PR #2, opened 2019-01-17
CRUD operations for snowflake databases

ryanking:
This will CRUD snowflake databases.

Auth is currently username + password only. Will look into okta later.


pkg/provider/provider.go
```diff
@@ -18,14 +26,68 @@ func Provider() *schema.Provider {
...
+
+func DSN(s *schema.ResourceData) (string, error) {
+	account := s.Get("account").(string)
+	username := s.Get("username").(string)
+	password := s.Get("password").(string)
+	region := s.Get("region").(string)
+	role := s.Get("role").(string)
+
+	// us-west-2 is their default region, but if you actually specify that it won't trigger their default code
+	//  https://github.com/snowflakedb/gosnowflake/blob/52137ce8c32eaf93b0bd22fc5c7297beff339812/dsn.go#L61
```
edulop91: :cry:

pkg/provider/provider.go
```diff
@@ -18,14 +26,68 @@ func Provider() *schema.Provider {
...
+		},
 		DataSourcesMap: map[string]*schema.Resource{},
+		ConfigureFunc:  ConfigureProvider,
+	}
+}
+
+func ConfigureProvider(s *schema.ResourceData) (interface{}, error) {
+	dsn, err := DSN(s)
+
+	log.Printf("[DEBUG] connecting to %s", dsn)
```
edulop91: why not logrus.Debug?

pkg/provider/provider_test.go
```diff
@@ -4,12 +4,64 @@ import (
 	"testing"
 
 	"github.com/chanzuckerberg/terraform-provider-snowflake/pkg/provider"
+	"github.com/hashicorp/terraform/helper/schema"
+	_ "github.com/snowflakedb/gosnowflake"
```
edulop91: just curious, what is being init'd here?

pkg/resources/database.go
```diff
@@ -0,0 +1,145 @@
...
+func (d *database) Create(data *schema.ResourceData, meta interface{}) error {
+	name := data.Get("name").(string)
+	comment := data.Get("comment").(string)
+	db := meta.(*sql.DB)
+
+	// TODO escape name
+	// TODO escape comment
+	// TODO name appears to get normalized to uppercase, should we do that? or maybe just consider it
+	// 	case-insensitive?
+	stmt := fmt.Sprintf("CREATE DATABASE %s COMMENT='%s'", name, comment)
```
edulop91: would this allow people to do terrible things?

pkg/resources/database.go
```diff
@@ -0,0 +1,145 @@
...
+	log.Printf("[DEBUG] stmt %s", stmt)
+	_, err := db.Exec(stmt)
+	if err != nil {
+		return errors.Wrapf(err, "error dropping database %s", name)
+	}
+
+	return nil
+}
+
+func (d *database) Update(data *schema.ResourceData, meta interface{}) error {
```
edulop91: neat

pkg/provider/provider.go
ryanking: That would work if I can get the format right (has to start with "[DEBUG]" for terraform to pass it through correctly.

pkg/provider/provider.go
ryanking: yeah that was an hour of my life I want back

pkg/provider/provider_test.go
ryanking: The sql driver for snowflake. https://github.com/snowflakedb/gosnowflake/blob/master/driver.go#L141

pkg/resources/database.go
ryanking: Without escaping, yes.

pkg/provider/provider.go
ryanking: I think I will leave it as-is for now.

pkg/resources/database.go
ryanking: I will follow up with a PR just on escaping. Need to research more.

pkg/provider/provider.go
edulop91: ah, interesting

PR #4, opened 2019-01-23
[feature] minimal warehouse support

ryanking:
Only a subset of fields so far, more to come.


pkg/resources/warehouse.go
```diff
@@ -0,0 +1,150 @@
...
+func ValidateWarehouseName(val interface{}, key string) ([]string, []error) {
+	return snowflake.ValidateIdentifier(val)
+}
+
+func (w *warehouse) Create(data *schema.ResourceData, meta interface{}) error {
+	name := data.Get("name").(string)
+	comment := data.Get("comment").(string)
+	db := meta.(*sql.DB)
+
+	err := DBExec(db, "CREATE WAREHOUSE %s COMMENT='%s", name, comment)
```
edulop91: this might be missing a `'` around the comment

pkg/resources/warehouse.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (w *warehouse) Delete(data *schema.ResourceData, meta interface{}) error {
+	db := meta.(*sql.DB)
+	name := data.Get("name").(string)
+
+	err := DBExec(db, "DROP WAREHOUSE %s", name)
+	if err != nil {
+		return errors.Wrapf(err, "error dropping warehouse %s", name)
+	}
+
```
edulop91: do you also have to `set.ID == ""`? Not sure

pkg/resources/warehouse.go
ryanking: indeed

pkg/resources/warehouse.go
ryanking: yeah I don't know i will look

ryanking: I don't see any examples of the aws provider doing this, so I assume its not needed.

codecov.yml
```diff
@@ -0,0 +1,26 @@
+codecov:
```
edulop91: i wonder if we should start a fogg for go repos

pkg/resources/warehouse.go
```diff
@@ -0,0 +1,202 @@
...
+	return &warehouse{}
+}
+
+func (w *warehouse) Create(data *schema.ResourceData, meta interface{}) error {
+	db := meta.(*sql.DB)
+	name := data.Get("name").(string)
+
+	var sb strings.Builder
+
+	sb.WriteString(fmt.Sprintf("CREATE WAREHOUSE %s", name))
```
edulop91: err being ignored

pkg/resources/warehouse.go
```diff
@@ -0,0 +1,202 @@
...
+	err = data.Set("comment", comment.String)
+	if err != nil {
+		return err
+	}
+
+	err = data.Set("warehouse_size", size.String)
+	if err != nil {
+		return err
+	}
+	return nil
```
edulop91: could just return `err` and save a branch

PR #7, opened 2019-01-24
database acceptance tests

ryanking:


pkg/resources/database_acceptance_test.go
```diff
@@ -0,0 +1,76 @@
+package resources_test
```
edulop91: you can also add annotations such as `// +build acceptance` and then `go test ... -tags=acceptance`

ryanking: Yeah, though I think I'll also have to set an env variable, based on https://github.com/hashicorp/terraform/blob/master/helper/resource/testing.go#L448

PR #11, opened 2019-01-28
ability to create roles

ryanking:


pkg/resources/role.go
```diff
@@ -0,0 +1,164 @@
...
+		Create: CreateRole,
+		Read:   ReadRole,
+		Delete: DeleteRole,
+		Update: UpdateRole,
+
+		Schema: map[string]*schema.Schema{
+			"name": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: func(val interface{}, key string) ([]string, []error) {
```
edulop91: maybe these inline functions should be de duped from the other resources

pkg/resources/role.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = data.Set("name", name.String)
+	if err != nil {
+		return err
+	}
+	err = data.Set("comment", comment.String)
+	if err != nil {
```
edulop91: could get rid of this `if` and save a branch

PR #13, opened 2019-01-31
RFC for granting priviledges in snowflake

ryanking:
I am trying to figure out the resource schemas for granting privileges in snowflake.

Below is a doc where I have some examples of how it *could* work. Please give feedback.

### DIFF
--- a/examples/priviledges/main.tf
+++ b/examples/priviledges/main.tf
@@ -1,0 +1,192 @@
+# NOTE that this is an RFC for how things *could* work.
+
+# This is the schema for granting priviledges
+# https://docs.snowflake.net/manuals/sql-reference/sql/grant-privilege.html
+# 
+# GRANT {  { globalPrivileges | ALL [ PRIVILEGES ] } ON ACCOUNT
+#        | { accountObjectPrivileges | ALL [ PRIVILEGES ] } ON { RESOURCE MONITOR | WAREHOUSE | DATABASE } <object_name>
+#        | { schemaPrivileges | ALL [ PRIVILEGES ] } ON { SCHEMA <schema_name> | ALL SCHEMAS IN DATABASE <db_name> }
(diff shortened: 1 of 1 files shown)


examples/priviledges/main.tf
```diff
@@ -0,0 +1,192 @@
...
+
+  schema {
+    schema_name = "bar"
+
+    # these two ^ v are mutually exclusive
+    # database_name = "bar"
+    #  if ^ then it means "all schemas in db"
+
+    # not sure if we should make these mutually exclusive or allow you to create one resource w/ a bunch of privs
+    table {
```
edulop91: not sure if you can have nested blocks like these in the current version of tf

ryanking: i see an example here https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html

edulop91: nvm, you can

examples/priviledges/main.tf
```diff
@@ -0,0 +1,192 @@
...
+#     USAGE
+# -- For FILE FORMAT, UDF, or SEQUENCE
+#     USAGE
+
+# Given that schema, here are some terraform resources we could support. We should map this out semi-fully, even though
+# we will build incrementally.
+
+# set global privs
+resource "snowflake_priviledge_grants" "a" {
+  name      = "foo1"
```
edulop91: does this name do anything other than provide an ID to track tf state?

ryanking: Just state, because there is no natural ID otherwise.

austinylin:
A few thoughts:

- Thanks for starting this!
- s/priviledges/privileges/g
- Just wanted to check my understanding, this is just how we assign privileges to roles? Users -> roles is another exercise?
- I think generally on the questions about how many permissions to enable to be nested in a single grant we should drive towards simplicity which to me would say let's make it easy to grant a ton of access in a single resource. I think that generally means we end up with one resource per role which is cognitively nice.

PR #24, opened 2019-02-13
cleanup resources

ryanking:
This consolidates a bunch of copy-pasta. For the straightforward CRUD resources we are able to share implementation for Create, Update and Delete. Read is separate for now because I haven't thought of a good way to implement without reflection.


pkg/resources/user.go
```diff
@@ -119,34 +118,10 @@ func User() *schema.Resource {
...
-				qb.SetBool(field, valBool)
-			}
-		}
-	}
-	err := DBExec(db, qb.Statement())
-
-	if err != nil {
-		return errors.Wrap(err, "error creating user")
-	}
+// func DeleteResource(t string, builder func(string) *snowflake.Builder) func(*schema.ResourceData, interface{}) error {
```
edulop91: rm

PR #25, opened 2019-02-13
need to close sql.Rows

ryanking:


pkg/resources/role_grants.go
```diff
@@ -139,6 +139,7 @@ func readGrants(db *sql.DB, roleName string) ([]*grant, error) {
 	stmt := fmt.Sprintf(`SHOW GRANTS OF ROLE "%s"`, roleName)
 	log.Printf("[DEBUG] stmt %s", stmt)
 	rows, err := sdb.Queryx(stmt)
+	defer rows.Close()
```
edulop91: Wonder if rows.Close() will panic when there's an error

PR #28, opened 2019-04-27
Enable support for externalbrowser Authenticator

austinylin:
This PR adds supports the "externalbrowser" authenticator. In this mode the user will be redirected to their browser to authenticate. 

This works well for accounts with SAML, but is a tad silly since it pops a browser tab for each connection. Long term we should figure out how to cache credentials here.


pkg/provider/provider.go
```diff
@@ -30,10 +30,16 @@ func Provider() *schema.Provider {
 			},
 			"password": &schema.Schema{
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_PASSWORD", nil),
 				Sensitive:   true,
 			},
+			"browser_auth": &schema.Schema{
```
ryanking: Both `password` and `browser_auth` should have [ConflictsWith](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/a19745031a0bd81ff66556212bfb3c7ebec730db/vendor/github.com/hashicorp/terraform/helper/schema/schema.go#L170) set to make it clear that you can't use both at the same time.

Issue #33, opened 2019-06-07
New resource: reader account


sjauld:
[Reference](https://docs.snowflake.net/manuals/user-guide/data-sharing-reader-create.html)

Reader accounts (formerly known as “read-only accounts”) enable providers to share data with consumers who are not already Snowflake customers, without requiring the consumers to become Snowflake customers.

PR #35, opened 2019-06-09
Adds a managed account resource (#33)

sjauld:
I'm waiting for Snowflake to get back to me re: a test account, so I haven't run the acceptance tests on this one yet, but I thought I'd submit it for feedback in the interim.


pkg/resources/managed_account.go
```diff
@@ -0,0 +1,193 @@
...
+		return err
+	}
+
+	if isReader {
+		err = data.Set("type", "READER")
+		if err != nil {
+			return err
+		}
+	} else {
+		return fmt.Errorf("is_reader should have been true but was %t", isReader)
```
ryanking: Unless I am missing something, this looks like it would break if and when snowflake introduced non-reader accounts?

pkg/validation/validation.go
```diff
@@ -0,0 +1,74 @@
...
+	asciiA = 65
+	asciiZ = 90
+	asciia = 97
+	asciiz = 122
+)
+
+// ValidateAccount checks that your account name is valid
+//
+// the identifier must start with an alphabetic character and cannot contain spaces or special characters unless
+// the entire identifier string is enclosed in double quotes (e.g. "My object").
```
ryanking: We (try to) always use the double-quote form, meaning that these validations are not relevant.

go.sum
```diff
@@ -169,6 +169,7 @@ github.com/hashicorp/memberlist v0.1.0/go.mod h1:ncdBp14cuox2iFOq3kDiquKU6fqsTBc
 github.com/hashicorp/serf v0.0.0-20160124182025-e4ec8cc423bb/go.mod h1:h/Ru6tmZazX7WO/GDmwdpS975F019L4t5ng5IgwbNrE=
 github.com/hashicorp/terraform v0.12.0 h1:It2vmod2dBMB4+r+aUW2Afx0HlftyUwzNsNH3I2vrJ8=
 github.com/hashicorp/terraform v0.12.0/go.mod h1:Ke0ig9gGZ8rhV6OddAhBYt5nXmpvXsuNQQ8w9qYBZfU=
+github.com/hashicorp/terraform v0.12.1 h1:A56pUiGbDZ372cvDKWixzVciTGZkqZ01hH58GTNQ4lg=
```
ryanking: We only recently switch from dep to go modules, so I my knowledge is not strong here, but it seems that having two different versions of hashicorp/terraform here is sub-optimal.

ryanking:
Current test failure–

```
=== RUN   TestAccManagedAccount
[DEBUG] sql-conn-exec [query CREATE MANAGED ACCOUNT "xezwtumzwk" ADMIN_NAME='ciwijpkjlb' ADMIN_PASSWORD='A1kotaqqisxm' COMMENT='Created by a Terraform acceptance test' NAME='xezwtumzwk' TYPE='READER' args []driver.NamedValue{ } err 001420 (22023): SQL compilation error: invalid property 'NAME' for 'MANAGED_ACCOUNT'] 
--- FAIL: TestAccManagedAccount (0.36s)
    testing.go:568: Step 0 error: errors during apply:
        
        Error: error creating this does not seem to be used: 001420 (22023): SQL compilation error:
        invalid property 'NAME' for 'MANAGED_ACCOUNT'
        
          on /var/folders/0q/52rv8xtd7mjcgljsgmbkk5yr0000gq/T/tf-test849392996/main.tf line 2:
          (source code not available)
        
```

ryanking:
Current test failure–

```
=== RUN   TestAccManagedAccount
[DEBUG] sql-conn-exec [query CREATE MANAGED ACCOUNT "xezwtumzwk" ADMIN_NAME='ciwijpkjlb' ADMIN_PASSWORD='A1kotaqqisxm' COMMENT='Created by a Terraform acceptance test' NAME='xezwtumzwk' TYPE='READER' args []driver.NamedValue{ } err 001420 (22023): SQL compilation error: invalid property 'NAME' for 'MANAGED_ACCOUNT'] 
--- FAIL: TestAccManagedAccount (0.36s)
    testing.go:568: Step 0 error: errors during apply:
        
        Error: error creating this does not seem to be used: 001420 (22023): SQL compilation error:
        invalid property 'NAME' for 'MANAGED_ACCOUNT'
        
          on /var/folders/0q/52rv8xtd7mjcgljsgmbkk5yr0000gq/T/tf-test849392996/main.tf line 2:
          (source code not available)
        
```

go.sum
sjauld: This is just the checksum, as I understand it the version used will still be from go.mod, it must have just pulled this in since I have it installed. There are a few other dependencies (e.g. aws-sdk) that also have multiple versions in the go.sum file.

pkg/resources/managed_account.go
sjauld: Yes, when more managed account types come in we would have to add them to the "type" field and add some logic here, something like:
```go
if isReader {
  // reader stuff
} else if isNewTypeOfAccount {
  // newType stuff
} else {
  // fail
}
```

pkg/validation/validation.go
sjauld: I originally read this as 
```
the identifier must start with an alphabetic character && 
(! contain spaces or special characters || the entire identifier string is enclosed in double quotes)
```

But I just checked and "1test" is valid, so yes I'll go ahead and remove this.

pkg/resources/managed_account_test.go
```diff
@@ -0,0 +1,48 @@
...
+		"name":           "test-account",
+		"admin_name":     "bob",
+		"admin_password": "abc123ABC",
+		"comment":        "great comment",
+	}
+	d := schema.TestResourceDataRaw(t, resources.ManagedAccount().Schema, in)
+	a.NotNil(d)
+
+	WithMockDb(t, func(db *sql.DB, mock sqlmock.Sqlmock) {
+		mock.ExpectExec(`^CREATE MANAGED ACCOUNT "test-account" ADMIN_NAME='bob' ADMIN_PASSWORD='abc123ABC' COMMENT='great comment' NAME='test-account' TYPE='READER'$`).WillReturnResult(sqlmock.NewResult(1, 1))
```
sjauld: This expectation is bad, I'll need to fix it.

ryanking:
acceptance test failure now–

```
--- FAIL: TestAccManagedAccount (3.24s)
    testing.go:568: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=10) "admin_name": (string) (len=10) "ntqolfcjxt"
        }
        
```

sjauld:
Damn - I'll try to get that acceptance test account up before fixing again. I think the Snowflake team are all recovering from the summit!

sjauld:
Ok, I managed to get a testing account up and running, I've run the acceptance tests now and they pass.

Issue #36, opened 2019-06-11
New resource: view


sjauld:
[Reference](https://docs.snowflake.net/manuals/user-guide/views-introduction.html)

A view allows the result of a query to be accessed as if it were a table.

PR #40, opened 2019-06-17
[feature] new resource: view

sjauld:
#36

I'm still having trouble with acceptance tests but I thought I'd open this because there are some other changes that may require discussion.

Fundamentally, I don't see any reason why we should be interpolating SQL statements in this provider. The approach taken in this resource allows us to pass arguments through to the SQL driver. I believe this is safer and cleaner, although please correct me if there's a reason for not doing this with the Snowflake driver.

## resources.DBExec 

I moved this function to `resources.go` since it's a generic function and didn't seem to belong in warehouse.go. I also modified it so that SQL arguments are passed through to the SQL driver instead of being interpolated by fmt.Sprintf.

## view

The view resource didn't fit the pattern of the generic Builder, so I created a new standalone ViewBuilder. Instead of using fmt.Sprintf to interpolate SQL queries, this builder spits out statements and arguments that are consumed by the SQL driver. 

## random stuff

There are probably some minor cosmetic changes from my previous PRs, sorry about that


sjauld:
Upon further investigation, it seems that the go snowflake driver doesn't support binding parameters. I'll rewrite as per the current implementation and raise a ticket with Snowflake

PR #63, opened 2019-08-09
fix build

### LINKED ISSUE #62: fix build

ryanking:
Temporarily turn off a few import tests that are not working. This will get us back to a green build from which we can fix things one-by-one.

I didn't want to disable the functionality because it may actually be working and in use by folks like @sjauld.

Fixes #62


Makefile
```diff
@@ -16,7 +16,7 @@ setup: ## setup development dependencies
 .PHONY: setup
 
 lint: ## run the fast go linters
-	golangci-lint run
+	./bin/golangci-lint run
```
edulop91: why this one?

ryanking: Yeah this doesn't actually affect the build, but I noticed that if you use `make setup` to install the dependencies it will put them in `./bin`, so this won't work.

pkg/resources/share.go
```diff
@@ -146,7 +146,8 @@ func ReadShare(data *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	accs := strings.Split(to.String, ", ")
+	// accs := strings.Split(to.String, ", ")
```
edulop91: should this be rmd?

edulop91: hmm, on second thought, why not keep strings.Split but change ", " to ","?

ryanking: yeah will do later

PR #70, opened 2019-08-29
switch from fmt to log for DEBUG

ryanking:
It seems that fmt.Printf does not make it through from the plugin to terraform's output.


pkg/resources/grant_helpers.go
```diff
@@ -3,6 +3,7 @@ package resources
 import (
 	"database/sql"
 	"fmt"
+	"log"
```
ryanking:
```suggestion
"notlog"
```

PR #73, opened 2019-09-09
Fix a small connection leak

pwnage101:
This fixes a connection leak which causes open connections to accumulate
for each schema that gets refreshed.  This probably doesn't manifest as
a user-facing bug, but it certainly can't hurt to merge a fix.

For an explanation on how I discovered and tested this, see the
following github issue comment:

https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/69#issuecomment-528959865

### DIFF
--- a/pkg/resources/schema.go
+++ b/pkg/resources/schema.go
@@ -264,11 +264,13 @@ func SchemaExists(data *schema.ResourceData, meta interface{}) (bool, error) {
 		return false, err
 	}
 
+	exists := false
 	if rows.Next() {
-		return true, nil
+		exists = true
 	}
 
-	return false, nil
+	rows.Close()
+	return exists, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/resources/schema.go
```diff
@@ -264,11 +264,13 @@ func SchemaExists(data *schema.ResourceData, meta interface{}) (bool, error) {
 		return false, err
 	}
 
+	exists := false
```
ryanking: Seems like it would be simpler to just use `defer rows.Close()`?

pwnage101: Okay, I'll update tomorrow.

ryanking: actually I did a pass to add this everywhere that I think is relevant- https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/74

PR #78, opened 2019-09-12
[feature] future object grants and add the table_grant resource

pwnage101:
Fixes https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/76

### DIFF
--- a/pkg/resources/view_grant.go
+++ b/pkg/resources/view_grant.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/hashicorp/terraform/helper/schema"
 	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/pkg/errors"
 
 	"github.com/chanzuckerberg/terraform-provider-snowflake/pkg/snowflake"
 )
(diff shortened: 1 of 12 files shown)


pwnage101:
@ryanking do you think this is going generally in the right direction?  I still haven't implemented table_grants yet, which is the main reason I'm doing this (to support future table grants), but this sets up most of the scaffolding for code and tests.

ryanking:
@pwnage101 I would prefer to add an on_future parameter to the existing grants objects

pwnage101:
@ryanking that's what I did, is it not?

https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/78/files#diff-4451f5a17d8648591cf4c1b4d6a23d48R57-R64

or, maybe you're referring to the naming of the parameter (I named it future_views for the view_grant resource)?

ryanking:
@pwnage101 yeah sorry, I didn't realize that the code had been updated. Will review today.

pkg/resources/view_grant.go
```diff
@@ -50,8 +51,16 @@ var viewGrantSchema = map[string]*schema.Schema{
 		Type:        schema.TypeSet,
 		Elem:        &schema.Schema{Type: schema.TypeString},
 		Optional:    true,
-		Description: "Grants privilege to these shares.",
+		Description: "Grants privilege to these shares (only valid if future_views = false).",
+		ForceNew:    true,
+	},
+	"future_views": &schema.Schema{
```
ryanking: I think we should name this "on_future" so that it maps more clearly to what SQL will be run.

pwnage101: The other option I was considering was to not add a bool field, and just let an empty view_name imply future views.  Just a thought, and isn't too hard to switch back and forth, but we need to come to a decision before merging and releasing this.

ryanking: I would worry a bit about that implicit semantic. I like having the boolean to make it more clear what is happening.

pkg/resources/view_grant_acceptance_test.go
```diff
@@ -9,7 +9,34 @@ import (
...
+				),
+			},
+			// IMPORT
+			{
+				ResourceName:      "snowflake_view_grant.test",
+				ImportState:       true,
+				ImportStateVerify: true,
+				// I don't understand why ImportStateVerifyIgnore is needed, but
+				// adding this is the only way to make acceptance tests pass.
+				ImportStateVerifyIgnore: []string{"roles", "shares"},
```
pwnage101: I had to add this to acceptance tests to git them to pass.  Acceptance tests for view grants also fail on master for the same reason (when not skipped).  As stated, I don't truly understand what's going on here and can't find additional documentation about ImportStateVerify beyond this section:

https://www.terraform.io/docs/extend/resources/import.html#importstateverifyignore-1

pwnage101:
@ryanking This should be ready for official review.  In summary I added the on_future field to snowflake_view_grant, and also created the snowflake_table_grant resource (also supporting on_future).  Unfortunately, I wasn't able to add acceptance tests for snowflake_table_grant because I would also needed to create the snowflake_table resource.

pwnage101:
BTW, for running acceptance tests, I just ran them against my own snowflake instance.  I also ran unit tests on my laptop.  Everything as-written is passing, with the caveat about ImportStateVerifyIgnore which I commented about above.

PR #81, opened 2019-09-23
Add support for granting ownership for schemas

tail:


pwnage101:
Doesn't granting ownership relinquish any rights to manage the resource with terraform?  If so, that seems like a caveat worth mentioning in the docs or field description.

tail:
> Doesn't granting ownership relinquish any rights to manage the resource with terraform? If so, that seems like a caveat worth mentioning in the docs or field description.

That's a good point and something I noted during testing.  The grant should include the the role that is running Terraform.  I can add a note in the README under `snowflake_schema_grant`.

pwnage101:
Keep in mind also the section of the README you modified is auto-generated so you will need to copy that description into the field definition as well, and run `make docs`

kylelundstedt:
Hi all

I built a local copy of the provider to use the current master branch code and get the ability to grant ownership for schemas.  Just as a heads up, if you change the privilege for an existing schema from, say, "USAGE" for roleA to "OWNERSHIP" for role B, TF fails during apply with an error like:

```
Error: 003036 (23001): SQL execution error: Dependent grant of privilege 'CREATE EXTERNAL TABLE' on securable 'DIYOTTA.FISERV_PROCESSED' to role 'DIYOTTA_LOADER' exists.  It must be revoked first.  More than one dependent grant may exist: use 'SHOW GRANTS' command to view them.  To revoke all dependent grants while transferring object ownership, use convenience command 'GRANT OWNERSHIP ON <target_objects> TO <target_role> REVOKE CURRENT GRANTS'.
```

So, it appears that some form of "REVOKE CURRENT GRANTS" is need to change ownership of an existing schema using the `snowflake_schema_grant` resource.  I don't know Go, so I apologize for not making this a pull request.  Thanks!

PR #85, opened 2019-10-03
[fix] removing blind | delimiting in Grant Creation

kuannie1:


pkg/resources/view_grant.go
```diff
@@ -107,14 +110,23 @@ func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {
...
-	// ID format is <db_name>|<schema_name>|<view_name>|<privilege>
-	// view_name is empty when on_future = true
-	if futureViews {
-		data.SetId(fmt.Sprintf("%v|%v||%v", dbName, schemaName, priv))
-	} else {
-		data.SetId(fmt.Sprintf("%v|%v|%v|%v", dbName, schemaName, viewName, priv))
+	data_identifiers := [][]string{{dbName, schemaName, viewName, ""}}
+
+	if !(futureViews) { // merge with above if/else statement if possible
+		data_identifiers[0][3] = priv
```
edulop91: seems like the thing that changes is if the `viewName` is present or not. but priv is always there

pkg/resources/view_grant.go
```diff
@@ -107,14 +110,23 @@ func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {
...
 	}
 
+	var buf bytes.Buffer
+	csv_writer := csv.NewWriter(&buf)
```
edulop91: in general, use camelCase for Go
```suggestion
	csvWriter := csv.NewWriter(&buf)
```

pkg/resources/view_grant.go
```diff
@@ -107,14 +110,23 @@ func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {
...
+	csv_writer.Comma = '|'
+	csv_writer.WriteAll(data_identifiers)
+
+	if err := csv_writer.Error(); err != nil {
+		log.Fatalln("error writing csv:", err)
```
edulop91: this will abruptly exit your code and call something like `os.Exit(1)`. https://github.com/golang/go/blob/go1.13.1/src/log/log.go#L329:6

edulop91: This function allows you to return an error `func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {`
so you can just `return err` and have it be someone else's problem to deal with

pkg/resources/view_grant.go
```diff
@@ -107,14 +110,23 @@ func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {
...
+	}
+
+	data.SetId(strings.TrimSuffix(buf.String(), "\n"))
```
edulop91: does it matter that newline is there?

pkg/resources/view_grant.go
kuannie1: @edulop91  

Getting a little frustrated about knowing how to format data_identifiers :/ I'm not sure if I should add priv to the end of dataIdentifiers (new variable name now)?

It's hard for me to make that variable adaptable. This is what I have so far:
```
	dataIdentifiers := make([][]string, 1, 4)
	dataIdentifiers[0][0] = dbName
	dataIdentifiers[0][1] = schemaName
	dataIdentifiers[0][2] = viewName
	dataIdentifiers[0][3] = priv
```
Everything else is the same.

pkg/resources/grant_helpers.go
```diff
@@ -54,12 +55,20 @@ type grant struct {
...
+
+	reader := csv.NewReader(strings.NewReader(v))
+	reader.Comma = '|'
+
+	lines, err := reader.ReadAll()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if len(lines) != 4 {
```
edulop91: lines is a `[][]string` I think you should check that `len(lines) == 1 && len(lines[0]) == 4`

pkg/resources/grant_helpers.go
```diff
@@ -54,12 +55,20 @@ type grant struct {
...
 func splitGrantID(v string) (string, string, string, string, error) {
-	arr := strings.Split(v, "|")
-	if len(arr) != 4 {
+
+	reader := csv.NewReader(strings.NewReader(v))
+	reader.Comma = '|'
+
+	lines, err := reader.ReadAll()
+	if err != nil {
+		log.Fatal(err)
```
edulop91: log.Fatal will call os.Exit, which we don't want. splitGrant allows you to return an error, so you could do `return "", "", "", "", err` and let the caller deal with the issue instead of crashing the program

pkg/resources/grant_helpers.go
```diff
@@ -54,12 +55,21 @@ type grant struct {
...
+	fmt.Printf("[DEBUG] splitGrantID input: %s", v)
+	reader := csv.NewReader(strings.NewReader(v))
+	reader.Comma = '|'
+
+	lines, err := reader.ReadAll()
+	if err != nil {
+		return "", "", "", "", err
+	}
+
+	if (len(lines) == 1) && (len(lines[0]) == 4) {
```
edulop91: I would do this the other way around, have the failure error up front and error out early if so.

```
if (len(lines) != 1 || len(lines[0]) !=4) {
    return """""', err
}
```

this reduces the cognitive load on the readers of the function since the further you go the less conditions you have to remember

pkg/resources/view_grant.go
```diff
@@ -107,14 +108,24 @@ func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	// ID format is <db_name>|<schema_name>|<view_name>|<privilege>
-	// view_name is empty when on_future = true
-	if futureViews {
-		data.SetId(fmt.Sprintf("%v|%v||%v", dbName, schemaName, priv))
-	} else {
-		data.SetId(fmt.Sprintf("%v|%v|%v|%v", dbName, schemaName, viewName, priv))
+	dataIdentifiers := make([][]string, 1)
```
edulop91: we got rid of a conditional that says  if `futureViews ==true` then `viewName` must be an empty string

edulop91: how it was coded before it would look like 
```
if futureViews {
  viewName = ""
}
```

kuannie1: Oh, so we're supposed to include that conditional?

edulop91: but you can also just assert that 
```
if futureViews && (viewName != "") {
    // error out since this looks like an error case
}
```

pkg/resources/view_grant.go
```diff
@@ -107,14 +108,24 @@ func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {
...
+	dataIdentifiers[0] = make([]string, 4)
+	dataIdentifiers[0][0] = dbName
+	dataIdentifiers[0][1] = schemaName
+	dataIdentifiers[0][2] = viewName
+	dataIdentifiers[0][3] = priv
+
+	var buf bytes.Buffer
+	csvWriter := csv.NewWriter(&buf)
+	csvWriter.Comma = '|'
+	csvWriter.WriteAll(dataIdentifiers) // calls Flush internally
```
edulop91:
```suggestion
	csvWriter.WriteAll(dataIdentifiers)
```
doesn't seem relevant/actionable to the users of this code so maybe we should not present them with this comment

pkg/resources/view_grant.go
```diff
@@ -107,14 +108,24 @@ func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	// ID format is <db_name>|<schema_name>|<view_name>|<privilege>
```
edulop91: we probably want to keep this documentation around

pkg/resources/view_grant.go
kuannie1: Can I add that line to the existing futureViews conditional that defines the builder?

```
	var builder snowflake.GrantBuilder
	if futureViews {
		builder = snowflake.FutureViewGrant(dbName, schemaName)
                viewName = ""
	} else {
		builder = snowflake.ViewGrant(dbName, schemaName, viewName)
	}
```

edulop91: I might do 
```
if futureViews && (viewName != "") {
    // error out since this looks like an error case
}
```
all the way at the top where you are doing validation, before any of the logic/processing

edulop91: since it is an invariant you want to hold throughout your function

edulop91:
@kuannie1 one last thing, I would change the pr title to be more descriptive of the issue you're fixing!

edulop91:
thanks for working on this, looks good!

edulop91:
waiting until we can fix the tests

edulop91:
@kuannie1 there are a bunch of other resources that create grants and have pipe-delimited ids. probably also worth looking into those!

PR #88, opened 2019-10-09
[feature] keypair auth

kuannie1:


kuannie1:
## INSTRUCTIONS
### Generating the snowflake keys
_Reference: https://docs.snowflake.net/manuals/user-guide/python-connector-example.html#using-key-pair-authentication_
1. Generate the unencrypted version of the private key. I would generate it in the ~/.ssh directory
`.ssh (you)$ openssl genrsa -out snowflake_key 2048`
2. After 1. is done, generate the public key by referencing the private key:
`openssl rsa -in snowflake_key -pubout -out snowflake_key.pub` 
3. Copy your public key (between the `BEGIN PUBLIC KEY` and `END PUBLIC KEY` lines) and paste it into this template inside your user entry in user_map in snowflake-infra/terraform/accounts/czitest/users.tf.
```
locals{
  user_map = {
    "(your username)" = {
        rsa_public_key = "(paste your public key here)"
    }
  }
}
```
4. In your terminal, run `make plan` in the snowflake-infra/terraform/accounts/czitest/ directory. If you see your public key in the plan, run `make apply`.

pkg/provider/provider.go
```diff
@@ -101,22 +110,59 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+	var dsn string
+	var err error
+
+	if len(pathPrivateKey) != 0 {
+
+		privateKeyFilename := "/Users/anneku/.ssh/snowflake_key"
+
+		// reading the private key
+		privateKeyBytes, err := ioutil.ReadFile(privateKeyFilename)
+		if err != nil || len(privateKeyBytes) == 0 { // both conditionals had to be false in order to get past this
```
edulop91: I would actually split these into two separate contidionals that way each can get a more specific error message. 
Also, go just released a feature to "wrap" error messages. so for errors, instead of %s you can do %w (to **w**rap)

pkg/provider/provider.go
```diff
@@ -101,22 +110,59 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+			return "", errors.Errorf("Could not read private key: %s", err)
```
edulop91: same here with the %w

pkg/provider/provider.go
```diff
@@ -101,22 +110,59 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+		}
+
+		// reads and unmarshals a private key
+		privateKey, err := ssh.ParseRawPrivateKey(privateKeyBytes)
+		if err != nil {
+			return "", errors.Errorf("Could not parse private key: %s", err)
```
edulop91: and here

pkg/provider/provider.go
```diff
@@ -101,22 +110,59 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+		dsn, err = gosnowflake.DSN(&gosnowflake.Config{
+			Account:       account,
+			User:          username,
+			Region:        region,
+			Role:          role,
+			PrivateKey:    rsaPrivateKey,
+			Authenticator: gosnowflake.AuthTypeJwt,
+		})
+
+		if err != nil {
```
edulop91: you can probably get rid of this since at the end we're returning dsn, error

pkg/provider/provider.go
```diff
@@ -36,7 +38,14 @@ func Provider() *schema.Provider {
...
 				Sensitive:     false,
-				ConflictsWith: []string{"password"},
+				ConflictsWith: []string{"password", "path_private_key"},
+			},
+			"path_private_key": &schema.Schema{
+				Type:          schema.TypeString,
+				Optional:      true,
+				DefaultFunc:   schema.EnvDefaultFunc("SNOWFLAKE_PRIVATE_KEY_PATH", nil),
+				Sensitive:     true,
+				ConflictsWith: []string{"browser_auth"},
```
edulop91: also conflicts with password I think

pkg/provider/provider.go
```diff
@@ -72,9 +81,8 @@ func Provider() *schema.Provider {
 }
 
 func ConfigureProvider(s *schema.ResourceData) (interface{}, error) {
-	dsn, err := DSN(s)
+	dsn, err := DSN(s) // got an error
```
edulop91: what's this `//got an error` comment?

kuannie1: oh, this was when I was trying to print something. The comment is outdated

pkg/provider/provider.go
```diff
@@ -101,22 +110,64 @@ func DSN(s *schema.ResourceData) (string, error) {
...
-		User:     username,
-		Region:   region,
-		Password: password,
-		Role:     role,
-	})
+	var dsn string
+	var err error
+
+	if len(pathPrivateKey) != 0 {
+		// JWT Authentication
```
edulop91:
```suggestion
		// JWT Authentication
		pathPrivateKey, err := homedir.Expand(pathPrivateKey)
		if err != nil{ bla bla}
```

pkg/provider/provider.go
```diff
@@ -101,23 +113,54 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+
+	} else {
+		config.Password = password
 	}
 
-	return dsn, err
+	return gosnowflake.DSN(&config)
+}
+
+func ParsePrivateKey(pathPrivateKey string) (rsa.PrivateKey, error) {
```
edulop91:
```suggestion
func ParsePrivateKey(pathPrivateKey string) (*rsa.PrivateKey, error) {
```

pkg/provider/provider.go
```diff
@@ -101,23 +113,54 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+	var err error
+
+	expandedPathPrivateKey, err := homedir.Expand(pathPrivateKey)
+
+	privateKeyBytes, err := ioutil.ReadFile(expandedPathPrivateKey)
+	if err != nil {
+		fmt.Printf("Could not read private key: %w\n", err)
+	}
+	if len(privateKeyBytes) == 0 {
+		fmt.Println("Private key is empty")
```
edulop91: return nil, fmt.Print...

pkg/provider/provider.go
```diff
@@ -101,23 +113,54 @@ func DSN(s *schema.ResourceData) (string, error) {
...
 	}
 
-	return dsn, err
+	return gosnowflake.DSN(&config)
+}
+
+func ParsePrivateKey(pathPrivateKey string) (rsa.PrivateKey, error) {
+	var err error
+
+	expandedPathPrivateKey, err := homedir.Expand(pathPrivateKey)
```
edulop91: check + return this error

pkg/provider/provider.go
```diff
@@ -29,14 +33,21 @@ func Provider() *schema.Provider {
...
 			},
 			"browser_auth": &schema.Schema{
 				Type:          schema.TypeBool,
 				Optional:      true,
 				DefaultFunc:   schema.EnvDefaultFunc("SNOWFLAKE_USE_BROWSER_AUTH", nil),
 				Sensitive:     false,
-				ConflictsWith: []string{"password"},
+				ConflictsWith: []string{"password", "path_private_key"},
+			},
+			"path_private_key": &schema.Schema{
```
ryanking: I think `"private_key_path"` sounds better.

pkg/provider/provider.go
```diff
@@ -89,9 +99,10 @@ func ConfigureProvider(s *schema.ResourceData) (interface{}, error) {
 
 func DSN(s *schema.ResourceData) (string, error) {
 	account := s.Get("account").(string)
-	username := s.Get("username").(string)
+	user := s.Get("username").(string)
 	password := s.Get("password").(string)
 	browserAuth := s.Get("browser_auth").(bool)
+	pathPrivateKey := s.Get("path_private_key").(string)
```
ryanking: rename this too, as above

pkg/provider/provider.go
```diff
@@ -101,23 +112,57 @@ func DSN(s *schema.ResourceData) (string, error) {
...
-		})
+	config := gosnowflake.Config{
+		Account: account,
+		User:    user,
+		Region:  region,
+		Role:    role,
 	}
 
-	return dsn, err
+	if len(pathPrivateKey) != 0 {
```
ryanking: Maybe I am missing something, but shouldn't this be `!= nil`?

kuannie1: I thought so too, but when I tried your suggestion I got this:
```
# github.com/chanzuckerberg/terraform-provider-snowflake/pkg/provider
pkg/provider/provider.go:122:20: invalid operation: privateKeyPath != nil (mismatched types string and nil)
```

pkg/provider/provider.go
```diff
@@ -1,13 +1,17 @@
...
+	"crypto/rsa"
+	"io/ioutil"
 
 	"github.com/chanzuckerberg/terraform-provider-snowflake/pkg/db"
 	"github.com/chanzuckerberg/terraform-provider-snowflake/pkg/resources"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/pkg/errors"
 	"github.com/snowflakedb/gosnowflake"
+	"golang.org/x/crypto/ssh"
+
```
edulop91: rm extra line, imports should be 2 groups. stdlib and everything else

pkg/provider/provider.go
```diff
@@ -101,23 +112,57 @@ func DSN(s *schema.ResourceData) (string, error) {
...
-		})
+	config := gosnowflake.Config{
+		Account: account,
+		User:    user,
+		Region:  region,
+		Role:    role,
 	}
 
-	return dsn, err
+	if len(privateKeyPath) != 0 {
```
edulop91: you can also do 
```suggestion
	if privateKeyPath != "" {
```
since the 0 value for strings is `""`

pkg/provider/provider.go
```diff
@@ -101,23 +112,57 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+
+		rsaPrivateKey, err := ParsePrivateKey(privateKeyPath)
+		if err != nil {
+			return "", errors.Wrap(err, "Private Key could not be parsed")
+		}
+		config.PrivateKey = rsaPrivateKey
+		config.Authenticator = gosnowflake.AuthTypeJwt
+
+	} else if browserAuth {
+
```
edulop91:
```suggestion
```

pkg/provider/provider.go
```diff
@@ -101,23 +112,57 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+		config.Authenticator = gosnowflake.AuthTypeExternalBrowser
+
```
edulop91:
```suggestion
```

pkg/provider/provider.go
```diff
@@ -101,23 +112,57 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+	} else {
+		config.Password = password
+	}
+
+	return gosnowflake.DSN(&config)
+}
+
+func ParsePrivateKey(privateKeyPath string) (*rsa.PrivateKey, error) {
+	var err error
```
edulop91: you don't need this variable declaration here

pkg/provider/provider.go
```diff
@@ -101,23 +112,57 @@ func DSN(s *schema.ResourceData) (string, error) {
...
+	privateKey, err := ssh.ParseRawPrivateKey(privateKeyBytes)
+	if err != nil {
+		return nil, errors.Wrap(err, "Could not parse private key")
+	}
+
+	rsaPrivateKey, ok := privateKey.(*rsa.PrivateKey)
+	if !ok {
+		return nil, errors.New("privateKey not of type RSA")
+	}
+	return rsaPrivateKey, err
```
edulop91:
```suggestion
	return rsaPrivateKey, nil
```
at this point, err must be `nil` so lets make it explicit

edulop91:
oh, and one last thing!
can you add a section on the Readme that shows how to generate the ssh keys?
something like:

```
### SSH Authentication

you can generate your keys with:

cd ~/.ssh
openssl genrsa -out snowflake 4096
openssl rsa -in snowflake -pubout -out snowflake.pub
```

kuannie1:
@edulop91  I think the README.md is the only thing left for you to review.

PR #91, opened 2019-10-10
Test changing to resource type List

edulop91:


pkg/resources/role_grants.go
```diff
@@ -207,20 +208,34 @@ func UpdateRoleGrants(data *schema.ResourceData, meta interface{}) error {
...
+		default:
+			return fmt.Errorf("Role granted_to unrecognized type(%s)", grantedTo)
 		}
-		os := o.(*schema.Set)
-		ns := n.(*schema.Set)
+	}
+
+	type funcRoleGrant func(db *sql.DB, resource string, target string) error
+	x := func(resource string, grant funcRoleGrant, revoke funcRoleGrant, existingDataSet *schema.Set) error {
+		iNewData := data.Get(resource)
```
edulop91: This happens to be always empty when 
```hcl
resource "snowflake_role_grants" "this_does_not_work" {
  role_name = snowflake_role.foo.name
  roles     = ["roleA"]
  users = [
    "userA",
    "userB",
    snowflake_user.user["userC"].name
  ]
}
```

but work fine with something like
```hcl
resource "snowflake_role_grants" "this_works" {
  role_name = snowflake_role.foo.name
  roles     = ["roleA"]
  users = [
    "userA",
    "userB",
    "userC",
  ]
}
```

PR #95, opened 2019-10-16
Update schema_grant.go

sokser:
Added ALL grant for schemas

### DIFF
--- a/pkg/resources/schema_grant.go
+++ b/pkg/resources/schema_grant.go
@@ -12,6 +12,7 @@ import (
 // Intentionally exclude the "ALL" alias because it is not a real privilege and
 // might not interact well with this provider.
 var validSchemaPrivileges = []string{
+	"ALL",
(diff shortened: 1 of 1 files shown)


ryanking:
@sokser I didn't write the code, so I don't know for certain, but there is a comment suggesting that ALL would not work well with this provider -

https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/95/files#diff-8209b694a558a394c1d0170f39b0cc2eR12

pwnage101:
I wrote that comment because I tried the ALL privilege with `snowflake_warehouse_grant`, and **IIRC** it was completely broken.  Like, it would constantly recreate the resource even if nothing changed since it could not find a grant literally called "ALL" when querying current grants.  I think adding ALL is okay if you are adding logic to make the provider aware that "ALL" is actually an alias, and you add an acceptance test specifically for "ALL".

PR #98, opened 2019-10-21
[fix] Refactored pipe delimiting fields to use grantID struct

kuannie1:


pkg/resources/grant_helpers.go
```diff
@@ -52,24 +53,67 @@ type grant struct {
...
+	Privilege    string
+}
+
+func (grantID *grantID) String() (string, error) {
+	var buf bytes.Buffer
+	csvWriter := csv.NewWriter(&buf)
+	csvWriter.Comma = '|'
+	dataIdentifiers := [][]string{{grantID.ResourceName, grantID.SchemaName, grantID.ViewName, grantID.Privilege}}
+	err := csvWriter.WriteAll(dataIdentifiers)
+
```
edulop91:
```suggestion
```
adding this newline adds overhead for readers since the err assignment is visually separated from usage of it

pkg/resources/grant_helpers.go
```diff
@@ -52,24 +53,67 @@ type grant struct {
...
+	ResourceName string
+	SchemaName   string
+	ViewName     string
+	Privilege    string
+}
+
+func (grantID *grantID) String() (string, error) {
+	var buf bytes.Buffer
+	csvWriter := csv.NewWriter(&buf)
+	csvWriter.Comma = '|'
```
edulop91: I would proabably make '|' a constant since it is also used in the read function and they must both be the same

pkg/resources/grant_helpers.go
```diff
@@ -52,24 +53,67 @@ type grant struct {
...
+	reader.Comma = '|'
 	lines, err := reader.ReadAll()
 	if err != nil {
-		return "", "", "", "", err
+		return nil, err
 	}
 
 	if len(lines) != 1 || len(lines[0]) != 4 {
-		return "", "", "", "", fmt.Errorf("ID %v is invalid", v)
+		return nil, fmt.Errorf("ID %v is invalid", stringID)
```
edulop91:
```suggestion
		return nil, fmt.Errorf("ID %s is invalid", stringID)
```
%s for strings

pkg/resources/grant_helpers.go
```diff
@@ -52,24 +53,67 @@ type grant struct {
...
+	grantResult := &grantID{
+		ResourceName: lines[0][0],
+		SchemaName:   lines[0][1],
+		ViewName:     lines[0][2],
+		Privilege:    lines[0][3],
+	}
+	return grantResult, nil
 }
 
+// // splitGrantID takes the <db_name>|<schema_name>|<view_name>|<privilege> ID and
```
edulop91: remove all of this commented out code since not used anymore

pkg/resources/grant_helpers.go
kuannie1: Should I make it a global variable of some sort? Not sure how to define & place the constant

pkg/resources/grant_helpers.go
```diff
@@ -52,24 +53,67 @@ type grant struct {
...
+	err := csvWriter.WriteAll(dataIdentifiers)
+
+	if err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
+}
 
+func grantIDFromString(stringID string) (*grantID, error) {
```
edulop91: this function can probably use a comment like the one `splitGrantID` used to have
```
// splitGrantID takes the <db_name>|<schema_name>|<view_name>|<privilege> ID and
...
```

pkg/resources/grant_helpers_internal_test.go
```diff
@@ -7,52 +7,52 @@ import (
...
-	if priv != "privilege" {
-		t.Errorf("Expected priv to be privilege, got %v", priv)
+	if grant.Privilege != "privilege" {
+		t.Errorf("Expected priv to be privilege, got %v", grant.Privilege)
 	}
 
 	// Bad ID
 	id = "database|name-privilege"
-	_, _, _, _, err = splitGrantID(id)
+	_, err = grantIDFromString(id)
```
edulop91: I would also try and do an assert on the type, contents, or message of the error. That way you can double check it is the error you are expecting

pkg/resources/schema_grant.go
```diff
@@ -92,43 +90,52 @@ func CreateSchemaGrant(data *schema.ResourceData, meta interface{}) error {
 	}
 
 	// ID format is <db_name>|<schema_name>||<privilege>
```
edulop91: can probably remove this comment now

kuannie1: Should I remove all the comments like this?

pkg/resources/table_grant.go
```diff
@@ -116,21 +114,34 @@ func CreateTableGrant(data *schema.ResourceData, meta interface{}) error {
...
-		data.SetId(fmt.Sprintf("%v|%v||%v", dbName, schemaName, priv))
-	} else {
-		data.SetId(fmt.Sprintf("%v|%v|%v|%v", dbName, schemaName, tableName, priv))
+	grantID := &grantID{
+		ResourceName: dbName,
+		SchemaName:   schemaName,
+		Privilege:    priv,
+	}
+	if !onFuture {
+		grantID.ViewName = tableName
```
edulop91: this feels a bit wonky since we're assigning a `tableName` to a struct field named `ViewName`. Is there a more suitable name for this field other than `ViewName`?

pkg/resources/view_grant.go
```diff
@@ -111,37 +108,34 @@ func CreateViewGrant(data *schema.ResourceData, meta interface{}) error {
...
 	}
 
-	// ID format is <db_name>|<schema_name>|<view_name>|<privilege>
-	dataIdentifiers := make([][]string, 1)
-	dataIdentifiers[0] = make([]string, 4)
-	dataIdentifiers[0][0] = dbName
-	dataIdentifiers[0][1] = schemaName
-	dataIdentifiers[0][2] = viewName
-	dataIdentifiers[0][3] = priv
+	// ID format is <db_name>|<schema_name>|<view_name>|<privilege> or <db_name>|||<privilege>
```
edulop91: probably not needed since this logic is now encapsulated in the grantID struct

edulop91: hmm, why did you add the `or <db_name>|||<privilege>` part to this comment? I might be missing something

pkg/resources/table_grant.go
kuannie1: the view_grant.go file doesn't use tableName, so I'm not sure what I should call this attribute :/

pkg/resources/grant_helpers_internal_test.go
```diff
@@ -1,59 +1,93 @@
...
-	}
+	grant, err = grantIDFromString(id)
+	r.NoError(err)
+	r.Equal("database_name", grant.ResourceName)
+	r.Equal("", grant.SchemaName)
+	r.Equal("", grant.ViewName)
+	r.Equal("privilege", grant.Privilege)
 
-	// Bad ID
+	// Bad ID -- not enough fields
```
edulop91: nice tests :)

pkg/resources/schema.go
```diff
@@ -52,6 +58,50 @@ var schemaSchema = map[string]*schema.Schema{
...
 }
 
+type schemaID struct {
+	DatabaseName string
+	SchemaName   string
+}
+
+// String() takes in a schemaID object and returns a pipe-delimited string:
+// DatabaseName|schemaName
+func (si *schemaID) String() (string, error) {
```
edulop91: these need some tests too

pkg/resources/schema.go
```diff
@@ -52,6 +58,50 @@ var schemaSchema = map[string]*schema.Schema{
...
+	if err != nil {
+		return "", err
+	}
+	strSchemaID := strings.TrimSpace(buf.String())
+	return strSchemaID, nil
+}
+
+// schemaIDFromString() takes in a pipe-delimited string: DatabaseName|schemaName
+// and returns a schemaID object
+func schemaIDFromString(stringID string) (*schemaID, error) {
```
edulop91: also needs tests

pkg/resources/table_grant.go
edulop91: maybe `ViewOrTable` ?

pkg/resources/schema.go
kuannie1: oh huh... I wrote schema_internal_test.go
not sure why it isn't being run

pkg/resources/table_grant.go
```diff
@@ -159,20 +169,24 @@ func ReadTableGrant(data *schema.ResourceData, meta interface{}) error {
...
 	}
+
+	return readGenericGrant(data, meta, builder, onFuture)
 }
 
 // DeleteTableGrant implements schema.DeleteFunc
 func DeleteTableGrant(data *schema.ResourceData, meta interface{}) error {
-	dbName, schemaName, tableName, _, err := splitGrantID(data.Id())
+	grantID, err := grantIDFromString(data.Id())
+
```
edulop91:
```suggestion
```
extra line makes this harder to read because it adds unnecessary space between the return args and where they are being used

PR #99, opened 2019-10-22
Make Valid Privileges Public

alldoami:
Need to validate privileges in terraformer so needed these vars to be public in order to reduce duplication


edulop91:
maybe add to the description why you're making them public

PR #100, opened 2019-10-23
connection explanation to snowflake-infra

kuannie1:


README.md
```diff
@@ -24,6 +24,11 @@ You can also just download a binary from our [releases](https://github.com/chanz
 
 TODO fogg config
 
+### Using it with snowflake-infra
+If you are making changes to your terraform-provider-snowflake repo and want to test your changes with snowflake-infra:
```
ryanking: This repo is public, we should not put docs that only apply internally here.

PR #102, opened 2019-10-24
[breaking] Reading shares and roles for grants

alldoami:
Grants (_grant) resources must now be unique for their corresponding entity. We are doing this since the `Read` functions on grants were not working properly (for example, when terraform importing resources).


pkg/resources/grant_helpers.go
```diff
@@ -152,45 +147,29 @@ func readGenericGrant(data *schema.ResourceData, meta interface{}, builder snowf
...
-			}
 			roles = append(roles, grant.GranteeName)
 		case "SHARE":
-			// Shares get the account appended to their name, remove this
-			granteeNameStrippedAccount := StripAccountFromName(grant.GranteeName)
-			if !stringInSlice(granteeNameStrippedAccount, sharesIn) {
-				continue
-			}
-
 			shares = append(shares, grant.GranteeName)
```
edulop91:
```suggestion
			shares = append(shares, granteeNameStrippedAccount)
```
I think that was one of the fixes from before. Maybe keep that line and the comment. but just remove the `if !stringInSlice.....`

main.go
```diff
@@ -40,6 +41,9 @@ func doc() {
 	for _, name := range names {
 		resource := resources[name]
 		fmt.Printf("\n### %s\n\n", name)
+		if strings.Contains(name, "_grant") {
+			fmt.Printf("%s are unique for a given %s\n\n", name, strings.Trim(name, "_grant"))
```
edulop91:
```suggestion
			fmt.Printf(`\`\`\`
			%s are unique for a given %s
			\`\`\`
			`, name, strings.Trim(name, "_grant"))
```

README.md
```diff
@@ -81,6 +81,7 @@ You can see a number of examples [here](examples).
 
 ### snowflake_database_grant
 
+**Note**: snowflake_database_grant are unique for a given snowflake_database.
```
ryanking: I think we need to expand on this a bit.

main.go
```diff
@@ -40,6 +41,16 @@ func doc() {
 	for _, name := range names {
 		resource := resources[name]
 		fmt.Printf("\n### %s\n\n", name)
+		if strings.HasSuffix(name, "_grant") {
+			grant_resource_name := strings.Trim(name, "_grant")
+			fmt.Printf(`
+			**Note**: The grant resource create exclusive attachments of grants.
```
edulop91:
```suggestion
			**Note**: The grant resource creates exclusive attachments of grants.
```

main.go
```diff
@@ -40,6 +41,16 @@ func doc() {
...
+			Across the entire AWS account, all of the users/roles/groups to which a single grant is attached must be declared
```
edulop91: `users/roles/groups` might not make sense in this context -> I wonder if you can fetch it from the schema that is available in the context

main.go
```diff
@@ -40,6 +41,16 @@ func doc() {
...
+			by a single %s resource. This means that even any %s that have the attached
+			grant via any other mechanism (including other Terraform resources) will have that attached grant revoked by this resource.
+			These resources do not enforce exclusive attachment of a grant.\n
```
edulop91:
```suggestion
			These resources do not enforce exclusive attachment of a grant, it is the user's responsibility to enforce this.\n
```

PR #104, opened 2019-10-24
[feature] Publish version suffix for binaries

kuannie1:
Adding versioning to the terraform-provider-snowflake binaries so it follows the format in this link: https://www.terraform.io/docs/configuration/providers.html#plugin-names-and-versions

By having the version at the end of the binary filename, we can pin provider versions in terraform files and avoid bugs that might arise from mismatched terraform to snowflake provider.


Makefile
```diff
@@ -37,7 +38,7 @@ release-snapshot: ## run a release
 .PHONY: release-snapshot
 
 build: ## build the binary
-	go build ${LDFLAGS} .
+	go build ${LDFLAGS} -o $(BINARY_NAME) .
```
ryanking: I don't think this build will be at BINARY_NAME, since it is not built by goreleaser

Makefile
```diff
@@ -67,7 +68,7 @@ install: ## install the terraform-provider-snowflake binary in $GOPATH/bin
 
 install-tf: build ## installs plugin where terraform can find it
 	mkdir -p $(HOME)/.terraform.d/plugins
-	cp ./terraform-provider-snowflake $(HOME)/.terraform.d/plugins
+	cp ./$(BINARY_NAME) $(HOME)/.terraform.d/plugins
```
ryanking: same as above, this will not find binaries build by `make build`

Makefile
kuannie1: Can you explain this a bit more when you get the chance? I modified the goreleaser files to refer to the same binaries:
```
builds:
  - binary: terraform-provider-snowflake_{{ .Version }}
```

kuannie1: Do the binaries themselves have to be generated from the goreleaser files?

ryanking: binaries get build 2 different ways–

`make build` -> `go build .` -> './terraform-provider-snowflake'

`make release-*` -> `goreleaser ...` -> 'dist/something/terraform-provider-snowflake_version' (with this change)

To see for yourself, you can run `git clean -fdx`, which will remove all files not tracked by git (things that are .gitignore'd) and then run just one of make build|release and see the results.

kuannie1: Hmm, so when I ran `make build`, I generated terraform-provider-snowflake_<version>. I haven't been running make release, but `make release-prerelease` executes `make build` first

ryanking: Oh i think i slightly misunderstood what was going on here. This change is probably fine then.

kuannie1:
A lot of files changed, but the main files are:
.gitignore
.goreleaser.prerelease.yml
.gorelearser.yml
Makefile
main.go (enables us to see version from a command, better function naming)
version.go (go file that helps us see latest version)
scripts/update-readme.sh (referring to versioned terraform binary rather than original name)

.goreleaser.prerelease.yml
```diff
@@ -12,14 +12,11 @@ builds:
     goarch:
       - amd64
     ldflags:
-      - "-w -s -X github.com/chanzuckerberg/terraform-provider-snowflake/util.GitSha={{.Commit}} -X github.com/chanzuckerberg/terraform-provider-snowflake/util.Version={{.Version}} -X github.com/chanzuckerberg/terraform-provider-snowflake/util.Dirty=false -X github.com/chanzuckerberg/terraform-provider-snowflake/util.Release=true"
+      - "-w -s -X github.com/chanzuckerberg/terraform-provider-snowflake/pkg/version.GitSha={{.Commit}} -X github.com/chanzuckerberg/terraform-provider-snowflake/pkg/version.Version={{.Version}} -X github.com/chanzuckerberg/terraform-provider-snowflake/pkg/version.Dirty={{.Env.DIRTY}}"
```
edulop91: neat

.goreleaser.prerelease.yml
```diff
@@ -12,14 +12,11 @@ builds:
...
 
 archive:
   files:
-    - none*
+  - none*
 
 release:
-  prerelease: true
-
-env_files:
```
edulop91: hmm it fails for me if this is not present. not sure why

Makefile
```diff
@@ -66,8 +67,9 @@ install: ## install the terraform-provider-snowflake binary in $GOPATH/bin
 .PHONY: install
 
 install-tf: build ## installs plugin where terraform can find it
+	version=`./$(BASE_BINARY_NAME) -version`;
```
edulop91: seems unused, we might not need this

scripts/update-readme.sh
```diff
@@ -7,7 +7,8 @@ CMD="$1"
 
 TMP=`mktemp`
 TMP2=`mktemp`
-./terraform-provider-snowflake -doc > "$TMP"
+VERSION="$(cat VERSION)"
+./terraform-provider-snowflake_"$VERSION" -doc > "$TMP"
```
edulop91: maybe BASE_BINARY_NAME is better, that way we're not duplicating how we assemble the name of the binary here and the makefile

PR #105, opened 2019-10-25
[feature] create a database from another one

yuriolive:
Adding the feature to create a database from another one, using the clone command in Snowflake. Similar to `from_share` feature.

https://docs.snowflake.net/manuals/sql-reference/sql/create-clone.html


pkg/resources/database.go
```diff
@@ -34,6 +34,12 @@ var databaseSchema = map[string]*schema.Schema{
 		Optional:    true,
 		ForceNew:    true,
```
edulop91:
```suggestion
		ForceNew:    true,
		ConflictsWith: []string{"from_database},
```

pkg/resources/database.go
```diff
@@ -55,10 +61,18 @@ func Database() *schema.Resource {
 
 // CreateDatabase implements schema.CreateFunc
 func CreateDatabase(data *schema.ResourceData, meta interface{}) error {
-	_, ok := data.GetOk("from_share")
-	if ok {
-		return createDatabaseFromShare(data, meta)
+	_, okShare := data.GetOk("from_share")
+	_, okDatabase := data.GetOk("from_database")
```
edulop91:
```suggestion
	_, ok = data.GetOk("from_database")
	if ok {
	  return createDatabaseFromDatabase(data, meta)
	}
```

pkg/resources/database.go
```diff
@@ -55,10 +61,18 @@ func Database() *schema.Resource {
 
 // CreateDatabase implements schema.CreateFunc
 func CreateDatabase(data *schema.ResourceData, meta interface{}) error {
-	_, ok := data.GetOk("from_share")
-	if ok {
-		return createDatabaseFromShare(data, meta)
+	_, okShare := data.GetOk("from_share")
```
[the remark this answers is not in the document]
edulop91: We can simplify this logic since terraform takes care of collisions for us.

yuriolive:
@edulop91 Changed the code to use ConflictWith and the others nit-picks. I'm trying to discover why is failing in the CI, I'm getting in the database acceptance test: `Missing required argument: The argument "username" is required, but was not set.` Any idea?

edulop91:
Thank you! TravisCI restricts outside collaborators from running jobs with protected vars. I’ll run them shortly.

edulop91:
Confirming tests pass

PR #106, opened 2019-10-25
[fix] view resource to read database and statement

alldoami:
[HACK] Need to fix importing the statement, only want to keep the `SELECT` statement and disregarding everything before `SELECT`. Cleaned up the statement by removing extra whitespaces in between words and `\n` that are in the string.


pkg/resources/view.go
```diff
@@ -129,6 +129,19 @@ func ReadView(data *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
+	// Want to only capture the Select part of the query because before that is the Create part of the view which we no longer care about
+	indexOfSelect := strings.Index(text.String, "SELECT")
```
edulop91: one thing to think about, and maybe it doesn't matter is a query like:
```
CREATE SELECT_ME_VIEW AS SELECT * FROM .........
```

edulop91: or even 
```
CREATE SELECT AS SELECT * FROM ....
```
not sure if this one is valid or not though

alldoami: Would looking for `AS SELECT` be safer?

edulop91: let's try `" AS SELECT "` (with the spaces) and see where it gets us. maybe also add a 
#HACK(adoami): (why are we doing this)

alldoami: Can't do `"AS SELECT " because for some reason the strings that we are receiving as the statement have extra spaces and new line characters within the string and there is no pattern for each value we are getting. I think using `SELECT` will work for now.

alldoami: actually I think I fixed it! And cleaner statement too!

PR #108, opened 2019-10-30
[feature] Generic Grant ALL Priv

alldoami:


pkg/resources/grant_helpers.go
```diff
@@ -147,22 +147,58 @@ func readGenericGrant(data *schema.ResourceData, meta interface{}, builder snowf
...
+			privileges, ok := rolePrivileges[roleName]
+			if !ok {
+				// If not there, create an empty set
+				privileges = map[string]struct{}{}
+			}
+			// Add privilege to the set
+			privileges[grant.Privilege] = struct{}{}
+			// Reassign set back
+			rolePrivileges[roleName] = privileges
+		// TODO: do something for shares
```
edulop91: rm

pkg/resources/grant_helpers.go
```diff
@@ -147,22 +147,58 @@ func readGenericGrant(data *schema.ResourceData, meta interface{}, builder snowf
 	}
 	priv := data.Get("privilege").(string)
 
-	var roles, shares []string
+	// Map of roles to privileges
+	rolePrivileges := map[string]map[string]struct{}{}
```
ryanking: I think if you are going to use a map as a set (which is reasonable) having `map[string]bool` might be clearer.

pkg/resources/grant_helpers.go
```diff
@@ -64,6 +64,12 @@ type grantID struct {
 	Privilege    string
 }
 
+type stringSet map[string]struct{}
+
+func (ss stringSet) Add(s string) {
```
edulop91: :+1:

pkg/resources/grant_helpers.go
```diff
@@ -147,22 +153,65 @@ func readGenericGrant(data *schema.ResourceData, meta interface{}, builder snowf
...
+	for roleName, privileges := range rolePrivileges {
+		// Where priv is not all so it should match exactly
+		if _, ok := privileges[priv]; ok {
+			roles = append(roles, roleName)
+		}
+		// TODO: these list of privs might include all and ownnership -- exclude those from
+		// 	     from our calculation
+		containsAllPrivs := true
+		if priv == "ALL" {
+			for _, p := range validprivileges {
```
edulop91: you can also add a `Equals(other stringStet) bool` on you string set

pkg/resources/grant_helpers.go
ryanking: I think it is better just because it is easier to understand/read. Also using a type alias probably makes sense here.

Example https://play.golang.org/p/kdViq4Mg0Pn

pkg/resources/grant_helpers.go
```diff
@@ -64,6 +64,95 @@ type grantID struct {
 	Privilege    string
 }
 
+type stringSet map[string]struct{}
+
+func (ss stringSet) add(s string) {
+	ss[s] = struct{}{}
+}
+
+func (ss stringSet) setEquals(validPrivs []string) bool {
```
edulop91: since setEquals has specific logic for privileges, I wonder if it is worth renaming `stringSet` to `privilegeSet`

pkg/resources/grant_helpers.go
```diff
@@ -64,6 +64,95 @@ type grantID struct {
...
+			continue
+		}
+		if _, ok := ss[p]; !ok {
+			return false
+		}
+	}
+	return true
+}
+
+// func writeToFile(filename string, data string) error {
```
edulop91: rm dead code

pkg/resources/grant_helpers.go
```diff
@@ -64,6 +64,95 @@ type grantID struct {
...
+// 	defer file.Close()
+
+// 	_, err = io.WriteString(file, data)
+// 	if err != nil {
+// 		return err
+// 	}
+// 	return file.Sync()
+// }
+
+func filterALLGrants(grantList []*grant, validPrivs []string) []*grant {
```
edulop91: this probably requires a lengthy comment answering why are we doing this?

pkg/resources/grant_helpers.go
```diff
@@ -64,6 +64,95 @@ type grantID struct {
...
+	// For each database_schema_role, figure out if the grant has all of the privileges
+	type databaseSchemaRole struct {
+		GrantName   string
+		GranteeType string
+		GranteeName string
+	}
+	groupedByRole := map[databaseSchemaRole]map[string]struct{}{}
```
edulop91: we can use our `stringSet` here

pkg/resources/grant_helpers.go
```diff
@@ -64,6 +64,95 @@ type grantID struct {
...
+			GranteeName: g.GranteeName,
+		}
+		if _, ok := groupedByRole[id]; !ok {
+			groupedByRole[id] = map[string]struct{}{}
+		}
+		groupedByRole[id][g.Privilege] = struct{}{}
+	}
+	for databaseSchemaRole, privs := range groupedByRole {
+		for _, p := range validPrivs {
+			if p == "ALL" || p == "OWNERSHIP" || p == "CREATE STREAM" {
```
edulop91: This can be the `stringSet.setEquals()`

pkg/resources/grant_helpers.go
edulop91: setEquals might also need a better name, such as `AllPrivsPresent` perhaps

pkg/resources/grant_helpers.go
```diff
@@ -145,24 +218,56 @@ func readGenericGrant(data *schema.ResourceData, meta interface{}, builder snowf
...
 	}
+
+	var roles, shares []string
+	// Now see which roles have our privilege
+	for roleName, privileges := range rolePrivileges {
+		// Where priv is not all so it should match exactly
+		if _, ok := privileges[priv]; ok || privileges.ALLPrivsPresent(validPrivileges) {
+			roles = append(roles, roleName)
+		}
+		// TODO: these list of privs might include all and ownnership -- exclude those from
```
edulop91: we've done this now, rm stale comment

pkg/resources/view_grant.go
```diff
@@ -162,10 +162,10 @@ func ReadViewGrant(data *schema.ResourceData, meta interface{}) error {
 	var builder snowflake.GrantBuilder
 	if futureViews {
```
edulop91: i would refactore `futureViews` to `futureViewsEnabled` to make it obvious it is a boolean value. I would then, instead of the if/else would just `return readGenericGrant(data, meta, builder, futureViewsEnabled, ValidViewPrivileges)`

PR #111, opened 2019-10-31
[fix] resource_monitor credit quota as float64, not int

chauvm:
Issue: https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/109

Note that if a float has more than 6 trailing numbers after the decimal point, `%f` rounds it up to have exactly 6. If fewer than 6 trailing numbers are provided, it appends zeros, for example, `66.00` becomes `66.000000`, which I think is fine?

### DIFF
--- a/pkg/resources/resource_monitor.go
+++ b/pkg/resources/resource_monitor.go
@@ -181,7 +181,7 @@ func ReadResourceMonitor(data *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	// Credit quota is the only integer
+	// Credit quota is a float
 	if rm.CreditQuota.Valid {
 		err = data.Set("credit_quota", rm.CreditQuota.Float64)
 	} else {
--- a/pkg/resources/resource_monitor.go
+++ b/pkg/resources/resource_monitor.go
@@ -44,7 +44,7 @@ var resourceMonitorSchema = map[string]*schema.Schema{
 		Type:        schema.TypeFloat,
 		Optional:    true,
 		Computed:    true,
-		Description: "The amount of credits allocated monthly to the resource monitor.",
+		Description: "The amount of credits allocated monthly to the resource monitor, round up to 2 decimal places.",
 		ForceNew:    true,
 	},
 	"frequency": &schema.Schema{
--- a/pkg/resources/resource_monitor.go
+++ b/pkg/resources/resource_monitor.go
@@ -15,20 +15,20 @@ import (
 )
 
 type resourceMonitor struct {
-	Name                 sql.NullString `db:"name"`
-	CreditQuota          sql.NullInt64  `db:"credit_quota"`
-	UsedCredits          sql.NullString `db:"used_credits"`
-	RemainingCredits     sql.NullString `db:"remaining_credits"`
-	Level                sql.NullString `db:"level"`
-	Frequency            sql.NullString `db:"frequency"`
-	StartTime            sql.NullString `db:"start_time"`
-	EndTime              sql.NullString `db:"end_time"`
-	NotifyAt             sql.NullString `db:"notify_at"`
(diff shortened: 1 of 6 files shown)


pkg/resources/resource_monitor.go
```diff
@@ -183,9 +183,9 @@ func ReadResourceMonitor(data *schema.ResourceData, meta interface{}) error {
 
 	// Credit quota is the only integer
```
edulop91: not an integer, maybe worth fixing that comment

pkg/snowflake/resource_monitor_test.go
```diff
@@ -22,14 +22,14 @@ func TestResourceMonitor(t *testing.T) {
...
+	ab.SetFloat("credit_quota", 66.6)
 	q = ab.Statement()
-	a.Equal(`ALTER RESOURCE MONITOR "resource_monitor" SET CREDIT_QUOTA=66`, q)
+	a.Equal(`ALTER RESOURCE MONITOR "resource_monitor" SET CREDIT_QUOTA=66.600000`, q)
 
 	cb := snowflake.ResourceMonitor("resource_monitor").Create()
 	cb.NotifyAt(80).NotifyAt(90).SuspendAt(95).SuspendImmediatelyAt(100)
 	cb.SetString("frequency", "YEARLY")
-	cb.SetInt("credit_quota", 666)
+	cb.SetFloat("credit_quota", 666.66666666)
```
edulop91: i wonder if it is also worth keeping a test where the input "looks like an integer" and making sure the parsing works properly

pkg/resources/resource_monitor.go
chauvm: done

pkg/snowflake/resource_monitor_test.go
chauvm: good point

pkg/snowflake/resource_monitor_test.go
```diff
@@ -22,14 +22,20 @@ func TestResourceMonitor(t *testing.T) {
...
 	cb.NotifyAt(80).NotifyAt(90).SuspendAt(95).SuspendImmediatelyAt(100)
 	cb.SetString("frequency", "YEARLY")
-	cb.SetInt("credit_quota", 666)
+
+	cb.SetFloat("credit_quota", 666.66666666)
+	q = cb.Statement()
+	a.Equal(`CREATE RESOURCE MONITOR "resource_monitor" FREQUENCY='YEARLY' CREDIT_QUOTA=666.666667 TRIGGERS ON 80 PERCENT DO NOTIFY ON 90 PERCENT DO NOTIFY ON 95 PERCENT DO SUSPEND ON 100 PERCENT DO SUSPEND_IMMEDIATE`, q)
+
+	// Check if credit quota can be parsed correctly to float if given an integer
+	cb.SetFloat("credit_quota", 666)
```
chauvm: Added this test, integers are converted to floats as expected

pkg/snowflake/generic.go
```diff
@@ -81,6 +87,10 @@ func (ab *AlterPropertiesBuilder) Statement() string {
 		sb.WriteString(fmt.Sprintf(" %s=%d", strings.ToUpper(k), v))
 	}
 
+	for k, v := range ab.floatProperties {
+		sb.WriteString(fmt.Sprintf(" %s=%f", strings.ToUpper(k), v))
```
edulop91:
```suggestion
		sb.WriteString(fmt.Sprintf(" %s=%2f", strings.ToUpper(k), v))
```
what do you think about setting the resolution to 2 decimals?

chauvm: if that's what users usually provide I think it makes more sense to round up to 2 figures. I'll fix the tests and documentation to reflect that.

ms32035:
That was quick. Thanks

PR #112, opened 2019-11-05
[fix] Always filter out the ALL schema_grants

edulop91:


pkg/resources/view_grant_acceptance_test.go
```diff
@@ -1,34 +1,31 @@
...
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/stretchr/testify/require"
 )
 
 func TestAccViewGrantBasic(t *testing.T) {
-	vName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	viewName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	databaseName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
```
alldoami: Can combine viewName and databaseName to same value all in one line? just a suggestion

pkg/resources/view_grant_acceptance_test.go
```diff
@@ -38,48 +35,44 @@ func TestAccViewGrantShares(t *testing.T) {
...
-				ImportStateVerify: true,
-			},
 		},
 	})
 }
 
 func TestAccFutureViewGrantChange(t *testing.T) {
-	vName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	viewName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	databaseName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
```
alldoami: same here?

pkg/resources/view_grant_acceptance_test.go
alldoami: viewName, databaseName := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)

edulop91: that won't work though since `acctest.RandStringFromCharSet` has a single string return value

alldoami: ah I see, I thought there'd be a way to give it the same value. Forgot that's how go interprets that syntax though

PR #119, opened 2019-12-19
[fix] schema managed access option

jong82:
Terraform state refresh was always setting the managed access option to the default, false.

### DIFF
--- a/pkg/resources/schema.go
+++ b/pkg/resources/schema.go
@@ -223,7 +223,7 @@ func ReadSchema(data *schema.ResourceData, meta interface{}) error {
 				if err != nil {
 					return err
 				}
-			case "MANAGED":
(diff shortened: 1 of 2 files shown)


jong82:
Built the provider locally and works as expected.  However tests are failing even though the errors don't seem to be related to the change I made?

jong82:
I think Travis CI removes sensitive environment variables for PR's coming from outside the repo.  However I'm getting permission denied when trying to create a new branch inside the repo.  How can I create a proper PR?

https://travis-ci.com/chanzuckerberg/terraform-provider-snowflake/jobs/269044056#L183
```Encrypted environment variables have been removed for security reasons.
See https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions```

PR #121, opened 2019-12-20
Add snowflake_stage resource

ajbosco:


pkg/resources/stage_grant.go
```diff
@@ -0,0 +1,153 @@
...
+	"stage_name": &schema.Schema{
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "The name of the stage on which to grant privileges.",
+		ForceNew:    true,
+	},
+	"schema_name": &schema.Schema{
+		Type:        schema.TypeString,
+		Optional:    true,
+		Default:     "PUBLIC",
```
ryanking: What is the reason for PUBLIC as a default? Defaulting to that level makes me a bit nervous that people will create stages that are more exposed than they expect. (I know that PUBLIC doesn't mean "public", but still).

pkg/resources/stage_grant_test.go
```diff
@@ -0,0 +1,61 @@
...
+	err := resources.StageGrant().InternalValidate(provider.Provider().Schema, true)
+	r.NoError(err)
+}
+
+func TestStageGrantCreate(t *testing.T) {
+	a := assert.New(t)
+
+	in := map[string]interface{}{
+		"stage_name":    "test-stage",
+		"schema_name":   "PUBLIC",
```
ryanking: Can we use something other than the default to test here?

pkg/resources/stage_grant_test.go
```diff
@@ -0,0 +1,61 @@
...
+		"database_name": "test-db",
+		"privilege":     "USAGE",
```
ryanking: Same as above - to be sure that we correctly handle non-default parameters, we should test with a non-default.

pkg/resources/stage_grant.go
ajbosco: Yeah, I'll just remove the default. No real need for it.

pkg/resources/stage_grant_test.go
ajbosco: Added a few stage access control privileges that I missed (https://docs.snowflake.net/manuals/user-guide/security-access-control-privileges.html#stage-privileges) and added a non-default test.

pkg/resources/stage.go
```diff
@@ -0,0 +1,414 @@
...
+		Type:        schema.TypeString,
+		Required:    true,
+		Description: "The database in which to create the stage.",
+		ForceNew:    true,
+	},
+	"schema": &schema.Schema{
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "The schema in which to create the stage.",
+		Default:     "PUBLIC",
```
ryanking: I think we should drop this default.

pkg/resources/stage.go
```diff
@@ -0,0 +1,414 @@
...
+	},
+	"file_format": &schema.Schema{
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies the file format for the stage.",
+	},
+	"copy_options": &schema.Schema{
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies the copy optionos for the stage.",
```
ryanking:
```suggestion
		Description: "Specifies the copy options for the stage.",
```

pkg/resources/stage.go
```diff
@@ -0,0 +1,414 @@
...
+	err = data.Set("aws_external_id", stageDesc.awsExternalID)
+	if err != nil {
+		return err
+	}
+
+	err = data.Set("snowflake_iam_user", stageDesc.snowflakeIamUser)
+	if err != nil {
+		return err
+	}
+
```
ryanking: IIUC some fields are omitted here because you can't read them back from snowflake via `DESCRIBE|SHOW STAGE`. Can you add a comment listing those and why they are omitted?

ryanking: Also can you check that there is no way to read those back? Sorry I don't know this part of snowflake well.

ajbosco: Added `file_format`, `copy_options` and `comment`. The other two fields (`credentials` and `encryption`) are not available in the either the `SHOW STAGES` or `DESCRIBE STAGE` queries.

doctoryes:
@ryanking This PR is great! Should it warrant a new `0.9.3` (or `0.10.0`) release? My vote: 👍

ryanking:
@doctoryes just released 0.9.3

doctoryes:
Thanks, @ryanking !

PR #128, opened 2020-01-16
Add pipes support

pevloshevskaya:


pkg/snowflake/pipe.go
```diff
@@ -0,0 +1,113 @@
...
+	q.WriteString(`CREATE`)
+
+	q.WriteString(fmt.Sprintf(` PIPE %v`, pb.QualifiedName()))
+
+	if pb.autoIngest {
+		q.WriteString(` AUTO_INGEST = TRUE`)
+	}
+
+	if pb.comment != "" {
+		q.WriteString(fmt.Sprintf(` COMMENT = '%v'`, pb.comment))
```
ryanking:
```suggestion
		q.WriteString(fmt.Sprintf(` COMMENT = '%v'`, EscapeString(pb.comment)))
```

We have been lax about escaping values but need to get better.

pkg/snowflake/pipe.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	if pb.copyStatement != "" {
+		q.WriteString(fmt.Sprintf(` AS %v`, pb.copyStatement))
```
ryanking: Not sure if this needs to be escaped as well.

pkg/snowflake/pipe.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	return q.String()
+}
+
+// ChangeComment returns the SQL query that will update the comment on the pipe.
+func (pb *PipeBuilder) ChangeComment(c string) string {
+	return fmt.Sprintf(`ALTER PIPE %v SET COMMENT = '%v'`, pb.QualifiedName(), c)
```
ryanking:
```suggestion
	return fmt.Sprintf(`ALTER PIPE %v SET COMMENT = '%v'`, pb.QualifiedName(), EscapeString(c))
```

pevloshevskaya: I fixed all the comments in [this commit ](https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/128/commits/b34292c2a74c3693501588a9fbd450eb0eb42812)

doctoryes:
🎉 !

Issue #129, opened 2020-01-16
Error when you create an internal stage


thalesmello:
If you try to create an internal stage like:

```
resource snowflake_stage my_stage {
  name     = "MY_STAGE"
  database = "TEST"
  schema   = "PUBLIC"
}
```

When you do a `terraform apply`, you get `Error: sql: Scan error on column index 9, name "region": converting NULL to string is unsupported`

This apparently is because [`region` cannot be `nil`](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/df3a407eb2436f1c957d8b4b269248fbb29c0f09/pkg/resources/stage.go#L430). I'm not sure what could be done to solve this problem.

thalesmello:
I have a pull request open at #130

ryanking:
@thalesmello we can consider this fixed, yeah?

thalesmello:
Correct. Closing the issue

PR #130, opened 2020-01-17
[fix] Fix cannot convert NULL to string when creating stage (#129)

thalesmello:
When I try to create an internal stage, Terraform breaks with an error
trying to convert a NULL value to a string, which is not possible in Go.

When you run SHOW STAGES in internal stages, the "region" and "cloud"
fields are NULL. This pull request fixes that by using *string pointers
in the SHOW STAGE struct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,10 +359,16 @@ We also run all tests in our [Travis-CI account](https://travis-ci.com/chanzucke
 
 Our CI jobs run the full acceptence test suite, which involves creating and destroying resources in a live snowflake account. Travis-CI is configured with environment variables to authenticate to our test snowflake account. For security reasons, those variables are not available to forks of this repo.
 
-If you are making a PR from a forked repo, you can set up Travis to build it by setting these environement variables:
+If you are making a PR from a forked repo, you can create a new Snowflake trial account and set up Travis to build it by setting these environement variables:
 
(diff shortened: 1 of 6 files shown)


thalesmello:
I can't put my company's credentials on Travis CI due to some security
constraints. Is there some other way to test this PR?

ryanking:
@thalesmello I understand the limitations on running in travis-ci. We only do it with a separate snowflake account dedicated to testing.

You can also run the acceptance tests locally from the machine you are working on. You need to set the same env variables - https://github.com/chanzuckerberg/terraform-provider-snowflake#pull-request-ci

thalesmello:
@ryanking Apparently Snowflake included two new fields over the weekend in the `SHOW STAGES` command, which actually breaks any code that uses the `stage` resource with the code currently in master branch. I'm pushing a bug fix in this pull request.

thalesmello:
@ryanking I'm not able to run the Travis with the correct environment variables in this PR, but you might want to check the build in my fork:

https://travis-ci.com/thalesmello/terraform-provider-snowflake/jobs/278105235

ryanking:
@thalesmello that build in your fork seems to skip the acceptance tests. When I ran it locally that is where I saw a failure.

Let me run again and see if I can get some output.

ryanking:
Ah tests are passing now.

PR #131, opened 2020-01-21
Add statement_timeout_in_seconds param support

pevloshevskaya:


pkg/resources/warehouse.go
```diff
@@ -132,6 +132,13 @@ var warehouseSchema = map[string]*schema.Schema{
 		Description: "Specifies whether the warehouse, after being resized, waits for all the servers to provision before executing any queued or new queries.",
 		Optional:    true,
 	},
+	"statement_timeout_in_seconds": &schema.Schema{
+		Type:        schema.TypeInt,
+		Optional:    true,
+		Default:     1000,
```
ryanking: I think our default should be the same as snowflake's. AFAICT from the docs that would be 0, which is interpreted as "no timeout".

pevloshevskaya: Fix in [this commit](https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/131/commits/b519b5d5c1706438f07274c569a5f87ef503b384)

pevloshevskaya: @ryanking Just a friendly reminder ^^

ryanking:
There is now a test failure when running the acceptance tests–

```
--- FAIL: TestAccWarehouse (3.84s)
    testing.go:635: Step 3 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=28) "statement_timeout_in_seconds": (string) (len=1) "0"
        }
      
```

PR #133, opened 2020-01-23
Add two missing schema grants

marcinef:
Snowflake supports two more schema privileges:

CREATE EXTERNAL TABLE
CREATE MATERIALIZED VIEW

Currently provider does not have above two privileges listed.


marcinef:
Hmm. the Travis CI failed.  Let me check and try fix it.

marcinef:
Should be all good now.  I do not have trial account so if you could run this in 
your test snowflake account it would be very convenient.

PR #136, opened 2020-01-31
Add AWS SNS topic to pipe resource

willejs:
Currently when creating a pipe you cant specify an SNS topic to associate a pipes SQS queue with. This pr add that option.


willejs:
Hmm, this worked on the first apply, but after destroying the pipe and sns topic and retrying it, it does not register the pipes sqs queue to the topic which we are currently debugging.

ryanking:
@willejs thanks for the PR - let us know if you need help debugging

willejs:
I am going to close this, as i couldn't get it working reliably and we decided don't need to use a SNS topic yet. For now we opted to attach the SQS queue created by the snowpipe stage instead.

PR #138, opened 2020-02-13
[feature] storage integration

### LINKED ISSUE #125: Add storage_integration
Storage Integrations consist in linking Snowflake accounts with AWS, Azure or GCP accounts via Roles.

It's currently the best practice for connecting with your cloud environment, and it definitely one of the main benefits from being on Terraform : you can create the AWS Role (for example) and pass it as parameter to the Snowflake StorageIntegration 👍 

https://docs.snowflake.net/manuals/sql-reference/sql/create-storage-integration.html

I think a first step would be AWS, then Azure and GCP (GCP was in preview until recently)

sjauld:
Resolves #125 

There is some strange stuff going on on my travis branch - I think that some build settings have changed since I set it up and I'm not super sure how to fix it.


sjauld:
Please hold off on looking at this one, I have found a bug. Will fix up shortly.

sjauld:
Ok should be all good now - I have also added the storage integration property onto stages.

sjauld:
I'm just gonna close this for now - there are a few missing bits and pieces needed for it to be really helpful at this stage. I'll do some more work then reopen.

PR #140, opened 2020-02-20
[feature] storage integration

### LINKED ISSUE #125: Add storage_integration

### LINKED ISSUE #144: snowflake_stage: support for storage_integration
Looking at the docs and the code, doesn't seem like `storage_integration` is not being handled.
We use storage_integrations on all of our stages, and can't make use of this provider for them.

Example from https://docs.snowflake.net/manuals/sql-reference/sql/create-stage.html#external-stages:

```
create or replace stage my_ext_stage
  url='s3://load/files/'
  storage_integration = myint;
```

sjauld:
Ok everything seems to be fixed up and working now.

This adds
* storage integration resource
* support to specify a storage integration for a stage

Closes #125 
Closes #144


pkg/resources/storage_integration.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+		stmt.SetString(`STORAGE_AWS_ROLE_ARN`, v.(string))
+	case "AZURE":
+		v, ok := data.GetOk("azure_tenant_id")
+		if !ok {
+			return fmt.Errorf("If you use the Azure storage provider you must specify an azure_tenant_id")
+		}
+		stmt.SetString(`AZURE_TENANT_ID`, v.(string))
+	}
+
```
alldoami: Looks like you need to add a case for `GCS` if you're allowing it here: https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/140/files#diff-ea99148ac4d4624cd63de6823b620da4R52

Would also be good to add a default statement so that it errors if there are any other providers given. Same goes for the other switch case statements.

pkg/resources/storage_integration.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+		case "STORAGE_AWS_ROLE_ARN":
+			if err = data.Set("storage_aws_role_arn", v.(string)); err != nil {
+				return err
+			}
+		case "STORAGE_AWS_EXTERNAL_ID":
+			if err = data.Set("storage_aws_external_id", v.(string)); err != nil {
+				return err
+			}
+		}
```
alldoami: Add default statement here to error if `k` is not any of these values.

sjauld: An error would break the provider if Snowflake adds a new property here, so I have printed a warning instead.

pkg/resources/storage_integration.go
sjauld: There is nothing to set for GCS so I have added a no op case.
I also added a default, but note that the provider validates the storage provider so it shouldn't be possible for this to happen.

sjauld:
Ooh also resolves #144

ryanking:
> Ooh also resolves #144

you can edit the description above to add that

pkg/snowflake/stage_test.go
```diff
@@ -68,6 +71,12 @@ func TestStageChangeCredentials(t *testing.T) {
 	a.Equal(s.ChangeCredentials("aws_role='arn:aws:iam::001234567890:role/mysnowflakerole'"), `ALTER STAGE "test_db"."test_schema"."test_stage" SET CREDENTIALS = (aws_role='arn:aws:iam::001234567890:role/mysnowflakerole')`)
 }
 
+func TestStageChangeStorageIntegration(t *testing.T) {
+	a := assert.New(t)
+	s := Stage("test_stage", "test_db", "test_schema")
+	a.Equal(s.ChangeStorageIntegration("MY_INTEGRATION"), `ALTER STAGE "test_db"."test_schema"."test_stage" SET STORAGE_INTEGRATION = 'MY_INTEGRATION'`)
```
brewkode: I was looking to build Storage integration and found this. I went through the test and found that the test assertion was _mildly_ wrong. Before that, the usage of `assert.Equal` is reversed in terms of expected & actual, hence the error message from assertion is flipped. 

Actual error from Travis
```--- FAIL: TestStageChangeStorageIntegration (0.00s)
    stage_test.go:77: 
        	Error Trace:	stage_test.go:77
        	Error:      	Not equal: 
        	            	expected: "ALTER STAGE \"test_db\".\"test_schema\".\"test_stage\" SET STORAGE_INTEGRATION = MY_INTEGRATION"
        	            	actual  : "ALTER STAGE \"test_db\".\"test_schema\".\"test_stage\" SET STORAGE_INTEGRATION = 'MY_INTEGRATION'"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-ALTER STAGE "test_db"."test_schema"."test_stage" SET STORAGE_INTEGRATION = MY_INTEGRATION
        	            	+ALTER STAGE "test_db"."test_schema"."test_stage" SET STORAGE_INTEGRATION = 'MY_INTEGRATION'
        	Test:       	TestStageChangeStorageIntegration```

Looks like `MY_INTEGRATION` in the "expected" string "ALTER STAGE ..." should NOT be Single quoted :)

@sjauld - Please see if you can do this minor edit. And, thank you for spending time on this.

sjauld: Thanks @brewkode! This was a bad oversight on my behalf.

Note that all of the assertions are the wrong way around so I'll just leave these for now.

brewkode:
Whole set of "Acc" related tests have been failing from the first commit after "go fmt" in this branch. @sjauld May be master was not stable when the feature branch fork happened ? I tried merging master into this branch and had lot of conflicts. Would need sometime to understand this fully.

sjauld:
I don't understand your issue with conflicts- this branch is forked  from the current HEAD of master, so there is nothing to merge. Have you tried merging this repo's master into your local master before merging my branch?

Travis tests will not pass on pull requests since they don't get access to snowflake creds. I think we should take an approach like the AWS provider but that's a seperate issue. 

Also, I have no idea why the docs tests are failing, they have changed since I last contributed here and I'm not sure what the issue is, but that seems to also affect all PRs.

brewkode:
Thanks for your response @sjauld . I did not spend time on merge failures, Im going to do now and will keep you posted. I did not realize the issue with snowflake creds - my bad!

brewkode:
@ryanking So, do you think, we can merge this PR ?

alldoami:
`make check-docs` failed so please run `make docs` and push again :) I'll run the test separately and then merge!

sjauld:
> `make check-docs` failed so please run `make docs` and push again :) I'll run the test separately and then merge! @sjauld

@alldoami `make check-docs` is passing now https://travis-ci.com/chanzuckerberg/terraform-provider-snowflake/jobs/296288487

alldoami:
Ran `make test-acceptance-ci` and looked good! Will merge.

yuriolive:
Any plans to have a release with this functionality?

brewkode:
@alldoami Thanks for taking a look and merging this PR.  @sjauld Thanks for helping with this feature 👍 

Can we please 🙏 cut a release with this functionality ?

lgwacker:
Hey guys! Is this available at the latest release? I'm trying to use it using release 0.13 but it says the provider doesn't have this resource.

PR #141, opened 2020-02-20
Add support for "ALL" privilege in snowflake_table_grant resource

aelesbao:
All Snowflake resources [support](https://docs.snowflake.net/manuals/user-guide/security-access-control-privileges.html) the `GRANT ALL` privilege, but the `snowflake_table_grant` resource doesn't:

```
Error: expected privilege to be one of [DELETE TRUNCATE REFERENCES SELECT INSERT UPDATE], got ALL
```

### DIFF
--- a/pkg/resources/table_grant.go
+++ b/pkg/resources/table_grant.go
@@ -9,6 +9,7 @@ import (
 )
 
 var validTablePrivileges = newPrivilegeSet(
+	privilegeAll,
 	privilegeSelect,
 	privilegeInsert,
 	privilegeUpdate,


pwnage101:
If you add the ALL privilege, all I ask is that you add a unit test for it that passes.  The problem with all is that in snowflake it is just an alias, not a real privilege.  It may not work they way you expect, but I'm happy to be proved wrong.

aelesbao:
@ryanking @pwnage101 thanks, I'll work on that soon. I confirmed in Snowflake documentation, and the [grant privilege syntax](https://docs.snowflake.net/manuals/sql-reference/sql/grant-privilege.html#syntax) support the keyword `ALL` for any object type. Although, they say:

> Multiple privileges can be specified for the same object type in a single `GRANT` statement (with each privilege separated by commas), or the special `ALL [ PRIVILEGES ]` keyword can be used to grant all applicable privileges to the specified object type. Note, however, that only privileges held and grantable by the role executing the `GRANT` command are actually granted to the target role. A warning message is returned for any privileges that could not be granted.

Since I'm new to writing Terraform providers, could you please guide me on how to print this warning message during the plan/apply? Is there a built-in option in Terraform for that?

ryanking:
> Since I'm new to writing Terraform providers, could you please guide me on how to print this warning message during the plan/apply? Is there a built-in option in Terraform for that?

As far as I know, there is no way to do that.

davehowell:
If you apply the `ALL` privilege from a role that only has permission to apply a subset then `ALL` should resolve to that subset and be stored in the state as that subset, not as `ALL`. It shouldn't be an error or a warning. 

e.g. you grant with a role that only has permission to apply `SELECT` permission, then applying `ALL` should be stored as just `SELECT`, but then trying to re-apply the exact same state with the same role - assuming the role's privilege is the same - should not force a new resource. 

If you later change the role to one that has privilege to grant more things, e.g. `[SELECT INSERT UPDATE]` then applying `ALL` should first resolve to `[SELECT INSERT UPDATE]` and then apply those additional permissions and update the tf state accordingly. Similarly if you change the role to one with less privilege then applying `ALL` should also change the state.

Using the correct role is a very strong theme in Snowflake and not something the Terraform Provider should obscure or abstract away.

funes79:
@davehowell I agree that this would be a nice feature to have, but to do that the provider would have to hard-code what ALL means for every single type of resource. On schema ALL means 17 privileges on database ALL means 5 privileges (MODIFY, MONITOR, USAGE, CREATE SCHEMA, IMPORTED PRIVILEGES).

In this way when you define a ALL privilege resource (let's say for database) the provider would have to create 5 resources in the state file. When somebody would revoke let's say CREATE SCHEMA from the role (outside of terraform), then the next run of the terraform should detect that this one is missing and thus grant it back to the role.

Similarly when a new change is applied in the tf code and ALL is changed to two resouces, let's say CREATE SCHEMA and USAGE, then the provider would have to detect that those 2 privileges are already covered in the 5 resources, so it just need to remove the MONITOR, MODIFY and IMPORT PRIVILEGES permission from db for this role.

Source: https://docs.snowflake.com/en/user-guide/security-access-control-privileges.html

aidan-melen:
I agree this would be nice and keep the resource DRY, but seems like an provider anti-pattern to have a single resource declaration manage multiple resources in the state.

I was going to make a local module and see how that goes. If it works nice I will share it on the terraform registry. That seems like it would clean up the code while not breaking the 1-to-1 relationship between a resource and a single snowflake object.

aidanmelen:
I started work on the [terraform-snowflake-grant-all](https://github.com/aidanmelen/terraform-snowflake-grant-all#complete-example) module. I realize that some of the Snowflake provider resources already support privilegeAll. However, the provider doesn't create each grant in terraform state, perhaps that is intended, but this module will create a resource for each expanded grant from ALL.

funes79:
thanks @aidanmelen

ryanking:
We are going to be [removing support for all ALL grants](https://github.com/chanzuckerberg/terraform-provider-snowflake/discussions/318).

Issue #152, opened 2020-03-19
Grant ownership fails on object with dependent grants


seediang:
Grant ownership fails when there is already a dependent grant on the object. 

> snowflake_role.role_one: Creating...
> snowflake_database.database_one: Creating...
> snowflake_role.role_two: Creating...
> snowflake_database.database_one: Creation complete after 2s [id=database_one]
> snowflake_schema.schema_one: Creating...
> snowflake_role.role_one: Creation complete after 2s [id=role_one]
> snowflake_role_grants.role_one: Creating...
> snowflake_role.role_two: Creation complete after 2s [id=role_two]
> snowflake_role_grants.role_two: Creating...
> snowflake_role_grants.role_one: Creation complete after 0s [id=role_one]
> snowflake_schema.schema_one: Creation complete after 0s [id=database_one|schema_one]
> snowflake_schema_grant.create_table: Creating...
> snowflake_role_grants.role_two: Creation complete after 1s [id=role_two]
> snowflake_schema_grant.create_table: Creation complete after 1s [id=database_one|schema_one||CREATE TABLE]
> snowflake_schema_grant.ownership: Creating...
> 
> Error: 003036 (23001): SQL execution error: Dependent grant of privilege 'CREATE TABLE' on securable '"database_one"."schema_one"' to role '"role_one"' exists.  It must be revoked first.  More than one dependent grant may exist: use 'SHOW GRANTS' command to view them.  To revoke all dependent grants while transferring object ownership, use convenience command 'GRANT OWNERSHIP ON <target_objects> TO <target_role> REVOKE CURRENT GRANTS'.
> 
>   on main.tf line 42, in resource "snowflake_schema_grant" "ownership":
>   42: resource "snowflake_schema_grant" "ownership" {

The following is an example used to create the error above
``` 
provider "snowflake" {
  account = "SNOWFLAKEACCOUNT"
  // running as ACCOUNTADMIN TO AVOID dealing with security
  // not a good practise
  role = "ACCOUNTADMIN"
}

resource "snowflake_database" "database_one" {
    name = "database_one"
}

resource "snowflake_schema" "schema_one" {
    name        = "schema_one"
    database    = snowflake_database.database_one.name
}

// CREATE ROLE_ONE and grant to SYSADMIN
resource "snowflake_role" "role_one" {
    name    = "role_one"
}
resource "snowflake_role_grants" "role_one" {
    role_name   = snowflake_role.role_one.name
[...]

PR #153, opened 2020-03-19
[fix] Add 'copy current grants' when performing ownership grant #152

saperry:


alldoami:
I ran `make test-acceptance-ci` and `make setup` and both yielded no errors.

pkg/snowflake/grant.go
```diff
@@ -138,7 +138,13 @@ func (gb *CurrentGrantBuilder) Share(n string) GrantExecutable {
...
-	return fmt.Sprintf(`GRANT %v ON %v %v TO %v "%v"`,
+	var template string
+	if p == `OWNERSHIP` {
+		template = `GRANT %v ON %v %v TO %v "%v" COPY CURRENT GRANTS`
+	} else {
+		template = `GRANT %v ON %v %v TO %v "%v"`
+	}
+	return fmt.Sprintf(template,
 		p, ge.grantType, ge.grantName, ge.granteeType, ge.granteeName)
 }
```
jas-grover, line 149: Despite this fix, I've hit the `Dependent grant of privilege` error a few times. What I've found is in the [grant resource](https://registry.terraform.io/providers/chanzuckerberg/snowflake/latest/docs/resources/stage_grant), I had specified `privilege = "ownership"` (lowercase) which meant that it did not match `p == OWNERSHIP` (uppercase) and ran the query without the `copy current grants`.

Changing it to uppercase worked. So seems like it should be handled in this function as the provider doesn't throw an error in general if a lowercase value is provided and the docs don't mention it being case sensitive.

PR #159, opened 2020-03-26
docs: Add a few "import" sections for some of the resources

pwnage101:
These "import" sections are available in 1st party provider docs, and I
found them to be helpful.  As a contributor to
terraform-provider-snowflake, I did not need the sections that I'm
contributing, but I can imagine how a non-contributor will find them to
be helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,14 @@ You can see a number of examples [here](examples).
 | from_share                  | map    | Specify a provider and a share in this map to create a database from a share. | true     | false     | false    | <nil>   |
 | name                        | string |                                                                               | false    | true      | false    | <nil>   |
 
+#### import
+
+Databases are simply identified by their name:
+
+```shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +80,14 @@ You can see a number of examples [here](examples).
 | from_share                  | map    | Specify a provider and a share in this map to create a database from a share. | true     | false     | false    | <nil>   |
 | name                        | string |                                                                               | false    | true      | false    | <nil>   |
 
+#### import
```
ryanking: I like what you are doing here, but all the docs between `<!-- START -->` and `<!-- END -->` is auto-generated  see the [docs make target](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/4346dee4430764fa7c608a761b5ffd3c53650631/Makefile#L94).

pwnage101: oops forgot about that, will look into that later.

PR #171, opened 2020-04-24
[fix] reduce spurious diffs to snowflake_view statement

ryanking:
Make an effort to normalize queries that create views.

Snowflake does not preserve formatting when round-tripping these
queries, so to compare changes we need to do some normalization. See the
comment in the code for limitations.

## Test Plan
* [ ] acceptance tests

## References
* A previous, unfinished attempt- #114
* #view


pkg/testhelpers/fixtures.go
```diff
@@ -0,0 +1,19 @@
+package testhelpers
+
+import (
+	"io/ioutil"
+	"path/filepath"
+)
+
+func Fixture(name string) string {
```
edulop91, line 8:
```suggestion
func MustFixture(name string) string {
```
I think that's the convention

pkg/resources/view.go
```diff
@@ -44,18 +44,30 @@ var viewSchema = map[string]*schema.Schema{
...
-		},
-	},
+// DiffSuppressStatement will suppress diffs between statemens if they differ in only case or in
+// runs of whitespace (\s+ = \s). This is needed because the snowflake api does not faithfully
+// round-trip queries so we cannot do a simple character-wise comparison to detect changes.
+//
+// Warnings: We will have false positives in cases where a change in case or run of whitespace is
+// semantically significant.
+//
+// If we can find a sql parser that can handle the snowflake dialect then we should switch to pasing
```
edulop91, line 66:
```suggestion
// If we can find a sql parser that can handle the snowflake dialect then we should switch to parsing
```

PR #182, opened 2020-04-28
[fix] statement extraction logic for view resource

ryanking:
Support CTE when extracting queries from view definitions.

When a view definition gets returned from snowflake it has the `create
view ...` at the beginning. To compare that sql statement to our state
we need to strip down to the query statement used to define the contents
of the view. Typically that statement starts with SELECT. When Common
Table Expressions (CTEs) are used WITH can start the statement.

As with some other PRs, I would love to have a proper SQL statement
parser here, but for now, just incrementally make our hacks better.

## Test Plan
* [x] acceptance tests
* [x] manual test on snowflake infra repo

## References
* https://docs.snowflake.com/en/user-guide/queries-cte.html

### DIFF
--- a/pkg/resources/view.go
+++ b/pkg/resources/view.go
@@ -3,6 +3,7 @@ package resources
 import (
 	"database/sql"
 	"fmt"
+	"log"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/resources/view.go
```diff
@@ -171,6 +169,31 @@ func ReadView(data *schema.ResourceData, meta interface{}) error {
 	return data.Set("database", databaseName.String)
 }
 
+func ExtractViewStatement(input string) string {
+	// Want to only capture the Select part of the query because before that is the Create part of the view which we no longer care about
+	targetStrings := []string{
+		" AS SELECT",
+		" AS WITH",
+		" AS (SELECT",
+		" AS (WITH",
```
jgadling, line 178: also need "AS ( WITH" ?

ryanking: does line 178 not cover that?

PR #185, opened 2020-04-29
[fix] a more robust parser for views

ryanking:
In order to extract the select statement from views we need a more robust parser.

So far it can parse all query shapes generated by this project and many others, which are relevant since they can come in via imports.

## Test Plan
* [x] acceptance tests
* [x] tested on our internal snowflake infra repo

## References
*  https://docs.snowflake.com/en/sql-reference/sql/create-view.html


pkg/resources/view.go
```diff
@@ -160,9 +160,14 @@ func ReadView(data *schema.ResourceData, meta interface{}) error {
 	}
 
 	// Want to only capture the Select part of the query because before that is the Create part of the view which we no longer care about
-	cleanString := space.ReplaceAllString(text.String, " ")
-	indexOfSelect := strings.Index(strings.ToUpper(cleanString), " AS SELECT")
-	substringOfQuery := cleanString[indexOfSelect+4:]
+
+	extractor := snowflake.NewViewSelectStatementExtractor(text.String)
+	substringOfQuery, err := extractor.Extract()
+
```
edulop91, line 166:
```suggestion
```

PR #196, opened 2020-05-19
[feature] tasks

henriblancke:
Attempt to add [task](https://docs.snowflake.com/en/user-guide/tasks-intro.html) support.

Couple of notes:
- There is no support for the copy grants option on tasks as we're never cloning or replacing an existing task
- `after` only supports tasks that are in the same schema (as per the [docs](https://docs.snowflake.com/en/sql-reference/sql/create-task.html#optional-parameters))
- This PR doesn't implement task grants but we could make it part of it?
- The task is automatically started after creation

[Travis build](https://travis-ci.com/github/henriblancke/terraform-provider-snowflake/builds/168942892)


pkg/resources/task_test.go
```diff
@@ -0,0 +1,52 @@
...
+		"sql_statement": "select hi from hello",
+		"comment":       "wow comment",
+	}
+
+	d := schema.TestResourceDataRaw(t, resources.Task().Schema, in)
+	r.NotNil(d)
+
+	WithMockDb(t, func(db *sql.DB, mock sqlmock.Sqlmock) {
+		mock.ExpectExec(
+			`^CREATE TASK "test_db"."test_schema"."test_task" WAREHOUSE = much_warehouse COMMENT = 'wow comment' AS select hi from hello$`,
```
ryanking, line 38: should the warehouse name be quoted?

pkg/snowflake/task.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// ChangeWarehouse returns the sql that will change the warehouse for the task.
+func (tb *TaskBuilder) ChangeWarehouse(newWh string) string {
+	return fmt.Sprintf(`ALTER TASK %v SET WAREHOUSE = %v`, tb.QualifiedName(), newWh)
+}
+
+// ChangeSchedule returns the sql that will change the schedule for the task.
+func (tb *TaskBuilder) ChangeSchedule(newSchedule string) string {
+	return fmt.Sprintf(`ALTER TASK %v SET SCHEDULE = '%v'`, tb.QualifiedName(), newSchedule)
```
ryanking, line 161: This (and the rest of the queries here) should probably use escaping - https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/37ea8346c3ac656088a1cc47efc4f602c9966965/pkg/snowflake/escaping.go#L7

pkg/resources/task.go
```diff
@@ -0,0 +1,478 @@
...
+	"schedule": &schema.Schema{
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "The schedule for periodically running the task. This can be a cron or interval in minutes.",
+	},
+	"session_parameters": &schema.Schema{
+		Type:        schema.TypeSet,
+		Elem:        &schema.Schema{Type: schema.TypeString},
+		Optional:    true,
+		Description: "Specifies a comma-separated list of session parameters to set for the session when the task runs. A task supports all session parameters.",
```
ryanking, line 53: It is not clear to me how this is supposed to be used?  A set of strings, each of which has a comma-separated list of paremeters?

Would `schema.TypeMap` be clearer?

pkg/resources/task.go
```diff
@@ -0,0 +1,478 @@
...
+
+	if data.HasChange("after") {
+		var (
+			q   string
+			err error
+		)
+		old, new := data.GetChange("after")
+
+		if old != "" {
+			q = builder.RemoveDependency(old.(string))
```
ryanking, line 354: It seems from the [docs](https://docs.snowflake.com/en/sql-reference/sql/alter-task.html) that removing will suspend the task and "must be restarted manually". Does adding a new `after` restart the task?

pkg/resources/task.go
henriblancke: Yes that would make more sense! It would simplify the logic of removing session parameters as well! Thanks, didn't think of `schema.TypeMap`

pkg/resources/task.go
henriblancke: Hmm, good catch! I'll take a look

henriblancke: I'm resuming the task every time `after` changes now.

ryanking:
@henriblancke it appears there is another PR covering this -> https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/198

Can you all work together on this?

chriskuchin:
@henriblancke @ryanking Happy to collaborate here the core of our 2 MRs seems very similar minus a few cosmetic differences. The biggest difference seems to be in mine I handle temporarily suspending the root task of a tree so that the child tasks can be altered and then resume it after the fact.

From the (docs)[https://docs.snowflake.com/en/sql-reference/sql/alter-task.html]:
```Modifying a standalone task requires that the task is suspended.

Modifying any task in a tree of tasks requires that the root task is suspended.
```

henriblancke:
@chriskuchin yeah your handling of task dependencies seems much better. Having the option to enable/disable the task is also nice! One thing I noticed tho is that you're not exposing `session_parameters` in your PR? Maybe we should just add that and we're good to go? It's up to @ryanking on how you all want to handle this :)

chriskuchin:
Yeah I got distracted and forgot about them. Happy to put a little more time into the PR if that's what we wanna do and add that

ryanking:
> One thing I noticed tho is that you're not exposing `session_parameters` in your PR?

Perhaps we can move forward by merging @chriskuchin's PR then adding session_params as a separate PR?

chriskuchin:
I nearly have the session_parameters piece done on my PR just testing and debugging now @ryanking

henriblancke:
@ryanking @chriskuchin I actually went ahead and added the `enabled` flag and logic to suspend root or standalone tasks before modifying / creating a child task. I think my use of `defer` to reactivate the suspended root task might be a little cleaner and safer when anything else fails during modifying or creating a new task. That way we'll always re-enable the root task? Let me know what you all think...

chriskuchin:
@henriblancke sounds good. Have you tested you session parameters? I had to modify them pretty heavily for them to actually work. Plus you weren't reading their state at all.

henriblancke:
@chriskuchin it did test em, but maybe not enough 😄? Can you elaborate? What do you mean by not reading state? Thanks!

chriskuchin:
You never read at least that I could find the parameters after you set them. The other problem I had was you were escaping the parameter values which caused them to be invalid as well as not wrapping them in quotes

pkg/resources/task.go
```diff
@@ -0,0 +1,645 @@
...
+	err = data.Set("sql_statement", t.Definition)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CreateTask implements schema.CreateFunc
+func CreateTask(data *schema.ResourceData, meta interface{}) error {
```
chriskuchin, line 318: @henriblancke you need root suspend logic in the create

pkg/resources/task.go
```diff
@@ -0,0 +1,645 @@
...
+		dbSchema = taskID.SchemaName
+		name     = taskID.TaskName
+		builder  = snowflake.Task(name, database, dbSchema)
+	)
+
+	root, err := getRootTaskAndSuspend(data, meta)
+	if err != nil {
+		return err
+	}
+	defer resumeTask(root, meta)
```
chriskuchin, line 415: @henriblancke you need to handle the case where the root was suspended

pkg/resources/task.go
```diff
@@ -0,0 +1,645 @@
...
+			return errors.Wrapf(err, "error updating task state %v", data.Id())
+		}
+		data.SetPartial("enabled")
+	}
+
+	return ReadTask(data, meta)
+}
+
+// DeleteTask implements schema.DeleteFunc
+func DeleteTask(data *schema.ResourceData, meta interface{}) error {
```
chriskuchin, line 587: @henriblancke you need to handle root task suspend/resume here as well

henriblancke:
> You never read at least that I could find the parameters after you set them. The other problem I had was you were escaping the parameter values which caused them to be invalid as well as not wrapping them in quotes

yep I did just run into the quoting issue. will look into that

chriskuchin:
Might be worthwhile to pull my acceptance test over into yours as it does a pretty good job of iterating a whole bunch of cases

pkg/resources/task.go
henriblancke: @chriskuchin it should be handled?

chriskuchin: Yeah you need to suspend the root to delete a child task

pkg/resources/task.go
henriblancke: @chriskuchin check `getRootTask` and `getRootTaskAndSuspend`. I added some logic there to not return the root task builder when the root task is already suspended that way the logic here and in other places stays concise and simple? Let me know if you think it is misleading or makes the logic hard to follow.

henriblancke:
> Might be worthwhile to pull my acceptance test over into yours as it does a pretty good job of iterating a whole bunch of cases

Agreed 👍

pkg/resources/task.go
chriskuchin: ahh nice I think though you are missing an edge case where you change multiple things including the state. i.e. i think if you were to alter some field plus change enabled -> false on a root node you would end up with enabled true at the end maybe?

pkg/resources/task.go
henriblancke: @chriskuchin thats what should be happening at line 600 here

pkg/resources/task.go
```diff
@@ -0,0 +1,645 @@
...
+	)
+	if err != nil {
+		return err
+	}
+
+	database := taskID.DatabaseName
+	schema := taskID.SchemaName
+	name := taskID.TaskName
+
+	root, err := getRootTaskAndSuspend(data, meta)
```
henriblancke, line 600: @chriskuchin here I'm suspending the root task before dropping the actual task on line 610

chriskuchin: 🤦 I must be blind

pkg/resources/task.go
henriblancke: @chriskuchin how would you handle that? It depends in what order terraform executes changes and how it builds its dag? If there is some order through tf variables, it would most likely alter the root task first before going onwards to the children or am I thinking this all the wrong way?

pkg/resources/task.go
chriskuchin: yeah you are good I just missed those lines

pkg/resources/task.go
chriskuchin: you get a single nodes updates all at once. So if the node you are modifying is the root and currently enabled You will suspend it then if there is a change to suspend the node you are on further down the method call you will go suspend it. Then your defer will trigger and reenable it

chriskuchin: The whole state machine around this took me forever and I really didn't love my code to handle it but my acceptance tests should exercise this case i believe

chriskuchin: I think you may be able to solve the problem by if you handle a change in state you set the root to nil when you handle the node state change so your resume is a no op

henriblancke: @chriskuchin added your acceptance tests

pkg/snowflake/task.go
```diff
@@ -128,7 +128,7 @@ func (tb *TaskBuilder) Create() string {
 		sort.Strings(sortedKeys)
 
 		for _, k := range sortedKeys {
-			sp = append(sp, EscapeString(fmt.Sprintf(`%v = %v`, k, tb.session_parameters[k])))
+			sp = append(sp, fmt.Sprintf(`%v = %v`, k, tb.session_parameters[k]))
```
chriskuchin, line 131: I think you also need to wrap the value in quotes since some of those values aren't valid ids

henriblancke: It's difficult here because in the case of `CLIENT_TIMESTAMP_TYPE_MAPPING = TIMESTAMP_LTZ` for example, snowflake won't recognize that if you quote it? I figured I'd leave it up to the end user here to make sure they quote what needs to be quotes?

chriskuchin: It will as long as they all caps it

chriskuchin: Since most of the other ids in teh system are auto quoted in the provider I would prefer just quoting everything

pkg/resources/task.go
henriblancke: @chriskuchin I want to understand better how you handled this edge case and how you would prevent things from being executed in parallel? I think by default terraform executes 10 operations in parallel as it walks the graph. If there is no relation that helps tf define the order when it builds the graph, there is no way to prevent it from running both operations in your edge case at the same time? Let me know if I'm looking at this in the wrong way

henriblancke: could people use `depends_on` to avoid this edge case?

chriskuchin: There are 2 edge cases 1 that i don't think you can handle which is you concurrently modify 2 nodes, but can be prevented by using references from tasks for the after clause. and the second that is if you disable an enabled task on a node that is the root node then you will reactivate it with your defer

chriskuchin: the operations from my testing happen on a node basis not a change in node basis. So you get all of a nodes changes in one call to the update method

pkg/resources/task.go
```diff
@@ -0,0 +1,645 @@
...
+
+		Schema: taskSchema,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+	}
+}
+
+// ReadTask implements schema.ReadFunc
+func ReadTask(data *schema.ResourceData, meta interface{}) error {
```
chriskuchin, line 244: >>>What do you mean by not reading state? Thanks!

Here is ReadTask you need to read the parameters to be able to add them to the state

pkg/resources/task.go
chriskuchin: depends_on or using variable references would fix the first concurrent execution problem but doesn't fix the other. Have you run my acceptance tests against your code?

chriskuchin: Also if it is easier I am happy to hop on a zoom for a quick conversation

pkg/resources/task.go
henriblancke: @chriskuchin does the `SHOW TASKS` return the `session_parameters`?

chriskuchin: no you have to do another query you can see in my PR. It's a show parameters query

pkg/resources/task.go
henriblancke: @chriskuchin lets hop on [a zoom call](https://us04web.zoom.us/j/75028769614?pwd=b2JPZWF0UXBUN1pGTE5Gd1Q4UmtmUT09)

henriblancke:
@chriskuchin any idea why [the acceptance test here](https://travis-ci.com/github/henriblancke/terraform-provider-snowflake/jobs/338886141) would be failing?

chriskuchin:
They aren't even running they error out about not being configured in Travis. Are they passing for you locally?

chriskuchin:
Ahh you have the acceptance tests hooked up to your fork... I have no idea that was never an error I saw when writing them or getting things working. Sounds like it's an error in the first step. You can enable tf_logs=debug to get more info but it's really noisy so might be hard. Can you run it locally so you run just that test with debug?

chriskuchin:
So it seems to have something to do with your state. The error occurs during a state refresh so it's probably in the read function I would guess

henriblancke:
@chriskuchin yeah just figured it out adding the id we get from snowflake during reading the task caused the error. It breaks `taskIDFromString`. I'm running into more problems tho now that this one is resolved

chriskuchin:
ahh that makes sense I didn't follow that compound id pattern in My MR 👍 let me know if i can help

ryanking:
> @ryanking not sure why the Acceptance tests aren't running in your CI but as long as those are passing the PR looks good to me

ok i will take a look

henriblancke:
@ryanking they should be all green [here](https://travis-ci.com/github/henriblancke/terraform-provider-snowflake/builds/167959488)

pkg/resources/task.go
```diff
@@ -0,0 +1,680 @@
...
+		dbSchema = taskID.SchemaName
+		name     = taskID.TaskName
+		builder  = snowflake.Task(name, database, dbSchema)
+	)
+
+	root, err := getRootTaskAndSuspend(data, meta)
+	if err != nil {
+		return err
+	}
+	defer resumeTask(root, meta)
```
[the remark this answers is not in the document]
henriblancke: Yeah we'd want to root task to be re-enabled on a TF or config failure. We don't want it to be suspended indefinitely because of an error.

pkg/resources/task.go
```diff
@@ -0,0 +1,680 @@
...
+		row := snowflake.QueryRow(db, q)
+		task, err := snowflake.ScanTask(row)
+
+		if err != nil && name != data.Get("name").(string) {
+			return nil, errors.Wrapf(err, "failed to locate the root node of: %v", name)
+		}
+
+		if task.Predecessors == nil {
+			log.Println(fmt.Sprintf("[DEBUG] found root task: %v", name))
+			// we only want to deal with suspending the root task when its enabled (started)
```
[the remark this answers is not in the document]
henriblancke: Yeah we can rename it. What do you propsose? My aim was to avoid checking the same condition at the caller levels in multiple places.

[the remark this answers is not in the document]
ryanking: Maybe getRootActiveTask ?

pkg/snowflake/task.go
```diff
@@ -0,0 +1,345 @@
...
+	return strings.ToLower(t.State) == "started"
+}
+
+func (t *task) GetPredecessorName() string {
+	if t.Predecessors == nil {
+		return ""
+	}
+
+	pre := strings.Split(*t.Predecessors, ".")
+	name, _ := strconv.Unquote(pre[len(pre)-1])
```
chriskuchin, line 311: Snowflake doesn't always wrap the name of the task in quotes and if there aren't quotes around the string then this errors and the result returned is an empty string

chriskuchin:
```suggestion
		name, err := strconv.Unquote(pre[len(pre)-1])
	if err != nil {
		return pre[len(pre)-1]
	}

```

chriskuchin: If you modify the root task's name to be all caps it will reproduce in the acceptance tests

PR #198, opened 2020-05-20
Add Tasks Support

chriskuchin:
It appears I had the same need as https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/196 A major difference between mine and that one seems to be that I will suspend a root task in a tree to enable changes or adding a task and then re enable it after the fact.

Added the `resource_task` type. Supports simple task trees and modifying child tasks by temporarily disabling the root task while modifying a child.

## Test Plan
- Wrote a combination of Unit & Internal Tests
  - Tested all helper tests and most SQL generators
* [ ] acceptance tests
 - Acceptance tests creates a 2 task tree and a standalone task and then iterates through changes verifying as it goes

## References
https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/157
### Snowflake Documentation
https://docs.snowflake.com/en/sql-reference/sql/create-task.html
https://docs.snowflake.com/en/sql-reference/sql/show-tasks.html
*


chriskuchin:
@ryanking @henriblancke Alright i borrowed heavily from the other MR for the session_parameters support. This seems to work I added pieces to my acceptance tests and they seem to work. Going to have to push one more commit with a readme update

PR #201, opened 2020-05-29
fix struct scan by exporting members of struct

chriskuchin:
The stage resource is no longer finding the properties. This change just exports the fields of the descStageRow struct.

## Test Plan
manually tested. this is currently breaking our terraform


chriskuchin:
@ryanking really simple MR that is breaking our stage terraform

PR #206, opened 2020-06-03
Don't Escape Task SQL Statements

chriskuchin:
Escaping the SQL statement causes problems when the SQL statement requires quotes and other typically escaped things.


alldoami:
Ran `make test-acceptance-ci`, and looks good.

PR #213, opened 2020-06-12
[feature] add monitor execution & execute tasks privileges

haley-roberts:


chriskuchin:
@ryanking any chance of getting some feedback here? We would like to manage these perms using terraform. Thanks

ryanking:
@chriskuchin Based on the title of the PR I was expecting some tests. Do you think some should be included?

chriskuchin:
If you think it's necessary I can work with haley to add some but the actual tests don't seem like they would be very useful or interesting based on the code changes. I think she titled it with test to try and kick off the test suite.

chriskuchin:
I will have her push a couple mockdb tests to ensure the queries come out right

haley-roberts:
@ryanking I added a couple of tests, let me know if there's anything else you would like

PR #216, opened 2020-06-23
[fix] download.sh

### LINKED ISSUE #215: download.sh breaks on zip-only releases
I'm cobbling together some CI pipelines that currently auto-download this plugin using the downloader.  (This will change soon, but give me a break, I just got it all working!)

I was sure this decision would take more than a week to come back to haunt me, but here we are - the downloader seems to be failing in a Linux AMD64 container.

The script fails thusly:

```
chanzuckerberg/terraform-provider-snowflake debug http_download https://github.com/chanzuckerberg/terraform-provider-snowflake/releases/download/v0.13.2/terraform-provider-snowflake_0.13.2_linux_amd64.tar.gz
 chanzuckerberg/terraform-provider-snowflake debug http_download_curl received HTTP status 404
```

As of this writing, there is no downloadable .tar.gz artifact for release v0.13.2.  But `download.sh` works like a champ on v0.13.0, e.g.:
```
curl https://raw.githubusercontent.com/chanzuckerberg/terraform-provider-snowflake/master/download.sh | bash -s -- -d -b $HOME/.terraform.d/plugins v0.13.0
```

By my reading of the situation, either the plugin release process should be generating tarballs on some platforms, or (probably easier) `download.sh` should be constructing .zip downloads in more situations.

ryanking:
Fix download.sh by specifying .zip and fixing the path to the checksums
file.

This broke with our changes to support releasing to the registry and
this should fix #215.

## Test Plan
* [x] acceptance tests
* [x] reproduce the problem with: `curl https://raw.githubusercontent.com/chanzuckerberg/terraform-provider-snowflake/master/download.sh | bash -s -- -d -b $HOME/.terraform.d/plugins v0.13.2
` and test the fix with `curl https://raw.githubusercontent.com/chanzuckerberg/terraform-provider-snowflake/ryanking/update_download.sh/download.sh | bash -s -- -d -b $HOME/.terraform.d/plugins v0.13.2`

## References
* Fixes #215


README.md
```diff
@@ -18,6 +18,8 @@ The easiest way is to run this command:
 curl https://raw.githubusercontent.com/chanzuckerberg/terraform-provider-snowflake/master/download.sh | bash -s -- -b $HOME/.terraform.d/plugins
 ```
 
+**Note that this will only work with recent releases, for older releaes, use the version of download.sh that corresponds to that release (replace master in that curl with the version).**
```
edulop91, line 21: should we specify a version?

ryanking: I honestly don't know which versions would work at this point.

PR #219, opened 2020-07-06
[feature] add or replace option for view creations

emancu:
`CREATE OR REPLACE VIEW` is similar, but if a view of the same name already exists, it is replaced. The new query must generate the same columns that were generated by the existing view query (that is, the same column names in the same order and with the same data types).

## Test Plan

* [x] Internal tests
* [x] acceptance tests

## References
* https://docs.snowflake.com/en/sql-reference/sql/create-view.html


pkg/resources/view.go
```diff
@@ -106,10 +116,6 @@ func CreateView(data *schema.ResourceData, meta interface{}) error {
 		builder.WithComment(v.(string))
 	}
 
-	if v, ok := data.GetOk("schema"); ok {
-		builder.WithSchema(v.(string))
-	}
-
```
emancu, line 112: The `WithSchema` method was applied in line 104 already.

emancu:
There is something odd with the acceptance-tests, that always fails.
But I noticed that is failing in ALL PRs submitted, so I think it is a bigger problem.

ryanking:
> There is something odd with the acceptance-tests, that always fails.
> But I noticed that is failing in ALL PRs submitted, so I think it is a bigger problem.

Have you gone through the steps in https://github.com/chanzuckerberg/terraform-provider-snowflake#pull-request-ci ?

emancu:
@ryanking I followed the steps and locally (as usual xD) the tests pass with my own account.
However, for this PR I can't make them work.

alldoami:
Ran acceptance tests and passed. Thanks for the work!

PR #228, opened 2020-07-29
Add pipe grant

mtranter:
Add a snowflake pipe grant resource. Currently pipes only support ALL and OWNERSHIP privilege statements. (https://docs.snowflake.com/en/user-guide/security-access-control-privileges.html#pipe-privileges)

This was almost a find and replace exercise on the stage_grant resource. 

I have added unit tests but not acceptance tests.

### DIFF
--- a/pkg/resources/pipe_grant.go
+++ b/pkg/resources/pipe_grant.go
@@ -1,0 +1,153 @@
+package resources
+
+import (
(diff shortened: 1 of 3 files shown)


alldoami:
@mtranter can you fix conflicts or close this PR?

PR #230, opened 2020-07-29
[fix] prerelease version

ryanking:
Our pre-releases were getting picked up by the registry as non-pre-releases. This will fix.

## Test Plan
* [x] acceptance tests
* [x] test release

## References
*


edulop91:
@ryanking I see an empty changeset for this one

PR #231, opened 2020-07-29
[fix] escape some strings in query builders

### LINKED ISSUE #127: Apostrophe character in comment breaks schema resource creation
If the `comment` parameter on a snowflake_schema resource contains an apostrophe, that's probably not escaped correctly during resource creation because it causes snowflake syntax errors and fails the creation.

I haven't tested other resources, but I suspect it might be also problematic elsewhere.

I'd paste the error I got, but it's a syntax error on the *next* token following the apostrophe, which isn't going to be consistent for this bug.

ryanking:
Escape comments and other string fields that are not currently escaped in our query builders.

## Test Plan

* [ ] acceptance tests

## References
* Fixes #127


pkg/snowflake/schema.go
```diff
@@ -120,12 +120,12 @@ func (sb *SchemaBuilder) Swap(targetSchema string) string {
 
 // ChangeComment returns the SQL query that will update the comment on the schema.
 func (sb *SchemaBuilder) ChangeComment(c string) string {
-	return fmt.Sprintf(`ALTER SCHEMA %v SET COMMENT = '%v'`, sb.QualifiedName(), c)
+	return fmt.Sprintf(`ALTER SCHEMA %v SET COMMENT = '%v'`, sb.QualifiedName(), EscapeString(c))
```
edulop91, line 123: why is the other QualifiedName escaped (in RemoveComment) but not this one? I can't really tell if there is a difference or not.

pkg/snowflake/view.go
```diff
@@ -139,7 +139,7 @@ func (vb *ViewBuilder) Unsecure() string {
 // Note that comment is the only parameter, if more are released this should be
 // abstracted as per the generic builder.
 func (vb *ViewBuilder) ChangeComment(c string) string {
-	return fmt.Sprintf(`ALTER VIEW %v SET COMMENT = '%v'`, vb.QualifiedName(), c)
+	return fmt.Sprintf(`ALTER VIEW %v SET COMMENT = '%v'`, vb.QualifiedName(), EscapeString(c))
```
edulop91, line 142: same here

pkg/snowflake/schema.go
ryanking: Good catch. I think qualified names should not be escaped. Or, at least, I wasn't trying to do that in this PR.

pkg/snowflake/view.go
ryanking: same

PR #239, opened 2020-08-03
[fix] skip test broken by snowflake

ryanking:
Snowflake appears to have changed something which is now causing this
test to fail. Skipping it for now while I find someone who can address
the problem.

cc @henriblancke  and @chriskuchin  who worked on this originally. Any chance one of you could look at this?

## Test Plan
* [ ] acceptance tests

## References
* https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/196

### DIFF
--- a/pkg/resources/task_acceptance_test.go
+++ b/pkg/resources/task_acceptance_test.go
@@ -157,7 +157,7 @@ var (
 )
 
 func Test_AccTask(t *testing.T) {
-	t.Skip("broken by a change to snowflake")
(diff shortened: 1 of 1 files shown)


pkg/resources/task_acceptance_test.go
```diff
@@ -157,6 +157,8 @@ var (
 )
 
 func Test_AccTask(t *testing.T) {
+	t.Skip("broken by a change to snowflake")
```
edulop91, line 160: Would you be able to link to a test with broken results or add some context(either here or the pr) so whomever works on fixing this can get a head start?

chriskuchin:
@ryanking I will see if i can find some time in the next couple of days to poke at it

PR #241, opened 2020-08-11
Chrisk/fix task tests

chriskuchin:
Snowflake recently made a change that will error if you try to activate a task without a schedule or predecessor. This fixes our test that was doing exactly this.

## Test Plan
* [x] Task Acceptance Tests pass

@ryanking This should fix that test


chriskuchin:
@ryanking the tests failed however it seems to have been a configuration error rather than the bad snowflake it was before

PR #242, opened 2020-08-20
[feature] table resource

### LINKED ISSUE #123: Add snowflake_table resource for creating and managing table objects
Add a resource to create and manage tables.  But I am curious how others are managing table objects without this resource available at this time.

johnpeterharvey:
## Summary
Resolves #123 
Addition of extra resource type `snowflake_table`, to enable terraform creation of tables in Snowflake.

The implementation is simple, allowing for the most basic creation. Can be extended with support for things like DROP CASCADE etc, but needed the basic table creation ability to be able to terraform out environments for our project.

Example usage:

```
resource "snowflake_table" "table" {
  database = "FULFILMENT"
  schema = "PUBLIC"
  name = "test"
  column {
    name = "column1"
    type = "VARCHAR"
  }
  column {
    name = "column2"
    type = "VARIANT"
  }
}
```

Output attribute data is currently only the table owner.

## Test Plan
Have added unit tests and base acceptance test in line with the test cases for other similar resources.

## References
Used https://docs.snowflake.com/en/sql-reference/sql/create-table.html for reference.

Thanks for this provider, we're finding it really useful! 🙂


robbruce:
@johnpeterharvey 

I can understand that right now your trying to get to an MVP point, but you could maybe change your schema so that this resource is extensible?

Maybe change columns to be be

```go
"column": {
	Type:     schema.TypeList,
	Required: true,
	MinItems: 1,
	ForceNew: true,
	Elem: &schema.Resource{
			"name": {
				Type:         schema.TypeString,
				Required:     true,
			},

			"type": {
				Type:         schema.TypeString,
				Required:     true,
			},
		},
	},
},
```

So that attributes can be added onto a column in the future, using a `schema.TypeMap` could be quite restrictive. A `schema.TypeList` is the only collection type which is ordered, so using that will ensure parity between the declared order and what's in Snowflake.

johnpeterharvey:
> @johnpeterharvey
> 
> I can understand that right now your trying to get to an MVP point, but you could maybe change your schema so that this resource is extensible?
> 
> Maybe change columns to be be
> 
> ```go
> "column": {
> 	Type:     schema.TypeList,
> 	Required: true,
> 	MinItems: 1,
> 	ForceNew: true,
> 	Elem: &schema.Resource{
> 		Schema: map[string]*schema.Schema{
> 			"name": {
> 				Type:         schema.TypeString,
> 				Required:     true,
> 			},
> 
> 			"type": {
> 				Type:         schema.TypeString,
> 				Required:     true,
> 			},
> 		},
> 	},
> },
> ```
> 
> So that attributes can be added onto a column in the future, using a `schema.TypeMap` could be quite restrictive. A `schema.TypeList` is the only collection type which is ordered, so using that will ensure parity between the declared order and what's in Snowflake.

Ah, thank you! Yes, that sounds like a good strategy. Appreciate the suggestion - I'm new to Golang and still learning a lot! I'll try to make that change today and update the merge request.

robbruce:
@alldoami I have a number enhancements to add to this resource, but as it's not in master currently would be tricky to issue a pull request. Do you have an indication as to when this might be reviewed?

Just set some context, our company wrote a terraform provider for snowflake but was planning on making open source once it was ready. However, after reviewing your provider we opted not continue development and contribute to this instead.

We have identified functionality that we have coded already that we are planning on converting, but for a table resource we would need this merging first (into master or another branch). All of the features listed above we have written for previously.

pkg/resources/table.go
```diff
@@ -0,0 +1,292 @@
...
+			columnDef[key] = val.(string)
+		}
+		columns = append(columns, columnDef)
+	}
+	builder := snowflake.TableWithColumnDefinitions(name, database, schema, columns)
+
+	// if len(columns) < 1 {
+	// 	return errors.New("At least one column definition is required")
+	// }
+	// builder.WithColumns(columns)
```
ryanking, line 154: uncomment or rm, please

pkg/snowflake/table.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Create returns the SQL statement required to create a table
+func (tb *TableBuilder) Create() string {
+	q := strings.Builder{}
+	q.WriteString(fmt.Sprintf(`CREATE TABLE %v`, tb.QualifiedName()))
+
+	q.WriteString(fmt.Sprintf(` (`))
+	columnDefinitions := []string{}
+	for _, columnDefinition := range tb.columns {
+		columnDefinitions = append(columnDefinitions, fmt.Sprintf(`"%v" %v`, columnDefinition["name"], columnDefinition["type"]))
```
ryanking, line 92: I think we either need to escape or validate the contents of column definitions. Otherwise we are open to injections.

ryanking:
When ready for review again, please click the "re-request review" button.

pkg/snowflake/table.go
johnpeterharvey: I've just spent time rewriting sections to make use of the bind arguments feature, to allow the SQL engine to handle escaping neatly for me, only to find that as far as I can tell you can't use it to bind column names in statements.
https://docs.snowflake.com/en/sql-reference/stored-procedures-usage.html#binding-variables
https://stackoverflow.com/questions/63025655/trying-to-pass-parameter-as-binding-variable-in-snowflake-statement

I'm unsure how best to approach adding this in, would you happen to have any suggestions of packages or approaches to look at @ryanking ? Thanks!

johnpeterharvey:
Sorry for the lack of action on this the last few weeks, had less time to revisit than I had planned. Still not really sure where to go with alterations. @ryanking , would you be able to give any guidance on my reply to your change request above please? Thanks!

pkg/snowflake/table.go
ryanking: Yes, binding arguments does not work at all with the go snowflake driver, so we do the escaping work ourselves.

johnpeterharvey: Ah, sorry @ryanking , think I was understanding at cross-purposes to what you meant. I've amended this to also run the column defs through the existing escaping function, to make it consistent with the other resources. Is that the desired result for you? Thanks!

brucedvgw:
Keen to see this come through.

pkg/resources/table.go
```diff
@@ -0,0 +1,287 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// UpdateTable implements schema.UpdateFunc
+func UpdateTable(data *schema.ResourceData, meta interface{}) error {
+	// https://www.terraform.io/docs/extend/writing-custom-providers.html#error-handling-amp-partial-state
+	data.Partial(true)
```
ryanking, line 211: If this goes in after https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/262, we need to remove partial usage.

ryanking (approved):

PR #262, opened 2020-09-25
Upgrade to terraform provider sdk v2

edulop91:
Upgrading to the v2 terraform provider sdk.

The big change here is that .Partial and .SetPartial are [deprecated/removed and discouraged](https://www.terraform.io/docs/extend/guides/v2-upgrade-guide.html#removal-of-helper-schema-resourcedata-setpartial).

Read/refresh is expected to take care of partial states.

## Test Plan

* [X] acceptance tests

* [ ] …

## References

*


Makefile
```diff
@@ -61,7 +61,7 @@ coverage: ## run the go coverage tool, reading file coverage.out
 .PHONY: coverage
 
 test: fmt deps ## run the tests
-	go test -race -coverprofile=coverage.txt -covermode=atomic $(TESTARGS) ./...
+	CGO_ENABLED=1 go test -race -coverprofile=coverage.txt -covermode=atomic $(TESTARGS) ./...
```
ryanking, line 64: what is cgo needed for?

pkg/db/db.go
```diff
@@ -7,7 +7,7 @@ import (
 	"log"
 	"regexp"
 
-	"github.com/ExpansiveWorlds/instrumentedsql"
+	"github.com/luna-duclos/instrumentedsql"
```
ryanking, line 10: Is this intended to be part of this change?

ryanking (changes requested):

Makefile
edulop91: `-race` requires it

pkg/db/db.go
edulop91: yeah, it was a fork I found to be more up-to-date since the ExpansiveWorlds project hasn't been touched since 2017 and won't compile anymore.

ryanking (approved):

pkg/resources/grant_helpers.go
```diff
@@ -301,10 +307,11 @@ func readGenericGrant(data *schema.ResourceData, meta interface{}, builder snowf
 	if err != nil {
 		return err
 	}
-	err = data.Set("shares", shares)
```
edulop91, line 304: this now panics

PR #263, opened 2020-10-02
[fix] Add shares to grant ResourceData

andybeeswax:
Grants don't recognize when privileges are already granted to a share.  So the grant resources are recreated on every apply.  This should resolve that.  The shares weren't being added to the ResourceData.

This should fix https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/226.  
It may fix https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/248 as well.

## Test Plan
* [x] acceptance tests


alldoami:
Could you write tests for this?  Thanks for the great work! 😄

andybeeswax:
@alldoami You're welcome!  Added some tests.

alldoami:
Passed `test-acceptance-ci`. Thanks @andybeeswax ! 🥳

alldoami (approved):

PR #264, opened 2020-10-02
[fix] If NULL_IF list is empty fix it with brackets

funes79:
## Test Plan

* [ ] acceptance tests

* [ x] …

## References

* This fix should avoid getting an SQL error when a stage has defined a NULL_IF = [] file format property. The provider now does not run `NULL_IF = []` but it executes `NULL_IF = ()`

### DIFF
--- a/pkg/snowflake/stage.go
+++ b/pkg/snowflake/stage.go
@@ -9,7 +9,6 @@ import (
 )
 
 func fixFileFormat(inputFileFormat string) string {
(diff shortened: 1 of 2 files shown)


funes79:
Fixes https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/254

ryanking (changes requested):

ryanking (changes requested):
Also, can we get a test for this?

ryanking (approved):

PR #271, opened 2020-10-09
[feature] Network Policy Resource

grahamplace:
Adds support for managing Snowflake Network Policy objects via the new `snowflake_network_policy` resource type.

## Test Plan

* [x] acceptance tests

* [x] tested built provider locally with trial Snowflake account

## References

* Issue: https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/221
* https://docs.snowflake.com/en/sql-reference/sql/create-network-policy.html


pkg/provider/provider.go
```diff
@@ -4,7 +4,7 @@ import (
 	"crypto/rsa"
 	"io/ioutil"
 
-	"github.com/chanzuckerberg/terraform-provider-snowflake/pkg/datasources"
+  "github.com/chanzuckerberg/terraform-provider-snowflake/pkg/datasources"
```
ryanking: You probably need to run `make fmt` on your code.

pkg/resources/network_policy.go
```diff
@@ -0,0 +1,319 @@
...
+}
+
+// CreateNetworkPolicy implements schema.CreateFunc
+func CreateNetworkPolicy(data *schema.ResourceData, meta interface{}) error {
+	db := meta.(*sql.DB)
+
+	name := data.Get("name").(string)
+	allowedIps := snowflake.IpListToString(expandStringList(data.Get("allowed_ip_list").(*schema.Set).List()))
+	blockedIps := snowflake.IpListToString(expandStringList(data.Get("blocked_ip_list").(*schema.Set).List()))
+	comment := data.Get("comment").(string)
```
ryanking, line 78: Comment needs to be escaped, see EscapeString func in the snowflake package.

pkg/resources/network_policy_acceptance_test.go
```diff
@@ -0,0 +1,55 @@
...
+	})
+}
+
+func networkPolicyConfig(name string) string {
+	return fmt.Sprintf(`
+resource "snowflake_network_policy" "test" {
+	name            = "%v"
+	comment         = "%v"
+	allowed_ip_list = ["192.168.0.100/24", "29.254.123.20"]
+	blocked_ip_list = ["192.168.0.101"]
```
ryanking, line 52: We should probably include testing adding to users here.

pkg/resources/network_policy_test.go
```diff
@@ -0,0 +1,74 @@
...
+	in := map[string]interface{}{
+		"name":            "test-network-policy",
+		"allowed_ip_list": []interface{}{"192.168.1.0/24"},
+		"blocked_ip_list": []interface{}{"155.548.2.98"},
+		"comment":         "great comment",
+	}
+	d := schema.TestResourceDataRaw(t, resources.NetworkPolicy().Schema, in)
+	r.NotNil(d)
+
+	// FIXME: this doesn't play nicely with the big post-processing SHOW/DESC sql
```
ryanking, line 37: Can we fix this before merging?

pkg/resources/network_policy.go
```diff
@@ -0,0 +1,319 @@
...
+
+	err := ensureUserAlterPrivileges(data, meta)
+	if err != nil {
+		return err
+	}
+
+	stmt := fmt.Sprintf(`CREATE NETWORK POLICY "%v" ALLOWED_IP_LIST=%v BLOCKED_IP_LIST=%v COMMENT="%v"`, name, allowedIps, blockedIps, comment)
```
ryanking, line 85: Can we move this sql generation to the snowflake package with the rest of the SQL?

ryanking (changes requested):
Thanks! This looks great.

A few minor bits of feedback and this should be good to go.

Thanks for writing acceptance tests.

ryanking:
Note that I am seeing some acceptance test failures, running `make test-acceptance` locally -

```
=== RUN   TestAccNetworkPolicy
    testing.go:683: Step 0 error: errors during apply:
        
        Error: 000606 (57P03): No active warehouse selected in the current session.  Select an active warehouse with the 'use warehouse' command.
        
        
          on /var/folders/ly/qwpw0g113mz15651mkc9wd400000gr/T/tf-test836548438/main.tf line 2:
          (source code not available)
        
        
    testing.go:744: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: errors during refresh: 000606 (57P03): No active warehouse selected in the current session.  Select an active warehouse with the 'use warehouse' command.
        
        
        State: <nil>
```

It seems that the select query needs a warehouse, which means we would need to add that to the provider configuration. In the past we've mostly avoided that and instead dealt with the the annoyances of working with just `show` and `describe`, which don't need a warehouse.

pkg/resources/network_policy_acceptance_test.go
grahamplace: I've added a `users` list to the test network policy resource definition (and a dummy user resource definition) so at least the user-granting code should be executing during this acceptance test now.

However, (as far as I can tell) I have to define:
`ImportStateVerifyIgnore: []string{"set_for_account", "users"},` 

because nowhere in `ReadNetworkPolicy` is there:
`data.Set("set_for_account", ...)` or
`data.Set("users", ...)`

This is the case because there doesn't appear to be a way in Snowflake to do "get all users with this Network Policy set" or "tell me if this network policy is set for the Snowflake account". Neither `SHOW` nor `DESC` returns any information about the active/applied status of a given Network Policy:

- Show: https://docs.snowflake.com/en/sql-reference/sql/show-network-policies.html
- Describe: https://docs.snowflake.com/en/sql-reference/sql/desc-network-policy.html

Is it acceptable to leave those out of `ReadNetworkPolicy` (and thus having to include them in `ImportStateVerifyIgnore`)?

grahamplace: FWIW this is why I had to write this resource with `ForceNew: true` for those two fields, with the "exclusive assignment" explanation in the docs — the simplest way I could figure out how to deal with this is to just always destroy/recreate and re-grant to users or re-set on account

grahamplace:
> Note that I am seeing some acceptance test failures, running `make test-acceptance` locally -
> 
> ```
> === RUN   TestAccNetworkPolicy
>     testing.go:683: Step 0 error: errors during apply:
>         
>         Error: 000606 (57P03): No active warehouse selected in the current session.  Select an active warehouse with the 'use warehouse' command.
>         
>         
>           on /var/folders/ly/qwpw0g113mz15651mkc9wd400000gr/T/tf-test836548438/main.tf line 2:
>           (source code not available)
>         
>         
>     testing.go:744: Error destroying resource! WARNING: Dangling resources
>         may exist. The full state and error is shown below.
>         
>         Error: errors during refresh: 000606 (57P03): No active warehouse selected in the current session.  Select an active warehouse with the 'use warehouse' command.
>         
>         
>         State: <nil>
> ```
> 
> It seems that the select query needs a warehouse, which means we would need to add that to the provider configuration. In the past we've mostly avoided that and instead dealt with the the annoyances of working with just `show` and `describe`, which don't need a warehouse.

Similar to my reply on your other comment — the issue here is that Snowflake's implementation of Network Policies is weird/somewhat different than other object types. 

More specifically in this case, I think we _have_ to do the `RESULT_SCAN` approach I used (which is the cause of acceptance tests user requiring a warehouse). This is because there is no way to `SHOW`/`DESC` the information about a _single_ Network Policy in Snowflake SQL:

- [DESC](https://docs.snowflake.com/en/sql-reference/sql/desc-network-policy.html) is limited to a single Network Policy, but it only returns the IP address lists: 
image

- [SHOW](https://docs.snowflake.com/en/sql-reference/sql/show-network-policies.html) _only_ allows `SHOW NETWORK POLICIES` — for some reason, unlike other object types, the output cannot be limited to a single Network Policy object via `LIKE`. Adding a `LIKE` does not limit the returned data:
image

Thus, in order to get the current `comment` on the Network Policy in question, the `SHOW` output has to be cut down to just `name = "mypolicy"` via a second query that uses `RESULT_SCAN('original_query_id')`


I worked around this by setting `DEFAULT_WAREHOUSE` on the Travis user. Not ideal obviously, but I don't have other ideas for how to implement `ReadNetworkPolicy` without the `RESULT_SCAN` approach given limitations outlined above. Open to ideas!

grahamplace:
^ while writing that I realized that I can probably just refactor to do the SHOW/DESC post-processing in Go instead of in Snowflake SQL, thus bypassing the `RESULT_SCAN` and the need for a Warehouse

Still posted as it's useful context, but I'll take a stab at that refactor now

pkg/resources/network_policy_acceptance_test.go
ryanking: Ah that makes sense!

Thinking about this a bit more, I think we should consider separating the attachment to a separate resource. This should allow us to avoid the scenario where we need to recreate the network policy resource every time one of the attachments changes. Of course the attachments may need to be recreated, but that might be tolerable. What do you think?

grahamplace: I'm on board with this — a separate attachment resource was my original plan, but I moved away from it because I (erroneously) observed that if you dropped a network policy, the users with that policy set continued to be restricted by the policy. Hence, I thought we needed to keep the attachment tightly coupled with the resource. 

However, I tested this understanding again and it turns out there's just a bit of lag (not more than a minute) between dropping the network policy and the user then being able to open a session from any IP address

So this approach _should_ work. I'll update PR

grahamplace: I guess one thing to think about is that if we do this, the `snowflake_network_policy_attachment` resource wouldn't behave like the similar *_grant resources (e.g. warehouse/database/etc) and create exclusive attachments

Because there's not a way of getting existing attachments from `SHOW`/`DESC`ing network policies/users/accounts, a network policy could be attached to a user outside of Terraform, and then changes to the `snowflake_network_policy_attachment` wouldn't destroy that attachment. That's probably fine? We should just document it

ryanking: Yeah I think that is fine given the limitations we are working with.

aidan-melen:
thanks for working on this feature @grahamplace and @ryanking!

alldoami (approved):
Ran `test-acceptancec-ci` locally and passed all the test! Thanks for the work! 🥳  I'll let @ryanking take one more look, but LGTM!

grahamplace:
Thanks for the review @alldoami! Per Ryan's [suggestion](https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/271#discussion_r504157297), I still need to refactor this to move the attachment into a separate resource. That work is on my todo list for today, so hopefully I can put that change up and tag you both for another look

pkg/resources/network_policy_acceptance_test.go
grahamplace: Done — the one thing I'm unsure about is `ReadNetworkPolicyAttachment` — given above limitations, there's no way to actually implement `schema.ReadFunc` for the attachment resource. I don't think I actually need `ReadNetworkPolicyAttachment` implemented for this, but `InternalValidate` fails without it. To get around that, I just implemented a dummy func 🤷 

lmk if there's a better way to handle this!

grahamplace:
Done! Ready for another look @ryanking @alldoami 🙂  Would appreciate input on this bit:
https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/271#discussion_r508133483

pkg/resources/network_policy_acceptance_test.go
```diff
@@ -0,0 +1,61 @@
...
+					resource.TestCheckResourceAttr("snowflake_network_policy_attachment.test", "network_policy_name", policyName),
+					resource.TestCheckResourceAttr("snowflake_network_policy_attachment.test", "set_for_account", "false"),
+					resource.TestCheckResourceAttr("snowflake_network_policy_attachment.test", "users.#", "2"),
+				),
+			},
+			// IMPORT
+			{
+				ResourceName:      "snowflake_network_policy_attachment.test",
+				ImportState:       true,
+				ImportStateVerify: false,
```
ryanking, line 34: Is there a reason for this to be false?

ryanking (changes requested):
I am currently getting acceptance test failures too:

```
pkg/resources/network_policy_attachment_acceptance_test.go:12:6: TestAccNetworkPolicyAttachment redeclared in this block
	previous declaration at pkg/resources/network_policy_acceptance_test.go:12:40
pkg/resources/network_policy_attachment_acceptance_test.go:40:6: networkPolicyAttachmentConfig redeclared in this block
	previous declaration at pkg/resources/network_policy_acceptance_test.go:40:55
```

grahamplace:
> I am currently getting acceptance test failures too:
> 
> ```
> pkg/resources/network_policy_attachment_acceptance_test.go:12:6: TestAccNetworkPolicyAttachment redeclared in this block
> 	previous declaration at pkg/resources/network_policy_acceptance_test.go:12:40
> pkg/resources/network_policy_attachment_acceptance_test.go:40:6: networkPolicyAttachmentConfig redeclared in this block
> 	previous declaration at pkg/resources/network_policy_acceptance_test.go:40:55
> ```

Ah oops, I accidentally overwrote `network_policy_acceptance_test.go` with `network_policy_attachment_acceptance_test.go`, should be corrected now

pkg/resources/network_policy_acceptance_test.go
grahamplace: I set this to false while developing because I wasn't able to get it to work, I thought due to the inability to _actually_ implement `ReadNetworkPolicyAttachment`

But it looks like it works now so switching it back to true

grahamplace: annnd I was wrong -- it's actually still not working:
```
=== RUN   TestAccNetworkPolicyAttachment
    TestAccNetworkPolicyAttachment: testing.go:683: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.

        (map[string]string) {
        }


        (map[string]string) (len=5) {
         (string) (len=19) "network_policy_name": (string) (len=10) "uowauijhoy",
         (string) (len=15) "set_for_account": (string) (len=5) "false",
         (string) (len=7) "users.#": (string) (len=1) "2",
         (string) (len=16) "users.2494182523": (string) (len=10) "test-user1",
         (string) (len=16) "users.3213322168": (string) (len=10) "test-user2"
        }

--- FAIL: TestAccNetworkPolicyAttachment (7.03s)
```

I think because `ReadNetworkPolicyAttachment` isn't really implemented?

ryanking (approved):

pkg/resources/network_policy_acceptance_test.go
grahamplace: Just confirming you saw this last comment/[commit](https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/271/commits/53a3f9a1e452f6f78245cb9de1282aec69ef890b) @ryanking? Does the merge mean you're fine skipping `ImportStateVerify` given Snowflake limitations that prevent us from show/desc'ing anything about a policy attachment on users/account?

ryanking: Yeah its fine.

grahamplace: Awesome, thanks for all the help getting this merged!

Issue #279, opened 2020-10-21
Grant SELECT on future views-tables in SCHEMA


funes79:
**Provider Version**

The provider version you are using: 0.17.1

**Terraform Version**

The version of Terraform you were using when the bug was encountered. v.0.13.4

**Describe the bug**

The provider is trying to constantly remove a `ENGINEER` role from a future select table grant, even though it is not defined any more in the tf files and it has been removed already.
```
-/+ resource "snowflake_table_grant" "in_db_table_grant_select" {
        database_name     = "IN_DB"
      ~ id                = "IN_DB|CUSTOMERDATA||SELECT|false" -> (known after apply)
        on_future         = true
        privilege         = "SELECT"
      ~ roles             = [ # forces replacement
            "LOADER",
          - "ENGINEER",
        ]
        schema_name       = "CUSTOMERDATA"
      - shares            = [] -> null
        with_grant_option = false
    }
```
When I apply these changes I see these comands (among others) in Snowflake (from newest to oldest).
```
SHOW FUTURE GRANTS IN SCHEMA "IN_DB"."CUSTOMERDATA"
GRANT SELECT ON FUTURE TABLES IN SCHEMA "IN_DB"."CUSTOMERDATA" TO ROLE "LOADER"
REVOKE SELECT ON FUTURE TABLES IN SCHEMA "IN_DB"."CUSTOMERDATA" FROM ROLE "LOADER"
REVOKE SELECT ON FUTURE TABLES IN SCHEMA "IN_DB"."CUSTOMERDATA" FROM ROLE "ENGINEER"
```
So this is OK - it revokes first the SELECT on FUTURE TABLES from `LOADER` and `ENGINEER` and then grants only `LOADER`

But then when I run the plan again, I get the same plan:
```
      ~ roles             = [ # forces replacement
            "LOADER",
          - "ENGINEER",
        ]
```
I checked in Snowflake, the `ENGINEER` **does not have SELECT ON FUTURE TABLES** by running the command:
```
SHOW FUTURE GRANTS IN SCHEMA "IN_DB"."CUSTOMERDATA"
````
It returns SELECT on VIEW but not on TABLE.  So can be this the reason of this bug? That the provider somehow interprets the Snowflake SELECT grant on VIEW as SELECT on TABLE?
So it looks to me that the provider interprets that this role has select on future table, but in reality it has not. 

**Expected behavior**

The provider should not try to remove a role when it has been already removed.

[...]

PR #281, opened 2020-10-25
[fix] future priv

funes79:
## Test Plan

* [ ] acceptance tests

* [x ] Test cases modified `table_grant_test.go` and `view_grant_test.go`

## References

* Fixes https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/279
* Some notes about the fix: as I am not really experienced with the provider I did not find any elegant solution how to exactly test that the `readGenericGrant` is executed on the `ReadTableGrant` or `ReadViewGrant` function call. So I had to reflect the builder interface but that was not enough, as the builder in both cases is FutureGrant. So I made a workaround by using a `validPrivileges` set, in VIEWS it contains only one item, the `SELECT` privilege.


ryanking (changes requested):
I think this is a reasonable approach. Only think I think should be added is an acceptance test. For that, a configuration that reproduced the problem in #279 would be great.

funes79:
Mhm, how to write and assert acceptance tests?

pkg/resources/grant_helpers.go
```diff
@@ -213,6 +214,18 @@ func readGenericGrant(data *schema.ResourceData, meta interface{}, builder snowf
 	priv := data.Get("privilege").(string)
 	grantOption := data.Get("with_grant_option").(bool)
 
+	// When onlySelect will be true, it means its a future privilege on View, thus we ignore TABLE privileges
```
ryanking, line 217: What is `onlySelect`?

pkg/resources/schema_grant_acceptance_test.go
```diff
@@ -38,6 +40,15 @@ func TestAccSchemaGrant(t *testing.T) {
...
 			},
+			// TABLE AND VIEW FUTURE GRANTS
+			{
+				Config: futureTableAndViewGrantConfig(sName, roleNameTable, roleNameView),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_view_grant.select_on_future_views", "roles.#", "1"),
+					resource.TestCheckResourceAttr("snowflake_table_grant.select_on_future_tables", "roles.#", "1")
+					resource.TestCheckResourceAttr("snowflake_view_grant.select_on_future_views", "privilege", "SELECT"),
+			//	),
+			//},
```
ryanking, line 51: This does not appear to be syntactically correct.

ryanking (changes requested):

pkg/resources/grant_helpers.go
funes79: Old comment, pushed a newer version.

pkg/resources/schema_grant_acceptance_test.go
funes79: Yes, fixed in newer commit.

funes79:
Just a side note about acceptance tests: All tests should be written with UPPERCASE ids as in my test because some clients can have `QUOTED_IDENTIFIERS_IGNORE_CASE` on Account level, so if the acceptance test creates a database `test_abc123` then the assertion fails that `TEST_ABC123` is not equal to `test_abc123`

funes79:
@ryanking please review again. Thank you

PR #291, opened 2020-10-28
[fix] use schema in show pipes

elliottohara:
use `SCHEMA` instead of `DATABASE` in show for stages and pipes.
Addresses https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/290

## Test Plan

* Current tests pass... 
* Have used in our current production environment with two workspaces where resources share names but are in different schemas. 

## References

*  https://docs.snowflake.com/en/sql-reference/sql/show-pipes.html (note use of SCHEMA or DATABASE)


alldoami:
This looks good! Can you pull in recent changes?

ryanking (approved):

Issue #292, opened 2020-10-30
Create resource snowflake_sql_script as failsafe for unsupported features


aidanmelen:
**Is your feature request related to a problem? Please describe.**

First things first, hats off to the maintainers for creating an awesome provider.

That being said, there are many features that I would like for this provider to support; such as granting all for all tables/schemas/futures or security integrations.

As a result, I am having to manually run commands like these to grant broad access to database tables and schemas:

```sql
-- grants.sql
-- table
GRANT SELECT ON ALL TABLES IN DATABASE MY_DATABASE TO ROLE MY_ROLE;
GRANT SELECT ON ALL VIEWS IN DATABASE MY_DATABASE TO ROLE MY_ROLE;
GRANT SELECT ON FUTURE TABLES IN DATABASE MY_DATABASE TO ROLE MY_ROLE;
GRANT SELECT ON FUTURE VIEWS IN DATABASE MY_DATABASE TO ROLE MY_ROLE;

-- schema
GRANT ALL ON ALL SCHEMAS IN DATABASE MY_DATABASE TO ROLE MY_ROLE;
GRANT ALL ON FUTURE SCHEMAS IN DATABASE MY_DATABASE TO ROLE MY_ROLE;
```

Today, this is simply not feasible to manage these statements with this provider. [I know I am not alone by manually running these commands along side the terraform](https://youtu.be/r4tAyiTgwRw?t=1703).

**Describe the solution you'd like**

I would like to create a failsafe resource that simply makes the user of the provider responsible for ensuring the CRUD logic with raw `snowsql` statements. I would call this the `snowflake_sql_script` resource and it would work similar to the [terraform-provider-shell](https://github.com/scottwinkler/terraform-provider-shell#using-the-provider). For example, I could have terraform manage one or many of the commands specified above e.g.

```hcl
resource "snowflake_sql_script" "grant-all-on-all-tables-in-database" {
  create = "GRANT SELECT ON ALL TABLES IN DATABASE MY_DATABASE TO ROLE MY_ROLE;"
  delete = "REVOKE SELECT ON ALL TABLES IN DATABASE MY_DATABASE ON ROLE MY_ROLE;"
}
```

This would give this provider a broad interface to run any Snowflake sql statement(s) and still have Terraform manage the CRUD operations.

**Describe alternatives you've considered**

I have tried using `table_grants` and `schema_grants` resources and they do not support my use case

**Additional context**

[...]

PR #294, opened 2020-11-02
snowflake_sql_script resource 

aidanmelen:
Related Issues:
- #292 

Successfully got the `create` and `delete` functionality working for the `snowflake_sql_script` resource.

now we can have Terraform manage the lifecycle of these grant statements

```sql
-- create
GRANT ALL ON ALL TABLES IN DATABASE TESTING_DATABASE TO ROLE TESTING_ROLE;

-- delete
REVOKE ALL ON ALL TABLES IN DATABASE TESTING_DATABASE FROM ROLE TESTING_ROLE;
```

with the following resource

```hcl
resource "snowflake_sql_script" "script" {
  name = "grant-all-on-all-tables-on-database-to-role"
  lifecycle_commands {
    create = join("", ["GRANT ALL ON ALL TABLES IN DATABASE ", snowflake_database.database.name, " TO ROLE ", snowflake_role.role.name ,";"])
    delete = join("", ["REVOKE ALL ON ALL TABLES IN DATABASE ", snowflake_database.database.name, " FROM ROLE ", snowflake_role.role.name ,";"])
  }
}
```

Here is a full example of this resource:

<details><summary>Click to show</summary>

```hcl
resource "snowflake_database" "test_database" {
	name = "test_database"
}

resource "snowflake_schema" "test_schema" {
	name = "test_schema"
	database = snowflake_database.database.name
}

resource "snowflake_table" "test_table" {
	name = "test_table"
	database = snowflake_database.database.name
	schema = snowflake_schema.schema.name
	column {
		name = "column1"
		type = "VARIANT"
	}
}

resource "snowflake_role" "test_role" {
	name = "test_role"
}

resource "snowflake_sql_script" "test_script" {
	depends_on = [
		snowflake_table.table,
	]
	name = "test_script"
	lifecycle_commands {
		create = join("", ["GRANT ALL ON ALL TABLES IN DATABASE ", snowflake_database.database.name, " TO ROLE ", snowflake_role.role.name, ";"])
		delete = join("", ["REVOKE ALL ON ALL TABLES IN DATABASE ", snowflake_database.database.name, " FROM ROLE ", snowflake_role.role.name, ";"])
	}
}
```
</details>

## Test Plan
I stubbed out the unittest and acceptance test. Having some trouble getting the unittest to mock out the `snowflake.Exec` call.


aidanmelen:
It would be inappropriate to use the `snowflake_sql_script` to manage a database, warehouse, role, or any other resource that is supported by this provider. These purpose built provider resources are more robust and are designed to handle the CRUD logic for a specific snowflake objects. Although you could use the `snowflake_sql_script` to managing these snowflake objects, it is not recommended. I created this provider resources to handle the situations and use cases that are unsupported by existing functionality. As a result, operators of this provider are required to deviate from the IaC by maintaining a few sql scripts to live along side the Terraform. Below are some examples of how operators of this provider may use the `snowflake_sql_script` resource to manage the lifecycle of snowflake objects:

1. If you have a OKTA integrated Snowflake account, then OKTA SCIM will automatically push over OKTA groups and users to corresponding Snowflake roles and users. This works great until you declare (in Terraform) to grant a TF managed service user to the OKTA SCIM managed role. The `snowflake_role_grant` resource will do what it is suppose to do, ensure that Terraform manages the grants to users for a specific role. As a result, Terraform will revoke the grants for all users granted by OKTA SCIM and will add the Terraform service users declared in Terraform. By having the `snowflake_sql_script` we could manage specific grants to a role without blowing away grants managed by OKTA SCIM.

2. It is a best practice to have all custom roles roll up into the `SYSADMIN` role as defined in the [Creating a Role Hierarchy](https://docs.snowflake.com/en/user-guide/security-access-control-configure.html#creating-a-role-hierarchy) Snowflake documentation. Just like (1), if you attempt to add a role grant to `SYSADMIN` from a role that is managed by OKTA SCIM, the Terraform will want to blow away all the user and role grants on that role. The `snowflake_sql_script` will allow admins to declare a subset of role grants for a non-terraform managed role and achieve a role hierarchy managed by Terraform.

3. The [OKTA SCIM integration requires a Snowflake Security Integration](https://docs.snowflake.com/en/user-guide/scim-okta.html#snowflake-configuration) which is currently unsupported by this provider. The Snowflake Security Integration is not a widely used resource and someone who wants it could be waiting a long time. The `snowflake_sql_script` provides a failsafe for missing features in this provider and allows users to manage their entire Snowflake Account with Terraform and not just the objects that are supported.

4. As Snowflake adds new objects, this provider will be able to manage them as soon as they are available to the console. e.g.
- [snowflake security integration](https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/277)
- snowflake organizations (BETA)
- [Streams and Tasks support](https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/157)
etc

5. Allow operators to manage a bundle of grants that would otherwise require many snowflake resources to declare the grants:
- `GRANT ALL ON ALL SCHEMAS IN DATABASE db TO ROLE role;`
- `GRANT ALL ON ALL FUTURE SCHEMA IN DATABASE db TO ROLE role;`
- `GRANT ALL ON ALL TABLES IN DATABASE db TO ROLE role;`
- `GRANT ALL ON ALL FUTURE TABLES IN DATABASE db TO ROLE role;`
- `GRANT ALL ON ALL VIEWS IN DATABASE db TO ROLE role;`
- `GRANT ALL ON ALL FUTURE VIEWS IN DATABASE db TO ROLE role;`

6. Bundling common grants with a single resource declaration. Consider this Terraform project layout:

```
snowflake
├── bin
│   ├── grants_create.sql
│   └── grants_delete.sql
└── main.tf
```

where the sql scripts contain the following content:

**grants_create.sql**
```sql
-- The snowflake_sql_script will use the as the 'C' in CRUD
GRANT ALL ON ALL SCHEMAS IN DATABASE db TO ROLE role;
GRANT ALL ON ALL FUTURE SCHEMA IN DATABASE db TO ROLE role;
GRANT ALL ON ALL TABLES IN DATABASE db TO ROLE role;
GRANT ALL ON ALL FUTURE TABLES IN DATABASE db TO ROLE role;
GRANT ALL ON ALL VIEWS IN DATABASE db TO ROLE role;
GRANT ALL ON ALL FUTURE VIEWS IN DATABASE db TO ROLE role;
```

**grants_delete.sql**
```sql
-- The snowflake_sql_script will use the as the 'D' in CRUD
REVOKE ALL ON ALL SCHEMAS IN DATABASE db ON ROLE role;
REVOKE ALL ON ALL FUTURE SCHEMA IN DATABASE db ON ROLE role;
REVOKE ALL ON ALL TABLES IN DATABASE db ON ROLE role;
REVOKE ALL ON ALL FUTURE TABLES IN DATABASE db ON ROLE role;
REVOKE ALL ON ALL VIEWS IN DATABASE db ON ROLE role;
REVOKE ALL ON ALL FUTURE VIEWS IN DATABASE db ON ROLE role;
```

then we can utilize the scripts in the `snowflake_sql_script` declaration:

```hcl
resource "snowflake_sql_script" "script" {
  name = "GRANT_VERBOSE_ACCESS"
  lifecycle_commands {
    create = file("${path.module}/bin/grants_create.sql")
    delete = file("${path.module}/bin/grants_delete.sql")
  }
}
```

In this way, we can write our CRUD logic in native SQL scripts and we can declare the orchestration of CRUD operations in just a few lines of Terraform.

pkg/snowflake/sql_script_test.go
```diff
@@ -0,0 +1,45 @@
...
+	WithMockDb(t, func(db *sql.DB, mock sqlmock.Sqlmock) {
+		mock.ExpectExec("CREATE DATABASE good_name").WillReturnResult(sqlmock.NewResult(1, 1))
+
+		err := resources.CreateSqlScript(d, db)
+		// The mock works when you call snowflake.Exec directly but not via
+		// resources.CreateSqlScript. Not sure why that is.
+		// err := snowflake.Exec(db, "CREATE DATABASE good_name")
+		r.NoError(err)
+	})
+}
```
alldoami, line 45: Do you also have a test for the `DeleteSqlScript`?

VERSION
```diff
@@ -1 +1 @@
-0.18.0
\ No newline at end of file
+0.19.0
```
alldoami, line 1: @ryanking we do releases on our own right?

aidan-melen: Sorry, I wasn't sure if I was suppose to rev the version or not.

pkg/snowflake/sql_script_test.go
aidan-melen: Not yet, but that was the plan. I was going to write the unittest for the delete lifecycle once I got the create lifecycle test mock working.

aidanmelen: @alldoami I just stubbed out the [DeleteSqlScript](https://github.com/aidanmelen/terraform-provider-snowflake/blob/main/pkg/snowflake/sql_script_test.go#L47-L71)

VERSION
ryanking: Yes, please do not change this file

aidanmelen: @ryanking it is back to 0.18.0

pkg/snowflake/sql_script_test.go
aidanmelen: @alldoami Do you know why the mock is failing? I am getting an index out of bound error. It was copied 1-to-1 from the [table_test.go](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/table_test.go#L46-L59). I think it might be failing because I have`ReadSqlScript`stubbed out to return `nil` instead of returning values from `snowflake.QueryRow(db, stmt)`

alldoami:
Hi @aidan-melen! Thanks for this PR, but I think we should keep the script out of the provider since it might bring people into trouble if they use it since it's not supported as an actual resource in Snowflake (might become convoluted to fix with little support).

aidanmelen:
> Hi @aidan-melen! Thanks for this PR, but I think we should keep the script out of the provider since it might bring people into trouble if they use it since it's not supported as an actual resource in Snowflake (might become convoluted to fix with little support).

I understand that this resource requires a greater user responsibility when compared to other resources, but there are existing resources in this provider that require user responsibility. Such as the `snowflake_database_grant` resource:

Screen Shot 2020-11-02 at 11 55 22 AM

I figured we could have a **Note** in the documentation stating something like:

> It is the user's responsibility to enforce CRUD when using the snowflake_sql_script resource. This is not a replacement for existing snowflake resources. This is intended to be used expert users who would like to manage fine grain control of Snowflake objects that are not currently supported by the provider.

mauza:
This seems like a game of trade offs. I understand the feeling that this PR shouldn't be necessary if we had all resources available and no bugs with them, but it is a useful workaround for when you need it. If you don't have this then you have to manage the resources that don't currently work in the provider outside of terraform and the remote state.

aidanmelen:
@alldoami are there any suggestions or changes that I could make to have you reconsider?

alldoami:
1. Fix the tests
2. If `Update` and `Read` don't work, take them out of the [schema list](https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/294/files#diff-66c427a5be0cf849f435c010bd68cb5da9330b2fad34d63c17f8c03b731d28c5R30-R37). Please add descriptions for `Create` and `Delete` [here](https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/294/files#diff-66c427a5be0cf849f435c010bd68cb5da9330b2fad34d63c17f8c03b731d28c5R26-R41).

cc: @ryanking

aidanmelen:
I think your gut was right

> might bring people into trouble

The reality is that this resource is fundamentally different than all the other resources in this provider. However, it is still fundamentally useful. For this reason, I decide to create a dedicated provider that utilizes the same provider.go authentication. The intention is for both providers to be used together. I think the separation will provide more clarity on the strengths of the terraform-provider-snowflake, but also why you might want to reach for the terraform-provider-snowsql.

Looking forward to contributing to this project in the future now that I have a pretty solid grasp of the inner workings.

aidanmelen:
I completed a provider for SnowSQL queries. Check out https://github.com/aidanmelen/terraform-provider-snowsql

aidanmelen:
I just created a provider that allows us to declare `GRANT ALL` statements in Terraform! Here is a code snippet from the [terraform-provider-snowsql/examples/simple/main.tf](https://github.com/aidanmelen/terraform-provider-snowsql/blob/main/examples/simple/main.tf#L66-L84).

```hcl
resource "snowsql_exec" "dcl" {
  name = local.name

  create {
    statements = <<-EOT
    GRANT ALL PRIVILEGES ON ALL TABLES IN DATABASE ${snowflake_database.database.name} TO ROLE ${snowflake_role.role.name};
    GRANT ALL PRIVILEGES ON FUTURE TABLES IN DATABASE ${snowflake_database.database.name} TO ROLE ${snowflake_role.role.name};
    EOT
  }

  delete {
    statements = <<-EOT
    REVOKE ALL PRIVILEGES ON ALL TABLES IN DATABASE ${snowflake_database.database.name} FROM ROLE ${snowflake_role.role.name};
    REVOKE ALL PRIVILEGES ON FUTURE TABLES IN DATABASE ${snowflake_database.database.name} FROM ROLE ${snowflake_role.role.name};
    EOT
  }

  delete_on_create = true
}
```

Check out the [terraform-provider-snowsql snowsql_exec resource documentation](https://registry.terraform.io/providers/aidanmelen/snowsql/latest/docs/resources/exec) for more information.

PR #298, opened 2020-11-06
Credit_quota update

andybrunner91:
Allow credit quota to be updated without destroying and recreating the resource monitor.

## Test Plan

* [ ] acceptance tests

### DIFF
--- a/pkg/resources/resource_monitor.go
+++ b/pkg/resources/resource_monitor.go
@@ -26,7 +26,6 @@ var resourceMonitorSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Computed:    true,
 		Description: "The number of credits allocated monthly to the resource monitor.",
-		ForceNew:    true,
 	},
 	"frequency": {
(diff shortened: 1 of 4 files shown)


pkg/resources/resource_monitor.go
```diff
@@ -231,6 +230,28 @@ func extractTriggerInts(s sql.NullString) ([]int, error) {
 	return out, nil
 }
 
+// UpdateResourceMonitor implements schema.UpdateFunc
+func UpdateResourceMonitor(data *schema.ResourceData, meta interface{}) error {
+	// https://www.terraform.io/docs/extend/writing-custom-providers.html#error-handling-amp-partial-state
+	data.Partial(true)
```
ryanking, line 236: I don't think you need to use partial if there is only one update happening.

ryanking (changes requested):

alldoami:
@andybrunner91 can you update PR or close?

PR #313, opened 2020-11-25
Add choices of privilege for table

kanata2:
~Resolve: https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/312~

#### Updated:
I only added an ownership privilege in this PR.
See also: #318


ryanking (changes requested):
We are [removing support for all ALL grants](https://github.com/chanzuckerberg/terraform-provider-snowflake/discussions/318), so please drop that from this PR and we can then merge it.

edulop91 (approved):

PR #320, opened 2020-12-02
[feature] ok-to-test allows testing on forked branches with secrets

alldoami:
**References:**
https://github.com/imjohnbo/ok-to-test

**Method:**
Copied over the `.github/workflows` files and modified to use the preferred method of authentication: 
`Preferred: GitHub App installation access token with contents: write and metadata: read permissions`

We had to follow [these instructions](https://docs.github.com/en/free-pro-team@latest/developers/apps/creating-a-github-app) to create a new Github App for the `chanzuckerberg` organization, not a personal account app. We were able to create and use the `app id` in the app settings and a private key after creating the app. Both are stored in Github secrets now. We changed the githash condition to be [more strict](https://github.com/chanzuckerberg/terraform-provider-snowflake/compare/adoami/ok-to-test?expand=1#diff-82454b2fc887e9985b9348b8f3bca03d2f839645a4c3ba1f3f850014d7da5c8bR49) than the [one written in the original project](https://github.com/imjohnbo/ok-to-test/blob/master/.github/workflows/integration.yml#L24-L27).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,118 @@
+# Run secret-dependent integration tests only after /ok-to-test approval
+on:
+  pull_request:
+  repository_dispatch:
+    types: [ok-to-test-command]
+
+name: Integration tests
+
+jobs:
+  # Branch-based pull request
+  integration-trusted:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
+    steps:
+
+    - name: Branch based PR checkout
+      uses: actions/checkout@v2
+
+    # Insert integration tests needing secrets
+    strategy:
+      matrix:
+        target:
+          - check-docs
+          - check-mod
+          - lint-ci
+          - test-acceptance-ci
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.15.2'
+      - name: Install dependencies
+        run: make setup
+
+      - name: make ${{ matrix.target }}
+        env:
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.REVIEWDOG_GITHUB_API_TOKEN }}
+          SNOWFLAKE_USER: ${{ secrets.SNOWFLAKE_USER }}
+          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
+          SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
+          SNOWFLAKE_ROLE: ${{ secrets.SNOWFLAKE_ROLE }}
+        run: make ${{ matrix.target }}
+
+  # Repo owner has commented /ok-to-test on a (fork-based) pull request
+  integration-fork:
+    runs-on: ubuntu-latest
+    if:
+      github.event_name == 'repository_dispatch' &&
+      github.event.client_payload.slash_command.sha == github.event.client_payload.pull_request.head.sha
+    steps:
+
+    # Check out merge commit
+    - name: Fork based /ok-to-test checkout
+      uses: actions/checkout@v2
+      with:
+        ref: 'refs/pull/${{ github.event.client_payload.pull_request.number }}/merge'
+
+    # Integration tests needing secrets
+    strategy:
+      matrix:
+        target:
+          - check-docs
+          - check-mod
+          - lint-ci
+          - test-acceptance-ci
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.15.2'
+      - name: Install dependencies
+        run: make setup
+
+      - name: make ${{ matrix.target }}
+        env:
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.REVIEWDOG_GITHUB_API_TOKEN }}
+          SNOWFLAKE_USER: ${{ secrets.SNOWFLAKE_USER }}
+          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
+          SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
+          SNOWFLAKE_ROLE: ${{ secrets.SNOWFLAKE_ROLE }}
+        run: make ${{ matrix.target }}
+
+    - run: |
+        echo "Integration tests... success! ;-)"
+
+    # Update check run called "integration-fork"
+    - uses: actions/github-script@v1
+      id: update-check-run
+      if: ${{ always() }}
+      env:
+        number: ${{ github.event.client_payload.pull_request.number }}
+        job: ${{ github.job }}
+        # Conveniently, job.status maps to https://developer.github.com/v3/checks/runs/#update-a-check-run
+        conclusion: ${{ job.status }}
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
+          const { data: pull } = await github.pulls.get({
+            ...context.repo,
+            pull_number: process.env.number
+          });
+          const ref = pull.head.sha;
+
+          const { data: checks } = await github.checks.listForRef({
+            ...context.repo,
+            ref
+          });
+
+          const check = checks.check_runs.filter(c => c.name === process.env.job);
+
+          const { data: result } = await github.checks.update({
+            ...context.repo,
+            check_run_id: check[0].id,
+            status: 'completed',
+            conclusion: process.env.conclusion
+          });
+
+          return result;
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -10,6 +10,7 @@ jobs:
   # Branch-based pull request
   integration-trusted:
     runs-on: ubuntu-latest
+    # Runs tests when a PR is opened from the original repo (not a forked repo), which protects the secrets and builds for trusted contributors
     if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
     strategy:
       matrix:
@@ -39,6 +40,7 @@ jobs:
   integration-fork:
     runs-on: ubuntu-latest
     if:
+    # Strict rule to check the latest commit sha with the one provided in the ok-to-test command
       github.event_name == 'repository_dispatch' &&
       github.event.client_payload.slash_command.sha == github.event.client_payload.pull_request.head.sha
     # Integration tests needing secrets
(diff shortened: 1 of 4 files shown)


.github/workflows/integration.yml
```diff
@@ -0,0 +1,118 @@
...
+  # Branch-based pull request
+  integration-trusted:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
+    steps:
+
+    - name: Branch based PR checkout
+      uses: actions/checkout@v2
+
+    # Insert integration tests needing secrets
```
ryanking, line 19:
```suggestion
```

.github/workflows/integration.yml
```diff
@@ -0,0 +1,118 @@
+# Run secret-dependent integration tests only after /ok-to-test approval
+on:
+  pull_request:
+  repository_dispatch:
+    types: [ok-to-test-command]
+
+name: Integration tests
+
+jobs:
+  # Branch-based pull request
```
ryanking, line 10: are there other kinds of PRs?

ryanking: does this mean not a fork?

.github/workflows/ok-to-test.yml
```diff
@@ -0,0 +1,34 @@
...
+        app_id: ${{ secrets.OK_TO_TEST_APP_ID }}
+        private_key: ${{ secrets.OK_TO_TEST_PRIVATE_KEY }}
+
+    - name: Slash Command Dispatch
+      uses: peter-evans/slash-command-dispatch@v1
+      env:
+        TOKEN: ${{ steps.generate_token.outputs.token }}
+      with:
+        token: ${{ env.TOKEN }} # GitHub App installation access token
+        # token: ${{ secrets.PERSONAL_ACCESS_TOKEN }} # PAT or OAuth token will also work
```
ryanking, line 29: rm if we are not using

.github/workflows/unit.yml
```diff
@@ -26,3 +31,4 @@ jobs:
           SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
           SNOWFLAKE_ROLE: ${{ secrets.SNOWFLAKE_ROLE }}
```
ryanking, line 32: rm these secrets, since we should be sure they are not being used in unit tests

also rm `test-acceptance-ci` from targets and replace with `test`

ryanking (changes requested):

.github/workflows/integration.yml
alldoami: From the example, it works for both forked PRs and branch-based PRs: https://github.com/imjohnbo/ok-to-test#example

.github/workflows/integration.yml
```diff
@@ -0,0 +1,117 @@
...
+      uses: actions/checkout@v2
+
+    strategy:
+      matrix:
+        target:
+          - check-docs
+          - check-mod
+          - lint-ci
+          - test
+    steps:
```
edulop91, line 26: hmm `steps` duped? is that intended?

.github/workflows/ok-to-test.yml
```diff
@@ -0,0 +1,33 @@
...
+    # See app.yml for an example app manifest
+    - name: Generate token
+      id: generate_token
+      uses: tibdex/github-app-token@v1
+      with:
+        app_id: ${{ secrets.OK_TO_TEST_APP_ID }}
+        private_key: ${{ secrets.OK_TO_TEST_PRIVATE_KEY }}
+
+    - name: Slash Command Dispatch
+      uses: peter-evans/slash-command-dispatch@v1
```
edulop91, line 24: there's a v2 for this action, I wonder if worth upgrading (either now or subsequent pr)

.github/workflows/unit.yml
```diff
@@ -1,15 +1,19 @@
-on: push
+# Run unit tests on any branch/fork pull request
+on:
+  pull_request
```
edulop91, line 3: not sure if worth specifying the `types` of pull_request events. See https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request and https://frontside.com/blog/2020-05-26-github-actions-pull_request/

.github/workflows/unit.yml
```diff
@@ -19,10 +23,5 @@ jobs:
...
 
       - name: make ${{ matrix.target }}
-        env:
-          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.REVIEWDOG_GITHUB_API_TOKEN }}
-          SNOWFLAKE_USER: ${{ secrets.SNOWFLAKE_USER }}
-          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
-          SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
-          SNOWFLAKE_ROLE: ${{ secrets.SNOWFLAKE_ROLE }}
         run: make ${{ matrix.target }}
+    - run: echo "Unit tests... success! ;-)"
```
edulop91, line 27: probably don't need these

.github/workflows/unit.yml
```diff
@@ -1,15 +1,19 @@
-on: push
+# Run unit tests on any branch/fork pull request
```
edulop91, line 1: maybe worth being explicit about these tests not having access to any secrets

.github/workflows/ok-to-test.yml
```diff
@@ -0,0 +1,33 @@
...
+    types: [created]
+
+jobs:
+  ok-to-test:
+    runs-on: ubuntu-latest
+    steps:
+    # Generate a GitHub App installation access token from an App ID and private key
+    # To create a new GitHub App:
+    #   https://developer.github.com/apps/building-github-apps/creating-a-github-app/
+    # See app.yml for an example app manifest
```
edulop91, line 15: I think these might be out of context here

.github/workflows/integration.yml
```diff
@@ -0,0 +1,117 @@
...
+    if:
+      github.event_name == 'repository_dispatch' &&
+      github.event.client_payload.slash_command.sha == github.event.client_payload.pull_request.head.sha
+    steps:
+
+    # Check out merge commit
+    - name: Fork based /ok-to-test checkout
+      uses: actions/checkout@v2
+      with:
+        ref: 'refs/pull/${{ github.event.client_payload.pull_request.number }}/merge'
```
edulop91, line 55: same here with the duplicate `steps` key, worth merging them?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,117 @@
...
+          const check = checks.check_runs.filter(c => c.name === process.env.job);
+
+          const { data: result } = await github.checks.update({
+            ...context.repo,
+            check_run_id: check[0].id,
+            status: 'completed',
+            conclusion: process.env.conclusion
+          });
+
+          return result;
```
edulop91, line 117: I might be wrong but it seems like the indenting of this block makes it so that it is not under the `steps` key, does it need an extra indent?

.github/workflows/ok-to-test.yml
alldoami: I think we should try using `v1` first and then try with `v2` since I'm expecting some other things might need to change besides the version number.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,102 @@
...
+  repository_dispatch:
+    types: [ok-to-test-command]
+
+name: Integration tests
+
+jobs:
+  # Branch-based pull request
+  integration-trusted:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
```
edulop91, line 13: I wonder if it is worth adding in a comment here explaining what this does and why it is important for "trusted builds"

.github/workflows/integration.yml
```diff
@@ -0,0 +1,102 @@
...
+        target:
+          - check-docs
+          - check-mod
+          - lint-ci
+          - test
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: 'refs/pull/${{ github.event.client_payload.pull_request.number }}/merge'
+      - uses: actions/checkout@v2
```
edulop91, line 56: you already have a checkout right above, do you need this one?

.github/workflows/ok-to-test.yml
```diff
@@ -0,0 +1,29 @@
...
+      uses: peter-evans/slash-command-dispatch@v1
+      env:
+        TOKEN: ${{ steps.generate_token.outputs.token }}
+      with:
+        token: ${{ env.TOKEN }} # GitHub App installation access token
+        reaction-token: ${{ secrets.GITHUB_TOKEN }}
+        issue-type: pull-request
+        commands: ok-to-test
+        named-args: true
+        permission: write
```
edulop91, line 29: I wonder how we verify this is enforced

.github/workflows/unit.yml
```diff
@@ -1,15 +1,20 @@
-on: push
+# Run unit tests that don't require secrets on any branch/fork pull request
+on:
+  pull_request:
+    types: [review_requested, edited]
```
edulop91, line 4:
```suggestion
    types: [review_requested, edited, synchronized]
```
I think so?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,102 @@
...
+    runs-on: ubuntu-latest
+    if:
+      github.event_name == 'repository_dispatch' &&
+      github.event.client_payload.slash_command.sha == github.event.client_payload.pull_request.head.sha
+    # Integration tests needing secrets
+    strategy:
+      matrix:
+        target:
+          - check-docs
+          - check-mod
```
edulop91, line 49: some of these are duplicated in the ci.yml, if they work there we shouldn't duplicate here

.github/workflows/integration.yml
```diff
@@ -0,0 +1,102 @@
...
+          - lint-ci
+          - test
```
edulop91, line 51:
```suggestion
          - test-acceptance-ci
```
is that the target that runs the acceptance tests?

alldoami: https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/320#discussion_r534560084

edulop91: yep, but you still want the `test-acceptance-ci` for the workflow that requires secrets. But only the `test` for the workflow with no secrets. If you look at https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/Makefile#L68, it has a `TF_ACC=1` which indicates we can actually go and create real resources (run acceptance tests).

edulop91: I'm actually not super clear why we have both `test-acceptance` and `test-acceptance-ci` I vote we rm `test-acceptance-ci` and just keep `test-acceptance` ?

.github/workflows/integration.yml
alldoami: I removed the `ci.yml` file, is that ok?

edulop91: oh yeah, I guess the now `unit.yml`. I think whatever runs there (without secrets) shouldn't be here to simplify and avoid duplicating work

edulop91 (approved):
lgtm :tada: 
idk if @ryanking wants to take a pass

PR #322, opened 2020-12-04
[breaking] Remove the ALL grant

edulop91:
Removing the ALL grant. See https://github.com/chanzuckerberg/terraform-provider-snowflake/discussions/318. I will follow up with more documentation in a separate PR

## Test Plan

* [X] acceptance tests

## References

*


pkg/resources/privileges.go
```diff
@@ -1,90 +1,77 @@
 package resources
 
-type privilege string
+type Privilege string
```
edulop91, line 3: made these public for a follow-up refactor I'm planning, apologies for the noise

alldoami (approved):

PR #326, opened 2020-12-07
[fix] storage_integration alter type

a129722:
This PR fixes https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/325.  While the type of EXTERNAL_STAGE must be supplied when creating a storage integration, the type does not actually appear in the `SHOW INTEGRATIONS` result set.  This causes the provider to try and alter the integration, which will fail as the type cannot be altered after the fact.

This change removes the altering of the storage integration.

### DIFF
--- a/pkg/resources/storage_integration.go
+++ b/pkg/resources/storage_integration.go
@@ -28,6 +28,7 @@ var storageIntegrationSchema = map[string]*schema.Schema{
 		Optional:     true,
 		Default:      "EXTERNAL_STAGE",
 		ValidateFunc: validation.StringInSlice([]string{"EXTERNAL_STAGE"}, true),
+		ForceNew: true,
 	},
 	"enabled": {
 		Type:     schema.TypeBool,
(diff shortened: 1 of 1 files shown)


slocke716:
Thanks you for doing this! This has become a problem for us again today since for some reason the lifecycle ignore of type is not working now

pkg/resources/storage_integration.go
```diff
@@ -236,11 +236,6 @@ func UpdateStorageIntegration(data *schema.ResourceData, meta interface{}) error
 		stmt.SetString("COMMENT", data.Get("comment").(string))
 	}
 
-	if data.HasChange("type") {
```
edulop91, line 239: If the type can't be changed I think we also should indicate that on the [schema for this resource](https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/326/files#diff-d93d3a19e447fb8fa7b3c85eba8698e994215c2a5b1d35f9e44299dc21189597R26) by setting `ForceNew: true`

edulop91 (commented):
thanks for the fix! I suggest adding one small addition to reflect this constraint at the resource schema level.

pkg/resources/storage_integration.go
a129722: Done.

edulop91:
https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/346 merged here, thanks!

PR #339, opened 2020-12-08
[fix] Remove resource from state file if its not found

### LINKED ISSUE #188: "sql: no rows in result set" error
We occasionally encounter this error when trying to run `plan`/`refresh` after a failed `apply` (e.g. can happen if the computer running `apply` gets terminated or if snowflake throws an error in the middle of an apply, like if we're trying to make terraform create an object that already exists...there are many reasons why an `apply` will error out without cleaning itself up): 

```
$ terraform refresh
... (a bunch of "Refreshing state..." lines)

Error: sql: no rows in result set
```

We typically remedy the issue by deleting grants from the state that don't exist in Snowflake, but it can be extremely difficult to manually find those instances since our terraform state has thousands of resources.

I think there's some grant-related code in the provider that is running a SQL query to get a list of grants and just blindly expecting the result to be non-empty.  I think the provider needs to handle this more elegantly, because I would hope that terraform `refresh` should be more robust since it is a prerequisite to running terraform `plan` and `apply`.

**Aside:** It is no surprise that the state is corrupt, since we ARE talking about the `apply` unexpectedly exiting or getting terminated.  So I agree that there are other provider bugs and issues with our infrastructure which must first exist and have happened before we encounter this bug, but I would rather treat those as separate bugs.  Since this happens to us so often and it's always the same error, I feel like fixing this is the low-hanging fruit.

lgwacker:
This PR is a redo of https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/305, since it was missing tests and some resources were not working.

Fixes #188


lgwacker:
@ryanking Can we merge this asap? My environment is kinda messy because of this issue. Would help a lot if we can add this to the latest release.

pkg/resources/managed_account_test.go
```diff
@@ -39,6 +40,19 @@ func TestManagedAccountCreate(t *testing.T) {
...
 
+func TestManagedAccountRead(t *testing.T) {
+	r := require.New(t)
+	d := managedAccount(t, "test-account", map[string]interface{}{"name": "test-account"})
+	q := snowflake.ManagedAccount(d.Id()).Show()
+
+	WithMockDb(t, func(db *sql.DB, mock sqlmock.Sqlmock) {
+		// Test when resource is not found
+		mock.ExpectQuery(q).WillReturnError(sql.ErrNoRows)
+		err := resources.ReadManagedAccount(d, db)
```
ryanking, line 51: I think we should find some way to assert that after these Read functions are called that the object has been removed from the state.

pkg/resources/managed_account.go
```diff
@@ -128,6 +128,12 @@ func ReadManagedAccount(d *schema.ResourceData, meta interface{}) error {
 	stmt := snowflake.ManagedAccount(id).Show()
 	row := snowflake.QueryRow(db, stmt)
 	a, err := snowflake.ScanManagedAccount(row)
+
+	if err == sql.ErrNoRows {
+		log.Printf("[WARN] managed account (%s) not found, removing from state file", d.Id())
```
ryanking, line 133: I think it will only be removed form the state file if running a refresh or apply operation, so the log message should avoid suggesting that it is being remove durably.

ryanking (changes requested):

lgwacker:
@ryanking I don't get why this check error is happening. go.mod is exactly like the one at the master branch.

pkg/resources/managed_account.go
ryanking: The level is fine, I just think that it would be more clear if we didn't state that it was removed from the state file, since it might only be removed from the transient, in-memory state used for a plan.

ryanking (approved):

PR #343, opened 2020-12-09
[fix] check-docs hacky fix for bug in terraform-plugin-docs

ryanking:
See upstream [bug report](https://github.com/hashicorp/terraform-plugin-docs/issues/12).

## Test Plan
* [ ] CI

## References
* https://github.com/hashicorp/terraform-plugin-docs/issues/12


alldoami (approved):
🔨

PR #356, opened 2020-12-10
[fix] schema grant acceptance test

ryanking:
Enable share-related tests that have never worked. The nature of the fix means that we will not be running tests that add accounts to shares, but since those have never been running in CI anyways, this seems acceptable.

## Test Plan
* [ ] acceptance tests

### DIFF
--- a/pkg/resources/database_grant.go
+++ b/pkg/resources/database_grant.go
@@ -81,7 +81,7 @@ func CreateDatabaseGrant(d *schema.ResourceData, meta interface{}) error {
 
 	err := createGenericGrant(d, meta, builder)
 	if err != nil {
-		return errors.Wrap(err, "error createing database grant")
+		return errors.Wrap(err, "error creating database grant")
 	}
 
 	grant := &grantID{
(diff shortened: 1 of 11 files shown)


pkg/resources/database_grant.go
```diff
@@ -80,7 +81,7 @@ func CreateDatabaseGrant(d *schema.ResourceData, meta interface{}) error {
 
 	err := createGenericGrant(d, meta, builder)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "error createing database grant")
```
alldoami, line 84:
```suggestion
		return errors.Wrap(err, "error creating database grant")
```

pkg/resources/database_grant.go
```diff
@@ -90,7 +91,7 @@ func CreateDatabaseGrant(d *schema.ResourceData, meta interface{}) error {
 	}
 	dataIDInput, err := grant.String()
 	if err != nil {
-		return err
+		return errors.Wrap(err, "error createing database grant")
```
alldoami, line 94:
```suggestion
		return errors.Wrap(err, "error creating database grant")
```

pkg/resources/view.go
```diff
@@ -94,6 +94,8 @@ func View() *schema.Resource {
 
 // CreateView implements schema.CreateFunc
 func CreateView(d *schema.ResourceData, meta interface{}) error {
+	log.Println("[DEBUG] XXX")
+
```
edulop91, line 98:
```suggestion
```

edulop91 (approved):

PR #257, opened 2020-09-17
[feature] masking policy schema grant

mhmcdonald:
This PR adds a missing schema grant of "CREATE MASKING POLICY".  

This Snowflake documentation shows this is a valid privilege in the schemaPrivileges section:
https://docs.snowflake.com/en/sql-reference/sql/grant-privilege.html#syntax

## Test Plan
I've ran the `make test` locally. This privilege is no different than any other schema privileges. 

## References
This should resolve this issue, which I opened: 
https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/256

PR #381, opened 2020-12-24
[Feature] Add apply masking policy grant on account

williampn:
v0.20.0 introduced grants to create masking policies per PR #257, but on top of being able to create masking policies, one should be able to apply them at the account level as per doc https://docs.snowflake.com/en/user-guide/security-column-ddm-use.html#step-1-grant-masking-policy-privileges-to-custom-role.

The following Snowflake documentation shows this is a valid privilege in the globalPrivileges section:
https://docs.snowflake.com/en/sql-reference/sql/grant-privilege.html#syntax

## Test Plan
I've ran the make test locally. This privilege is no different than any other already existing account privilege.


williampn:
#396 as a replacement

PR #391, opened 2021-01-04
Putting if in bash vs in action step

alldoami:


.github/workflows/integration.yml
```diff
@@ -27,10 +27,15 @@ jobs:
...
           echo \"${{ github.event.client_payload.slash_command.args.named.sha }}\"
           SHAINPUT=$(echo ${{github.event.client_payload.slash_command.args.named.sha}} | cut -c1-7)
-          echo ${#SHAINPUT}
           if [ ${#SHAINPUT} -le 6 ]; then echo "error::input sha not at least 7 characters long" ; exit 1
           else echo "done"
           fi
+          SHAHEAD=$(echo ${{github.event.client_payload.pull_request.head.sha}} | cut -c1-7)
+          echo ${#SHAINPUT}
+          echo ${#SHAHEAD}
+          if [ "${SHAHEAD}" != "${SHAINPUT}" ]; then echo "sha input from slash command does not equal the head sha" ;
```
edulop91, line 36: should this also fail?

alldoami: ah yes! sorry took it out for testing

alldoami:
Putting this here is for added security from my last PR when I removed this criteria from the initial if statement, but I don't think it will fix the problem of not running on forked PRs. It looks like the `integration tests` are running only on `master` and not on the sha provided in the `ok-to-test` slash command. You can see that the shas provided in the logs are the ones in the slash command, but the code that is built is the commit most recently merged to master. Its status also isn't reflected in the PR itself.

edulop91:
> Putting this here is for added security from my last PR when I removed this criteria from the initial if statement, but I don't think it will fix the problem of not running on forked PRs. It looks like the `integration tests` are running only on `master` and not on the sha provided in the `ok-to-test` slash command. You can see that the shas provided in the logs are the ones in the slash command, but the code that is built is the commit most recently merged to master. Its status also isn't reflected in the PR itself.

Is this what the checkout + merge step is supposed to be taking care of?

edulop91 (approved):

alldoami:
@edulop91 looking at this: https://github.com/actions/checkout, it looks like we can supply the sha we give in the slash command to build on it instead of the commit that was recently merged

PR #380, opened 2020-12-23
simplify id codec code

ryanking:
Refactor to simplify and DRY up our code for generating IDs.

## Test Plan
* [ ] acceptance tests

PR #392, opened 2021-01-04
[feature] file format resource

ajbosco:
## Test Plan

* [x] acceptance tests

* [x] unit tests

## References

*


pkg/resources/file_format.go
```diff
@@ -0,0 +1,1018 @@
...
+		Description: "Boolean that specifies whether the XML parser disables automatic conversion of numeric and Boolean values from text to native representation.",
+	},
+	"comment": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies a comment for the file format.",
+	},
+}
+
+type fileFormatID struct {
```
ryanking, line 214: I refactored the way we encode and decode IDs recently in #380. You should be able to use that to drop this.

pkg/resources/file_format.go
```diff
@@ -0,0 +1,1018 @@
...
+
+		Schema: fileFormatSchema,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+	}
+}
+
+// CreateFileFormat implements schema.CreateFunc
+func CreateFileFormat(data *schema.ResourceData, meta interface{}) error {
```
ryanking, line 277: Did you consider providing some validation on field / parameter compatibility? Not all format types support all parameters. As it stands, if a user were to use an incompatible combination, they would get an sql error from snowflake, which might be confusing.

ryanking (changes requested):

pkg/resources/file_format.go
ajbosco: Just added some validation to the Create method.

pkg/resources/file_format.go
ajbosco: Updated

pkg/resources/file_format.go
ajbosco: The validation gets messier though since each Format Type has their own defaults too. Thinking they should each be separate resources to make it easier to take advantage of `ValidateFunc`. Thoughts?

ryanking: Unfortunately there is no good place to do cross-attribute validations except for in the Create and Update functions. ValidateFunc works at the scope of a single attribute.

I looked at two paths of tackling this awhile ago-

1. create a separate resource for each file format type. Probably ok since you can't change the format type anyway. Would add a lot of resources with a fair amount of duplication of code.

2. rather than take flat approach (which you are doing), I tried creating a structured type for each format. My WIP code is in https://github.com/chanzuckerberg/terraform-provider-snowflake/tree/ryanking/file_format. I gave up because I was trying to do something too hard - using the presence of and structure to signal the type (even if empty). Terraform doesn't do well with empty / zero stuff.

I think maybe a hybrid approach of 2 and what you are doing might work - create a separate structured attribute for each type, and also have a type attribute. Then we get some deal of validation via the type system.

Make sense?

ajbosco: Yep, I think that hybrid approach makes sense. I'll see if I can sort it out.

acompa: @ryanking I'll be picking this work up from @ajbosco in the coming days. Would (1) above be acceptable as an approach to this problem? I see how lots of resources would be added, but the approach is more straightforward (if verbose) to implement.

ryanking: I think we should either take option 2 or revert to @ajbosco's original approach of a flat list of options.

alldoami:
@ajbosco can you fix the conflicts and i'll take a look after?

elliottohara:
Any status on this? Would love the resource!

philippeboyd:
Any news on this?

acompa:
@ajbosco and I deprioritized this work, but anyone else can feel free to pick up this branch and finalize.

PR #398, opened 2021-01-07
Remove matrix for tests

alldoami:
Trying inline tests instead of matrix to see how github actions responds to forked PRs.


ryanking (approved):

edulop91 (approved):
let's try it

PR #404, opened 2021-01-08
Updating comments even on failures

alldoami:


.github/workflows/integration.yml
```diff
@@ -68,14 +68,46 @@ jobs:
...
             Integration tests ran for [${{ github.event.client_payload.slash_command.args.named.sha }}](https://github.com/chanzuckerberg/terraform-provider-snowflake/actions/runs/${{ github.run_id }})
-          reactions: '+1'
+
+      - name: create comment
+        if: (${{ failure() }} || ${{ cancelled() }} || ${{ success() }}) && ${{ steps.fc.outputs.comment-id == 0 }} && github.event_name == 'pull_request'
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          issue-number: ${{ github.event.client_payload.pull_request.number }}
+          body: |
+            Integration tests ran for [${{ github.event.pull_request.head.sha }}](https://github.com/chanzuckerberg/terraform-provider-snowflake/actions/runs/${{ github.run_id }})
```
edulop91, line 94: ok to leave out the pul_request ones since they get linked automatically in the tests?

.github/workflows/integration.yml
```diff
@@ -68,14 +68,46 @@ jobs:
...
+        if: ${{ always() }}
+        uses: peter-evans/find-comment@v1
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: Integration tests ran for
+
+      - name: create comment
+        if: (${{ failure() }} || ${{ cancelled() }} || ${{ success() }}) && ${{ steps.fc.outputs.comment-id == 0 }} && (github.event_name == 'repository_dispatch')
```
edulop91, line 81: huh, surprised all of these failure, cancelled, scucess work but... always doesnt?

alldoami: i'm not sure if it works... but it might? I feel like it's more specific than always so it doesn't just ignore everything else?

.github/workflows/integration.yml
alldoami: ah yes you're right, I was thinking this but got distracted.

edulop91 (approved):

PR #407, opened 2021-01-08
[fix] Revert "simplify id codec code (#380)"

edulop91:
This reverts commit ccce88c5c5bc766bb2b73db6a0a90a326290e53a.

Seeing a bunch of:
>Error: wrong number of fields in record

I think reverting this commit should fix things temporarily.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


ryanking:
hmm I thought I fixed all of those

PR #412, opened 2021-01-13
[fix] ok-to-test: Report status checks on failed jobs

edulop91:
ok to test would not report job status on failed jobs. change status check conditional to always run (even after failed tests)

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -88,7 +88,7 @@ jobs:
 
       - name: set fork job status
         uses: actions/github-script@v3
-        if: github.event_name == 'repository_dispatch'
+        if: ${{ always() }}
         id: update-check-run
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -88,17 +88,23 @@ jobs:
 
       - name: set fork job status
         uses: actions/github-script@v3
-        if: github.event_name == 'repository_dispatch'
+        if: ${{ always() }}
```
ryanking, line 91: why leave the if?

ryanking (approved):

.github/workflows/integration.yml
edulop91: GHA defaults to "success()", so this step wouldn't run if previous failed otherwise

PR #429, opened 2021-01-22
Fix import format for snowflake_database_grant

kanata2:
## Description
According to the following code, we must pass 4 or 5 pipe-delimited values in order to use `terraform import`, but the documentation says that you should pass 3 values.
Indeed, `terraform import snowflake_database_grant.xxx 'xxx|||USAGE|false' works correctly, so I fixed the documentation.

https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/a4240bb6bd399141f3614b3a5e1ff54476de70b4/pkg/resources/grant_helpers.go#L114-L116

### Current Behaviour

```
$ terraform import snowflake_database_grant.xxx 'xxx|||USAGE|false'
...
snowflake_database_grant.xxx: Refreshing state... [id=xxx|USAGE|false]

Error: 4 or 5 fields allowed
```

### DIFF
--- a/docs/resources/database_grant.md
+++ b/docs/resources/database_grant.md
@@ -42,6 +42,6 @@ resource snowflake_database_grant grant {
 Import is supported using the following syntax:
 
 ```shell
-# format is database name | privilege | true/false for with_grant_option
-terraform import snowflake_database_grant.example 'databaseName|USAGE|false'
+# format is database name | | | privilege | true/false for with_grant_option
+terraform import snowflake_database_grant.example 'databaseName|||USAGE|false'
 ```
--- a/docs/resources/schema_grant.md
+++ b/docs/resources/schema_grant.md
@@ -46,6 +46,6 @@ resource snowflake_schema_grant grant {
 Import is supported using the following syntax:
 
 ```shell
-# format is schema name | privilege | true/false for with_grant_option
-terraform import snowflake_schema_grant.example 'schemaName|MONITOR|false'
+# format is database name | schema name | | privilege | true/false for with_grant_option
+terraform import snowflake_schema_grant.example 'databaseName|schemaName||MONITOR|false'
 ```
--- a/docs/resources/stage_grant.md
+++ b/docs/resources/stage_grant.md
@@ -49,6 +49,6 @@ resource snowflake_stage_grant grant {
 Import is supported using the following syntax:
 
 ```shell
-# format is stage name | privilege | true/false for with_grant_option
-terraform import snowflake_stage_grant.example 'stageName|USAGE|true'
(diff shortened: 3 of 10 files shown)


kanata2:
Ah, I found same cases which should be fixed and will fix its later.

kanata2:
Fixed in 8cdb7ca

ryanking:
@kanata2 what version was this on? We had a regression in, I think v0.20, related to this which was later reverted.

kanata2:
@ryanking I used v0.18.1, but main branch seems to have same problem.

kanata2:
I tried `terraform import snowflake_database_grant.xxx 'xxx|USAGE|false'`again, but got the same result.

```
$ terraform import snowflake_database_grant.xxx 'xxx|USAGE|false'
snowflake_database_grant.xxx: Importing from ID "xxx|USAGE|false"...
snowflake_database_grant.xxx: Import prepared!
  Prepared snowflake_database_grant for import
snowflake_database_grant.xxx: Refreshing state... [id=xxx|USAGE|false]

Error: 4 or 5 fields allowed


$ terraform import snowflake_database_grant.xxx 'xxx|||USAGE|false'
snowflake_database_grant.xxx: Importing from ID "xxx|||USAGE|false"...
snowflake_database_grant.xxx: Import prepared!
  Prepared snowflake_database_grant for import
snowflake_database_grant.xxx: Refreshing state... [id=xxx|||USAGE|false]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

docs/resources/database_grant.md
```diff
@@ -42,6 +42,6 @@ resource snowflake_database_grant grant {
 Import is supported using the following syntax:
 
 ```shell
-# format is database name | privilege | true/false for with_grant_option
-terraform import snowflake_database_grant.example 'databaseName|USAGE|false'
+# format is database name | | | privilege | true/false for with_grant_option
+terraform import snowflake_database_grant.example 'databaseName|||USAGE|false'
```
ryanking, line 46: You need to update this file - https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/examples/resources/snowflake_database_grant/import.sh 

and then run `make docs`.

ryanking (changes requested):
same for all the other files

docs/resources/database_grant.md
kanata2: Ah, sorry. I'll run later.

kanata2: ran at c8f9e2e

alldoami (approved):

alldoami:
@kanata2 can you update your branch with main?

PR #432, opened 2021-01-26
[fix] Grants check if a resource doesn't exist

edulop91:
This is hacky and I don't really know of a better way of managing this. 

Unfortunately, we can land ourselves into situations where snowflake TF enters a permanent error loop. This can happen, for example, if you:
- create a table and a table grant, but don't "associate them" through terraform
- delete the table
- err, since the grant can't detect the table it will keep erring out. 

We instead make it so that if we can't find the grantee resource, we delete the tf resource.    

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*

### DIFF
--- a/pkg/resources/database_grant_acceptance_test.go
+++ b/pkg/resources/database_grant_acceptance_test.go
@@ -62,7 +62,7 @@ func TestAccDatabaseGrant_basic(t *testing.T) {
 }
 
 // TODO(el): fix this test
-// func TestAccDatabaseGrant_roleNotExists(t *testing.T) {
+// func TestAccDatabaseGrant_dbNotExists(t *testing.T) {
 // 	dbName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
 // 	roleName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
 
--- a/pkg/resources/database_grant_acceptance_test.go
+++ b/pkg/resources/database_grant_acceptance_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/pkg/errors"
 )
 
 func testRolesAndShares(t *testing.T, path string, roles, shares []string) func(*terraform.State) error {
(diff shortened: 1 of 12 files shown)


ryanking (approved):
I think this is the way it has to be :(

igungor:
I'm experiencing a similar issue and wanted to chime in. How about checking the [error number](https://github.com/snowflakedb/gosnowflake/blob/master/errors.go#L11)? For demo purposes;

```go
package main

import (
	"database/sql"

	"github.com/jmoiron/sqlx"
	"github.com/kr/pretty"
	"github.com/snowflakedb/gosnowflake"
)

func main() {
	cfg := &gosnowflake.Config{
		Account:       "xxx",
		Database:      "xxx",
		User:          "xxx",
		Role:          "xxx",
		Authenticator: gosnowflake.AuthTypeExternalBrowser,
	}
	dsn, err := gosnowflake.DSN(cfg)
	if err != nil {
		panic(err)
	}

	grantsInSchema := `SHOW FUTURE GRANTS IN SCHEMA "foo"."bar"`
	db, err := sqlx.Open("snowflake", dsn)
	if err != nil {
		panic(err)
	}
	row := db.QueryRowx(grantsInSchema)
	r := &schema{}
	if err := row.StructScan(r); err != nil {
		snowflakeErr, ok := err.(*gosnowflake.SnowflakeError)
		if !ok {
			panic(err)
		} else {
			pretty.Println(snowflakeErr)
		}
	}
}

type schema struct {
	Name          sql.NullString `db:"name"`
	DatabaseName  sql.NullString `db:"database_name"`
	Comment       sql.NullString `db:"comment"`
	Options       sql.NullString `db:"options"`
	RetentionTime sql.NullString `db:"retention_time"`
}
```

```bash
tmp  Δ go run main.go
[...]
INFO[0004]connection.go:146 gosnowflake.(*snowflakeConn).exec Success: false, Code: 2003
ERRO[0004]connection.go:325 gosnowflake.(*snowflakeConn).QueryContext error: 002003 (02000): SQL compilation error:
Database '"foo"' does not exist or not authorized.
&gosnowflake.SnowflakeError{
    Number:         2003,
    SQLState:       "02000",
    QueryID:        "xxx",
    Message:        "SQL compilation error:\nDatabase '\"foo\"' does not exist or not authorized.",
    MessageArgs:    nil,
    IncludeQueryID: false,
}
```

I'm not very confident if `2003` is an error number we can rely on but it could be better than checking the error string. What do you think?

pkg/resources/database_grant_acceptance_test.go
```diff
@@ -61,6 +62,36 @@ func TestAccDatabaseGrant_basic(t *testing.T) {
 	})
 }
 
+func TestAccDatabaseGrant_roleNotExists(t *testing.T) {
+	dbName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
+	roleName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
+
+	resource.ParallelTest(t, resource.TestCase{
```
cziie, line 69: **[tfproviderlint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>AT001: missing CheckDestroy

edulop91:
@igungor good idea, thank you. Added the extra check and a corresponding integration test.

PR #442, opened 2021-02-03
add link to getting-started guide from snowflake

jdanielmyers:
I've added a link to the main readme to a getting-started guide written by Snowflake for this provider.

https://guides.snowflake.com/guide/terraforming_snowflake/

for the source of the guide (pull requests welcome!) take a look here: 

https://github.com/Snowflake-Labs/sfguides/blob/master/site/devlabs/terraforming_snowflake.md


ryanking (changes requested):
I think we should put these docs in this repo. I would be happy to give folks at snowflake commit rights, CODEOWNERS, etc.

Though if there are process or other reasons that prevent that, we can talk.

PR #468, opened 2021-02-23
schema: handle retention_time='' while describing a schema

### LINKED ISSUE #467: Provider can't handle schema retention_time=''
**Provider Version**

v0.23.2 (latest as of this writing)

**Terraform Version**

v0.14.4

**Describe the bug**

Provider can't handle `retention_time` column while describing the schema. `retention_time` column is empty string, instead of an integer or NULL value, which is why the provider can't handle. Please see the below output.

```
show schemas like 'midtables';

created_on,name,is_default,is_current,database_name,owner,comment,options,retention_time
2016-02-19 07:09:03.721 +0000,MIDTABLES,N,N,XXXXXXXX,DATA_ADMIN,,,


Error: sql: Scan error on column index 8, name "retention_time": converting driver.Value type string ("") to a int64: invalid syntax

---------------------

module.schema_midtables.snowflake_schema.schema: Refreshing state... [id=XX|MIDTABLES]
2021-01-19T19:33:43.898+0300 [DEBUG] plugin.terraform-provider-snowflake_v0.14.4: 2021/01/19 19:33:43 [DEBUG] query stmt SHOW SCHEMAS LIKE 'MIDTABLES' IN DATABASE "XX"
2021-01-19T19:33:44.193+0300 [DEBUG] plugin.terraform-provider-snowflake_v0.14.4: 2021/01/19 19:33:44 [DEBUG] sql-conn-query [query SHOW SCHEMAS LIKE 'MIDTABLES' IN DATABASE "XX" err <nil> duration 294.483756ms args {}] 
2021-01-19T19:33:44.193+0300 [DEBUG] plugin.terraform-provider-snowflake_v0.14.4: 2021/01/19 19:33:44 [DEBUG] sql-rows-next [err <nil> duration 6.299µs]
```

**Expected behavior**

Provider should handle empty string as well as NULL and integer values.

**Code samples and commands**

Pasted above.

**Additional context**

I'm going to send a fix after creating this issue.

igungor:
Snowflake returns empty string when you describe the schema with `SHOW
SCHEMAS LIKE 'schema_name'`.

Fixes #467
Updates #217


pkg/resources/schema.go
```diff
@@ -203,9 +204,22 @@ func ReadSchema(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	err = d.Set("data_retention_days", s.RetentionTime.Int64)
-	if err != nil {
-		return err
+	// "retention_time" may sometimes be empty string instead of an integer
+	{
```
alldoami, line 208: Is there a reason why you're using blocks nest and influence scoping here?

igungor: There're multiple set and parse operations in this function. This scope is there to isolate the variables needed for retention time parsing logic from other set operations, just to be safe, not that there's shadowing.

alldoami (approved):

Issue #137, opened 2020-02-10
Feature Request: to be able to modify a grant without re-creating it


pwnage101:
The ability to modify grants (database grants, schema grants, warehouse grants etc.) without completely re-creating them.

Motivation: As our Snowflake userbase grows, more people (and automation users) are running queries at any given time.  More and more often, people complain that they don't have permission to access certain resources that they should have access to.  This is typically caused by the temporarily loss of some grants while we are applying a terraform plan due to grant resources that have `ForceNew` enabled on the roles list.

Via the Snowflake web UI, I believe it is technically possible to modify individual grants to users/roles without affecting other users/roles which share that same grant object.

ryanking:
@pwnage101 Yeah that sounds painful.

I spent some time looking through the [snowflake docs](https://docs.snowflake.net/manuals/sql-reference/sql-all.html) and can't figure out how one would do this. Is is possible that the snowflake web ui is doing as revoke followed by a grant?

pwnage101:
Perhaps introducing the Web UI into the conversation was confusing---lets just talk about just pure snow-SQL.  For instance, I can grant to a bunch of roles:
```sql
GRANT USAGE ON SCHEMA myschema TO ROLE role1;
GRANT USAGE ON SCHEMA myschema TO ROLE role2;
```
Then, later I can grant the same permissions to another role:
```sql
GRANT USAGE ON SCHEMA myschema TO ROLE role3;
```
The last statement does not revoke (even temporarily) usage on myschema to role1 or role2.  However, If I were to do the last step via terraform, I'd get a diff that looked something like this:
```diff
  # snowflake_schema_grant.foo must be replaced
-/+ resource "snowflake_schema_grant" "foo" {
        database_name = "BAR"
      ~ id            = "BAR|||USAGE" -> (known after apply)
        privilege     = "USAGE"
      ~ roles         = [ # forces replacement
            "ROLE1",
            "ROLE2",
          + "ROLE3",
        ]
      - shares        = [] -> null
    }
```
role1 and role2 would have their permissions interrupted by the apply.

ryanking:
@pwnage101 got it and that makes sense and seems like something that could be implemented.

jstachowiak:
@ryanking, any idea how we could go about implementing this?

md5:
[...]

PR #469, opened 2021-02-23
account: update grants in-place instead of remove/add method

igungor:
Terraform supports updating a resource in-place if a Terraform schema
key is not marked as `ForceNew` and an `Update` method is provided. In
fact, it is implemented for 2 resources already:

- https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/schema_grant.go#L90
- https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/role_grants.go#L18

This PR aims to add in-place support for `account grants`.

We're testing this method at @peak for a while. This is critical because
'force delete/add' method makes services hiccup while trying to access a
Snowflake resource.

I plan to add more of this for:
- database grants
- table grants
- view grants
- warehouse grants

also update:
- schema grants
- role grants

to have the same structure as the others if this PR looks good to you.

Updates #137


alldoami (approved):
Thank you!

PR #472, opened 2021-02-23
provider: read private key from environment variable

### LINKED ISSUE #471: Read private key from environment variable

We're using [atlantis](https://github.com/runatlantis/atlantis) on AWS Fargate to manage our all resources, including Snowflake. Provider has private-key support but only reads from the file. It'd be best if it could read the key content directly instead of given file path. 

A new provider attribute named `private_key` (just like `private_key_path`) which holds the private key content. Also it can be passed via environment variable `SNOWFLAKE_PRIVATE_KEY`, which enables tools like Atlantis to run on a serverless platform in a more simpler way.

**Additional context**

I'll send a PR for a more concrete approach. I understand that doing nearly the same thing with two provider attributes is not very nice, but on the other hand, it's very practical for users like us.

igungor:
If you keep your secrets in Vault (or similar secret manager), you need
to render a file for that secret, which complicates things on a
serverless platform. It'd be simpler to just export an environment
variable instead of managing files.

Fixes #471


pkg/provider/provider.go
```diff
@@ -54,7 +54,14 @@ func Provider() *schema.Provider {
 				Optional:      true,
 				DefaultFunc:   schema.EnvDefaultFunc("SNOWFLAKE_PRIVATE_KEY_PATH", nil),
 				Sensitive:     true,
-				ConflictsWith: []string{"browser_auth", "password", "oauth_access_token"},
+				ConflictsWith: []string{"browser_auth", "password", "oauth_access_token", "private_key"},
```
ryanking, line 57: Please add `"private_key"` to `ConflictsWith` for all others that currently have `"private_keyPath"`.

ryanking (changes requested):

pkg/provider/provider.go
igungor: done. PTAL.

ryanking (approved):

thenaturalist:
Hi there, has this been released to the registry by now?

PR #480, opened 2021-03-03
table_grant: update table grants in-place instead of recreating

igungor:
Related to #469
Updates #137

### DIFF
--- a/pkg/resources/table_grant.go
+++ b/pkg/resources/table_grant.go
@@ -41,6 +41,7 @@ var tableGrantSchema = map[string]*schema.Schema{
 		Optional:     true,
 		Description:  "The privilege to grant on the current or future table.",
 		Default:      privilegeSelect.String(),
+		ForceNew:     true,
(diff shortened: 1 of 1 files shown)


pkg/resources/table_grant.go
```diff
@@ -42,28 +42,24 @@ var tableGrantSchema = map[string]*schema.Schema{
 		Description:  "The privilege to grant on the current or future table.",
 		Default:      privilegeSelect.String(),
 		ValidateFunc: validation.ValidatePrivilege(validTablePrivileges.ToList(), true),
-		ForceNew:     true,
```
edulop91, line 45: I might be missing something but it seems to me we should keep `ForceNew = true` here. The update function doesn't currently know how to handle `privilege` or `on_future` updates.

edulop91 (commented):
one small comment, otherwise lgtm. Thank you

pkg/resources/table_grant.go
igungor: yes, we should keep this as is. reverting now.

edulop91 (approved):

PR #482, opened 2021-03-03
schema_grant: refactor schema grant Update function

igungor:
Related to #469

### DIFF
--- a/pkg/resources/schema_grant.go
+++ b/pkg/resources/schema_grant.go
@@ -67,6 +67,7 @@ var schemaGrantSchema = map[string]*schema.Schema{
 		Optional:      true,
 		Description:   "When this is set to true, apply this grant on all future schemas in the given database. The schema_name and shares fields must be unset in order to use on_future.",
 		Default:       false,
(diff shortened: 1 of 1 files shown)


pkg/resources/schema_grant.go
```diff
@@ -61,14 +61,12 @@ var schemaGrantSchema = map[string]*schema.Schema{
...
 		Optional:    true,
 		Description: "Grants privilege to these shares (only valid if on_future is unset).",
-		ForceNew:    true,
 	},
 	"on_future": {
 		Type:          schema.TypeBool,
 		Optional:      true,
 		Description:   "When this is set to true, apply this grant on all future schemas in the given database. The schema_name and shares fields must be unset in order to use on_future.",
 		Default:       false,
-		ForceNew:      true,
```
edulop91, line 71: Should this remain `true`?

igungor: you're right. reverted.

edulop91 (approved):

PR #491, opened 2021-03-10
Add new 'api integration' and 'external function' resources

asaintsever:
- Add support for new `snowflake_api_integration` and `snowflake_external_function` resources
- Add unit and acceptance tests
- Add examples & updated documentation

## Test Plan

Changes have been compiled and tested successfully (including acceptance tests)

* [x] unit tests
* [x] acceptance tests

## References

- https://docs.snowflake.com/en/sql-reference/external-functions-introduction.html
- https://docs.snowflake.com/en/sql-reference/external-functions-creating-aws-options.html
- https://docs.snowflake.com/en/sql-reference/external-functions-creating-azure-options.html
- https://docs.snowflake.com/en/sql-reference/sql/create-api-integration.html
- https://docs.snowflake.com/en/sql-reference/sql/create-external-function.html


asaintsever:
Hi @edulop91,

Is there something more I should provide to have this PR considered? I saw some PR rotting since several months and I am keen to have those changes merged as we are already making use of these new resources in Terraform configurations.

Thanks in advance for your guidance.
Best.

alldoami:
Sorry to do this, but would it be possible to separate this PR into two PRs? It is a lot to review 😅

asaintsever:
@alldoami 
Well, it doesn't really make sense to split this contribution into several PR. If not already done, I would suggest you read https://docs.snowflake.com/en/sql-reference/external-functions-introduction.html : it is not possible to create an external function without creating an api integration. They come both at play. For distinct and not related resources, I would have submitted different PR of course but not the case here.

Moreover, by splitting into 2 PR the risk is to merge one and not the other. So you'll have to review both in the end and make sure both are part of same release. Is it what you want?

pkg/resources/api_integration.go
```diff
@@ -0,0 +1,340 @@
...
+
+	d.SetId(name)
+
+	return ReadAPIIntegration(d, meta)
+}
+
+// ReadAPIIntegration implements schema.ReadFunc
+func ReadAPIIntegration(d *schema.ResourceData, meta interface{}) error {
+	db := meta.(*sql.DB)
+	id := d.Id()
```
alldoami, line 153: If you're going to set this id here, replace the `d.Id()`s below for consistency.

pkg/resources/api_integration.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}
+
+	return err
+}
+
+// UpdateAPIIntegration implements schema.UpdateFunc
+func UpdateAPIIntegration(d *schema.ResourceData, meta interface{}) error {
+	db := meta.(*sql.DB)
+	id := d.Id()
```
alldoami, line 240: Same here with one instance below.

pkg/resources/api_integration.go
```diff
@@ -0,0 +1,340 @@
...
+			stmt.SetString("AZURE_TENANT_ID", d.Get("azure_tenant_id").(string))
+		}
+		if d.HasChange("azure_ad_application_id") {
+			runSetStatement = true
+			stmt.SetString("AZURE_AD_APPLICATION_ID", d.Get("azure_ad_application_id").(string))
+		}
+	}
+
+	if runSetStatement {
+		if err := snowflake.Exec(db, stmt.Statement()); err != nil {
```
alldoami, line 299: Can you log the statement being run? Potentially large.

alldoami (commented):
Few NITs

pkg/resources/api_integration.go
asaintsever: Done

pkg/resources/api_integration.go
asaintsever: There is one already (the `snowflake.Exec()` method has `log.Print("[DEBUG] exec stmt ", query)`).

alldoami (commented):
I'm looking at the syntax docs on the API Integration and it looks like the `API_PROVIDER` value should not be in quotes: https://docs.snowflake.com/en/sql-reference/sql/create-api-integration.html#syntax, I think the statement you query wraps the value in quotes.

asaintsever:
That's true. It happens that using quotes works fine as it is the case with any query arg/param.
The documentation states that API_PROVIDER "

alldoami:
@asaintsever I believe this comment is the same for the external function: https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/491#pullrequestreview-616727059, seeing it here in the example: https://docs.snowflake.com/en/sql-reference/sql/create-external-function.html#examples.

asaintsever:
@alldoami sorry I do not understand. What do you mean? As for identifier, Snowflake accepts both quoted and unquoted forms: https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html

So, except for the API_PROVIDER that should be unquoted as stated by the documentation (even if it works with quotes ...), you are free to have quotes or not.

alldoami (approved):

Issue #446, opened 2021-02-10
snowflake_table: table column actions  will force-destroy the table and recreate it


chenlu-wang:
**Provider Version**

0.22.0

**Terraform Version**

0.14.5

**Describe the bug**

Add, alter, drop columns actions on table are attribute changes, but it will cause that the table is destroyed and recreated. 
For example, to add a column on table using `snowflake_table` resource, the table will be dropped and recreated. 
It makes unexpected data lost in whole table by column operations.

**Expected behavior**
When columns are updated, the table resource should be updated rather than destroyed. 
For example, to add a new column on table, the table should just update the column with empty data. 
In snowflake, it is the equivalent action of `alter table table_name add column column_1 data type;`
[snowflake documentation ](https://docs.snowflake.com/en/sql-reference/sql/alter-table.html#table-column-actions-tablecolumnaction)

**Code samples and commands**
If I add a column {name="SOURCE", type="VARCHAR"} on mytable, terraform will destroy the table and add it again: 
```
-/+ resource "snowflake_table" "mytable" {
      ~ id       = "mydb|myschena|mytable" -> (known after apply)
        name     = "mytable"
      ~ owner    = "SYSADMIN" -> (known after apply)
        # (3 unchanged attributes hidden)

      + column { # forces replacement
          + name = "SOURCE"
          + type = "VARCHAR"
        }
    }
Plan: 1 to add, 0 to change, 1 to destroy.
```
The expected resource should be a resource change.

nikolayskovpin:
Also version 0.23.2:
even changes were applied to snowflake table, every time I run `terraform plan` it shows previous state and plans to apply the same change

hragarwal:
I see a merged PR, when this change would be available ?

Issue #421, opened 2021-01-21
Support the ability to alter table...modify column data type


sfc-gh-dsandler:
As a data modeler, I need to modify a column's data type. I would like to use the snowflake_table resource to support this action, provided [the action is supported by Snowflake ](https://docs.snowflake.com/en/sql-reference/sql/alter-table-column.html#examples)

From pure-DDL perspective, it would like this: 

`alter table mydb.myschema.mytbl modify column col1 varchar(1234);`

to be supported with the nested schema for column (as such):

```
resource "snowflake_table" "my_tbl" {
  name     = "MYTBL"
  database = "MYDB"
  schema   = "MYSCHEMA"
  column   {
            name = "COLUMN1"
            old_name = "COL1"
            type = "varchar(1234)" 
  } 
  comment  = "test table"

  depends_on = [snowflake_schema.schema]
}
```

PR #503, opened 2021-03-25
Support adding, removing and changing the type of columns in tables

edwardbrowncross:
The current behaviour is to drop the table and all its data when a column is added, removed, or modified (#446, #421). This PR mostly fixes this. 

Now supported without having to drop and recreate the table:
* Adding columns
* Removing columns
* Changing the data type of a column

Partially supported:
* Renaming columns - the column is dropped and recreated leading to _some_ data loss. There is not currently any better way to correlate columns than their name.

## Test Plan
* [ ✔] acceptance tests (plus made the existing ones more rigorous)
* [✔ ] snowflake unit tests

## References
#446 
#421

[ALTER TABLE documentation](https://docs.snowflake.com/en/sql-reference/sql/alter-table.html)


AdamDewberry:
Ooo this looks good, excited for a codeowner to review

alldoami:
@edwardbrowncross could you update your branch with `main`?

alldoami (approved):

PR #520, opened 2021-04-19
role_grants: update read to only use grants specific to the tf resource

cstkpk:
The current behavior is to query Snowflake for all grants to the specified role when a role grant is read, updated, or deleted. This means the update or delete to applies to _all_ grants to the specified role rather than just the grants that are part of the terraform resource. This causes issues when there are multiple resources handling grants to the same role, as well as when there is a role hierarchy (see issues linked below for more info).

These changes modify the role grant read to look at only the grants that are part of the specified terraform resource. This allows for more flexible use of terraform modules and eliminates issues with role hierarchies.

## Test Plan
* [x] acceptance tests
* [x] unit tests

## References
[189](https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/189)
[512](https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/512)

### DIFF
--- a/pkg/resources/role_grants.go
+++ b/pkg/resources/role_grants.go
@@ -98,6 +98,9 @@ func ReadRoleGrants(d *schema.ResourceData, meta interface{}) error {
 	db := meta.(*sql.DB)
 	roleName := d.Id()
 
+	tfRoles := expandStringList(d.Get("roles").(*schema.Set).List())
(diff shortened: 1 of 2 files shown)


alldoami:
@cstkpk could you look at the errors and update your branch 😄

alldoami (changes requested):
Review errors and update branch

alldoami (approved):

PR #524, opened 2021-04-23
[feature] masking policy grant resource

ajbosco:
Adds resource to create grants for specific masking policies. 

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/resources/masking_policy_grant.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var validMaskingPoilcyPrivileges = NewPrivilegeSet(
+	privilegeOwnership,
+	privilegeApply,
+)
+
+var maskingPolicyGrantSchema = map[string]*schema.Schema{
+	"masking_policy_name": {
```
alldoami, line 15: Nit: alphabetize these please!

pkg/resources/masking_policy_grant_acceptance_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	out := bytes.NewBuffer(nil)
+	tmpl := template.Must(template.New("view)").Parse(config))
+	err := tmpl.Execute(out, map[string]string{
+		"database_name":       database_name,
+		"schema_name":         schema_name,
+		"masking_policy_name": masking_policy_name,
+		"role_name":           role,
+	})
+	r.NoError(err)
```
alldoami, line 81: All of our acceptance tests are formatted like:
```
return fmt.Sprintf(`
...
`, name, ...)
```

Is there a reason you didn't want to do it that way? Wondering if this will confuse others since it's not the convention. Would you be opposed to switching it to the above format?

alldoami (changes requested):

pkg/resources/masking_policy_grant.go
ajbosco: Updated

pkg/resources/masking_policy_grant_acceptance_test.go
ajbosco: I was just copying how the `file_format_grant` test was written. Updated to use the normal convention now.

alldoami (approved):
Thank you!

PR #548, opened 2021-05-18
Feature/table clustering

berosen:
This PR does the following: 

* Adds the ability to define clustering keys on a table resource. 
* Adds applicable tests
* Updates docs

## Test Plan

* [x] acceptance tests

Tested with
`TF_ACC=1 go test -v ./... -run TestAcc_Table` 

## References

* #242
* https://docs.snowflake.com/en/user-guide/tables-clustering-keys.html


berosen:
@alldoami Would you be able to give me a pointer for fixing the docs failure?

alldoami:
@berosen if you run `make docs` it should fix the errors! You shouldn't be manually editing the docs.

berosen:
> @berosen if you run `make docs` it should fix the errors! You shouldn't be manually editing the docs.

Thanks @alldoami! It seems like `tfplugindocs` generates docs differently depending on the directory it's being run from. I believe it should be good to go now.

alldoami (approved):
Thank you!

PR #552, opened 2021-05-24
Lower limit on warehouse auto-suspend setting

jnoynaert:
The warehouse auto-suspend setting can actually be an arbitrary integer (you can test this with `alter warehouse reporting set auto_suspend = <x>`). The current check for 60 seconds imposes an arbitrary restriction on the provider that's not present in the service.

Negative values are just interpreted as "never". Based on the intent of the provider surface it looks like the "never" case should be handled by removing the auto-suspend setting, so a 1-second minimum seemed reasonable.


alldoami (approved):
Thank you!

wookasz:
Thank you!

Issue #549, opened 2021-05-19
Support Resource Security Integration (SCIM)


ChrisIsidora:
**Is your feature request related to a problem? Please describe.**

I would be nice to create the Snowflake Security Integrations with this provider.

**Describe the solution you'd like**

Snowflake documentation: https://docs.snowflake.com/en/user-guide/scim.html
 
```
create or replace security integration aad_provisioning
    type = scim
    scim_client = 'azure'
    run_as_role = 'AAD_PROVISIONER';
```

PR #556, opened 2021-05-31
Add SCIM Integration Resource

ChrisIsidora:
Added SCIM Integration Resource (#549)

## Test Plan
* [x] acceptance tests
* [x] unit tests…

## References
See https://docs.snowflake.com/en/user-guide/scim.html

*


ChrisIsidora:
@alldoami This is related to the feature request #549

pkg/resources/scim_integration.go
```diff
@@ -0,0 +1,229 @@
...
+			"OKTA", "AZURE", "CUSTOM",
+		}, true),
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+			normalize := func(s string) string {
+				return strings.ToUpper(strings.Replace(s, "-", "", -1))
+			}
+			return normalize(old) == normalize(new)
+		},
+	},
+	"run_as_role": {
```
alldoami, line 35: Wondering if there is a clearer name for this... possibly `provisioner_role`?

ChrisIsidora: I wanted to keep it the same as in Snowflake itself. But good point provisioner_role is more clear

ChrisIsidora: @alldoami Renamed to `provisioner_role`

ChrisIsidora:
@alldoami can you trigger the test after the latest changes?

ChrisIsidora:
@alldoami  Removed Enabled property which cannot be set for SECURITY INTEGRATION with type=SCIM

ChrisIsidora:
@alldoami Added the dependent resources to the acceptance tests. Can you check if it's ok now?

alldoami:
This would be best if you could test locally and request my review when ready.

ChrisIsidora:
@alldoami You are right, sorry for the inconvenience. I tested the acceptance test locally, made some fixes and it seems to work now.

image

ChrisIsidora:
@alldoami Not sure why the acc test are failing. It does work on my snowflake account. See below, I made some changes to the resource id maybe that might help. For the unit test I saw that for the Generic alter builder there was no sorting which means that my alter unit test might fail sometimes because of the random order of the statement. That should be fixed now.

image

ChrisIsidora:
@alldoami  Seems that okta provider already exists on the acc snowflake account. I removed Okta tests

ChrisIsidora:
@alldoami can this be merged so that I can work / update #557? Thanks in advance for your patience

pkg/snowflake/generic.go
```diff
@@ -108,24 +109,53 @@ func (ab *AlterPropertiesBuilder) Statement() string {
...
-		sb.WriteString(fmt.Sprintf(" %s='%s'", strings.ToUpper(k), EscapeString(v)))
+	sortedStringProperties := make([]string, 0)
+	for k := range ab.stringProperties {
+		sortedStringProperties = append(sortedStringProperties, k)
 	}
+	sort.Strings(sortedStringProperties)
 
-	for k, v := range ab.stringListProperties {
-		sb.WriteString(fmt.Sprintf(" %s=%s", strings.ToUpper(k), formatStringList(v)))
+	for _, k := range sortedStringProperties {
```
jacobrheath, line 118: This is a little verbose for all the fields. Can you make a helper function that returns a sorted list of string?
```suggestion
	for _, k := range sortStrings(ab.stringProperties) {
```

`sortStrings` would look something like `func sortStrings(strs []string) []string` and would return a list of a sorted strings from the list passed in.

pkg/snowflake/generic.go
```diff
@@ -108,24 +109,53 @@ func (ab *AlterPropertiesBuilder) Statement() string {
 
 	sb.WriteString(ab.rawStatement)
 
-	for k, v := range ab.stringProperties {
-		sb.WriteString(fmt.Sprintf(" %s='%s'", strings.ToUpper(k), EscapeString(v)))
+	sortedStringProperties := make([]string, 0)
```
jacobrheath, line 112:
```suggestion
	sortedStringProperties := []string{}
```

jacobrheath: I think this is the convention we have going.

jacobrheath (changes requested):

pkg/snowflake/generic.go
ChrisIsidora: @jacobrheath I copied this from the Create function which is also in Generic, but I will try your suggestion

ChrisIsidora:
@alldoami  and @jacobrheath Can you please review the changes requested. Thanks!

alldoami (approved):

PR #561, opened 2021-06-01
Remove snowflake_role_grants name example

dfedde:
`name` is not a field on the snowflake_role_grants resource making the example invalid if copied exactly.

## Test Plan
This is just an example change and does not not currently have any tests other than manually checking the terraform.

### DIFF
--- a/docs/resources/role_grants.md
+++ b/docs/resources/role_grants.md
@@ -33,8 +33,6 @@ resource "snowflake_role" "other_role" {
 }
 
 resource "snowflake_role_grants" "grants" {
-  name = "foo"
-
   role_name = "${snowflake_role.role.name}"
 
   roles = [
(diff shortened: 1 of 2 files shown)


alldoami:
Could you run `make-docs`?

dfedde:
I looks like I am doing something wrong or something changed in the way docs get built. `make docs` is generating doc files with the full name of the resource and not just the part without the provider name in it (`docs/resources/database.md` -> `docs/resources/snowflake_database.md`). If this is something I am doing wrong how do I fix this? and if it is a change to the way docs are built should the generated doc changes be committed separately? In this PR?

alldoami:
You can put it in this PR! We did change the way docs are generated. We use https://github.com/hashicorp/terraform-plugin-docs. That change does seem correct though.

alldoami (approved):

PR #575, opened 2021-06-15
feat: Add SHOW_INITIAL_ROWS to stream resource

alldoami:
Addressing this feature request https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/537
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/resources/stream.go
```diff
@@ -170,6 +177,7 @@ func CreateStream(d *schema.ResourceData, meta interface{}) error {
 
 	builder.WithOnTable(resultOnTable.DatabaseName, resultOnTable.SchemaName, resultOnTable.OnTableName)
 	builder.WithAppendOnly(appendOnly)
+	builder.WithShowInitialRows(showInitialRows)
```
edulop91, line 180: you need to add this prop to the Read func.

For the update side of things, is this property "updateable"? If not, you might want to "ForceNew" on the Resource schema

edulop91: Seems like the only thing you can update are comments so this will require a ForceNew (in fact, every property except comment should be this way)

edulop91: https://docs.snowflake.com/en/sql-reference/sql/alter-stream.html

pkg/snowflake/stream.go
```diff
@@ -59,6 +60,16 @@ func (sb *StreamBuilder) WithAppendOnly(b bool) *StreamBuilder {
...
 }
 
+func (sb *StreamBuilder) WithShowInitialRows(b bool) *StreamBuilder {
+	sb.showInitialRows = false
+
+	if b {
+		sb.showInitialRows = b
+	}
+
+	return sb
```
edulop91, line 70:
```suggestion
	sb.showInitialRows = b
	return sb
```
are they equivalent?

alldoami: Since it's optional I don't think it has a value unless you give it a value of false...

edulop91: `func(b bool)`, `b` must be either true or false in this context

edulop91: how I read the existing code is "if b is True then set sb.showInitialRows to True, otherwise False" which is the same as `sb.showInitialRows = b`

edulop91 (approved):
lgtm other than that small simplification

PR #580, opened 2021-06-17
fix: Escape String for AS in external table

alldoami:
Addresses: https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/542
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*

### DIFF
--- a/pkg/snowflake/external_table.go
+++ b/pkg/snowflake/external_table.go
@@ -113,7 +113,7 @@ func (tb *ExternalTableBuilder) Create() string {
 	q.WriteString(fmt.Sprintf(` (`))
 	columnDefinitions := []string{}
 	for _, columnDefinition := range tb.columns {
-		columnDefinitions = append(columnDefinitions, fmt.Sprintf(`"%v" %v AS %v`, EscapeString(columnDefinition["name"]), EscapeString(columnDefinition["type"]), EscapeString(columnDefinition["as"])))
+		columnDefinitions = append(columnDefinitions, fmt.Sprintf(`"%v" %v AS %v`, EscapeString(columnDefinition["name"]), EscapeString(columnDefinition["type"]), columnDefinition["as"]))
 	}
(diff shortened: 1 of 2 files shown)


pkg/snowflake/external_table.go
```diff
@@ -113,7 +113,7 @@ func (tb *ExternalTableBuilder) Create() string {
 	q.WriteString(fmt.Sprintf(` (`))
 	columnDefinitions := []string{}
 	for _, columnDefinition := range tb.columns {
-		columnDefinitions = append(columnDefinitions, fmt.Sprintf(`"%v" %v AS %v`, EscapeString(columnDefinition["name"]), EscapeString(columnDefinition["type"]), EscapeString(columnDefinition["as"])))
+		columnDefinitions = append(columnDefinitions, fmt.Sprintf(`"%v" %v AS %v`, EscapeString(columnDefinition["name"]), EscapeString(columnDefinition["type"]), columnDefinition["as"]))
```
edulop91, line 116: I wonder if a more sustainable fix is to make it so that you can do `columnDefinition.Get("name")` and it'll return behind the scenes `EscapeString(columnDefinition["name"])`

edulop91 (approved):

pkg/snowflake/external_table.go
alldoami: Hmm `columnDefinition` is a map of strings so there's no `Get` method.

edulop91: it is now, but it doesn't have to be. I don't think you should make that change in the PR though

alldoami: Ohhh you were saying to change the structure of `columnDefinitions` so that we could use `Get`...

alldoami: What would the difference be from what I changed it to?

Issue #509, opened 2021-03-31
snowflake_sequence


tannon5d:
**Is your feature request related to a problem? Please describe.**

We have a couple of sequences that can't be deployed by Terraform. Neither can tables using sequences for columns.

**Describe the solution you'd like**

Something like:
```
resource snowflake_sequence sequence {
  database = "dbname"
  schema   = "testschema"
  name     = "testsequence"
  comment  = "A sequence."
  initial_value = 1
  sequence_interval = 1
}

resource snowflake_table table {
  database = "dbname"
  schema   = "testschema"
  name     = "testtable"
  comment  = "A table."
  owner    = "me"

  column {
    name = "id"
    type = "int"
    default = "dbname.testschema.testsequence.nextval"
  }
}
```

**Describe alternatives you've considered**

ATM our sequences and tables referencing sequences are not source controlled.

**Additional context**

N/A

sfc-gh-asawicki:
We are closing this issue as part of a cleanup described in [announcement](https://github.com/Snowflake-Labs/terraform-provider-snowflake/discussions/2755). If you believe that the issue is still valid in v0.89.0, please open a new ticket.

PR #582, opened 2021-06-18
feat: Add a resource to manage sequences

dfedde:
Addresses feature-request #509 

## Test Plan

* [x] acceptance tests
* [x] used terraform to run test creates, updates, destroys and fixing changes made outside of terraform.
* [x] has been run and is working in rxrevu's infrastructure

## References
* https://docs.snowflake.com/en/sql-reference/sql/create-sequence.html


pkg/resources/sequence.go
```diff
@@ -0,0 +1,200 @@
...
+		Description: "The database in which to create the sequence. Don't use the | character.",
+	},
+	"schema": {
+		Type:        schema.TypeString,
+		Required:    true,
+		Description: "The schema in which to create the sequence. Don't use the | character.",
+	},
+	"next_value": {
+		Type:        schema.TypeInt,
+		Description: "The next value the sequance will provide.",
```
alldoami, line 44:
```suggestion
		Description: "The next value the sequence will provide.",
```

alldoami:
I'm not totally familiar with sequences, but should we try and provide a `start` optional parameter?

dfedde:
We could add a start but it is a little confusing on how to implement it. It is only useful on create. On an update we don't want to regress the counter back to its start. We would also need to be smart in the cases where the sequence is recreated.

I was unable to figure out a way to implement it cleanly and was not useful for our use case so I left it out.

alldoami:
@dfedde cool, wanna update your branch and commit the change I suggested above?

alldoami (approved):

PR #598, opened 2021-07-07
fix: Warehouse create and alter properties

alldoami:
Addressing: https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/583

Adding optional parameters to list of properties. Already had `statement_timeout_in_seconds` but added the other two parameters (which are available in both create and alter).

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/resources/warehouse.go
```diff
@@ -11,12 +11,13 @@ import (
 )
 
 // warehouseCreateProperties are only available via the CREATE statement
-var warehouseCreateProperties = []string{"initially_suspended", "wait_for_provisioning", "statement_timeout_in_seconds"}
+var warehouseCreateProperties = []string{"initially_suspended", "wait_for_provisioning"}
```
edulop91, line 14: these should probably ForceNew

pkg/resources/warehouse.go
```diff
@@ -103,6 +104,20 @@ var warehouseSchema = map[string]*schema.Schema{
 		ForceNew:    false,
 		Description: "Specifies the time, in seconds, after which a running SQL statement (query, DDL, DML, etc.) is canceled by the system",
 	},
+	"statement_queued_timeout_in_seconds": {
```
edulop91, line 107: need to add these to Read+Update

edulop91: maybe just Read actually

edulop91 (approved):

PR #610, opened 2021-07-20
fix: Add AWS_SQS_IAM_USER_ARN to notification integration

gouline:
Notification integration fixes missed in #595:

- Add `AWS_SQS_IAM_USER_ARN` read-only/computed parameter
- Fix `AWS_SQS_EXTERNAL_ID` parameter to be read-only/computed, instead of optional
- Documentation, example and import

## Test Plan

* [x] unit tests
* [x] acceptance tests


alldoami (approved):
Thanks!

PR #614, opened 2021-07-22
feat: Add support for table column comments and to control a tables data retention and change tracking settings

robbruce:
Adds the following options to the table resource

* comments on columns
* property to control `DATA_RETENTION_TIME_IN_DAYS` on a table
* property to control `CHANGE_TRACKING` on a table

## Test Plan

* [x] acceptance tests

```
$ TF_ACC=1 go test -v ./... -run TestAcc_Table
?   	github.com/chanzuckerberg/terraform-provider-snowflake	[no test files]
testing: warning: no tests to run
PASS
ok  	github.com/chanzuckerberg/terraform-provider-snowflake/pkg/datasources	(cached) [no tests to run]
?   	github.com/chanzuckerberg/terraform-provider-snowflake/pkg/db	[no test files]
testing: warning: no tests to run
PASS
ok  	github.com/chanzuckerberg/terraform-provider-snowflake/pkg/provider	(cached) [no tests to run]
=== RUN   TestAcc_Table
=== PAUSE TestAcc_Table
=== CONT  TestAcc_Table
--- PASS: TestAcc_Table (57.92s)
PASS
ok  	github.com/chanzuckerberg/terraform-provider-snowflake/pkg/resources	58.328s
testing: warning: no tests to run
PASS
ok  	github.com/chanzuckerberg/terraform-provider-snowflake/pkg/snowflake	(cached) [no tests to run]
?   	github.com/chanzuckerberg/terraform-provider-snowflake/pkg/testhelpers	[no test files]
testing: warning: no tests to run
PASS
ok  	github.com/chanzuckerberg/terraform-provider-snowflake/pkg/validation	(cached) [no tests to run]
?   	github.com/chanzuckerberg/terraform-provider-snowflake/pkg/version	[no test files]
```

## References

* None


pkg/resources/table.go
```diff
@@ -101,6 +108,19 @@ var tableSchema = map[string]*schema.Schema{
 			},
 		},
 	},
+	"data_retention_days": {
+		Type:         schema.TypeInt,
+		Optional:     true,
+		Default:      1,
```
berosen, line 114: Will this default value overwrite the inherited value from a higher level object? I believe the default behavior in snowflake is to inherit from the "higher object". For example, if I were to set `data_retention_days=20` for my schema, I'd expect that to be inherited by the table. This seems like it might break that?

alldoami: I also think since the variable is called `data_retention_time_in_days` we should also name it that here. https://docs.snowflake.com/en/sql-reference/parameters.html#data-retention-time-in-days

It does look like the `Default` value is 1 in the docs though.

berosen: More context 
_If a retention period is specified for a database or schema, the period is inherited by default for all objects created in the database/schema_

from https://docs.snowflake.com/en/user-guide/data-time-travel.html

robbruce: @alldoami @berosen - I really wasn't sure what to use as `snowflake_schema` has `data_retention_days` and `snowflake_database` has `data_retention_time_in_days` so there's inconsistencies already.

https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/schema.go#L54
https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/database.go#L25

@berosen - yes you're right it will override the default from a parent object, I could change this so that if its nil then the setting is not applied to the table if that's more suitable. Which in the case of using this provider, it will be the schema as that has the logic I have here, e.g. it defaults to 1, rather than defaults to inherit.

@alldoami - I can change the behaviour of the `snowflake_schema` resource too, but obviously that will have some side effects for existing deployments. I have no idea why on the database the value is computed though.

berosen: @robbruce , personally I like the idea of using `nil`. I feel the change would be more in-line with expected behaviour. I think making the change in `snowflake_schema` would be a good idea as well. However, I'll defer to whatever @alldoami feels is best.

robbruce: @berosen - I have been looking into this, and I'm not sure its possible with how the current typing system works for `schema.TypeInt` - see https://github.com/hashicorp/terraform-plugin-sdk/issues/261

Essentially , if I set Default to be `nil`, then the SDK returns this as `0` and so there doesn't appear to be a nice way to inherit.

In addition, when reading a table with `SHOW TABLES`, then even if the value was not supplied (so it inherits) then it returns the value defined in it's parent schema, so even then there is no way to detect if there is a difference.

alldoami: Ah ok I see your reasoning now with the naming @robbruce.

berosen: @robbruce That's a good point, I think the work around for that would be complicated. Maybe it's enough to write a note in the description that the inheritance needs to be explicitly defined by passing the same value as the parent object? ie
```
snowflake_table foo {
   ...
   data_retention_days=snowflake_schema.name. data_retention_days
}
```
Sorry for the noise all!

robbruce: @berosen @alldoami - updated the description as suggested and the table example too.

robbruce:
@alldoami - what needs to be done to sort out the integration-trusted test?

alldoami:
@robbruce can you add `change_tracking` to https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/stream_grant_acceptance_test.go#L78-L91

robbruce:
> @robbruce can you add `change_tracking` to `True` to https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/stream_grant_acceptance_test.go#L78-L91

@alldoami - done, and rebased with main

alldoami:
Need to do the same here: https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/39631932d6e90e4707a73cca9c5f1237cf3c3a1c/pkg/resources/stream_acceptance_test.go#L47-L60

robbruce:
> Need to do the same here:
> 
> https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/39631932d6e90e4707a73cca9c5f1237cf3c3a1c/pkg/resources/stream_acceptance_test.go#L47-L60

@alldoami - all done

alldoami (approved):
Woo! Nice

PR #619, opened 2021-07-24
feat: Procedures

funes79:
* New resource **snowflake_procedure**
* As a user using Standard edition of Snowflake, had to add environment variables to prevent the creation of Materialized views and Masking policies.

## Test Plan

* [x] acceptance tests

* [x] unit tests

## References

* Implements part of https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/532 and https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/273
* https://docs.snowflake.com/en/sql-reference/sql/create-procedure.html


pkg/resources/procedure.go
```diff
@@ -0,0 +1,464 @@
...
+		ForceNew: true,
+		// We do not use STRICT, because Snowflake then in the Read phase returns RETURNS NULL ON NULL INPUT
+		ValidateFunc: validation.StringInSlice([]string{"CALLED ON NULL INPUT", "RETURNS NULL ON NULL INPUT"}, false),
+		Description:  "Specifies the behavior of the procedure when called with null inputs.",
+	},
+	"return_behavior": {
+		Type:         schema.TypeString,
+		Optional:     true,
+		Default:      "VOLATILE",
+		Required:     false,
```
alldoami, line 85: don't need this line

pkg/resources/procedure.go
```diff
@@ -0,0 +1,464 @@
...
+		ForceNew:     true,
+		ValidateFunc: validation.StringInSlice([]string{"VOLATILE", "IMMUTABLE"}, false),
+		Description:  "Specifies the behavior of the function when returning results",
+	},
+	"comment": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Default:     "user-defined function",
```
alldoami, line 93: from the docs, it looks like this should be `stored procedure`

pkg/resources/procedure.go
```diff
@@ -0,0 +1,464 @@
...
+		Description: "Specifies a comment for the procedure.",
+	},
+	"statement": {
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      "Specifies the javascript code used to create the procedure.",
+		ForceNew:         true,
+		DiffSuppressFunc: DiffSuppressStatement,
+	},
```
alldoami, line 102: Can you move this argument above where the other required fields are?

ChrisIsidora:
@funes79 Can you take a look at the comments of @alldoami so this can be merged? I have a dependency on this feature. Thanks in advance

pkg/resources/procedure.go
funes79: I agree, Required removed.

pkg/resources/procedure.go
funes79: No, the default comment for a stored procedure is "user-defined function". Its wierd but I tested it on Snowflake many times.

pkg/resources/procedure.go
funes79: done

ChrisIsidora:
@funes79 Can you also run the `make docs` so that the .md (markdown) is created? (optionally an example tf in the example folder could be nice) Thanks in advance

funes79:
@ChrisIsidora  docs and examples added.

funes79:
@alldoami is it ok to merge?

alldoami:
If you run `make docs` do you get a diff?

ChrisIsidora:
@alldoami Added SKIP_SCIM_INTEGRATION_TESTS in #635 to fix the SCIM integration issue as it doesn't clean up properly if another test previously fails at some point or if Integrations tests are ran in parallel for different PRs.

ChrisIsidora:
> If you run `make docs` do you get a diff?

@alldoami I think @funes79  needs to remove the new docs for the procedure. But after merge there needs to be a new PR created after running `make docs`  because the new docs for resources/datasources are not automatically created after commit.

funes79:
I do not understand. On my local `make docs` runs fine. But I see that it failed in the Github actions. Is the any info why it causing the failure?

funes79:
> If you run `make docs` do you get a diff?

Yes I got a diff, so now I pushed a new commit to update the docs for procedure.

ChrisIsidora:
> I do not understand. On my local `make docs` runs fine. But I see that it failed in the Github actions. Is the any info why it causing the failure?

@funes79 I don't know exactly why but I had similar issues some time ago when I also introduced a new resources. I suspect that the `make docs` checks against the `main` branch and compares the docs in the PR branch and if the comparison yields a difference then it fails, but in this case for new resources it's logical that there is a difference because the resource is not there yet in the `main` branch. So I think the use case for the `make docs` check is primarily for flagging existing resources docs that needs to be updated if that resource was changed in the PR branch.

So fastest solution for now is to remove the new .md and after the merge, open a PR with only the new .md.

(Related to #632)

alldoami (approved):

PR #618, opened 2021-07-23
feat: Implemented Schemas, Tables and Views Datasources

ChrisIsidora:
Implemented following DataSources (which can be later on used to iterate over other resources):
- Schemas
- Tables
- Views
- Materialized Views 

## Test Plan
* [x] acceptance tests

PR #622, opened 2021-07-27
feat: Added Several Datasources Part 2

ChrisIsidora:
Part 2 of #618  (Sorry in advanced for the big PR)
Implemented following DataSources (which can be later on used to iterate over other resources):

- Stages
- File Formats
- Sequences
- Streams
- Tasks
- Pipes
- Masking Policies
- External Functions
- External Tables
- Warehouses
- Resource Monitors
- Storage Integrations

## Test Plan
* [x] acceptance tests

## References
https://docs.snowflake.com/en/sql-reference/sql/show.html
*


ChrisIsidora:
@alldoami Streams Acceptance Test fixed. Not sure why the warehouses, resource_monitors and storage_integrations are failing. Locally against my Snowflake account it works (see screenshot below), can you take a look if it fails again?
image

ChrisIsidora:
@alldoami Can you take a look why the warehouses, resource_monitors and storage_integrations are failing? Locally against my Snowflake account it works

ChrisIsidora:
@alldoami Did some research and it seems that the depends_on in the acceptance tests is triggering it, maybe it works locally on my side because I'm using TF 1.02? I found that there is a Test parameter for this `PreventPostDestroyRefresh` and I've added this, it still works locally on my side. Maybe you can run the integrations tests again to see if that helps? Thanks in advance.

ChrisIsidora:
@alldoami I was able to reproduce this locally. This tests are failing indeed due to the `depends_on`  in datasource but ONLY  in Terraform prior to `0.13`. This was solved in version `0.13` (See https://github.com/hashicorp/terraform/issues/17034) . When I use terraform prior to `0.13` like `0.12.31` I get the same error as in the integration run.
What are the options here, is it possible to upgrade to `>0.13` ? If not what do you suggest we do to make these tests pass? I really do need these datasources atm.

alldoami:
I'll upgrade tf for testing! Could you get rid of your workaround 😅

ChrisIsidora:
@alldoami Done!

alldoami (approved):

ChrisIsidora:
@alldoami Can you run it again? Seems it failing for no apparent reason?

PR #625, opened 2021-07-29
fix: Upgrade tf for testing

alldoami:
To address this problem: https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/622#issuecomment-888879621

Upgrading our tf version for testing!


Makefile
```diff
@@ -4,7 +4,7 @@ export DIRTY=$(shell if `git diff-index --quiet HEAD --`; then echo false; else
 LDFLAGS=-ldflags "-w -s -X github.com/chanzuckerberg/go-misc/ver.GitSha=${SHA} -X github.com/chanzuckerberg/go-misc/ver.Version=${VERSION} -X github.com/chanzuckerberg/go-misc/ver.Dirty=${DIRTY}"
 export BASE_BINARY_NAME=terraform-provider-snowflake_v$(VERSION)
 export GO111MODULE=on
-export TF_ACC_TERRAFORM_VERSION=0.12.29
+export TF_ACC_TERRAFORM_VERSION=0.13
```
edulop91, line 7: maybe add the readme that there's now a version requirement for the provider

edulop91 (approved):

Makefile
ChrisIsidora: @alldoami @edulop91 I think this is failing because it should be 0.13.x in this case 0.13.0 not only 0.13

ChrisIsidora:
@alldoami I couldn't clone your branch, I upgraded locally and figured it out  to fix the remaining failing tests:

For SCIM integration test failures:
- Somehow because of the change in the behavior of datasources in 0.13 and the previous failed tests the role AAD_PROVISIONER was never destroyed. It should be dropped in the Acceptance Tests Snowflake Account manually. The reason it's not a random role name is because snowflake only expects a set of roles with a specific name

For the current datasources the behavior has changed with 0.13 so to pass these tests the direct dependency should be added:
- `materialized_views_acceptance_test.go` => Add `depends_on = [snowflake_materialized_view.v]` in the data block
- `schemas_acceptance_test.go` => Add `depends_on = [snowflake_schema.s]` in the data block
- `tables_acceptance_test.go` => Add `depends_on = [snowflake_table.t]` in the data block
- `views_acceptance_test.go` => Add `depends_on = [snowflake_view.v]` in the data block
- `system_generate_scim_access_token_acceptance_test.go` => Add `depends_on = [snowflake_scim_integration.azured]` in the data block

PR #626, opened 2021-07-31
fix: Fixed Browser PKG

ChrisIsidora:
Attempt to fix https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/624#issuecomment-890273120


go.sum
```diff
@@ -215,21 +201,16 @@ github.com/bgentry/go-netrc v0.0.0-20140422174119-9fd32a8b3d3d h1:xDfNPAt8lFiC1U
...
 github.com/census-instrumentation/opencensus-proto v0.2.1/go.mod h1:f6KPmirojxKA12rnyqOA5BBL4O983OfeGPqjHWSTneU=
-github.com/certifi/gocertifi v0.0.0-20200211180108-c7c1fbc02894/go.mod h1:sGbDF6GwGcLpkNXPUTkMRoywsNa/ol15pxFe6ERfguA=
 github.com/cespare/xxhash v1.1.0/go.mod h1:XrSqR1VqqWfGrhpAt58auRo0WTKS1nRRg3ghfAqPWnc=
-github.com/chanzuckerberg/aws-oidc v0.23.1/go.mod h1:AbN6f7tGQ1aR8kJdr5Mpnyk3iabBONFFLlAVQG8UgqM=
-github.com/chanzuckerberg/go-misc v0.0.0-20200814145037-61fd36eedee6/go.mod h1:5GljuH2+xw/T7ovs4sRj9WbNRrgePgom8sMzJJPlAAo=
 github.com/chanzuckerberg/go-misc v0.0.0-20201208181439-ea3e3e39e194 h1:Te8+Bfs5/1LQ9v0r8h7ocrCR7WXGp+Y61dbQhPIVtuc=
 github.com/chanzuckerberg/go-misc v0.0.0-20201208181439-ea3e3e39e194/go.mod h1:I+XR+Hhcty6cSO2ZBXNmlvxeGOr6yhE+JfEDQyOFtv4=
-github.com/chanzuckerberg/go-misc v0.0.0-20210727205931-6daf95f6727e h1:LyRaq/Sz1omD+bBuVwZBVxAv3qEUHAzgol5IFdG1YtY=
-github.com/chanzuckerberg/go-misc v0.0.0-20210727205931-6daf95f6727e/go.mod h1:mQwH9ExhxX/18odRhummSPc21vJ7TqB5y8u4uYv+1h4=
+github.com/cheggaaa/pb v1.0.27/go.mod h1:pQciLPpbU0oxA0h+VJYYLxO+XeDQb5pZijXscXHm81s=
```
ChrisIsidora, line 213: @alldoami Is you run `go mod graph` you could see that `go-misc v0.0.0-20210727205931-6daf95f6727e` was added the browser depedency that is breaking the release?

alldoami (approved):

alldoami:
Thanks so much @ChrisIsidora !

go.sum
alldoami: Could you explain how this works for future reference 😅

ChrisIsidora: @alldoami I also don't know. All I know is that when the go mods where updated the go-misc mod of chanzuckerberg was also updated and in the new updated version it's pulling a new version of the indirect dependency browser pkg which broke the release.  Using `go mod graph` you can find out where the indirect dependency is coming from, which in this case was flgo-misc

PR #632, opened 2021-08-03
fix: Compiled missing docs

robbruce:
Docs hadn't been ran for new resources and data sources. This PR adds in the newly generated documents.


ChrisIsidora:
@robbruce It used to be automated in the pipeline/actions? Is this not the case anymore?

robbruce:
> @robbruce It used to be automated in the pipeline/actions? Is this not the case anymore?

@ChrisIsidora - no idea, I'm not a maintainer, just an active contributor!

@alldoami - some guidance on what you'd like people to do with docs would be great, if people need to build themselves then maybe an extra check should be in place to see if the built docs would differ from whats in the PR?

alldoami (approved):

PR #634, opened 2021-08-04
fix: Added missing view privilege

ChrisIsidora:
Added missing view privilege: References

## References
https://docs.snowflake.com/en/user-guide/security-access-control-privileges.html#view-privileges

*

### DIFF
--- a/pkg/resources/view_grant.go
+++ b/pkg/resources/view_grant.go
@@ -9,6 +9,7 @@ import (
 
 var validViewPrivileges = NewPrivilegeSet(
 	privilegeSelect,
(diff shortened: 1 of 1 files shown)


alldoami (approved):

pkg/resources/view_grant.go
```diff
@@ -9,6 +9,7 @@ import (
 
 var validViewPrivileges = NewPrivilegeSet(
 	privilegeSelect,
+	privilegeReferences,
 	privilegeOwnership,
```
alldoami, line 13: Nit: alphabetize this list please :)

PR #639, opened 2021-08-09
feat: Add private key passphrase support

### LINKED ISSUE #251: Passphrase for private key
Hi contributors!
 would it be possible to add a passphrase support for private key authentication? 
Because the existing Snowsql documentation requires an encrypted private key, and here the provider requires a non-encrypted key. 

https://docs.snowflake.com/en/user-guide/snowsql-start.html
_The private key file must be encrypted,  and the passphrase must be specified in the environment variable SNOWSQL_PRIVATE_KEY_PASSPHRASE_

Thank you!

robbruce:
Add support for a private key passphrase, this does add a new dependency on github.com/youmark/pkcs8.

## Test Plan

Could not determine best approach to test, as there aren't any existing unit tests for a private key. However have tested with terraform code with a local install against both with and without a private key being required.

## References

* Resolves #251


alldoami:
Hmm it would be great if we could add some testing to this...

robbruce:
> Hmm it would be great if we could add some testing to this...

@alldoami - I agree, but could do with some guidance of an acceptable approach from the code maintainers as I notice that there aren't any existing tests for any private key authentication. Essentially, are you happy for private keys to be held in the test code?

alldoami:
This looks good to me right now, hopefully we can add testing later! If you could keep an eye out for issues (if they do come up) related to this, would love your input there!

alldoami (approved):

robbruce:
> This looks good to me right now, hopefully we can add testing later! If you could keep an eye out for issues (if they do come up) related to this, would love your input there!

@alldoami - will do, thanks for accepting

JimMcKenzieSmith:
@robbruce I was having the same issue as @advaitvdeo, so I followed the link you posted about the library only supporting certain ciphers.  This did get me a step further.  However, now I'm getting a new error:
```
Error: error creating api integration: 390144 (08004): JWT token is invalid.
```
Here is the key I generated, per that documentation:
```
cd ~/.ssh
openssl genrsa -out rsa_key -4096
openssl rsa -in rsa_key -pubout -out rsa_key.pub
openssl pkcs8 -topk8 -inform pem -in rsa_key -outform PEM -v2 aes-256-cbc -out rsa_key.p8
```
One interesting thing, when I use this key to authenticate via snowsql on the command line, it works!  So, the key and passphrase seem to be good, it just doesn't work with this terraform.  Let me know if you have any ideas on what to try.  Thanks,

JimMcKenzieSmith:
Update: it appears that the "JWT token is invalid" seems to be only happening on my local machine.  Once I ran this from Jenkins it was a success.  I suspect that this is due to a timing issue or a clock difference between my local machine and Snowflake servers.

PR #653, opened 2021-08-20
fix: Stream append only

koslam:
This change updates the logic used when reading the data for the `snowflake_stream` resource to determine whether or not the stream is append only.

Per the documentation at <https://docs.snowflake.com/en/sql-reference/sql/show-streams.html>, we can determine whether a stream is append only by checking the value of the `mode` for the stream. If the value is `APPEND_ONLY`, then the stream is append only.

## Test Plan
* [X] Unit Tests
* [x] Acceptance Tests

If possible I would like feedback on the acceptance tests. I'm not terribly familiar with terraform provider testing.

## References

* Related issue: <https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/652>


pkg/resources/stream_acceptance_test.go
```diff
@@ -71,3 +71,67 @@ resource "snowflake_stream" "test_stream" {
 `
 	return fmt.Sprintf(s, name, name, name)
 }
+
+func TestAcc_Stream_AppendOnly(t *testing.T) {
```
alldoami, line 75: Let's move this up to where the other test is `TestAcc_Stream`. Around line 33

alldoami: Just like we do here: https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/schema_grant_acceptance_test.go#L30-L35

koslam: I've moved the test into the existing method and updated the `streamConfig` function to allow the caller to specify whether the stream is append only. I've tried running the acceptance tests both locally and through Travis, but I seem to be missing something in the configuration. Are there docs for how to run the acceptance tests so I can verify they work?

koslam: Never mind, I managed to execute them using Travis CI. I wasn't setting the Go version correctly and it was defaulting to 1.10. The `TestAcc_Stream` test passed.

alldoami (approved):

PR #661, opened 2021-08-24
feat: Add option to create streams which reference EXTERNAL TABLEs

momer:
Allow the creation of `STREAMS` on `EXTERNAL TABLE`s.

## Test Plan

* [x] acceptance tests
  * Note: I haven't run these against an account, and there's some copy/pasta from the `external_table` which I modified a bit, so is a bit fragile.
* [x] unit tests 

## References

* https://docs.snowflake.com/en/sql-reference/sql/create-stream.html#creating-an-insert-only-stream-on-an-external-table

### DIFF
--- a/pkg/resources/stream_acceptance_test.go
+++ b/pkg/resources/stream_acceptance_test.go
@@ -40,6 +40,18 @@ func TestAcc_Stream(t *testing.T) {
 					checkBool("snowflake_stream.test_stream", "show_initial_rows", false),
 				),
 			},
+			{
+				Config: externalTableStreamConfig(accName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "name", accName),
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "database", accName),
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "schema", accName),
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "on_table", fmt.Sprintf("%s.%s.%s", accName, accName, "STREAM_ON_TABLE")),
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "comment", "Terraform acceptance test"),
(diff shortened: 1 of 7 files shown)


pkg/resources/stream_acceptance_test.go
```diff
@@ -40,6 +40,18 @@ func TestAcc_Stream(t *testing.T) {
 					checkBool("snowflake_stream.test_stream", "show_initial_rows", false),
 				),
 			},
+			{
+				Config: externalTableStreamConfig(accName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "name", accName),
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "database", accName),
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "schema", accName),
+					resource.TestCheckResourceAttr("snowflake_stream.test_stream", "on_table", fmt.Sprintf("%s.%s.%s", accName, accName, "STREAM_ON_TABLE")),
```
alldoami, line 49: Is this supposed to be `on_external_table`?

alldoami: And I think the name is supposed be `accName`, as well... not `STREAM_ON_TABLE`

momer: Hey @alldoami `on_external_table` is a `bool`, and creating a stream on an external table without `... EXTERNAL TABLE` will raise an error; so, the acceptance test is simply ensuring that the table referenced by `on_table` (which is an external table created via `externalTableStreamConfig`) exists. 

That said, I should definitely have added a check that `on_external_table` is `true` - see the most recent commits, which also resolve the `STREAM_ON_TABLE` reference - thanks! I'll have to set up an environment to run these tests locally which will reduce impact on you via these PRs, sorry.

michael-robbins:
How's this going? Just stumbled across this issue myself!

alldoami:
@momer Please reopen when ready for another review.

PR #692, opened 2021-09-27
fix: Allow empty result when looking for storage integration on refresh

bneijt:
Trying to refresh the terraform state after deleting an integration, I get the error:
`Error: Could not show storage integration: sql: no rows in result set`

The issue seems to be that looking for a specific integration does deal with not finding that specific integration, triggering an error instead of allow the refresh flow to continue.

Steps to reproduce:
- Create a `snowflake_storage_integration` using the provider
- Delete the resulting integration
- Execute `terraform refresh` to update the state

Expected result:
- The provider should notice the resource gone and update the state
Actual result:
- The provider errors out with the message: `Error: Could not show storage integration: sql: no rows in result set`

Terraform version: 1.0.7
Provider version: 0.25.19

I'm not a go developer, so this PR probably needs tests.


pkg/resources/storage_integration.go
```diff
@@ -160,6 +160,10 @@ func ReadStorageIntegration(d *schema.ResourceData, meta interface{}) error {
 	// Some properties can come from the SHOW INTEGRATION call
 
 	s, err := snowflake.ScanStorageIntegration(row)
+	if err == sql.ErrNoRows {
+		// Integration not found, maybe it was already deleted and we are running a refresh
+		return err
+	}
```
edulop91, line 166:
```suggestion
		if err == sql.ErrNoRows {
			// If not found, mark resource to be removed from statefile during apply or refresh
			log.Printf("[DEBUG] storage integration (%s) not found", d.Id())
			d.SetId("")
			return nil
		}
```
I think you might want something like this so we won't error out and clear the state for this resource when not found

bneijt:
Turns out you can nolonger `terraform plan` anything if the storage integration from your current state is missing, so I filed a bug to make it more visible to other users.

#696

alldoami:
@bneijt could you fix the conflicts and I'll look at it again after? Sorry for the late review!

alldoami:
Thanks for working on this!

PR #697, opened 2021-09-28
docs: Add extensive example of setting up a storage integration

bneijt:
If you do not define the role as a variable, you will end up with a cycle between role and integration, so I thought it would be nice to add a full example for other people to see that relation.

## References

https://docs.snowflake.com/en/user-guide/data-load-s3-config-storage-integration.html


alldoami:
Hello! Please update this file here: https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/examples/resources/snowflake_storage_integration/resource.tf instead of changing it in this md. We use a plugin that gathers various information across files and compiles it into one.

bneijt:
The PR has been open for a few months now.

Is there something I'm missing?

alldoami:
Ah sorry @bneijt could you fix the conflicts and I'll merge after!

bneijt:
Ah, totally forgot about this PR. The changes are still good, but never got to rebase it.

Better luck next time.

PR #728, opened 2021-10-21
docs: formatting and consistency

farfromunique:
* Updated the provider block example with more comments, and broke the auth-related parameters into groups for easier understanding
* Added corresponding environment variables to provider block example and to Argument Reference
* Improved formatting and consistency of Argument reference
* Added "Conflicts with" data from [provider.go](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/provider/provider.go)

## Test Plan
Visual validation (if possible) that Terraform's Markdown formatter supports footnotes with the `[^n]` syntax


alldoami:
@farfromunique could you reopen once you've fixed?

PR #745, opened 2021-11-02
fix: refactor ReadWarehouse function to correctly read object parameters

### LINKED ISSUE #739: Terraform plan always finds diff for warehouse parameters even if changes are already applied
**Provider Version**

v0.25.23

**Terraform Version**

 v1.0.9
 
**Describe the bug**

When we run `terraform plan`, terraform detects changes for following warehouse parameters even if they are already applied.
```
max_concurrency_level
statement_queued_timeout_in_seconds
statement_timeout_in_seconds
```

_Example Diff:_ 

```
~ resource "snowflake_warehouse" "warehouse" {
      id                                  = "BACKUP"
    + max_concurrency_level               = 0
      name                                = "BACKUP"
    + statement_queued_timeout_in_seconds = 0
      # (9 unchanged attributes hidden)
  }
```

_Query:_
```
show parameters in warehouse backup;
```

_Result:_

```tsv
key	value	default	level	description	type
MAX_CONCURRENCY_LEVEL	8	8	WAREHOUSE	Maximum number of SQL statements a warehouse cluster can execute concurrently before queuing them. Small SQL statements count as a fraction of 1.	NUMBER
STATEMENT_QUEUED_TIMEOUT_IN_SECONDS	0	0		Timeout in seconds for queued statements: statements will automatically be canceled if they are queued on a warehouse for longer than this amount of time; disabled if set to zero.	NUMBER
STATEMENT_TIMEOUT_IN_SECONDS	0	172800	WAREHOUSE	Timeout in seconds for statements: statements are automatically canceled if they run for longer; if set to zero, max value (604800) is enforced.	NUMBER
```

As you can see from the query result, `terraform plan` reports different values (zero-value) for these warehouse parameters.  
After some digging, I found that provider is using `SHOW WAREHOUSES` query to find values for these parameters ([ref](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/snowflake/warehouse.go#L63)). But according to [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/sql/show-warehouses.html), output of `SHOW WAREHOUSES`does not return any of these columns. 

**Expected behavior**

It should not find any changes.

sonmezonur:
Fixes #739 

#598 has introduced 3 additional parameters (`statement_timeout_in_seconds`, `statement_queued_timeout_in_seconds` and `max_concurrency_level`) for warehouse resource. After adding this feature, `terraform plan` has started to consistently generate diff for these parameters even when they are already applied.  Currently, provider is using `SHOW WAREHOUSES LIKE $warehouse_name` query but [according to doc](https://docs.snowflake.com/en/sql-reference/sql/show-warehouses.html), these columns are not included in the query output. That is why generated diff always reports `zero-value` for [these fields](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/snowflake/warehouse.go#L45-L47).

After this update, ReadWarehouse function will run the additional query (`SHOW PARAMETERS IN WAREHOUSE $warehouse_name`) to fetch warehouse parameters. Not sure if this is the best way but I created a new struct (`WarehouseBuilder`) by embedding/overriding generic [Builder](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/8e1ad2b399e39463797171aff6bc864bc79046ce/pkg/snowflake/generic.go#L26) ([TaskBuilder](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/bde252ef2b225b128728e2cd4f2dcab62a65ba58/pkg/resources/task.go#L336-L343) uses a similar approach). Updating `schema.ResourceData` by using the output of this query does not generate diff anymore.


alldoami:
Could you give a bit more description on what you did in this PR?

sonmezonur:
Hello @alldoami . Updated PR description to give more details on this issue. I hope that makes it clearer.

sonmezonur:
Hello @alldoami. I would like to fix integration tests but don't know how to test and work on it locally. Requires account and username to run tests. Could you please help with this?

alldoami:
You can set these account variables: https://github.com/chanzuckerberg/terraform-provider-snowflake#pull-request-ci
and run `make test-acceptance` locally.

sonmezonur:
Hello @alldoami . I updated integration tests and run `make test-acceptance` locally as you suggested. Could you approve workflows to run tests again?

pkg/resources/warehouse.go
```diff
@@ -105,7 +106,7 @@ var warehouseSchema = map[string]*schema.Schema{
 	"statement_timeout_in_seconds": {
 		Type:        schema.TypeInt,
 		Optional:    true,
-		Default:     0,
+		Default:     172800,
```
sonmezonur, line 109: Default value for this parameter is `172800` (https://docs.snowflake.com/en/sql-reference/parameters.html#statement-timeout-in-seconds).  I think provider defaults should be the same as Snowflake defaults. What do you think?

sonmezonur:
Should fix https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/601

alldoami (approved):

PR #748, opened 2021-11-05
fix: Procedure Resource Panics with NOT NULL in Return Type

BrentSouza:
The procedure resource does a regex match for the return type it receives out of DESCRIBE PROCEDURE which fails to match against return types that specify a NOT NULL constraint.  It also updates state to just the type of the match and does not include length which causes thrash on the resources.

This PR will do three things:

1. Support optional specification of NOT NULL in the regex
2. Check that match has a length greater than 0 before indexing into it (this was causing the panic)
3. Set the state to match[0] which is the full type to avoid plan thrash

## Test Plan
* Additional unit tests to verify the above outcomes


sfc-gh-swinkler:
closing this since we got rid of regex expression. if you feel this needs to be changed, please open a new PR

PR #768, opened 2021-11-26
feat: Allow setting resource monitor on account

### LINKED ISSUE #721: [Question] Assign resource monitor to account level?
Hey, 
it's possible to create `snowflake_resource_monitor` resources as well as assigning them to a `snowflake_warehouse`.
What I did not find is assigning a monitor to the account e.g. `ALTER ACCOUNT SET RESOURCE_MONITOR = "resource_monitor_account";`

Did I miss anything or is this currently not supported/implemented?

gouline:
Allow setting resource monitor on the account scope.

Fixes #721

## Test Plan

* [x] acceptance tests

## References

* [ALTER ACCOUNT](https://docs.snowflake.com/en/sql-reference/sql/alter-account.html)


alldoami:
```
=== CONT  TestAcc_ResourceMonitor
    resource_monitor_acceptance_test.go:16: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=15) "set_for_account": (string) (len=4) "true"
        }
--- FAIL: TestAcc_ResourceMonitor (6.76s)
```

pkg/resources/resource_monitor_acceptance_test.go
```diff
@@ -23,6 +23,14 @@ func TestAcc_ResourceMonitor(t *testing.T) {
 					resource.TestCheckResourceAttr("snowflake_resource_monitor.test", "credit_quota", "100"),
 				),
 			},
+			{
+				Config: resourceMonitorConfigSetForAccount(name, true),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_resource_monitor.test", "name", name),
+					resource.TestCheckResourceAttr("snowflake_resource_monitor.test", "credit_quota", "100"),
+					resource.TestCheckResourceAttrSet("snowflake_resource_monitor.test", "set_for_account"),
```
alldoami, line 31:
```suggestion
					resource.TestCheckResourceAttrSet("snowflake_resource_monitor.test", "set_for_account", true),
```

I think?

gouline:
Hopefully this fixes it. Not quite sure what's causing it and cannot reproduce locally (without a test Snowflake account), so best I can do is get it as close to network_policy_attachment_acceptance_test.go that it's based on as possible.

gouline:
Any ideas? Everything looks identical to network_policy_attachment and yet it fails.

gouline:
I think I understand now: network_policy_attachment just disables `ImportStateVerify` because it doesn't read the resource state and hence has no way to verify it. Resource monitors have a `level` column that's set to `ACCOUNT` when `set_for_account` is true, so I added that to `ReadResourceMonitor`. 

Does this look right?

alldoami (approved):

alldoami:
Thank you!

Issue #677, opened 2021-09-13
Can't create user with USERADMIN, only with SECURITYADMIN with resource snowflake_user


jaisojar:
**Provider Version**

0.25.18

**Terraform Version**

Terraform v1.0.4

**Describe the bug**
Trying to create a new user using the USERADMIN role throws:
│ Error: 003001 (42501): SQL access control error:
│ Insufficient privileges to operate on account 'XXXX'
│ 
│   with snowflake_user.baby

I can use the SECURITYADMIN role, but it seems I should be able to use USERADMIN as well t create users.

**Expected behavior**

A user appears in snowflake, as specified in the terraform resource snowflake_user.user. I want to use the USERADMIN snowlfkae built in role. Only SECURITYADMIN works.

**Code samples and commands**

This does not work:
```
resource "snowflake_user" "baby" {
  provider = snowflake.useradmin
  name     = "BABY"
}
```
However this work just fine:
```
resource "snowflake_user" "baby" {
  provider = snowflake.securityadmin
  name     = "BABY"
}
```

Also it would be helpful to have clearer logging output, which SQL statement threw the error.

**Additional context**

NA

sshah90:
After debugging for some time finally arrived at this issue.

I am facing the same issue, although from Snowflake UI I am able to create user with `useradmin` role.

PR #769, opened 2021-11-26
fix: use "DESCRIBE USER" in ReadUser, UserExists

mcrlc:
This PR is supposed to solve #677 by using `DESCRIBE USER ...` instead of `SHOW USERS LIKE ...` in the `ReadUser` and `UserExists` functions.
While `SHOW USERS ...` requires the `MANAGE_GRANTS` global privilege, which usually requires the `SECURITYADMIN` role, the `DESCRIBE USER ...` does not.

## Test Plan
* [ ] acceptance tests

## References

* https://docs.snowflake.com/en/sql-reference/sql/show-users.html#usage-notes
* https://docs.snowflake.com/en/sql-reference/sql/desc-user.html#usage-notes

### DIFF
--- a/pkg/snowflake/user_test.go
+++ b/pkg/snowflake/user_test.go
@@ -15,6 +15,9 @@ func TestUser(t *testing.T) {
 	q := u.Show()
 	r.Equal("SHOW USERS LIKE 'user1'", q)
 
+	q = u.Describe()
+	r.Equal(`DESCRIBE USER "user1"`, q)
+
 	q = u.Drop()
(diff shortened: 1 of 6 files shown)


pkg/snowflake/user_test.go
```diff
@@ -15,6 +15,9 @@ func TestUser(t *testing.T) {
 	q := u.Show()
```
alldoami, line 15: Would we get rid of this now?

alldoami (approved):

PR #789, opened 2021-12-14
Fix: Shares can't be updated on table_grant resource

chriskuchin:
The difference function in the table_grant resource had a hard coded string meaning it only ever calculated the difference for roles not shares.


chriskuchin:
@alldoami  would love any guidance to get this incorporated as soon as possible. Let me know if you would like to see anything more on this seems pretty straightforward to me

alldoami:
Would be great if you could add a test here! https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/resources/external_table_grant_test.go

chriskuchin:
@alldoami I am a little confused does the external table grant use the table_grant?

chriskuchin:
@alldoami I added a test to `table_grant_test.go` I have verified that it fails when bug is present. Assumed this is what you wanted above. Let me know if there is anything else needed here.

alldoami:
Thank you for adding the test! It looks good.

alldoami (approved):

PR #792, opened 2021-12-19
fix: Recreate notification integration when type changes

gouline:
Edge case when changing the type of notification integration (e.g. from `AWS_SQS` to `AWS_SNS`) with the same name, Terraform attempts to `ALTER` it, which results in an error because required fields are different depending on the type.

This change forces recreation when any of the integration types change.

## Test Plan

* [x] acceptance tests

## References

* [ALTER NOTIFICATION INTEGRATION](https://docs.snowflake.com/en/sql-reference/sql/alter-notification-integration.html)


alldoami:
run `make docs`

gouline:
Sorry forgot 🤦  Done now!

alldoami (approved):

PR #795, opened 2021-12-22
fix: tagging for db, external_table, schema

sfc-gh-swinkler:
tagging for database, external_table and schema were not working. this is caused by two separate problems 1) the command for updating tags never runs and 2) the command itself was incorrect. This PR aims to fix both of these issues. In addition, tags were previously marked as force new, when they really shouldn't be. It would be very bad to accidentally delete a database just because the tags are getting changed, for example. A consequence of removing force new means that we now have to implement update command for external_table, which uses tags, but does not have its own update command. 

## Test Plan
* [ ] acceptance tests

## References

*

### DIFF
--- a/pkg/resources/resource.go
+++ b/pkg/resources/resource.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"database/sql"
+	"log"
 
 	"github.com/chanzuckerberg/terraform-provider-snowflake/pkg/snowflake"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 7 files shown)


pkg/resources/resource.go
```diff
@@ -34,15 +35,13 @@ func CreateResource(
 				case schema.TypeInt:
 					valInt := val.(int)
 					qb.SetInt(field, valInt)
-				case schema.TypeList:
```
alldoami, line 37: Why are you getting rid of this?

alldoami:
Could you add tests to test your new update functions?

pkg/resources/resource.go
sfc-gh-swinkler: all that does is type conversion, but it didnt actually work, as it wasn't converting the types properly. we already have a function getTags() that does work. so using that instead

sfc-gh-swinkler:
> Could you add tests to test your new update functions?

sure i will work on this

sfc-gh-swinkler:
i added a test case for update and also simplified the update to only handle tags, as that is the purpose of this PR. leaving force new for all other attributes, as that should be handled in a separate pr.

alldoami (approved):

PR #798, opened 2021-12-26
feat: Support create function with Java language

kurochan:
I implemented `snowflake_function` for Java UDF.

## References
* https://docs.snowflake.com/en/sql-reference/sql/create-function.html


alldoami (approved):
Thanks!

PR #803, opened 2021-12-30
docs: update docs to accommodate upstream changes

fernandoataoldotcom:
Adjustments to documentation to reflect changes in `go` and snowflake `tasks`. 

* README.md: replace `go get` command
* docs/resources/task.md: add `MINUTE` component to scheduler example

## Test Plan
No tests plan, since only documentation was updated

## References
* https://go.dev/doc/go-get-install-deprecation
* https://docs.snowflake.com/en/sql-reference/sql/create-task.html


alldoami (changes requested):
You'll need to update https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/examples/resources/snowflake_task/resource.tf in order to change these docs. Please update the resource.tf file and then run `make docs`.

fernandoataoldotcom:
> You'll need to update https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/examples/resources/snowflake_task/resource.tf in order to change these docs. Please update the resource.tf file and then run `make docs`.

Sorry I missed that, and thank you for the notes on what to do next.  I've made the change and pushed.

Issue #616, opened 2021-07-23
Support for `SECURITY INTEGRATION`


LouisAmon:
**Describe the solution you'd like**

Support for for the `SECURITY INTEGRATION` resource

**Additional context**

https://docs.snowflake.com/en/sql-reference/sql/create-security-integration.html

There's only one per account, so we should specify that in the docs I think (same as the `network_policy_attachment`)

ChrisIsidora:
@LouisAmon There are different types of Security Integration. The SCIM type is already implemented. Is there a particular one that you require?

LouisAmon:
@ChrisIsidora I meant the SAML2 integration : it typically involves a lot of back and forth between two different services so it's a great use-case for IaC ;)

kmertz-rga:
Agreed, SSO/SAML2 needs to be available in the terraform provider.

intinfinite:
I think this issue can be closed with recent pull request

sfc-gh-asawicki:
We are closing this issue as part of a cleanup described in [announcement](https://github.com/Snowflake-Labs/terraform-provider-snowflake/discussions/2755). If you believe that the issue is still valid in v0.89.0, please open a new ticket.

PR #805, opened 2022-01-05
feat: Allow creation of saml2 integrations (#616)

intinfinite:
- Allows creation of security integrations with type=SAML2 ([example](https://docs.snowflake.com/en/sql-reference/sql/create-security-integration.html#id6))

## Test Plan
Created integration and acceptance tests

## References
This PR should address issue #616 
https://docs.snowflake.com/en/sql-reference/ddl-user-security.html#security-integrations
https://docs.snowflake.com/en/sql-reference/sql/create-security-integration.html#id6


intinfinite:
> Integration tests failure for [45875c5](https://github.com/chanzuckerberg/terraform-provider-snowflake/actions/runs/1659742853)

Is this an issue on my end?

alldoami:
@intinfinite yes, you can see the logs if you click on the link in the comment.

intinfinite:
Thanks, I think these should fix it

intinfinite:
Looks like it's failing because a SAML2 integration already exists on the account.

```
=== RUN   TestAcc_SamlIntegration
    saml_integration_acceptance_test.go:20: Step 1/1 error: Error running apply: exit status 1
        
        Error: error creating security integration: 390956 (22023): There can only be one enabled SAML2 integration
        
          on terraform_plugin_test.tf line 2, in resource "snowflake_saml_integration" "test_saml_int":
           2: 	resource "snowflake_saml_integration" "test_saml_int" {
        
        
--- FAIL: TestAcc_SamlIntegration (2.40s)
```

Do you know if there is an active SAML2 integration on the runner's Snowflake account the tests are run on or if the integration I created wasn't dropped since the first test failed?

You can check by seeing if any records are returned when running the below with ACCOUNTADMIN:
```sql
show security integrations;
```

If there is a SAML2 integration being used on the account, I can change the test to create a disabled security integration instead.

Thanks!

alldoami:
Hi @intinfinite! Thanks for the thorough comment. Yes we are using the SAML2 Integration, so the test to disable sounds good.

intinfinite:
Sorry for the multiple iterations.  Think this should be the last one *fingers crossed*.

It looks like there is a remnant `SECURITY INTEGRATION` from one of the previous failures that's preventing this from being run and would need to be manually dropped/destroyed before this test will pass.

```
=== RUN   TestAcc_SamlIntegration
    saml_integration_acceptance_test.go:20: Step 1/1 error: Error running apply: exit status 1
        
        Error: error creating security integration: 390957 (22023): Issuer URLs must be distinct across SAML2 integrations.
        
          on terraform_plugin_test.tf line 2, in resource "snowflake_saml_integration" "test_saml_int":
           2: 	resource "snowflake_saml_integration" "test_saml_int" {
        
        
--- FAIL: TestAcc_SamlIntegration (3.09s)
```

Can view if there are any test integrations like this:
```sql
SHOW SECURITY INTEGRATIONS LIKE '%test%';
```

And then drop them like this:
```
DROP SECURITY INTEGRATION "test_saml_integration";
```

alldoami (changes requested):
We cannot test this without modifying our snowflake account's saml integration. There can only be one saml integration at a time. Please do not run this until there are changes made to not create another saml integration.

intinfinite:
You should be able to have multiple security integrations of the same type, just cannot have more than one enabled at a time (see screenshot below).

I updated the test to create a disabled integration, but I think there might be a test integration that's leftover from a previous run that failed before it could be removed.

image

alldoami:
I checked for the `test_saml_integration` but do not see it on my end...

intinfinite:
Thanks for checking.  I think I see what the issue is. 

I used to have the name hard-coded as "test-saml-integration", but changed it to be random to be more in line with the other tests (like the API integration or scim integration).

You can be sure that it's the dummy integration that was created before by describing it using:

```sql
desc integration "jumblednameitgenerated";
```

If the `SAML2_SSO_URL` or `SAML2_ISSUER` has test in it, it's probably the remnant, like below:

image

alldoami:
yup I believe I deleted the one with the random string. Cool that the tests succeeded!

intinfinite:
Awesome, thanks for putting up with me.  First time contributing to this project.

alldoami:
Going to have @edulop91 do a passover, but in the meantime, can you pull in the most recent changes @intinfinite?

intinfinite:
Sounds good. Merged from main.

alldoami:
Hm still looks out of sync with `main`

intinfinite:
All set

pkg/resources/saml_integration.go
```diff
@@ -0,0 +1,476 @@
...
+			return errors.Wrap(err, "error updating security integration")
+		}
+	}
+
+	return ReadSAMLIntegration(d, meta)
+}
+
+// DeleteSAMLIntegration implements schema.DeleteFunc
+func DeleteSAMLIntegration(d *schema.ResourceData, meta interface{}) error {
+	return DeleteResource("", snowflake.ScimIntegration)(d, meta)
```
edulop91, line 475: `snowflake.ScimIntegration` seems off -- is this snowflake.SAMLIntegration?

intinfinite: good catch, fixed in latest commit

alldoami (approved):

pkg/resources/saml_integration.go
edulop91: thank you!

PR #819, opened 2022-01-13
fix: read Name, Database and Schema during Procedure import

dkduo:
Update ReadProcedure() to set name, database and schema during import.

Refactor ArgumentsSignature() to both remove the RETURN portion and to
uppercase the output. The RETURN is technically not part of the unique signature 
of a procedure and will not exist when performing an import unless the ID
is updated to include it. Uppercasing the output means the signature will match 
the output of the Snowflake SHOW command.

Change default value of comment to `user-defined procedure`.

Refactor unit test for Read to more thoroughly test reading an existing procedure
with the minimal initial values provided as would be the case during an import.

Update acceptance test to include testing an import.

## Test Plan
* [X] unit tests
* [X] acceptance tests

## References
* https://docs.snowflake.com/en/sql-reference/sql/show-procedures.html
* https://docs.snowflake.com/en/sql-reference/sql/create-procedure.html
"stored procedures are identified and resolved by the combination of the name and argument types"

### DIFF
--- a/pkg/snowflake/procedure.go
+++ b/pkg/snowflake/procedure.go
@@ -42,9 +42,10 @@ func (pb *ProcedureBuilder) QualifiedNameWithoutArguments() (string, error) {
 	return fmt.Sprintf(`"%v"."%v"."%v"`, pb.db, pb.schema, pb.name), nil
 }
 
-// Returns the arguments signature of the procedure in a form <procedure>(<type>, <type>, ..) RETURN <type>
+// Returns the arguments signature of the procedure in a form <procedure>(<type>, <type>, ..)
+// Removed the RETURNS portion because it will not exist during import and is not part of the unqiue signature
 func (pb *ProcedureBuilder) ArgumentsSignature() (string, error) {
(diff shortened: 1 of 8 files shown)


pkg/snowflake/procedure.go
```diff
@@ -230,7 +230,7 @@ type procedure struct {
 	Name         sql.NullString `db:"name"`
 	SchemaName   sql.NullString `db:"schema_name"`
 	Text         sql.NullString `db:"text"`
-	DatabaseName sql.NullString `db:"database_name"`
+	DatabaseName sql.NullString `db:"catalog_name"`
```
alldoami, line 233: This looks like a mistake?

dkduo: For some reason, Snowflake returns the database name in a column called `catalog_name` when calling the SHOW PROCEDURES query.
https://docs.snowflake.com/en/sql-reference/sql/show-procedures.html#output

alldoami: well that's weird and not uniform...

alldoami (approved):

PR #820, opened 2022-01-18
fix: add missing quotes to resources in documentation

florianMalbranque:
Fix documentation by adding quotes to resources naming & generate doc.

## Test Plan

seems like no test is needed.


alldoami:
In newer versions of terraform, quotes are no longer needed for resources. Going to close this PR

PR #823, opened 2022-01-19
fix: File format optionals

gouline:
Current behaviour requires you to explicitly specify all optional parameters for file formats, like the Parquet example:

```hcl
resource "snowflake_file_format" "parquet" {
  database    = "DATABASE"
  schema      = "PUBLIC"
  name        = "PARQUET"
  format_type = "PARQUET"

  # Optionals
  compression    = "AUTO"
  binary_as_text = true
  null_if        = []
  trim_space     = false
}
```

These optional parameters are rarely changed from the sensible defaults that Snowflake uses if you omit them from creation. For some format types, there are many more of these parameters, e.g. 15 for JSON! So it would be nice if you could just write the following:

```hcl
resource "snowflake_file_format" "parquet" {
  database    = "DATABASE"
  schema      = "PUBLIC"
  name        = "PARQUET"
  format_type = "PARQUET"
}
```

Equivalent to the following SQL used frequently in Snowflake documentation:

```sql
create file format database.public.parquet type = parquet;
```

This pull request fixes the current behaviour where omitted optional parameters are still included in the create SQL statement as their type-specific 'zero' values (nil, false, 0, [], etc) and marks them as computed to prevent it from attempting to alter them back to those zero values on resource update.

## Test Plan

* [x] acceptance tests

## References

* [CREATE FILE FORMAT](https://docs.snowflake.com/en/sql-reference/sql/create-file-format.html)


gouline:
Disregard this, there appears to be no way to distinguish between omitted and zero value, so consciously setting a boolean value to false results in Snowflake's default being applied (which may be true, depending on the value). What I'm trying to do doesn't seem possible.

PR #830, opened 2022-01-25
feat: Task error integration

gouline:
Support for private preview task error integration feature. Identical to #595 with pipes, except for tasks.

## Test Plan

* [x] acceptance tests

## References

* https://docs.snowflake.com/en/LIMITEDACCESS/task-notifications.html


alldoami:
Do you think you'd be able to add any integration tests for this?

gouline:
Integration tests would be easy to add, but they would fail for anyone running against a Snowflake account that doesn't have this private preview feature enabled.

I added the new field to a few unit tests (and fixed an issue that it discovered), so hopefully that's good enough for now.

alldoami (approved):

PR #831, opened 2022-01-25
fix: add warehouses attribute to resource monitor

sfc-gh-swinkler:
Currently resource monitors can only be set at the account level, or not at all. This PR adds an attribute to the resource monitor called "warehouses" which is a set of string corresponding to warehouse names. These warehouses will be assigned to the resource monitor when it is created. 

https://docs.snowflake.com/en/user-guide/resource-monitors.html#assigning-warehouses-to-a-resource-monitor

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


alldoami:
run `make docs`

sfc-gh-swinkler:
okay i ran `make docs`.

pkg/resources/resource_monitor.go
```diff
@@ -77,6 +77,13 @@ var resourceMonitorSchema = map[string]*schema.Schema{
 		Default:     false,
 		ForceNew:    true,
 	},
+	"warehouses": {
+		Type:		schema.TypeSet,
+		Optional: true,
+		Description: "A list of warehouses to apply the resource monitor to.",
+		Elem:		&schema.Schema{Type: schema.TypeString},
+		ForceNew: true,
+	},
```
alldoami, line 86: Could you also run `make fmt` to fix the spacing here.

sfc-gh-swinkler: okay done

alldoami (approved):

PR #840, opened 2022-01-31
feat: Release GH workflow

alldoami:
Following these instructions https://learn.hashicorp.com/tutorials/terraform/provider-release-publish?in=terraform/providers#add-github-action-workflow


.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
+on:
+  schedule:
+    - cron:  '0 11 * * 1'
```
edulop91, line 3: not cron -- maybe on merge to main

.github/workflows/release.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - uses: actions/checkout@v2
+        if: ${{ steps.release.outputs.release_created }}
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+        if: ${{ steps.release.outputs.release_created }}
+
+      - name: Run GoReleaser
```
edulop91, line 56: if: ${{ steps.release.outputs.release_created }} 

also needs that

edulop91 (approved):

.github/workflows/release.yml
edulop91: (you can then release mondays by merging the release pr)

PR #848, opened 2022-02-07
fix: Release tag

alldoami:
Fixing release process with updated tag information


edulop91 (approved):

.github/workflows/release.yml
```diff
@@ -46,7 +46,11 @@ jobs:
           changelog-types: ${{ steps.configure-changelog.outputs.result }}
           token: ${{ steps.generate_token.outputs.token }}
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v2'
```
edulop91, line 49:
```suggestion
      - uses: actions/checkout@v2
```
think shouldn't be there

PR #852, opened 2022-02-07
fix: Remove keybase since moving to github actions

alldoami:
Fixing: 
```
   ⨯ release failed after 651.76s error=sign failed: terraform-provider-snowflake_0.26.2_SHA256SUMS: template: tmpl:1:6: executing "tmpl" at <.Env.KEYBASE_KEY_ID>: map has no entry for key "KEYBASE_KEY_ID"
```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -33,18 +33,3 @@ archives:
 checksum:
   name_template: '{{ .ProjectName }}_{{ .Version }}_SHA256SUMS'
   algorithm: sha256
-
-signs:
-  - artifacts: checksum
-    cmd: keybase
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -33,18 +33,3 @@ archives:
 checksum:
   name_template: '{{ .ProjectName }}_{{ .Version }}_SHA256SUMS'
   algorithm: sha256
-
-signs:
```
edulop91, line 37: maybe now also rm related release stuff from the Makefile, wdyt?

edulop91 (approved):

PR #861, opened 2022-02-15
feat: Data source for list databases

jmrobles:
Add the ability to read the current databases in Snowflake (as far as the databases that the role used has the USAGE privilege on those databases).

## Test Plan

It's easy to test. Just run and retrieve the Snowflake Account's databases

[X] Acceptance test

## References

Feature Request: https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/859


alldoami:
Would it be possible to instead create a `snowflake_database` data source instead?

jmrobles:
Sure! why not? The reason to use it plural is due to avoid some confusion respect the resource name "snowflake_database", but if you consider there is no problem with that I can update the PR with the data source in singular :)

alldoami:
@jmrobles ya because we use `data` before the name I don't think there should be confusion! I think it'd be more standard to import as a singular resource than a list.

alldoami:
hmmm @jmrobles sorry I don't think I was clear 😅  Instead of outputting a list of databases, do you think it's possible to create the data source to just read one database from a specified `database_name`? If people want to use this data source and want a list of all databases, I think they'd have to specify a string list of database names in a `locals` variable and then use a `for_each` to create a data source for each one. I think this use case of getting a list of databases is a bit too niche of a situation.

jmrobles:
Ah, ok I understand you.

The motivation of this PR is because in a project that I'm working on, we need to create a user account per database. 

The idea is that automatically those user will be maintained without the need of declare in locals or vars section.

alldoami:
@jmrobles ah ok, then let's change it back to plural. Would be great if you could create a new data source with the singular form so that people can use it as a singular data source instead of a list.

examples/data-sources/snowflake_database/data-source.tf
```diff
@@ -0,0 +1,8 @@
+data "snowflake_database" "this" {
+    name = "DEMO_DB"
+}
+
+resource "snowflake_database" "backup" {
```
alldoami, line 5: I think you can remove this resource, you can just keep the data source as the example. Same for the one below

jmrobles:
I needed to write a custom TestStep check function. 
Tested against Snowflake (own account).

alldoami:
Could you run `make docs`?

jmrobles:
Sure! Done, sorry, I forgot to re-run it.

alldoami (approved):

PR #872, opened 2022-02-22
feat: Support DIRECTORY option on stage create

zapo:
This adds passing the `DIRECTORY` option of `CREATE STAGE` (as `directory`) in the stage resource.
This is only supported on create as the ALTER capabilities don't really align. As a result this PR suggests to ForceNew on `directory` change.

## Test Plan
This is currently not tested using acceptance tests

* [ ] acceptance tests


zapo:
I just realized I forgot to add support in ReadStage(). Let me add it.

alldoami (approved):

Issue #835, opened 2022-01-27
snowflake terraform resource not escaping backslashes properly upon apply


Pete-rePete:
## **Provider Version**: 
0.25.35 (current latest)

## **Terraform Version** : 
1.1.3

## **Describe the bug**
We are seeing inconsistent character escaping when performing `terraform import` versus `terraform apply`. Quick example to reproduce the bug:

1. Make sure logging level is INFO. `export TF_LOG=INFO`
2. In snowflake, create a file_format resource like so. **Take special note that escaping unenclosed fields is 1 escaped backslash.**
`CREATE FILE FORMAT DB1.SCHEMA1.CSV TYPE = 'CSV' ESCAPE_UNENCLOSED_FIELD = '\\'`

3. Create 2 nearly-identical Terraform resources. 1 to mirror the remote state, and 1 we are going to create via `terraform apply`.  The only field you need to pay attention to is `escape_unenclosed_field`. 
```hcl
resource "snowflake_file_format" "db1_schema1_csv" {
  escape_unenclosed_field        = "\\"
  name                           = "CSV"
  database                       = "DB1"
  schema                         = "SCHEMA1"
  format_type                    = "CSV" 
  binary_format                  = "HEX"
  compression                    = "AUTO"
  date_format                    = "AUTO"
  empty_field_as_null            = true
  encoding                       = "UTF8"
  error_on_column_count_mismatch = true
  escape                         = "NONE"
  field_delimiter                = ","
  field_optionally_enclosed_by   = "NONE"
  null_if                        = [
    "\\N",
  ]
  record_delimiter               = <<-EOT

EOT
  skip_byte_order_mark           = true
  time_format                    = "AUTO"
  timestamp_format               = "AUTO"
  validate_utf8                  = true 
}

resource "snowflake_file_format" "db1_schema1_csv2" {
  escape_unenclosed_field        = "\\"
  name                           = "CSV2"
  database                       = "DB1"
  schema                         = "SCHEMA1"
  format_type                    = "CSV" 
  binary_format                  = "HEX"
  compression                    = "AUTO"
  date_format                    = "AUTO"
  empty_field_as_null            = true
  encoding                       = "UTF8"
  error_on_column_count_mismatch = true
  escape                         = "NONE"
  field_delimiter                = ","
  field_optionally_enclosed_by   = "NONE"
  null_if                        = [
    "\\N",
  ]
  record_delimiter               = <<-EOT

EOT
[...]

PR #877, opened 2022-02-27
fix: escape string escape_unenclosed_field(#835)

ugmuka:
I faced the same problem as issue #835.

In `file_format.go`, escape optional value `escape_unenclosed_field`.

## Test Plan

* [ ] acceptance tests

## References

* #835


pkg/resources/file_format.go
```diff
@@ -391,7 +391,7 @@ func CreateFileFormat(data *schema.ResourceData, meta interface{}) error {
 	}
 
 	if v, ok, err := getFormatTypeOption(data, formatType, "escape_unenclosed_field"); ok && err == nil {
-		builder.WithEscapeUnenclosedField(v.(string))
+		builder.WithEscapeUnenclosedField(snowflake.EscapeString(v.(string)))
```
alldoami, line 394:
```suggestion
		builder.WithEscapeUnenclosedField(EscapeString(v.(string)))
```

alldoami (approved):

PR #878, opened 2022-02-28
chore: Upgarde all dependencies to latest

### LINKED ISSUE #866: Vulnerability issue in AWS S3 Crypto SDK for golang
I believe there is a vulnerability issue with the AWS S3 Crypto SDK for golang.
Is it possible to address this issue ?

References:
https://github.com/advisories/GHSA-76wf-9vgp-pj7w
https://github.com/google/security-research/security/advisories/GHSA-76wf-9vgp-pj7w

edulop91:
Updating all dependencies to latest available. Fixes #866.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


jacobrheath:
Go 1.17 keeps references to the dependency graph for Go 1.16 as it made the breaking change to how `go mod tidy` worked. Essentially, if the `go` directive in our go.mod file is "1.17", `go mod tidy` changes its behavior. I ran the tests and it doesn't seem like we have any references to the specific behavior of go version 1.16, so I have changed our `go mod tidy` commands to use `go mod tidy -compat=1.17` instead to remove the references. I think the security scanner simply looked at the go.sum file and associated the CVEs with the versions, but it doesn't seem like we actually use them.

edulop91 (commented):
lgtm

jacobrheath:
I've also incorporated an exclude directive because there was a breaking change in one of the terraform-exec packages that no longer exists. The exclude directive should prevent us from accidentally upgrading to that version and breaking things. From what I gathered it was still a requirement by another dependency of Hashicorp so until they update, we will have to keep that exclude in there.

PR #890, opened 2022-03-07
feat: add 'snowflake_role' data source

z4f1r0v:
This PR introduces a `snowflake_role` data source. With this functionality, one will be able to have access to roles that are created on Snowflake outside of Terraform's state.

I acknowledge that the quality here may not be up to par with the standard for this repository. Kindly, point out how I can up the game here and I will do my best. 

## Test Plan
- An acceptance test has been added

## References
This PR is an attempt to address [this](https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/883) issue.


z4f1r0v:
@alldoami, I tried running the suggested command but get nothing generated. `check-docs` gives back 0 as output. I added an example of the new data source but that doesn't help. Looking at the source code of other data sources and commands, I'm not sure what I'm missing in order to force the `tfplugindocs` to generate the necessary docs page. Any suggestions?

edulop91:
think it needs to be added [here](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/pkg/provider/provider.go#L213)

z4f1r0v:
Thanks, @edulop91. That did it!

edulop91 (approved):

pkg/datasources/role.go
```diff
@@ -0,0 +1,55 @@
...
+	err = d.Set("name", role.Name.String)
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
+		return err
+	}
+
+	return err
```
edulop91, line 54:
```suggestion
	return nil
```
is sufficient here!

alldoami:
I believe you'll need to address the error from the test ran above

z4f1r0v:
@alldoami, as I see it the problem in the tests is:
```
=== CONT  TestRole
[226](https://github.com/chanzuckerberg/terraform-provider-snowflake/runs/5516781232?check_suite_focus=true#step:9:226)
    testing_new_config.go:111: no "id" found in attributes
[227](https://github.com/chanzuckerberg/terraform-provider-snowflake/runs/5516781232?check_suite_focus=true#step:9:227)
    testing_new.go:53: no "id" found in attributes
[228](https://github.com/chanzuckerberg/terraform-provider-snowflake/runs/5516781232?check_suite_focus=true#step:9:228)
--- FAIL: TestRole (3.17s)
```
This supports my suspicion around how IDs are found when none are provided. I have been looking at the code and adding an `Importer` to the `Resource` seems like the missing link. I'm not entirely sure how that is done but it seems that this is the way to bring resources not managed by Terraform into its scope. 

Can somebody enlighten me on this matter? I see that the original author @ryanking is no longer supporting these repos so I don't know who is the go-to person.

z4f1r0v:
Same errors as before. It seems that the `ImportStatePassthroughContext` is not enough to get the ID. I'll continue exploring the code. Meanwhile, if anyone can provide some insights, I'd really appreciate it.

ChrisIsidora:
@z4f1r0v From what i can see you are not setting an id in ReadRole. I think you do need to set an id if i look at the existing datasources

pkg/datasources/role.go
```diff
@@ -0,0 +1,54 @@
...
+		return err
+	}
+
+	err = d.Set("name", role.Name.String)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
gary-beautypie, line 54:
```suggestion
func ReadRole(d *schema.ResourceData, meta interface{}) error {
	db := meta.(*sql.DB)
	id := d.Get("name").(string)

	row := snowflake.QueryRow(db, fmt.Sprintf("SHOW ROLES LIKE '%s'", id))
	role, err := snowflake.ScanRole(row)

	if err == sql.ErrNoRows {
		log.Printf("[DEBUG] role (%s) not found", d.Id())
		d.SetId("")
		return nil
	}
	if err != nil {
		return err
	}

    d.SetId(role.Name.String)
	err = d.Set("name", role.Name.String)
	if err != nil {
		return err
	}

	return nil
}
```

PR #906, opened 2022-03-10
fix: update snowflake identifier validation condition

wli-brex:
Has been blocked by the same issues as others reported: 
[issue-846](https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/846)
[issue-862](https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/862)

For context, Snowflake object identifier can use special characters like `-` as long as it's [double quoted](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#double-quoted-identifiers). However currently the validation rule only allow characters used in [unquoted Snowflake object identifier](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#unquoted-identifiers).

As mentioned in the issue tickets, currently creating a role with `-` is allowed by the provider, but granting role with `-` to user or other roles is blocked by this validation. So I believe we should enable the dash character to provide consistent experience as in Snowflake console.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*

### DIFF
--- a/pkg/snowflake/validation.go
+++ b/pkg/snowflake/validation.go
@@ -42,6 +42,7 @@ func isIdentifierRune(r rune) bool {
 
 func isInitialIdentifierRune(r rune) bool {
 	return (r == '_' ||
+		r == '-' ||
 		(r >= 'A' && r <= 'Z') ||
 		(r >= 'a' && r <= 'z'))
 }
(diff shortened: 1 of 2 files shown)


wli-brex:
@alldoami May I ask what else do I need to do to release this change? I'm a first time committer, not sure if I need to setup a new releaser follow the README, or will the goreleaser handles that automatically?

alldoami (approved):

Issue #460, opened 2021-02-19
Support for grant ownership on role


ericmjalbert:
**Is your feature request related to a problem? Please describe.**

I'm setting up my Snowflake account with a [Vault Database Engine](https://github.com/Snowflake-Labs/sfguide-vault-snowflakepasswords) to manage the user passwords for applications. As part of this solution I need to `grant ownership on role ROLENAME to role VAULTROLE`.

I don't think there is a way to currently do this with the terraform snowflake provider.

**Describe the solution you'd like**

I'd like to see a new resource that lets you define role ownership to a specific user. Maybe something like a new resource named `snowflake_ownership_grant` that covers the functionality of https://docs.snowflake.com/en/sql-reference/sql/grant-ownership.html.

It might looks something like:
```
resource snowflake_role_ownership_grant grant {
  on_role = "ROLENAME"
  to_role = "VAULTROLE"

  current_grants = "COPY|REVOKE"
}
```

**Describe alternatives you've considered**

I haven't thought of any alternatives yet. I think I'll be able to solve this using unrelated automation scripts, but those exist outside the terraform ecosystem.

**Additional context**

NA

yohei1126:
Have you try `role_grants`? https://registry.terraform.io/providers/chanzuckerberg/snowflake/latest/docs/resources/role_grants

ericmjalbert:
I don't think that `role_grants` can do this since it only does the [`GRANT ROLE ...`](https://docs.snowflake.com/en/sql-reference/sql/grant-role.html) type sql (see [source code](https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/565c8c6991f7959f207645d97837de7bed1b9d0c/pkg/snowflake/role_grant.go#L36)) . This feature request is specifically for the granting of `OWNERSHIP`.

malonsocasas:
I have a similar issue here. 

An alternative proposal, although I don't know if this completely fixes ericmjalbert's problem: would it be possible to specify parent role when creating a new role with `snowflake_role` 

Something like :
```
resource snowflake_role role {
  name    = "role1"
  comment = "A role."
  parent_role = "parentrole1"
}
```

PChambino:
My use case for this is that I want role A (created by terraform) to be own by role B so that role B can grant role A to users without having to give MANAGE GRANTS at the account level to role B.

[...]

PR #916, opened 2022-03-12
snowflake_ownership_grant

aidanmelen:
attempting to add `snowflake_ownership_grant`

Adding support for granting ownership to a role. Initial feature only supports role to role.

More in the issue: #460

## Test Plan
- 

## References
- https://docs.snowflake.com/en/sql-reference/sql/grant-ownership.html


aidanmelen:
The provider seems to be written to best support ownership via the `snowflake_grant_roles` resource. closing for now

PR #939, opened 2022-03-18
feat: support host option to pass down to driver

Dirrk:
Allows the provider to accept a host value that is passed down to the snowflakego driver.  In addition, if Host is set it unsets the region value due to the way the driver attempts to hack the region back into the host [incorrectly](https://github.com/snowflakedb/gosnowflake/blob/678701a9ecf52cbfb5ca425daf540ed7e70a5125/dsn.go#L381-L383) for privatelink.

## Test Plan

* [ ] acceptance tests

## References

* We built this module into our terraform container and have tested this to validate that it is working with our snowflake setup which only allows privatelink connections. 
* Securing snowflake to only privatelink connections requires blocking public ips via the network security policy and using a [custom privatelink domain](https://docs.snowflake.com/en/user-guide/admin-security-privatelink.html#creating-additional-cname-records) ie: `account.region.privatelink.snowflakecomputing.com`
* This should solve https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/926 as I worked directly with scott on this
* Should essentially solve https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/447 while being backward compatible


alldoami:
Should we not close #926 if this fixes it? Is this a replacement PR for it?

alldoami:
Could you run `make docs`?

alldoami (approved):

aidanmelen:
forgot to update the provider docs 😉

- https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/959

PR #941, opened 2022-03-21
fix: Legacy role grantID to work with new grant functionality

### LINKED ISSUE #922: Error: 5 or 6 fields allowed (using snowflake_role_grants)
**Provider Version**
0.28.6

**Terraform Version**
1.1.7

**Describe the bug**

Error: 5 or 6 fields allowed (using snowflake_role_grants)

**Expected behavior**

In version 0.27.0 it works normally

**Code samples and commands**

```
resource "snowflake_role_grants" "dadosfera" {
  role_name = snowflake_role.dadosfera.name

  users = [
    snowflake_user.dadosfera.name
  ]
}
```

**Additional context**

image

daniepett:
These changes will allow legacy grantID for role grants (Just role name), to work with the new grantID structure/functionality.

Implementing these changes requires less stringent testing on the IDs, but gives flexibility that allows legacy roleGrantIDs to be using the same function, therefore limiting code duplication.

* [ ] acceptance tests
* [x] unit tests

## References

* Closes #922


ChrisIsidora:
@daniepett Does this change comes with the caveat that grants applied OUTSIDE Terraform will not be revoked. If so is there a reason why this behavior is changed? This would mean I can never achieve a desired state with this new behavior

daniepett:
@ChrisIsidora The behaviour was made to allow similar grants to be created in different resources. Resource A won't know if the grant is being created in Resource B, and therefore won't revoke it.

daniepett:
I do agree, but for implementation of this functionality it was a trade off. It was clearly stated in the original PR that this would be the case, I left it up to maintainers to make that call. It can be reverted if need be.

ChrisIsidora:
@alldoami @edulop91 @robbruce What do you think?

daniepett:
@ChrisIsidora I've pushed some changes as an example of a feature flag implementation. Is there a better way to implement this than on a resource level?

ChrisIsidora:
> @ChrisIsidora I've pushed some changes as an example of a feature flag implementation. Is there a better way to implement this than on a resource level?

@daniepett I think it's a good implementation. Normally this is implemented a different way but in this case I think it's fine. Usually for these partial/multiple cases you see that some providers chose to introduce a new resources with only that specific functionality. A fair example is for instance azurerm_keyvault and azurerm_keyvault_access_policy. Access policy is also defined on azurerm_keyvault.

Thanks for the changes :-)

daniepett:
@ChrisIsidora It will potentially force replacement of a few grant resources as some of the grant resources don't have update functions and the config for them has changed. Should hopefully be a compromise everyone can live with.

Thanks for the feedback, do appreciate that :)

ChrisIsidora:
@alldoami I think with the new changes and the feature flag set by default to false it should have the same way of working as before. So i think now it should be ok. I can so a test only my side when you release.

robbruce:
@daniepett @alldoami @ChrisIsidora - Can a StateMigration function be added? With different schema versions? So that the the provider know's how to migrate the old id format to the new id format? I'm not sure if can be used with the id, I've never done it myself.

https://www.terraform.io/plugin/sdkv2/resources/state-migration

ChrisIsidora:
@daniepett Can you fix the tests?

daniepett:
@ChrisIsidora @alldoami I've ignored the feature flag for imports, won't affect functionality. It can't be read from current state, so better to use the default false once it's imported.

ChrisIsidora:
@alldoami Can we release a new version? Thanks in advance

alldoami (approved):

Issue #959, opened 2022-04-05
Provider docs haven't been updated with the new `host` arg


aidanmelen:
**Provider Version**

v0.30.0

**Describe the bug**

The provider docs haven't been updated with the new [`host`](https://registry.terraform.io/providers/chanzuckerberg/snowflake/latest/docs) argument

yohei1126:
@aidanmelen Would you provide a PR?

aidanmelen:
> @aidanmelen Would you provide a PR?

If it is an easy as running `make docs` then sure. @Dirrk did you forget to build the docs or is something broke with that workflow?

Dirrk:
@aidanmelen I ran `make docs` and pushed up https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/939/commits/2b1d9eb06ae0bcc1a2017e03b75d8f6a43ee8a23

aidanmelen:
> @aidanmelen I ran `make docs` and pushed up [2b1d9eb](https://github.com/chanzuckerberg/terraform-provider-snowflake/commit/2b1d9eb06ae0bcc1a2017e03b75d8f6a43ee8a23)

You definitely did. I also noticed that you added Description to the `provider.go` schema. That was a nice touch on your part.

Dirrk:
I have another PR coming exposing the port and protocol options to the driver which we have tested locally for almost 2 weeks now.  I will make sure the docs get updated in that PR as well before submitting.

aidanmelen:
@alldoami it looks like the `docs/index.md` file was generated correctly, but the terraform registry is showing the same docs as version `v0.29.0`

Link: https://undefined/providers/chanzuckerberg/snowflake/latest/docs

PR #960, opened 2022-04-06
docs: Update host provider docs

aidanmelen:
Added missing `host` provider documentation. But then realized that the provider.go schema was missing descriptions that `terraform-plugin-docs` is capable of auto generating for us. So I went ahead and fixed that too.

## References

* #959

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -45,25 +45,25 @@ provider "snowflake" {
 
 ### Required
 
-- **account** (String)
-- **username** (String)
+- **account** (String) The name of the Snowflake account. Can also come from the `SNOWFLAKE_ACCOUNT` environment variable.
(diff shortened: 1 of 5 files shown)


aidanmelen:
@alldoami this should improve the provider doc generation for future builds.

aidanmelen:
@alldoami any clue why these checks failed?

Screen Shot 2022-04-06 at 12 30 28 PM

aidanmelen:
@alldoami it looks like a there is a breaking change to the `snowflake_procedures` data resource such that the procedures count is return the wrong value. I believe this is unrelated to these incoming doc changes.

alldoami:
oh so weird! ya definitely unrelated

alldoami:
As long as `check-docs` runs clean, i'll just merge it.

PR #1007, opened 2022-05-17
feat: Adding Database Replication

sfc-gh-jalin:
Adding database replication namely enabling accounts to replicate a database from a primary source

- Added new parameters to the database resource: `from_replica_config` and `replication_configuration`. 
- `from_replica_config` allows one to input the three parameters that are required for database replication. The format is `'organization'.'account'.'primaryDbName'`
- `replication_configuration` allows one to add accounts that are allowed to create a replica database from
## Test Plan

* [ ] acceptance tests

* [x] added unit tests and are passing
* [x] able to create replication databases in personal account

## References

https://docs.snowflake.com/en/user-guide/database-replication-config.html
*


pkg/resources/database.go
```diff
@@ -82,9 +128,37 @@ func CreateDatabase(d *schema.ResourceData, meta interface{}) error {
...
+	if _, ok := d.GetOk("from_replica_config"); ok {
+		return createDatabaseFromReplicaConfig(d, meta)
+	}
+
+	// If set, verify parameters are valid and attempt to enable replication
+	if v, ok := d.GetOk("replication_configuration"); ok {
+		replicationConfiguration := v.([]interface{})[0].(map[string]interface{})
+		ignoreEditionCheck := replicationConfiguration["ignore_edition_check"].(bool)
+
+		if ignoreEditionCheck == false {
```
cziie, line 140: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>S1002: should omit comparison to bool constant, can be simplified to `!ignoreEditionCheck` (gosimple)

docs/resources/database.md
```diff
@@ -38,10 +38,34 @@ resource "snowflake_database" "test2" {
 - **data_retention_time_in_days** (Number)
 - **from_database** (String) Specify a database to create a clone from.
 - **from_replica** (String) Specify a fully-qualified path to a database to create a replica from.
+- **from_replica_config** (Block List, Max: 1) Specifies the fully qualified path to a database for replication purposes. (see [below for nested schema](#nestedblock--from_replica_config))
```
scottwinkler, line 41: so how is this different than from_replica?

pkg/resources/database.go
```diff
@@ -142,6 +216,22 @@ func createDatabaseFromReplica(d *schema.ResourceData, meta interface{}) error {
 	return ReadDatabase(d, meta)
 }
 
+func createDatabaseFromReplicaConfig(d *schema.ResourceData, meta interface{}) error {
```
scottwinkler, line 219: this looks exactly like createDatabaseFromReplica(). if all you are doing is concatenating the organization_name account_name and primary_db_name, i dont know that this adds much value compared to using the from_replica argument. i am wondering if this can instead be an output of the database resource that has replication_configuration enabled. e.g. "database_identifier". Another option is create a datasource to construct identifiers based on input arguments.

pkg/resources/database.go
```diff
@@ -48,6 +49,51 @@ var databaseSchema = map[string]*schema.Schema{
...
 		ConflictsWith: []string{"from_share", "from_database"},
 	},
+	"from_replica_config": {
+		Type:          schema.TypeList,
+		Description:   "Specifies the fully qualified path to a database for replication purposes.",
+		Optional:      true,
+		ForceNew:      true,
+		MaxItems:      1,
+		ConflictsWith: []string{"from_share", "from_database", "from_replica"},
+		Elem: &schema.Resource{
```
scottwinkler, line 59: we should add validation here

scottwinkler (changes requested):

docs/resources/database.md
sfc-gh-jalin: It's not different, I wanted to add this so it would be more clear for users to understand what the fully qualified path is and removes the need for repeatedly escaping the `"` character. It makes more sense to me than the current implementation but I also wanted to allow backwards compatibility so I did not remove the other one but added a `conflicts_with`

alldoami:
@sfc-gh-jalin looks like you need to run `make docs` and fix a test related to `TestAcc_ApiIntegration`.

alldoami:
@sfc-gh-jalin you'll need to run `make docs`

sfc-gh-swinkler (approved):
LGTM

PR #1008, opened 2022-05-18
fix: Update go dependencies

alldoami:
ran: `go get -u ./...` at the root to update go dependencies.

Also ran `go get github.com/hashicorp/terraform-exec@v0.15.0`

because this https://github.com/hashicorp/terraform-exec/blob/v0.16.0/tfinstall/tfinstall.go does not exist. Needed to downgrade to `v0.15.0`


alldoami:
This should fix failing tests here: https://github.com/chanzuckerberg/terraform-provider-snowflake/pull/1007

sfc-gh-jalin (approved):

sfc-gh-jalin (approved):

Issue #982, opened 2022-04-21
snowflake_procedure_grant documentation is wrong


jrobison-sb:
**Provider Version**

```
$ terraform version
Terraform v0.14.11
+ provider registry.terraform.io/chanzuckerberg/snowflake v0.25.36
```

**Describe the bug**

The docs for `snowflake_procedure_grant` are wrong:
https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/examples/resources/snowflake_procedure_grant/resource.tf#L6-L15

```
  arguments   = [
    {
      "name": "a",
      "type": "array"
    },
    {
      "name": "b",
      "type": "string"
    }
  ]
```

This results in an error:

```
Error: Unsupported argument

  on ../../modules/aws/application/snowflake.tf line 1147, in resource "snowflake_procedure_grant" "foo":
1147:   arguments = [

An argument named "arguments" is not expected here. Did you mean to define a
block of type "arguments"?
```

**Expected behavior**

It should instead show the correct data type, which is a block, not a list:
```
  arguments {
    name = "a"
    type = "array"
  }

  arguments {
    name = "b"
    type = "string"
  }

```

coryveilleux:
This applies to `snowflake_function_grant` as well:

https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/main/examples/resources/snowflake_function_grant/resource.tf#L6-L15

sfc-gh-jalin:
Issue has been resolved by above PR, closing issue

PR #1010, opened 2022-05-19
feat: Support for selecting language in snowflake_procedure

emancu:
Implements https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/948

The previous behaviour was always setting `JAVASCRIPT` as language.
The implementation in this PR breaks the "backwards compatibility" with that rule to honour the Snowflake default: `SQL`.

In addition, documentation for `procedure_grant` and `function_grant` is updated too.

## References

* https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/948
* https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/982


emancu:
Do I need to do something in particular to get the approval to run the workflows?

alldoami:
Hi! Thank you for being patient! We were fixing some go package updates, we'll take a look at this soon! In the meantime, can you fix the conflicts?

emancu:
@alldoami Done, now the docs should be up to date too.

However, I'm struggling here. The definition of `arguments` for `procedure` is the same as `procedure_grant` and `function_grant`, however, the documentation generated is different!

In fact, the docs generated for `procedure_grant` and `function_grant` is incorrect (https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/982) but I can't figure out how to fix it 😓

emancu:
> Our docs are generated from the resource definition schemas! So for `procedure_grant`, you'll want to modify the descriptions here: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/procedure_grant.go, and for `function_grant`, you'll want to modify the descriptions here: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/function_grant.go. Does that make sense?

Yes and no 😅 . The descriptions get updated as expected, but the problem is with the sample code generated.

Running `make checks-docs` generates the following example for `procedure_grant` and `function_grant`

```
arguments   = [
    {
      "name": "a",
      "type": "array"
    },
    {
      "name": "b",
      "type": "string"
    }
  ]
```

However, for `procedure` it generates the following (which in fact it is the correct)

```
  arguments {
    name = "a"
    type = "array"
  }

  arguments {
    name = "b"
    type = "string"
  }
```


ℹ️ The three of them share the same* definition for arguments, which it is

```go
	"arguments": {
		Type: schema.TypeList,
		Elem: &schema.Resource{
				"name": {
					Type:     schema.TypeString,
					Required: true,
					Description:      "The argument name",
				},
				"type": {
					Type:     schema.TypeString,
					Required: true,
					Description:      "The argument type",
				},
			},
		},
		Optional:    true,
		Description: "List of the arguments for the procedure",
		ForceNew:    true,
	},
```
> (*) procedure uses `DiffSuppressFunc: DiffTypes` but it shouldn't affect.



Am I missing something obvious? 🤔

alldoami:
@emancu examples are written here manually :) https://github.com/Snowflake-Labs/terraform-provider-snowflake/tree/main/examples/resources

emancu:
@alldoami Thank you so much!
Docs are updated now 🎉

sfc-gh-swinkler (approved):

PR #1031, opened 2022-05-30
fix: handling not exist gracefully

israel:
This is my first PR here so I have made changes to only some of the resources. Once approved I will do the same for the rest. Thanks

Changed handling of not exists scenarios for read resource functions for User, API Integration and External Function resources

## Test Plan
Updated user_test to reflect the change.


sfc-gh-swinkler:
Hello @israel, thank you for your PR!

I think this is good, just I would ask for you to create a new file called `errors.go` at `pkg/snowflake/errors.go` to declare this error as a constant. For example, take a look at how this provider did the same thing you are doing:
https://github.com/hashicorp/terraform-provider-tfe/blob/main/tfe/resource_tfe_team.go#L155-L162

and then they declared that error in the errors.go file:
https://github.com/hashicorp/go-tfe/blob/main/errors.go#L13

This is not something we are currently doing, but its a pattern I would like to start abiding by. If you could please do this I would be much obliged.

israel:
Hi @sfc-gh-swinkler 
Apologies for the delayed response, I have not seen the notification for some reason...
I have moved the generic error from external_function.go to a common errors file as you requested.
Thanks,
Israel

pkg/resources/api_integration.go
```diff
@@ -150,7 +150,13 @@ func ReadAPIIntegration(d *schema.ResourceData, meta interface{}) error {
 
 	s, err := snowflake.ScanApiIntegration(row)
 	if err != nil {
-		return fmt.Errorf("Could not show api integration: %w", err)
+		// If no such resource exists, it is not an error but rather not exist
+		if err.Error() == "sql: no rows in result set" {
```
sfc-gh-swinkler, line 154: this should be ErrNoRowInRS i think

sfc-gh-swinkler (changes requested):

pkg/resources/api_integration.go
israel: done

sfc-gh-swinkler (approved):

sfc-gh-swinkler:
Thank you for your contribution! @israel

PR #1047, opened 2022-06-08
fix: Correctly read INSERT_ONLY mode for streams

adamantike:
This change removes the `append_only`/`insert_only` fields from the Stream row struct, as they are not returned neither by `SHOW STREAMS` nor `DESCRIBE STREAM`.

To find if an existing stream is `INSERT_ONLY`, it now checks the `mode` field, as it has been already done for the `APPEND_ONLY` mode.

## Test Plan
* [x] acceptance tests
* [x] unit tests


sfc-gh-swinkler:
this looks good.

Issue #1054, opened 2022-06-10
Snowflake role with dot (.) in its name can be created but not granted using terraform


poojagujar93:
**Provider Version**

 v0.35.0.

**Terraform Version**

v1.1.8

**Describe the bug**

When using terraform to grant a role to a role/user, the role (snowflake_role) can be created with a dash in the name, but a grant to a role (snowflake_role_grants) with a dot(.) in its name fails with Error: '.' is not a valid identifier character.

**Expected behavior**

A grant should be able to be created for any roles that have been successfully created with snowflake_role.

**Code samples and commands**
# Create Role
resource "snowflake_role" "roles" {
  provider = snowflake.securityadmin
  name     = "DU.DPLTF.DEV.READERS"
  comment  = "this is head of all Dev Readers role"
}

# Grant roles
resource "snowflake_role_grants" "roles" {
  provider   = snowflake.securityadmin
  role_name  = "DU.DPLTF.DEV.READERS"
  roles      = ["sp-terraform"]
  depends_on = [snowflake_role.roles]
}

After running plan:
│ Error: '.' is not a valid identifier character.
│
│   with snowflake_role_grants.grants,
│   on main.tf line 5, in resource "snowflake_role_grants" "grants":
│    5:   role_name   =  "DU.DPLTF.DEV.READERS"
│
╵
time=2022-06-10T16:18:40+05:30 level=error msg=1 error occurred:
        * exit status 1

**Additional context**

If role_name does not have .(dot), then it works. But our projects requirement is to have .(dot) in the role name.

sfc-gh-jcieslak:
Hey 👋
Closing the issue as it's referring to the deprecated resource and we're not supporting them. Please, try to use the latest provider version with the non-deprecated resources (e.g. snowflake_grant_privileges_to_account_role). If the issue persists on the non-deprecated resource (or the feature is present in Snowflake, but not in the provider), please create a new GitHub issue. The [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/docs/technical-documentation/resource_migration.md) may help during migration to the latest resources. Here's a list of the latest, non-deprecated, grant resources:

Resources
[snowflake_grant_privileges_to_database_role](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_privileges_to_database_role)
[...]

PR #1059, opened 2022-06-14
fix: allow custom characters to be ignored from validation

sfc-gh-jalin:
Adding fix for issue: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1054

Created a new struct parameter for adding characters to exclude from the validation check.

role_ownership_grants is now able to take in periods or "." in the role_name parameter

## Test Plan

* [ ] acceptance tests
tested locally and was able to resolve the error for ` Error: '.' is not a valid identifier character.`

## References

https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1054
*


sfc-gh-swinkler (approved):
LGTM

PR #1063, opened 2022-06-15
feat: add python language support for functions

sfc-gh-kmaurya:
Added python language support for snowflake_function resource.

## Test Plan
This PR need to be tested for acceptance tests

## References

* [Create Function](https://docs.snowflake.com/en/sql-reference/sql/create-function.html#create-function)
* [Creating Python UDFs](https://docs.snowflake.com/en/developer-guide/udf/python/udf-python-creating.html#creating-python-udfs)


sfc-gh-afedorov (commented):
I don't think we should define the warehouse that's used to create the function on the function

docs/resources/function.md
```diff
@@ -21,18 +21,21 @@ description: |-
...
 
 - `arguments` (Block List) List of the arguments for the function (see [below for nested schema](#nestedblock--arguments))
 - `comment` (String) Specifies a comment for the function.
-- `handler` (String) the handler method for Java function.
-- `imports` (List of String) jar files to import for Java function.
+- `handler` (String) the handler method for Java and python function.
+- `imports` (List of String) jar files to import for Java function or for importing python files.
 - `language` (String) The language of the statement
 - `null_input_behavior` (String) Specifies the behavior of the function when called with null inputs.
+- `packages` (List of String) For java the value should be of the form package_name:version_number, where package_name is snowflake_domain:package and for python use it as packages = ('numpy','pandas','xgboost==1.5.0').
```
sfc-gh-swinkler, line 34: List of package imports to use for Java / Python functions. For Java, package imports should be of the form: package_name:version_number, where package_name is snowflake_domain:package. For Python use it should be:  ('numpy','pandas','xgboost==1.5.0').

pkg/resources/function.go
```diff
@@ -179,11 +200,38 @@ func CreateFunction(d *schema.ResourceData, meta interface{}) error {
 		builder.WithLanguage(v.(string))
 	}
 
+	// Set optionals, runtime version for python
+	if v, ok := d.GetOk("runtime_version"); ok {
+		builder.WithRuntimeVersion(v.(string))
+	}
+
+	// Set optionals, warehouse in which to create the function
```
sfc-gh-swinkler, line 208: so it looks like warehouse is getting set even for non-Python functions. Is this okay? Also i could not find in the documentation where this was supported. Can you pleaselink?

pkg/resources/function.go
```diff
@@ -102,26 +108,41 @@ var functionSchema = map[string]*schema.Schema{
 		Default:     "user-defined function",
 		Description: "Specifies a comment for the function.",
 	},
+	"runtime_version": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		ForceNew:    true,
+		Description: "runtime version for python.",
```
sfc-gh-swinkler, line 115: Required for Python functions. Specifies Python runtime version.

pkg/resources/function.go
```diff
@@ -102,26 +108,41 @@ var functionSchema = map[string]*schema.Schema{
...
+	},
+	"packages": {
+		Type: schema.TypeList,
+		Elem: &schema.Schema{
+			Type: schema.TypeString,
+		},
+		Optional:    true,
+		ForceNew:    true,
+		Description: "For java the value should be of the form package_name:version_number, where package_name is snowflake_domain:package and for python use it as packages = ('numpy','pandas','xgboost==1.5.0').",
```
sfc-gh-swinkler, line 124: these should be updated as above

pkg/resources/function.go
```diff
@@ -192,12 +240,12 @@ func CreateFunction(d *schema.ResourceData, meta interface{}) error {
...
 	}
 
-	// handler for Java
+	// handler for Java / python
 	if v, ok := d.GetOk("handler"); ok {
 		builder.WithHandler(v.(string))
 	}
 
-	// target path for Java
+	// target path for Java / python
```
sfc-gh-swinkler, line 248: Python should be capitalized

sfc-gh-swinkler (changes requested):

pkg/snowflake/function.go
```diff
@@ -140,6 +161,13 @@ func Function(db, schema, name string, argTypes []string) *FunctionBuilder {
 	}
 }
 
+func (pb *FunctionBuilder) UseWarehouse() (string, error) {
+	var q strings.Builder
+
+	q.WriteString(fmt.Sprintf("USE WAREHOUSE %v", pb.warehouse))
```
sfc-gh-afedorov, line 167: this shouldn't be set on a single resource but on the session. within tf, it should be set per instance of the provider.

sfc-gh-afedorov: cc @sfc-gh-swinkler to clarify where the warehouse goes in the other piece

sfc-gh-swinkler: @sfc-gh-afedorov it is not set on the instance of the provider, since the provider can create multiple warehouses and not one of them can be considered a default warehouse.

i don't like having to put this here, but in discussion with @sfc-gh-kmaurya this is the best place to put it.

sfc-gh-swinkler (approved):

sfc-gh-swinkler (approved):

sfc-gh-afedorov (commented):
I don't think we should release this -- setting the warehouse on the session when creating a function is not a side-effect we want to introduce when creating the object

pkg/resources/function.go
sfc-gh-afedorov: @sfc-gh-kmaurya  please don't ignore feedback from senior engineers

PR #1072, opened 2022-06-23
fix: Handling 2022_03 breaking changes

gouline:
2022_03 bundle (enabled in 6.21 release) changes the following behavior:

* Share `NAME` column in accounts using the new org naming is returned as `ORG.ACCOUNT.SHARE`, instead of the old `ACCOUNT.SHARE`
* Task `PREDECESSORS` column is now always returned as a JSON array (even when it's empty)
* File format `VALIDATE_UTF8` option is obsoleted, attempting to set it to false now fails

Note that the existing task implementation only handles one task root and this change continues that. While you can only specify one `AFTER` in `CREATE TASK`, you can add more with `ALTER TASK ... ADD AFTER`. Supporting this would require a breaking change to make the `after` parameter a list and handle state for multiple roots, this should be done in a separate contribution.

## Test Plan

* [x] acceptance tests

## References

* [6.21 Release Notes](https://community.snowflake.com/s/article/6-21-Behavior-Change-Release-Notes-June-22-23-2022)
* [Pending Behavior Change Log](https://community.snowflake.com/s/article/Pending-Behavior-Change-Log)
* [SHOW TASKS Command & TASK_DEPENDENTS Function: PREDECESSORS / PREDECESSOR Column](https://community.snowflake.com/s/article/SHOW-TASKS-Command-and-TASK-DEPENDENTS-Function-PREDECESSORS-PREDECESSOR-Column)
* [SHOW SHARES Command & Data Sharing UI: Changes to Output](https://community.snowflake.com/s/article/show-shares-command-ui-data-sharing-changes)
* [VALIDATE_UTF8 File Format Option](https://community.snowflake.com/s/article/VALIDATE-UTF8-File-Format-Option-Obsoleted)

### DIFF
--- a/pkg/datasources/file_formats_acceptance_test.go
+++ b/pkg/datasources/file_formats_acceptance_test.go
@@ -65,6 +65,7 @@ func fileFormats(databaseName string, schemaName string, fileFormatName string) 
 		null_if = ["NULL"]
 		error_on_column_count_mismatch = true
 		replace_invalid_characters = true
+		validate_utf8 = true
 		empty_field_as_null = false 
 		skip_byte_order_mark = false
 		encoding = "UTF-16"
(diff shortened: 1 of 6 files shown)


gouline:
Tests should be fixed now, ready for another run 🤞

sfc-gh-swinkler (approved):

ejaoua:
@sfc-gh-swinkler do you know when the release will be available ? it's quite urgent for us regarding the synchronization of share names between terraform and snowflake

Issue #1076, opened 2022-06-24
Plugin did not respond: error: The terraform-provider-snowflake_v0.33.1 plugin crashed!


kbroughton:
**Provider Version**
terraform-provider-snowflake_v0.33.1

**Terraform Version**
$ terraform -v
Terraform v1.0.11
on darwin_amd64
+ provider registry.terraform.io/chanzuckerberg/snowflake v0.33.1
+ provider registry.terraform.io/hashicorp/tls v3.3.0

** Remote
Using TFC as remote

**Describe the bug**
```
Error: The terraform-provider-snowflake_v0.33.1 plugin crashed!

This is always indicative of a bug within the plugin. It would be immensely
helpful if you could report the crash with the plugin's maintainers so that it
can be fixed. The output above should help diagnose the issue.

╷
│ Error: Plugin did not respond
│
│   with snowflake_role_grants.ur_to_br_grants["Example:User"],
│   on main.tf line 317, in resource "snowflake_role_grants" "ur_to_br_grants":
│  317: resource snowflake_role_grants "ur_to_br_grants" {
│
│ The plugin encountered an error, and failed to respond to the
│ plugin.(*GRPCProvider).ReadResource call. The plugin logs may contain more
│ details
```

It appears there is a provider-side issue.

**Expected behavior**

* terraform plan should run without errors
* better error messaging. If this is related to the provider upgrade (which also forces tf version upgrade)
   the error should direct the user to upgrade.

**Code samples and commands**

terraform plan

```
terraform {
  backend "remote" {
    organization = "example-security"

    workspaces {
      name = "security_snowflake"
    }
  }
}


terraform {
  required_providers {
    snowflake = {
      source  = "chanzuckerberg/snowflake"
      version = "0.33.1"
    }
  }
}

provider "snowflake" {
  alias           = "sys_admin"
  role            = "SYSADMIN"
  account         = "PFXXXXXX"
  host            = "example.snowflakecomputing.com"
  username        = "EXAMPLE_SNOW"
  private_key     = var.EXAMPLE_RSA_PRIVATE_KEY

}


data "snowflake_tables" "panther_logs" {
    provider    = snowflake.sys_admin
    database = "PANTHER_LOGS"
    schema   = "PUBLIC"
}

```

**Additional context**
I ran this about 2 weeks ago. Made very minimal changes to main.tf.
Now I cannot do anything with the snowflake provider.

**Stack Trace**

```
[...]

PR #1077, opened 2022-06-27
fix: update README with provider version -> 0.36.0

sfc-gh-dszmolka:
This is because of issue#1076 . If someone tries to actually use the version from the example, they might face the locale issue which has already been fixed with https://github.com/snowflakedb/gosnowflake/pull/571 and released in gosnowflake 1.6.9, which is bundled with the 0.36.0 version of the provider.

## Test Plan
Don't think this minor doc change needs tests?

## References
* https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1076


sfc-gh-dszmolka:
can i please ask you @sfc-gh-swinkler for a review?

PR #1090, opened 2022-06-30
fix: Add org name to database from_share

gouline:
After 2022_03 bundle, creating a database from share now consists of `<org>.<account>.<share>`. This PR proposes adding "org" as a component of the "from_share" map.

Having said that, I can see "from_replica" uses a fully-qualified name string instead of the map approach. What do you think would be best for consistency?

## Test Plan

* [x] acceptance tests

## References

* #1072


sfc-gh-swinkler:
i like using fully qualified names better. Also i think adding a data source that can create fully qualified names using maps  (or equivalent) would be a good idea.

gouline:
My gripe with the current implementation of fully-qualified names is that you're expected to inject only the _middle_ quotes, e.g. `from_share = "org\".\"account\".\"share"` for SQL `FROM SHARE "org"."account"."share"`, instead of allowing just `from_share = "org.account.share"` for simple names and `from_share = "\"org\".\"account\".\"share\""` for ones with special characters. What do you think?

Either way, if fully-qualified names are the preference, how can we make `from_share` consistent with `from_replica`? Breaking change to migrate it from a map to a fully-qualified name string?

sfc-gh-swinkler:
> Either way, if fully-qualified names are the preference, how can we make from_share consistent with from_replica? Breaking change to migrate it from a map to a fully-qualified name string?

since the from_replica was already done this way i am okay keeping it consistent. just know that in the future it will likely be a breaking change to change it to fully qualified name strings. i agree with the gripe of middle quotes, which is why i would favor a data source to create these.

gouline:
Sounds reasonable. In that case, I'm closing this pull request and will think about the data source for fully-qualified names and migrating `from_share` to that.

PR #1092, opened 2022-06-30
fix: warehouse import when auto_suspend is set to null

ezeql:
Fixes 
```
sql: Scan error on column index 8, name "auto_suspend": converting NULL to int64 is unsupported
```
by using `sql.NullInt64` instead of `int64`
 

- [x] unit tests


sfc-gh-swinkler (approved):
LGTM

sfc-gh-swinkler:
Thanks @ezeql for your contribution!

PR #1114, opened 2022-07-10
feat: Adding support for orgaznization and account names when creating db f…

### LINKED ISSUE #1107: Database from share using organization.account as a provider

Following new snowflake [release](https://community.snowflake.com/s/article/show-shares-command-ui-data-sharing-changes), when creating a database from share instead of account locator you can specify source as `ORGANIZATION.ACCOUNT.SHARE`. This is not supported by current (0.37.1) version of the provider. Adding `ORGANIZATION.ACCOUNT` in the `provider` field results in error, because of the quoting [here](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/f34f012195d0b9718904ffa7a3a529f58167a74e/pkg/snowflake/database.go#L38).

The clearest for me would be to add fields `organization`, `account` and `account_locator`, that would replace `provider` in `snowflake_database` resource. Where user has to provide either `organization` and `account` or `account_locator`.

**Describe alternatives you've considered**

Could also live with `provider` properly handling `ORGANIZATION.ACCOUNT` format.

**Additional context**

marcin-vt:
Allowes to create database from share using `organization_name.account_name` notation. Add deprecation warning when using `provider` (based on [this](https://community.snowflake.com/s/article/show-shares-command-ui-data-sharing-changes) and [this](https://docs.snowflake.com/en/user-guide/admin-account-identifier.html#format-2-legacy-account-locator-in-a-region)).

Supports adding comments when creating a database from share (didn't work before).

```
resource "snowflake_database" "db_from_share" {
  comment = "This is a comment"
  from_share = {
	  account_name = "ACCOUNT_NAME"
	  organization_name = "ORGANIZATION_NAME"
	  share = "SHARE_NAME"
	}
  name = "DB_FROM_SHARE"
}
```

## Test Plan

* [x] acceptance tests

* [x] tested in own environment
* [ ] acceptance tests 

## References
resolves #1107

### DIFF
--- a/docs/resources/database.md
+++ b/docs/resources/database.md
@@ -49,13 +49,26 @@ resource "snowflake_database" "test3" {
 - `data_retention_time_in_days` (Number)
 - `from_database` (String) Specify a database to create a clone from.
 - `from_replica` (String) Specify a fully-qualified path to a database to create a replica from. A fully qualified path follows the format of "<organization_name>"."<account_name>"."<db_name>". An example would be: "myorg1"."account1"."db1"
-- `from_share` (Map of String) Specify a provider and a share in this map to create a database from a share.
+- `from_share` (Map of String) Specify a organization_name, account_name and a share in this map to create a database from a share. (see [below for block shema](#block--from_share))
 - `replication_configuration` (Block List, Max: 1) When set, specifies the configurations for database replication. (see [below for nested schema](#nestedblock--replication_configuration))
 - `tag` (Block List) Definitions of a tag to associate with the resource. (see [below for nested schema](#nestedblock--tag))
 
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+
+<a id="block--from_share"></a>
+### Nested Schema for `from_share`
+
+Required:
+
+- `organization_name` (String)
+- `account_name` (String)
+
+Optional:
+
+- `provider` (String) Account locator. Deprecated alternative to `organization_name.account_name`, see more [here](https://docs.snowflake.com/en/user-guide/admin-account-identifier.html#format-2-legacy-account-locator-in-a-region)
+
 
 <a id="nestedblock--replication_configuration"></a>
 ### Nested Schema for `replication_configuration`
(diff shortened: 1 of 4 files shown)


marcin-vt:
Just noticed this PR #1090

sfc-gh-swinkler:
this looks fine, but its failing the check-docs test. its unclear why, just fails with exit code 2. could you please look into this and we can get this merged?

marcin-vt:
@sfc-gh-swinkler - thanks for checking, I did correct that. 
There has been some discussions around this feature here: #1107 -  and in #1090. Don't know what is the approach in the end.
Also I might be unreachable for next two weeks - feel free to merge once ok with changes.

FYI @gouline.

gouline:
@sfc-gh-swinkler Is the breaking change really warranted? At the very least `provider` should be kept and `organization` added as an optional extension (to avoid `\".\"`). This will cause existing databases from shared to be needlessly recreated despite no actual changes (since Terraform state is different). 

Also `from_replica` still uses `\".\"`, so not quite sure why inconsistency in new changes is encouraged.

marcin-vt:
@gouline - `prov` is kept and, only 'marked' as deprecated. This should not force recreation of the db. There is a point that it is a bit  inconsistent. Additional open issue not covered here is importing existing object to state and refreshing it.

gouline:
Deprecation means it will be removed eventually, my point is that it's unnecessary. If the `\".\"` way of specifying the `provider` works for folks, it shouldn't be deprecated and `organization` should only be _added_ as an optional extra for anyone who prefers it.

I'd understand the deprecation if it was a new way that improves consistency. This does the opposite.

marcin-vt:
But yeah I am fine with removing "deprecation" - as you said using `prov` is also working...

gouline:
> deprecation was just a way of pointing out that snowflake has started to call it legacy

Current `provider` is unrelated to the account identifier format, it works with either. As I [responded](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1107#issuecomment-1190814728) in your issue, you can specify `provider = "ORGANIZATION\".\"ACCOUNT"`. This is consistent with `from_replica` and preference expressed by @sfc-gh-swinkler in #1090 (which is why I closed it).

> But yeah I am fine with removing "deprecation" - as you said using `prov` is also working...

Maintaining two ways of defining the same feature doesn't make much sense either. I'm sorry, this PR feels like duplication of an existing feature that enforces a semantic preference without introducing/fixing anything that cannot already be done, while also introducing an inconsistent precedent (i.e. not using fully-qualified names like in `from_replica`)...

sfc-gh-swinkler:
I am inclined to agree with @gouline that supporting two different methods is not proper. However, I think it should be made more clear in the documentation that `provider = "ORGANIZATION\".\"ACCOUNT"`works. I will go ahead and close this PR

Issue #1113, opened 2022-07-10
Gov Cloud Support


sfc-gh-pkommini:
**Is your feature request related to a problem? Please describe.**
Please support S3 Storage Integration `storage_provider  = 'S3GOV'` and API Integration `api_provider = AWS_GOV_API_GATEWAY`.

Right now it causes an error saying that the 2 options above are not supported.

**Describe the solution you'd like**
Add the above to options to the respective allow list for storage and api Integration providers.

PR #1118, opened 2022-07-12
feat: add AWS GOV support in api_integration

sfc-gh-kmaurya:
Added aws_gov_api_gateway and aws_gov_private_api_gateway in api_provider list in api_integration resource
## Test Plan

* [x] careful review

## References

* API Integration [doc](https://docs.snowflake.com/en/sql-reference/sql/create-api-integration.html)
* issue ref #1113


pkg/resources/api_integration_test.go
```diff
@@ -24,15 +24,15 @@ func TestAPIIntegrationCreate(t *testing.T) {
...
+		"api_provider":         "aws_gov_api_gateway",
 		"api_aws_role_arn":     "arn:aws:iam::000000000001:/role/test",
 	}
 	d := schema.TestResourceDataRaw(t, resources.APIIntegration().Schema, in)
 	r.NotNil(d)
 
 	WithMockDb(t, func(db *sql.DB, mock sqlmock.Sqlmock) {
 		mock.ExpectExec(
-			`^CREATE API INTEGRATION "test_api_integration" API_PROVIDER=aws_api_gateway API_AWS_ROLE_ARN='arn:aws:iam::000000000001:/role/test' API_ALLOWED_PREFIXES=\('https://123456.execute-api.us-west-2.amazonaws.com/prod/'\) ENABLED=true$`,
+			`^CREATE API INTEGRATION "test_api_integration" API_PROVIDER=aws_gov_api_gateway API_AWS_ROLE_ARN='arn:aws:iam::000000000001:/role/test' API_ALLOWED_PREFIXES=\('https://123456.execute-api.us-west-2.amazonaws.com/prod/'\) ENABLED=true$`,
```
sfc-gh-jalin, line 35: Should we create a new test rather than overwrite an existing test?

sfc-gh-kmaurya: Yes sure, I will create new test.

sfc-gh-afedorov (approved):
lgtm, but remove the tests because we don't have a GovCloud account to test on (we need this because we're building one!)

PR #1127, opened 2022-07-13
feat: tag grants

berosen:
* Adds support for granting a role `APPLY`/`OWNERSHIP`  privileges on a tag
*  Adds new unit and acceptance tests

## Test Plan

* [x] acceptance tests

* [x] unit tests

Tested with 
`TF_ACC=1 go test -v ./... -run TestAccTagGrant`


sfc-gh-swinkler:
sorry, i am really confused on this. why would you need to grant roles to a tag? Does this have any utility purpose?

berosen:
> sorry, i am really confused on this. why would you need to grant roles to a tag? Does this have any utility purpose?

Hi @sfc-gh-swinkler, sorry for any confusion. 

This resource isn't for granting a role to a tag, but for granting a _privilege_ on a tag to a specific role. Currently - from what I can tell - there is no resource that supports doing this. If a user creates a tag using the `snowflake_tag` resource, the role that the resource is deployed under will own that tag. In order to transfer ownership of that tag to another role, a user would need to go into Snowflake and run `grant ownership on tag <tag_name> to <role>`. The same is true if a user would like to give other roles the ability to apply a created tag to an object.

For example, `role_1` creates a tag in a schema and `role_2` would like to apply that tag to a column on a table `role_2` has created. Until `role_1` runs `grant apply on tag <tag> to role role_2`, `role_2` will encounter an ` Insufficient privileges to operate on...` error.  This PR creates a resource for handling those operations, the same way grants are handled for other resources in this provider. 

Please let me know if I'm missing something or can add further clarification.

sfc-gh-swinkler:
okay i get it now. thank you for clarifying, and i see how this could be useful now.

sfc-gh-swinkler (approved):

sfc-gh-swinkler:

Issue #1104, opened 2022-07-06
SQL compilation error - warehouse doest not exist


vlucas-caylent:
**Provider Version**
0.37.1

**Terraform Version**
0.14.0

**Describe the bug**

A little resume about it. My currently version of **snowflake-provider is 0.25.3** and i'm trying to update to the latest 0.37.1 I need to be able to use resource_monitor fully (And just after **Version 0.25.36** i can). But in the version **0.25.36** has a [bug ](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1075) and was fixed in the **0.37.0** . So i did some necessary default changes.  But i still receiving this error during the **Terraform Plan**

**Error below:**
```
2022-07-06T21:31:40.622Z [INFO]  plugin.terraform-provider-snowflake: 2022/07/06 21:31:40 [DEBUG] sql-conn-query [query SHOW PARAMETERS IN WAREHOUSE "admin_wh_xs" err 002003 (02000): SQL compilation error: Warehouse '"admin_wh_xs"' does not exist or not authorized. duration 212.728089ms args {}]: timestamp=2022-07-06T21:31:40.622Z
2022/07/06 21:31:40 [ERROR] <root>: eval: *terraform.EvalRefresh, err: 002003 (02000): SQL compilation error:
Warehouse '"admin_wh_xs"' does not exist or not authorized.
2022/07/06 21:31:40 [ERROR] <root>: eval: *terraform.EvalSequence, err: 002003 (02000): SQL compilation error:
Warehouse '"admin_wh_xs"' does not exist or not authorized.
```

**Expected behavior**

The warehouse should be validate, that warehouse exist and in version 0.25.3 is working good.

**Code samples and commands**

**warehouses.tf**
```
resource "snowflake_warehouse" "warehouses" {
  provider                     = snowflake.snowflake_sysadmin
  for_each                     = var.warehouses
  name                         = upper(each.key)
  comment                      = "${each.value.comment}${local.comment_suffix}"
  auto_resume                  = lookup(each.value, "auto_resume", var.warehouse_defaults.auto_resume)
  auto_suspend                 = lookup(each.value, "auto_suspend", var.warehouse_defaults.auto_suspend)
  max_cluster_count            = lookup(each.value, "max_cluster_count", var.warehouse_defaults.max_cluster_count)
  min_cluster_count            = lookup(each.value, "min_cluster_count", var.warehouse_defaults.min_cluster_count)
[...]

PR #1135, opened 2022-07-15
fix: Show Parameters Warehouse

lucasp0r:
When I try to use the command **SHOW PARAMETERS** with double quotes or quotes I receive an error in the Terraform, the problem is the print where I did the change. From my fork now it is working well.

The ignore quote cant be added during the process in the terraform plan/apply (terraform opens multiple sessions in snowflake) that's why we need to fix them.

## Test Plan

* [x] acceptance tests

* [x] …

## References

Link with the issue where I've open some days ago. 

https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1104
*

### DIFF
--- a/.github/workflows/team-slack-bot.yml
+++ b/.github/workflows/team-slack-bot.yml
@@ -1,11 +1,15 @@
 name: slack-notifications-prs-issues
 
-on: [pull_request, issues]
+on:
+  pull_request:
+    types: [opened]
+  issues:
+    types: [opened]
 
 jobs:
   slack-action:
-    runs-on: windows-latest
-    name: Test 2 [windows-latest]
+    runs-on: ubuntu-latest
+    name: Notify Team in Slack
 
     steps:
       - name: Send Slack Message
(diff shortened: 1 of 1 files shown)


sfc-gh-swinkler:
this causes the unit tests to fail. i am also worried that changing the way quotes work in this instance could impact other potential use cases. maybe there is a better way to do this. probably we should have a helper function that can handle formatting quotes properly.

```
SHOW VIEWS LIKE 'good_name' IN SCHEMA "test_db"."test_schema"
--- FAIL: TestWarehouseCreate (0.00s)
    warehouse_test.go:36: 
        	Error Trace:	warehouse_test.go:36
        	            				helpers.go:20
        	            				warehouse_test.go:32
        	Error:      	Received unexpected error:
        	            	call to Query 'SHOW PARAMETERS IN WAREHOUSE good_name' with args [] was not expected
        	Test:       	TestWarehouseCreate
--- FAIL: TestWarehouseRead (0.00s)
    warehouse_test.go:58: 
        	Error Trace:	warehouse_test.go:58
        	            				helpers.go:20
        	            				warehouse_test.go:55
        	Error:      	Received unexpected error:
        	            	call to Query 'SHOW PARAMETERS IN WAREHOUSE good_name' with args [] was not expected
        	Test:       	TestWarehouseRead
FAIL
```

sfc-gh-swinkler:
@lucasp0r are you able to fix the problem with the test case?

sfc-gh-swinkler:
@lucasp0r since I have not heard a response from you in over 60 days I will close this PR. If you think it is still an issue, you are welcome to open a new PR.

Issue #793, opened 2021-12-21
Can't set owner in snowflake_pipe


Dzvezdana:
**Provider Version**
0.25.30

**Terraform Version**
 0.14.6

**Describe the bug**

When running `terraform plan`, setting the owner in `snowflake_pipe`:

```
resource "snowflake_pipe" "pipe" { 
     ...
     owner = "ROLE1" 
}
```

fails with: 

```
Error: Value for unconfigurable attribute
on ../snowflake-pipe.tf line 65, in resource "snowflake_pipe" "pipe":
65:   owner             = "ROLE1"

Can't configure a value for "owner": its value will be decided automatically
based on the result of applying this configuration.
```

**Expected behavior**

According to the [documentation](https://registry.terraform.io/providers/chanzuckerberg/snowflake/latest/docs/resources/pipe), it should be possible to set the owner of Snowpipe.

senmm:
I have the same issue.

**Provider Version**
v0.37.1

**Terraform Version**
v1.2.2

sfc-gh-asawicki:
We are closing this issue as part of a cleanup described in [announcement](https://github.com/Snowflake-Labs/terraform-provider-snowflake/discussions/2755). If you believe that the issue is still valid in v0.89.0, please open a new ticket.

PR #1136, opened 2022-07-15
fix: doc of snowflake_pipe.

senmm:
Fix doc #793 

## Test Plan
N/A

## References

#793 

* `owner` is Read-Only. So I think that `owner` should not defined in `snowflake_pipe` resource.
  * https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/pipe.go


sfc-gh-swinkler (approved):

sfc-gh-swinkler:

PR #1143, opened 2022-07-18
feat: tag based masking policy

### LINKED ISSUE #1116: Add ability to set a masking policy on a tag
Snowflake now has the ability to set a masking policy on a tag object. See [Tag-based Masking Policies](https://docs.snowflake.com/en/user-guide/tag-based-masking-policies.html). 

For maximum flexibility given the variety of management styles Snowflake supports for tags and masking policies, this could take the form of a new `snowflake_tag_masking_policy_attachment` resource. Another benefit of that approach is it neatly resolves an issue arising from this pair of rules from the documentation:

> - A tag cannot be dropped if it is assigned to a masking policy.
> - A masking policy cannot be dropped if it is assigned to a tag.

Having a separate attachment resource which depends on both the tag and masking policy objects means that terraform will destroy the attachment before destroying the tag or masking policy.

The structure of the resource would be something like:

```hcl
resource "snowflake_tag_masking_policy_attachment" "test" {
    tag_database = "test_db"
    tag_schema = "test_schema"
    tag_name = "tag_name"

    masking_policy_database = "test_db"
    masking_policy_schema = "test_schema"
    masking_policy_name = "policy_name"
}
```

berosen:
* Adds the ability to attach a masking policy to a tag

closes #1116 

**Note** `SNOWFLAKE_WAREHOUSE` needs to be provided for this resource to query the  [POLICY_REFERENCES](https://docs.snowflake.com/en/sql-reference/functions/policy_references.html) table function.

## Test Plan

* [x] acceptance tests
* [x] unit tests 

Tested with `TF_ACC=1 go test -v ./... -run TestAcc_TagMaskingPolicyAssociation`

## References

* https://docs.snowflake.com/en/user-guide/tag-based-masking-policies.html


sfc-gh-swinkler:
sorry for the long delay in responding to this PR, my wife died in July and I was on grievance leave. I have been catching up over the past couple weeks. This is a complicated PR so i have been thinking about it for a few days now.

We are actually in the process of changing how tags work in the provider. It won't break backwards compatibility, but it does affect how new resources that involved tagging should be implemented. What I would like to do is wait until we get the change implemented, which should be later this week, and then address this masking policy attachment after that. I do like this PR, but there will need to be a few changes to get it to align with the new model. If you are not able to make the change, then i can take over and make sure this gets done in the next sprint.

berosen:
> sorry for the long delay in responding to this PR, my wife died in July and I was on grievance leave. I have been catching up over the past couple weeks. This is a complicated PR so i have been thinking about it for a few days now.
> 
> We are actually in the process of changing how tags work in the provider. It won't break backwards compatibility, but it does affect how new resources that involved tagging should be implemented. What I would like to do is wait until we get the change implemented, which should be later this week, and then address this masking policy attachment after that. I do like this PR, but there will need to be a few changes to get it to align with the new model. If you are not able to make the change, then i can take over and make sure this gets done in the next sprint.

I'm really sorry to hear that, sending you my condolences. 

Sounds good to me. Feel free to keep me posted with what changes need to be made once the implementation change is complete. I'll let you know if I'm unable to work on them in a timely manner. Thank you for taking a look at this.

sfc-gh-swinkler:
basically how it will work is that there is a new resource called "snowflake_tag_attachment". it will replace the current tag schema on each resource. below is sample configuration

```
resource "snowflake_tag_attachment" "cost_center_tag_attachment" {
tag_id = snowflake_tag.cost_center.id
value = "finance"
resource_id = snowflake_database.my_db.id
resource_type = "DATABASE"
}
```

as you can see we are using a tag_id rather than database, schema and name. the tag_id is a string which has the same infomration, and is flexible in the way that the data is passed in, for example:
"database"."schema"."name"
"database.schema.name"
"database|schema.name" (structure of the snowflake_tag resource id)

are all allowed.

We should be getting this PR done either tomorrow or Monday.

sfc-gh-swinkler:
https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1187 @berosen would appreciate you taking a look at this

berosen:
> #1187 @berosen would appreciate you taking a look at this

Will do

sfc-gh-swinkler:
@berosen i merged that change because we promised a customer to get the release done this week. i also had to fix some integration test problems and so now your branch has some merge conflicts that need to be resolved. would you like me to work on this PR for you to get it up-to-date? or are you comfortable making the changes yourself? sorry for the inconvenience, i really want to help you get this merged.

berosen:
> 

Thanks for this @sfc-gh-swinkler. I'll try and get this updated in the next couple days. Mid next week at the very latest, if that's alright. The change seems fairly straight forward. Just need to carve out a few minutes for it.

sfc-gh-swinkler:
thanks @berosen please let me know how i can assist in any way

berosen:
Hey @sfc-gh-swinkler , the merge conflicts should be resolved and I've changed the resource to use the `tag_id`. This should be ready for a review when you have a chance. Thanks!

sfc-gh-swinkler:
okay good good. so now i just have two further suggestions:
1. instead of doing 
```
  masking_policy_database = snowflake_database.test.name
  masking_policy_schema   = snowflake_schema.test.name
  masking_policy_name     = snowflake_masking_policy.example_masking_policy.name
```
you can do 
```
masking_policy_id = snowflake_masking_policy.example_masking_policy.id
```
the reason why is because this id already has the db, schema and name https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/masking_policy.go#L68-L69

2. change the name of the resource to `snowflake_tag_masking_policy_association`. The work "association" seems to be favored in Snowflake, so i just want to be consistent with naming convention.

berosen:
Good point, will do. Thanks, I'll make those changes.

sfc-gh-swinkler:
looking good, looking real good. as long as this passes integration tests i will merge this

berosen:
@sfc-gh-swinkler I've updated the CI [here](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1143/commits/227a37f89423db88f00a6cab6899e14d824b2e33) to include `SNOWFLAKE_WAREHOUSE`. Could you add a value for this in your CI secrets? That should fix the failing test.

sfc-gh-swinkler:
@berosen that environment variable is already set

sfc-gh-swinkler:
not sure what is going on here. the `SNOWFLAKE_WAREHOUSE` environment variable is definitely set on the repo settings. I also updated it just in case
Screen Shot 2022-08-26 at 2 32 41 PM

Its also not clear to me why this even needs to be set for the masking policy association. Nowhere in the documentation does it state that this needs to be set: https://docs.snowflake.com/en/user-guide/tag-based-masking-policies.html. As far as i know this only needs to be set for Python UDF. Are you setting this locally when you test the resource? If so, what exact line is causing this to fail if the environment variable is not set? Maybe can add some debugging there to figure out what the problem is. Because default warehouse gets set in the provider from environment variable here:
https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/provider/provider.go#L366

berosen:
hmm that's odd. Yes I'm setting that variable locally. The line that needs the warehouse is [this one](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/227a37f89423db88f00a6cab6899e14d824b2e33/pkg/snowflake/tag.go#L173) because it uses the `policy_references` table function described in the docs [here in step 5](https://docs.snowflake.com/en/user-guide/tag-based-masking-policies.html#:~:text=A%20masking%20policy%20that%20uses,masking%20policy%20to%20the%20tag.). I'm not super familiar with GitHub actions but I'm wondering if [this](https://github.com/actions/checkout/discussions/277) is what's going on and the CI change just isn't being picked up.

sfc-gh-swinkler (approved):

sfc-gh-swinkler:
going to merge this and try to figure out what the problem is on my own. i think you are right, that its probably not picking up the changes.

sfc-gh-swinkler:

berosen:
Thanks @sfc-gh-swinkler , look like that was the issue and ended up passing [here](https://github.com/Snowflake-Labs/terraform-provider-snowflake/runs/8121289933?check_suite_focus=true). Appreciate your help on this.

alldoami:
@berosen wanna add the warehouse requirement to the example in the docs 😄 Thanks for adding this to the provider 🥳

Issue #1152, opened 2022-07-22
Allow create transient databases


sebaap:
**Is your feature request related to a problem? Please describe.**

Snowflake [documentation](https://docs.snowflake.com/en/sql-reference/sql/create-database.html#optional-parameters) for `CREATE DATABASE` mentions the `TRANSIENT` optional parameter, however it's not currently possible to create a transient database with the database resource

**Describe the solution you'd like**

Ideally, a new optional parameter `is_transient` in the database resource, same as for the [schema resource](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/schema#is_transient), that allows to create a transient database.

**Describe alternatives you've considered**
I guess an option would be to use snowsql provider and write custom queries for it, but it would overcomplicate things.

**Additional context**
—

PR #1165, opened 2022-08-03
feat: transient database

sebaap:
This PR adds support for transient databases (closes [#1152](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1152)). 

It was a bit tricky to add this functionality with the current database implementation so I've slightly refactored it to an iterative approach, following how it's done for the [schema resource](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/schema.go). Hopefully it will make it easier to implement other functionalities.
I only modified the code for database and database clone, so database from share and database from replication are still using the old implementation.
I've added some extra tests and ran the acceptance tests using a trial snowflake account as suggested in the readme.
Also installed the provider locally and did some extra testing, everything seems to be working fine.

## Test Plan

* [x] acceptance tests
* [x] added extra unit tests
* [x] install provider locally for extra testing

## References
* https://docs.snowflake.com/en/sql-reference/sql/create-database.html#optional-parameters

### DIFF
--- a/docs/resources/database.md
+++ b/docs/resources/database.md
@@ -46,6 +46,7 @@ resource "snowflake_database" "test3" {
 ### Optional
 
 - `comment` (String)
+- `is_transient` (Boolean) Specifies a database as transient. Transient databases do not have a Fail-safe period so they do not incur additional storage costs once they leave Time Travel; however, this means they are also not protected by Fail-safe in the event of a data loss.
 - `data_retention_time_in_days` (Number)
 - `from_database` (String) Specify a database to create a clone from.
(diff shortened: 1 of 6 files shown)


sfc-gh-swinkler:
i actually like this change. the generic resource builder was very confusing and hard to debug. i think we should be refactoring more resources to each have their own builder.

sfc-gh-swinkler:
can you run `make docs` to regenerate docs? since you are updating some documentation

sebaap:
> can you run `make docs` to regenerate docs? since you are updating some documentation which is causing check-docs to fail

sure thing, done

sfc-gh-swinkler (approved):

sfc-gh-swinkler:

PR #1180, opened 2022-08-19
feat: Creating new resource for Tag reference

sfc-gh-jalin:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/snowflake/tag_attachment.go
```diff
@@ -0,0 +1,107 @@
...
+	return fmt.Sprintf(`ALTER %v %v SET TAG "%v"."%v"."%v" = '%v'`, tb.objectType, tb.resourceId, tb.databaseName, tb.schemaName, tb.tagName, tb.tagValue)
+}
+
+// Drop returns the SQL query that will remove a tag from a resource.
+func (tb *TagAttachmentBuilder) Drop() string {
+	return fmt.Sprintf(`ALTER %v %v UNSET TAG %v`, tb.objectType, tb.resourceId, tb.tagName)
+}
+
+func ListTagAttachments(tb *TagAttachmentBuilder, db *sql.DB) ([]tagAttachment, error) {
+	stmt := fmt.Sprintf(`SELECT * FROM SNOWFLAKE.ACCOUNT_USAGE.TAG_REFERENCES WHERE TAG_NAME = '%v' AND DOMAIN = '%v' AND TAG_VALUE = '%v'`, tb.tagName, tb.objectType, tb.tagValue)
```
berosen, line 92: @sfc-gh-jalin  Quick question on this, from what I see in the [docs](https://docs.snowflake.com/en/sql-reference/account-usage.html) the `TAG_REFERENCES` view can have latency up to 2 hours, where as there isn't any latency using `INFORMATION SCHEMA`. I'm not super familiar with these views, but do you see a scenario where a tag has been attached but a subsequent read operation doesn't reflect the attachment because the view isn't up to date yet? This may lead to spurious diffs and some possibly unexpected behavior.

sfc-gh-jalin: Hi @berosen, our team did some manual testing and found that the latency is about one hour on average using the TAG_REFERENCES view. We haven't looked into the INFORMATION_SCHEMA yet but it's probably something we should look into. For now we have a retry mechanism

Issue #1050, opened 2022-06-09
Return Type TABLE() for snowflake_procedure is not working


srgva:
**Provider Version**

0.35.0

**Terraform Version**

1.1.7

**Describe the bug**

When creating a snowflake_procedure with return_type="TABLE()", the procedure is created in Snowflake, but not saved in the state file and an error is thrown:
 resource "snowflake_procedure" "eap_network_policies_show_application_policy" {
│
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call.

**Expected behavior**

The procedure is saved in the state file. 

**Code samples and commands**

Please add code examples and commands that were run to cause the problem.

**Additional context**

The problem seems to be cause by the regex https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/f9036e8914b9c139eb6798276124c5544a083eb8/pkg/resources/procedure.go#L272

Snowflake returns "TABLE ()" (with additional empty space after TABLE) from "describe procedure".

remi-f-artelia:
This bug is still not fixed in `0.91.0`
```
snowflake_procedure.alert2_proc must be replaced
-/+ resource "snowflake_procedure" "alert2_proc" {
      ~ id                  = "\"DEMO\".\"SCHEMA_TEST\".\"alert2_proc\"" -> (known after apply)
        name                = "alert2_proc"
      ~ return_type         = "TABLE ()" -> "TABLE" # forces replacement
      ~ statement           = <<-EOT
         [....]
         EOT
```
Impact: recreation of the resource and import not allowed

```
Error: Invalid import id argument
│
│   on procedures.tf line 60, in import:
│   60:   id       = snowflake_procedure.alert2_proc.id
│
│ The import block "id" argument depends on resource attributes that cannot be determined until apply, so Terraform cannot plan to import this resource.
```

sfc-gh-jmichalak:
Hi @remi-f-artelia 👋 

We are reworking procedures and have this issue on our list. We'll let you know about a release with the fix.

sfc-gh-asawicki:
Hey @remi-f-artelia @srgva.

The new procedure resources from v0.100.0 (also from v.1.0.0) should handle empty return tables correctly. PLease check them out and let us know.

sfc-gh-dszmolka:
[...]

PR #1181, opened 2022-08-19
fix: add support in procedure for return type table()

doino-gretchenliev:
Snowflake returns `return_type` `table()` as `table ()`. The provider regex doesn't match the space and caused the create and import operations to fail. As part of this PR the space matching is fixed.

## Test Plan
I've added the required specific test case in the test suite.

## References
https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1050


sfc-gh-swinkler (approved):
LGTM

sfc-gh-swinkler:

PR #1187, opened 2022-08-23
feat: tag association resource

sfc-gh-swinkler:
This PR changes the way that tags work. Instead of using tagschema on each resource, it creates a new resource specifically for associating tags with resources. this fixes a lot of the weird dependency problems people had using the old method. the tagschema is still supported but marked as deprecated and will be removed in a future release.

also there was a lot of weird problems with integration tests failing that are fixed that probably should have been done in a separate PR,

## Test Plan

* [x] acceptance tests

## References

*


sfc-gh-jlove (approved):

pkg/snowflake/tag_association.go
```diff
@@ -0,0 +1,121 @@
...
+	return fmt.Sprintf(`ALTER %v "%v" SET TAG "%v"."%v"."%v" = '%v'`, tb.objectType, tb.objectName, tb.databaseName, tb.schemaName, tb.tagName, tb.tagValue)
+}
+
+// Drop returns the SQL query that will remove a tag from an object.
+func (tb *TagAssociationBuilder) Drop() string {
+	return fmt.Sprintf(`ALTER %v "%v" UNSET TAG "%v"."%v"."%v"`, tb.objectType, tb.objectName, tb.databaseName, tb.schemaName, tb.tagName)
+}
+
+func ListTagAssociations(tb *TagAssociationBuilder, db *sql.DB) ([]tagAssociation, error) {
+	stmt := fmt.Sprintf(`SELECT * FROM SNOWFLAKE.ACCOUNT_USAGE.TAG_REFERENCES WHERE TAG_NAME = '%v' AND DOMAIN = '%v' AND TAG_VALUE = '%v'`, tb.tagName, tb.objectType, tb.tagValue)
```
christophkreutzer, line 106: @sfc-gh-swinkler Would you be so kind to provide a reasoning why the tags are read this way?

I see multiple potential issues here:
* minor nitpick: tags on deleted objects are not filtered (`OBJECT_DELETED IS NULL`)
* the query filters only on tag name instead of the fully qualified identifier (missing `TAG_DATABASE` and `TAG_SCHEMA`) - if I have two tags with the same name in different schemas, this would also select the other one
* why via the ACCOUNT_USAGE view? It would be possible to query the tags without any latency using [`SYSTEM$GET_TAG`](https://docs.snowflake.com/en/sql-reference/functions/system_get_tag.html) and compare the value afterwards with the state, if this is needed.
* Also, currently the query doesn't include the object name, so we are just querying if there is a tag on any object of the same domain with the given tag value set?

sfc-gh-swinkler: @christophkreutzer thank you for pointing these out. i must say that i am not an expert at the Snowflake product, so I did not know about this SYSTEM$GET_TAG function. These would be good changes to make. If you can make a request with your account manager, then i can prioritize getting this done.

christophkreutzer: @sfc-gh-swinkler thanks for the fast response! I created a first draft in PR #1261, and I'll let our account team know of the issues if that helps you taking a look at it.

Issue #967, opened 2022-04-08
Provider revoke grants created outside terraform.


gal-pal:
**Provider Version**

0.30.0

**Terraform Version**

1.1.7

**Describe the bug**

Granting resources no longer working properly.
For example, I had a resource snowflake_account_grant created in an earlier version of provider 0.28.8

```
resource "snowflake_account_grant" "grantCreateDatabase"  {
    roles               = ["SANDBOX_SYSADMIN_ROLE"]
    privilege           = "CREATE DATABASE"
}
```

I upgraded snowflake provider to version 0.30.0 I modified the resource to compatibly with existing environment:

```
 resource "snowflake_account_grant" "grantCreateDatabase"  {
    roles               = ["SANDBOX_SYSADMIN_ROLE"]
    privilege           = "CREATE DATABASE"
    enable_multiple_grants = true
}

```
Unfortunately, regardless of what value I give for the parameter "enable_multiple_grants", each time resources created outside the terraform are selected for deletion.

```
13:57:23  Note: Objects have changed outside of Terraform
13:57:23  
13:57:23  Terraform detected the following changes made outside of Terraform since the
13:57:23  last "terraform apply":
13:57:23  
13:57:23    # module.snowflake.snowflake_account_grant.grantCreateDatabase has changed
13:57:23    ~ resource "snowflake_account_grant" "grantCreateDatabase" {
13:57:23          id                = "ACCOUNT|||CREATE DATABASE|SANDBOX_SYSADMIN_ROLE|false"
13:57:23        ~ roles             = [
13:57:23            + "DEVEL_ROLE",
13:57:23            + "DEVEL_SYSADMIN_ROLE",
13:57:23              # (1 unchanged element hidden)
13:57:23          ]
13:57:23          # (2 unchanged attributes hidden)
13:57:23      }

13:57:23    # module.snowflake.snowflake_account_grant.grantCreateDatabase will be updated in-place
13:57:23    ~ resource "snowflake_account_grant" "grantCreateDatabase" {
13:57:23        + enable_multiple_grants = false
13:57:23          id                     = "ACCOUNT|||CREATE DATABASE|SANDBOX_SYSADMIN_ROLE|false"
13:57:23        ~ roles                  = [
13:57:23            - "DEVEL_ROLE",
13:57:23            - "DEVEL_SYSADMIN_ROLE",
13:57:23              # (1 unchanged element hidden)
13:57:23          ]
[...]

PR #824, opened 2022-01-19
fix: Allow multiple resources of the same object grant

daniepett:
- Filtering grants on read to only return those managed by the resource.
- Updated grantID to contain a list of all the roles in the current grant resource.

These changes allows grants to multiple roles to be done in different resources.

**This change comes with the caveat that grants applied OUTSIDE Terraform will not be revoked. This is because the resource won't know if it's maintained by another resource or created elsewhere**

Previously the resource id had the structure:
resourceName|schemaName|ObjectName|Privilege|GrantOption
For example: DWH|TEST||USAGE|FALSE

New structure:
resourceName|schemaName|ObjectName|Privilege|Roles|GrantOption

```hcl
# id = DWH|TEST||USAGE|ROLE_A,ROLE_B|FALSE
resource "snowflake_schema_grant" "grant_a_b" {
  database_name = "DWH"
  schema_name   = "TEST"

  privilege = "USAGE"
  roles     = ["ROLE_A", "ROLE_B"]
  with_grant_option = false
}

# id = DWH|TEST||USAGE|ROLE_C|FALSE
resource "snowflake_schema_grant" "grant_c" {
  database_name = "DWH"
  schema_name   = "TEST"

  privilege = "USAGE"
  roles     = ["ROLE_C"]
  with_grant_option = false
}
```

## Test Plan

* [ ] acceptance tests
* [X] unit tests

## References

* https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/210
* https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/560
* https://github.com/chanzuckerberg/terraform-provider-snowflake/issues/733

PR #1225, opened 2022-09-21
fix: Reverted enable_multiple_grants

ChrisIsidora:
Revert for enable_multiple_grants because it does not work properly as intended and is causing side effects. One of the biggest side effects is that because the roles are part of the id, when a role is added to an existing object grant, the id changes, which in turn triggers the resource to be replaced which means that the current roles will be revoked then granted again. This causes a significant disruption when users that are using the affected roles are running a query when the grants are being rolled out.

This side effect was only noticed by us now because we were creating new grants with the new id format which includes the roles in the id itself. For our old grant resources this is not the case because it was still using the id format that excluded the roles prior to the introduction of the enable_multiple_grants feature.

## Test Plan
* [x] acceptance tests
* [x] unit tests

## References
* #824 
* #922
* #941
* #967

### DIFF
--- a/docs/resources/account_grant.md
+++ b/docs/resources/account_grant.md
@@ -25,7 +25,6 @@ resource snowflake_account_grant grant {
 
 ### Optional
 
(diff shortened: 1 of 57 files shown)


ChrisIsidora:
@sfc-gh-swinkler Can you please review?

PR #1228, opened 2022-09-21
feat: Added Missing Grant Updates + Removed ForceNew

ChrisIsidora:
Added missing update function on resource grants and removed ForceNew from Shares + Roles. This is added to mitigate the current side effects of ForceNew when you add a Share or Role to an existing Grant. Currently existings grants are revoked and then recreated when you add a share or role to a existing grant that has no implementation of the update function and has ForceNew set to true. This causes current queries / users sessions that uses the existing grants to fail because the existing grants where revoked to be recreated a couple of seconds later. This feature makes it possible to only apply the difference. 

This has already been done for a couple of resource grants and works properly.

## Test Plan
* [x] acceptance tests
* [x] unit tests

### DIFF
--- a/pkg/resources/external_table_grant.go
+++ b/pkg/resources/external_table_grant.go
@@ -45,14 +45,12 @@ var externalTableGrantSchema = map[string]*schema.Schema{
 		Elem:        &schema.Schema{Type: schema.TypeString},
 		Optional:    true,
 		Description: "Grants privilege to these roles.",
-		ForceNew:    true,
 	},
 	"shares": {
 		Type:        schema.TypeSet,
 		Elem:        &schema.Schema{Type: schema.TypeString},
 		Optional:    true,
 		Description: "Grants privilege to these shares (only valid if on_future is false).",
-		ForceNew:    true,
 	},
 	"on_future": {
 		Type:        schema.TypeBool,
@@ -84,6 +82,7 @@ func ExternalTableGrant() *TerraformGrantResource {
 			Create: CreateExternalTableGrant,
 			Read:   ReadExternalTableGrant,
 			Delete: DeleteExternalTableGrant,
+			Update: UpdateExternalTableGrant,
 
 			Schema: externalTableGrantSchema,
 			Importer: &schema.ResourceImporter{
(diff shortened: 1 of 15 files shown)


ChrisIsidora:
@sfc-gh-swinkler Can this be reviewed and then released (it's quite blocking for us)

ChrisIsidora:
Can someone take a look at this? @sfc-gh-jlove @sfc-gh-swinkler?

pkg/resources/external_table_grant.go
```diff
@@ -216,3 +215,56 @@ func DeleteExternalTableGrant(d *schema.ResourceData, meta interface{}) error {
 	}
 	return deleteGenericGrant(d, meta, builder)
 }
+
+// UpdateExternalTableGrant implements schema.UpdateFunc
+func UpdateExternalTableGrant(d *schema.ResourceData, meta interface{}) error {
+	// for now the only thing we can update are roles or shares
+	// if nothing changed, nothing to update and we're done
+	if !d.HasChanges("roles", "shares") {
```
sfc-gh-swinkler, line 223: since everything else is marked as forcenew, this code will actually never run. but at least it makes it clear what future work needs to be done

sfc-gh-swinkler (approved):

sfc-gh-swinkler:
> --- FAIL: TestProvider (0.03s)
>     provider_test.go:22: 
>         	Error Trace:	/home/runner/work/terraform-provider-snowflake/terraform-provider-snowflake/pkg/provider/provider_test.go:22
>         	Error:      	Received unexpected error:
>         	            	1 error occurred:
>         	            		* resource snowflake_row_access_policy_grant: No Update defined, must set ForceNew on: []string{"roles"}

@ChrisIsidora this looks good, but you missed one of the updates functions to the snowflake_row_access_policy_grant schema.

pkg/resources/external_table_grant.go
ChrisIsidora: @sfc-gh-swinkler The ForceNew should be removed that's why the Update was implemented

ChrisIsidora:
@sfc-gh-swinkler I added the missing Update Function for snowflake_row_access_policy_grant. Can you run the tests again?

sfc-gh-swinkler:
thank you for your contribution!

PR #1231, opened 2022-09-22
feat: add connection param for snowhouse

sfc-gh-swinkler:
Adding in a "Application" connection param which can be used to determine where requests are coming from.
## Test Plan

* [x] acceptance tests

* [ ] …

## References

*


sfc-gh-jlove (approved):
LGTM

PR #1242, opened 2022-09-28
fix: Update 'user_ownership_grant' schema validation

antonblr:
Make `user_ownership_grant` resource schema validation consistent with upstream resources schemas - [user.name](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/da886d4e05f7bb9443168f0fa04b8b397a1db5c7/pkg/resources/user.go#L36) and [role.name](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/da886d4e05f7bb9443168f0fa04b8b397a1db5c7/pkg/resources/role.go#L17) 

Also per https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#double-quoted-identifiers user name could be anything. And it is double-quoted as per https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/da886d4e05f7bb9443168f0fa04b8b397a1db5c7/pkg/snowflake/user_ownership_grant.go#L36

## Test Plan

* [ ] acceptance tests
* [ ] unut tests

## References

* Fixes https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1184

PR #1245, opened 2022-09-29
fix: Table Tags Acceptance Test

ChrisIsidora:
Fixed Table Tags Acceptance Test. @sfc-gh-swinkler  Some reason this test was failing, this is  the fix for that test

## Test Plan
* [x] acceptance tests

## References
* #1228 
* #1242 
* #1244


sfc-gh-swinkler:
@ChrisIsidora thanks for this. I am not sure why these tests are suddenly failing. Actually we could even remove them, since tags block on tables has been deprecated in favor of the `snowflake_tag_association` resource.

sfc-gh-swinkler (approved):

PR #1247, opened 2022-09-30
fix: function resource - python packages

chasegiles-built:
### Problem statement
The snowflake describe function outputs the python packages from a function as a list of strings like `['pytz','pandas']`. The resource was reading this in as one item instead of two. This would result in terraform detecting changes when there actually were no changes.

Something like:
```
-"'pytz','pandas'",
+"pytz",
+"pandas",
```

### Summary of changes
Corrected the split by removing a space and removed single quotes from the string values.

Verified this was working as expected in my fork.

## Test Plan
* [ ] acceptance tests

### DIFF
--- a/pkg/resources/function.go
+++ b/pkg/resources/function.go
@@ -336,17 +336,17 @@ func ReadFunction(d *schema.ResourceData, meta interface{}) error {
 				}
 			}
 		case "packages":
-			packagesString := strings.ReplaceAll(strings.ReplaceAll(desc.Value.String, "[", ""), "]", "")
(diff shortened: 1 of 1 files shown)


sfc-gh-swinkler (approved):

sfc-gh-swinkler:
@chasegiles-built sfc-gh-jalin no longer works at Snowflake. So unfortunately for the time being it is only me who can approve PRs. Hopefully we can backfill position soon.

sfc-gh-swinkler:
@chasegiles-built thank you for your contribution!

PR #1261, opened 2022-10-07
fix: clean up tag association read

christophkreutzer:
As discussed in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1187#discussion_r989292991, there are some potential issues in the current implementation:
* minor nitpick: tags on deleted objects are not filtered (OBJECT_DELETED IS NULL)
* the query filters only on tag name instead of the fully qualified identifier (missing TAG_DATABASE and TAG_SCHEMA) - if I have two tags with the same name in different schemas, this would also select the other one
* why via the ACCOUNT_USAGE view? It would be possible to query the tags without any latency using SYSTEM$GET_TAG and compare the value afterwards with the state, if this is needed.
* Also, currently the query doesn't include the object name, so we are just querying if there is a tag on any object of the same domain with the given tag value set?

This is only a first draft for discussion.

## Test Plan
* [ ] acceptance tests (should be handled by the existing tests?)

## References
* Track the tags: https://docs.snowflake.com/en/user-guide/object-tagging.html#step-5-track-the-tags
* `SYSTEM$GET_TAG`: https://docs.snowflake.com/en/sql-reference/functions/system_get_tag.html


pkg/snowflake/tag_association.go
```diff
@@ -21,19 +21,7 @@ type TagAssociationBuilder struct {
...
-	ObjectDeleted  sql.NullString `db:"OBJECT_DELETED"`
-	ObjectId       sql.NullString `db:"OBJECT_ID"`
-	ObjectName     sql.NullString `db:"OBJECT_NAME"`
-	ObjectSchema   sql.NullString `db:"OBJECT_SCHEMA"`
-	TagDatabase    sql.NullString `db:"TAG_DATABASE"`
-	TagId          sql.NullString `db:"TAG_ID"`
-	TagName        sql.NullString `db:"TAG_NAME"`
-	TagSchema      sql.NullString `db:"TAG_SCHEMA"`
-	TagValue       sql.NullString `db:"TAG_VALUE"`
+	TagValue sql.NullString `db:"TAG_VALUE"`
```
christophkreutzer, line 24: It seems like `tagAssociation` is never accessed, so removing the attributes shouldn't harm anything.

`ListTagAssociations` is only called twice in `pkg/resources/tag_association.go`, one time the result isn't used at all, and one time only the object count of the array is accessed.

pkg/snowflake/tag_association.go
```diff
@@ -103,7 +92,7 @@ func (tb *TagAssociationBuilder) Drop() string {
 }
 
 func ListTagAssociations(tb *TagAssociationBuilder, db *sql.DB) ([]tagAssociation, error) {
-	stmt := fmt.Sprintf(`SELECT * FROM SNOWFLAKE.ACCOUNT_USAGE.TAG_REFERENCES WHERE TAG_NAME = '%v' AND DOMAIN = '%v' AND TAG_VALUE = '%v'`, tb.tagName, tb.objectType, tb.tagValue)
+	stmt := fmt.Sprintf(`SELECT SYSTEM$GET_TAG('"%v"."%v"."%v"', '%v', '%v') TAG_VALUE WHERE TAG_VALUE IS NOT NULL`, tb.databaseName, tb.schemaName, tb.tagName, tb.objectName, tb.objectType)
```
christophkreutzer, line 95: I tried a minimal-invasive way to switch to the `SYSTEM$GET_TAG` function, probably it could be improved quite a bit (like checking if the returned `TAG_VALUE` is actually the same value as applied in the resource to detect diffs).

Behavior of the `SYSTEM$GET_TAG` function:
* If the tag identifier references a non-existing or unauthorized tag object, the query fails.
* If the object identifier references a non-existing or unauthorized object, the query fails.
* If the tag is not set on the object, the function returns a SQL NULL
* If the tag is set on the object, the function returns a string with the tag value.

I didn't expect using the column alias would work in the WHERE clause, but it does:
image

christophkreutzer (commented):
I added some comments explaining the changes, hope it helps understanding.

Probably the whole implementation could be streamlined a bit, e.g.
* the timeouts in `pkg/resources/tag_association.go` are probably no longer needed, as the query is based purely on the Metadata
image
* one could actually use the returned tag value

However, I'm a bloody rookie when it comes to Terraform Providers and Go in general...

sfc-gh-swinkler:
there is also an attribute called "skip_validation" that could be marked as deprecated since it was only used to disable the read. otherwise looks good

christophkreutzer:
> there is also an attribute called "skip_validation" that could be marked as deprecated since it was only used to disable the read.

I tried to mark it as deprecated - feel free to propose a better text :)

I also added Read/Update similar to how it is done for tags itself - so it should now be able to reconcile manually modified tag associations. Hope you are fine with that, otherwise I'm fine discarding these commits, too.

christophkreutzer:
@sfc-gh-swinkler could you please start the workflows again?

christophkreutzer:
Hi @sfc-gh-swinkler @sfc-gh-jlove,
sorry to bother you, but could you please trigger the tests again and maybe give this a look?

christophkreutzer:
> Integration tests failure for [8f21957](https://github.com/Snowflake-Labs/terraform-provider-snowflake/actions/runs/3292380869)

[c9d57f4](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1261/commits/c9d57f47335ec8da5d62d9074d5128698e9fec6b) should hopefully look better :)

sfc-gh-swinkler (approved):

sfc-gh-swinkler:
im working on this resource right now anyways so i will make sure the integrations tests pass before new release is created. thanks @christophkreutzer !

PR #1273, opened 2022-10-12
chore: update docs style of grant modules

mnagaa:
## Summary

In this PR, I want to fix the styles of the terraform snowflake provider documentation to be the same format. There are some expression of the variables in the example (like below) and it is not suitable for the official documentation.

This PR only includes updates for modules that grant privileges:

- use UPPERCASE for granting privilege (`privilege = "select"` => `privilege = "SELECT"`)
- do not use abbr like `database_name = "db"` => `database_name = "database"` and `warehouse_name = "wh"` => `warehouse_name = "warehouse"` because some other examples already use "database" for their definitions.

Furthermore, changing the definitions with double quotations seems better as follows.
**before**
```
resource snowflake_warehouse_grant grant {
  warehouse_name = "wh"
  privilege      = "MODIFY"

  roles = [
    "role1",
  ]

  with_grant_option = false
}
```

**after**
```
resource "snowflake_warehouse_grant" "grant" {
  warehouse_name = "warehouse"
  privilege      = "MODIFY"

  roles = ["role1", "role2"]

  with_grant_option = false
}
```

### DIFF
--- a/docs/resources/database_grant.md
+++ b/docs/resources/database_grant.md
@@ -13,8 +13,8 @@ description: |-
 ## Example Usage
 
 ```terraform
-resource snowflake_database_grant grant {
-  database_name = "db"
+resource "snowflake_database_grant" "grant" {
+  database_name = "database"
 
   privilege = "USAGE"
   roles     = ["role1", "role2"]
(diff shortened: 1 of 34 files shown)


mnagaa:
I will open further pull requests for fixing documentation if this idea is good for this project.

PR #1291, opened 2022-10-20
feat: task with allow_overlapping_execution option

mlorek:
Added ALLOW_OVERLAPPING_EXECUTION option to the task resource

## Test Plan
test included

https://docs.snowflake.com/en/sql-reference/sql/create-task.html#optional-parameters

### DIFF
--- a/pkg/resources/task.go
+++ b/pkg/resources/task.go
@@ -16,7 +16,8 @@ import (
 )
 
 const (
-	taskIDDelimiter = '|'
+	taskIDDelimiter           = '|'
+	AllowOverlappingExecution = "allow_overlapping_execution"
 )
 
 var taskSchema = map[string]*schema.Schema{
@@ -106,6 +107,12 @@ var taskSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Description: "Specifies the name of the notification integration used for error notifications.",
 	},
+	AllowOverlappingExecution: {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     false,
+		Description: "By default, Snowflake ensures that only one instance of a particular DAG is allowed to run at a time, setting the parameter value to TRUE permits DAG runs to overlap.",
+	},
 }
 
 type taskID struct {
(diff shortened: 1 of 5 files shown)


pkg/snowflake/task.go
```diff
@@ -233,6 +244,16 @@ func (tb *TaskBuilder) RemoveComment() string {
 	return fmt.Sprintf(`ALTER TASK %v UNSET COMMENT`, tb.QualifiedName())
 }
 
+// ChangeAllowOverlappingExecution returns the sql that will change the ALLOW_OVERLAPPING_EXECUTION for the task.
+func (tb *TaskBuilder) ChangeAllowOverlappingExecution(flag bool) string {
```
sfc-gh-swinkler, line 248: this function signature should probably be changed to `SetAllowOverlappingExecution(overlappingExecution bool) string {}`

pkg/resources/task.go
```diff
@@ -619,6 +635,21 @@ func UpdateTask(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	if d.HasChange(AllowOverlappingExecution) {
+		var q string
+		n := d.Get(AllowOverlappingExecution)
```
sfc-gh-swinkler, line 640: why not use d.GetChange() instead? https://pkg.go.dev/github.com/hashicorp/terraform/helper/schema#ResourceData.GetChange.

pkg/resources/task.go
```diff
@@ -619,6 +635,21 @@ func UpdateTask(d *schema.ResourceData, meta interface{}) error {
...
+		flag := n.(bool)
+		if flag == false {
+			q = builder.UnsetAllowOverlappingExecution()
+		} else {
+			q = builder.ChangeAllowOverlappingExecution(flag)
```
sfc-gh-swinkler, line 645: since this function only gets called if flag is true, why not simplify to:
```
if !flag {
 q = builder.UnsetAllowOverlappingExecution()
} else {
q = builder.SetAllowOverlappingExecution()
}
```

sfc-gh-swinkler:
this looks fine but its failing the check-docs step. im just going to merge it and run the command myself, because i would like to get this in the next release. in the future, if you are changing anything in the examples folder or a schema for a resource you need to run `make docs`

sfc-gh-swinkler (approved):

sfc-gh-swinkler:
@mlorek thanks for your contribution!

PR #1292, opened 2022-10-20
fix: typo on GitHub bug template

froblesmartin:
Easy peasy 😄


sfc-gh-swinkler:
> Behavior is the preferred spelling in American English. Behaviour is preferred everywhere else.
this is not a typo. In America we just like to drop the "U" for some reason. I actually did not even know there was an alternative spelling of this word until just now.

Issue #1250, opened 2022-10-03
Add or_replace to task resources


PaulGolabowski:
We recently had a failed Terraform Apply due to the plugin timing out.  The apply was to create 2 new tasks (a root task and a dependent).  We usually fix the timing out issue by simply re-running the job from GitHub Actions.  This time, though, we got a Snowflake error saying that the dependent task already existed.  The only fix was to suspend the root task and drop the dependent via SYSADMIN.

What we'd like to happen is have the explicit "or_replace" option added to the Task resource so that it can be created as a "CREATE OR REPLACE" command instead of a pure "CREATE".  Ideally, this option should exist in every resource type that Snowflake supports "CREATE OR REPLACE" syntax.

An alternative would be an easier way to resolve conflicts like this from within Terraform.  Or to find a way to fix this plugin timed out bug.

sfc-gh-swinkler:
i am conflicted on this. on the one hand i want to support every feature that the Snowflake API does. On the other hand, this is not a very "Terraform" way of doing things. Terraform does not normally overwrite existing resources, its better to throw an error.

One thing that might be a better fix is to set partial state, although this could get complicated as it may require special customizeddiff functions to detect if the resource is partially created. 

Another possibility is to set the id immediately after creation, and before trying to resume the task. See here: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/task.go#L495-L518. I would move the line `d.SetId(dataIDInput)` immediately before `if enabled {`. That we even if there is a timeout when trying to resume the task, at least the ID will be set. And you shouldnt have this same problem.

Also, you said that you needed to manually resolve this conflict. Was it not possible to use `terraform import` in this case?

danu165:
> On the other hand, this is not a very "Terraform" way of doing things. Terraform does not normally overwrite existing resources, its better to throw an error.

For what it’s worth, the AWS provider actually always overwrites existing resources/configurations. Many people prefer this method because it ensures that if someone edits the resource outside of terraform then terraform will remedy it to its intended state. Implementing a provider this way or not is ultimately up to the provider manager, but I wouldn’t say it’s not a terraform way of doing things.

sfc-gh-jcieslak:
Hey @PaulGolabowski @danu165 
[...]

PR #1307, opened 2022-10-28
feat: task with or replace

mlorek:
added optional OR REPLACE option to the task resource - #1250
test included

## Test Plan
* [x] acceptance tests

## References
* https://docs.snowflake.com/en/sql-reference/sql/create-task.html#syntax


sfc-gh-swinkler:
Hello @mlorek!
i was working on refactoring this task resource as part of this ticket: https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1342

i believe the problem was that id was not being set until after waiting to resume the task. so instead i just switched the order. immediantly after the task is created the id will be set, then the waiting for starting the task will happen. This way you don't have the issue with the resource failing to persist to state.

in general i think that the `or_replace` functionality is not a good practice. despite it being part of the snowflake API, its very much anti-terraform best practices. If you have any issues in the future with this resource, please let me know.

Issue #1177, opened 2022-08-18
UNIQUE key for CREATE TABLE


yohei1126:
**Is your feature request related to a problem? Please describe.**

We want to use UNIQUE key for columns in CREATE TABLE statements 
https://docs.snowflake.com/en/sql-reference/constraints-overview.html

but can not find in current implementation 

https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/table.go
https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/table

**Describe the solution you'd like**

* add new boolean value `unique` to column in snowflake_table

```
resource "snowflake_table" "table" {
  database            = snowflake_schema.schema.database
  schema              = snowflake_schema.schema.name
  name                = "table"
  comment             = "A table."

  column {
    name     = "id"
    type     = "int"
    nullable = true
    unique = true
  }
```

**Additional context**

sfc-gh-asawicki:
Closing because of https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/table_constraint existence.

PR #1308, opened 2022-10-28
feat: column definition with unique constraint

mlorek:
column definition with an optional unique constraint option

## Test Plan
* [x] acceptance tests

## References
* #1177


sfc-gh-swinkler:
Table constraints should be created using the new table constraint resource. Please refer to snowflake_table_constraint resource https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/table_constraint

I don't want to overload the table resource with doing more than it already is at this point.

Issue #1068, opened 2022-06-21
snowflake_resource_monitor is missing the notify_users attribute


jrobison-sb:
**Provider Version**

```
$ terraform --version
Terraform v1.2.0
on darwin_arm64
+ provider registry.terraform.io/aidanmelen/snowsql v0.3.0
+ provider registry.terraform.io/chanzuckerberg/snowflake v0.25.36
```

**Describe the bug**

snowflake_resource_monitor is missing the notify_users attribute.

https://docs.snowflake.com/en/sql-reference/sql/create-resource-monitor.html#optional-parameters

**Expected behavior**

snowflake_resource_monitor has the the notify_users attribute.

**Code samples and commands**

I would expect the schema to support something like this:

```
resource "snowflake_resource_monitor" "monitor" {
  notify_users = [
    snowflake_user.foo.name,
  ]
  ...
}
```

preynacc:
The `feature-request` tag might be more appropriate, but this is also something that we would like to use.

nafarlee:
I can put a PR in for this in the near future (assuming there is enough interest)

miditha:
@nafarlee thanks for adding this feature.
I get `NOTIFY_USERS fields of account-level resource monitors must be null.` when try to use `notify_users` for an  account level resource monitor which can be done via Snow SQL.

nafarlee:
Hmm, I get the same thing in my testing environment, though the [Snowflake](https://docs.snowflake.com/en/sql-reference/sql/create-resource-monitor.html#usage-notes) [documentation](https://docs.snowflake.com/en/user-guide/resource-monitors.html#setting-a-resource-monitor-for-your-account) seems to imply it should work, and the SQL queries being generated are consistent with their recommendations:

```sql
CREATE RESOURCE MONITOR "PMFDSRXHCF" NOTIFY_USERS=('A', 'B');
ALTER ACCOUNT SET RESOURCE_MONITOR = "PMFDSRXHCF"; -- NOTIFY_USERS fields of account-level resource monitors must be null.
```

As I mentioned in that PR, I am far from a Snowflake expert, so I am not sure why this is causing an error (and I especially do not know why it works in Snow SQL). I hope someone else can shed light on this situation. It still works correctly when setting Resource Monitors at the Warehouse level.

On second thought @miditha, it might be better to open a separate bug report, as the scope of this issue was the initial introduction of the feature and should likely be closed.

[...]

PR #1340, opened 2022-11-03
feat: add support for `notify_users` to `snowflake_resource_monitor` resource

nafarlee:
This PR adds support for managing the `notify_users` attribute on resource monitors via the `snowflake_resource_monitor` Resource. It does not include any changes to the `snowflake_resource_monitors` Data Source.

## Test Plan
- [x] internal tests
- [x] unit tests
- [ ] acceptance tests

All new and existing tests pass, though one new test is a bit special and worth calling out explicitly.

The new test in `pkg/resources/resource_monitor_acceptance_test.go` is particularly tricky, as the users listed must have a verified email address. As far as I know, this verification process may only be completed manually via the Snowflake interface. To circumvent this, that test depends on a new `RESOURCE_MONITOR_NOTIFY_USERS_TEST` environment variable. If the variable is unset, or set to the empty string, the test will be skipped (similar to the `SKIP_*` variable for many other acceptance tests). However, if the variable is set, it is expected to be a comma-delimited list of users that already exist with verified email addresses. Open to suggestions on how this can be improved.

## References
- https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1068
- https://docs.snowflake.com/en/sql-reference/sql/create-resource-monitor.html#optional-parameters

### DIFF
--- a/docs/resources/resource_monitor.md
+++ b/docs/resources/resource_monitor.md
@@ -24,6 +24,8 @@ resource snowflake_resource_monitor monitor {
   notify_triggers            = [40]
   suspend_triggers           = [50]
   suspend_immediate_triggers = [90]
+
+  notify_users = ["USERONE", "USERTWO"]
 }
 ```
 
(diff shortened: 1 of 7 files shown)


sfc-gh-swinkler (approved):

sfc-gh-swinkler:
@nafarlee this looks good. Thank you for your contribution!

Without a better SDK we don't really have a way to validate emails easily.

I am planning to rethink some of these environment variables since they are starting to clutter up the test cases. I will eventually go through and redo these in a different way. Maybe have a .env file with default values or detect whether the account is standard of enterrpise during provider initialization and enable / disable tests as a result.

nafarlee:
@sfc-gh-swinkler You are welcome! I am far from experienced with Snowflake, so I don't have a good sense for how difficult/inconsistent automatic detection would be. Personally, a `.env` that centralizes the options and their individual "considerations" would be a helpful stopgap

PR #1378, opened 2022-11-20
feat(ci): add goimports and makezero

mnagaa:
This PR adds new linters `goimports` and `makezero`. The current code meets the condition to pass these linters so there is no change here.

In the future, I will try to add more linters to maintain the quality of codes👍


sfc-gh-swinkler:
@mnagaa thank you!

PR #1379, opened 2022-11-20
fix: wrong usage of testify Equal() function

mnagaa:
## Summary

This PR fixes the wrong usages of testify package.

Function arguments of `Equal()`  from the `testify` package should be first `expected` and second `actual` values. The actual values are often the function's return to be tested. However, many tests with `Equal()` in this repo are incorrect.

The definition is as follows:

```
func (a *Assertions) Equal(expected interface{}, actual interface{}, msgAndArgs ...interface{}) {
	if h, ok := a.t.(tHelper); ok {
		h.Helper()
	}
	Equal(a.t, expected, actual, msgAndArgs...)
}
```

https://pkg.go.dev/github.com/stretchr/testify/require#Assertions.Equal

This leads to showing the unexpected verbose output when the test fails.


sfc-gh-swinkler:
@mnagaa thank you!

PR #1382, opened 2022-11-22
Remove double quotes from external table column definition

spcvalente:
this is inconsistent with the partition_by field, for example, this generated SQL:

``` sql
CREATE EXTERNAL TABLE "x"."y"."z" ("partition_date" date AS '2020-01-01')) PARTITION BY ( partition_date ) WITH LOCATION ...
```

it throws an error because `"partition_date"` is not the same as `partition_date`


spcvalente:
ℹ️ upper case on column names solves the issue

PR #1386, opened 2022-11-23
feat: Added (missing) API Key to API Integration

ChrisIsidora:
Added (missing) API Key property to the API Integration

## Test Plan
* [x] acceptance tests
* [x] unit tests

## References
* https://docs.snowflake.com/en/sql-reference/sql/create-api-integration.html

### DIFF
--- a/pkg/resources/api_integration.go
+++ b/pkg/resources/api_integration.go
@@ -77,6 +77,12 @@ var apiIntegrationSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Description: "Lists the endpoints and resources in the HTTPS proxy service that are not allowed to be called from Snowflake.",
 	},
+	"api_key": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Default:     "",
+		Description: "The API key (also called a “subscription key”).",
+	},
 	"enabled": {
 		Type:        schema.TypeBool,
 		Optional:    true,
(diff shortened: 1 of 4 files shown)


ChrisIsidora:
@sfc-gh-swinkler Can you approve, merge and release? Thanks in advance :-)

pkg/resources/api_integration.go
```diff
@@ -77,6 +77,12 @@ var apiIntegrationSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Description: "Lists the endpoints and resources in the HTTPS proxy service that are not allowed to be called from Snowflake.",
 	},
+	"api_key": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Default:     "",
```
sfc-gh-swinkler, line 83: if this is optional you don't need to set a default value to an empty string. instead you can rely on the `d.GetOk()` function instead.
```
if _, ok := d.GetOk("api_key);ok {
 // api_key was set
} 
// else no need to do anything
```

ChrisIsidora: @sfc-gh-swinkler Great catch, I made the changes. Thanks for reviewing

ChrisIsidora:
> Hello @ChrisIsidora this looks good. If you could just make a minor change, i would greatly appreciate.
@sfc-gh-swinkler I pushed the suggested minor change. Thanks!

sfc-gh-swinkler (approved):

sfc-gh-swinkler:
@ChrisIsidora thank you for your contribution!

PR #1402, opened 2022-12-07
fix: Removed Read for API_KEY

ChrisIsidora:
Fixed Integration Tests for API Integration implementation of API_KEY introduced with #1386 

## Test Plan
* [x] Acceptance tests
* [x] Unit tests

## References
*  #1386 
* #1401

### DIFF
--- a/pkg/resources/api_integration.go
+++ b/pkg/resources/api_integration.go
@@ -80,6 +80,7 @@ var apiIntegrationSchema = map[string]*schema.Schema{
 	"api_key": {
 		Type:        schema.TypeString,
 		Optional:    true,
+		Sensitive:   true,
 		Description: "The API key (also called a “subscription key”).",
 	},
 	"enabled": {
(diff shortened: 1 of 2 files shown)


pkg/resources/api_integration.go
```diff
@@ -212,10 +213,6 @@ func ReadAPIIntegration(d *schema.ResourceData, meta interface{}) error {
 					return err
 				}
 			}
-		case "API_KEY":
```
sfc-gh-swinkler, line 215: why is this being removed?

ChrisIsidora: @sfc-gh-swinkler  API_KEY is a sensitive value in Snowflake, so we can't Read it (back) with Terraform. 
* See also the user resource which also omits the password property from the Read. 
* Integration Test error: https://github.com/Snowflake-Labs/terraform-provider-snowflake/actions/runs/3641738297/jobs/6148030735#step:9:284

sfc-gh-swinkler: thank you for the clarification. approved

PR #1410, opened 2022-12-08
fix: go syntax

mnagaa:
This PR fixes lines to use better Go syntax.
- increment-decrement: should replace found += 1 with found++
- removing underscore

```
pkg/snowflake/table.go:118:2: var-naming: don't use underscores in Go names; struct field masking_policy should be maskingPolicy (revive)
        masking_policy string

```


sfc-gh-swinkler (approved):

sfc-gh-swinkler:
@mnagaa thanks again for your contribution!

Issue #1412, opened 2022-12-09
Support for Stream On Stage


vgw-parham:
**Is your feature request related to a problem? Please describe.**

We are trying to create a Stream resource on a Stage. However, the way the the create syntax for it works, is currently not supported by the Snowflake Terraform provider.

**Describe the solution you'd like**

To add support for a `on_stage` field on the `snowflake_stream` resource which then references the correct stage directory table.

**Describe alternatives you've considered**

The current alternative for us right now is to use either a manual query directly through the snowflake worksheet or run it through something like Flyway.

PR #1413, opened 2022-12-09
feat: add ON STAGE support for Stream resource

vgw-parham:
This PR adds support for Streams to work with Stages.

## Test Plan
I've just used the `make test` command locally

* [ ] acceptance tests

* [ ] compatibility with https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/stream.go#L257
* [ ] compatibility with https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/snowflake/stream.go#L150

^ I wasnt sure on whether those 2 functions needed to be updated or not, so left it for now.

## References
Full disclosure: this is my first time contributing to a terraform provider project, so I'm not 100% sure if I've changed all the correct things. If I've missed anything, let me know and I'll add it in.

* https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1412

### DIFF
--- a/pkg/resources/stream.go
+++ b/pkg/resources/stream.go
@@ -47,14 +47,21 @@ var streamSchema = map[string]*schema.Schema{
 		Optional:     true,
 		ForceNew:     true,
 		Description:  "Name of the table the stream will monitor.",
-		ExactlyOneOf: []string{"on_table", "on_view"},
+		ExactlyOneOf: []string{"on_table", "on_view", "on_stage"},
 	},
 	"on_view": {
 		Type:         schema.TypeString,
 		Optional:     true,
 		ForceNew:     true,
 		Description:  "Name of the view the stream will monitor.",
-		ExactlyOneOf: []string{"on_table", "on_view"},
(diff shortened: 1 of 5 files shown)


sfc-gh-swinkler (changes requested):
Looking at this resource, i would like to redo this eventually. all this `on_xxx` stuff is not really the way we want to be doing things going forward. But since this has already been done for this resource, i will allow it.

FYI the pattern i have been using for new resources would use `object_identifier` and `object_type`.

For example,

```
resource "snowflake_stream" "s" {
object_type = "" // one of TABLE, EXTERNAL TABLE, VIEW, STAGE
object_identifier = "" (removing the need for schema and database to be set)
 
// all optional parameters next
append_only = true /false
comment = ""
...
}
```
As for right now, as long as you can make the requested changes then i will merge.

vgw-parham:
@sfc-gh-swinkler thanks for the review! I've updated the code as requested

Issue #1310, opened 2022-10-28
Updating the credit_quota in snowflake_resource_monitor breaks the resource monitor


bdumford-OM1:
**Provider Version**

0.40.0

**Terraform Version**

0.13.4

**Describe the bug**

Updating the credit quota in "snowflake_resource_monitor" causes the resource monitor to be replaced such that it is no longer attached to the warehouse it is monitoring.  Additionally, any monitoring grants on the resource monitor provider to roles are lost.

```
# module.sample_warehouse.snowflake_resource_monitor.warehouse_monitor must be replaced
-/+ resource "snowflake_resource_monitor" "warehouse_monitor" {
!       credit_quota               = 1500 -> 1000 # forces replacement
        frequency                  = "MONTHLY"
!       id                         = “SAMPLE_MONITOR" -> (known after apply)
        name                       = “SAMPLE_MONITOR"
        notify_triggers            = []
        set_for_account            = false
!       start_timestamp            = "2022-10-01T00:00:00Z" -> "2022-11-01T00:00:00Z"
-       suspend_immediate_triggers = [] -> null
-       suspend_triggers           = [] -> null
    }
```
Workaround: A subsequent plan and apply will fix both of these issues
```
# module.sample_warehouse.snowflake_resource_monitor_grant.warehouse_monitor_grants must be replaced
-/+ resource "snowflake_resource_monitor_grant" "warehouse_monitor_grants" {
        enable_multiple_grants = false
!       id                     = “SAMPLE_MONITOR" -> (known after apply)
        monitor_name           = "SAMPLE_MONITOR"
        privilege              = "MONITOR"
!       roles                  = [ # forces replacement
+           “ROLE1”,
+           "ROLE2”,
        ]
        with_grant_option      = false
    }

  # module.sample_warehouse.snowflake_warehouse.warehouse will be updated in-place
! resource "snowflake_warehouse" "warehouse" {
        auto_resume                         = true
        auto_suspend                        = 300
        id                                  = “SAMPLE”
        initially_suspended                 = true
        max_cluster_count                   = 1
        max_concurrency_level               = 1
        min_cluster_count                   = 1
        name                                = “SAMPLE”
!       resource_monitor                    = "null" -> “SAMPLE_MONITOR"
        scaling_policy                      = "STANDARD"
        statement_queued_timeout_in_seconds = 0
        statement_timeout_in_seconds        = 14400
        warehouse_size                      = "X-Large"
    }

[...]

Issue #876, opened 2022-02-24
Alter resource monitor is actually destroy plus create


kumarjoshi:
**Provider Version**

0.25.36

**Terraform Version**
v1.1.6

The version of Terraform you were using when the bug was encountered.
v1.1.6

**Describe the bug**
I create a resource monitor `EVENTS_DEVELOPER_MONITOR` using ACCOUNTADMIN role and then granted all privileges on this monitor to the role DATA_ADMIN, which is used to run terraform.  That means the role DATA_ADMIN should have MODIFY and MONITOR privileges according to :
https://docs.snowflake.com/en/user-guide/security-access-control-privileges.html#resource-monitor-privileges
GRANT CREATE MONITOR is currently not supported by Snowflake.

When I try to alter one of the properties of this resource monitor using Terraform, it gives me an error:
```
│ Error: error deleting resource monitor EVENTS_DEVELOPER_MONITOR: 003001 (42501): SQL access control error:
│ Insufficient privileges to operate on resource_monitor 'EVENTS_DEVELOPER_MONITOR'
```
That means, `terraform apply` is not modifying the resource monitor but instead deleting and re-creating it. 

**Expected behavior**
With role `DATA_ADMIN` I should be able to modify existing resource monitors since I granted ALL privileges on this monitor to role DATA_ADMIN

**Code samples and commands**
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # snowflake_resource_monitor.events_developer_monitor must be replaced
-/+ resource "snowflake_resource_monitor" "events_developer_monitor" {
      ~ credit_quota               = 10000 -> 8000 # forces replacement
      ~ id                         = "EVENTS_DEVELOPER_MONITOR" -> (known after apply)
        name                       = "EVENTS_DEVELOPER_MONITOR"
      - suspend_immediate_triggers = [] -> null
      - suspend_triggers           = [] -> null
        # (4 unchanged attributes hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.
snowflake_resource_monitor.events_developer_monitor: Destroying... [id=EVENTS_DEVELOPER_MONITOR]
╷
│ Error: error deleting resource monitor EVENTS_DEVELOPER_MONITOR: 003001 (42501): SQL access control error:
[...]

PR #1456, opened 2023-01-05
feat: add in more functionality for UpdateResourceMonitor 

sfc-gh-jlove:
resource_monitor now can update without creation and delation for 

- notify_users
- suspend_triggers
- notify_triggers
- set_for_account
- warehouses

Also fixed a typo in a comment implents to implements

## Test Plan

* [x] acceptance tests

## References

- https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1310
- https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/876

### DIFF
--- a/pkg/resources/resource_monitor.go
+++ b/pkg/resources/resource_monitor.go
@@ -287,22 +287,22 @@ func inSlice(n string, h []string) bool {
 	return false
 }
 
-// convertTerraformSetToStringSlice turns a terraform set into a string slice
+// convertTerraformSetToStringSlice turns a terraform set into a string slice.
 func convertTerraformSetToStringSlice(i interface{}) []string {
-	var s []string
-	for _, w := range i.(*schema.Set).List() {
-		s = append(s, w.(string))
+	s := make([]string, len(i.(*schema.Set).List()))
+	for x, w := range i.(*schema.Set).List() {
+		s[x] = w.(string)
 	}
 	return s
 }
 
 // compareTerraformSets compares two terraform sets and returns a string slice of all values in the first set that is
 // not in the second set.
-func compareTerraformSets(firstSet interface{}, SecondSet interface{}) []string {
+func compareTerraformSets(firstSet interface{}, secondSet interface{}) []string {
 	res := make([]string, 0)
 	sliceOne := convertTerraformSetToStringSlice(firstSet)
 
-	sliceTwo := convertTerraformSetToStringSlice(SecondSet)
+	sliceTwo := convertTerraformSetToStringSlice(secondSet)
 
 	for _, s := range sliceOne {
 		if !inSlice(s, sliceTwo) {
@@ -369,7 +369,7 @@ func UpdateResourceMonitor(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	// Remove from account
-	if d.HasChange("set_for_account") && d.Get("set_for_account").(bool) == false {
+	if d.HasChange("set_for_account") && !d.Get("set_for_account").(bool) {
 		if err := snowflake.Exec(db, stmt.UnsetOnAccount()); err != nil {
 			return fmt.Errorf("error unsetting resource monitor %v on account err = %w", id, err)
 		}
--- a/pkg/resources/resource_monitor.go
+++ b/pkg/resources/resource_monitor.go
@@ -25,7 +25,6 @@ var resourceMonitorSchema = map[string]*schema.Schema{
 	"notify_users": {
 		Type:        schema.TypeSet,
 		Optional:    true,
-		ForceNew:    true,
 		Description: "Specifies the list of users to receive email notifications on resource monitors.",
 		Elem: &schema.Schema{
 			Type: schema.TypeString,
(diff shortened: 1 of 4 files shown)


pkg/resources/resource_monitor.go
```diff
@@ -283,13 +277,54 @@ func extractTriggerInts(s sql.NullString) ([]int, error) {
 	return out, nil
 }
 
+// inSlice returns true if n is in string slice h otherwise false.
+func inSlice(n string, h []string) bool {
```
sfc-gh-swinkler, line 281: there is a slices.Contains() function you should use instead of rolling your own function https://pkg.go.dev/golang.org/x/exp/slices#Contains

pkg/resources/resource_monitor.go
```diff
@@ -283,13 +277,54 @@ func extractTriggerInts(s sql.NullString) ([]int, error) {
...
+	}
+	return res
+}
+
+// UpdateResourceMonitor implements schema.UpdateFunc.
 func UpdateResourceMonitor(d *schema.ResourceData, meta interface{}) error {
 	db := meta.(*sql.DB)
 	id := d.Id()
 
 	stmt := snowflake.NewResourceMonitorBuilder(id).Alter()
```
sfc-gh-swinkler, line 320: this was a bit confusing since its not actually the statement (which would be a string), but the alterBuilder instead. This should be renamed for clarity

pkg/resources/resource_monitor.go
```diff
@@ -310,12 +345,65 @@ func UpdateResourceMonitor(d *schema.ResourceData, meta interface{}) error {
...
+	}
+
 	if runSetStatement {
 		if err := snowflake.Exec(db, stmt.Statement()); err != nil {
 			return fmt.Errorf("error updating resource monitor %v\n%w", id, err)
 		}
 	}
 
+	// Remove from account
+	if d.HasChange("set_for_account") && !d.Get("set_for_account").(bool) {
```
sfc-gh-swinkler, line 372: this could be combined with the add to account. something like

```
if d.HasChange("set_for_account")
  old, new := d.GetChange("set_for_account")
 if new == nil {
  // remove from account
} else {
 // set on new account
}
f d.Get("set_
```

pkg/resources/resource_monitor.go
```diff
@@ -283,13 +277,54 @@ func extractTriggerInts(s sql.NullString) ([]int, error) {
...
+	s := make([]string, len(i.(*schema.Set).List()))
+	for x, w := range i.(*schema.Set).List() {
+		s[x] = w.(string)
+	}
+	return s
+}
+
+// compareTerraformSets compares two terraform sets and returns a string slice of all values in the first set that is
+// not in the second set.
+func compareTerraformSets(firstSet interface{}, secondSet interface{}) []string {
```
sfc-gh-swinkler, line 301: firstly, helper functions like this should be grouped with the other helper functions. they are spread out a bit now, but i think we should start putting them together in the helpers package https://github.com/Snowflake-Labs/terraform-provider-snowflake/tree/main/pkg/helpers.

second, this function is unclear in arguments. The input arguments are both generic interface type, but the result is a slice of string. I think if you want to deal with slices, then deal with slices. Don't do this extra conversion behind the scenes. 

Finally, the function is unclear in name. What is it exactly that you are comparing? I dont think you are actually comparing anything, what you are doing is finding either the union, intersection or difference.

pkg/resources/resource_monitor.go
```diff
@@ -283,13 +277,54 @@ func extractTriggerInts(s sql.NullString) ([]int, error) {
...
+	for _, v := range h {
+		if v == n {
+			return true
+		}
+	}
+	return false
+}
+
+// convertTerraformSetToStringSlice turns a terraform set into a string slice.
+func convertTerraformSetToStringSlice(i interface{}) []string {
```
sfc-gh-swinkler, line 291: the expandStringList() function already does this functionality so we actually don't need this function

 Also there doesn't seem to be any benefit in having an interface{} argument. Just accept `*schema.Set` since its easy to cast and you also make it explicitly clear by the name that it only accepts terraform sets.

pkg/snowflake/resource_monitor.go
```diff
@@ -44,6 +44,14 @@ type ResourceMonitorCreateBuilder struct {
...
+type ResourceMonitorAlterBuilder struct {
+	AlterPropertiesBuilder
+
+	// triggers consist of the type (DO SUSPEND | SUSPEND_IMMEDIATE | NOTIFY) and
+	// the threshold (a percentage value)
+	triggers []trigger
+}
+
 type trigger struct {
 	action    string
```
sfc-gh-swinkler, line 56: should action be an enum since it can only be one of three values?

sfc-gh-swinkler (changes requested):

pkg/resources/resource_monitor.go
sfc-gh-jlove: This was intentional the issue is you can't have it set for both an account and a warehouse. So if you are going from setting a resource monitor from the account level to the warehouse level or vice versa  you have to make sure it is unset first.

sfc-gh-jlove:
@sfc-gh-swinkler Addressed all your feedback.

pkg/resources/resource_monitor.go
sfc-gh-terraform-sa: in that case you should also use"ConflictsWith", it works like setting ForceNew on an attirbute. Will prevent people from setting both at the same time.

```
// ConflictsWith is a set of schema keys that conflict with this schema.
// This will only check that they're set in the _config_. This will not
// raise an error for a malfunctioning resource that sets a conflicting
// key.
ConflictsWith []string
```

sfc-gh-terraform-sa (changes requested):

sfc-gh-swinkler (approved):

Issue #1343, opened 2022-11-04
Create Warehouse Fails due to new parameter


sfc-gh-cdube:
**Provider Version**

v0.49.0

**Terraform Version**

>= 1.1.7

**Describe the bug**

When creating a ressource warehouse, the command passed to Snowflake adds the optional parameter query_acceleration_max_scale_factor, even though this is not a parameter in the ressource code. Since this is a parameter that needs enable_query_acceleration = True to works, this fails the statement. The Enable_query_acceleration is not even deployed in my region as we speak.

**Expected behavior**

Pass all of the default, which would pass the enable_query_acceleration to False, hence will not fail the statement.

asnodgrass-nid:
```
│ Error: error altering warehouse: 001423 (22023): SQL compilation error:
│ invalid property 'QUERY_ACCELERATION_MAX_SCALE_FACTOR'; feature 'Query Acceleration Service' not enabled
```

The workaround for me is to explicitly set `query_acceleration_max_scale_factor = 0` on all my `snowflake_warehouse` resources.

sfc-gh-cdube:
I still needed to add enable_query_acceleration = False for it to work. and per documentation:
`Setting the QUERY_ACCELERATION_MAX_SCALE_FACTOR to 0 eliminates the limit and allows queries to lease as many resources as necessary and as available to service the query.`

asnodgrass-nid:
It was enough in my case to just set that one property. The [PR](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1239) I found that added this change seems to set the boolean flag to `false` by default.

asnodgrass-nid:
Given your region comment in the original post, it might be that a full workaround will vary depending on whether the preview is available in the target region.

mnagaa:
I also encountered the same issue in the same provider version v0.49 of this thread, which caused failures in creating all warehouses. We currently use the standard edition for Snowflake. 
Perhaps, this came from the standard edition, but any warehouse without new accelerator features should be created for any edition.
If you find a workaround for this, please let me know.

The error case is as follows (with the standard edition):

```
resource "snowflake_warehouse" "XXX" {
  provider = snowflake.sys_admin
  name     = XXX
  comment  = "XXX"

  warehouse_size               = "small"
  auto_resume                  = true
  auto_suspend                 = 60
  statement_timeout_in_seconds = 3600
  initially_suspended          = true
  enable_query_acceleration    = false
[...]

PR #1474, opened 2023-01-12
fix: dont set query_acceleration_max_scale_factor when not enabled

mnagaa:
One of my snowflake accounts is using the Standard edition.
Recently, a functionality for query acceleration as the warehouse parameter has been implemented in Snowflake.

For standard editions, the warehouse parameter `enable_query_acceleration` is always set to false because this acceleration can be used for Enterprise edition users. 

The following queries are invalid for  Standard user and raises error like attached image of snowflake console.
`alter warehouse WAREHOUSE_NAME set query_acceleration_max_scale_factor = 0;`
`alter warehouse WAREHOUSE_NAME set query_acceleration_max_scale_factor = 8;`

As a workaround for this, I change lines to alter `query_acceleration_max_scale_factor` only if enable_query_acceleration is true.

I am not sure the implementation around this functionality, please review carefully.


mnagaa:
Related to #1343

sfc-gh-swinkler:
This seems like a reasonable solution. If you can't read the value with Standard edition, then don't bother even trying.

sfc-gh-swinkler (approved):

belacevic:
This, unfortunately doesn't seem to have helped with creating new warehouses, as Terraform still adds the default value `8` to the (non supported) attribute, even when setting:

```
enable_query_acceleration = false
```

and 

```
  lifecycle {
    ignore_changes = [
      query_acceleration_max_scale_factor
    ]
  }
```

forest:
Duplicate: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1944

PR #1476, opened 2023-01-13
Update system_get_privatelink_config.md

jreslock:
`zone_id` value for the `snowflake_private_link_ocsp_url` is incorrect.

The value should be `aws_route53_zone.snowflake_private_link.zone_id` based on the resources configured in this example.

## Test Plan
N/A this is just a tiny documentation update.


sfc-gh-swinkler:
@jreslock the files in the docs folder are generated with the `make docs` command and will get overwritten if you write to them directly. Suggest to update the example and then run `make docs`. Here is the file that needs to be changed: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/examples/data-sources/snowflake_system_get_privatelink_config/data-source.tf#L47

PR #1522, opened 2023-02-06
fix: cleanup grant logic

sfc-gh-swinkler:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


yinxu0619:
Hi there, I do believe snowflake_warehouse_grant has the same problem as when applying the tf I can see it grants privileges on the warehouse to the role over and over... even it didn't cause any problem but I hope we can fix this issue.

Issue #1538, opened 2023-02-16
snowflake_resource_monitor parameters wrongly configured


froblesmartin:
**Provider Version**

v0.56.3

**Terraform Version**

v1.3.7

**Describe the bug**

The [snowflake_resource_monitor resource](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/resource_monitor) has two arguments configured as `Set of Number`, while those can be only a single Number:
* suspend_immediate_triggers
* suspend_triggers

If you put more than one number, like `[10, 20]`, the `terraform apply` returns:
> ... err = 090251 (42601): A resource monitor can have at most one suspend_immediate trigger.

**Expected behavior**

The parameter expects just a Number

PR #1545, opened 2023-02-17
fix: change resource monitor suspend properties to number

froblesmartin:
Fixing https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1538

Both the `suspend trigger` and `suspend immediate trigger` properties for a resource monitor are just a single number, not a set of numbers.

Also adding missing information for importing the resource monitor Terraform resource.

Also adding a note for `set_for_account` property, as it defaults to false.

NOTE: Important to notice that due to the change of the properties (removing the s) this is a breaking change

## Test Plan

* [ ] acceptance tests

## References
https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/resource_monitor
https://docs.snowflake.com/en/sql-reference/sql/create-resource-monitor

### DIFF
--- a/docs/resources/resource_monitor.md
+++ b/docs/resources/resource_monitor.md
@@ -21,9 +21,9 @@ resource "snowflake_resource_monitor" "monitor" {
   start_timestamp = "2020-12-07 00:00"
   end_timestamp   = "2021-12-07 00:00"
 
-  notify_triggers            = [40]
-  suspend_triggers           = [50]
-  suspend_immediate_triggers = [90]
+  notify_triggers            = [40, 50]
+  suspend_triggers           = 50
+  suspend_immediate_triggers = 90
 
   notify_users = ["USERONE", "USERTWO"]
 }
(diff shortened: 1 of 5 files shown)


froblesmartin:
Important to notice that due to the change of the properties (removing the `s`) this is a breaking change

sfc-gh-swinkler:
@froblesmartin for breaking changes like this it is important to mark the old attribute as deprecated, while continuing to support both methods until a major release would warrant such a change. See: https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-attribute-removal.

Also now that i am looking at this resource, i realize it would be better if we instead had a triggers block. e.g. something like:

```
resource "snowflake_resource_monitor" "monitor" {
  name         = "monitor"
  credit_quota = 100

  frequency       = "DAILY"
  start_timestamp = "2020-12-07 00:00"
  end_timestamp   = "2021-12-07 00:00"

  trigger {
    threshold_percent = <threshold>
    do  = { SUSPEND | SUSPEND_IMMEDIATE | NOTIFY }
  }  

  notify_users = ["USERONE", "USERTWO"]
}
```

froblesmartin:
Sure thing @sfc-gh-swinkler, I leave it up to you to include this change in the next major release. I do not know if you are planning to do so in the near future. If that is not the case, I could try to spend some time on changing this PR to keep working in both the old and new way, but also it would be nice to know if that `trigger` block is actually what you want to go for. 😄

Issue #1529, opened 2023-02-10
Refresh of database schema fails after deleting a database manually


Relativity74205:
**Provider Version**

0.56.3

**Terraform Version**

Terraform v1.2.9

**Describe the bug**

When a database with at least a schema is manually deleted in Snowflake, terraform refresh for the schema object fails with 
```
snowflake_schema.foo: Refreshing state... [id=FOO_DATABASE|FOO_SCHEMA]
╷
│ Error: 002043 (02000): SQL compilation error:
│ Object does not exist, or operation cannot be performed.
│ 
│   with snowflake_schema.foo,
│   on database.tf line 8, in resource "snowflake_schema" "foo":
│    8: resource "snowflake_schema" "foo" {
```

**Expected behavior**

A `terraform refresh` should delete the database and all its objects from the terraform state.

**Code samples and commands**

Setup with terraform:
```
resource "snowflake_database" "foo" {
  provider = snowflake.sys
  name = "FOO_DATABASE"
}

resource "snowflake_schema" "foo" {
  provider = snowflake.sys
  database = snowflake_database.foo.name
  name = "FOO_SCHEMA"
}
```
Drop the "FOO_DATABASE" object and run `terraform plan` or `terraform refresh`

**Additional context**

Relativity74205:
Closed by #1568

PR #1568, opened 2023-02-23
fix: schema read now checks first if the corresponding database exists

Relativity74205:
Fixes the problem when a schema refresh is performed, when the schema and the corresponding database had been deleted in snowflake.

## Test Plan
* [x] unit/internal tests
* [x] manual tests with a snowflake environment

## References
* https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1529

PS: I have the impression that similar problems might exist with other resources. I might have a look on this as the next step.


sfc-gh-swinkler (approved):

sfc-gh-swinkler:
@Relativity74205 thank you for your contribution!

Issue #1590, opened 2023-03-01
Renaming role of role_grants resource is causing an error


Relativity74205:
**Provider Version**

0.57.0

**Terraform Version**

1.2.9

**Describe the bug**

When renaming the role which is granted to other roles with the role_grants resource, an error is thrown:

`│ Error: 002003 (02000): SQL compilation error:
│ Role 'FOO_TEST2' does not exist or not authorized.
│ 
│   with snowflake_role_grants.foo_to_bar,
│   on role_test.tf line 9, in resource "snowflake_role_grants" "foo_to_bar":
│    9: resource "snowflake_role_grants" "foo_to_bar" {
`

**Expected behavior**

Role is renamed without error.

**Code samples and commands**

Setup:
`
resource "snowflake_role" "foo_test" {
    name = "FOO_TEST2"
}

resource "snowflake_role" "bar_test" {
    name = "BAR_TEST2"
}

resource "snowflake_role_grants" "foo_to_bar" {
    role_name = snowflake_role.foo_test.name
    
    roles = [snowflake_role.bar_test.name]
}
`

Change name of snowflake_role.foo_test to e.g. "FOO_TEST".

**Additional context**

The problem is that the role_grants resource is not dropped before the role is renamed. A workaround is using a lifecycle policy, which forces the drop of the role_grants, when the name of the role is changed.

Relativity74205:
I haved created a PR for this issue: #1591

PR #1591, opened 2023-03-01
fix: added force_new option to role grant when the role_name has been changed

Relativity74205:
- role_grants resource is now recreated if the role_name is changed
- when refreshing the resource, the code checks now first, if the grantee role still exists (in case it was manually deleted in snowflake)

## Test Plan
* [x] unit tests
* [x] manuel tests
* [x] acceptance tests of role_grants module

## References

Issue is described in #1590

### DIFF
--- a/pkg/resources/role_grants.go
+++ b/pkg/resources/role_grants.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"database/sql"
+	"errors"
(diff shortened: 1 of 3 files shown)


sfc-gh-swinkler (approved):

sfc-gh-swinkler:
looks good, thanks!

PR #1597, opened 2023-03-02
fix: 

Relativity74205:
- readded imported privileges special case for database grants

It was removed in #1522, however it is needed for the edge case of the shared databases as the snowflake database. In order to give custom roles the permission to use the shared databases, the privilege must be given "IMPORTED PRIVILEGES". However, in the information_schema tables the privilege is shown as "USAGE".

Without this code snippet, the terraform provider wants to regive this permission every time.

## Test Plan
* [x] unit tests
* [x] manual test with snowflake

## References
(https://docs.snowflake.com/en/user-guide/data-exchange-marketplace-privileges)

### DIFF
--- a/pkg/resources/database_grant.go
+++ b/pkg/resources/database_grant.go
@@ -123,6 +123,13 @@ func ReadDatabaseGrant(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	// IMPORTED PRIVILEGES is not a real resource, so we can't actually verify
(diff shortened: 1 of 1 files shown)


Relativity74205:
@sfc-gh-swinkler I would be great, if this small fix could be included in the next release as this bug is quite annoying ;)

Relativity74205:
@sfc-gh-swinkler Any chance, if this fix could be merged? The bug is a little anoying

sfc-gh-swinkler:
@Relativity74205. Thank you, i didn't understand why this was in there previously.

Issue #1609, opened 2023-03-08
null_if value in file_format cannot be set to ''


Lesly-prem:
**Provider Version**

The provider version you are using.
0.41.0

**Expected behavior**

**Code samples and commands**

Please add code examples and commands that were run to cause the problem.

**Additional context**

sfc-gh-jmichalak:
Hi @Lesly-prem  👋 

We've just released v2.19.0 ([release notes](https://github.com/snowflakedb/terraform-provider-snowflake/releases/tag/v2.19.0), [migration guide](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v218x--v2190)). In this version, we introduce new file formats resources and deprecate the old `snowflake_file_format`. Enjoy!

sfc-gh-jmichalak:
Closing due to inactivity.

Issue #708, opened 2021-10-07
null_if value in file_format cannot be set to ''


syldor:
**Provider Version**

0.25.19

**Terraform Version**

1.0.3

**Describe the bug**

It's impossible to create a file_format with a `null_if` that compiles in snowflake to

```
"NULL_IF":[""]
```
It should be possible with:

```
null_if                        = [""]
``` 

but this gives in Snowflake

```
"NULL_IF":[]
```

In the code of the provider here: https://github.com/chanzuckerberg/terraform-provider-snowflake/blob/e3e412942c6706942b5dd21af8ea51658ea0ff5b/pkg/snowflake/file_format.go#L339

Should we change the default to be 
```
		q.WriteString(` NULL_IF = ('')`)
```
as it's the one to consider an empty value as null?

adharshdlx:
Thanks for raising this. I am having the same issue. Would like to see a fix for this please.

caiolauro:
Would also like to see a fix for this

mattthaber:
also experiencing this, and saddened to see zero address of a bug after so many months

sfc-gh-swinkler:
@syldor 
I have tested this in the latest version (0.44.0) and this does not seem to be an issue anymore. My example terraform configuration is:

```
resource "snowflake_file_format" "example_file_format" {
  name        = "EXAMPLE_FILE_FORMAT"
  database    = snowflake_database.test.name
  schema      = snowflake_schema.test.name
  format_type = "CSV"
  null_if                        = [""]
}
``` 

and the resulting query in Snowflake is:

```
CREATE FILE FORMAT "test"."test"."EXAMPLE_FILE_FORMAT" TYPE = 'CSV' NULL_IF = () SKIP_BLANK_LINES = false TRIM_SPACE = false ERROR_ON_COLUMN_COUNT_MISMATCH = false REPLACE_INVALID_CHARACTERS = false EMPTY_FIELD_AS_NULL = false SKIP_BYTE_ORDER_MARK = false
```

Which, as you can see  the value is correctly set to `NULL_IF = ()`

If you can please upgrade to the latest version, that should resolve your issue.

yo0824:
I think the correct value is `NULL_IF = ('')` , not `NULL_IF = ()`.

Lesly-prem:
Whether Still this issue exist ,since i am facing still the issue ..its going as NULL_IF = () for ""

PR #1617, opened 2023-03-14
feat[file format]: allow empty string for null_if argument

tmlp-lv:
## Summary

The Snowflake Terraform provider handles file formats, but does not allow empty string to be added in the `null_if` directive. This PR aims to fix so that the go string `""` is not ignored anymore, and can be handled by Terraform without having to use the `ignore_changes` lifecycle option. This was solved by adding a new expansion function in `pkg/resources/helper_expansion.go` called `expandStringListAllowEmpty`, and using it in `pkg/resources/file_format.go` instead of `expandStringList`.

An acceptance test was added in the CSV test with an empty value.

## Test Plan

* [ ] acceptance tests - needs to be validated
* [x] A provider was build using the latest version creation/plan/modification were tested

## References

* could reference #1609
* #708


tmlp-lv:
After closer investigation with Snowflake, it seems there is a bug on Snowflake, here is the answer from the support:

> We have tested it out and we were able to reproduce the problem. It looks like a unexpected behaviour as parameter EMPTY_FIELD_AS_NULL doesn't get honoured when FIELD_OPTIONALLY_ENCLOSED_BY is used for the enclosed fields.

Closing the MR then.

sfc-gh-swinkler:
@tmlp-lv thank you for your contribution

Issue #787, opened 2021-12-11
Add on_existing_and_future option for grant resources where on_future is an option


chriscardillo:
**Is your feature request related to a problem? Please describe.**

For all grant resources that allow `on_future`, I would like some mechanism to guarantee that all existing objects (tables/views/etc.) will also be available to the users/roles listed in the grant resource. More details below.

When using a grant resource like `snowflake_table_grant`, I am able to specify `on_future` for the database or schema and assign that grant to a role or set of roles. To the best of my knowledge, the `on_future` parameter is only useful if I never change the role that has these grants.

Let's take a simple example, where I grant SELECT to all future tables in a new schema to "role_a" using the `snowflake_table_grant`.

1. The initial `snowflake_table_grant` specifies "role_a" and `on_future = true`. "role_a" can now SELECT all tables I create in this new schema
2. I create two tables - "table_1" and "table_2", which "role_a" can SELECT because of the `on_future` specified in the above resource.
3. I then adjust this `snowflake_table_grant` resource, and I replace "role_a" with "role_b".
4. "role_b" would __not__ be able to SELECT on "table_1" and "table_2", because "role_b" has only been granted SELECT on future tables, not existing tables.

**Describe the solution you'd like**

It would be nice to have something like `on_existing_and_future`, which would conflict with `on_future`.

```terraform
resource "snowflake_table_grant" "select" {
  privilege              = "SELECT"
  database_name          = "db"
  schema_name            = "schema"
  roles                  = ["role"]
  on_existing_and_future = true
  with_grant_option      = false
}
```

The above would ensure that a role change on the resource would not result in a loss of data accessibility for the new role.

One important concern of adding this feature is the resulting terraform id. I do not know the intricacies of this package, but I believe that the resulting terraform id for this grant resource is the combination of `database_name|schema_name|table_name|privilege|on_future|with_grant_option`. To implement this feature in such a way that it would not drastically affect the resulting terraform id, the `|on_future|` id slot - which currently can be nil, `true`, or `false` - could render as `|oeaf|` if `on_existing_and_future = true`.

**Describe alternatives you've considered**

- In tandem with the `snowflake_table_grant`, use the `snowflake_tables` data source, issue per-table grants to the same role that is receiving the `on_future` grants. This would alleviate the issue but is does not scale desirably. This option all doesn't work when standing up infrastructure from scratch, so one would need to add this capability after the initial apply.
[...]

Issue #284, opened 2020-10-25
GRANT SELECT TO ALL 


aaronsteers:
**Is your feature request related to a problem? Please describe.**

I'm not sure how to replicate my existing snowflake grants into terraform without access to an option for `ALL TABLES IN SCHEMA/DATABASE`.

**Describe the solution you'd like**

Just as "future" tables/views are supported in grants, the request would be to support granting access to "all" tables/views.

**Describe alternatives you've considered**

I considered creating a "task" resource to manually declare the grants - however, that "task" resource appears to be missing an "undo_sql_statement" to provide an "undo"/"destroy" capability, for instance when permission is changed/revoked.

**Additional context**

tnightengale:
Would love to help implement this!

JimFawkes:
It would be great to have this!

a2m1:
Would be great to have `GRANT ON ALL TABLES/VIEWS/SCHEMAS`

JavierLopezT:
Having this would be awesome

juangesino:
+1 on the awesomeness of having something like this 👍

gnilrets:
+1 - I can specify that I want to grant select on all future tables, but not all existing ones!

aidanmelen:
Hey @tnightengale, @JimFawkes, @a2m1, @JavierLopezT, @juangesino , and @gnilrets...  I just created a provider that allows us to declare `GRANT ALL` statements in Terraform! Here is a code snippet from the [terraform-provider-snowsql/examples/simple/main.tf](https://github.com/aidanmelen/terraform-provider-snowsql/blob/main/examples/simple/main.tf#L66-L84).

```hcl
resource "snowsql_exec" "dcl" {
  name = local.name

  create {
    statements = <<-EOT
    GRANT ALL PRIVILEGES ON ALL TABLES IN DATABASE ${snowflake_database.database.name} TO ROLE ${snowflake_role.role.name};
    GRANT ALL PRIVILEGES ON FUTURE TABLES IN DATABASE ${snowflake_database.database.name} TO ROLE ${snowflake_role.role.name};
    EOT
  }

  delete {
    statements = <<-EOT
    REVOKE ALL PRIVILEGES ON ALL TABLES IN DATABASE ${snowflake_database.database.name} FROM ROLE ${snowflake_role.role.name};
[...]

Issue #250, opened 2020-09-02
Question: GRANT SELECT ON ALL TABLES


funes79:
Hi, based on the docs (and I tested) the tables grant for future tables grants SELECT only for future tables.
But when I unset (to false) the future tables then the provider requires a table name.

How can I achieve a 
```
GRANT SELECT ON ALL TABLES IN SCHEMA <schema> TO ROLE <role>
```
using terraform resource?

Thanks

mhmcdonald:
I don't believe you can with this terraform provider. It's not currently supported.

tjtaill:
This is a feature that is greatly missed maybe if schema or table names were * wildcard values we could apply GRANT SELECT ON ALL TABLES IN SCHEMA or GRANT <PRIVILEGE> TO ALL SCHEMAS IN DATABASE ?

shaunmaris:
We would also greatly benefit from this feature

multiprocessingh:
Bumping this. Unsure where we landed with this. This is a bit of a deal breaker for a lot of folks who use third party loading services like fivetran or segment where it's difficult to administrate the number of tables being created in a database on a table by table basis. An example is the fivetran salesforce integration which has over 100+ tables being ingested.

adamrosenblit-wex:
My team would also greatly benefit from this feature.

kumarjoshi:
This will be an extremely useful feature

aicharles:
ditto

rdouglashg:
We also would find this useful!

samrogers226:
I want this

sfc-gh-bculberson:
The way terraform lifecycle works, this would only execute once and ONLY grant access to all tables that exist at that time. It wouldn't know when a new table gets added.  It's probably much better to create a resource that grants selects on all the tables you want.

adamrosenblit-wex:
> The way terraform lifecycle works, this would only execute once and ONLY grant access to all tables that exist at that time. It wouldn't know when a new table gets added. It's probably much better to create a resource that grants selects on all the tables you want.

This new resource, _combined_ with the `snowflake_table_grant` with the `on_future=true` flag set, would cover all the tables without you needing to know when a new one was created.  We have a specific use-case at my company for this combination of grants.

sfc-gh-bculberson:
[...]

PR #1626, opened 2023-03-16
feat: add GRANT ... ON ALL TABLES IN ...

Relativity74205:
In an number of issues, the snowflake feature GRANT SELECT ON **_ALL_** TABLES ... (or similar) is requested. I created a prototype of this feature in this branch. Currently it only works for the table resource (not for others).

Before I finish this work (add unit and acceptance tests, implement it on other resources, etc.) I would like to have an opinion if this feature makes sense. Not only how it was implemented, but also generally. Because:

When running e.g. GRANT SELECT ON ALL TABLES IN ..., then Snowflake creates a grant for each individual existing table and there is no way to attribute existing table grants to a GRANT SELECT ON ALL TABLES grant. Thus they cannot be checked (or removed).

## Test Plan
tbd

* [ ] acceptance tests

* [ ] …

## References
https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/787
https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/250
https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/284


sfc-gh-swinkler:
@Relativity74205 this is very good. I like this a lot. I think i will merge this, get the tests working, and it should be good to go.

Relativity74205:
@sfc-gh-swinkler Thanks for the review and acceptance. :)

sfc-gh-swinkler:
@Relativity74205 your code was pretty good as it was, and i didn't have to make too many changes. Its currently released as part of 0.60. Thank you so much!

TerjeRusska:
@Relativity74205 Awesome feature, I've been looking for this for quite a few months :smile: Dropped my resource count from table level 4096 to schema level 106 :tada:

Issue #1442, opened 2022-12-29
support for database roles


cpmub1:
snowflake has introduced DATABASE ROLES. 
https://docs.snowflake.com/en/sql-reference/sql/create-database-role.html

can support for DATABASE ROLE be introduced ?

bennylu2:
I'm not sure what the policy is for the provider to support features that are in preview

MLdavion:
wholeheartedly wanted this feature, this feature makes data share sooooo much easier to configure and control

bennylu2:
I can put up a PR for this but I'm unsure what policy is for provider to support features in preview. I'm also unsure of the status of the provider and if there's going to a be a major version release in the near future.

I'd be happy to pick this up if given the go ahead by the maintainers

cpmub1:
Thanks @bennylu2 
who should give ahead to move on ? Let me try to tag them here or get their attention

cpmub1:
Though this feature is said to be in "preview", Snowflake has implemented it on the SNOWFLAKE database & enabled it for all accounts, with no restrictions. More over, almost everyone has their own version of database role, thru custom roles - which does not provide a defined scope boundary across databases.
hence, i strongly believe, Snowflake has it as "preview" for paper-document reasons, rather than for functionality.

cpmub1:
@sfc-gh-swinkler I see that you have recently committed some changes to this repo. 
Can you please suggest on who can give a go ahead for making this change ?

@bennylu2 was willing to do a PR , but there was no ahead & the effort stopped.

bgabrhelik-com:
Now there is the `snowflake_database_role` resource, however it seems there is not support for granting of database role. Granting model seems to be quit difficult and duplication of all the resources doesn't make sense. There is small difference in the syntax.
* FQN for role
* TO DATABASE ROLE abcd;

E.g. 
GRANT USAGE ON SCHEMA DB1.PUBLIC TO DATABASE ROLE DB_READ;

sfc-gh-asawicki:
Closing because done in #1654.

Database role grants are currently being developed (#2034).

PR #1654, opened 2023-03-24
feat: database role resource

mlorek:
added database role resource: #1442 

## Test Plan
* [x ] acceptance tests

## References
https://docs.snowflake.com/en/user-guide/security-access-control-considerations#label-access-control-considerations-database-roles
*


pkg/resources/database_role.go
```diff
@@ -0,0 +1,211 @@
...
+
+	database := dbRoleID.DatabaseName
+	name := dbRoleID.RoleName
+
+	builder := snowflake.NewDatabaseRoleBuilder(name, database)
+	qry := builder.Show()
+	row := snowflake.QueryRow(db, qry)
+	// FIXME scan for name as there is LIKE pattern syntax
+	databaseRole, err := snowflake.ScanDatabaseRole(row)
+	databaseRole.DatabaseName = database // Nasty.
```
sfc-gh-swinkler, line 113: not understanding what the point of this is. If database name is a ForceNew attribute then theres no reason to read  or overwrite it at all since its never going to change.

pkg/resources/database_role_acceptance_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"strings"
+	"testing"
+	"text/template"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+type (
+	TestAccDatabaseRoleSettings struct {
```
sfc-gh-swinkler, line 15: i think this is a bit overkill. probably better to just have a function like `databaseRoleConfig(warehouseName, databaseName, roleName`)

sfc-gh-swinkler:
@mlorek this is interesting an highlights a deficiency in the currently provider. we will also need to a `snowflake_database_role_grant` implemented as well, but that can be in a separate PR

sfc-gh-swinkler (approved):

PR #1678, opened 2023-03-30
feat: mark PII as sensitive in account and user resources.

sfc-gh-ngaberel:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


sfc-gh-swinkler (approved):

hlleb-lizunkou:
@sfc-gh-ngaberel can it be a configurable option? By default, it can be enabled but we can disable it for some users. 
We have users in snowflake which is used for operational purposes and does not contain any PII information.

PR #1686, opened 2023-03-31
feat: added on_all grants for view, stage, schema and materialized view resource

Relativity74205:
Added on_all option to the following grants (accordingly to the table_grant on_all option):
- schema_grant
- stage_grant
- view_grant
- materialized_view_grant

I plan to add the on_all option to the rest of the grants in additional pull request(s)

## Test Plan
* [x] acceptance tests
* [x] unit tests
* [x] manual tests

## References
https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/250#issuecomment-1487419577


Relativity74205:
@sfc-gh-swinkler I have added the on_all option to additional grants. I would welcome a review.

 I plan to add it to the rest of the grants in an additional PR, probably next week.

sfc-gh-swinkler:
Hi @Relativity74205 this looks good, but we will also need acceptance tests for each of the resources. There is an example already for table grant. Can you please do this and then I can merge?

Relativity74205:
@sfc-gh-swinkler Sure.

However, when I started doing this, I noticed that I also had to fix some existing acceptance test. And here, I noticed something interesting:

The acceptance test `view_grant_acceptance_test.go:TestAcc_FutureViewGrantChange` fails during the last step (IMPORT), because the view_grant cannot be imported due to 
```
While attempting to import an existing object to "snowflake_view_grant.test", the provider detected that no object exists with the given id. Only pre-existing objects can be imported; check that the id is correct and that it is associated with the provider's configured region or endpoint, or use "terraform
│ apply" to create a new remote object for this resource.
```

I can reproduce the error during manual testing. The cause is the removal of the following code (I must admit I didn't understood the meaning of the code before):
```
	on_future := false
	if grantID.ObjectName == "" {
		on_future= true
	}
	err = d.Set("on_future", on_future)
```
The problem is that the Read*Grant (this issue is relevant for all future/all grants) logic relies purely on the grant ID, when determining the grant attributes during the import (no state exists at this point). But now we have a second similar attribute (`on_all`) and with this logic we cannot distinguish on_all and on_future grants during import. Thus I have introduced the following logic for Read*Grants, so it will work during import:
```
	onFuture := false
	onAll := false
	if grantID.ObjectName == "" {
		db := meta.(*sql.DB)
		var grantDetails []snowflake.GrantDetail
		if grantID.SchemaName == "" {
			grantDetails, err = snowflake.ShowFutureGrantsIn(db, "DATABASE", grantID.DatabaseName)
		} else {
			schemaName := grantID.DatabaseName + "." + grantID.SchemaName
			grantDetails, err = snowflake.ShowFutureGrantsIn(db, "SCHEMA", schemaName)
		}
		if len(grantDetails) > 0 {
			onFuture = true
		} else {
			onAll = true
		}
	}
```
This logic is only needed for importing, otherwise
```
	onFuture := d.Get("on_future").(bool)
	onAll := d.Get("on_all").(bool)
```
works as well, because the attributes can be imported from the state file. However, I don't know how to distinguish the origin of the Read*Grant call inside the function.

Relativity74205:
@sfc-gh-swinkler Sure.

However, when I started doing this, I noticed that I also had to fix some existing acceptance test, mose especially an import test of an on_future grant. And for this I could not find a solution so far.

The problem is as follows: Importing the grant resource with the on_all option is not possible as there is now way to determine this state inside of snowflake. I tried several approaches which all failed due to some reason:

- Setting `on_future = true`, when `grantID.ObjectName == "" && onAll == false`. This leads to an on_future import when actually importing an on_all grant... (on_all/on_future are always nil and thus false, when importing the resource)
- Determining the on_future state from snowflake. This works well in all cases except one: when one wants to import an on_all grant where the exact same on_future grant exists. Then importing the on_all grant would not be possible, because the logic would determine the grant every time as an on_future grant.


I have only two ideas left:
- preventing the import, when `on_all=true` was defined in the resource definition. Do you know, if this is possible?
- saving the on_all option in the ID (as in my first solution)

sfc-gh-swinkler (dismissed):

sfc-gh-swinkler:
@Relativity74205 I have been thinking a lot about resource IDs recently and I will share with you my thoughts. In Terraform, an ID is meant to uniquely identify a resource. Typically this would be something you can use to query the cloud provider in order to get the rest of the information about the resource. As an example, if there was a `widgets` API, i could do a GET to `/api/v1/widgets/<widget_id>` in order to get the rest of the information about the widget. So the default ResourceImporter function works exactly like this: given an ID, it will use the Read() function to populate the rest of the unknown values during the Import step. Except, as you pointed out, the ID would have to be updated to hold this new value since it otherwise isn't known. If we do that though we will also have to update all the existing resources which use the older ID. And then rewrite the ID during the Update() step, if ever this changes - since this happens to be a ForceNew attribute, it would not apply in this case. Another solution would be to use a custom import function, in which case the saved ID can actually be different than the imported ID.

Let me go ahead and merge your PR and then resolve this import problem since its a more general problem that affects other resources as well. I will tag you on the PR to get your feedback. Sound good?

Relativity74205:
@sfc-gh-swinkler Thanks for the explanation, it helped my understanding of the issue. And yes, your plan sounds good. However, before we merge this MR, let me add the acceptance tests to the rest of the resources, I only added them for the view_grants so far. I will try to do this this evening.

Relativity74205:
@sfc-gh-swinkler I am done with the acceptance tests, however, now the linter is complaining and I have no idea why. I cannot reproduce the issue locally.

sfc-gh-swinkler (approved):

Issue #1607, opened 2023-03-07
snowflake_account resource ignores "must_change_password" configuration argument


sfc-gh-klim:
**Provider Version**

0.58.0

**Terraform Version**

v1.3.9

**Describe the bug**

When attempting to create a new account with the "[must_change_password](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/account#must_change_password)" configuration value (set to false), the configuration argument "must_change_password" is ignored, i.e., the actual SQL query that ran doesn't have the "must_change_password" parameter at all.

The newly created account still requires the default admin user to reset their password upon initial 1st-time login.

**Expected behavior**

The admin user created for the new account should not be required to reset the password upon first login

**Code samples and commands**

```
terraform {
	required_providers {
		snowflake = {
      			source  = "Snowflake-Labs/snowflake"
      			version = "~> 0.35"
    		}
  	}
}

provider "snowflake" {
	alias = "orgadmin"
	role  = "ORGADMIN"
}

resource "snowflake_account" "account" {
	name = "tfaccount"
	admin_name = "tf_admin"
	admin_password = "P@ssw0rd"
  	must_change_password = false
  	edition = "STANDARD"
  	region = "AWS_AP_SOUTHEAST_1"
  	email = "your@email.com"
  	provider = snowflake.orgadmin
 	comment = "SNOWFLAKE"
}
```

**Additional context**

Debug log shows:
`2023-03-07T18:47:55.812+0100 [INFO]  provider.terraform-provider-snowflake_v0.58.0: 2023/03/07 18:47:55 [DEBUG] sql-conn-exec [query CREATE ACCOUNT tfaccount  ADMIN_NAME = tf_admin ADMIN_PASSWORD = 'P@ssw0rd' EMAIL = 'your@email.com' EDITION = STANDARD REGION_GROUP = PUBLIC REGION = AWS_AP_SOUTHEAST_1 COMMENT = 'SNOWFLAKE' err <nil> duration 6.771643517s args {}]: timestamp=2023-03-07T18:47:55.812+0100`
Note that the "[must_change_password](https://docs.snowflake.com/en/sql-reference/sql/create-account#optional-parameters)" parameter is missing.

Natalique:
This is still happening in version "0.67.0"
Getting the following error after creating the account 
[...]

PR #1699, opened 2023-04-10
fix: don't ignore MUST_CHANGE_PASSWORD on account creation

sfc-gh-swinkler:
#1607 
Must change password is an optional parameter which defaults to false. See: https://docs.snowflake.com/en/sql-reference/sql/create-account

> MUST_CHANGE_PASSWORD = TRUE | FALSE
> Specifies whether the new user created to administer the account is forced to change their password upon first login into the account.
> 
> Default: FALSE

Therefore in the source code for the account resource we previously only set the MUST_CHANGE_PASSWORD attribute if it is true:

```
	if b.mustChangePassword {
		q.WriteString(fmt.Sprintf(` MUST_CHANGE_PASSWORD = %t`, b.mustChangePassword))
	}
```

However it appears like the default might actually be true, and the documentation is incorrect. Regardless, it is a simple fix to remove the conditional check.

### DIFF
--- a/pkg/datasources/shares.go
+++ b/pkg/datasources/shares.go
@@ -10,7 +10,7 @@ import (
 )
 
 var sharesSchema = map[string]*schema.Schema{
-	pattern: {
+	"pattern": {
 		Type:        schema.TypeString,
 		Optional:    true,
 		Description: "Filters the command output by object name.",
--- a/pkg/resources/file_format.go
+++ b/pkg/resources/file_format.go
@@ -502,7 +502,6 @@ func CreateFileFormat(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	q := builder.Create()
-
 	err := snowflake.Exec(db, q)
 	if err != nil {
 		return fmt.Errorf("error creating file format %v err = %w", fileFormatName, err)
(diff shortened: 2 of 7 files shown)


sfc-gh-ngaberel (approved):
Should we change the default value in the resource as well? https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/account.go#L143

sfc-gh-swinkler:
@sfc-gh-ngaberel  https://docs.snowflake.com/en/sql-reference/sql/create-account#optional-parameters the default value is false, as is stated in the documentation. The difference is that we always will set this in the templated SQL statement. The end result should be identical. I guess there is no reason not to set it in the templated SQL anyways.

Natalique:
Sorry @ sfc-gh-swinkler  but this is still happening in version "0.67.0" :(
Getting the following error after creating the account 
`│ Error: could not retrieve session id err = 390106 (08004): Specified password has expired.  Password must be changed using the Snowflake web console.`

TF apply
``` # module.snowflake-account.snowflake_account.account["search"] will be created
  + resource "snowflake_account" "account" {
      + admin_name           = "name"
      + admin_password       = (sensitive value)
      + comment              = "..."
      + edition              = "ENTERPRISE"
      + email                = (sensitive value)
      + first_name           = (sensitive value)
      + id                   = (known after apply)
      + is_org_admin         = (known after apply)
      + must_change_password = false
      + name                 = "SEARCH"
      + region               = "AWS_EU_CENTRAL_1"
```

PR #1702, opened 2023-04-11
feat: Password policy resource

### LINKED ISSUE #1581: Password Policy Object

Not related to a problem just an object in Snowflake that isn't available to be created through Terraform.

I would like to be able to create the Snowflake object [Password Policy](https://docs.snowflake.com/en/sql-reference/sql/create-password-policy) through Terraform. We manage objects like this through Terraform so being able to create and manage this through TF would be helpful.

**Describe alternatives you've considered**

It just isn't available so I don't know of any alternatives

**Additional context**

I created a [discussion ](https://github.com/Snowflake-Labs/terraform-provider-snowflake/discussions/1446 )on this as well.

sfc-gh-ngaberel:
Add support for the [Password policy](https://docs.snowflake.com/en/sql-reference/sql/create-password-policy) resource.
Fixes #1581 

## Test Plan

* [X] acceptance tests

* [X] password policy builder unit tests


pkg/resources/password_policy.go
```diff
@@ -0,0 +1,447 @@
...
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+	}
+}
+
+// CreatePasswordPolicy implements schema.CreateFunc.
+func CreatePasswordPolicy(d *schema.ResourceData, meta interface{}) error {
+	manager, err := snowflake.NewPasswordPolicyManager()
+	if err != nil {
+		return fmt.Errorf("couldn't create password policy builder: %w", err)
```
sfc-gh-swinkler, line 132:
```suggestion
		return fmt.Errorf("couldn't create password policy manager: %w", err)
```

pkg/resources/password_policy.go
```diff
@@ -0,0 +1,447 @@
...
+	input := &snowflake.PasswordPolicyCreateInput{
+		PasswordPolicy: snowflake.PasswordPolicy{
+			SchemaObjectIdentifier: snowflake.SchemaObjectIdentifier{
+				Database:   d.Get("database").(string),
+				Schema:     d.Get("schema").(string),
+				ObjectName: d.Get("name").(string),
+			},
+
+			MinLength:           d.Get("min_length").(int),
+			MinLengthOk:         manager.Ok(d.GetOk("min_length")),
```
sfc-gh-swinkler, line 144: manager.Ok should probably be a helper method in the terraform provider, not SDK

pkg/resources/password_policy.go
```diff
@@ -0,0 +1,447 @@
...
+		Schema: passwordPolicySchema,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+	}
+}
+
+// CreatePasswordPolicy implements schema.CreateFunc.
+func CreatePasswordPolicy(d *schema.ResourceData, meta interface{}) error {
+	manager, err := snowflake.NewPasswordPolicyManager()
```
sfc-gh-swinkler, line 130: the word "manager" is not very descriptiive or useful. What i was thinking is that we could have a Client struct that has references to each object API. e.g.

```
type Client struct {
	conn *sql.DB

	Users           Users
	Roles           Roles
	Warehouses      Warehouses
	Databases       Databases
	Schemas         Schemas
	Tables          Tables
	NetworkPolicies NetworkPolicies
}
```
where Roles for example would be:

```
// Roles describes all the roles related methods that the
// Snowflake API supports.
type Roles interface {
	// List all the roles by pattern.
	List(ctx context.Context, options RoleListOptions) ([]*Role, error)
	// Create a new role with the given options.
	Create(ctx context.Context, options RoleCreateOptions) (*Role, error)
	// Read an role by its name.
	Read(ctx context.Context, role string) (*Role, error)
	// Update attributes of an existing role.
	Update(ctx context.Context, role string, options RoleUpdateOptions) (*Role, error)
	// Delete a role by its name.
	Delete(ctx context.Context, role string) error
	// Rename a role name.
	Rename(ctx context.Context, old string, new string) error
}

```
then you could reference password policeis as: `client.PasswordPolicies.Create(<CreatePasswordPolicyInput>)`

I also think addtion a validation method to the input would be useful. Example:

```
// RoleCreateOptions represents the options for creating a role.
type RoleCreateOptions struct {
	*RoleProperties

	// Required: Identifier for the role; must be unique for your account.
	Name string
}

func (o RoleCreateOptions) validate() error {
	if o.Name == "" {
		return errors.New("name must not be empty")
	}
	return nil
}
```
then you could validate input before doing generic request as so:

```
func (r *roles) List(ctx context.Context, options RoleListOptions) ([]*Role, error) {
	if err := options.validate(); err != nil {
		return nil, fmt.Errorf("validate list options: %w", err)
	}
...
```

pkg/resources/password_policy.go
```diff
@@ -0,0 +1,447 @@
...
+		IfNotExistsOk: manager.Ok(d.GetOk("if_not_exists")),
+	}
+
+	stmt, err := manager.Create(input)
+	if err != nil {
+		return fmt.Errorf("couldn't generate create statement: %w", err)
+	}
+
+	db := meta.(*sql.DB)
+	_, err = db.Exec(stmt)
```
sfc-gh-swinkler, line 177: still not fond of having this db.Exec right here. Would be better to put this logic in the client struct (see earlier comment)

pkg/resources/password_policy.go
```diff
@@ -0,0 +1,447 @@
...
+	if err != nil {
+		return fmt.Errorf("error executing create statement: %w", err)
+	}
+
+	d.SetId(PasswordPolicyID(&input.PasswordPolicy))
```
sfc-gh-swinkler, line 182: a lot of resources share the same kind of ID format. This could probably be abstracted. Also i am not sure it needs to accept the whole PasswordPolicy reference, it really just needs the schema object identifier.

pkg/resources/password_policy.go
```diff
@@ -0,0 +1,447 @@
...
+
+// UpdatePasswordPolicy implements schema.UpdateFunc.
+func UpdatePasswordPolicy(d *schema.ResourceData, meta interface{}) error {
+	manager, err := snowflake.NewPasswordPolicyManager()
+	if err != nil {
+		return fmt.Errorf("couldn't create password policy builder: %w", err)
+	}
+
+	runAlter := false
+	alterInput := &snowflake.PasswordPolicyUpdateInput{
```
sfc-gh-swinkler, line 256: should we stick with the Snowflake naming convention? in which case this would be `PasswordPolicyAlterInput`?

pkg/resources/password_policy.go
```diff
@@ -0,0 +1,447 @@
...
+			},
+		},
+	}
+
+	if d.HasChange("min_length") {
+		val, ok := d.GetOk("min_length")
+		if ok {
+			alterInput.MinLength = val.(int)
+			alterInput.MinLengthOk = true
+			runAlter = true
```
sfc-gh-swinkler, line 281: is this flag really necessary? perhaps it could be set internally? Also how to differentiate between a non-set value, and a set value? ex, lets say MinLength is not set. It would default to false when you instantiate the struct. So it might look like its being set to false when actually its not being set at all.

pkg/resources/password_policy.go
```diff
@@ -0,0 +1,447 @@
...
+	db := meta.(*sql.DB)
+	_, err = db.Exec(stmt)
+	if err != nil {
+		return fmt.Errorf("error executing drop statement: %w", err)
+	}
+
+	return nil
+}
+
+func PasswordPolicyID(pp *snowflake.PasswordPolicy) string {
```
sfc-gh-swinkler, line 445: this doesn't seem like a very useful function

pkg/snowflake/builder.go
```diff
@@ -0,0 +1,419 @@
...
+	String     ParamType = "string"
+	StringList ParamType = "stringlist"
+)
+
+type SQLParameter struct {
+	name      string
+	paramType ParamType
+}
+
+type BuilderConfig struct {
```
sfc-gh-swinkler, line 24: can we call this SQLBuilderConfig?

pkg/snowflake/builder.go
```diff
@@ -0,0 +1,419 @@
...
+	beforeObjectType map[string]string
+	afterObjectType  map[string]string
+	parameters       map[string]*SQLParameter
```
sfc-gh-swinkler, line 27: could this be `[]*SQLParameter`? If you use the name as a field on the struct it should be okay. Benefit is that it preserves order.

pkg/snowflake/builder.go
```diff
@@ -0,0 +1,419 @@
...
+)
+
+type SQLParameter struct {
+	name      string
+	paramType ParamType
+}
+
+type BuilderConfig struct {
+	beforeObjectType map[string]string
+	afterObjectType  map[string]string
```
sfc-gh-swinkler, line 26: Should these be `[]*SQLKeyword`?

pkg/snowflake/builder.go
```diff
@@ -0,0 +1,419 @@
...
+	parameters       map[string]*SQLParameter
+}
+
+type NewBuilder struct {
+	objectType       string
+	objectTypePlural string
+	createConfig     *BuilderConfig
```
sfc-gh-swinkler, line 33: maybe make a nested struct for configs. since you have so many of them. also why even does the config need to be saved? couldn't it be purely functional?

pkg/snowflake/builder.go
```diff
@@ -0,0 +1,419 @@
...
+
+	return sb.String(), nil
+}
+
+func (b *NewBuilder) Create(obj Identifier) (string, error) {
+	sb := strings.Builder{}
+	sb.WriteString("CREATE")
+
+	// eg. "OR REPLACE"
+	before, err := b.renderKeywords(obj, b.createConfig.beforeObjectType)
```
sfc-gh-swinkler, line 229: yeah i think the createConfig should probably be an input

pkg/snowflake/password_policy.go
```diff
@@ -0,0 +1,111 @@
...
+type PasswordPolicyUpdateInput struct {
+	PasswordPolicy
+
+	IfExists   bool `pos:"afterObjectType" value:"IF EXISTS"`
+	IfExistsOk bool
+}
+
+func (m *PasswordPolicyManager) Update(x *PasswordPolicyUpdateInput) (string, error) {
+	return m.genericBuilder.Alter(x)
+}
```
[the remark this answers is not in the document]
sfc-gh-swinkler: is there a way to automatically run gofumpt whenever a PR is made? this is a really annoying error message

sfc-gh-swinkler (changes requested):

pkg/snowflake/password_policy.go
```diff
@@ -0,0 +1,114 @@
...
+	return m.genericBuilder.Create(x)
+}
+
+type (
+	PasswordPolicyReadInput  = SchemaObjectIdentifier
+	PasswordPolicyReadOutput = PasswordPolicy
+)
+
+func (m *PasswordPolicyManager) Read(x *PasswordPolicyReadInput) (string, error) {
+	return m.genericBuilder.Describe(x)
```
sfc-gh-swinkler, line 78: how about SQLBuilder?

pkg/snowflake/password_policy.go
```diff
@@ -0,0 +1,114 @@
...
+func (m *PasswordPolicyManager) Create(x *PasswordPolicyCreateInput) (string, error) {
+	return m.genericBuilder.Create(x)
+}
+
+type (
+	PasswordPolicyReadInput  = SchemaObjectIdentifier
+	PasswordPolicyReadOutput = PasswordPolicy
+)
+
+func (m *PasswordPolicyManager) Read(x *PasswordPolicyReadInput) (string, error) {
```
sfc-gh-swinkler, line 77: Describe instead of Read?

pkg/resources/password_policy.go
sfc-gh-ngaberel: 1. yes, an update is an ALTER SET query and/or an ALTER UNSET query, we need to know which ones are necessary.
2. That's what the ok flag is for for each field (true = flag is present)

pkg/snowflake/password_policy.go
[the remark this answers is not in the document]
sfc-gh-ngaberel: Yes, I'll add that with the vscode config changes.

pkg/snowflake/password_policy.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func NewPasswordPolicyManager() (*PasswordPolicyManager, error) {
+	builder, err := newBuilder(
+		"PASSWORD POLICY",
+		"PASSWORD POLICIES",
+		reflect.TypeOf(PasswordPolicyCreateInput{}),
+		reflect.TypeOf(PasswordPolicyUpdateInput{}),
+		reflect.TypeOf(PasswordPolicyUpdateInput{}),
+		reflect.TypeOf(PasswordPolicyDeleteInput{}),
```
sfc-gh-swinkler, line 45: i really don't like this way of registering functions with the password policy manager. i am sure there is a better way to do this.

sfc-gh-swinkler (approved):

PR #1703, opened 2023-04-11
feat: provider config support insecureMode

sfc-gh-swinkler:
Allow support for optional insecureMode parameter to bypass OCSP checks


pkg/provider/provider.go
```diff
@@ -164,6 +164,12 @@ func Provider() *schema.Provider {
 				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_PROTOCOL", "https"),
 			},
+			"insecure_mode": {
+				Type:        schema.TypeBool,
+				Description: "If true, bypass the Online Certificate Status Protocol (OCSP) certificate revocation check. IMPORTANT: Change the default value for testing or emergency situations only..",
```
sfc-gh-ngaberel, line 169:
```suggestion
				Description: "If true, bypass the Online Certificate Status Protocol (OCSP) certificate revocation check. IMPORTANT: Change the default value for testing or emergency situations only.",
```

docs/data-sources/shares.md
```diff
@@ -0,0 +1,38 @@
+---
```
sfc-gh-ngaberel, line 1: Should this be in a different PR?

sfc-gh-swinkler: this is just the result of "make docs". so it means current main is not up to date.

sfc-gh-ngaberel (approved):

PR #1713, opened 2023-04-13
fix: grant imports

sfc-gh-swinkler:
normalizes grant IDs, no matter what the format used to be. fixes import issues for grant IDs.


sfc-gh-swinkler:
@Relativity74205 this fixes the ID for imports. i decided to just rewrite all the IDs so we don't have so many different kinds of IDs. I tested this with a few grant resources having older IDs and it does correctly change the ID, but only during an Apply. And import works as expected.

sfc-gh-ngaberel (dismissed):

pkg/resources/view_grant.go
```diff
@@ -139,6 +144,9 @@ func CreateViewGrant(d *schema.ResourceData, meta interface{}) error {
 	if _, ok := d.GetOk("view_name"); ok {
 		viewName = d.Get("view_name").(string)
 	}
+	if err := d.Set("view_name", viewName); err != nil {
+		return err
+	}
 	var schemaName string
 	if _, ok := d.GetOk("schema_name"); ok {
 		schemaName = d.Get("schema_name").(string)
```
Relativity74205, line 152: Is the block from line 143-153 still needed? Or would it be enough to just use?
```suggestion
viewName := d.Get("view_name").(string)
schemaName := d.Get("schema_name").(string)
```

Relativity74205, line 152: Is this block still needed? Or would it be enough to just use?
```suggestion
viewName := d.Get("view_name").(string)
schemaName := d.Get("schema_name").(string)
```
Especially this seems odd:
```
	if _, ok := d.GetOk("view_name"); ok {
		viewName = d.Get("view_name").(string)
	}
	if err := d.Set("view_name", viewName); err != nil {
		return err
	}
```

I don't know, if this is special to this grant, I just have picked this grant for reviewing/testing.

Relativity74205:
> @Relativity74205 this fixes the ID for imports. i decided to just rewrite all the IDs so we don't have so many different kinds of IDs. I tested this with a few grant resources having older IDs and it does correctly change the ID, but only during an Apply. And import works as expected.

@sfc-gh-swinkler That looks great, I guess this solves a lot of problems with IDs. I have noticed only a small cosmetic issue (see above).

pkg/resources/view_grant.go
sfc-gh-swinkler: I agree this doesn't make a lot of sense, its more to uphold existing contract. if `on_future` or `on_all` is set to true  then "view_name" (or "table_name", "materialized_view_name", etc.) is expected to be an empty string. We can either remove that expectation, or add in a default value to the object name. I am thinking removing that  expectation is sufficient.

sfc-gh-ngaberel: I think you're right @Relativity74205, good catch! @sfc-gh-swinkler what was the intent for this block? From what I can tell the only difference is that it would now set view_name to an empty string if it wasn't set by the user, but I assume that's already the default value?

sfc-gh-ngaberel (approved):

Issue #1674, opened 2023-03-29
Broken snowflake_database resource in 0.60.0 provider version


marcincuber:
**Provider Version**

0.60.0

**Terraform Version**

1.4.1

**Describe the bug**

Snowflake databases seem to disappeared from state file and terraform plan wants to re-create already deployed databases.

**Expected behavior**

Database should exists and plan should report no changes.

**Code samples and commands**

```
resource "snowflake_database" "eurowind" {
  name = "EUROWIND"
}
```

**Additional context**

Another problem is that with this provider version, even import command doesn't work.
```
terraform import snowflake_database.eurowind EUROWIND                                                                                                                          
╷
│ Error: Cannot import non-existent remote object
│
│ While attempting to import an existing object to "snowflake_database.eurowind", the provider detected that no object exists with the given id. Only pre-existing objects can be imported; check that the id is
│ correct and that it is associated with the provider's configured region or endpoint, or use "terraform apply" to create a new remote object for this resource.
```

This release is super broken. Fix was to downgrade to provider and re-import database resources.

mlorek:
also getting the following err on resource monitor options change
11:37:20  
11:37:20  Stack trace from the terraform-provider-snowflake_v0.60.0 plugin:
11:37:20  
11:37:20  panic: interface conversion: interface {} is *schema.Set, not int

adamhealy:
+1 on this issue. Same issue with databases, same fix to roll back an re-import

kietay:
If the provider fails to query the dbs it just thinks the resource has been deleted. It should be checking for a nil, not an error https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/database.go#L231-L235

sfc-gh-swinkler:
@marcincuber what is the ID of your database resource? I would like to help resolve this issue, but I cannot replicate the issue in the 0.61 version. 

configuration:
```
resource "snowflake_database" "eurowind" {
  name = "EUROWIND"
}

```

apply is succesful
```
snowflake_database.eurowind: Creating...
snowflake_database.eurowind: Creation complete after 1s [id=EUROWIND]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

no changes during subsequent plan

```
snowflake_database.eurowind: Refreshing state... [id=EUROWIND]

No changes. Your infrastructure matches the configuration.
```

removing the resource from TF state and reimporting:

```
[...]

PR #1725, opened 2023-04-18
fix: be more cautious when setting db id to empty

sfc-gh-swinkler:
fix for #1674 
this adds back in the initial scan on database, and only sets the database id to "" if the error is ErrNoRows, and throws exceptions for all else, without setting the id to "". This is the original behavior. Retaining the ListDatabases function which is necessary in case there is more than one match for the given database name.

Also reverted read to use d.Id() instead of custom import function, which really wasn't necessary here anyways.


pkg/resources/database.go
```diff
@@ -227,11 +220,25 @@ func createDatabaseFromReplica(d *schema.ResourceData, meta interface{}) error {
...
+	_, err := snowflake.ScanDatabase(row)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			// If not found, mark resource to be removed from statefile during apply or refresh
+			log.Printf("[DEBUG] database (%s) not found", name)
+			d.SetId("")
+			return nil
+		}
+		return fmt.Errorf("unable to scan row for SHOW DATABASES")
+	}
```
sfc-gh-ngaberel, line 236: Why do we need the SHOW query? I assume this is to make sure the database with that name actually exists, would be good to leave a comment to explain why this is here.

sfc-gh-ngaberel (approved):

sfc-gh-ngaberel (approved):

PR #1727, opened 2023-04-18
feat: alerts datasource

mlorek:
alerts datasource

## Test Plan

* [x] acceptance tests

## References

https://docs.snowflake.com/en/sql-reference/sql/show-alerts
*


sfc-gh-swinkler:
Hello @mlorek . I think I would rather design the interface for this one to be slightly different , so as it follows the structure of the Snowflake API more closely (see below code). It would also be good to see this use the new v2 SDK format, although I realize that is asking a lot. If this is urgent, I could see merging this with some lesser modifications, but then just know that i would eventually make some breaking changes in the future to make it look more like the structure shown below.

```
data "snowflake_alerts" "d" {
terse = true/false (default false)
like {
 pattern = "<filter pattern>"
}

in {
 account = true / false
 database = "db_name"
 schema = "schema_name>
}

starts_with = "<name_string>"
limit {
 rows = <row_nb>
 from = "<name_string>"
}

alerts = [
 ... (should be complete alert properties and metadata)
]
}

```

sfc-gh-swinkler:
Thanks @mlorek

Issue #1738, opened 2023-04-20
snowflake_stream always perform forces replacement for stream on view


nullcoder:
**Provider Version**

0.61

**Terraform Version**

1.1.6

**Describe the bug**

`snowflake_stream` always perform forces replacement for stream on view.
It thought that the current state was the stream on table.

Example plan:
```
-/+ resource "snowflake_stream" "stream" {
      ~ id                = "MY_DATABASE|MY_SCHEMA|MY_STREAM" -> (known after apply)
        name              = "MY_STREAM"
      - on_table          = "MY_DATABASE.MY_SCHEMA.MY_VIEW" -> null # forces replacement
      + on_view           = "MY_DATABASE.MY_SCHEMA.MY_VIEW" # forces replacement
      ~ owner             = "MY_USER" -> (known after apply)
        # (5 unchanged attributes hidden)
    }
```

**Expected behavior**

It should not force replace the stream if nothing changed.

**Code samples and commands**

```terraform
resource "snowflake_stream" "stream" {
  database = "MY_DATABASE"
  schema   = "MY_SCHEMA"
  name     = "MY_STREAM"

  on_view     = "MY_DATABASE.MY_SCHEMA.MY_VIEW"
}
```

**Additional context**

I think the bug caused by `ListStreams` which there is no column `view_name` returned from [`SHOW STREAMS`](https://docs.snowflake.com/en/sql-reference/sql/show-streams#output).
https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/929801b7b6f0892db32b3c103adbe8a974a08217/pkg/snowflake/stream.go#L182-L184

It should check `source_type` and populate `on_view` with `table_name` instead of `view_name`.
https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/929801b7b6f0892db32b3c103adbe8a974a08217/pkg/resources/stream.go#L320-L332

PR #1112, opened 2022-07-09
feat: Streams on views

gouline:
Allow streams on views (rather than tables), now that it's in public preview.

## Test Plan

* [x] acceptance tests

## References

* [CREATE STREAM](https://docs.snowflake.com/en/sql-reference/sql/create-stream.html)

PR #1740, opened 2023-04-21
fix: get `on_view` from `table_name` for `stream`

nullcoder:
This PR fixes https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1738

There is no column `view_name` returned from [`SHOW STREAMS`](https://docs.snowflake.com/en/sql-reference/sql/show-streams#output).
This causes the resource `snowflake_stream` to always force replace the stream on view even there was no changes on any attributes.

I believe the original PR https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1112 was done when this feature was in public preview so this probably changed since GA?

## Test Plan

* [X] internal tests
* [X] unit tests
* [X] acceptance tests -- Note: TestAcc_Stream Step 6 `STREAM_ON_STAGE` was failed since before this change. I believe the `table_name` returned from `SHOW STREAMS` for `source_type='Stage'` is not prefixed with database and schema name.

## References
* https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1738
* https://docs.snowflake.com/en/sql-reference/sql/show-streams
* https://docs.snowflake.com/en/sql-reference/sql/create-stream


sfc-gh-swinkler (approved):
@nullcoder this looks good, thank you!

PR #1739, opened 2023-04-21
feat: masking policy application resource

sfc-gh-ngaberel:
Add support for the `snowflake_table_column_masking_policy_application` resource.

## Test Plan
* [ ] acceptance tests
* [ ] sdk unit tests

## References

* Fixes https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1719

PR #1756, opened 2023-04-26
column name should be "secret"

liamjamesfoley:
I think the example column should be "secret" to match the table created in the example.


liamjamesfoley:
FYI @sfc-gh-ngaberel re: https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1739
Thanks for adding this! Excited to start using it.

sfc-gh-swinkler (approved):
LGTM!

PR #1771, opened 2023-05-02
feat: adding CREATE STREAMLIT as privilege

danielhstahl:
With Streamlit support on the horizon, adding CREATE STREAMLIT privilege to Terraform.

## Test Plan
I added this as a configuration in privileges, I don't believe any logic was changed.


christyvp:
+1 we would also like to see this feature

sfc-gh-swinkler (approved):
LGTM!

sfc-gh-swinkler:
@danielhstahl thank you for your contribution!

PR #1792, opened 2023-05-11
chore: Migrate Warehouse resource + datasource to new SDK

sfc-gh-ngaberel:
Some API changes:
- Tags field on `snowflake_warehouse` resource has been removed. Use `snowflake_tag_association` instead.
- `snowflake_warehouse` resource fields `warehouse_size`, `scaling_policy` and `warehouse_type` now accept a subset of the options they previously accepted.

## Test Plan
* [X] previous acceptance tests are mostly unchanged
* [X] new SDK is covered by SQL unit tests + full integration tests

## References

*


pkg/sdk/warehouse.go
```diff
@@ -0,0 +1,476 @@
...
+	EnableQueryAcceleration         *bool          `ddl:"parameter" db:"ENABLE_QUERY_ACCELERATION"`
+	QueryAccelerationMaxScaleFactor *uint8         `ddl:"parameter" db:"QUERY_ACCELERATION_MAX_SCALE_FACTOR"`
+
+	// Object params
+	MaxConcurrencyLevel             *uint `ddl:"parameter" db:"MAX_CONCURRENCY_LEVEL"`
+	StatementQueuedTimeoutInSeconds *uint `ddl:"parameter" db:"STATEMENT_QUEUED_TIMEOUT_IN_SECONDS"`
+	StatementTimeoutInSeconds       *uint `ddl:"parameter" db:"STATEMENT_TIMEOUT_IN_SECONDS"`
+}
+
+type WarehouseUnsetField string
```
sfc-gh-swinkler, line 164: Would be better to use an Unset struct rather than making an enum type for this

pkg/sdk/warehouse.go
```diff
@@ -0,0 +1,476 @@
...
+		opts.Set,
+		opts.Unset,
+		opts.SetTags,
+		opts.UnsetTags,
+	}
+	if err := checkExclusivePointers(exclusivePointers); err != nil {
+		return fmt.Errorf("exactly one of Suspend, Resume, AbortAllQueries, NewName, Set, Unset, SetTags and UnsetTags must be set: %w", err)
+	}
+
+	if opts.Suspend != nil && *opts.Suspend && opts.Resume != nil && *opts.Resume {
```
sfc-gh-swinkler, line 203: checkEclsuviePointers is a cool function. we should do something similiar for checking that both are not true. maybe making a rules validation lib

pkg/sdk/warehouse.go
```diff
@@ -0,0 +1,476 @@
...
+	Owner                           string
+	Comment                         string
+	EnableQueryAcceleration         bool
+	QueryAccelerationMaxScaleFactor uint8
+	ResourceMonitor                 string
+	Actives                         string
+	Pendings                        string
+	Failed                          string
+	Suspended                       string
+	Uuid                            string
```
sfc-gh-swinkler, line 308: should be UUID

pkg/sdk/type_helpers.go
```diff
@@ -19,6 +19,16 @@ func Int(i int) *int {
 	return &i
 }
 
+// Uint returns a pointer to the given uint.
+func Uint(i uint) *uint {
```
sfc-gh-swinkler, line 23: not sure that i like this over having int type

sfc-gh-swinkler (approved):

astanfel:
@sfc-gh-ngaberel This is a breaking change for existing declaration's of a given warehouse that are in lower case. 

I got this error when testing another change I was working on 

```bash 

│ Error: expected warehouse_size to be one of [XSMALL SMALL MEDIUM LARGE XLARGE XXLARGE XXXLARGE X4LARGE X5LARGE X6LARGE], got xsmall

```

AnatEzra81:
@sfc-gh-ngaberel this introduced a breaking change for the `max_cluster_count` warehouse property. 
this [pr](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1057) allowed values greater than 10 since this is a soft limit. we are using warehouses with more clusters than 10 and cannot reduce the value because it will affect production load

sfc-gh-ngaberel:
Hi @AnatEzra81, thanks for reporting this, I'm talking to other teams to figure out what the validation rules should actually be (if any). Could you create a separate issue so we can track this?

Issue #1795, opened 2023-05-12
Can not create external stage with storage integration


toni-moreno:
**Provider Version**

+ provider registry.terraform.io/snowflake-labs/snowflake v0.64.0

**Terraform Version**

Terraform v1.4.6
on linux_amd64

**Problem description**

When trying to create new external stages with `storage_integration` this error happens

```
╷
│ Error: error creating stage MYSTAGE
│ 
│   with snowflake_stage.stage["MYDB|PUBLIC|MYSTAGE"],
│   on db-objects.tf line 174, in resource "snowflake_stage" "stage":
│  174: resource "snowflake_stage" "stage" {
│ 
╵
```
**Expected behavior**

Create a new storage integration

**Code samples and commands**

```tf
resource "snowflake_stage" "stage" {
  for_each = local.stage_map

  database            = MYDB
  schema              = PUBLIC
  name                = MYSTAGE
  url                 = "azure://xxxxxx.blob.core.windows.net/snowflakecsvx"
  storage_integration = "azure_integration"

}
```

**Additional context**

After review some logs setting TF_LOG env var,  I've noticed that the provider is building erroneously the query .  It creates the following query

```sql
CREATE STAGE "MYDB"."PUBLIC"."MYSTAGE" URL = 'azure://xxxxxx.blob.core.windows.net/snowflakecsv' STORAGE_INTEGRATION = "azure_integration"
```

but the correct query should built without double quotes in "azure_integration"

```sql
CREATE STAGE "MYDB"."PUBLIC"."MYSTAGE" URL = 'azure://xxxxxx.blob.core.windows.net/snowflakecsv' STORAGE_INTEGRATION = azure_integration
```

danu165:
If you change `storage_integration = "azure_integration"` to `storage_integration = "AZURE_INTEGRATION"` then that should fix your issue

toni-moreno:
@danu165  you are right . 

But the provider should work with both cases upper and lowercase integration names. IMHO The bug still exist.

These are the 4 possible options: lower/upper with double quotes and without

**With double quotes**
```sql
CREATE STAGE "MYDB"."PUBLIC"."MYSTAKE" URL = 'azure://xxxxx' STORAGE_INTEGRATION = "azure_integration"; --> FAIL  
[...]

PR #1798, opened 2023-05-12
fix: create external stages with storage integration (#1795)

toni-moreno:
This fix the #1795  bug and now we are able to create external stages with storage integrations.


toni-moreno:
Closed because this solution won't fix the error in all cases. I will review later

PR #1804, opened 2023-05-15
fix: warehouse for sdk v2

sfc-gh-swinkler:
- refactor warehouse so that set and unset tags work.
- also change purpose of ddl:"list" to now operate on structs 
- remove "ddl:"command" which wasn't very useful
- cleanup sql builder engine


pkg/sdk/warehouse.go
```diff
@@ -71,22 +71,21 @@ type WarehouseCreateOptions struct {
...
 	Comment                         *string        `ddl:"parameter,single_quotes" db:"COMMENT"`
 	EnableQueryAcceleration         *bool          `ddl:"parameter" db:"ENABLE_QUERY_ACCELERATION"`
-	QueryAccelerationMaxScaleFactor *uint8         `ddl:"parameter" db:"QUERY_ACCELERATION_MAX_SCALE_FACTOR"`
+	QueryAccelerationMaxScaleFactor *int           `ddl:"parameter" db:"QUERY_ACCELERATION_MAX_SCALE_FACTOR"`
 
 	// Object params
-	MaxConcurrencyLevel             *uint            `ddl:"parameter" db:"MAX_CONCURRENCY_LEVEL"`
-	StatementQueuedTimeoutInSeconds *uint            `ddl:"parameter" db:"STATEMENT_QUEUED_TIMEOUT_IN_SECONDS"`
-	StatementTimeoutInSeconds       *uint            `ddl:"parameter" db:"STATEMENT_TIMEOUT_IN_SECONDS"`
-	Tags                            []TagAssociation `ddl:"list,parentheses" db:"TAG"`
```
sfc-gh-ngaberel, line 89: Why remove the Tags field? we won't use it in the Terraform provider but the SDK should probably still support this operation.

pkg/sdk/warehouse.go
```diff
@@ -135,81 +134,99 @@ type WarehouseAlterOptions struct {
...
 
 	// Object params
-	MaxConcurrencyLevel             *uint `ddl:"parameter" db:"MAX_CONCURRENCY_LEVEL"`
-	StatementQueuedTimeoutInSeconds *uint `ddl:"parameter" db:"STATEMENT_QUEUED_TIMEOUT_IN_SECONDS"`
-	StatementTimeoutInSeconds       *uint `ddl:"parameter" db:"STATEMENT_TIMEOUT_IN_SECONDS"`
+	MaxConcurrencyLevel             *int `ddl:"parameter" db:"MAX_CONCURRENCY_LEVEL"`
+	StatementQueuedTimeoutInSeconds *int `ddl:"parameter" db:"STATEMENT_QUEUED_TIMEOUT_IN_SECONDS"`
+	StatementTimeoutInSeconds       *int `ddl:"parameter" db:"STATEMENT_TIMEOUT_IN_SECONDS"`
+
+	Tag []TagAssociation `ddl:"keyword" db:"TAG"`
```
sfc-gh-ngaberel, line 161: Pretty sure I tried this last week but turns out tags can't be set at the same time as normal parameters, it's either `SET param = 'val';` or `SET TAG db.s.tag = 'val';`, we can't mix both. Same for unset.
Hence why I created a separate field in `WarehouseAlterOptions`.

pkg/sdk/warehouse.go
```diff
@@ -135,81 +134,99 @@ type WarehouseAlterOptions struct {
...
-		return fmt.Errorf("exactly one of Suspend, Resume, AbortAllQueries, NewName, Set, Unset, SetTags and UnsetTags must be set: %w", err)
-	}
-
-	if opts.Suspend != nil && *opts.Suspend && opts.Resume != nil && *opts.Resume {
-		return fmt.Errorf(`"Suspend" and "Resume" cannot both be true`)
+	if everyValueSet(opts.Suspend, opts.Resume) && (*opts.Suspend && *opts.Resume) {
+		return fmt.Errorf("Suspend and Resume cannot both be true")
 	}
-	if opts.IfSuspended != nil && *opts.IfSuspended && (opts.Resume == nil || !*opts.Resume) {
+	if valueSet(opts.IfSuspended) && !valueSet(opts.Resume) {
```
sfc-gh-ngaberel, line 203: We also need to check the values of IfSuspended and Resume, not just their values:
```suggestion
	if valueSet(opts.IfSuspended) && *opts.ifSuspended && (!valueSet(opts.Resume) || !*opts.Resume) {
```

pkg/sdk/warehouse_test.go
```diff
@@ -33,55 +33,44 @@ func TestWarehouseCreate(t *testing.T) {
...
 		}
 		clauses, err := builder.parseStruct(opts)
 		require.NoError(t, err)
 		assert.Equal(t,
-			`CREATE OR REPLACE WAREHOUSE IF NOT EXISTS "completewarehouse" WAREHOUSE_TYPE = 'STANDARD' WAREHOUSE_SIZE = 'X4LARGE' MAX_CLUSTER_COUNT = 8 MIN_CLUSTER_COUNT = 3 SCALING_POLICY = 'ECONOMY' AUTO_SUSPEND = 1000 AUTO_RESUME = true INITIALLY_SUSPENDED = false RESOURCE_MONITOR = "myresmon" COMMENT = 'hello' ENABLE_QUERY_ACCELERATION = true QUERY_ACCELERATION_MAX_SCALE_FACTOR = 62 MAX_CONCURRENCY_LEVEL = 7 STATEMENT_QUEUED_TIMEOUT_IN_SECONDS = 29 STATEMENT_TIMEOUT_IN_SECONDS = 89 TAG ("db"."schema"."tag1" = 'v1',"db2"."schema2"."tag2" = 'v2')`,
+			`CREATE OR REPLACE WAREHOUSE IF NOT EXISTS "completewarehouse" WAREHOUSE_TYPE = 'STANDARD' WAREHOUSE_SIZE = 'X4LARGE' MAX_CLUSTER_COUNT = 8 MIN_CLUSTER_COUNT = 3 SCALING_POLICY = 'ECONOMY' AUTO_SUSPEND = 1000 AUTO_RESUME = true INITIALLY_SUSPENDED = false RESOURCE_MONITOR = "myresmon" COMMENT = 'hello' ENABLE_QUERY_ACCELERATION = true QUERY_ACCELERATION_MAX_SCALE_FACTOR = 62 MAX_CONCURRENCY_LEVEL = 7 STATEMENT_QUEUED_TIMEOUT_IN_SECONDS = 29 STATEMENT_TIMEOUT_IN_SECONDS = 89 TAG "db"."schema"."tag1" = 'v1',"db2"."schema2"."tag2" = 'v2'`,
```
sfc-gh-ngaberel, line 54: Shouldn't we remove the TAG bit?

pkg/sdk/warehouse_test.go
```diff
@@ -95,20 +84,19 @@ func TestWarehouseAlter(t *testing.T) {
...
+				Tag: []ObjectIdentifier{
+					NewSchemaObjectIdentifier("db", "schema", "tag1"),
+				},
 			},
 		}
 		clauses, err := builder.parseStruct(opts)
 		require.NoError(t, err)
 		assert.Equal(t,
-			`ALTER WAREHOUSE "mywarehouse" UNSET WAREHOUSE_SIZE,MAX_CLUSTER_COUNT,AUTO_RESUME`,
+			`ALTER WAREHOUSE "mywarehouse" UNSET WAREHOUSE_SIZE,MAX_CLUSTER_COUNT,AUTO_RESUME,TAG "db"."schema"."tag1"`,
```
sfc-gh-ngaberel, line 99: This is invalid SQL, cf previous comment about SET + TAG.

pkg/sdk/common_types.go
```diff
@@ -1,5 +1,10 @@
 package sdk
 
+type LimitPagination struct {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: because it is specifically used for limit. actually this is used for show shares: https://docs.snowflake.com/en/sql-reference/sql/show-shares, so i can remove it and add it to that PR where it might make more sense. Most resources "limit" only a number of rows, but this one also has "from" which is used for pagination purposes of the limit function. Hence "LimitPagination"

pkg/sdk/identifier_helpers.go
```diff
@@ -5,11 +5,107 @@ import (
...
+		accountIdentifier: NewAccountIdentifier(orgName, accountName),
+	}
+}
+
+func (i ExternalObjectIdentifier) Name() string {
+	return i.objectIdentifier.Name()
+}
+
+func (i ExternalObjectIdentifier) FullyQualifiedName() string {
+	return fmt.Sprintf(`%v.%v`, i.accountIdentifier.Name(), i.objectIdentifier.FullyQualifiedName())
```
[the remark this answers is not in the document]
sfc-gh-swinkler: correct. external objects id does not use wrapped double quotes. what distinugishes external ids is that they are prefixed with `organization_name-account_name:<object_identifier>`

pkg/sdk/sql_builder.go
```diff
@@ -7,82 +7,164 @@ import (
...
-	NoQuotes     quoteType = "no_quotes"
-	DoubleQuotes quoteType = "double_quotes"
-	SingleQuotes quoteType = "single_quotes"
+	NoQuotes     quoteModifier = "no_quotes"
+	DoubleQuotes quoteModifier = "double_quotes"
+	SingleQuotes quoteModifier = "single_quotes"
 )
 
-func (qt quoteType) Quote(v interface{}) string {
+func (qm quoteModifier) Modify(v any) any {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: because it implements the modify interface

pkg/sdk/warehouse.go
sfc-gh-swinkler: Tags are not parts of the create operation, they are part of alter

pkg/sdk/warehouse.go
sfc-gh-swinkler: i can do validation to ensure that the user only sets one or the other but i prefer to keep it this way

pkg/sdk/sql_builder.go
```diff
@@ -7,82 +7,149 @@ import (
...
-		escapedString := strings.ReplaceAll(s, qt.String(), qt.String()+qt.String())
-		return fmt.Sprintf(`%v%v%v`, qt.String(), escapedString, qt.String())
+		escapedString := strings.ReplaceAll(s, qm.String(), qm.String()+qm.String())
+		return fmt.Sprintf(`%v%v%v`, qm.String(), escapedString, qm.String())
 	case SingleQuotes:
-		escapedString := strings.Trim(s, qt.String())
-		return fmt.Sprintf(`%v%v%v`, qt.String(), escapedString, qt.String())
+		// https://docs.snowflake.com/en/sql-reference/data-types-text#single-quoted-string-constants
+		// replace all single quotes with \'
+		escapedString := strings.ReplaceAll(s, qm.String(), `\'`)
```
sfc-gh-ngaberel, line 42: Single quotes use the same escaping strategy as double quotes, they're doubled.
```suggestion
		escapedString := strings.ReplaceAll(s, qm.String(), qm.String()+qm.String())
```

pkg/sdk/warehouse.go
sfc-gh-ngaberel: Yes they are. :)
https://docs.snowflake.com/en/sql-reference/sql/create-warehouse
image
I was even able to use that in the "unset tags" integration tests.

pkg/sdk/warehouse.go
```diff
@@ -135,81 +134,102 @@ type WarehouseAlterOptions struct {
...
+				return fmt.Errorf("AutoSuspend must be greater than or equal to 0")
+			}
+		}
+		if opts.Set.QueryAccelerationMaxScaleFactor != nil {
+			if ok := validateIntInRange(*opts.Set.QueryAccelerationMaxScaleFactor, 0, 100); !ok {
+				return fmt.Errorf("QueryAccelerationMaxScaleFactor must be between 0 and 100")
+			}
+		}
+		if valueSet(opts.Set.Tag) && !everyValueNil(opts.Set.AutoResume, opts.Set.EnableQueryAcceleration, opts.Set.MaxClusterCount, opts.Set.MinClusterCount, opts.Set.AutoSuspend, opts.Set.QueryAccelerationMaxScaleFactor) {
+			return fmt.Errorf("Tag cannot be set with any other Set parameter")
```
sfc-gh-ngaberel, line 231: Same for unset

pkg/sdk/warehouse_test.go
sfc-gh-ngaberel: image

pkg/sdk/warehouse.go
```diff
@@ -135,81 +135,107 @@ type WarehouseAlterOptions struct {
...
+		}
+	}
+	if opts.Unset != nil {
+		if valueSet(opts.Unset.Tag) && !everyValueNil(opts.Unset.AutoResume, opts.Unset.EnableQueryAcceleration, opts.Unset.MaxClusterCount, opts.Unset.MinClusterCount, opts.Unset.AutoSuspend, opts.Unset.QueryAccelerationMaxScaleFactor) {
+			return fmt.Errorf("Tag cannot be set with any other Unset parameter")
```
sfc-gh-ngaberel, line 237:
```suggestion
			return fmt.Errorf("Tag cannot be unset with any other Unset parameter")
```

pkg/sdk/warehouse_test.go
```diff
@@ -95,20 +96,19 @@ func TestWarehouseAlter(t *testing.T) {
...
-				// 	NewSchemaObjectIdentifier("db1", "schema1", "tag1"),
-				// 	NewSchemaObjectIdentifier("db2", "schema2", "tag2"),
-				// },
+			Unset: &WarehouseUnset{
+				WarehouseSize:   Bool(true),
+				MaxClusterCount: Bool(true),
+				AutoResume:      Bool(true),
+				Tag: []ObjectIdentifier{
+					NewSchemaObjectIdentifier("db", "schema", "tag1"),
+				},
```
sfc-gh-ngaberel, line 105: Doesn't that return an error now?

sfc-gh-ngaberel (approved):

pkg/sdk/warehouse_test.go
sfc-gh-swinkler:
```
--- PASS: TestWarehouseAlter (0.00s)
    --- PASS: TestWarehouseAlter/with_unset (0.00s)
PASS
ok  	github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk	0.413
```
it doe not error

sfc-gh-ngaberel (approved):

PR #1807, opened 2023-05-16
feat: Grants sdkv2

sfc-gh-swinkler:
adds a few endpoints for grants. particularly, grant privilege and revoke privilege from share, and show grants.


pkg/sdk/grants.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (opts *grantPrivilegeToShareOptions) validate() error {
+	if opts.to.FullyQualifiedName() == "" {
+		return fmt.Errorf("to is required")
+	}
+	if opts.On == nil || opts.objectPrivilege == "" {
+		return fmt.Errorf("on and objectPrivilege are required")
+	}
+	return nil
```
sfc-gh-ngaberel, line 85: We should also check that "On" only has one field, right?

sfc-gh-ngaberel (approved):

PR #1813, opened 2023-05-17
feat: add shares to sdk v2

sfc-gh-swinkler:
adding shares to sdk v2


pkg/sdk/databases.go
```diff
@@ -1,12 +1,103 @@
...
+
+type Databases interface {
+	// Create creates a database.
+	Create(ctx context.Context, id AccountObjectIdentifier, opts *DatabaseCreateOptions) error
+	// Alter modifies an existing database
+	Alter(ctx context.Context, id AccountObjectIdentifier, opts *DatabaseAlterOptions) error
+	// Drop removes a database.
+	Drop(ctx context.Context, id AccountObjectIdentifier, opts *DatabaseDropOptions) error
+	// Show returns a list of databases.
+	Show(ctx context.Context, opts *DatabaseShowOptions) ([]*Database, error)
```
sfc-gh-ngaberel, line 16: Can you add ShowById here so we don't forget to add it later?

pkg/resources/share.go
```diff
@@ -112,92 +123,127 @@ func setAccounts(d *schema.ResourceData, meta interface{}) error {
...
+			if err != nil {
+				return fmt.Errorf("error removing accounts from share err = %w", err)
+			}
+		} else {
+			accountIdentifiers := accountIdentifiersFromSlice(newAccounts)
+			err := setShareAccounts(ctx, client, sdk.NewAccountObjectIdentifier(d.Id()), accountIdentifiers)
+			if err != nil {
+				return err
+			}
+		}
```
sfc-gh-ngaberel, line 222: If an account has been removed in the terraform config then this won't remove it from the share, this only adds accounts to the share. Correct? I think there's a bug there.

pkg/datasources/shares.go
```diff
@@ -63,36 +63,35 @@ func Shares() *schema.Resource {
...
-	listShares, err := snowflake.ListShares(db, sharePattern)
-	if errors.Is(err, sql.ErrNoRows) {
-		log.Printf("[DEBUG] no shares found in account (%s)", d.Id())
-		d.SetId("")
-		return nil
-	} else if err != nil {
-		log.Println("[DEBUG] failed to list shares")
-		d.SetId("")
-		return nil
+	pattern := d.Get(pattern).(string)
```
sfc-gh-ngaberel, line 66: I don't understand this, how can `pattern` depend on itself?

pkg/resources/share.go
```diff
@@ -63,45 +59,60 @@ func Share() *schema.Resource {
...
+		for i, account := range accounts {
+			parts := strings.Split(account, ".")
+			orgName := parts[0]
+			accountName := parts[1]
+			accountIdentifiers[i] = sdk.NewAccountIdentifier(orgName, accountName)
+		}
+		err := setShareAccounts(ctx, client, shareID, accountIdentifiers)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
sfc-gh-swinkler: not sure. i am just keeping the functionality as it was. i think since set accounts is part of the shares API, it should not be a separate resource.

pkg/resources/share.go
sfc-gh-swinkler: according to documentation set accounts does the following:
> Specifies the account(s) to replace all previous accounts with which the share was shared. T
so yes it will remove accounts. unless the list is zero length, in which case i would have it 
we already have a test that goes from 1 account down to 0 accounts. I added a test that goes from 2 -> 1 also. Noticed there was an issue with the ordering of the result (fixed), but it does correctly remove accounts

pkg/datasources/shares.go
sfc-gh-swinkler: if pattern was set, or not empty string then we do `SHOW SHARES LIKE '<pattern>'` otherwise we do `SHOW SHARES`. theres no dependency

sfc-gh-ngaberel: confusing variable nane

sfc-gh-ngaberel (approved):

Issue #1691, opened 2023-04-05
0.61.0 -  unexpected number of ID parts (1)


thomassandslyst:
**Provider Version**

0.61.0

**Terraform Version**

1.2.5

**Describe the bug**

When planning using 0.61.0 I'm getting "unexpected number of ID parts (1)" errors on all snowflake grant resources. These resources were last applied with 0.56.3

**Expected behavior**

No errors

**Code samples and commands**

An example of a resource that is failing with "unexpected number of ID parts (1), expected 6":
```hcl
resource "snowflake_stage_grant" "paid_branch_usage" {
  database_name = snowflake_database.lyst.name
  schema_name   = snowflake_schema.lyst_analytics.name
  stage_name    = snowflake_stage.paid_branch.name
  privilege     = "USAGE"
  roles = [
    snowflake_role.dbt.name,
  ]
}
```

**Additional context**

The tfstate for the above resource:
```json
    {
      "mode": "managed",
      "type": "snowflake_stage_grant",
      "name": "paid_branch_usage",
      "provider": "provider[\"registry.terraform.io/snowflake-labs/snowflake\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "database_name": "LYST",
            "enable_multiple_grants": null,
            "id": "LYST|LYST_ANALYTICS|PAID_BRANCH|USAGE",
            "on_future": false,
            "privilege": "USAGE",
            "roles": [
              "DBT"
            ],
            "schema_name": "LYST_ANALYTICS",
            "stage_name": "PAID_BRANCH",
            "with_grant_option": false
          },
          "sensitive_attributes": [],
          "private": "<redacted>",
          "dependencies": [
            "snowflake_database.lyst",
            "snowflake_role.dbt",
            "snowflake_schema.lyst_analytics",
            "snowflake_stage.paid_branch"
          ]
        }
      ]
    }
```

Looks to be related to https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1642

ivica-k:
Seems to be fixed in `0.64.0`.

PR #1816, opened 2023-05-17
fix: fixed enable_multiple_grants behaviour for role_grants

Shanjohn:
Previously enable_multiple_grants parameter for the snowflake_role_grants resource did not revoke any permissions given outside of the terraform. My fix actually enables such behavior: if parameter is set to false, every role/user from `SHOW GRANTS OF ROLE "%s"` is being accounted by the resource.

Since everyone who used the resource previously had resource acting as the enable_multiple_grants were set to true, I suggest we change the default for this parameter to true as well, in order not to break existing configurations. Only the ones who set the parameter explicitly will see the change.

## Test Plan
I added couple new unit tests around it, I don't know if any new acceptance tests are required.

### DIFF
--- a/docs/resources/role_grants.md
+++ b/docs/resources/role_grants.md
@@ -3,11 +3,12 @@
 page_title: "snowflake_role_grants Resource - terraform-provider-snowflake"
 subcategory: ""
 description: |-
-  
+
 ---
 
 # snowflake_role_grants (Resource)
 
+!> **WARNING:** If parameter `enable_multiple_grants` is set to `false`, the snowflake_role_grants resource creates **exclusive** grants for the selected role. Creating multiple resources for the same role with `enable_multiple_grants = false` will lead to permanent changes in plan.
 
 
 ## Example Usage
@@ -55,7 +56,7 @@ resource "snowflake_role_grants" "grants" {
 
 ### Optional
 
-- `enable_multiple_grants` (Boolean) When this is set to true, multiple grants of the same type can be created. This will cause Terraform to not revoke grants applied to roles and objects outside Terraform.
+- `enable_multiple_grants` (Boolean) When this is set to false, multiple grants of the same type cannot be created. This will cause Terraform to revoke grants applied to roles and objects outside Terraform.
 - `roles` (Set of String) Grants role to this specified role.
 - `users` (Set of String) Grants role to this specified user.
 
--- a/pkg/resources/role_grants.go
+++ b/pkg/resources/role_grants.go
@@ -49,8 +49,8 @@ func RoleGrants() *schema.Resource {
 			"enable_multiple_grants": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Description: "When this is set to true, multiple grants of the same type can be created. This will cause Terraform to not revoke grants applied to roles and objects outside Terraform.",
-				Default:     false,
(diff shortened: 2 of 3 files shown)


sfc-gh-swinkler:
@Shanjohn good point - everyone was setting this to true anyways. Thanks for your contribution.

ivica-k:
I believe this was a breaking change for the organization I work in, please correct me if i'm wrong. Granted, we did not have the provider version pinned - that's on us, but this change removed all users synchronized from Azure AD that were assigned to roles in Snowflake.

Screenshot 2023-05-30 at 15 39 10


Be careful if you have a similar setup and always pin your provider versions.

attekei:
> We are using Okta to provision our users, and no matter the value of enable_multiple_grants all of the users are being removed from externally managed role grants as of v0.66.1

Same here, seems like this change is not working as expected. Also, older provider versions aren't working anymore, so I'm effectively unable to provision any resources at the moment.

Relativity74205:
@Shanjohn @sfc-gh-swinkler 
My interpretation of `enable_multiple_grants` was actually different. It simply allowed multiple grants of the same resource within terraform and this worked quite well. And I don't see why terraform should have the option/power to revoke grants, which were granted outside of terraform. This should be totally out of scope of the provider, because the problems resulting out of this are totally unforseenable. And this change obviously broke a lot of stuff.

I would suggest to reverse this change.

Shanjohn:
@Relativity74205 the documentation implied that Terraform should be able to revoke permissions outside of its scope:
> `enable_multiple_grants` - This will cause Terraform to not revoke grants applied to roles and objects outside Terraform.
And in my case it was functionality I really needed because I wanted to be sure that my terraform configuration is the only source of truth. Do you mean that you wanted to ensure there is a single resource in terraform configuration for a given role? What if the role is in multiple different configurations?

Anyway, prior to this change `enable_multiple_grants` option was basically always set to `True` despite its value (and for all default configurations. That's why I suggested to change default to `True`.

@ivica-k @jakemingolla I think the root cause of the issue is not that the value changed to `True`, but because of the changes order: the list of grants is actually processed based on current value of `enable_multiple_grants` (which was `False` by default), not on the future one (`True`)

@sfc-gh-swinkler I agree with the reversing the change because of breaking changes, but I don't agree that it is not needed at all. I will look into the order of operations to see if it is possible resolve the issue above

Shanjohn:
@attekei could you please elaborate on older provider version not working?

attekei:
@Shanjohn These were the issues I hit:
0.63-> or so: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1868 
Prior to 0.64: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1691

Please note that this could be an interim solution to the first one, letting you safely downgrade:
https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1868#issuecomment-1585576889

Relativity74205:
@Shanjohn I see your point, also that such a feature might be useful. In general we have the same requirement, however, a few grants are done by AzureAD through SCIM.

Up to know I have only operated based on the first part of the docs:

> [enable_multiple_grants](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/role_grants#enable_multiple_grants) (Boolean) When this is set to false, multiple grants of the same type cannot be created. [...]

And we have been using this feature, e.g. we grant usage on a certain database in several places in terraform.

PR #1822, opened 2023-05-22
chore: Account SDK

sfc-gh-ngaberel:
Implements Accounts API in new SDK and use it in account resource.

## Test Plan
* [X] SQL unit tests
* [ ] no integration or acceptance tests :/

### DIFF
--- a/pkg/sdk/accounts.go
+++ b/pkg/sdk/accounts.go
@@ -1,0 +1,375 @@
+package sdk
+
+import (
+	"context"
+	"fmt"
+	"time"
+)
+
+type Accounts interface {
+	// Create creates an account.
+	Create(ctx context.Context, name AccountObjectIdentifier, opts *AccountCreateOptions) error
+	// Alter modifies an existing account
+	Alter(ctx context.Context, opts *AccountAlterOptions) error
+	// Show returns a list of accounts.
+	Show(ctx context.Context, opts *AccountShowOptions) ([]*Account, error)
+	// ShowByID returns an account by ID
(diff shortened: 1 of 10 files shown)


pkg/sdk/accounts.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+	sql, err := structToSQL(opts)
+	if err != nil {
+		return err
+	}
+	_, err = c.client.exec(ctx, sql)
+	return err
+}
+
+type AccountShowOptions struct {
```
sfc-gh-swinkler, line 261: Since there is also a "SHOW MANAGED ACCOUNTS" command it would make more sense to called this ShowOrganizationAccounts or something like that

pkg/sdk/comments_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestComments(t *testing.T) {
```
sfc-gh-swinkler, line 11: We are starting to get more and more functions. Wondering if we should put these into a separate package to make it more clear what these are or if it is sufficient to call them "XXXFunctions". If the latter, then perhaps this should be in the "GeneralDDLFunctions" file. Seems a bit much to make it its own high level thing, but also it is kinda unique. We can discuss this

sfc-gh-swinkler (changes requested):

sfc-gh-swinkler (approved):

sfc-gh-swinkler (approved):

Issue #1746, opened 2023-04-24
Password Policy Attachment


marenronneseth:
**Is your feature request related to a problem? Please describe.**

Not related to a problem just that Password Policy Attachment in Snowflake isn't available through Terraform.

**Describe the solution you'd like**

I would like to be able to attache Password Policy to an Snowflake account through Terraform . We manage objects like this through Terraform so being able to create and manage this through TF would be helpful.

**Describe alternatives you've considered**

It just isn't available so I don't know of any alternatives

**Additional context**

Link to Snowflake documentation: https://docs.snowflake.com/en/sql-reference/sql/alter-account

sfc-gh-ngaberel:
Hi @marenronneseth, I've added a new `snowflake_account_password_policy_attachment` resource to the provider to support this use case, it will be available with the [next release](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1880), probably by next week.
Please note that, due to a limitation of the SQL API, this resource can only change the password policy of the **current account** (ie. the one specified in the provider block). If you need to change the password policy on a different account you'll need to use an [aliases provider](https://developer.hashicorp.com/terraform/language/providers/configuration#alias-multiple-provider-configurations).

PR #1824, opened 2023-05-23
feat: account password policy attachment

sfc-gh-ngaberel:
Adds support for attaching a password policy to the current account.

Due to limitations of the underlying API, this feature only supports setting the password policy of the **current account** (ie. the one from the `provider snowflake {}` block). To set the password policy of a different account, configure an aliased provider for that account.

## Test Plan
* [X] acceptance tests

## References

* Implements #1746 
* Depends on #1822


pkg/resources/account_password_policy_attachment.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var accountPasswordPolicyAttachmentSchema = map[string]*schema.Schema{
+	"password_policy": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: "Qualified name (`\"db\".\"schema\".\"policy_name\"`) of the password policy to apply to the current account.",
```
sfc-gh-swinkler, line 18: if we are going to do this, we should probably make data source(s) to help make it easier for people to make fully qualified names from db / schema / object names

pkg/resources/account_password_policy_attachment.go
```diff
@@ -0,0 +1,85 @@
...
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var accountPasswordPolicyAttachmentSchema = map[string]*schema.Schema{
```
sfc-gh-swinkler, line 13: should this be renamed to "password_policy_account_attachment"?

sfc-gh-ngaberel: password policies can be attached to either accounts or users, so I figured we'd have a separate resource later user_password_policy_attachment

pkg/resources/account_password_policy_attachment.go
sfc-gh-ngaberel: I added the qualified name as an attribute of the password policy resource, but the data source is a good idea too.

docs/resources/password_policy.md
```diff
@@ -39,5 +39,6 @@ A password policy specifies the requirements that must be met to create and rese
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+- `qualified_name` (String) The qualified name for the password policy.
```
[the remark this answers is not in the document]
sfc-gh-ngaberel: This matches the attribute on masking policies, but we can change them.

PR #1826, opened 2023-05-23
feat: set query tag for terraform sessions 

DmitryMaletin:
Set query_tag  from env variable  to terraform sessions 
## Usage
export SNOWFLAKE_QUERY_TAG="test" && terraform apply

## References
[Query_tag parameter](https://docs.snowflake.com/en/sql-reference/parameters#query-tag)
[Connection parameters](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#hdr-Connection_Parameters)
*


sfc-gh-swinkler:
I do think that the provider configuration needs some refactoring, but i am not sure if adding query_tags is the way to go. Why not set this at the account level instead of session level? Curious what your use case is.

DmitryMaletin:
I use terraform in CI/CD pipeline  triggered on PR. Set query tag  on sessions level allow to see when objects were deployed.

DmitryMaletin:
@sfc-gh-swinkler 

> I do think that the provider configuration needs some refactoring, but i am not sure if adding query_tags is the way to go
I think they are not related. If there are any other way to set session properties?

sfc-gh-swinkler:
@DmitryMaletin you can set the Session properties using the Sessions manager. i.e. `client.Sessions.AlterSession(...)`

Sorry for the delay. I have been thinking about this PR and the conclusion that I came to is that any changes to the provider configuration should match what is in the gosnowflake connector as best as possible. So since you are wanting to change the Params, then you should support map[string]string, not just one for query_tag. e.g your provider configuration could look like:

```
provider "snowflake" {
...
params = {
 "query_tag" = <query_tag_value>
"some_other_param" = <some_other_param_value
}
}
```

You also don't need to set an environment variable for this, i think its fine as is. If you can make this change then I will accept the PR.

sfc-gh-swinkler:
Also have you verified that this actually works as intended?I have not used the params field before, so unsure how it works. Can you please provide some results to demonstrate the behavior?

DmitryMaletin:
@sfc-gh-swinkler 


> @DmitryMaletin you can set the Session properties using the Sessions manager. i.e. `client.Sessions.AlterSession(...)`
> 
> Sorry for the delay. I have been thinking about this PR and the conclusion that I came to is that any changes to the provider configuration should match what is in the gosnowflake connector as best as possible. So since you are wanting to change the Params, then you should support map[string]string, not just one for query_tag. e.g your provider configuration could look like:
> 
> ```
> provider "snowflake" {
> ...
> params = {
>  "query_tag" = <query_tag_value>
> "some_other_param" = <some_other_param_value
> }
> }
> ```
> 
> You also don't need to set an environment variable for this, i think its fine as is. If you can make this change then I will accept the PR.

Fixed.  I have added session_params as you proposed and removed Env. variable

DmitryMaletin:
> Also have you verified that this actually works as intended?I have not used the params field before, so unsure how it works. Can you please provide some results to demonstrate the behavior?

Yes. It's working as expected.  Use similar approach in python based project.

sfc-gh-swinkler:
@DmitryMaletin this looks really good, thanks!

PR #1828, opened 2023-05-24
feat: adding a subset of missing parameters

adamhealy:
There are many missing parameters from the provider that are in the snowflake [docs](https://docs.snowflake.com/en/sql-reference/parameters). The ones I've added here are those I have an urgent need for at the moment.

## Test Plan
Added a test for the parameter `CLIENT_METADATA_USE_SESSION_DATABASE` as it is not documented that it applies to the account as well as the session.
* [x] acceptance tests

## References
* [Parameters | Snowflake Documentation](https://docs.snowflake.com/en/sql-reference/parameters)


sfc-gh-swinkler (approved):
LGTM

sfc-gh-swinkler:
@adamhealy thank you for your contribution!

PR #1869, opened 2023-06-09
Added password_mfa option for mfa caching

sfc-gh-cconner:
We have had several reports of not being able to use MFA caching even though the go driver supports it.  THe reason is you must use a specific authenticator for password MFA:

AuthTypeUsernamePasswordMFA

This PR adds a "password_mfa" authentication option to the snowflake provider to utilize the above authenticator and allow for MFA caching.


sfc-gh-ngaberel:
Thanks @sfc-gh-cconner, do you think there is a way to write tests for this feature? I'm not sure whether we can automate mfa auth but worth looking into.

PR #1871, opened 2023-06-12
feat: alerts sdk v2

sfc-gh-pbosak:
* add alerts to sdk v2
* add unit and integration tests for alerts
* add new "condition" ddl type
## Test Plan

* [X] unit tests
* [X] integration tests


pkg/sdk/alert.go
```diff
@@ -0,0 +1,421 @@
+package sdk
```
sfc-gh-swinkler, line 1: this file should be names "alerts" for consistency

pkg/sdk/alert.go
```diff
@@ -0,0 +1,421 @@
...
+	Owner        string
+	Comment      *string
+	Warehouse    string
+	Schedule     AlertSchedule
+	State        string
+	Condition    string
+	Action       string
+}
+
+type AlertSchedule interface {
```
sfc-gh-swinkler, line 228: i think its an interesting idea to have this as a separate type.  I guess the reason i had settled on using strings is because its a string type in the API documentation.

Note that are actually other places which use schedules, for example [ Failover Groups](https://docs.snowflake.com/en/sql-reference/sql/create-failover-group) and [Replication Groups ](https://docs.snowflake.com/en/sql-reference/sql/create-replication-group). In the failover groups file, you can see i just treat it like a string instead of using a separate type. We could have a "type Schedule struct" which could have a String() method to convert it into a schedule. Alternatively i can edit the sql_builder to be able to allow wrapping keywords with quotes. Or we can just keep it as a *string. Let me know your preference. I do not like having a interface though, i think thats a little too different than the pattern we have been going with for the rest of the SDK.

pkg/sdk/alert.go
```diff
@@ -0,0 +1,421 @@
...
+	} else if withoutCron := dropUsingCron(parts); len(withoutCron) == 6 {
+		return AlertScheduleCronExpression{
+			Expression: strings.Join(withoutCron[0:len(withoutCron)-1], " "),
+			TimeZone:   withoutCron[len(withoutCron)-1],
+		}, true
+	} else {
+		return nil, false
+	}
+}
+func dropUsingCron(parts []string) []string {
```
sfc-gh-swinkler, line 246: lets try not have helper functions at the global level

pkg/sdk/alert.go
```diff
@@ -0,0 +1,421 @@
...
+	Owner        string    `db:"owner"`
+	Comment      *string   `db:"comment"`
+	Warehouse    string    `db:"warehouse"`
+	Schedule     string    `db:"schedule"`
+	State        string    `db:"state"` // suspended, started
+	Condition    string    `db:"condition"`
+	Action       string    `db:"action"`
+}
+
+func (row alertDBRow) toAlert() *Alert {
```
sfc-gh-swinkler, line 295: okay so the point of having separate alertDBRow and an Alert struct is so that we can transform the data to be more user friendly. For example, instead of just keeping state as a string, we could convert. it to an enum

pkg/sdk/alert.go
```diff
@@ -0,0 +1,421 @@
...
+	TimeZone   string
+}
+
+func (expression AlertScheduleCronExpression) String() string {
+	return fmt.Sprintf("USING CRON %v %v", expression.Expression, expression.TimeZone)
+}
+
+func (AlertScheduleCronExpression) alertSchedule() {}
+
+func (v *Alert) IsEnabled() bool {
```
sfc-gh-swinkler, line 273: these functions are not useful. we should instead have enum types, and as part of the toAlert() function, it should convert to that enum type.

pkg/sdk/alert.go
```diff
@@ -0,0 +1,421 @@
...
+}
+
+func (AlertScheduleInterval) alertSchedule() {}
+
+type AlertScheduleCronExpression struct {
+	Expression string
+	TimeZone   string
+}
+
+func (expression AlertScheduleCronExpression) String() string {
```
sfc-gh-swinkler, line 267: the more i am looking at this the more i think it is too different than what we have been doing with other structs. Can you just use *string for schedule? And make a note to revisit this later. I do think we can have Schedule as a type, but it should not be a special implementation compared to how we build SQL statements for everything else.

pkg/sdk/alert.go
```diff
@@ -0,0 +1,421 @@
...
+		Condition:    row.Condition,
+		Action:       row.Action,
+	}
+}
+
+func (v *alerts) Show(ctx context.Context, opts *ShowAlertOptions) ([]*Alert, error) {
+	if opts == nil {
+		opts = &ShowAlertOptions{}
+	}
+	sql, err := structToSQL(opts)
```
sfc-gh-swinkler, line 316: you are missing the validation check

pkg/sdk/alert.go
```diff
@@ -0,0 +1,421 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = v.client.exec(ctx, sql)
+	return err
+}
+
+type state string
```
sfc-gh-swinkler, line 86: a type called "state" is too generic. Lets call it "AlertState" instead

sfc-gh-swinkler (changes requested):

pkg/sdk/alert.go
sfc-gh-pbosak: okay

sfc-gh-pbosak: okay

sfc-gh-pbosak: Ok

pkg/sdk/alert.go
sfc-gh-pbosak: Ok, I was lookin

sfc-gh-pbosak: Ok, I was lookin

sfc-gh-pbosak: Ok, I was looking at masking_policy

sfc-gh-pbosak: Ok, I was looking at masking_policy

sfc-gh-pbosak: Ok

sfc-gh-pbosak: Ok

sfc-gh-pbosak: Ok

pkg/sdk/alert.go
sfc-gh-pbosak: Yes, I did not notice other resources

sfc-gh-pbosak: Test,
test

sfc-gh-pbosak: Test,
test

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for type

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for type

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding some type safety

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding some type safety

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding some

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding some

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I adde

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the \"cleaner\"

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the \"cleaner\"

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the cleanest way of encoding sum types in Go that I know of, but I think a struct will do.

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the cleanest way of encoding sum types in Go that I know of, but I think a struct will do.

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the cleanest way of encoding sum types in Go that I know of, but I think a struct will do.
side note: In `Create`

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the cleanest way of encoding sum types in Go that I know of, but I think a struct will do.
side note: In `Create`

sfc-gh-pbosak: Fir

sfc-gh-pbosak: Fir

sfc-gh-pbosak: Test comment

sfc-gh-pbosak: Test comment

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I added

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I added

sfc-gh-pbosak: abc

sfc-gh-pbosak: abc

sfc-gh-pbosak: Test comment
Test commentt

sfc-gh-pbosak: Test comment
Test commentt

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the cleanest way of encoding sum types in Golang that I know of, but a struct will do too.
side note: In `Create` I can use `AlertSchedule` type, but in `CreateAlertOptions` I'm forced to use string.
What do you think about updating sql_builder so it also can handle types implementing `Stringer` interface?

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the cleanest way of encoding sum types in Golang that I know of, but a struct will do too.
side note: In `Create` I can use `AlertSchedule` type, but in `CreateAlertOptions` I'm forced to use string.
What do you think about updating sql_builder so it also can handle types implementing `Stringer` interface?

pkg/sdk/alert.go
sfc-gh-pbosak: ok

sfc-gh-pbosak: ok

pkg/sdk/alert.go
sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the cleanest way of encoding sum types in Golang that I know of, but a struct will do too.
side note: In `Create` I can use `AlertSchedule` type, but in `CreateAlertOptions` I'm forced to use string.
What do you think about updating sql_builder so it also can handle types implementing `Stringer` interface?
If `AlertSchedule` was not required, it would not be a separate argument in the function, so even in the interface we would 
just have a plain string.

sfc-gh-pbosak: Yes, I did not notice it's used by other resources
In general, I'm all for adding more types.
I used an interface because that's the cleanest way of encoding sum types in Golang that I know of, but a struct will do too.
side note: In `Create` I can use `AlertSchedule` type, but in `CreateAlertOptions` I'm forced to use string.
What do you think about updating sql_builder so it also can handle types implementing `Stringer` interface?
If `AlertSchedule` was not required, it would not be a separate argument in the function, so even in the interface we would 
just have a plain string.

pkg/sdk/alert.go
sfc-gh-pbosak: I'm not sure what \"global\" means in this context and where I should put it instead. 
I could put in some `helpers.go` inside sdk package, but this is an auxiliary function meant to be used only by `AlertScheduleFromString`

sfc-gh-pbosak: I'm not sure what \"global\" means in this context and where I should put it instead. 
I could put in some `helpers.go` inside sdk package, but this is an auxiliary function meant to be used only by `AlertScheduleFromString`

sfc-gh-pbosak: I'm not sure what \"global\" means in this context and where I should put it instead. 
I could put in some `helpers.go` inside sdk package, but this is an auxiliary function meant to be used only by `AlertScheduleFromString`

sfc-gh-pbosak: I'm not sure what \"global\" means in this context and where I should put it instead. 
I could put in some `helpers.go` inside sdk package, but this is an auxiliary function meant to be used only by `AlertScheduleFromString`

pkg/sdk/alert.go
sfc-gh-pbosak: fixed

pkg/sdk/alert.go
sfc-gh-pbosak: fixed

sfc-gh-pbosak: fixed

pkg/sdk/alert.go
sfc-gh-swinkler: so because SDK is one big package, any function you create will be available for use anywhere in the package. Thats a lot of files. I would suggest either using an anonymous function or creating a "helper" struct and putting the functions in that struct so at least they are all in one place and not cluttering the namespace.

pkg/sdk/alerts.go
```diff
@@ -0,0 +1,382 @@
...
+	alter    bool                   `ddl:"static" sql:"ALTER"` //lint:ignore U1000 This is used in the ddl tag
+	alert    bool                   `ddl:"static" sql:"ALERT"` //lint:ignore U1000 This is used in the ddl tag
+	IfExists *bool                  `ddl:"keyword" sql:"IF EXISTS"`
+	name     SchemaObjectIdentifier `ddl:"identifier"`
+
+	// One of
+	Operation       *AlertOperation `ddl:"keyword"`
+	Set             *AlertSet       `ddl:"keyword" sql:"SET"`
+	Unset           *AlertUnset     `ddl:"keyword" sql:"UNSET"`
+	ModifyCondition *string         `ddl:"condition,parentheses" sql:"MODIFY CONDITION EXISTS"`
```
sfc-gh-swinkler, line 111: again, this could be done using a list type. or adding parentheses modifier to parameter type. so then it would be like `ddl:"parameter,parantheses,no_equals" sql:"MODIFY CONDITION EXISTS"`

pkg/sdk/alerts.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	sql, err := structToSQL(opts)
+	if err != nil {
+		return err
+	}
+	_, err = v.client.exec(ctx, sql)
+	return err
+}
+
+type DropAlertOptions struct {
```
sfc-gh-swinkler, line 168: this could be make private since it shouldn't be used by outside consumers. e.g. `dropAlertOptions`

sfc-gh-swinkler (changes requested):

pkg/sdk/alerts.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	sql, err := structToSQL(opts)
+	if err != nil {
+		return err
+	}
+	_, err = v.client.exec(ctx, sql)
+	return err
+}
+
+type AlertOperation string
```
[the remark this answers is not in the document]
sfc-gh-pbosak: I went with operation as we already have a `modify action`, but it makes sense to stay with the documentation terms

pkg/sdk/alerts.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// required
+	warehouse AccountObjectIdentifier `ddl:"identifier,equals" sql:"WAREHOUSE"`
+	schedule  string                  `ddl:"parameter,single_quotes" sql:"SCHEDULE"`
+
+	// optional
+	Comment *string `ddl:"parameter,single_quotes" sql:"COMMENT"`
+
+	// required
+	condition AlertCondition `ddl:"condition,parentheses"   sql:"IF"`
```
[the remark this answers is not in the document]
sfc-gh-pbosak: Yes, I've considered all the options you mention, but all of them felt slightly "hacky" to me.
I will go with the `list` option

pkg/sdk/alerts.go
sfc-gh-pbosak: fixed using `list` type, but that means instead of:
```
alterOptions.ModifyCondition = &condition
```
I need to do 
```
alterOptions.ModifyCondition = &[]string{condition}
```
In this case it's not a big problem, but I was wondering if you considered using separate types for constructing sql statements(structs with `ddl` tags) and for arguments inside client's interface.

Structs with ddl struct tags need to have a certain structure so that we can turn them into SQL statements, but since the same type is used in client's interface, we allow users of that interface to make a mistake, e.g. passing multiple conditions to `ModifyCondition`

I think this might prove useful if we were to convert sdk package into a stand-alone library in the future.
I'm not saying we should make this change now, but I am curious what are your thoughts on it.

pkg/resources/alert.go
```diff
@@ -264,169 +225,152 @@ func ReadAlert(d *schema.ResourceData, meta interface{}) error {
...
+		alterOptions.Set = &sdk.AlertSet{
+			Schedule: &alertSchedule,
+		}
+		err := client.Alerts.Alter(ctx, objectIdentifier, alterOptions)
+		if err != nil {
+			return fmt.Errorf("error updating schedule on alert %v", objectIdentifier.Name())
 		}
 	}
 
 	if d.HasChange("comment") {
```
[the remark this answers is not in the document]
sfc-gh-pbosak: fixed

pkg/sdk/alerts.go
```diff
@@ -0,0 +1,382 @@
...
+	// Resume makes a suspended alert active.
+	Resume AlertAction = "RESUME"
+	// Suspend puts the alert into a “Suspended” state.
+	Suspend AlertAction = "SUSPEND"
+)
+
+type AlertState string
+
+var (
+	AlertStarted   AlertState = "started"
```
sfc-gh-swinkler, line 97: This should be AlertStateStarted

pkg/sdk/alerts.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	_, err = v.client.exec(ctx, sql)
+	return err
+}
+
+type AlertAction string
+
+var (
+	// Resume makes a suspended alert active.
+	Resume AlertAction = "RESUME"
```
sfc-gh-swinkler, line 89: This Should be called AlertActionResume

sfc-gh-swinkler (changes requested):

sfc-gh-swinkler (approved):

sfc-gh-swinkler:
LGTM!

Issue #1868, opened 2023-06-08
err = missing destination name kind in *[]snowflake.Database


rjoelnorgren:
**Provider Version**

0.64

**Terraform Version**

1.4.6

**Describe the bug**
Snowflake [release 7.13 adds the kind column ](https://community.snowflake.com/s/article/SHOW-DATABASES-Command-New-Column-in-Output) to the show databases command. This behavior change was [enabled by default in the 7.19 release](https://community.snowflake.com/s/article/Pending-Behavior-Change-Log#disabled):
```sql
select current_version();
-- +-------------------+                                                           
-- | CURRENT_VERSION() |
-- |-------------------|
-- | 7.19.2            |
-- +-------------------+
```

I believe this is now causing the state refresh for our existing databases to fail:

```
Error: unable to scan row for SHOW DATABASES LIKE 'ANALYTICS' err = missing destination name kind in *[]snowflake.Database

  with snowflake_database.analytics,
  on databases.tf line 1, in resource "snowflake_database" "analytics":
   1: resource "snowflake_database" "analytics" {
```

**Expected behavior**

The state of the database refreshes successfully.

**Code samples and commands**

```
resource "snowflake_database" "analytics" {
  name                        = "ANALYTICS"
  data_retention_time_in_days = 14
  is_transient                = false
}
```

**Additional context**
A patch for this change would be appreciated. In particular, https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1845 is a major friction to us upgrading beyond 0.64.

rjoelnorgren:
As a work around, we have disabled bundle 2023_03 which appears to have resolved the issue for now:
```sql
select system$disable_behavior_change_bundle('2023_03');
```

Assuming this is fixed in a release before the opt out period has ended, we ideally would prefer to not have to remove and re-import most of the role grants in our account from terraform state to upgrade (https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1845).

jackwillisupside:
Hitting this same issue!

walarue:
also hitting this, it's stopping our whole terraform configuration

mcurtisslalom:
Hitting the same issue on 0.63

yanlonghe:
Hitting the same issue on terraform: 1.4.6 and snowflake 0.64

willwooten:
same issue 0.61

skwon615:
[...]

PR #1872, opened 2023-06-12
fix: missing 'kind' field

mlorek:
missing 'kind' field in database struct leading to:
_Error: unable to scan row for SHOW DATABASES LIKE 'SCRATCH' err = missing destination name kind in *[]snowflake.Databas_

## Test Plan
N/A
* [ ] acceptance tests
N/A
* [ ] …

## References

#1868
*


boblannon-fp:
Is this fix likely to release this week or next? It's currently blocking our terraform work even though we haven't made any database changes.  If it's likely to stay open for a while, we'll need to explore rolling back the behavior change on our snowflake account, but we'd really rather not...

sfc-gh-swinkler (approved):

PR #1892, opened 2023-06-19
feat: resource monitor sdk v2

sfc-gh-pbosak:
* refactored resource monitor to sdk v2
* added unit and integration tests
* minor fix in masking policy tests

## Test Plan

* [X] unit tests
* [X] acceptance tests

* [ ] …


sfc-gh-swinkler (changes requested):

pkg/sdk/resource_monitors.go
```diff
@@ -24,17 +29,114 @@ type resourceMonitors struct {
...
+	SetForAccount            bool
+	Comment                  string
+	NotifyUsers              []string
 }
 
 type resourceMonitorRow struct {
-	Name string `db:"name"`
+	Name               string         `db:"name"`
+	CreditQuota        sql.NullString `db:"credit_quota"`
+	UsedCredits        sql.NullString `db:"used_credits"`
```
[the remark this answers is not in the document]
sfc-gh-pbosak: fixed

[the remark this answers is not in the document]
sfc-gh-pbosak: fixed

pkg/sdk/resource_monitors.go
```diff
@@ -75,29 +197,126 @@ func (v *resourceMonitors) Create(ctx context.Context, id AccountObjectIdentifie
 	return err
 }
 
+type TriggerDefinition struct {
+	on            bool          `ddl:"static" sql:"ON"`      //lint:ignore U1000 This is used in the ddl tag
+	Threshold     int           `ddl:"keyword"`              //lint:ignore U1000 This is used in the ddl tag
```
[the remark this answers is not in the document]
sfc-gh-pbosak: fixed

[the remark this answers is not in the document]
sfc-gh-pbosak: fixed

pkg/sdk/resource_monitors.go
```diff
@@ -47,9 +173,18 @@ func (v *ResourceMonitor) ObjectType() ObjectType {
 
 // CreateResourceMonitorOptions contains options for creating a resource monitor.
 type CreateResourceMonitorOptions struct {
-	create          bool                    `ddl:"static" sql:"CREATE"`           //lint:ignore U1000 This is used in the ddl tag
+	create          bool                    `ddl:"static" sql:"CREATE"` //lint:ignore U1000 This is used in the ddl tag
+	OrReplace       *bool                   `ddl:"keyword" sql:"OR REPLACE"`
 	resourceMonitor bool                    `ddl:"static" sql:"RESOURCE MONITOR"` //lint:ignore U1000 This is used in the ddl tag
 	name            AccountObjectIdentifier `ddl:"identifier"`
+	With            *bool                   `ddl:"keyword" sql:"WITH"`
```
sfc-gh-swinkler, line 180: With is a static value, i think. Its not a required or optional parameter, its just there.

pkg/sdk/resource_monitors.go
```diff
@@ -47,9 +173,18 @@ func (v *ResourceMonitor) ObjectType() ObjectType {
...
+
+	CreditQuota    *int                 `ddl:"parameter,equals" sql:"CREDIT_QUOTA"`
+	Frequency      *Frequency           `ddl:"parameter,equals" sql:"FREQUENCY"`
+	StartTimeStamp *string              `ddl:"parameter,equals,single_quotes" sql:"START_TIMESTAMP"`
```
sfc-gh-swinkler, line 184: minor point but Timestamp is typically written as one word. so this should be `StartTimestamp` and `EndTimestamp`

pkg/sdk/resource_monitors.go
```diff
@@ -24,17 +29,138 @@ type resourceMonitors struct {
...
 type ResourceMonitor struct {
-	Name string
+	Name                     string
+	CreditQuota              float64
+	UsedCredits              float64
+	RemainingCredits         float64
+	Frequency                Frequency
+	StartTime                string
+	EndTime                  string
+	SuspendTriggers          []TriggerDefinition
```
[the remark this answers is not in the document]
sfc-gh-swinkler: it may make sense to have this just be an int. i.e. `SuspendAt *int`

pkg/sdk/resource_monitors_integration_test.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+func TestInt_ResourceMonitorsShow(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	resourceMonitorTest, resourceMonitorCleanup := createResourceMonitor(t, client)
+	t.Cleanup(resourceMonitorCleanup)
+
+	_, resourceMonitorCleanup2 := createResourceMonitor(t, client)
```
sfc-gh-swinkler, line 18: why do you create two resource montiors if you only use one of them?

pkg/sdk/resource_monitors.go
```diff
@@ -75,29 +221,131 @@ func (v *resourceMonitors) Create(ctx context.Context, id AccountObjectIdentifie
...
+	}
+	_, err = v.client.exec(ctx, sql)
+	return err
+}
+
+type ResourceMonitorSet struct {
+	// at least one
+	CreditQuota    *int       `ddl:"parameter,equals" sql:"CREDIT_QUOTA"`
+	Frequency      *Frequency `ddl:"parameter,equals" sql:"FREQUENCY"`
+	StartTimeStamp *string    `ddl:"parameter,equals,single_quotes" sql:"START_TIMESTAMP"`
```
sfc-gh-swinkler, line 329: should this be *time.Time?

pkg/sdk/resource_monitors_test.go
```diff
@@ -0,0 +1,169 @@
...
+			Frequency:      &frequency,
+			StartTimeStamp: &startTimeStamp,
+			EndTimeStamp:   &endTimeStamp,
+			NotifyUsers:    &NotifyUsers{notifiedUsers},
+			Triggers:       &triggers,
+		}
+
+		actual, err := structToSQL(opts)
+		require.NoError(t, err)
+		expected := fmt.Sprintf(`CREATE OR REPLACE RESOURCE MONITOR %s WITH CREDIT_QUOTA = %d FREQUENCY = %s START_TIMESTAMP = '%s' END_TIMESTAMP = '%s' NOTIFY_USERS = ("FIRST_USER", "SECOND_USER") TRIGGERS ON %d PERCENT DO %s ON %d PERCENT DO %s`,
```
sfc-gh-swinkler, line 55: not that its really an issue, but it may be easier to read to just have the expected be a hardcoded string instead of templated. Anyways the values are already known. It also saves you a bit of time because you can basically copy paste the result into expected instead of spending time to templating it.

sfc-gh-swinkler (changes requested):
mostly good, just a few minor things

pkg/sdk/resource_monitors.go
```diff
@@ -75,29 +221,131 @@ func (v *resourceMonitors) Create(ctx context.Context, id AccountObjectIdentifie
...
+
+type NotifiedUser struct {
+	Name string `ddl:"keyword,double_quotes"`
+}
+
+type Frequency string
+
+func FrequencyFromString(s string) (*Frequency, error) {
+	s = strings.ToUpper(s)
+	if monthly := string(FrequencyMonthly); monthly == s {
```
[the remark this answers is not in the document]
sfc-gh-pbosak: you're right, fixed

pkg/resources/resource_monitor.go
```diff
@@ -138,65 +137,68 @@ func CreateResourceMonitor(d *schema.ResourceData, meta interface{}) error {
...
+		users := []sdk.NotifiedUser{}
+		for _, name := range userNames {
+			users = append(users, sdk.NotifiedUser{Name: name})
+		}
+		opts.NotifyUsers = &sdk.NotifyUsers{Users: users}
 	}
+
 	if v, ok := d.GetOk("credit_quota"); ok {
-		cb.SetInt("credit_quota", v.(int))
+		opts.CreditQuota = sdk.Pointer(v.(int))
```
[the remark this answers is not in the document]
sfc-gh-pbosak: fixed

pkg/sdk/resource_monitors.go
```diff
@@ -24,17 +29,138 @@ type resourceMonitors struct {
...
-	Name string
+	Name                     string
+	CreditQuota              float64
+	UsedCredits              float64
+	RemainingCredits         float64
+	Frequency                Frequency
+	StartTime                string
+	EndTime                  string
+	SuspendTriggers          []TriggerDefinition
+	SuspendImmediateTriggers []TriggerDefinition
```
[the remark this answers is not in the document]
sfc-gh-pbosak: same as above, waiting until we agree on the solution

pkg/sdk/resource_monitors.go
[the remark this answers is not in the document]
sfc-gh-pbosak: Ok, I guess I didn't want to change the behaviour. I thought `suspend_triggers` still might be used under some circumstances.
So what should we do about the `suspend_triggers` argument then?
I think we can:
1) just get rid of it
2) ignore it and just take the value from `suspend_trigger`
3) use the trigger with the lowest threshold from `suspendTrigger` and `suspend_triggers`

I think we should either remove it or at least explain the changed semantics. Otherwise it might be confusing for the users.

pkg/sdk/resource_monitors.go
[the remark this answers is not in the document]
sfc-gh-pbosak: same as above, waiting until we agree on the solution

pkg/sdk/resource_monitors_integration_test.go
sfc-gh-pbosak: probably copy-paste from other resource, fixed

pkg/sdk/resource_monitors.go
sfc-gh-pbosak: Well, if we create empty resource monitor, then the query won't compile(`create resource monitor 'foo' with` is not ok)

pkg/sdk/resource_monitors.go
sfc-gh-pbosak: I did initially use time.Time here, but I think the issue of timestamps is quite tricky.

pkg/sdk/resource_monitors.go
sfc-gh-pbosak: fixed

pkg/sdk/resource_monitors.go
sfc-gh-pbosak: Well, if we create empty resource monitor, then the query won't compile(`create resource monitor 'foo' with` is not ok)

pkg/sdk/resource_monitors.go
sfc-gh-pbosak: fixed

pkg/sdk/resource_monitors.go
sfc-gh-pbosak: I did initially use time.Time here, but I think the issue of timestamps is quite tricky.

pkg/sdk/resource_monitors_integration_test.go
sfc-gh-pbosak: probably copy-paste from other resource, fixed

pkg/sdk/resource_monitors_test.go
sfc-gh-pbosak: ok

sfc-gh-pbosak: ok

pkg/sdk/resource_monitors.go
sfc-gh-swinkler: okay fair point. String is a good choice if "IMMEDIATELY" is one of the options.

pkg/sdk/resource_monitors.go
sfc-gh-swinkler: In that case would it make sense to have With be a separate type?

for example: 
```
With *ResourceMonitorWith `ddl:"keyword" sql:"WITH"`
```
and then later have the ResourceMonitorWith type declared which has all the parameters in it?

pkg/sdk/resource_monitors.go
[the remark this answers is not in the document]
sfc-gh-swinkler: suspend_triggers is a deprecated attribute since it never makes sense to have more than one suspend trigger. We have to continue to support it until the next major provider release (1.0). What i would do is just choose the lowest value in the list and use that

pkg/sdk/resource_monitors.go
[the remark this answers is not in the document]
sfc-gh-swinkler: Yeah just use the lowest value, thats the only thing that makes sense anyways

pkg/sdk/resource_monitors.go
sfc-gh-pbosak: ok

pkg/sdk/resource_monitors.go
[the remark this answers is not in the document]
sfc-gh-pbosak: ok

pkg/sdk/resource_monitors.go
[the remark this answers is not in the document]
sfc-gh-pbosak: ok

sfc-gh-swinkler (approved):

Issue #1700, opened 2023-04-10
Connection Caching For External Browser Auth


DustinMoriarty:
**Is your feature request related to a problem? Please describe.**

While using the snowflake provider with externalbrowser authentication, hundreds of tabs are opened in the browser as it re-authenticates for every resource. This makes the provider very hard to use with externalbrowser authentication.

**Describe the solution you'd like**

Connection caching is now supported by gosnowflake. https://github.com/snowflakedb/gosnowflake/issues/486
We need to determine what additional work is needed to get it to work with terraform. 

**Describe alternatives you've considered**

The alternative is to either accept very slow performance and the provider taking over my browser for the entire time of an apply due to opening many tabs or to use another form of authorization. For companies where SSO or MFA is required for human users, using other authorization is not an option. 

**Additional context**

Terraform Version: 1.4.1
Provider Version: 0.61.0
OS Version: MacOS 13.3

kallangerard:
+1 to this. Human users should definitely not be using Private Keys and Passwords, so without Terraform Cloud there's no compliant way to perform operations from local.

fh-dustin-winslow:
+1 I'm having the same issue, our team is having to do all operations via CI - can't do anything from local without connection caching.

DustinMoriarty:
@kallangerard and @fh-dustin-winslow : thank you for the bump.  The best way to get priority for these tickets is to raise a support ticket within snowflake.  Of course submitting a PR is even better.  However, my Go skills are pretty basic so I have yet to contribute to this repo.

jcourteau:
@DustinMoriarty I took a stab at this here - https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1913. I'm actually not sure why this wasn't auto-enabled on non-Linux platforms, my read of the code is that it _should_ have worked but clearly it didn't.

Note that you need to explicitly enable ID token support on the Snowflake account, using ACCOUNTADMINISTRATOR.

DustinMoriarty:
@jcourteau : Fantastic!  Thanks for taking this on!

thomas-tomlinson:
[...]

PR #1913, opened 2023-06-27
Add client credential caching option

jcourteau:
* Add new provider configuration option - `cache_temporary_credential`, default `false`

This is normally automatically enabled or disabled based on the platform on the upstream gosnowflake SDK. However, even on mac, with the option enabled on the Snowflake account, the provider wasn't using caching.

This PR adds an explicit flag to enable / disable credential caching so that the exact behaviour can be set. Since Linux doesn't offer a password store like Windows and Mac do, this feature is a bit of a security risk there.

This must be [enabled on the snowflake account](https://docs.snowflake.com/en/user-guide/admin-security-fed-auth-use#using-connection-caching-to-minimize-the-number-of-prompts-for-authentication-optional)

Specifically, as ACCOUNTADMIN:

```
alter account set allow_id_token = true;
```

## Test Plan

* [ ] acceptance tests

* [x] added unit tests for the generated DSN, following existing examples
* [x] updated existing unit tests to reflect new config option

I also verified that caching worked with a provider config like this:

```
provider "snowflake" {
  account  = "<my SF account>"
  username = "<me>"
  region   = "us-west-2"

  cache_temporary_credential = true
  browser_auth               = true

  role = "SECURITYADMIN"
}
```

With `cache_temporary_credential=true` I was prompted to save the password in the OSX keychain; and subsequent runs didn't prompt for credentials.

With it set to false, I went through the normal browser auth cycle, with no caching; the same as the pre-patch behaviour.
 
(I would appreciate if someone could test this on Windows, as I don't have that readily available.)

## References

* https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1700 issue requesting this feature
* [Snowflake SSO connection caching](https://docs.snowflake.com/en/user-guide/admin-security-fed-auth-use#using-connection-caching-to-minimize-the-number-of-prompts-for-authentication-optional)
* [gosnowflake caching becomes configurable, optional](https://github.com/snowflakedb/gosnowflake/pull/710)
* [gosnowflake original implementation](https://github.com/snowflakedb/gosnowflake/pull/661)


cablespaghetti:
@sfc-gh-jcieslak @sfc-gh-asawicki @sfc-gh-swinkler sorry for the mention but is there any possibility of this getting merged soon? Without it the terraform provider unusable with SSO/Browser auth.

mlavaert:
I hoped that this was going to be included in PR https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2109 to use the authenticator. Sadly browser authentication is still impossible to use as it opens 100s of browser tabs right now.

sfc-gh-swinkler:
This has already been added: https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs#client_store_temporary_credential

Closing as completed

cablespaghetti:
> This has already been added: https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs#client_store_temporary_credential
> 
> Closing as completed
> 
> Note that we are still having issues with the token for Windows / Darwin not caching token despite adding this. We need to build these two binaries with CGO enabled, and this is proving to be somewhat challenging to do in github-actions. We hope to get a solution out soon

As this feature is not working correctly, is this being tracked in an issue I can watch?

Thanks

sfc-gh-swinkler:
@cablespaghetti you can check the related issue https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2047. We will close this issue only once the feature is working correctly.

PR #1914, opened 2023-06-27
feat: Parameters sdk v2

sfc-gh-pbosak:
* updates session and account parameter resources to use new sdk implementation


pkg/sdk/parameters.go
```diff
@@ -1,6 +1,382 @@
...
+	SetSessionParameterForAccount(ctx context.Context, parameter SessionParameter, value string) error
+	SetObjectParameterForAccount(ctx context.Context, parameter ObjectParameter, value string) error
+}
+
+type parameters struct {
+	client *Client
+}
+
+func (parameters *parameters) SetAccountParameter(ctx context.Context, parameter AccountParameter, value string) error {
+	opts := AlterAccountOptions{Set: &AccountSet{Parameters: &AccountLevelParameters{AccountParameters: &AccountParameters{}}}}
```
sfc-gh-ngaberel, line 20: This is hard to read, can you split this over multiple lines?

pkg/resources/session_parameter.go
```diff
@@ -56,6 +58,8 @@ func CreateSessionParameter(d *schema.ResourceData, meta interface{}) error {
 	db := meta.(*sql.DB)
 	key := d.Get("key").(string)
 	value := d.Get("value").(string)
+	client := sdk.NewClientFromDB(db)
+	ctx := context.Background()
 
 	parameterDefault := snowflake.GetParameterDefaults(snowflake.ParameterTypeSession)[key]
```
sfc-gh-ngaberel, line 64: Can we get rid of this dependency on the snowflake package?

pkg/resources/session_parameter.go
```diff
@@ -64,70 +68,59 @@ func CreateSessionParameter(d *schema.ResourceData, meta interface{}) error {
...
+	var err error
 	if onAccount {
-		builder.SetOnAccount(onAccount)
+		err := client.Parameters.SetSessionParameterForAccount(ctx, parameter, value)
+		if err != nil {
+			return err
+		}
 	} else {
 		if user == "" {
 			return fmt.Errorf("user is required if on_account is false")
```
sfc-gh-ngaberel, line 83: Can we move this logic to a validation function in the resource schema?

pkg/resources/session_parameter.go
```diff
@@ -64,70 +68,59 @@ func CreateSessionParameter(d *schema.ResourceData, meta interface{}) error {
...
 		}
+		// add quotes to value if it is a string
+		typeString := reflect.TypeOf("")
+		if reflect.TypeOf(parameterDefault.DefaultValue) == typeString {
+			value = fmt.Sprintf("'%s'", snowflake.EscapeString(value))
+		}
+		builder := snowflake.NewSessionParameter(key, value, db)
```
sfc-gh-ngaberel, line 90: Still using the old SDK here.

pkg/resources/session_parameter.go
```diff
@@ -64,70 +68,59 @@ func CreateSessionParameter(d *schema.ResourceData, meta interface{}) error {
...
+		}
 	} else {
 		if user == "" {
 			return fmt.Errorf("user is required if on_account is false")
 		}
+		// add quotes to value if it is a string
+		typeString := reflect.TypeOf("")
+		if reflect.TypeOf(parameterDefault.DefaultValue) == typeString {
+			value = fmt.Sprintf("'%s'", snowflake.EscapeString(value))
+		}
```
sfc-gh-ngaberel, line 89: This should already be handled by the new SDK, cf. https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/sdk/sql_builder.go#L52

pkg/resources/session_parameter.go
```diff
@@ -64,70 +68,59 @@ func CreateSessionParameter(d *schema.ResourceData, meta interface{}) error {
...
 	onAccount := d.Get("on_account").(bool)
-	var p *snowflake.Parameter
 	var err error
+	var p *sdk.Parameter
 	if onAccount {
-		p, err = snowflake.ShowAccountParameter(db, key)
-		if err != nil {
-			return fmt.Errorf("error reading session parameter err = %w", err)
-		}
+		p, err = client.Sessions.ShowAccountParameter(ctx, sdk.AccountParameter(parameter))
```
sfc-gh-ngaberel, line 114: It's a bit weird to have `client.Sessions.ShowAccountParameter` and `client.Parameters.SetSessionParameterForAccount`, maybe we should have all parameter-related methods under the same object.

pkg/resources/session_parameter.go
```diff
@@ -143,28 +136,43 @@ func UpdateSessionParameter(d *schema.ResourceData, meta interface{}) error {
...
+		err = client.Parameters.SetSessionParameterForAccount(ctx, parameter, defaultValue)
+		if err != nil {
+			return fmt.Errorf("error creating session parameter err = %w", err)
+		}
 	} else {
 		user := d.Get("user").(string)
+		if user == "" {
+			return fmt.Errorf("user is required if on_account is false")
+		}
+		parameterDefault := snowflake.GetParameterDefaults(snowflake.ParameterTypeSession)[key]
```
sfc-gh-ngaberel, line 161: Remove dependency on old SDK.

pkg/resources/object_parameter.go
```diff
@@ -91,6 +93,9 @@ func CreateObjectParameter(d *schema.ResourceData, meta interface{}) error {
 	db := meta.(*sql.DB)
 	key := d.Get("key").(string)
 	value := d.Get("value").(string)
+	client := sdk.NewClientFromDB(db)
+	ctx := context.Background()
+	parameter := sdk.ObjectParameter(key)
 
 	parameterDefault := snowflake.GetParameterDefaults(snowflake.ParameterTypeObject)[key]
```
sfc-gh-ngaberel, line 100: Old SDK here too.

pkg/resources/account_parameter.go
```diff
@@ -45,49 +46,38 @@ func CreateAccountParameter(d *schema.ResourceData, meta interface{}) error {
 	db := meta.(*sql.DB)
 	key := d.Get("key").(string)
 	value := d.Get("value").(string)
+	client := sdk.NewClientFromDB(db)
+	ctx := context.Background()
+	parameter := sdk.AccountParameter(key)
 
-	parameterDefault := snowflake.GetParameterDefaults(snowflake.ParameterTypeAccount)[key]
+	parameterDefault := snowflake.GetParameterDefaults(snowflake.ParameterTypeSession)[key]
```
sfc-gh-ngaberel, line 53: Old SDK here.

pkg/resources/account_parameter_acceptance_test.go
```diff
@@ -48,19 +48,3 @@ func TestAcc_AccountParameter_PREVENT_LOAD_FROM_INLINE_URL(t *testing.T) {
...
-			{
-				Config: accountParameterBasic("CLIENT_METADATA_USE_SESSION_DATABASE", "true"),
-				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("snowflake_account_parameter.p", "key", "CLIENT_METADATA_USE_SESSION_DATABASE"),
-					resource.TestCheckResourceAttr("snowflake_account_parameter.p", "value", "true"),
-				),
-			},
-		},
-	})
-}
```
sfc-gh-ngaberel, line 66: Why are we removing this test? It should still pass, right?

sfc-gh-ngaberel (changes requested):

pkg/resources/session_parameter.go
sfc-gh-swinkler: in the snowflake api Account parameters are set at the  `ALTER ACCOUNT SET <>` while Session parameters are set using `ALTER SESSION SET <>` or for users with `ALTER USER SET <>`, or at the account level using the same `ALTER ACCOUNT SET <>`. This is a bit confusing. I think we should consolidate all the parameter methods into the client while still maintaining compatability with the original API. So `client.Parameters.SetSessionParameterForAccount()` would do the same thing as `client.Accounts.Alter(..)`. What we don't want it someone looking for the "ALTER SESSION SET PARAMETER" method and not being able to find it.

sfc-gh-swinkler (changes requested):

pkg/sdk/parameters.go
sfc-gh-pbosak: done

pkg/resources/session_parameter.go
sfc-gh-pbosak: done

pkg/resources/session_parameter.go
sfc-gh-pbosak: I'm not sure if we can do that since we only check if it's empty if on_account is false. validation functions in resource schema cannot access another attribute value afaik

pkg/resources/session_parameter.go
sfc-gh-pbosak: fixed

pkg/resources/session_parameter.go
sfc-gh-pbosak: fixed

pkg/resources/session_parameter.go
sfc-gh-pbosak: All those function are part of `Parameters` interface now

pkg/resources/session_parameter.go
sfc-gh-pbosak: fixed

pkg/resources/object_parameter.go
sfc-gh-pbosak: I don't think we can get rid of that in this case, at least not fully. In order to make use of new sdk here, we need to have new sdk implementation for all objects which can take parameters and that's not the case just yet. It can be improved in the future, but I think it has to use old sdk for now

pkg/resources/account_parameter.go
sfc-gh-pbosak: fixed

pkg/resources/account_parameter_acceptance_test.go
sfc-gh-pbosak: This test tries to set CLIENT_METADATA_USE_SESSION_DATABASE for account.
CLIENT_METADATA_USE_SESSION_DATABASE is a session parameter and I think we should only allow to set it using a session_parameter. For now I modified it to use an actual account parameter, but if you think we should allow to set session parameters using account_parameter resource then we can discuss it

pkg/resources/account_parameter.go
```diff
@@ -101,24 +91,18 @@ func UpdateAccountParameter(d *schema.ResourceData, meta interface{}) error {
...
 	if err != nil {
-		return fmt.Errorf("error creating account parameter err = %w", err)
+		return err
 	}
-	_, err = snowflake.ShowAccountParameter(db, key)
+	defaultValue := defaultParameter.Default
+	err = client.Parameters.SetAccountParameter(ctx, parameter, defaultValue)
 	if err != nil {
-		return fmt.Errorf("error reading account parameter err = %w", err)
+		return fmt.Errorf("error creating account parameter err = %w", err)
```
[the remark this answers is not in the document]
sfc-gh-pbosak: fixed

sfc-gh-swinkler (approved):

Issue #1923, opened 2023-07-05
ENABLE_UNREDACTED_QUERY_SYNTAX_ERROR parameter cannot be set


pkotlar:
**Provider Version**

0.67.0

**Terraform Version**

1.4.0

**Describe the bug**

ENABLE_UNREDACTED_QUERY_SYNTAX_ERROR is a valid user parameter that can be also set at the account level per https://docs.snowflake.com/en/sql-reference/parameters#enable-unredacted-query-syntax-error .

No matter if I use snowflake_account_parameter or snowflake_object_parameter I get:

```
Error: expected key to be one of [...], got ENABLE_UNREDACTED_QUERY_SYNTAX_ERROR
```

**Expected behavior**

Terraform should apply the parameter with no issues.

**Code samples and commands**

```
resource "snowflake_account_parameter" "p1" {
  key        = "ENABLE_UNREDACTED_QUERY_SYNTAX_ERROR"
  value      = "TRUE"
}
```
or:
```
resource "snowflake_object_parameter" "p2" {
  key         = "ENABLE_UNREDACTED_QUERY_SYNTAX_ERROR"
  value       = "TRUE"
  object_type = "USER"
  object_identifier {
    name = "TESTUSER"
  }
}
```

**Additional context**

It is not included in parameters.go though a similar user level parameter (PREVENT_UNLOAD_TO_INTERNAL_STAGES) is.

sfc-gh-swinkler:
Going to close this issue since it was resolved by a merged PR which has since been released. Thanks for your contribution!

PR #1924, opened 2023-07-05
fix: snowflake_object_parameter add missing option

pkotlar:
Add missing option as per https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1923


sfc-gh-swinkler (approved):
LGTM

PR #1945, opened 2023-07-11
feat: Users sdk v2

sfc-gh-pbosak:
* Refactored users to sdk v2
* Added unit and integration tests

## Test Plan

* [X] unit tests
* [X] acceptance tests


pkg/resources/user.go
```diff
@@ -181,61 +254,210 @@ func ReadUser(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	u, err := snowflake.ScanUserDescription(rows)
-	if err != nil {
-		return err
-	}
-	if err = d.Set("name", u.Name.String); err != nil {
-		return err
+	if user.Name != nil {
```
sfc-gh-asawicki, line 257: similar comment could applied here.

Maybe we should rearrange both of these and introduce something like a really simple builder which we could use like (pseudocode):
```
b
  .set("name", user.Name)
  .set("comment", user.Comment)
  .set("login_name", user.LoginName)
  .build() // this last part would try to apply all above and return first error which we can check against OR something like that
```

\* and this is just an idea, which we can discuss and implement it in following PRs if accepted

pkg/resources/user.go
```diff
@@ -133,9 +137,6 @@ var userSchema = map[string]*schema.Schema{
 		Sensitive:   true,
 		Description: "Last name of the user.",
 	},
-	"tag": tagReferenceSchema,
```
sfc-gh-asawicki, line 136: so why are tags removed from here, but they still are present e.g. in `CreateUserOptions` (and set in tests etc)? Just want to confirm: is it because we differentiate between sdk capabilities, but from the terraform resource side we would like to use tag_association resource to set these tags (in this case)?

pkg/sdk/common_types.go
```diff
@@ -96,9 +103,43 @@ func (row *propertyRow) toStringProperty() *StringProperty {
 }
 
 func (row *propertyRow) toIntProperty() *IntProperty {
+	var value *int
```
sfc-gh-asawicki, line 106: can we add unit tests both to `toIntProperty` and `toBoolProperty`?

pkg/resources/password_policy.go
```diff
@@ -206,32 +206,51 @@ func ReadPasswordPolicy(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return err
 	}
-	if err := d.Set("min_length", passwordPolicyDetails.PasswordMinLength.Value); err != nil {
-		return err
+	if passwordPolicyDetails.PasswordMinLength != nil && passwordPolicyDetails.PasswordMinLength.Value != nil {
```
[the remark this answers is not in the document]
sfc-gh-pbosak: Ok, I did write such  functions, I think the code looks cleaner now.

pkg/resources/user.go
sfc-gh-pbosak: We could do that if we care about short-circuiting(i.e. stopping after first encountered error).
Another option would be to use `errors.Join` so we would have sth like
```
err := errors.Join(
d.set("name", user.Name) 
...
)

```
I wonder if aggregating errors like that isn't a better option. Those actions are independent and we probably want to give as much information to the user as possible, if we discover multiple errors, we should report all of them.

I'm not modifying the code just yet, we should discuss it further

pkg/resources/user.go
sfc-gh-pbosak: Yes, this is something  we should talk about, probably @scottwinkler would be really helpful in this conversation.
We have a separate `tag_association` resource which can be used for setting tags, and from what I could gather we want to move towards that approach. I based my implementation on recent warehouse sdk rewrite, which also removed the `tag` attribute from the schema.
One obvious downside of just removing "tag" attribute from the schema is that it breaks compatibility which I guess we want to avoid. Rewriting `tagReferenceSchema` to using new sdk implementation seems like quite a bit of work, but maybe it's still worth it

pkg/sdk/common_types.go
sfc-gh-pbosak: ok

pkg/resources/password_policy.go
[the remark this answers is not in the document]
sfc-gh-swinkler: this is a good idea. Another popular strategy is to create a struct with flatten / expand functions. The "flatten" function saves it to TF state, while the expand function instantiates the struct from TF state. It could be combined with the function here I think.

pkg/resources/user.go
sfc-gh-swinkler: i dont like use a builder pattern for this kind of problem. In the new plugin framework, which we will eventually be migrating to, they use models to read from and save the state. See https://developer.hashicorp.com/terraform/plugin/framework/resources/create.

You might have a ThingResource with a model like:
```
// ThingResourceAPIModel describes the API data model.
type ThingResourceAPIModel struct {
    Name string `json:"name"`
    Id   string `json:"id"`
}
```
which then can be marshalled / unmarshalled using helper functions:

```
  // Read Terraform plan data into the model
    resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
```

and

```
    // Save data into Terraform state
    resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)

```

pkg/resources/user.go
sfc-gh-swinkler: we cannot introduce any more breaking changes to the provider prior to the next major release. We have to keep tags in, if it was already supported by the resource. I think we should support both ways of tagging resources. So perhaps deprecating this attribute was the wrong idea.

sfc-gh-pbosak: Now that I look at it, I'm not sure if tags were actually supported for users resource. I might be mistaken, but it seems they were part of the schema, but we didn't do anything with them. In resources like `schema` we make use of the "tag" attribute and perform corresponding SQL queries, but for users and warehouse(on which I based this implementation) resources didn't really support tags as part of the resource.
So I think maybe we could just leave the tag schema as it was and try to add actual support for tags for the resources like users and warehouses(in a separate PR though, this one is already taking too long).

pkg/resources/user.go
sfc-gh-pbosak: Ok, I guess I'll keep it for now then

pkg/sdk/users_integration_test.go
```diff
@@ -0,0 +1,240 @@
...
+	t.Run("when user exists", func(t *testing.T) {
+		userDetails, err := client.Users.Describe(ctx, user.ID())
+		require.NoError(t, err)
+		assert.Equal(t, user.Name, userDetails.Name.Value)
+	})
+
+	t.Run("when user does not exist", func(t *testing.T) {
+		id := NewAccountObjectIdentifier("does_not_exist")
+		_, err := client.Users.Describe(ctx, id)
+		assert.ErrorIs(t, err, ErrObjectNotExistOrAuthorized)
```
sfc-gh-pbosak, line 218: One thing I find quite annoying is when we raise an error like this, we get no information about where the error was initially raised. Since errors are regular values in Go, they don't carry any stack trace with them.
I think we might consider 
1) using a third-party library(https://github.com/go-errors/errors)
2) converting those errors into a thunk function type, like this:
```
	ErrObjectNotExistOrAuthorized = func() error {
		return errors.New("object does not exist or not authorized")
	}
```
that way we can add a breakpoint inside it and see the stack trace with a debugger
3) add some basic info to the error, like what kind of object does not exist etc.

sfc-gh-asawicki: I like 1) and 2).

sfc-gh-asawicki (approved):

pkg/resources/password_policy.go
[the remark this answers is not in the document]
sfc-gh-asawicki: Okay, my part of the comment is fine, but I am not resolving it because of the above comment which awaits @sfc-gh-swinkler answer.

pkg/resources/user_test.go
```diff
@@ -8,7 +8,6 @@ import (
 	sqlmock "github.com/DATA-DOG/go-sqlmock"
```
sfc-gh-swinkler, line 8: this file is a remnant of some of the older tests which we are moving away from. this entire file can be deleted

pkg/sdk/users_integration_test.go
sfc-gh-swinkler: perhaps wrapping the errors at each place where they are thrown. that way you can see the stacktrace as part of the error message as a sort of linked list
`error: <err msg> occurred in <func> -> <struct> -> <other func> -> <other struct>`

sfc-gh-pbosak: Right, @sfc-gh-jcieslak and I were discussing an option like this, but it requires quite a lot of effort when rethrowing errors everywhere. We need to also agree what is it that we want to achieve here.
My main pain point was not knowing when the error initially occurred during debugging a failing test.
I'm not sure what functionality can an external library provide, but we can either:
a) use the suggestion from point 2)
b) add a trace using runtime package when the error initially enters our application(i.e when we create one ourselves or get one from an external library)
I find such information more useful in this case because I can just go directly to the line when the error was raised.
Having a wrapper for errors will still be useful, but it:
a) needs additional effort on every rethrow
b) still requires a bit of decoding if we want to go to a specific line when the error happened.
Maybe such an error would be more helpful for the customer? I don't know, but I think those 2 approaches are not mutually exclusive, but whether we need the latter if we have the former is something is something I'm 100% sure about.

pkg/resources/user.go
```diff
@@ -181,61 +254,200 @@ func ReadUser(d *schema.ResourceData, meta interface{}) error {
...
+		defaultRoles, _ := strings.CutPrefix(user.DefaultSecondaryRoles.Value, "[\"")
+		defaultRoles, _ = strings.CutSuffix(defaultRoles, "\"]")
+		defaultSecondaryRoles = strings.Split(defaultRoles, ",")
 	}
 	if err = d.Set("default_secondary_roles", defaultSecondaryRoles); err != nil {
 		return err
 	}
-	if err = d.Set("default_namespace", u.DefaultNamespace.String); err != nil {
-		return err
+	if user.DefaultNamespace != nil {
```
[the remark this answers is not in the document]
sfc-gh-pbosak: fixed

pkg/resources/user_test.go
sfc-gh-pbosak: ok

sfc-gh-asawicki (approved):

PR #1956, opened 2023-07-17
Fix file format integration tests

sfc-gh-ngaberel:
Fixed expected value for owner type.

## Test Plan
* [X] updated integration test

## References

*


sfc-gh-pbosak (approved):
LGTM

PR #1965, opened 2023-07-20
fix: support generic csim client

sfc-gh-pbosak:
* fixes validation function for scim integration, allowing for "generic" integration


sfc-gh-swinkler (approved):
This resource should have been named `scim_security_integration`. We will need to change at a later date, I think.

PR #1983, opened 2023-07-28
fix: unable to alter stage if url and si have changes  (#1982)

### LINKED ISSUE #1982: Unable to alter stage when both storage integration & allowed bucket url are changed
**Provider Version**
0.67.0

**Terraform Version**
1.5.4

**Describe the bug**

Scenario :
-  Storage Integration "SI_ONE" has a bucket url called "gcs://abc"  in its allowed location
- Stage "STAGE_ONE" has been mapped to "SI_ONE" & url "gcs://abc" 
- Now, user changes the SI name from "SI_ONE" to "SI_NEW" and allowed_bucket url to "gcs://def" (previous gcs://abc is removed as part of this change)
- After this change Stage updates comes in picture and tries to change the SI name & url and this where it fails because of the reasons below:
- Now, stage tries to update the url to "gcs://def" as per the sequence in the code and it would fail because "SI_ONE" which is mapped to stage doesn't exist
- Even if you try to change the SI name to "SI_NEW" first then also it would fail because the updated SI has different bucket url in allowed list, and stage still has older url which is not allowed any more

**Expected behavior**

When both SI & url are changed then it should be updated as a single alter query rather two separate queries

```sql
ALTER STAGE "TEST_DB"."TEST_SCHEMA"."TEST_STAGE" 
SET STORAGE_INTEGRATION = "SI_NEW"
URL="gcs://def"
```

**Code samples and commands**

Plan showing changes to SI name & url for the stage, previous SI "SI_ONE" has dropped & new "SI_NEW" is mapped to "gcs://def" as stated above

```
 # snowflake_stage.stage will be updated in-place
  ~ resource "snowflake_stage" "stage" {
        id                  = "TEST_DB|TEST_SCHEMA|TEST_STAGE"
        name                = "TEST_STAGE"
      ~ storage_integration = "SI_ONE" -> "SI_NEW"
      ~ url                 = "gcs://abc" -> "gcs://def"
        # (2 unchanged attributes hidden)
    }
```

Executing terraform apply in debug mode results:
`[DEBUG] sql-conn-exec [query ALTER STAGE "TEST_DB"."TEST_SCHEMA"."TEST_STAGE" SET URL = 'gcs://def' err 003139 (42601): SQL compilation error: Integration 'SI_ONE' associated with the stage 'TEST_STAGE' cannot be found. duration 485.574375ms args {}]
`

If I just try to change the SI name also for the stage, see the plan:
``` # snowflake_stage.stage has changed
  ~ resource "snowflake_stage" "stage" {
        id                  = "TEST_DB|TEST_SCHEMA|TEST_STAGE"
        name                = "TEST_STAGE"
      ~ storage_integration = "SI_NEW" -> "SI_ONE"
        # (3 unchanged attributes hidden)
    }
```

Executing the terraform apply now for the same:
`[DEBUG] sql-conn-exec [query ALTER STAGE "TEST_DB"."TEST_SCHEMA"."TEST_STAGE" SET STORAGE_INTEGRATION = "SI_NEW" err 003127 (42601): SQL compilation error: Location 'gcs://abc' is not allowed by integration SI_NEW. Please use DESC INTEGRATION to check out allowed and blocked locations. duration 481.370625ms args {}]
`

**Additional context**

This problem would only happen if the existing bucket location is removed from the allowed list from the storage integration & its name is also changed at the same time

malduarte:
Previously, if both `URL` and `STORAGE_INTEGRATION` had been updated the provider would generate separate ALTER STAGE commands - one for `URL`, and another for `STORAGE` integration.  This can create an invalid change because:
- When changing `URL`: the new `URL` could be invalid with the existing storage integration `STORAGE_ALLOWED_LOCATIONS`
- When changing STORAGE_INTEGRATION: existing `URL` not present in the new storage integration `STORAGE_ALLOWED_LOCATIONS`

This MR now checks if both `URL` and `STORAGE_INTEGRATION` have changed and issues a single `ALTER` statement with both changes in one go.

Before:

ALTER STAGE the_stage SET URL = 's3://the_new_url'
ALTER STAGE the_stage SET STORAGE_INTEGRATION  = "new_storage_integration"

Now:

ALTER STAGE the_stage SET STORAGE_INTEGRATION  = "new_storage_integration" URL = 's3://the_new_url'

## Test Plan
* We have added tests for `ChangeStorageIntegrationAndUrl` (newly added function)
* We have added tests for `Stage` resource update to ensure:
  * The combined `ALTER STAGE` is only issued when both `URL` and `STORAGE_INTEGRATION` change
  * Existing behavior remains unchanged when only one of `URL`  and `STORAGE_INTEGRATION` change
* We have added  acceptance tests that replicate the problem and prove the fix in `pkg/resources/stage_acceptance_test.go`

## References
* Fixes #1982 
* https://docs.snowflake.com/en/sql-reference/sql/alter-stage

### DIFF
--- a/pkg/resources/stage.go
+++ b/pkg/resources/stage.go
@@ -319,13 +319,6 @@ func UpdateStage(d *schema.ResourceData, meta interface{}) error {
 	builder := snowflake.NewStageBuilder(stage, dbName, schema)
 
 	db := meta.(*sql.DB)
-	if d.HasChange("url") {
-		url := d.Get("url")
-		q := builder.ChangeURL(url.(string))
-		if err := snowflake.Exec(db, q); err != nil {
-			return fmt.Errorf("error updating stage url on %v", d.Id())
-		}
-	}
 
 	if d.HasChange("credentials") {
 		credentials := d.Get("credentials")
@@ -335,11 +328,28 @@ func UpdateStage(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
-	if d.HasChange("storage_integration") {
+	if d.HasChange("storage_integration") && d.HasChange("url") {
 		si := d.Get("storage_integration")
-		q := builder.ChangeStorageIntegration(si.(string))
+		url := d.Get("url")
+		q := builder.ChangeStorageIntegrationAndUrl(si.(string), url.(string))
 		if err := snowflake.Exec(db, q); err != nil {
-			return fmt.Errorf("error updating stage storage integration on %v", d.Id())
+			return fmt.Errorf("error updating stage storage integration and url on %v", d.Id())
+		}
+	} else {
+		if d.HasChange("storage_integration") {
+			si := d.Get("storage_integration")
+			q := builder.ChangeStorageIntegration(si.(string))
+			if err := snowflake.Exec(db, q); err != nil {
+				return fmt.Errorf("error updating stage storage integration on %v", d.Id())
+			}
+		}
+
+		if d.HasChange("url") {
+			url := d.Get("url")
+			q := builder.ChangeURL(url.(string))
+			if err := snowflake.Exec(db, q); err != nil {
+				return fmt.Errorf("error updating stage url on %v", d.Id())
+			}
 		}
 	}
 
--- a/pkg/resources/stage_acceptance_test.go
+++ b/pkg/resources/stage_acceptance_test.go
@@ -1,0 +1,68 @@
+package resources_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccAlterStageWhenBothURLAndStorageIntegrationChange(t *testing.T) {
+	name := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+
+	resource.ParallelTest(t, resource.TestCase{
+		Providers:    providers(),
+		CheckDestroy: nil,
+		Steps: []resource.TestStep{
+			{
+				Config: stageIntegrationConfig(name, "si1", "s3://foo/"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_stage.test", "name", name),
+					resource.TestCheckResourceAttr("snowflake_stage.test", "url", "s3://foo/"),
+				),
+				Destroy: false,
+			},
+			{
+				Config: stageIntegrationConfig(name, "changed", "s3://changed/"),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_stage.test", "name", name),
+					resource.TestCheckResourceAttr("snowflake_stage.test", "url", "s3://changed/"),
+				),
+			},
+		},
+	})
+}
+
+func stageIntegrationConfig(name string, siNameSuffix string, url string) string {
+	resources := `
+resource "snowflake_database" "test" {
+	name = "%s"
+	comment = "Terraform acceptance test"
+}
+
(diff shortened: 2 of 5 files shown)


sfc-gh-ngaberel:
HI @malduarte, thanks for submitting this.
We're going through a major refactor of our SQL builder and are slowly removing the mock DB tests for resources, focusing on acceptance tests instead. Could you add an acceptance test that changes both attributes at once so we can avoid a regression when we refactor the stage resource?

malduarte:
@sfc-gh-ngaberel Will do!

malduarte:
@sfc-gh-ngaberel What cloud provider do you use to run acceptance tests? I'm assuming AWS. Unlike other resources [stage](https://docs.snowflake.com/en/sql-reference/sql/create-stage) and the linked [storage integration](https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration) URL formats /configuration are snowflake provider specific. How do I configure an acceptance test in such a way that it _only_ runs on a specific Snowflake cloud provider?

sfc-gh-ngaberel:
Oh good point I didn't think about that. We'll definitely start with S3, although it would be nice to test every supported provider in the future.
For the test you can create two storage integrations using S3, we'll pass the role ARN and allowed locations as action secrets / environment variables. I don't have values for these yet but you can assume the following environment variables exist when writing the tests, I'll get those values and do the rest of the plumbing:
- STORAGE_S3_ROLE_ARN_1
- STORAGE_S3_URL_1: compatible with STORAGE_S3_ROLE_ARN_1
- STORAGE_S3_ROLE_ARN_2
- STORAGE_S3_URL_2: compatible with STORAGE_S3_ROLE_ARN_2

malduarte:
@sfc-gh-ngaberel That might not be necessary (*) for stages and storage integrations. Maybe my perception is biased because on GCS the URLs don't need to refer to an existing bucket. Creating SI/Stages is perfectly fine with test bogus urls.  Validation only occurs when you actually try to use them.

That being said we still need to have cloud provider-specific tests because the DDL validation depends on the account under which snowflake runs. 


(*) I suspect it would be necessary for things that actually use them. Creating snowpipes comes to mind

image

malduarte:
@sfc-gh-ngaberel Currently creating an aws trial account to run the acceptance tests on AWS

malduarte:
@sfc-gh-ngaberel Added [acceptance test](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/ed0b3d19504994d072b56a027a062ef53d63aa15/pkg/resources/stage_acceptance_test.go) that:
- replicates the bug on the previous version
- proves the bug is fixed on the new version

```
/Users/malduarte/.asdf/installs/golang/1.20.1/go/bin/go tool test2json -t /Users/malduarte/Library/Caches/JetBrains/IntelliJIdea2023.2/tmp/GoLand/___7TestAccStage_in_github_com_Snowflake_Labs_terraform_provider_snowflake_pkg_resources.test -test.v -test.paniconexit0 -test.run ^\QTestAccStage\E$
=== RUN   TestAccStage
=== PAUSE TestAccStage
=== CONT  TestAccStage
    stage_acceptance_test.go:14: Step 2/2 error: Error running apply: exit status 1
        
        Error: error updating stage url on rxfkxeyghq|rxfkxeyghq|rxfkxeyghq
        
          with snowflake_stage.test,
          on terraform_plugin_test.tf line 22, in resource "snowflake_stage" "test":
          22: resource "snowflake_stage" "test" {
        
--- FAIL: TestAccStage (11.06s)

FAIL
```

sfc-gh-ngaberel:
This is great, thanks @malduarte!!

sfc-gh-ngaberel (approved):

sfc-gh-ngaberel:
Please fix the formatting of `pkg/resources/stage_acceptance_test.go` and we'll merge.

malduarte:
@sfc-gh-ngaberel done!

sfc-gh-ngaberel (approved):

PR #1996, opened 2023-08-04
fix: Always set MUST_CHANGE_PASSWORD on account creation

doyleish:
- Previoulsy, GetOK was returning false on a falsely-set value for this, stopping it from going into the query string.  We can safely use Get instead to always get this clause into the query string (The resource param has a default of false already).
- This has the added benefit of not relying on the default behavior on the snowflake side, which has changed in the past.  So tf users shouldn't be caught off guard if it changes again.
- I wrote a test for this before I realized this issue lied in the provider code

this addresses https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1607

## Test Plan

* [x] acceptance tests
* [x] unit tests

* [x] local testing, debug logs, and activity logs on snowflake side

## References

*

### DIFF
--- a/pkg/resources/account.go
+++ b/pkg/resources/account.go
@@ -242,9 +242,11 @@ func CreateAccount(d *schema.ResourceData, meta interface{}) error {
 	if v, ok := d.GetOk("last_name"); ok {
 		createOptions.LastName = sdk.String(v.(string))
 	}
-	if v, ok := d.GetOk("must_change_password"); ok {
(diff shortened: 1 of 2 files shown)


sfc-gh-ngaberel (approved):

sfc-gh-ngaberel:
Thanks @doyleish! This will be available in the next release, 0.70.0.

doyleish:
Awesome! Thanks for the merge

PR #2000, opened 2023-08-07
fix: warehouse size validations

sfc-gh-bculberson:
max clusters of 10 is a soft limit and can be lifted via support. this removes some validation on the max clusters to allow for these soft limit elevations.


sfc-gh-ngaberel:
@sfc-gh-bculberson Thanks for fixing this! Can you add a unit test with MinCount > 10 to ensure we don't introduce a regression later?

sfc-gh-ngaberel (approved):

PR #2006, opened 2023-08-10
feat: Migrate external tables to new sdk

sfc-gh-jcieslak:
## Test Plan
* [x] unit tests (SQL generation + validations)
* [x] integration tests

## References
- Create External Table - https://docs.snowflake.com/en/sql-reference/sql/create-external-table


pkg/sdk/external_tables_integration_test.go
```diff
@@ -0,0 +1,392 @@
+package sdk
+
+import (
+	"context"
+	"fmt"
+	"github.com/stretchr/testify/assert"
```
sfc-gh-asawicki, line 6: This time golangci is right, we keep internal and external imports separated (github.com/stretchr/* should be on the bottom of import, with line separating them from context and fmt)

pkg/sdk/errors.go
```diff
@@ -16,8 +16,12 @@ var (
 	ErrInvalidObjectIdentifier = errors.New("invalid object identifier")
 )
 
-func errOneOf(fieldNames ...string) error {
-	return fmt.Errorf("fields %v are incompatible and cannot be set at once", fieldNames)
+func errOneOf(structName string, fieldNames ...string) error {
```
sfc-gh-asawicki, line 19: AS I remember this method is used in generator, so you have to adjust it there.

pkg/sdk/validations.go
```diff
@@ -79,7 +79,7 @@ func valueSet(value interface{}) bool {
 		reflectedValue = reflectedValue.Elem()
 	}
 	switch reflectedValue.Kind() {
-	case reflect.Slice:
+	case reflect.Slice, reflect.String:
```
sfc-gh-asawicki, line 82: hmmm, I am not sure about this change, at least add a test to TestValueSet pls

pkg/sdk/external_tables_integration_test.go
```diff
@@ -0,0 +1,392 @@
+package sdk
```
sfc-gh-asawicki, line 1: As I see, some of your tests are not passing. Did you try to running the same queries by hand? Did you try to adjust them according to the SQL error suggestions?

For tag, try with simpler name, I guess sometimes the strangely generated names are not working (not for all resources).
Getting tag in one of the tests is also suggesting to use slightly different syntax.

sfc-gh-asawicki (changes requested):

pkg/sdk/validations.go
sfc-gh-jcieslak: I don't remeber particular reason of why I added it here, but here's my understanding right now. If we want to enforce that some field has to be set and we're validating it with valueSet and maybe some other validating function, we want to make sure that this field is not empty (I don't think there might be case where some value is needed, but empty string is correct value, there might be one but I cannot think about one, because then I would assume it's optional). Plus I think it's correct to assume that string is just a syntactic sugar around []byte which is a slice of bytes (which was already there - I mean reflect.Slice).

pkg/sdk/external_tables_integration_test.go
sfc-gh-jcieslak: Yes, 3 tests are failing, because although I think queries are correct according to the Snowflake documentation they're not correct, because they're failing to run (In the test and by hand). I'll try to run them with different names(ids) or variants of syntax for tags to see if anything will work.

pkg/sdk/validations.go
sfc-gh-asawicki: And what about e.g. comment? Let's say, you want to set it to empty string (which is correct behaviour AFAIK). Then you cannot use valueSet for it, right?

pkg/sdk/poc/generator/poc.go
```diff
@@ -192,7 +193,7 @@ func (v *Validation) Error() string {
 	case ValidIdentifier:
 		return fmt.Sprintf("ErrInvalidObjectIdentifier")
 	case ConflictingFields:
-		return fmt.Sprintf("errOneOf(%s)", strings.Join(v.paramsQuoted(), ","))
+		return fmt.Sprintf("errOneOf(%s, %s)", wrapWith(v.Struct.Name, `"`), strings.Join(v.paramsQuoted(), ","))
```
sfc-gh-asawicki, line 196: both other methods here (`Condition` and `TodoComment`) take field as an argument, so I guess it would be more consistent to avoid adding this Struct field to Validation, and just pass it from context like in aforementioned methods.

pkg/sdk/poc/generator/poc_builders.go
```diff
@@ -60,6 +60,9 @@ func (field *Field) WithFields(Fields []*Field) *Field {
 }
 
 func (field *Field) WithValidations(Validations []*Validation) *Field {
+	for _, v := range Validations {
```
sfc-gh-asawicki, line 63: and them, this can be removed.

sfc-gh-asawicki (changes requested):

pkg/sdk/validations.go
sfc-gh-jcieslak: Hmm, yeah you're right, but today I stumbled upon similar case with slice, because sometimes you may want to set empty slice because you would like to 'unset' something with it. I think we have to either change this function to support this kind of edge cases or create new valueSet function that will be able to do it with "better" defaults or more arguments / options.

pkg/sdk/external_tables_integration_test.go
sfc-gh-jcieslak: Well, Delta table is passing, but Set/Unset Tags are still not passing. This is something different and weird, because I believe it says "tags" after "set" is "Syntax Error" while in the documentation "tags" keyword is clearly after "set"

pkg/sdk/external_tables.go
```diff
@@ -0,0 +1,472 @@
...
+	Cloud               string
+	Region              string
+	NotificationChannel sql.NullString
+	LastRefreshedOn     sql.NullTime
+	TableFormat         string
+	LastRefreshDetails  sql.NullString
+	OwnerRoleType       string
+}
+
+func (e externalTableRow) ToExternalTable() ExternalTable {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: can we do var _ con... = (*externalTableRow)(nil) to avoid allocating memory ? I'm not sure if it will, because maybe it should be optimized out of final binary, but still, wdyt ?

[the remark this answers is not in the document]
sfc-gh-asawicki: I like the `new` syntax much more TBH, but either will work, so I don't care that much which

pkg/sdk/external_tables_impl.go
```diff
@@ -0,0 +1,99 @@
+package sdk
+
+import "context"
+
+var (
+	_ ExternalTables                = (*externalTables)(nil)
```
sfc-gh-asawicki, line 6: You have this twice. PLease remove it from other file (I mean the ExternalTables one)

pkg/sdk/external_tables_impl.go
```diff
@@ -0,0 +1,99 @@
...
+	if !validObjectidentifier(req.id) {
+		return nil, ErrInvalidObjectIdentifier
+	}
+
+	externalTables, err := v.client.ExternalTables.Show(ctx, NewShowExternalTableRequest().WithLike(String(req.id.Name())))
+	if err != nil {
+		return nil, err
+	}
+
+	for _, t := range externalTables {
```
sfc-gh-asawicki, line 61: why not `findOne` here?

pkg/sdk/external_tables_impl.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (v *externalTables) DescribeColumns(ctx context.Context, req *DescribeExternalTableColumnsRequest) ([]ExternalTableColumnDetails, error) {
+	rows, err := validateAndQuery[externalTableColumnDetailsRow](v.client, ctx, &describeExternalTableColumns{
+		name: req.id,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]ExternalTableColumnDetails, len(rows))
```
sfc-gh-asawicki, line 78: why not `convert` method also here and in the next method? You have different rows, so you can support entirely different convert methods and reuse the function instead of copy-pasting.

pkg/sdk/external_tables_integration_test.go
```diff
@@ -0,0 +1,397 @@
...
+		err = client.ExternalTables.Alter(
+			ctx,
+			NewAlterExternalTableRequest(externalTableID).
+				WithIfExists(Bool(true)).
+				WithAutoRefresh(Bool(true)),
+		)
+		require.NoError(t, err)
+	})
+
+	// TODO: Uncomment when the problem with alter external table set / unset tags is solved
```
sfc-gh-asawicki, line 236: Add slack and jira link here pls, so that it's obvious from code what we are waiting for.

sfc-gh-asawicki (changes requested):

pkg/sdk/external_tables_integration_test.go
sfc-gh-jcieslak: Added Jira ticket id

pkg/sdk/external_tables_impl.go
sfc-gh-jcieslak: Used findOne

pkg/sdk/external_tables_impl.go
sfc-gh-jcieslak: Used convert with convertRows

pkg/sdk/external_tables.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: It looks like `new` doesn't allocate in global interface assertions, so I can change it for simplicity and consistency with other files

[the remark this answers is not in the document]
sfc-gh-asawicki: Well, we are using both :) For validatable we have sh*tload of `new()` :)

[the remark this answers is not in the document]
sfc-gh-asawicki: (you can close after reading)

pkg/sdk/external_tables.go
```diff
@@ -0,0 +1,472 @@
...
+
+import (
+	"context"
+	"database/sql"
+	"time"
+)
+
+var (
+	_ ExternalTables = (*externalTables)(nil)
+	_ validatable    = (*CreateExternalTableOptions)(nil)
```
sfc-gh-asawicki, line 11: hmm, you still have this line (and other) in this file and in external_tables_validations, why?

pkg/sdk/external_tables_validations.go
```diff
@@ -0,0 +1,247 @@
+package sdk
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	_ validatable = (*CreateExternalTableOptions)(nil)
```
sfc-gh-asawicki, line 9: here.

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

PR #2007, opened 2023-08-10
feat: Migrate roles to the new sdk

sfc-gh-jcieslak:
Migrate roles to the new sdk

## Test Plan
[x] unit tests
[x] integration tests

## References
doc reference - https://docs.snowflake.com/en/sql-reference/sql/create-role


pkg/sdk/roles_integration_test.go
```diff
@@ -0,0 +1,323 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInt_Roles(t *testing.T) {
+	client := testClient(t)
+	ctx := context.Background()
+
+	database, databaseCleanup := createDatabase(t, client)
+	t.Cleanup(databaseCleanup)
```
sfc-gh-asawicki, line 16: Nice, only db cleanup, I like it <3

pkg/sdk/roles_integration_test.go
```diff
@@ -0,0 +1,323 @@
...
+		assert.Equal(t, roleID.Name(), role.Name)
+	})
+
+	t.Run("create if not exists", func(t *testing.T) {
+		roleID := randomAccountObjectIdentifier(t)
+		err := client.Roles.Create(ctx, roleID, &CreateRoleOptions{
+			IfNotExists: Bool(true),
+		})
+		require.NoError(t, err)
+		t.Cleanup(func() {
```
sfc-gh-asawicki, line 42: This cleanup is repeated, and it is used also in helper_test.go, so maybe extract it and reuse it?

sfc-gh-asawicki (changes requested):

pkg/sdk/roles_integration_test.go
sfc-gh-jcieslak: Right, but not much gains here, because roles / users should have their own cleanup functions in the tests

pkg/sdk/roles_integration_test.go
sfc-gh-jcieslak: I actually created createRoleWithOptions which creates role with passed options and it returns the cleanup function. I just didn't see that it's not used everywhere

pkg/sdk/roles_validations.go
```diff
@@ -0,0 +1,57 @@
+package sdk
+
+import "errors"
+
+func (opts *CreateRoleOptions) validateProp() error {
```
sfc-gh-asawicki, line 5: And what about nil options? If we use DTOs we can enforce it, but now nil options here will result in panic

pkg/sdk/roles.go
```diff
@@ -44,51 +48,121 @@ func (v *Role) ObjectType() ObjectType {
...
-	sql := `SHOW ROLES`
-	err := v.client.query(ctx, &rows, sql)
-	if err != nil {
-		return nil, err
-	}
+// ShowRoleOptions based on https://docs.snowflake.com/en/sql-reference/sql/show-roles
+type ShowRoleOptions struct {
+	show  bool  `ddl:"static" sql:"SHOW"`
+	roles bool  `ddl:"static" sql:"ROLES"`
+	Like  *Like `ddl:"keyword" sql:"LIKE"`
```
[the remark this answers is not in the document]
sfc-gh-swinkler: what even is a class? i cannot seem to figure that out.

pkg/sdk/roles_integration_test.go
```diff
@@ -0,0 +1,323 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInt_Roles(t *testing.T) {
```
sfc-gh-swinkler, line 11: this is one long function... is it really not possible to break this up into helper functions?

e.g.
```
func TestInt_Roles(t *testing.T){
// setup work
tests = []func(t *testing.T, ...) error {Test1, Test2, Test3}
for test, _ := range tests {
t.Run(test)
}
}
```

pkg/sdk/roles_integration_test.go
sfc-gh-swinkler: my question is -- do we even need separate functions for createXXX vs createXXXWithOptions? We almost never us the createXXXWithOptions. How about we just have createXXX

pkg/sdk/sessions.go
```diff
@@ -286,3 +288,22 @@ func (v *sessions) UseSchema(ctx context.Context, schema SchemaIdentifier) error
 	_, err := v.client.exec(ctx, sql)
 	return err
 }
+
+func (v *sessions) UseRole(ctx context.Context, role AccountObjectIdentifier) error {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: in context_functions_integration_test.go

pkg/sdk/roles.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, I don't know how I missed that. I'll add it

pkg/sdk/roles.go
```diff
@@ -44,51 +48,121 @@ func (v *Role) ObjectType() ObjectType {
...
+}
 
-	roles := make([]*Role, len(rows))
-	for i, row := range rows {
-		var role Role = row
-		roles[i] = &role
-	}
+func (opts *ShowRoleOptions) validateProp() error {
+	return nil
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: 👍 added

pkg/sdk/roles_validations.go
sfc-gh-jcieslak: added nil checks

pkg/sdk/context_functions.go
```diff
@@ -49,6 +52,40 @@ func (c *contextFunctions) CurrentRole(ctx context.Context) (string, error) {
 	return s.CurrentRole, nil
 }
 
+type CurrentSecondaryRoles struct {
+	Roles []string
+	Value string
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yeah, I created one in the UseSecondaryRoles, but forgot to use it here.

pkg/sdk/roles_integration_test.go
sfc-gh-jcieslak: I used it a few times in this test and I think we can leave with two create functions since creating createXXX function is as simple as calling the one with opts but passing nil or default options there. We can do either, but like I said I used those two in roles_integration_test.go go get the cleanup function for free, so I didn't need to copy cleanup functions from other tests

pkg/sdk/roles_test.go
```diff
@@ -0,0 +1,217 @@
+package sdk
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRolesCreate(t *testing.T) {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: added

pkg/sdk/roles.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: https://docs.snowflake.com/en/sql-reference-classes Classes are similar to Java classes I guess, you create instances of them and call functions on those instances. I think we'll know more about classes when I'll make the integration test for it

pkg/sdk/roles_integration_test.go
sfc-gh-jcieslak: It sounds like this custom framework to encapsulate common test cases, which should be done in its own pr + we discussed that breaking up into multiple functions is the same as breaking up the tests with t.Run() and with one function we can create resources (like database) use it across the tests and then release these resources at the end of all tests

pkg/sdk/roles.go
```diff
@@ -2,38 +2,42 @@ package sdk
...
 	"context"
-	"fmt"
+	"database/sql"
+	"time"
 )
 
 type Roles interface {
 	// Create creates a role.
-	Create(ctx context.Context, id AccountObjectIdentifier, opts *RoleCreateOptions) error
+	Create(ctx context.Context, id AccountObjectIdentifier, opts *CreateRoleOptions) error
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Added dtos :)

pkg/sdk/errors.go
```diff
@@ -8,6 +8,9 @@ import (
 )
 
 var (
+	ErrNilOptions                    = errors.New("options cannot be nil")
```
sfc-gh-asawicki, line 11: why this is exported? do we need these errors to be referenced outside the SDK package?

pkg/sdk/roles_impl.go
```diff
@@ -0,0 +1,66 @@
...
+
+	roles := make([]Role, len(rows))
+	for i, row := range rows {
+		roles[i] = row.toRole()
+	}
+
+	return roles, nil
+}
+
+func (v *roles) ShowByID(ctx context.Context, req *ShowRoleByIdRequest) (*Role, error) {
```
sfc-gh-asawicki, line 39: u can use findOne (check pipes_impl)

pkg/sdk/roles_impl.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (v *roles) Alter(ctx context.Context, req *AlterRoleRequest) error {
+	return validateAndExec(v.client, ctx, req.ToOpts())
+}
+
+func (v *roles) Drop(ctx context.Context, req *DropRoleRequest) error {
+	return validateAndExec(v.client, ctx, req.ToOpts())
+}
+
+func (v *roles) Show(ctx context.Context, req *ShowRoleRequest) ([]Role, error) {
```
sfc-gh-asawicki, line 25: you can use convertRows (check pipes_impl)

pkg/sdk/roles_validations.go
```diff
@@ -0,0 +1,93 @@
+package sdk
+
+import "errors"
+
+func (opts *CreateRoleOptions) validate() error {
```
sfc-gh-asawicki, line 5: please add 
```
var (
	_ validatable = new(CreateRoleOptions)
	...
)
```
block (we have discussed and agreed to have this compiler check present)

pkg/sdk/roles_validations.go
```diff
@@ -0,0 +1,93 @@
...
+	if opts == nil {
+		return ErrNilOptions
```
sfc-gh-asawicki, line 7: I was returning for such case `errors.Join(errNilOptions)`, so that we always return the "same" thing, but I guess that output should be really similar in this case

pkg/sdk/roles.go
```diff
@@ -44,51 +48,115 @@ func (v *Role) ObjectType() ObjectType {
...
+	GrantedRoles    int            `db:"granted_roles"`
+	Owner           sql.NullString `db:"owner"`
+	Comment         sql.NullString `db:"comment"`
+}
 
-func (v *roles) Create(ctx context.Context, id AccountObjectIdentifier, opts *RoleCreateOptions) error {
-	sql := fmt.Sprintf(`CREATE ROLE %v`, id.FullyQualifiedName())
-	_, err := v.client.exec(ctx, sql)
-	return err
+func (row *roleDBRow) toRole() Role {
```
sfc-gh-asawicki, line 64: please make it a `convert` method and add `var _ convertibleRow[Role] = new(roleDBRow)` (check more in database_role.go)

pkg/sdk/errors.go
sfc-gh-jcieslak: No, I'll make them private

pkg/sdk/roles_validations.go
sfc-gh-jcieslak: Yeah, It doesn't matter in this case. The only thing is different is if you want to assert on error with wrapping you're asserting stringError inside joinedError and this is only stringError. I think we can leave it as is, because it may be changed after we'll figure out how do we want handle errors with new custom error type.

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

PR #2012, opened 2023-08-14
feat: Use database role from SDK

sfc-gh-asawicki:
## Description
Use SDK in existing resource and datasource implementations. Remove old implementation.

## Test Plan

* [x] acceptance tests


pkg/datasources/database_roles.go
```diff
@@ -1,11 +1,11 @@
 package datasources
 
 import (
+	"context"
 	"database/sql"
-	"errors"
 	"log"
 
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/snowflake"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
```
sfc-gh-swinkler, line 8:
```
"database": {
		Optional:    true,
		Description: "The database from which to return the database roles from.",
	},
```

this should not be optional, it should be required.

pkg/resources/database_role.go
```diff
@@ -1,21 +1,16 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-const (
-	databaseRoleIDDelimiter = '|'
-)
-
 var databaseRoleSchema = map[string]*schema.Schema{
```
sfc-gh-swinkler, line 14: again, isn't this schema missing some computed attributed like `created_at`, `is_default` and `is_current`?

pkg/datasources/database_roles.go
```diff
@@ -52,34 +52,30 @@ func DatabaseRoles() *schema.Resource {
...
 
-	log.Printf("[DEBUG] list roles: %v", listRoles)
+	databaseRoles := make([]map[string]any, 0, len(extractedDatabaseRoles))
+	for _, databaseRole := range extractedDatabaseRoles {
+		databaseRoleMap := map[string]any{}
 
-	roles := []map[string]interface{}{}
-	for _, role := range listRoles {
-		roleMap := map[string]interface{}{}
+		databaseRoleMap["name"] = databaseRole.Name
```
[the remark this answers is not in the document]
sfc-gh-pbosak: I think that would cause changes in the interface of this resource which is something we were trying to avoid before v1, right?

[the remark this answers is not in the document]
sfc-gh-asawicki: Well, they were the only ones supported till this moment. The goal of this task and this PR was to use SDK from resource and from datasource and not to support all missing functionalities/fields/etc (we have separate epic for this), so I would rather not add them here, in this PR.

pkg/datasources/database_roles.go
sfc-gh-asawicki: I have changed it to Required true.

But one thing that worries me here is that this is a breaking change in theory (in practice omitting database must have resulted in a runtime error, because it was always needed underneath).

pkg/resources/database_role.go
sfc-gh-asawicki: Same as above.

PR #2020, opened 2023-08-18
feat: Convert alert datasource to new sdk

sfc-gh-pbosak:
* adds alert data source to new sdk


pkg/resources/alert.go
```diff
@@ -370,3 +370,48 @@ func DeleteAlert(d *schema.ResourceData, meta interface{}) error {
 	d.SetId("")
 	return nil
 }
+
+func waitResumeAlert(ctx context.Context, client *sdk.Client, id sdk.SchemaObjectIdentifier) error {
```
sfc-gh-asawicki, line 374: nit-picking: these two are almost identical, maybe extract helper?

not-nit-picking: shouldn't we put some info for user who is waiting? (e.g. alert not started yet, retrying in 10sec)

sfc-gh-asawicki (approved):

pkg/resources/alert.go
sfc-gh-pbosak: 1. That's a good suggestion, but after a couple of tries I decided against it. I had to not only extract inputs to the function but also handle returned errors differently. When we also add some additional logging for the users this problem becomes only more acute. Of course you can suggest your own version if you found an elegant solution :) 

2. This again is a good suggestion, I added the logs.
I think it's unfortunate we have to use logs here because this seems like a fairly important message. If the user doesn't set `TF_LOG` flag they won't see them and might think the application hung or whatever.
Also, I set log level to INFO, because I don't consider it a warning or an error but at the same time this message might be lost in all the other logs or not displayed at all.

sfc-gh-swinkler: I think we can make use of more general retry function (perhaps one which allows for retryable errors) or a promise function.
```
alterAlertAction := func(ctx context.Context, client *sdk.Client, id sdk.SchemaObjectIdentifier, action sdk.AlertAction) func() error {
   return func() error {
     err := client.Alerts.Alert(ctx,id, &opts) 
     if err != nil {
       return err
     }
     if alert.State != action {
      // a retryable error
      return error.New("not ready yet")
     }
   return nil
   }
}


Retry(5, 10 time.Seconds, alterAlertAction(sdk.AlertActionSuspend)) error 

```

sfc-gh-pbosak: Ok, I did refactor a separate retry function, I guess I was trying to extract too much into a common function

pkg/resources/alert.go
```diff
@@ -370,3 +370,50 @@ func DeleteAlert(d *schema.ResourceData, meta interface{}) error {
...
+	opts := sdk.AlterAlertOptions{Action: &sdk.AlertActionResume}
+	// try to resume the alert, and verify that it was resumed.
+	// if it's not resumed then try again up until a maximum of 5 times
+	for i := 0; i < 5; i++ {
```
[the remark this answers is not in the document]
sfc-gh-pbosak: done

pkg/datasources/alerts.go
```diff
@@ -82,27 +83,32 @@ func Alerts() *schema.Resource {
...
+		Like: &like,
+	})
+	if err != nil {
+		log.Printf("[DEBUG] failed to list alerts in schema (%s)", d.Id())
 		d.SetId("")
-		return nil
+		return err
 	}
-
 	log.Printf("[DEBUG] list alerts: %v", listAlerts)
```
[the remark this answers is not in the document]
sfc-gh-pbosak: ok

pkg/resources/alert.go
```diff
@@ -370,3 +370,51 @@
...
+		alert, err := client.Alerts.ShowByID(ctx, id)
+		if err != nil {
+			return err, false
+		}
+		return nil, alert.State == sdk.AlertStateSuspended
+	}
+	return retry(5, 10*time.Second, suspendAlert)
+}
+
+func retry(attempts int, sleepDuration time.Duration, f func() (error, bool)) error {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: :o

sfc-gh-asawicki (approved):

sfc-gh-swinkler (approved):

PR #2029, opened 2023-08-23
fix: SNOW-59564 remove default data retention in days value

### LINKED ISSUE #1938: snowflake_object_parameter resource for the key "DATA_RETENTION_TIME_IN_DAYS" is not working as expected
**Provider Version**
0.68.1

The provider version you are using.
0.68.1

**Terraform Version**
1.5.2

The version of Terraform you were using when the bug was encountered.
1.5.2

**Describe the bug**
The documentation says that the data_retention_days argument is deprecated for Snowflake tables. When I use the data_retention_days argument under a table resource, Terraform tells me to use the snowflake_object_parameter resource instead. When I used the snowflake_object_parameter resource, using "DATA_RETENTION_TIME_IN_DAYS" as the key and set the value to 30 and run terraform apply, the data retention time for the table will be set to 30. But if I make changes to the table schema or any other snowflake resources, the data retention time for the table reverts back to 1 (which is the default value for Snowflake time travel)

**Expected behavior**
When I used the snowflake_object_parameter resource, using "DATA_RETENTION_TIME_IN_DAYS" as the key and set the value to 30 and run terraform apply, the data retention time for the table will be set to 30. If I make changes to the table schema or any other snowflake resources, the data retention time for the table should be 30.

When I used the snowflake_object_parameter resource, using "DATA_RETENTION_TIME_IN_DAYS" as the key and set the value to 30 and run terraform apply, the data retention time for the table will be set to 30. If I make changes to the table schema or any other snowflake resources, the data retention time for the table should be 30.

**Code samples and commands**

Step-1:
create a table called "SALES"

resource "snowflake_table" "sales_t" {
  database = "TF_DEMO"
  schema   = "SALES"
  name     = "SALES"
  
  column {
    name = "ID"
    type = "NUMBER(38,0)"
  }

  column {
    name = "SALE_NAME"
    type = "VARCHAR(10)"
  }
}

run below commands to create the above table
terraform init
terraform validate
terraform plan
terraform apply

at this point, data_retention_time of the table is 1 which is the default value.

Step - 2:
change data_retention_time of the table

resource "snowflake_object_parameter" "data_retention_time" {
  key         = "DATA_RETENTION_TIME_IN_DAYS"
  value       = "30"
  object_type = "TABLE"
  object_identifier {
    database = "TF_DEMO"
    schema   = "SALES"
    name      = "SALES"
  }
}

run below commands to change the data_retention_time
terraform init
terraform validate
terraform plan
terraform apply

Now, data_retention_time of the table will be changes to 30 which is correct.

Step-3:
create a new table called "BOOKS"

resource "snowflake_table" "book" {
  database = "TF_DEMO"
  schema   = "SALES"
  name     = "BOOKS"
  
  column {
    name = "ID"
    type = "NUMBER(38,0)"
  }

  column {
    name = "BOOK_NAME"
    type = "VARCHAR(10)"
  }
}

run below commands to create an another table.
terraform init
terraform validate
terraform plan
terraform apply

this is where things get weird. If I check the data_retention_time of the table (SALES) which was created earlier, the value will be changed to 1. This is not the intended behaviour.

sfc-gh-pbosak:
- removes default value for data_retention_time_in_days
Having a default value caused problems when value for `data_retention_time_in_days` was set with a separate resource.
Terraform would read the current value set by the dedicated parameter resource, compare it to the default value for the table resource and plan to update it since the two values don't match. This gave the illusion that creating another table somehow interferes with the first one, where in fact any `terraform update` would give a similar result
- (BREAKING) renames `data_retention_days` -> `data_retention_time_in_days` for consistency
- marks `data_retention_time_in_days` as not deprecated

fixes #1938


sfc-gh-swinkler (changes requested):

pkg/resources/table_acceptance_test.go
```diff
@@ -25,7 +25,7 @@ func TestAcc_Table(t *testing.T) {
 					resource.TestCheckResourceAttr("snowflake_table.test_table", "name", accName),
 					resource.TestCheckResourceAttr("snowflake_table.test_table", "database", accName),
 					resource.TestCheckResourceAttr("snowflake_table.test_table", "schema", accName),
-					resource.TestCheckResourceAttr("snowflake_table.test_table", "data_retention_days", "1"),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "data_retention_time_in_days", "1"),
```
sfc-gh-asawicki, line 28: Do we have any acceptance test showing bad behaviour before and the right one after? I think it would be nice to proceed with bugs by first introducing a test showing the error and then fixing it.

sfc-gh-swinkler: its not really bad behavior, its changing the default value of this parameter, as well as changing the name, and adding documentation on how to use this table resource alongside object parameter.

sfc-gh-swinkler (dismissed):

PR #2045, opened 2023-09-04
chore: Change return type of show

sfc-gh-jcieslak:
Remove unnecessary pointers from Show methods return type + minor changes

### DIFF
--- a/pkg/sdk/accounts.go
+++ b/pkg/sdk/accounts.go
@@ -345,8 +345,8 @@ type accountDBRow struct {
 	IsOrgAdmin                           bool           `db:"is_org_admin"`
 }
 
-func (row accountDBRow) convert() Account {
-	acc := Account{
+func (row accountDBRow) convert() *Account {
+	acc := &Account{
 		OrganizationName:                     row.OrganizationName,
 		AccountName:                          row.AccountName,
 		RegionGroup:                          "",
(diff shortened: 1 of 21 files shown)


pkg/sdk/accounts.go
```diff
@@ -345,8 +345,8 @@ type accountDBRow struct {
 	IsOrgAdmin                           bool           `db:"is_org_admin"`
 }
 
-func (row accountDBRow) toAccount() *Account {
-	acc := &Account{
+func (row accountDBRow) toAccount() Account {
```
sfc-gh-asawicki, line 348: 2 things here:
1. I think that * is better here.
2. Check `convertibleRow` (pipes and database role) -> if you are touching these functions, you can rename them to convert.

sfc-gh-asawicki (changes requested):

pkg/sdk/accounts.go
sfc-gh-jcieslak: 1. I don't know, * means that it may return nil, but that's conversion where theoretically nothing can happen. Btw convert interface maybe should also return T, not *T. (toOpts should also be non pointer, but that can be later discussed). Right now, I'll change convert to convert() T, and rename functions to convert and change their implementation if possible. We can still discuss pointer part and change it in this pr, because it's not a lot of work
2. Sure, I'll also use common Show implementation where possible

sfc-gh-asawicki: Well, that's one of the reasons to use pointers. IMO, the second one is to avoid the instantiation of a new struct every time. So, for me, using a pointer is better in both cases. :D We can discuss it with the whole team, but not necessarily in this PR.

You can rewrite them; why not? :)

sfc-gh-asawicki (approved):

PR #2046, opened 2023-09-04
docs: Update table.md

jkodroff:
Add a link to the Snowflake docs which contain all the valid column types.


sfc-gh-jcieslak (approved):

docs/resources/table.md
```diff
@@ -111,7 +111,7 @@ resource "snowflake_table" "table" {
 Required:
 
 - `name` (String) Column name
-- `type` (String) Column type, e.g. VARIANT
+- `type` (String) Column type, e.g. VARIANT. For a full list of column types, see [Summary of Data Types](https://docs.snowflake.com/en/sql-reference/intro-summary-data-types).
```
sfc-gh-asawicki, line 114: This file is auto-generated, so it should not be updated directly.

You can instead add it to `pkg/resources/table.go` and run `make docs`; that will update this file.

sfc-gh-asawicki (changes requested):

sfc-gh-asawicki:
@jkodroff, thanks for the suggestion!

I have added one comment during the review; can you correct it accordingly?

Also, please add the `docs: ` prefix to the name of the pull request.

PR #2052, opened 2023-09-12
feat: Poc custom error type

sfc-gh-jcieslak:
## Summary

This is a proposal for new custom error type that will be able to provide better error messages that will be more readable for developers and more useful for users.

It contains:
- custom error type in order to create stack trace and prefix errors with file name + line number, because sometimes it's hard to track from where given error was returned originally
- top level error - still thinking about how / where we'll be able return it to be always "top level error" - it will wrap error(s) that were returned from lower levels of program execution. On error it prints useful information for the user that something wrong happened and in order to improve the product user can click the link that will lead to this repository /create-issue endpoint with already partially filled issue description
- some factory methods for creating / wrapping custom (or any other) errors
- tests / examples of usage documentation

Mind TODOs, I left them on purpose where I had an idea or wasn't sure and wanted to discuss certain parts (or where work still has to be done, but I would like to wait till we discuss some topics, before implementing / replacing things)

## Update

After discussions, I've applied the following changes
- previously called Error constructors NewPredefinedError NewPredefinedError2 where not that necessary and for already predefined errors like ErrInvalidObjectIdentifier value and errOneOf function now return sdk.Error with just a NewError for values or newSDKError for function (with function
  - I did it that way, because then file and line information will be retrieved where the function was called and not where it was declared)
  - with values I've left it as is, so the file and line number will be taken from the location of declaration
- error output building was tested
- easy adoption, because we don't have to do much replacing predefined errors with NewError is good enough for now, but I've made example use in dynamic_table_validations.go
- what we could do later on (in validations)
  - remove unnecessary join for single errors
  - replace custom errors (errors.New or fmt.Errorf) with sdk.NewError
  - replace errors.Join with sdk.JoinErrors
  - the rest should be predefined values or functions which already return new error


pkg/sdk/errors.go
```diff
@@ -45,3 +45,174 @@ func decodeDriverError(err error) error {
 
 	return err
 }
+
+const ghIssueBodyTemplate = `
```
sfc-gh-jcieslak, line 49: template should be changed to be more matching with the current one, e.g. fields like Provider version / TF version should also be here

pkg/sdk/errors.go
```diff
@@ -45,3 +45,174 @@
...
+		filename = fileSplit[len(fileSplit)-1]
+	} else {
+		filename = fileSplit[0]
+	}
+	return line, filename
+}
+
+func writeTree(e error, builder *strings.Builder, indent int) {
+	var sdkErr *sdkError
+	if joinedErr, ok := e.(interface{ Unwrap() []error }); ok {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Not true, errors.As will recursively Unwrap to search for first encountered sdkError which is terrible, this should stay as is

pkg/sdk/errors.go
```diff
@@ -45,3 +45,174 @@
...
+			fieldsBuilder.String(),
+		),
+		2,
+	)
+}
+
+// TODO We can force to use sdkError as wrapper
+// WrapErrors TODO
+func WrapErrors(wrapper error, errs ...error) error {
+	if err, ok := wrapper.(*sdkError); ok {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Same as below

pkg/sdk/errors.go
```diff
@@ -48,3 +47,230 @@ func decodeDriverError(err error) error {
...
+-->
+
+Errors:
+%s
+`
+
+// NewTopLevelError wraps an error with "final" error message of sdk.
+// It should be placed in the highest place of call stack to catch as much error context as possible.
+// TODO Is it possible to call this function somewhere high in the callstack, so it always be top level ? If not maybe this would be overkill / too hard to use
+// TODO It's possible to wrap errors multiple times, this function will try to keep only the one with most error context.
```
sfc-gh-jcieslak, line 87: If confusing this TODO and the one under is referring following case
err (type unknown error)
|
- err1 (type SDKError)
- err2 (type TopLevelError)
|
- err3 (type SDKError)

err has child err2 which is TopLevelError, we want to unwrap it and wrap err, so the error will have more context (TopLevelError now will also include err1) and we should end up with

err (TopLevelError)
|
- err1 (SDKError)
- err3 (SDKError

pkg/sdk/errors_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	e := Example{}
+
+	printErr(NewErrorNotSet(&e, e.ExportedField, e.unexportedField, e.unexportedFieldPointer, e.ExportedFieldSlice))
+
+	// Output:
+	// [file:line] fields of struct Example: [ExportedField string, unexportedField int, unexportedFieldPointer *string, ExportedFieldSlice []int] are required and should be set
+}
+
+func ExampleNewTopLevelError() {
```
sfc-gh-swinkler, line 124: if this is a test, then it should be prefixed with "Test" and should take as argument t *testing.T

pkg/sdk/errors.go
```diff
@@ -48,3 +47,230 @@ func decodeDriverError(err error) error {
...
+func NewTopLevelError(err error) error {
+	// TODO if called multiple times, unwrap lower level errors and wrap err to have more context
+
+	return fmt.Errorf(`
+Snowflake Terraform Provider error!
```
sfc-gh-swinkler, line 92: in general, i prefer to keep all references to Terraform Provider out of the SDK. Is there any way this can be specified through the client config?

pkg/sdk/errors.go
```diff
@@ -48,3 +47,230 @@ func decodeDriverError(err error) error {
...
+
+Errors:
+%s`,
+		url.QueryEscape("New issue"),
+		url.QueryEscape(fmt.Sprintf(ghIssueBodyTemplate, err.Error())),
+		err.Error(),
+	)
+}
+
+type SDKError struct { //nolint:all
```
sfc-gh-swinkler, line 108: i don't like SDKError as a type name. because the package is SDK. so it would be sdk.SDKError, which is awkward. Would be better to rename this to SnowflakeError or possibly SnowflakeSDKError

pkg/sdk/errors.go
```diff
@@ -48,3 +47,230 @@ func decodeDriverError(err error) error {
...
+// we would have to change every return statement of this error to function call (because NewPredefinedError returns factory method for predefined err)
+// and the nature of NewPredefinedError2 is that we can return value as with ErrInvalidObjectIdentifier example.
+// The only disadvantage of this is that we'll get file + line from declaration which may cause issues that we have right now
+// (not knowing places where errors have been returned from - one function can return ErrInvalidObjectIdentifier in two places
+// and we'll know from which place with this approach unfortunately) and this error should solve them.
+// We can use both (e.g. for new errors use NewPredefinedError2 and for replacing old use NewPredefinedError)
+// or we can just choose one approach over the other. (of course NewPredefinedError2 will be renamed to something that makes more sense)
+// TODO this approach to predefined errors, this is lazy evaluated error, so the only filename + line number
+// TODO if we're going with this approach, change other methods to handle SDKPredefinedError type
+func NewPredefinedError2(message string) SDKPredefinedError {
```
sfc-gh-swinkler, line 159: this is a terrible name. at least call it "NewPredfinedLegacyError" or something to that effect.

pkg/sdk/errors_test.go
sfc-gh-asawicki: This is a golang "testable example" feature, you can read about it here: https://go.dev/blog/examples.

pkg/sdk/errors.go
sfc-gh-asawicki: Yeah, that's a good point keeping in mind that later we want to extract SDK as a separate project.

pkg/sdk/errors.go
sfc-gh-asawicki: You can add this comment to the comment in code, to make it understandable also from there.

pkg/sdk/errors.go
```diff
@@ -48,3 +47,230 @@ func decodeDriverError(err error) error {
...
+	err := newSDKError(message, 2)
+	return func() error {
+		return err
+	}
+}
+
+func NewErrorOneOf(structPointer any, fields ...any) error {
```
sfc-gh-asawicki, line 166: I am not sure if we need reflection to build such simple errors, TBH I liked more the previous simple way, where fields are just specified.

Moreover if we want to depend on things like that, we have to test them thoroughly.

So for me this is too much of a revolution with this new errors. Addition of callstack file and line number is a valid idea, but I wouldn't go that far. Baby steps.

pkg/sdk/errors.go
```diff
@@ -48,3 +47,230 @@ func decodeDriverError(err error) error {
...
+	var filename string
+	if len(fileSplit) > 1 {
+		filename = fileSplit[len(fileSplit)-1]
+	} else {
+		filename = fileSplit[0]
+	}
+	return line, filename
+}
+
+func writeTree(e error, builder *strings.Builder, indent int, fileInfo bool) {
```
sfc-gh-asawicki, line 253: This method needs to be thoroughly tested.

sfc-gh-asawicki (changes requested):

pkg/sdk/errors.go
```diff
@@ -53,3 +54,85 @@ func decodeDriverError(err error) error {
...
+	}
+	if len(notNilErrs) == 0 {
+		return nil
+	}
+	err := newSDKError("joined error", 2)
+	err.nestedErrors = notNilErrs
+	return err
+}
+
+func newSDKError(message string, skip int, nested ...error) *Error {
```
sfc-gh-swinkler, line 93: instead of calling this SDK, should we call it "SnowflakeError"?

sfc-gh-swinkler (approved):

pkg/sdk/errors.go
sfc-gh-jcieslak: Yeah, forgot about that one. This is pretty much a constructor for sdk.Error, but it was meant to be used only inside errors.go. it's kinda low level in a sense that "skip" parameter can be confusing. I would prefer not to export this and call it "newError" or how you proposed, but in unexported version "snowflakeError" or "newSnowflakeError". I'll rename it to "newError" and comment it that it's supposed to be used only inside this file. Sounds good ?

pkg/sdk/errors_test.go
sfc-gh-jcieslak: Removed, right now the tests are done with Test... functions and output is determined by regex

pkg/sdk/errors.go
sfc-gh-jcieslak: Removed, so no reference to TF Provider is left

pkg/sdk/errors.go
sfc-gh-jcieslak: Changed to sdk.Error

pkg/sdk/errors.go
sfc-gh-jcieslak: Removed (according to discussions), so no changes applied

sfc-gh-asawicki (dismissed):

pkg/sdk/errors.go
sfc-gh-jcieslak: Removed (according to discussions), so no changes applied

pkg/sdk/errors.go
sfc-gh-jcieslak: Added tests

sfc-gh-asawicki (approved):

PR #2059, opened 2023-09-17
fix: add dynamic tables to grants

sfc-gh-swinkler:
* adds dynamic tables to schema object grant   see: https://docs.snowflake.com/en/sql-reference/sql/grant-privilege#syntax


sfc-gh-jcieslak (approved):

pkg/sdk/poc/example/sdk_definitions.go
```diff
@@ -30,16 +30,19 @@ type (
 )
 
 func randomAccountObjectIdentifier(t *testing.T) AccountObjectIdentifier {
+	t.Helper()
```
sfc-gh-asawicki, line 33: this helpers are not necessary here, they are not used (check file description on the top)

sfc-gh-asawicki (approved):

PR #2063, opened 2023-09-21
fix: database grant read

### LINKED ISSUE #1863: Grant read operation fails when a privilege has been granted to a database role
**Provider Version**

0.65.0

**Terraform Version**

1.3.7

**Describe the bug**

Grant read operation fails when a privilege has been granted to a database role:

```
Error: error reading database grant: unknown grantee type DATABASE_ROLE
```

**Expected behavior**

Provider ignores database role grants, as they are not currently supported.

**Code samples and commands**

```hcl
resource "snowflake_database_grant" "grant" {
  database_name = "DB"

  privilege = "USAGE"
  roles     = ["role1", "role2"]
  shares    = ["share1", "share2"]

  with_grant_option = false
}
```

**Executed outside of Terraform**

```sql
GRANT USAGE ON DATABASE DB TO DATABASE ROLE DB.EXAMPLE_ROLE
```

**Additional context**

I guess support for database role grants is on the roadmap, but this can be fixed temporarily by ignoring `DATABASE_ROLE` grants in read operations. It can be as simple as adding the following to https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/grant_helpers.go#L251:

```go
if currentGrant.GranteeType == "DATABASE_ROLE" {
  // Temporary measure so that Terraform will not fail
  // when a database role has been granted privileges
  continue
}
```

and the equivalent to https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/grant_helpers.go#L287:

```go
if futureGrant.GranteeType == "DATABASE_ROLE" {
  // Temporary measure so that Terraform will not fail
  // when a database role has been granted privileges
  continue
}
```

sfc-gh-swinkler:
fixes #1863 
grantee type "database_role" was not recognized in the switch/case statement. adding a default fixes that.

before:
```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: error reading database grant: unknown grantee type DATABASE_ROLE
│
│   with snowflake_database_grant.grant,
│   on main.tf line 38, in resource "snowflake_database_grant" "grant":
│   38: resource "snowflake_database_grant" "grant" {
│
╵
```

after:
```
No changes. Your infrastructure matches the configuration.
```

### DIFF
--- a/pkg/resources/grant_helpers.go
+++ b/pkg/resources/grant_helpers.go
@@ -187,6 +187,8 @@ func readGenericGrant(
 			privileges.addString(grant.Privilege)
 			// Reassign set back
 			sharePrivileges[granteeNameStrippedAccount] = privileges
+		default:
+			log.Printf("[WARN] unexpected grantee type: %s", grant.GranteeType)
 		}
 	}
 


sfc-gh-jcieslak (approved):

pkg/resources/grant_helpers.go
```diff
@@ -187,6 +187,8 @@ func readGenericGrant(
 			privileges.addString(grant.Privilege)
 			// Reassign set back
 			sharePrivileges[granteeNameStrippedAccount] = privileges
+		default:
```
sfc-gh-jcieslak, line 190: How the default case fixes the issue ? Right now, the program behaviour is the same except it logs on unknown case in switch

sfc-gh-jcieslak (changes requested):

pkg/resources/grant_helpers.go
sfc-gh-swinkler: actually this was already fixed in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2037/files
released as part of 0.70.1.

I guess this PR is not needed. it does make it more explicit though

sfc-gh-jcieslak: Hmm, in the pr you've linked there's also only change with just logging which shouldn't break / fix anything, right ? I don't understand why this or linked pr is fixing anything

sfc-gh-jcieslak (approved):

PR #2064, opened 2023-09-21
feat: Add grant ownership to SDK

sfc-gh-asawicki:
## Changes
* Implement granting ownership to schema objects in SDK:
  * Add implementation based on references
* What was NOT changed:
  * Grants interface is still not using our new Requests, new methods are also not using them, to make them consistent inside this interface and not to do too much refactoring in this PR

## Test Plan
* [x] unit tests (sql generation and validations)
* [x] integration tests (sql invocation)
* [x] acceptance tests (nothing should break yet)

## References
* [GRANT OWNERSHIP](https://docs.snowflake.com/en/sql-reference/sql/grant-ownership)


pkg/sdk/grants.go
```diff
@@ -250,3 +252,32 @@ func (row grantRow) convert() *Grant {
...
+// GrantOwnershipOptions is based on https://docs.snowflake.com/en/sql-reference/sql/grant-ownership#syntax.
+type GrantOwnershipOptions struct {
+	grantOwnership bool                    `ddl:"static" sql:"GRANT OWNERSHIP"`
+	On             OwnershipGrantOn        `ddl:"keyword" sql:"ON"`
+	To             OwnershipGrantTo        `ddl:"keyword" sql:"TO"`
+	CurrentGrants  *OwnershipCurrentGrants `ddl:"-"`
+}
+
+type OwnershipGrantOn struct {
+	SchemaObject GrantOnSchemaObject `ddl:"-"`
```
sfc-gh-swinkler, line 265: grant ownership could also be given to account level objects like warehouses and users, not just schema objects

pkg/sdk/grants.go
```diff
@@ -250,3 +252,32 @@ func (row grantRow) convert() *Grant {
...
+	AccountRoleName  *AccountObjectIdentifier  `ddl:"identifier" sql:"ROLE"`
+}
+
+type OwnershipCurrentGrants struct {
+	RevokeOrCopy  OwnershipRevokeOrCopy `ddl:"-"`
+	currentGrants bool                  `ddl:"static" sql:"CURRENT GRANTS"`
+}
+
+type OwnershipRevokeOrCopy struct {
+	// One of
```
sfc-gh-swinkler, line 280: this could be an enum called GrantOwnershipOutboundPrivileges with operations: REVOKE | COPY | NONE

sfc-gh-swinkler (changes requested):

pkg/sdk/grants.go
sfc-gh-asawicki: Not according to the doc which is a base for this change: https://docs.snowflake.com/en/sql-reference/sql/grant-ownership: `Transfers ownership of an object or all objects of a specified type in a schema from one role to another role. Role refers to either a role or a database role.`, and also from syntax.

pkg/sdk/grants.go
sfc-gh-asawicki: NONE is not a valid option, so I don't want to make it more complicated than necessary. I can change to enum with just two values.

sfc-gh-jcieslak (approved):

pkg/sdk/grants.go
sfc-gh-asawicki: Added.

sfc-gh-jcieslak (approved):

sfc-gh-swinkler (approved):

PR #2065, opened 2023-09-22
feat: support drop for accounts and fix read race condition

sfc-gh-swinkler:
Snowflake now supports drop and undrop operations on account

https://docs.snowflake.com/en/sql-reference/sql/drop-account

Also after creating an account, it doesn't always show up immediately, so add in some retry logic.

Finally, comments actually cannot be set on account so changing that back to force new. Also cannot rename an account that is not in the same account so removing that as well


pkg/sdk/accounts.go
```diff
@@ -417,3 +415,56 @@ func (c *accounts) ShowByID(ctx context.Context, id AccountObjectIdentifier) (*A
...
+	return nil
+}
+
+func (c *accounts) Drop(ctx context.Context, id AccountObjectIdentifier, gracePeriodInDays int, opts *DropAccountOptions) error {
+	if opts == nil {
+		opts = &DropAccountOptions{}
+	}
+	opts.name = id
+	opts.gracePeriodInDays = gracePeriodInDays
+	if err := opts.validate(); err != nil {
```
sfc-gh-asawicki, line 443: please use `validateAndExec(v.client, ctx, opts)`

pkg/sdk/accounts_integration_test.go
```diff
@@ -63,6 +60,13 @@ func TestInt_AccountCreate(t *testing.T) {
 		}
 		err = client.Accounts.Create(ctx, accountID, opts)
 		require.NoError(t, err)
+
+		t.Logf("sleeping for 5 seconds to wait for account to be created")
```
sfc-gh-asawicki, line 64: I have two remarks to that, one for now and the other for later:
1. this code is repeated, so please extract the sleep function with the seconds parameter and description parameter.
2. Sleeping is generally not the most excellent solution in tests; we are adding the fixed time, which makes our tests longer. A better solution is to use the await test pattern, which, in general, works like this:
-  there should be a function `awaitUntil` where we can specify timeout and condition check interval
- we use the function above providing a condition to check: in this case, fetch the account and check if there is no error
- method checks the condition several times using the check interval provided and ends with the first success; after the timeout and no successes it fails, and the test should be marked as failed
- this way, we do not wait these 5 seconds every time, but we wait 5 seconds at most

Let's introduce the second idea in a separate PR.

pkg/sdk/accounts_integration_test.go
```diff
@@ -96,6 +107,32 @@ func TestInt_AccountCreate(t *testing.T) {
...
+
+		// drop account
+		err = client.Accounts.Drop(ctx, newAccountID, 3, &DropAccountOptions{
+			IfExists: Bool(true),
+		})
+		require.NoError(t, err)
+
+		// check if account is dropped
+		_, err = client.Accounts.ShowByID(ctx, newAccountID)
+		require.Error(t, err)
```
sfc-gh-asawicki, line 119: can we check that the error matches not existing account?

pkg/sdk/accounts_integration_test.go
```diff
@@ -96,6 +107,32 @@ func TestInt_AccountCreate(t *testing.T) {
...
+
+		// undrop account
+		err = client.Accounts.Undrop(ctx, newAccountID)
+		require.NoError(t, err)
+
+		// check if account is undropped
+		_, err = client.Accounts.ShowByID(ctx, newAccountID)
+		require.NoError(t, err)
+
+		// drop account again
```
sfc-gh-asawicki, line 129: why are we dropping it again?

pkg/resources/account.go
```diff
@@ -335,42 +357,37 @@ func ReadAccount(d *schema.ResourceData, meta interface{}) error {
 
 // UpdateAccount implements schema.UpdateFunc.
 func UpdateAccount(d *schema.ResourceData, meta interface{}) error {
-	db := meta.(*sql.DB)
-	client := sdk.NewClientFromDB(db)
-	ctx := context.Background()
-
-	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
+	/*
```
sfc-gh-asawicki, line 360: do we have to leave this as a comment? if so, please add a description at the top of the comment about why we are leaving this

pkg/sdk/accounts_integration_test.go
```diff
@@ -63,6 +60,13 @@ func TestInt_AccountCreate(t *testing.T) {
 		}
 		err = client.Accounts.Create(ctx, accountID, opts)
 		require.NoError(t, err)
+
+		t.Logf("sleeping for 5 seconds to wait for account to be created")
+		for i := 0; i < 5; i++ {
+			t.Logf("%d", 5-i)
+			time.Sleep(1 * time.Second)
+		}
```
sfc-gh-jcieslak, line 68: maybe you could do 3 tries of ShowByID instead of sleeping (Something similar to Retries you've done in this change) ? I would prefer it that way, because sometimes it could proceed with the test faster than 5 sec and won't add additional seconds to the integration tests

pkg/resources/account.go
```diff
@@ -197,6 +198,12 @@ var accountSchema = map[string]*schema.Schema{
 		Computed:    true,
 		Description: "Indicates whether the ORGADMIN role is enabled in an account. If TRUE, the role is enabled.",
 	},
+	"grace_period_in_days": {
```
sfc-gh-asawicki, line 201: you have to regen docs with new param

sfc-gh-asawicki (changes requested):

pkg/sdk/accounts_integration_test.go
scottwinkler: done

pkg/sdk/accounts.go
scottwinkler: done

pkg/sdk/accounts_integration_test.go
scottwinkler: i have used the github.com/avast/retry-go instead of writing a helper function. we were already using this library somewhere else. i guess it will work.

pkg/sdk/accounts_integration_test.go
scottwinkler: added a new test case for ShowByID

pkg/sdk/accounts_integration_test.go
scottwinkler: create -> drop -> undrop -> drop again

its just a final cleanup.

pkg/resources/account.go
scottwinkler: made a note

pkg/resources/account.go
scottwinkler: done

sfc-gh-jcieslak (approved):

sfc-gh-asawicki (approved):

PR #2078, opened 2023-09-27
docs: Update database resource to remove escaped characters

jreslock:
- The example for the `snowflake_database` resource that uses `from_replica` is incorrect. It currently shows escaped characters in the `from_replica` input which will cause a failure at apply time with terraform 1.5.x/snowflake provider version 0.71.0.

## Test Plan
I did not include a test plan as this is a very small doc change and does not modify or add any functionality.

## References
https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/database


sfc-gh-asawicki:
Superseded by #2527.

Issue #2067, opened 2023-09-22
snowflake_current_account not able to fetch the account url correctly in account spanning multiple regions


mhmcdonald:
**Provider Version**
0.68.2

**Terraform Version**
1.2.5

**Describe the bug**
If you're in an account that spans multiple regions, the data source `snowflake_current_account` is not able to fetch the account's url correctly. The reason for this is related to this [AccountURL() function](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/snowflake/current_account.go#L72). 

When you're in an account in an organization that spans multiple region groups, and you call current_region(), the value returned is prefixed with `PUBLIC.`. It's documented [here](https://docs.snowflake.com/en/sql-reference/functions/current_region#examples). However, in the current implementation of the data source is not expecting the `PUBLIC.` string prefix.  

**Expected behavior**
Regardless of the account you're in, using the snowflake_current_account data source should return a valid URL for the account, and should not raise [this error](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/538efa5ee9b0370e3090cd7fb867f3df4bbfd0ed/pkg/snowflake/current_account.go#L79C17-L79C95).

**Code samples and commands**

Please add code examples and commands that were run to cause the problem.

`
data "snowflake_current_account" "this" {
}

resource "resource_name" "thing" {
     url  = data.snowflake_current_account.this.url
}
`

PR #2080, opened 2023-09-29
fix: allow AccountURL function to handle accounts in orgs spanning multiple regions

mhmcdonald:
This PR addresses an issue with the `snowflake_current_account` data source. Currently, it's not able to return the account url when the account belongs to an organization that spans multiple region groups. It's currently returning [this error](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/538efa5ee9b0370e3090cd7fb867f3df4bbfd0ed/pkg/snowflake/current_account.go#L79C17-L79C95) when it tries to construct the account URL. The underlying cause of this error is that when an organization spans multiple regions, the current_account() function returns `<region_group>.<region>`, which is different behavior from when the organization's accounts are in a single region. 

The current_region logic is documented [here](https://docs.snowflake.com/en/sql-reference/functions/current_region#examples). 

This PR also addresses an error in the regionMapping for the `aws_us_gov_east_1` region. 

This PR changes the query that's used for fetching the current account and current region: it adds a new column to that query output to contain the `region_group`. In case there is ever a need for region_group in the future, I've added that to the CurrentAccount type. 

It also corrects the url string mapping for `aws_us_gov_east_1` region. 

## Test Plan

* [x] acceptance tests

## References

Closes issue [2067](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2067)

### DIFF
--- a/pkg/snowflake/current_account.go
+++ b/pkg/snowflake/current_account.go
@@ -55,9 +55,9 @@ func SelectCurrentAccount() string {
 }
 
 type CurrentAccount struct {
-	Account string `db:"account"`
+	Account     string `db:"account"`
 	RegionGroup string `db:"region_group"`
-	Region  string `db:"region"`
+	Region      string `db:"region"`
 }
 
 func ScanCurrentAccount(row *sqlx.Row) (*CurrentAccount, error) {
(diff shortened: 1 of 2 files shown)


sfc-gh-asawicki:
Hey @mhmcdonald. We all slowly progressing through all issues and PRs in the repository.

The code you have altered is already gone. It was moved during the SDK rewrite to `ContextFunctions` interface. We will address the problems you mentioned as a part of an upcoming rework of existing resources and datasources https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#our-roadmap.

PR #2082, opened 2023-09-29
feat: use SDK in schema resource and datasource

sfc-gh-jcieslak:


pkg/resources/helpers.go
```diff
@@ -61,3 +61,41 @@ func setBoolProperty(d *schema.ResourceData, key string, property *sdk.BoolPrope
...
 }
+
+func getTagObjectIdentifier(v map[string]any) sdk.ObjectIdentifier {
+	if _, ok := v["database"]; ok {
+		if _, ok := v["schema"]; ok {
+			return sdk.NewSchemaObjectIdentifier(v["database"].(string), v["schema"].(string), v["name"].(string))
+		}
+		return sdk.NewDatabaseObjectIdentifier(v["database"].(string), v["name"].(string))
+	}
+	return sdk.NewAccountObjectIdentifier(v["name"].(string))
```
sfc-gh-jcieslak, line 72: Not sure about this. I think tags are database or database+schema objects and without them it cannot be created / retrieved (unless there's active database or database+schema in the client session), so I should return error here. What do you think ?

pkg/datasources/schemas.go
```diff
@@ -48,11 +49,20 @@ func Schemas() *schema.Resource {
...
 
-	currentSchemas, err := snowflake.ListSchemas(databaseName, db)
+	currentSchemas, err := client.Schemas.Show(ctx, &sdk.ShowSchemaOptions{
+		In: &sdk.SchemaIn{
+			Database: sdk.Bool(true),
+			Name:     databaseID,
+		},
+	})
+
 	if errors.Is(err, sql.ErrNoRows) {
```
sfc-gh-asawicki, line 66: I guess that now this won't be the same error.

For other datasources we have moved to plan `err != nil` when migrating to SDK (e.g. check `database_roles.go`).

pkg/resources/schema.go
```diff
@@ -290,60 +226,148 @@ func UpdateSchema(d *schema.ResourceData, meta interface{}) error {
...
+	if err != nil {
 		return fmt.Errorf("error deleting schema %v err = %w", d.Id(), err)
 	}
 
 	d.SetId("")
 
 	return nil
 }
+
+type schemaID struct {
```
sfc-gh-asawicki, line 334: this and the methods below should be gone. Use `helpers.DecodeSnowflakeID`.

pkg/resources/helpers.go
```diff
@@ -61,3 +61,41 @@ func setBoolProperty(d *schema.ResourceData, key string, property *sdk.BoolPrope
...
+				Name:  getTagObjectIdentifier(v),
+				Value: v["value"].(string),
+			}
+		}
+		return to
+	}
+	return nil
+}
+
+func getPropertyAsPointer[T any](d *schema.ResourceData, property string) *T {
```
sfc-gh-asawicki, line 91: add test for it please

pkg/resources/helpers.go
sfc-gh-asawicki: Hmmm, I think tags are indeed on schema level: https://docs.snowflake.com/en/user-guide/object-tagging#what-is-a-tag and I guess this should be the only option here.

Both schema and database are required in schema resource definition, so it is safe to assume they will be present here.

@scottwinkler, can you also chip in?

pkg/resources/helpers.go
```diff
@@ -61,3 +61,41 @@ func setBoolProperty(d *schema.ResourceData, key string, property *sdk.BoolPrope
...
+	if _, ok := v["database"]; ok {
+		if _, ok := v["schema"]; ok {
+			return sdk.NewSchemaObjectIdentifier(v["database"].(string), v["schema"].(string), v["name"].(string))
+		}
+		return sdk.NewDatabaseObjectIdentifier(v["database"].(string), v["name"].(string))
+	}
+	return sdk.NewAccountObjectIdentifier(v["name"].(string))
+}
+
+func getPropertyTags(d *schema.ResourceData, key string) []sdk.TagAssociation {
```
sfc-gh-asawicki, line 75: What was the plan for tags and tags associations @scottwinkler? I can see that now they are supported in two places: for some resource, like schema here, they have `"tag": tagReferenceSchema` specified, but `tagReferenceSchema` has `Deprecated:  "Use the 'snowflake_tag_association' resource instead."`.

So I believe that for now we want to support both options, and with V1 we cut out this deprecated option, right?

sfc-gh-asawicki, line 75: Was it not handled yet in any other resource?

sfc-gh-asawicki (changes requested):

pkg/resources/helpers.go
sfc-gh-jcieslak: Hmm, I wonder what would happen if only database would be specified. It would process with "PUBLIC" schema or threw an error.

pkg/resources/helpers.go
sfc-gh-jcieslak: Not yet, as far as I checked everywhere getTags (resources/tag.go - last function) is used which is used for old SDK tags and this function is for new SDK tags

sfc-gh-swinkler: well i think it is a similar problem to the snowflake_parameters vs having the same attribute be on a resource. We also have to weigh the downside of basically doubling the number of resources, if customers want to tag every resource. i didn't like the implementation of tag blocks previously, because it was difficult to know when the order changed, and also it made for really long configuration code, but I could see maybe adding them back later with a slightly different syntax, e.g.:

```
snowflake_database "d" {
....
tags = [
{
tag_name = ""
tag_value = ""
},
{
tag_name = ""
tag_value = ""
},
...
]
}
```

That being said, if we add tags back in it would be problematic for people who use the existing tag association resource. They would need to set `ignore_changes = ["tags"] ` on every resource which uses tags so as not dissociate every tag they had set. https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes. At the time it was too complicated to have a tags block on every resource, which is why it was made into its own resource. Now that we have a more stable API, I'm not sure its beneficial anymore.

sfc-gh-swinkler (approved):

PR #2088, opened 2023-10-02
feat: Add session policy to SDK

sfc-gh-asawicki:
## Changes
* Add session policy to SDK:
  * Add implementation based on references
  * Remove existing partial implementation
* Misc:
  * Add missing issue number to TODO comments.
  * Fix network policy validation (at least one of -> exactly one of).
  * Comment out printing generations to stdout (because sometimes there are problems with it - and we don't really need it, maybe we can remove it).

As a part of this implementation, the generator was used (check the `session_policies_def.go` file first). After generation, the following changes were applied:
* Filled out unit tests.
* Added unit tests missing from the generated code (validations for precisely one, cases for alter statement).
* Removed generated unnecessary unit tests (usually full and basic were the same in this simple resource).
* Filled out acceptance tests.
* Added mappings (convert functions).
* Added `ID()` method to session policy.
* Added `ShowID()` interface method and implementation (not yet generated).

## Test Plan
* [x] unit tests (SQL generation and validations)
* [x] integration tests (SQL invocation)
* [x] acceptance tests (nothing should break yet)

## References
* [CREATE](https://docs.snowflake.com/en/sql-reference/sql/create-session-policy)
* [ALTER](https://docs.snowflake.com/en/sql-reference/sql/alter-session-policy)
* [DROP](https://docs.snowflake.com/en/sql-reference/sql/drop-session-policy)
* [SHOW](https://docs.snowflake.com/en/sql-reference/sql/show-session-policies)
* [DESCRIBE](https://docs.snowflake.com/en/sql-reference/sql/desc-session-policy)


pkg/sdk/poc/main.go
```diff
@@ -18,14 +18,15 @@ import (
...
+	"session_policies_def.go": sdk.SessionPoliciesDef,
 }
 
 func main() {
 	file := os.Getenv("GOFILE")
 	fmt.Printf("Running generator on %s with args %#v\n", file, os.Args[1:])
 	definition := getDefinition(file)
 
-	runAllTemplatesToStdOut(definition)
+	//runAllTemplatesToStdOut(definition)
```
sfc-gh-asawicki, line 29: It is not working sometimes, I believe that this should be our default.

pkg/sdk/session_policies_def.go
```diff
@@ -0,0 +1,110 @@
...
+).
+	CreateOperation(
+		"https://docs.snowflake.com/en/sql-reference/sql/create-session-policy",
+		g.QueryStruct("CreateSessionPolicy").
+			Create().
+			OrReplace().
+			SQL("SESSION POLICY").
+			IfNotExists().
+			Name().
+			OptionalIntAssignment("SESSION_IDLE_TIMEOUT_MINS", g.ParameterOptions().NoQuotes()).
```
sfc-gh-jcieslak, line 20: Since we decided to have "Text" instead of "String" I would also consider "Number" instead of "Int", but that's nitpicking and either works for me

pkg/sdk/session_policies_def.go
```diff
@@ -0,0 +1,110 @@
...
+			WithValidation(g.ConflictingFields, "OrReplace", "IfNotExists"),
+	).
+	AlterOperation(
+		"https://docs.snowflake.com/en/sql-reference/sql/alter-session-policy",
+		g.QueryStruct("AlterSessionPolicy").
+			Alter().
+			SQL("SESSION POLICY").
+			IfExists().
+			Name().
+			Identifier("RenameTo", g.KindOfTPointer[SchemaObjectIdentifier](), g.IdentifierOptions().SQL("RENAME TO")).
```
sfc-gh-jcieslak, line 33: Should be OptionalIdentifier, but that also works

pkg/sdk/session_policies_def.go
```diff
@@ -0,0 +1,110 @@
...
+		g.DbStruct("showSessionPolicyDBRow").
+			Field("created_on", "string").
+			Field("name", "string").
+			Field("database_name", "string").
+			Field("schema_name", "string").
+			Field("kind", "string").
+			Field("owner", "string").
+			Field("comment", "string").
+			Field("options", "string"),
+		g.PlainStruct("SessionPolicy").
```
sfc-gh-jcieslak, line 77: Hmm, now when I look at it, maybe PlainStruct could be omitted and should be auto-generated from the DbStruct defined above, because it will always have the same number of fields with similar name and type (there could be some kind of options for unusual cases)

sfc-gh-jcieslak (approved):

pkg/sdk/session_policies_def.go
sfc-gh-asawicki: I will change it, you are right

pkg/sdk/session_policies_def.go
sfc-gh-asawicki: Yes, I think we will tackle this with the mapping topics (there are at least two mentioned in the readme).

Not changing anything now.

pkg/sdk/session_policies_def.go
sfc-gh-asawicki: Done.

pkg/sdk/session_policies_def.go
sfc-gh-asawicki: Done.

sfc-gh-jcieslak (approved):

PR #2104, opened 2023-10-09
feat: dynamic table resource + data source

sfc-gh-swinkler:
Adds a Terraform managed resource for dynamic table, and corresponding data source. Also refactor acceptance tests into separate package, with some utility functions, and upgrade to latest test package used by plugin framework.


pkg/sdk/file_format.go
```diff
@@ -1,3 +1,4 @@
+//lint:file-ignore U1000 Ignore all unused code, it's generated
```
sfc-gh-asawicki, line 1: is it really generated?

pkg/datasources/dynamic_tables.go
```diff
@@ -0,0 +1,293 @@
...
+	"context"
+	"database/sql"
+	"log"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var dynamicTablesSchema = map[string]*schema.Schema{
+	"like": {
```
sfc-gh-asawicki, line 13: Is this the first datasource using like + in + startsWith + limit? (I've searched a bit, and I see that database datasource - `pkg/datasource/databases.go` - has both `pattern` and `starts_with`)

I am asking because definitions for these four parameters should be the same (or almost the same) throughout our codebase. Maybe we could extract their descriptions? Or even extract `Schema`s for them (I mean for like, in, etc)?

This shouldn't be done inside this PR of course, but I am putting it here as a food for thought.

pkg/datasources/dynamic_tables_acceptance_test.go
```diff
@@ -0,0 +1,100 @@
...
+					resource.TestCheckResourceAttrSet(dataSourceName, "records.0.data_timestamp"),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckDynamicTableDestroy(s *terraform.State) error {
+	db := acc.TestAccProvider.Meta().(*sql.DB)
+	client := sdk.NewClientFromDB(db)
```
sfc-gh-asawicki, line 87: also here it would be nice to not initialize the client again

pkg/resources/dynamic_table_test.go
```diff
@@ -0,0 +1,36 @@
...
+	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func init() {
+	resource.AddTestSweepers("snowflake_dynamic_table", &resource.Sweeper{
+		Name: "snowflake_dynamic_table",
+		F: func(profile string) error {
+			client, err := sdk.NewDefaultClient()
```
sfc-gh-asawicki, line 16: And here this is initialized again.

pkg/resources/dynamic_table_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+	"fmt"
+
+	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func init() {
+	resource.AddTestSweepers("snowflake_dynamic_table", &resource.Sweeper{
```
sfc-gh-asawicki, line 13: I like that we can addSweepers like this. 👌

pkg/resources/helpers_test.go
```diff
@@ -6,10 +6,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/stretchr/testify/require"
 
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources"
 )
 
+// todo: remove the rest of these which are not used. also this file should be renamed for clarity to make it clear it is for testing only
```
sfc-gh-asawicki, line 12: Can we have a jira to handle this, so that this is not left hanging here. You can add it e.g. to SNOW-926148 and put the number here in the comment

pkg/acceptance/testing.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func TestAccProviders() map[string]*schema.Provider {
+	p := provider.Provider()
+	return map[string]*schema.Provider{
+		"snowflake": p,
+	}
+}
+
+func TestAccPreCheck(t *testing.T) {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I have left multiple comments around the `sdk.NewDefaultClient()` initialization. Let's not correct it as a part of this PR, I have added this as a topic to remember to SNOW-926148 (SNOW-935947).

sfc-gh-asawicki (changes requested):

pkg/acceptance/testing.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/hashicorp/terraform-plugin-mux/tf5to6server"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+)
+
+func TestMain(m *testing.M) {
+	resource.TestMain(m)
+}
+
+var (
```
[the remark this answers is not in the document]
sfc-gh-swinkler: fixed

pkg/acceptance/testing.go
```diff
@@ -0,0 +1,86 @@
...
+
+func TestAccPreCheck(t *testing.T) {
+	client, err := sdk.NewDefaultClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ctx := context.Background()
+
+	// create test database
+	dbId := sdk.NewAccountObjectIdentifier("terraform_test_database")
```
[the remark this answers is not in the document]
sfc-gh-swinkler: fixed

pkg/acceptance/testing.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func TestAccPreCheck(t *testing.T) {
+	client, err := sdk.NewDefaultClient()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ctx := context.Background()
+
+	// create test database
```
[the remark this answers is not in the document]
sfc-gh-swinkler: i simplified it, so no longer needing to do the read before the create. since its already only one function, I don't think breaking it into separate private functions makes sense

pkg/acceptance/testing.go
```diff
@@ -0,0 +1,86 @@
...
+	dbId := sdk.NewAccountObjectIdentifier("terraform_test_database")
+	_, err = client.Databases.ShowByID(ctx, dbId)
+	if err != nil {
+		if err := client.Databases.Create(ctx, dbId, &sdk.CreateDatabaseOptions{
+			OrReplace: sdk.Bool(true),
```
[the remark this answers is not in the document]
sfc-gh-swinkler: ifnotexists is okay

pkg/acceptance/testing.go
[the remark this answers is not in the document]
sfc-gh-swinkler: putting this in test main is okay, I suppose.i also removed the client init  everytime form pre-check

pkg/sdk/file_format.go
sfc-gh-swinkler: no haha i just was getting annoyed by the linter showing me all the warnings of unused variables. i will remove it

pkg/datasources/dynamic_tables_acceptance_test.go
sfc-gh-swinkler: i would like to take a page out of azure terraform provider playbook and put this into some helper methods in the acceptance package.  this destroy method would be a good candidate. actually you could argue we don't even need this function, since destroying the test table already removes the dynamic table as it is a schema level object. i just have this here as a placeholder mainly.

pkg/resources/dynamic_table_test.go
sfc-gh-swinkler: i think we only need a handful of sweeper functions. one for each account level object. then we can use "make sweep" to remove them. actually this code is not working right now, its just placeholder for "best practice". as i was reading through all the documentation on how to write acceptance tests for terraform, i saw this one and thought we should at least have a placeholder implementation

sfc-gh-asawicki (approved):

pkg/datasources/external_functions_acceptance_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
```
sfc-gh-jcieslak, line 8: should be also updated

pkg/resources/dynamic_table_acceptance_test.go
```diff
@@ -0,0 +1,118 @@
...
+				ConfigVariables:   variableSet2,
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func testAccCheckDynamicTableDestroy(s *terraform.State) error {
```
sfc-gh-jcieslak, line 103: I think check destroy functions could be made into one common function that would do the same logic 

and for uncommon things like 
```go
id := sdk.NewSchemaObjectIdentifier(rs.Primary.Attributes["database"], rs.Primary.Attributes["schema"], rs.Primary.Attributes["name"])
dynamicTable, err := client.DynamicTables.ShowByID(ctx, id)
```
there would be a function passed as a parameter that receives resource ("rs") and client and returns an error that is common operation (or the err could be checked in the function)

sfc-gh-swinkler: true, but i don't want to do that right now. i think we will need to have a separate discussion on this, similar to what was done for integration tests. for now, at least it works

pkg/datasources/dynamic_tables_acceptance_test.go
```diff
@@ -0,0 +1,100 @@
...
+					resource.TestCheckResourceAttrSet(dataSourceName, "records.0.is_clone"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "records.0.is_replica"),
+					resource.TestCheckResourceAttrSet(dataSourceName, "records.0.data_timestamp"),
+				),
+			},
+		},
+	})
+}
+
+func testAccCheckDynamicTableDestroy(s *terraform.State) error {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: this is the more general solution, in case there were multiple such resources declared in the configuration. also if you did it the way you suggest, then there would need to be a wrapper function because the interface would not be the same as what is expected for a checkDestroy function

pkg/datasources/dynamic_tables_acceptance_test.go
```diff
@@ -0,0 +1,100 @@
...
+			"name":       config.StringVariable(name),
+			"database":   config.StringVariable(acc.TestDatabaseName),
+			"schema":     config.StringVariable(acc.TestSchemaName),
+			"warehouse":  config.StringVariable(acc.TestWarehouseName),
+			"query":      config.StringVariable(fmt.Sprintf("select \"id\" from \"%v\".\"%v\".\"%v\"", acc.TestDatabaseName, acc.TestSchemaName, tableName)),
+			"comment":    config.StringVariable("Terraform acceptance test"),
+			"table_name": config.StringVariable(tableName),
+		}
+	}
+	variableSet1 := m()
```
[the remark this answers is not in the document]
sfc-gh-swinkler: ok

pkg/datasources/dynamic_tables_acceptance_test.go
```diff
@@ -0,0 +1,100 @@
...
+	name := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
+	dataSourceName := "data.snowflake_dynamic_tables.dts"
+	tableName := name + "_table"
+	m := func() map[string]config.Variable {
+		return map[string]config.Variable{
+			"name":       config.StringVariable(name),
+			"database":   config.StringVariable(acc.TestDatabaseName),
+			"schema":     config.StringVariable(acc.TestSchemaName),
+			"warehouse":  config.StringVariable(acc.TestWarehouseName),
+			"query":      config.StringVariable(fmt.Sprintf("select \"id\" from \"%v\".\"%v\".\"%v\"", acc.TestDatabaseName, acc.TestSchemaName, tableName)),
```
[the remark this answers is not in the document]
sfc-gh-swinkler: good point

sfc-gh-asawicki (approved):

pkg/resources/dynamic_table.go
```diff
@@ -0,0 +1,342 @@
...
+		if err := d.Set("or_replace", true); err != nil {
+			return err
+		}
+	} else {
+		if err := d.Set("or_replace", false); err != nil {
+			return err
+		}
+	}
+	// trim up to " ..AS" and remove whitespace
+	query := strings.TrimSpace(text[strings.Index(text, "AS")+3:])
```
rmargarint-nydig, line 202: Hello. Turns out this will be incorrect when anything in the table description text contains the 'AS' letter sequence, either the DB name, schema name, table name, comment, or warehouse name. It will end up splitting at the incorrect location.

PR #1989, opened 2023-08-01
feat: [PoC] Set up integration tests - with setup

sfc-gh-asawicki:
## Changes
This is a short PoC of introducing [TestMain wrapper](https://pkg.go.dev/testing#hdr-Main) to all integration tests to avoid common setup and cleanup (with the aim to reduce boilerplate and tests time).

There are two PRs:
- #1988 - the one used as an reference
- #1989 - this one where test client and context are set up just once

On local machine the difference in running example test was around 20% time reduction (accordingly without and with set up):

```txt
2023/08/01 13:41:02 [DEBUG] tests took 35.309400583s
ok      github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/integration_tests        35.629s
```

```txt
2023/08/01 13:39:18 [DEBUG] tests took 27.170710792s
ok      github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/integration_tests        28.069s
```

## Test Plan
* [x] none, this is just PoC

PR #2111, opened 2023-10-10
feat: Move integration tests to separate package

sfc-gh-asawicki:
This is the first PR to have SDK integration tests in a separate package. It is based on PoC: https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1989.

It was essential to move the tests fast so that we could work on them in isolation and that there were not many strange conflicts to resolve. A few issues will be handled in subsequent PRs; they are described below.

## Changes
* (Almost) all integration tests moved to separate package
  * setup I contained in the `setup_integration_test.go` file. It introduces two temporary methods, `testClient` and `testContext,` allowing easier test transition to the separate package. They should be inlined in subsequent changes.
  * `client_integration_test.go` was left in the SDK package because this test is a bit different; it tests the client itself and not any resource in particular; therefore, its setup and behavior are different (I think it should be left there).
  * This required prefixing some of the methods with `sdk.`.
  * Some unexported methods were needed in tests (and helpers), so they had to be exported somehow. To avoid big renames throughout the SDK package in this PR, they were copied/exported by delegating in the `integration_test_imports.go` file. The file contains a description; it will be one of the issues that must be handled in subsequent PRs.
  * `helper_test.go` were almost completely moved to `integration_tests/helpers.go` file. Some methods are still needed in the SDK package (for sweepers and the aforementioned `client_integration_test.go`). It may change when we decide what to do with the sweepers.
  * Two new precisely the same files, `random.go`, were introduced in both packages. We need to decide how to handle the need to have randomly generated content in both the sdk and integration_tests packages. A comment is left in the `random.go` file in the integration_tests package.

## Next steps (for subsequent PRs) - summarized:
- decide what is the right way to reuse client throughout the tests (currently accessor function `testClient` vs. `itc.client` vs. top-level `client` vs. something else?); the same with the `testContext` function
- deal with the contents of `integration_test_imports.go` file
- deal with two `random.go` files
- decide what to do with sweepers
- the tests were only moved, and so the common setup was not extracted:
  - we can extract setting up database/schema to the `testMain` method
  - we can merge separate functions in a single file to have less setup steps 

## Test Plan
* [x] unit tests (nothing should break)
* [x] integration tests (the essence of the tests was not changed, so they should pass)
* [x] acceptance tests (nothing should break)

## References
* https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1989


pkg/sdk/integration_test_imports.go
```diff
@@ -0,0 +1,75 @@
...
+// For each of them we will have to decide what do we do:
+// - do we expose the field/method (e.g. errors or ValidObjectIdentifier)
+// - do we keep the workaround
+// - do we copy the code (e.g. ExecForTests)
+// - do we move the code to other place and use it from both places (e.g. findOne)
+// - something else.
+// This will be handled in subsequent PRs, so that the main difficulty (moving) is already merged.
+
+var (
+	ErrObjectNotExistOrAuthorized = errObjectNotExistOrAuthorized
```
sfc-gh-swinkler, line 21: i don't have an issue making errors public

pkg/sdk/integration_test_imports.go
```diff
@@ -0,0 +1,75 @@
...
+	return result, decodeDriverError(err)
+}
+
+// ValidObjectIdentifier is just a delegate to existing unexported validObjectidentifier
+func ValidObjectIdentifier(objectIdentifier ObjectIdentifier) bool {
+	return validObjectidentifier(objectIdentifier)
+}
+
+// GetName is just an accessor to unexported name field
+func (r *CreateNetworkPolicyRequest) GetName() AccountObjectIdentifier {
```
sfc-gh-swinkler, line 38: Looks like we could have the helper functions which create network policy / other test objects also return the id. e.g. `req, name := getDefaultNetworkPolicy()`. So that way we do not need these getter functions. However, there is also no particular reason not to support getter functions. Perhaps we should add getter functions for all the request objects.

pkg/sdk/integration_test_imports.go
```diff
@@ -0,0 +1,75 @@
...
+	return r.name
+}
+
+// GetColumns is just an accessor to unexported name field
+func (s *CreateExternalTableRequest) GetColumns() []*ExternalTableColumnRequest {
+	return s.columns
+}
+
+// GetAccountLocator is an accessor to unexported accountLocator, which is needed in some tests
+func (c *Client) GetAccountLocator() string {
```
sfc-gh-swinkler, line 63: i don't have a problem moving this to the config.go file. I think this plus the GetConfig() function are useful in general, not just for int tests.

pkg/sdk/integration_test_imports.go
```diff
@@ -0,0 +1,75 @@
...
+	return c.accountLocator
+}
+
+// GetConfig is an accessor to unexported config, which is needed in some tests
+func (c *Client) GetConfig() *gosnowflake.Config {
+	return c.config
+}
+
+// FindOne just delegates to our util findOne from SDK
```
sfc-gh-swinkler, line 72: i guess this begs the questions: should we move such helper functions into their own package, and have them all be exported? i don't see why not.

pkg/sdk/integration_tests/accounts_integration_test.go
```diff
@@ -1,13 +1,13 @@
-package sdk
+package sdk_integration_tests
```
sfc-gh-swinkler, line 2: This package name goes againt golang convention. https://go.dev/blog/package-names. recommend to rename to simply "integration"

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps. They are often simple nouns, such as:
> time (provides functionality for measuring and displaying time)
> list (implements a doubly linked list)
> http (provides HTTP client and server implementations)
> The style of names typical of another language might not be idiomatic in a Go program. Here are two examples of names that might be good style in other languages but do not fit well in Go:
> 
> computeServiceClient
> priority_queue

sfc-gh-swinkler (changes requested):

pkg/sdk/integration_tests/accounts_integration_test.go
sfc-gh-asawicki: Good point. I do not like simple integration though because I would expect that integration contains some plugins to integrate with some external systems or something like that. We usually start the name of the integration test with `TestInt_` prefix, so maybe let's rename it to `testint`?

pkg/sdk/integration_test_imports.go
sfc-gh-jcieslak: I think we can consider using internal packages for utils / helpers and other similar packages.
Here's more on internal packages - https://go.dev/doc/go1.4#internalpackages

sfc-gh-asawicki: Yes, that's what I would also recommend for this particular one. @sfc-gh-jcieslak do you agree?

Still, I would prefer to work on the contents of this file in subsequent PR to not inflate this PR any further (which would happen by using extracted util in existing places).

pkg/sdk/integration_test_imports.go
sfc-gh-asawicki: I will add it to our generator improvement doc, to generate at least a getter for the name for now.

pkg/sdk/integration_test_imports.go
```diff
@@ -0,0 +1,75 @@
...
+// - something else.
+// This will be handled in subsequent PRs, so that the main difficulty (moving) is already merged.
+
+var (
+	ErrObjectNotExistOrAuthorized = errObjectNotExistOrAuthorized
+	ErrDifferentDatabase          = errDifferentDatabase
+)
+
+// ExecForTests is an exact copy of exec (that is unexported), that some integration tests/helpers were using
+func (c *Client) ExecForTests(ctx context.Context, sql string) (sql.Result, error) {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Yes, it's just temporary, I didn't want to do refactoring in this PR.

sfc-gh-jcieslak (approved):

pkg/sdk/integration_tests/accounts_integration_test.go
sfc-gh-asawicki: Done.

pkg/sdk/integration_test_imports.go
sfc-gh-asawicki: As discussed I will do it in separate PR.

pkg/sdk/integration_test_imports.go
sfc-gh-asawicki: As discussed I will do it in separate PR.

pkg/sdk/integration_test_imports.go
sfc-gh-asawicki: As discussed I will do it in separate PR.

pkg/sdk/integration_test_imports.go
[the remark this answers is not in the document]
sfc-gh-asawicki: As discussed I will do it in separate PR.

pkg/sdk/integration_test_imports.go
sfc-gh-asawicki: As discussed I will do it in separate PR.

PR #2115, opened 2023-10-11
chore: Adjust integration tests after moving to separate package

sfc-gh-asawicki:
This is the second PR to have SDK integration tests in a separate package. It is a continuation of #2111. It deals with the previous discussions and addresses the topics mentioned in the last PR.

## Changes
* deal with the contents of `integration_test_imports.go` file:
  * there is only one method left (with the comment)
  * errors exported from sdk
  * getter methods added to some request structs
  * validation method extracted
* deal with two `random.go` files
  * `random.go` from the integration tests package is entirely removed
  * common randoms are moved to `pkg/sdk/internal/random.go`
  * SDK-specific random helpers are left with a comment in the `sdk` package
  * all `randomX` functions signatures were changed (`t` parameter was removed)
* generator templates and code had to be adjusted to methods moved and/or renamed

## Next steps (for subsequent PRs) - summarized:
- decide what is the right way to reuse client throughout the tests (currently accessor function `testClient` vs. `itc.client` vs. top-level `client` vs. something else?); the same with the `testContext` function
- decide what to do with sweepers
- the tests were only moved, and so the common setup was not extracted:
  - we can extract setting up database/schema to the `testMain` method
  - we can merge separate functions in a single file to have fewer setup steps 

## Test Plan
* [x] unit tests (nothing should break)
* [x] integration tests (the essence of the tests was not changed, so they should pass)
* [x] acceptance tests (nothing should break)

## References
* https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/1989 
* https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2111


pkg/sdk/database_role_test.go
```diff
@@ -2,10 +2,12 @@ package sdk
 
 import (
 	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/internal/random"
 )
 
 func TestDatabaseRoleCreate(t *testing.T) {
-	id := randomDatabaseObjectIdentifier(t)
+	id := RandomDatabaseObjectIdentifier(t)
```
sfc-gh-swinkler, line 10: shouldnt this be random.DatabaseObjectIdentifier()?

pkg/sdk/database_role_test.go
```diff
@@ -15,13 +17,13 @@ func TestDatabaseRoleCreate(t *testing.T) {
 
 	t.Run("validation: nil options", func(t *testing.T) {
 		var opts *createDatabaseRoleOptions = nil
-		assertOptsInvalidJoinedErrors(t, opts, errNilOptions)
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
```
sfc-gh-swinkler, line 20: should errors also be move into the internal package? Especially since jan is going to do some work with that. i guess it work make package imports weird. like errors.ErrNilOptions, also golang doesn't like it when you overwrite the standard package names

pkg/sdk/internal/collections/collection_helpers.go
```diff
@@ -0,0 +1,16 @@
+package collections
+
+import (
+	"errors"
+)
+
+var ErrObjectNotFound = errors.New("object does not exist")
+
+func FindOne[T any](collection []T, condition func(T) bool) (*T, error) {
```
sfc-gh-swinkler, line 9: this feels like such a sad package. wondering what else we can put in here. Also does it make sense to have the error in this package included? then elsewhere it would have to be referenced as collections.ErrObjectNotFound

pkg/sdk/internal/random/random_helpers.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"testing"
+
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/hashicorp/go-uuid"
+	"github.com/stretchr/testify/require"
+)
+
+func Uuid(t *testing.T) string {
```
sfc-gh-swinkler, line 11: my friend, this should be UUID

sfc-gh-asawicki: I think we have another discussion coming. :D I strongly prefer having abbreviations written in only the first letter capitalized, but you are right; for now, our codebase has them capitalized, so I will rename them.

pkg/sdk/internal/collections/collection_helpers.go
sfc-gh-asawicki: Well, there may be much more convenience methods for collections there, and even if it stays "sad", I don't have a problem with it ;)

About errors, we discussed with @sfc-gh-jcieslak today; we were considering three options:
1. have this error here
2. move SDK errors to the internal package and use one of them here
3. change the signature of the method so that error type can be given to this function

We have decided to go with 1. However, based on your other comment above about moving errors to internal, I will go with 2.

pkg/sdk/database_role_test.go
sfc-gh-asawicki: I will move them (to package `sdkerr` for now). And we will revisit this topic with the errors rework. cc: @sfc-gh-jcieslak

pkg/sdk/database_role_test.go
sfc-gh-asawicki: I started doing this earlier, but:
- these `RandomXIdentifier(t)` are using underneath the `random.Y(t)`; it means that I would have to rework also `random.Y(t)` ones
- some `random.Y(t)` on the other hand are using `require.NoError(t, err)` which require this `t` param.

But I agree that having them without the `t` part is better, so I will rework them.

sfc-gh-asawicki: Done.

pkg/sdk/internal/random/random_helpers.go
sfc-gh-asawicki: Done.

pkg/sdk/random.go
```diff
@@ -1,77 +1,43 @@
...
-	return NewDatabaseObjectIdentifier(randomAlphanumericN(t, 12), randomAlphanumericN(t, 12))
-}
-
-func randomAccountObjectIdentifier(t *testing.T) AccountObjectIdentifier {
-	t.Helper()
-	return NewAccountObjectIdentifier(randomStringN(t, 12))
+// Helper methods in this file are used both in SDK tests and also in integration tests.
+// These methods are not supposed to be used in production code, so it would be better to not export them.
+// Simply moving them to internal/random along other random helper methods is not an easy option, because it would create import cycle:
+// sdk needed here -> sdk imports this to use the methods for tests.
```
sfc-gh-jcieslak, line 10: maybe moving unit tests into test package could be an option ?

sfc-gh-jcieslak: Didn't read the solutions below, but after reading them, separate package for unit tests seems the best solution (imo). Don't know if separate package or sdk_test would be better, but the overall idea of extracting unit tests is pretty good + solution to avoid circular deps.

pkg/sdk/poc/README.md
```diff
@@ -97,6 +97,8 @@ find a better solution to solve the issue (add more logic to the templates ?)
 - generate full tests for common types (e.g. setting/unsetting tags)
 - generate common resources for integration tests
 - cleanup the design of builders in DSL (e.g. why transformer has to be always added?)
+- generate getters for requests, at least for identifier/name
```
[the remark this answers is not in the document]
sfc-gh-asawicki: getter, exported field can be modified, and ultimately we don't want to allow it (only through our builder methods)

pkg/sdk/random.go
sfc-gh-asawicki: Thanks for the response. I like the most the option with extracting identifiers to separate package inside SDK package.

The option with separate package for tests is also fine, if we agree to change the approach in unit tests for black-box one.

sfc-gh-swinkler (approved):

PR #2118, opened 2023-10-12
chore: Integration tests setup db once

sfc-gh-asawicki:
After moving integration tests to a separate package in #2111 and #2115, it is possible now to reuse a common setup. This PR is about setting up a database only once for all integration tests and cleaning it up at the end.

## Changes
* Create a database in integration tests setup; clean it up in cleanup.
* Not every test uses the created database (database tests, renames, and tests specifically needing two databases are still creating their own database).
* Database id is the same string with added UUID so that different test runs do not interfere with each other.

### DIFF
--- a/pkg/sdk/testint/setup_integration_test.go
+++ b/pkg/sdk/testint/setup_integration_test.go
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
-	"github.com/stretchr/testify/require"
+	"github.com/hashicorp/go-uuid"
 )
 
 var itc integrationTestContext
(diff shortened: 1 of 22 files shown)


pkg/sdk/testint/setup_integration_test.go
```diff
@@ -31,27 +32,59 @@ func setup() {
...
+	u, err := uuid.GenerateUUID()
+	if err != nil {
+		return nil, nil, err
+	}
+	id := sdk.NewAccountObjectIdentifier("int_test_db_" + u)
+	err = client.Databases.Create(ctx, id, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	database, err := client.Databases.ShowByID(ctx, id)
```
sfc-gh-jcieslak, line 82: check the err or ignore with _ if you're not using it

pkg/sdk/testint/setup_integration_test.go
```diff
@@ -31,27 +32,59 @@ func setup() {
...
 		log.Printf("Integration test context initialisation failed with %s\n", err)
+		cleanup()
 		os.Exit(1)
 	}
 }
 
 func cleanup() {
 	log.Println("Running integration tests cleanup")
+	if itc.databaseCleanup != nil {
+		defer itc.databaseCleanup()
```
sfc-gh-jcieslak, line 43: I think defer is not needed in this context

sfc-gh-swinkler (approved):

pkg/sdk/testint/setup_integration_test.go
sfc-gh-asawicki: It will be when we have more resources to clean here.

pkg/sdk/testint/setup_integration_test.go
sfc-gh-asawicki: what do you mean? It is returned from the method.

sfc-gh-jcieslak: ah, you're right, didn't see it

sfc-gh-jcieslak (approved):

PR #2119, opened 2023-10-13
chore: Use helper methods in old unit tests

sfc-gh-asawicki:
Use our helper methods `assertOptsValidAndSQLEquals` and `assertOptsInvalid` in older unit tests.

## Changes
* Replaced existing assertions with oneliner.
* Fixed some of the tests and added some more validations. Many resources still need essential validations (issue incoming).
  * The main reason for tests failing was that the previous implementation was only creating opts and invoking `structToSQL,` whereas our helper functions are also calling `validate().` So, some of the generated SQLs that were tested are not achievable through a normal execution and were not correct before. Wrong setups were also corrected.
* Added some TODOs, which should be addressed later in another change (I will create an issue with the description).


pkg/sdk/file_format_test.go
```diff
@@ -85,14 +75,10 @@ func TestFileFormatsCreate(t *testing.T) {
 				JSONStripOuterArray:          Bool(true),
 				JSONStripNullValues:          Bool(true),
 				JSONReplaceInvalidCharacters: Bool(true),
-				JSONIgnoreUTF8Errors:         Bool(true),
```
sfc-gh-jcieslak, line 88: why was it removed ?

pkg/sdk/file_format_test.go
```diff
@@ -146,17 +126,13 @@ func TestFileFormatsCreate(t *testing.T) {
 
 			FileFormatTypeOptions: FileFormatTypeOptions{
 				ParquetCompression:              &ParquetCompressionLzo,
-				ParquetSnappyCompression:        Bool(true),
```
sfc-gh-jcieslak, line 149: why was this removed ?

pkg/sdk/file_format_test.go
```diff
@@ -168,22 +144,18 @@ func TestFileFormatsCreate(t *testing.T) {
 			Type:        FileFormatTypeXML,
 
 			FileFormatTypeOptions: FileFormatTypeOptions{
-				XMLCompression:              &XMLCompressionZstd,
-				XMLIgnoreUTF8Errors:         Bool(true),
-				XMLPreserveSpace:            Bool(true),
-				XMLStripOuterElement:        Bool(true),
-				XMLDisableSnowflakeData:     Bool(true),
-				XMLDisableAutoConvert:       Bool(true),
-				XMLReplaceInvalidCharacters: Bool(true),
```
sfc-gh-jcieslak, line 177: why was this removed ?

pkg/sdk/file_format_test.go
sfc-gh-asawicki: Because this is not valid configuration (cant be used with JSONReplaceInvalidCharacters)

pkg/sdk/file_format_test.go
sfc-gh-asawicki: Because this is not valid configuration (cant be used with ParquetCompression)

pkg/sdk/file_format_test.go
sfc-gh-asawicki: Because this is not valid configuration (cant be used with XMLIgnoreUTF8Errors)

sfc-gh-jcieslak (approved):

PR #2122, opened 2023-10-13
chore: Return multiple errors in existing validations

sfc-gh-jcieslak:
Some of the other stuff done in this pr
- Returned always joined errors for Query structs (didn't applied that to nested structures, because It didn't matter that much - because if there's an error It'll be wrapped anyways by Query struct validation)
- Added common number error messages in errors.go
- Merged some of the validations, making them shorter (e.g. instead of check every combination and returning unique message for that OneOf combination. I've used `exactlyOneValueSet`)
- Replaced plain errors.New() errors with our predefined ones from errors.go (I've left some that had unique or not that much repeating error message that would be worth to make into a separate function in errors.go)
- Adjusted generator's validation files generation

Notes
To be consistent with new validation I've adjusted Query struct (...Options structs) validation functions to include at the top.
```go
if opts == nil {
	return errors.Join(ErrNilOptions)
}
```
and, If there was only one validation after opts nil check I've inlined the join, so instead
```go
var errs []error
if !ValidObjectIdentifier(opts.name) {
	errs = append(errs, ErrInvalidObjectIdentifier)
}
return errors.Join(errs...)
```
I've returned
```go
if !ValidObjectIdentifier(opts.name) {
	return errors.Join(ErrInvalidObjectIdentifier)
}
return nil
```

The advantage of always using joined errors is that we can have one assert function in the unit tests `assertOptsInvalidJoinedErrors` (I can remove the other one now (`assertOptsInvalid`). I didn't, but lmk if we can do it or should we return single errors e.g. like in the above case).

### DIFF
--- a/pkg/sdk/poc/generator/validation.go
+++ b/pkg/sdk/poc/generator/validation.go
@@ -22,8 +22,6 @@ const (
 	ExactlyOneValueSet
 	AtLeastOneValueSet
 	ValidateValue
-	ValidateIntValue
-	ValidateIntRange
 )
 
 type Validation struct {
--- a/pkg/sdk/poc/generator/validation.go
+++ b/pkg/sdk/poc/generator/validation.go
@@ -22,6 +22,8 @@ const (
 	ExactlyOneValueSet
 	AtLeastOneValueSet
 	ValidateValue
+	ValidateIntValue
+	ValidateIntRange
 )
 
 type Validation struct {
@@ -57,13 +59,13 @@ func (v *Validation) Condition(field *Field) string {
 	case ValidIdentifier:
 		return fmt.Sprintf("!ValidObjectIdentifier(%s)", strings.Join(v.fieldsWithPath(field), ","))
 	case ValidIdentifierIfSet:
-		return fmt.Sprintf("valueSet(%s) && !ValidObjectIdentifier(%s)", strings.Join(v.fieldsWithPath(field), ","), strings.Join(v.fieldsWithPath(field), ","))
+		return fmt.Sprintf("%s != nil && !ValidObjectIdentifier(%s)", strings.Join(v.fieldsWithPath(field), ","), strings.Join(v.fieldsWithPath(field), ","))
 	case ConflictingFields:
 		return fmt.Sprintf("everyValueSet(%s)", strings.Join(v.fieldsWithPath(field), ","))
 	case ExactlyOneValueSet:
-		return fmt.Sprintf("ok := exactlyOneValueSet(%s); !ok", strings.Join(v.fieldsWithPath(field), ","))
+		return fmt.Sprintf("!exactlyOneValueSet(%s)", strings.Join(v.fieldsWithPath(field), ","))
 	case AtLeastOneValueSet:
-		return fmt.Sprintf("ok := anyValueSet(%s); !ok", strings.Join(v.fieldsWithPath(field), ","))
+		return fmt.Sprintf("!anyValueSet(%s)", strings.Join(v.fieldsWithPath(field), ","))
 	case ValidateValue:
 		return fmt.Sprintf("err := %s.validate(); err != nil", strings.Join(v.fieldsWithPath(field.Parent), ","))
 	}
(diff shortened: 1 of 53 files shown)


sfc-gh-swinkler (approved):

pkg/sdk/tags_validations.go
```diff
@@ -39,20 +38,20 @@ func (opts *createTagOptions) validate() error {
...
 func (v *TagSet) validate() error {
 	var errs []error
 	if !exactlyOneValueSet(v.MaskingPolicies, v.Comment) {
-		errs = append(errs, errOneOf("MaskingPolicies", "Comment"))
+		errs = append(errs, errOneOf("TagSet", "MaskingPolicies", "Comment"))
 	}
 	if valueSet(v.MaskingPolicies) {
-		if ok := validateIntGreaterThanOrEqual(len(v.MaskingPolicies.MaskingPolicies), 1); !ok {
-			errs = append(errs, fmt.Errorf("number of the MaskingPolicies must be greater than zero"))
+		if !validateIntGreaterThanOrEqual(len(v.MaskingPolicies.MaskingPolicies), 1) {
```
sfc-gh-asawicki, line 53: why not validateIntGreaterThan? (it would be more consistent with the error which is only greater, and not greater or equal)

pkg/sdk/testint/dynamic_table_integration_test.go
```diff
@@ -133,7 +134,7 @@ func TestInt_DynamicTableAlter(t *testing.T) {
 
 		err := client.DynamicTables.Alter(ctx, sdk.NewAlterDynamicTableRequest(dynamicTable.ID()).WithSuspend(sdk.Bool(true)).WithResume(sdk.Bool(true)))
 		require.Error(t, err)
-		expected := "alter statement needs exactly one action from: set, unset, refresh"
+		expected := fmt.Sprintf("exactly one of %s fileds %v must be set", "alterDynamicTableOptions", []string{"Suspend", "Resume", "Refresh", "Set"})
```
sfc-gh-asawicki, line 137: why Sprintf and not ExactlyOneOf here? Because accessibility? Maybe just add a delegation to integration_test_imports.go for now, and we can think later, I would for sure prefer not to have repeated Sprintf like this here.

pkg/sdk/poc/generator/validation.go
```diff
@@ -57,13 +59,13 @@ func (v *Validation) Condition(field *Field) string {
 	case ValidIdentifier:
 		return fmt.Sprintf("!ValidObjectIdentifier(%s)", strings.Join(v.fieldsWithPath(field), ","))
 	case ValidIdentifierIfSet:
-		return fmt.Sprintf("valueSet(%s) && !ValidObjectIdentifier(%s)", strings.Join(v.fieldsWithPath(field), ","), strings.Join(v.fieldsWithPath(field), ","))
+		return fmt.Sprintf("%s != nil && !ValidObjectIdentifier(%s)", strings.Join(v.fieldsWithPath(field), ","), strings.Join(v.fieldsWithPath(field), ","))
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: So, %s != nil should be more correct, because this validation rule should be applied to optional identifiers (...IfSet implies that) and we have them as pointers, that's why %s != nil, you're right that valueSet(%s) should be correct, but it's not right now and we don't know when we'll change it. Plus before this validation wouldn't happen because valueSet underneath calls ValidObjecIdentifier for object identifiers, so then it's really `!ValidObjectIdentifier(%s) && !ValidObjectIdentifier(%s)` which never will be true

pkg/sdk/poc/generator/validation.go
```diff
@@ -77,11 +79,11 @@ func (v *Validation) ReturnedError(field *Field) string {
 	case ValidIdentifierIfSet:
 		return "ErrInvalidObjectIdentifier"
 	case ConflictingFields:
-		return fmt.Sprintf(`errOneOf("%s", %s)`, field.Name, strings.Join(v.paramsQuoted(), ","))
+		return fmt.Sprintf(`errOneOf("%s%s", %s)`, field.Name, field.Path(), strings.Join(v.paramsQuoted(), ","))
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm what I wanted to achieve is when we want to validated nested structs inside e.g. Options struct the struct name passed to the error should be Option struct + path, so e.g. we have this validation
```go
if valueSet(opts.Set) {
 if err := opts.Set.validate(); err != nil {
   errs = append(errs, errOneOf("SomeOpts.Set", err)
 }
}
```

pkg/sdk/poc/generator/validation.go
```diff
@@ -22,6 +22,8 @@ const (
 	ExactlyOneValueSet
 	AtLeastOneValueSet
 	ValidateValue
+	ValidateIntValue
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Oh, I forgot to remove it then, because at first I wanted to add them, but they needed extra work, so I just removed the switch branches but forgot about this value. I'll also check if we have that in the readme

pkg/sdk/tags_validations.go
sfc-gh-jcieslak: Yeah, I didn't see that I have to confirm then which one is correct

sfc-gh-jcieslak: True, I just automatically refactored it as is. Changed.

pkg/sdk/testint/dynamic_table_integration_test.go
sfc-gh-jcieslak: Since this is now in testint, we cannot call errOneOf because it's in the SDK package. We can either export it or leave it as is

sfc-gh-swinkler (dismissed):

pkg/sdk/testint/dynamic_table_integration_test.go
sfc-gh-jcieslak: kk declared in integration_test_imports.go and used in the test

pkg/sdk/poc/generator/validation.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: So, I've made similar function to .Path but with root included and it works as intended now

sfc-gh-asawicki (dismissed):

PR #2126, opened 2023-10-16
feat: Provider config refactor

sfc-gh-swinkler:
Adds all missing attributes as per config block of gosnowflake package. https://pkg.go.dev/github.com/snowflakedb/gosnowflake#Config

Renames any attributes which didn't follow convention. Cleaned up oauth logic. Removed pointless tests.


pkg/provider/provider.go
```diff
@@ -52,13 +46,247 @@ func Provider() *schema.Provider {
...
+			},
+			"protocol": {
+				Type:        schema.TypeString,
+				Description: "Either http or https. Can be sourced from `SNOWFLAKE_PROTOCOL` environment variable.",
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_PROTOCOL", "https"),
+			},
+			"host": {
+				Type:        schema.TypeString,
+				Description: "Supports passing in a custom host value to the snowflake go driver for use with privatelink.",
```
sfc-gh-jcieslak, line 98: Can you add in every description that it can be retrieved from other source e.g. here env (sometimes it's described and like in this case, it's not). From user's perspective it's pretty important to know how to setup the client without reading the code to know what's going on

pkg/provider/provider.go
```diff
@@ -52,13 +46,247 @@ func Provider() *schema.Provider {
...
+				Optional:    true,
+			},
+			"external_browser_timeout": {
+				Type:        schema.TypeInt,
+				Description: "The timeout in seconds for the external browser to complete the authentication. Default is 120 seconds.",
+				Optional:    true,
+			},
+			"insecure_mode": {
+				Type:        schema.TypeBool,
+				Description: "If true, bypass the Online Certificate Status Protocol (OCSP) certificate revocation check. IMPORTANT: Change the default value for testing or emergency situations only.",
```
sfc-gh-jcieslak, line 171: Add env variable to the desc

sfc-gh-jcieslak: Add in others if missing

pkg/provider/provider.go
```diff
@@ -52,13 +46,247 @@ func Provider() *schema.Provider {
...
+			"params": {
+				Type:        schema.TypeMap,
+				Description: "Sets other connection (i.e. session) parameters. [Parameters](https://docs.snowflake.com/en/sql-reference/parameters)",
+				Optional:    true,
+			},
+			"session_params": {
+				Type:        schema.TypeMap,
+				Description: "Sets session parameters. [Parameters](https://docs.snowflake.com/en/sql-reference/parameters)",
+				Optional:    true,
+				Deprecated:  "Use `params` instead",
```
sfc-gh-jcieslak, line 83: It's deprecated. Shouldn't it be put with other deprecated parameters at the bottom ?

pkg/provider/provider.go
```diff
@@ -52,13 +46,247 @@ func Provider() *schema.Provider {
...
+				Description: "Specifies the role to use by default for accessing Snowflake objects in the client session. Can be sourced from the `SNOWFLAKE_ROLE` environment variable.",
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_ROLE", nil),
+			},
+			"region": {
+				Type:        schema.TypeString,
+				Description: "Snowflake region, such as \"eu-central-1\", with this parameter. However, since this parameter is deprecated, it is best to specify the region as part of the account parameter. For details, see the description of the account parameter. [Snowflake region](https://docs.snowflake.com/en/user-guide/intro-regions.html) to use.  Required if using the [legacy format for the `account` identifier](https://docs.snowflake.com/en/user-guide/admin-account-identifier.html#format-2-legacy-account-locator-in-a-region) in the form of `<cloud_region_id>.<cloud>`. Can be sourced from the `SNOWFLAKE_REGION` environment variable.",
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_REGION", "us-west-2"),
+				Optional:    true,
+				Deprecated:  "Specify the region as part of the account parameter",
```
sfc-gh-jcieslak, line 67: It's deprecated. Shouldn't it be put with other deprecated parameters at the bottom ?

pkg/provider/provider.go
```diff
@@ -525,218 +593,136 @@ func DSN(
...
 
-	return gosnowflake.DSN(config)
-}
-
-func ReadPrivateKeyFile(privateKeyPath string) ([]byte, error) {
-	expandedPrivateKeyPath, err := homedir.Expand(privateKeyPath)
-	if err != nil {
-		return nil, fmt.Errorf("invalid Path to private key err = %w", err)
+	if v, ok := s.GetOk("login_timeout"); ok {
+		config.LoginTimeout = time.Duration(int64(v.(int)))
```
sfc-gh-jcieslak, line 616: I think every conversion to time.Duration is incorrect. time.Duration as a single value represents nano second. In the documentation you've mentioned you expect users to pass seconds. Then it should be assigned like this
```go
config.LoginTimeout = time.Second * time.Duration(v.(int))
```

pkg/provider/provider_helpers.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/mitchellh/go-homedir"
+	"github.com/youmark/pkcs8"
+	"golang.org/x/crypto/ssh"
+)
+
+func mergeSchemas(schemaCollections ...map[string]*schema.Resource) map[string]*schema.Resource {
+	out := map[string]*schema.Resource{}
+	for _, schemaCollection := range schemaCollections {
+		for name, s := range schemaCollection {
+			out[name] = s
```
sfc-gh-jcieslak, line 26: Shouldn't we check for any unexpected overwriting here ?

sfc-gh-jcieslak (changes requested):

pkg/provider/provider_helpers.go
sfc-gh-swinkler: we have had this code in the provider for years now, and it has not caused a problem, so i don't think we need to change it. i was contemplating removing this entirely, as its only purpose is to segregating grant resource from other resources.  Once we are using the provider v2, then this can be removed.

docs/index.md
```diff
@@ -51,31 +51,61 @@ provider "snowflake" {
 
 ### Optional
 
-- `account` (String) The name of the Snowflake account. Can also come from the `SNOWFLAKE_ACCOUNT` environment variable. Required unless using profile.
-- `authenticator` (String) Specifies the [authentication type](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#AuthType) to use when connecting to Snowflake. Valid values include: Snowflake, OAuth, ExternalBrowser, Okta, JWT, TokenAccessor, UsernamePasswordMFA
+- `account` (String) Specifies your Snowflake account identifier assigned, by Snowflake. For information about account identifiers, see the (Snowflake documentation)[https://docs.snowflake.com/en/user-guide/admin-account-identifier.html]. Can also be sourced from the `SNOWFLAKE_ACCOUNT` environment variable.
```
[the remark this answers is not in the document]
sfc-gh-swinkler: fixed

pkg/provider/provider.go
```diff
@@ -1,48 +1,42 @@
...
-	"golang.org/x/crypto/ssh"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/datasources"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/db"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 )
 
-// Provider is a provider.
+// Provider returns a Terraform Provider using configuration from https://github.com/snowflakedb/gosnowflake/blob/master/dsn.go#L43
```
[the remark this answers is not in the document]
sfc-gh-swinkler: fixed

pkg/provider/provider.go
```diff
@@ -1,48 +1,42 @@
...
 func Provider() *schema.Provider {
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
 			"account": {
 				Type:        schema.TypeString,
-				Description: "The name of the Snowflake account. Can also come from the `SNOWFLAKE_ACCOUNT` environment variable. Required unless using profile.",
+				Description: "Specifies your Snowflake account identifier assigned, by Snowflake. For information about account identifiers, see the (Snowflake documentation)[https://docs.snowflake.com/en/user-guide/admin-account-identifier.html]. Can also be sourced from the `SNOWFLAKE_ACCOUNT` environment variable.",
```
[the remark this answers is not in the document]
sfc-gh-swinkler: fixed

pkg/provider/provider.go
sfc-gh-swinkler: done

pkg/provider/provider.go
sfc-gh-swinkler: done

pkg/provider/provider.go
sfc-gh-swinkler: done

pkg/provider/provider.go
sfc-gh-swinkler: done

pkg/provider/provider.go
sfc-gh-swinkler: fixed

sfc-gh-jcieslak (dismissed):

pkg/provider/provider_helpers.go
```diff
@@ -0,0 +1,137 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mitchellh/go-homedir"
+	"github.com/youmark/pkcs8"
+	"golang.org/x/crypto/ssh"
+)
+
+func mergeSchemas(schemaCollections ...map[string]*schema.Resource) map[string]*schema.Resource {
```
sfc-gh-asawicki, line 22: It would be nice to unit test the methods from this file (I know that there are other helpers throughout the project, which also need testing, but I would like to move towards testing such things). It does not have to be done in this PR, it does not necessarily have to be the next one - it can be added as a part of SNOW-933289 (please create a subtask there if you choose to postpone it).

pkg/provider/provider.go
```diff
@@ -1,221 +1,393 @@
...
+			},
+			"role": {
+				Type:        schema.TypeString,
+				Description: "Specifies the role to use by default for accessing Snowflake objects in the client session. Can also be sourced from the `SNOWFLAKE_ROLE` environment variable. .",
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_ROLE", nil),
+			},
+			"validate_default_parameters": {
+				Type:        schema.TypeBool,
+				Description: "ValidateDefaultParameters disable the validation checks for Database, Schema, Warehouse and Role at the time a connection is established. Can also be sourced from the `SNOWFLAKE_VALIDATE_DEFAULT_PARAMETERS` environment variable.",
```
sfc-gh-asawicki, line 65: param is `validate_default_parameters` and in the description it's camelCased `ValidateDefaultParameters`, is there a reason for it?

pkg/provider/provider.go
```diff
@@ -1,221 +1,393 @@
...
+				Description: "IP address for network checks. Can also be sourced from the `SNOWFLAKE_CLIENT_IP` environment variable.",
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_CLIENT_IP", nil),
+			},
+			"protocol": {
+				Type:        schema.TypeString,
+				Description: "Either http or https, defaults to https. Can also be sourced from the `SNOWFLAKE_PROTOCOL` environment variable.",
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_PROTOCOL", nil),
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
```
sfc-gh-asawicki, line 85: it would be nice to have this function extracted and unit-tested

pkg/provider/provider.go
```diff
@@ -1,221 +1,393 @@
...
+				Description: "Support custom port values to snowflake go driver for use with privatelink. Can also be sourced from the `SNOWFLAKE_PORT` environment variable. ",
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_PORT", nil),
+			},
+			"authenticator": {
+				Type:        schema.TypeString,
+				Description: "Specifies the [authentication type](https://pkg.go.dev/github.com/snowflakedb/gosnowflake#AuthType) to use when connecting to Snowflake. Valid values include: Snowflake, OAuth, ExternalBrowser, Okta, JWT, TokenAccessor, UsernamePasswordMFA. Can also be sourced from the `SNOWFLAKE_AUTHENTICATOR` environment variable.",
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_AUTHENTICATOR", nil),
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
```
sfc-gh-asawicki, line 112: the same here, I think all such functions should be extracted and not defined in place, I also think that this may be a topic for a followup task, which could be added as a subtask of SNOW-933289

pkg/provider/provider.go
```diff
@@ -1,221 +1,393 @@
...
+			"passcode": {
+				Type:          schema.TypeString,
+				Description:   "Specifies the passcode provided by Duo when using multi-factor authentication (MFA) for login. Can also be sourced from the `SNOWFLAKE_PASSCODE` environment variable. ",
+				Optional:      true,
+				ConflictsWith: []string{"passcode_in_password"},
+				DefaultFunc:   schema.EnvDefaultFunc("SNOWFLAKE_PASSCODE", nil),
+			},
+			"passcode_in_password": {
+				Type:          schema.TypeBool,
+				Description:   "False by default. Set to true if the MFA passcode is embedded in the login password. Appends the MFA passcode to the end of the password. Can also be sourced from the `SNOWFLAKE_PASSCODE_IN_PASSWORD` environment variable. ",
```
sfc-gh-asawicki, line 131: description states that this is optional by default, but the default is taken from env variable

pkg/provider/provider.go
```diff
@@ -1,221 +1,393 @@
...
+					Description: "Should retried request contain retry reason",
+					Optional:    true,
+				},*/
+			"profile": {
+				Type:        schema.TypeString,
+				Description: "Sets the profile to read from ~/.snowflake/config file. Can also be sourced from the `SNOWFLAKE_PROFILE` environment variable.",
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_PROFILE", "default"),
+			},
+			// Deprecated attributes
```
sfc-gh-asawicki, line 301: do we add any note somewhere else that these ones are deprecated?

Do we add it in the description of each of these parameters? ( I don't like single comment stating about deprecations, someone may add other param after this comment, and it will be misleading - that's why I would prefer having it added to every one of them)

pkg/provider/provider.go
```diff
@@ -525,218 +626,136 @@ func DSN(
...
-		return nil, err
+
+	/*
+		Feature not yet released as of latest gosnowflake release
+		https://github.com/snowflakedb/gosnowflake/blob/master/dsn.go#L103
+		if v, ok := s.GetOk("include_retry_reason"); ok && v.(bool) {
+			config.IncludeRetryParameters = v.(bool)
+		}
+	*/
+	if v, ok := s.GetOk("profile"); ok && v.(string) != "" {
```
sfc-gh-asawicki, line 736: such logic should be tested, is it tested anywhere?

sfc-gh-asawicki (changes requested):

pkg/provider/provider.go
sfc-gh-swinkler: only if it isn't nil. And the default value is from the config, not from us setting the default intentionally. I just searched through the gosnowflake package to find out what the defaults are.

pkg/provider/provider.go
sfc-gh-swinkler: its enough to have the "deprecated" message, as it will show up in the documentation, and already "marks" resources as deprecated. The comment is not super important, its just for code organization I suppose.

> [tag](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/table#tag) (Block List, Deprecated) Definitions of a tag to associate with the resource. (see [below for nested schema](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/table#nestedblock--tag))

pkg/provider/provider.go
sfc-gh-swinkler: i suppose we could, firstly there aren't too many validators, and secondly, the framework does validators a bit differently. so even if we extract this code and put it somewhere else, it would eventually become obsolete.

pkg/provider/provider.go
sfc-gh-swinkler: one thing at a time...

pkg/provider/provider.go
sfc-gh-swinkler: copied from the gosnowflake.Config docs. Changed

pkg/provider/provider_helpers.go
sfc-gh-swinkler: i would sooner refactor the code to not use this function at all than to write a test for it, because this function is not very useful anyways. but I get what you are saying

sfc-gh-asawicki (approved):

PR #2129, opened 2023-10-18
feat: Use streams from the new SDK in resource / datasource

sfc-gh-jcieslak:
Usage of the new SDK Streams inside existing resource + datasource.
Things to point out
- Mind the TODO comment in stream resource Read function
- Helper function for decoding . (dot) delimited identifiers from config
- I've changed the doc to explicitly specify that OnXXX attributes are identifiers, not names. In the old implementation it wasn't allowed to put only name and I've kept the same behaviour, but I wondered what if we allowed that and if only name is specified or only "schema|name" was specified and the rest was read from the resource config ? For example
```tf
resource "snowflake...stream" "mystream" {
  "name": "abc",
  "database": "db",
  "schema": "sch",
  "on_table": "xyz", // only valid option -> "on_table": "db.sch.xyz"
}
```
Right now, this wouldn't work, because on_table is not fully qualified identifier. We could assume that the user meant to create stream db.sch.abc onto a table, which fully qualified identifier would look like: db.sch.xyz ?
We could achieve it by providing following function
```go
func constructWithFallbacks(databaseName string, schemaName string, identifier sdk.ObjectIdentifier) sdk.SchemaObjectIdentifier {
	switch id := identifier.(type) {
	case sdk.AccountObjectIdentifier:
		return sdk.NewSchemaObjectIdentifier(databaseName, schemaName, id.Name())
	case sdk.DatabaseObjectIdentifier:
		return sdk.NewSchemaObjectIdentifier(id.DatabaseName(), schemaName, id.Name())
	case sdk.SchemaObjectIdentifier:
		return id
	default:
		return sdk.NewSchemaObjectIdentifier(databaseName, schemaName, id.Name())
	}
}
```
where identifier can be either AccountObjectIdentifier, DatabaseObjectIdentifier or SchemaObjectIdentifier (let's say we don't care about the rest of the identifiers for now) and If any of the passed identifiers is not as qualified as SchemaObjectIdentifiers we'll use fallback (or default) values from parameters (in this case resource level databaseName and schemaName).


sfc-gh-swinkler:
> func constructWithFallbacks(databaseName string, schemaName string, identifier sdk.ObjectIdentifier) s

This is a weird function, but I guess its a weird behavior that would should eventually fix. Maybe calling this WithDefault() would be better.

pkg/helpers/helpers.go
```diff
@@ -100,6 +105,25 @@ func DecodeSnowflakeID(id string) sdk.ObjectIdentifier {
 	}
 }
 
+func DecodeSnowflakeParameterID(id string) sdk.ObjectIdentifier {
+	parts := strings.Split(id, ParameterIDDelimiter)
```
sfc-gh-swinkler, line 109: just FYI this could be problematic if you have "." in your name. I.e.

> "test.db"."myschema"."mystream" 


This is a perfectly valid identifier, but when split would result in 4 parts. The correct thing to do is to split on double quotes, if there are double quotes.

pkg/resources/stream_test.go
```diff
@@ -1,238 +1,239 @@
...
-	r := require.New(t)
-
-	in := map[string]interface{}{
+	d := stream(t, "database_name|schema_name|stream_name", map[string]any{
 		"name":              "stream_name",
 		"database":          "database_name",
 		"schema":            "schema_name",
 		"comment":           "great comment",
-		"on_table":          "target_db.target_schema.target_table",
+		"on_table":          "\"target_db\".\"target_schema\".\"target_table\"",
```
sfc-gh-swinkler, line 36: would be good if we could support both ways. With and without double quotes. Some people don't like using double quotes.

pkg/resources/stream_test.go
```diff
@@ -1,238 +1,239 @@
 package resources_test
```
sfc-gh-swinkler, line 1: this whole file could be deleted, as this sort of test is not useful

sfc-gh-swinkler (changes requested):

pkg/resources/stream_test.go
sfc-gh-jcieslak: I just saw in the other pr and somewhere else that we're using the approach with \". I agree we should support both (but imo a way with \" is so bad). I'll adjust the decoder, so It'll parse both

pkg/helpers/helpers.go
sfc-gh-asawicki: Yeah, but it's also not that easy, because we should enforce that if the user wants to use quotes, they should use them for each part of the id (e.g. "db".schema."table" should not be accepted).

This is a much broader topic than just fixing it in this PR. Let's not aim to solve it perfectly here and instead let's dedicate a separate task to discuss and handle it correctly.

sfc-gh-jcieslak: I'll fix it and write tests for it, so it will be clear that we wouldn't get any unintentional behaviour

sfc-gh-asawicki: And please add a parametrized test for such a method.

pkg/resources/stream_test.go
sfc-gh-asawicki: Yeah, it is not useful in such a state and we were removing those for other resources, because these tests should be unaware of SQL statements generated. It should be aware of the interaction with SDK though. Let's discuss these kinds of tests in our upcoming test stabilization efforts, because this topic is much wider than the PR discussion.

pkg/resources/stream_test.go
sfc-gh-asawicki: Yes, and this is a bit of breaking change if we stop to support these now, right?

pkg/resources/stream.go
```diff
@@ -109,295 +105,207 @@ func Stream() *schema.Resource {
...
-		DatabaseName: lines[0][0],
-		SchemaName:   lines[0][1],
-		StreamName:   lines[0][2],
-	}
-	return streamResult, nil
-}
-
-// streamOnObjectIDFromString() takes in a dot-delimited string: DatabaseName.SchemaName.TableName
-// and returns a streamOnObjectID object.
-func streamOnObjectIDFromString(stringID string) (*streamOnObjectID, error) {
```
sfc-gh-asawicki, line 165: Shouldn't we keep this method to not introduce breaking (or non-breaking) changes with this change about using the SDK and not changing the way we expect the user to type in identifiers in the config?

sfc-gh-asawicki (changes requested):

pkg/helpers/helpers_test.go
```diff
@@ -0,0 +1,92 @@
+package helpers
+
+import (
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDecodeSnowflakeParameterID(t *testing.T) {
```
sfc-gh-swinkler, line 10: FYI this would be a good place to use table-driven tests. https://dave.cheney.net/2019/05/07/prefer-table-driven-tests. But not that important

sfc-gh-asawicki: yeah, I support that, we already have them in few other places

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

sfc-gh-swinkler (approved):

PR #2130, opened 2023-10-18
fix: provider muxing for plugin-framework

sfc-gh-swinkler:
Adds muxing to the provider, to allow it to be able to work with new provider framework


internal/provider/provider.go
```diff
@@ -0,0 +1,818 @@
...
+							Sensitive:   true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func (p *SnowflakeProvider) Configure(ctx context.Context, req provider.ConfigureRequest, resp *provider.ConfigureResponse) {
```
sfc-gh-jcieslak, line 396: Parameter setting looks like it could be simplified by creating common functions for setting (at least) simple values. e.g.

```go
// instead of
config := &gosnowflake.Config{
	Application: "terraform-provider-snowflake",
}
account := os.Getenv("SNOWFLAKE_ACCOUNT")
if data.Account.ValueString() != "" {
	account = data.Account.ValueString()
}
if account != "" {
	config.Account = account
}
// we could something like this
config := &gosnowflake.Config{
	Application: "terraform-provider-snowflake",
	Account:     GetStringValue(data.Account, "SNOWFLAKE_ACCOUNT"),
	// Other params...
}

type StringValuer interface {
	ValueString() string
}

func GetStringValue(valuer StringValuer, envVariable string) string {
	if value := valuer.ValueString(); value != "" {
		return value
	}
	return os.Getenv(envVariable)
}
```
I know there can be different variations for the same type (e.g. with secondary env variable, with default value, etc.), but It's hard to clearly see what is going on and to check if every if statement is correct.

pkg/provider/provider.go
```diff
@@ -372,13 +372,13 @@ func Provider() *schema.Provider {
...
-				Optional:      true,
-				Sensitive:     false,
-				DefaultFunc:   schema.EnvDefaultFunc("SNOWFLAKE_USE_BROWSER_AUTH", nil),
-				Deprecated:    "Use `authenticator` instead",
-				ConflictsWith: []string{"password", "private_key_path", "private_key", "private_key_passphrase", "oauth_access_token", "oauth_refresh_token"},
+				Type:        schema.TypeBool,
+				Description: "Required when `oauth_refresh_token` is used. Can also be sourced from `SNOWFLAKE_USE_BROWSER_AUTH` environment variable.",
+				Optional:    true,
+				Sensitive:   false,
+				// DefaultFunc:   schema.EnvDefaultFunc("SNOWFLAKE_USE_BROWSER_AUTH", nil),
```
sfc-gh-jcieslak, line 379: remove if not used ?

internal/provider/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+
+const (
+	CreateOperation tfOperation = "CREATE"
+	ReadOperation   tfOperation = "READ"
+	UpdateOperation tfOperation = "UPDATE"
+	DeleteOperation tfOperation = "DELETE"
+)
+
+func formatSQLPreview(operation tfOperation, resourceName string, id string, commands []string) string {
+	var c color.Color
```
sfc-gh-asawicki, line 23: Instead of doing a switch every time we could define a map to just get the color from it. This is just nit-picking, it can stay like this.

internal/provider/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	return sb.String()
+}
+
+type sensitiveAttributes struct {
+	m map[string]bool
+}
+
+var (
+	sa   *sensitiveAttributes
+	lock = sync.Mutex{}
```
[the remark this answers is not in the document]
sfc-gh-asawicki: +1 for init function and logic simplification

internal/provider/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func isSensitive(s string) bool {
```
sfc-gh-asawicki, line 52: 1. Do we have any tests for this method?
2. It won't find such a file on the CI right now, we should add it or at least have a task to remember about it.

internal/provider/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+				return false
+			}
+			// sensitive path is ~/.snowflake/sensitive.
+			f := filepath.Join(dir, ".snowflake", "sensitive")
+			dat, err := os.ReadFile(f)
+			if err != nil {
+				return false
+			}
+			lines := strings.Split(string(dat), "\n")
+			r := regexp.MustCompile("(data[.])?snowflake_(.*)[.](.+)[.](.+)")
```
sfc-gh-asawicki, line 71: is this regexp coming from some terraform doc? Let's test it and document it.

internal/provider/resource_monitor_resource.go
```diff
@@ -0,0 +1,807 @@
...
+	SetForAccount            types.Bool    `tfsdk:"set_for_account"`
+	Warehouses               types.Set     `tfsdk:"warehouses"`
+	Id                       types.String  `tfsdk:"id"`
+}
+
+func resourceMonitorSchemaV0() schema.Schema {
+	return schema.Schema{
+		Version: 0,
+		Attributes: map[string]schema.Attribute{
+			"name": schema.StringAttribute{
```
sfc-gh-asawicki, line 61: is there any way to use `resourceMonitorModelV0` to not write everything here (or use this to generate the above struct)?

What I mean is that in the model above I have to say that the `Name` will be a `string` and that it uses the `"name"` key. Here I have to specify it again (`"name"`, `StringAttribute`).

It seems super-clunky. Aren't there any tools to make it less repetitive?

internal/provider/resource_monitor_resource.go
```diff
@@ -0,0 +1,807 @@
...
+}
+
+func upgradeResourceMonitorStateV0toV1(ctx context.Context, req resource.UpgradeStateRequest, resp *resource.UpgradeStateResponse) {
+	var resourceMonitorDataV0 resourceMonitorModelV0
+	resp.Diagnostics.Append(req.State.Get(ctx, &resourceMonitorDataV0)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	name := resourceMonitorDataV0.Name
```
sfc-gh-asawicki, line 157: We could work a bit on making this function a bit more readable. This is just a nit-picking but suggestions I have are:
- do not define all these variables upfront if they are needed in if statements much later in this function
- scope these assignments inside if condition checks
- I would like to see a builder pattern here for `resourceMonitorModelV1`; maybe not even the basic one, but the one where e.g. WithSuspendTriggers takes resourceMonitorDataV0.SuspendTriggers as an input param, ad does all the logic from if there, and finally sets the right value on resourceMonitorModelV1 struct

internal/provider/resource_monitor_resource.go
```diff
@@ -0,0 +1,807 @@
...
+		Frequency:      frequency,
+		StartTimestamp: startTimestamp,
+		EndTimestamp:   endTimestamp,
+		Triggers:       triggersSet,
+		Id:             resourceMonitorDataV0.Id,
+	}
+	resp.Diagnostics.Append(resp.State.Set(ctx, resourceMonitorV1)...)
+}
+
+type resourceMonitorModelV1 struct {
```
sfc-gh-asawicki, line 229: Later it would be nice to split into different version-prefixed files. (nit-picking)

internal/provider/resource_monitor_resource.go
```diff
@@ -0,0 +1,807 @@
...
+		}
+		with.Triggers = triggers
+	}
+
+	if setWith {
+		opts.With = with
+	}
+	err := client.ResourceMonitors.Create(ctx, id, opts)
+
+	if dryRun {
```
sfc-gh-asawicki, line 558: I like the dry run

internal/provider/resource_monitor_resource.go
```diff
@@ -0,0 +1,807 @@
...
+	} else {
+		// UPDATE
+		_, updateLogs, _ := r.update(ctx, plan, state, true)
+		_, readLogs, _ := r.read(ctx, plan, true)
+		updateLogs = append(updateLogs, readLogs...)
+		tflog.Debug(ctx, formatSQLPreview(UpdateOperation, resourceName, state.Id.ValueString(), updateLogs))
+	}
+}
+
+func (r *ResourceMonitorResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
```
sfc-gh-asawicki, line 477: I am not commenting much below, because it seems like logic similar to the one in the existing provider, I see that there are some additions and shortcuts but we will probably talk about them on Monday.

main.go
```diff
@@ -1,21 +1,64 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6/tf6server"
+	"github.com/hashicorp/terraform-plugin-mux/tf5to6server"
+	"github.com/hashicorp/terraform-plugin-mux/tf6muxserver"
 )
 
-const ProviderAddr = "registry.terraform.io/Snowflake-Labs/snowflake"
+var version string = "dev" // goreleaser can pass other information to the main package, such as the specific commit
+// https://goreleaser.com/cookbooks/using-main.version/
 
 func main() {
```
sfc-gh-asawicki, line 21: Let's test this even manually, so that we are sure, that we are not introducing any incompatibility with the setup change (because as we discussed some things are slipping through our tests)

pkg/sdk/client.go
```diff
@@ -186,19 +205,34 @@
 
 // Exec executes a query that does not return rows.
 func (c *Client) exec(ctx context.Context, sql string) (sql.Result, error) {
+	if c.dryRun {
+		c.traceLogs = append(c.traceLogs, sql)
+		log.Printf("[DEBUG] sql-conn-exec-dry: %v\n", sql)
```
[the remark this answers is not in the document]
sfc-gh-asawicki: 😮 
Should we care about this comment? Maybe we are really exposing some sensitive info in this log?

pkg/sdk/client.go
```diff
@@ -18,6 +18,8 @@ type Client struct {
 	db             *sqlx.DB
 	sessionID      string
 	accountLocator string
+	dryRun         bool
```
sfc-gh-asawicki, line 21: I like adding this dry run here, but it's not ideal. Now we will not get to the logging have the SDK validation errors (but maybe that's also fine). Let's come back to this later, it is for sure fine for now.

internal/provider/provider.go
sfc-gh-asawicki: ...or a config builder

internal/provider/provider.go
```diff
@@ -0,0 +1,818 @@
+package provider
```
sfc-gh-asawicki, line 1: I believe that this file mimics the current provider config.

Is something worth mentioning here which is an additional thing? (we can talk about it on Monday).

internal/provider/provider_helpers.go
```diff
@@ -0,0 +1,186 @@
...
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/mitchellh/go-homedir"
+	"github.com/snowflakedb/gosnowflake"
+	"github.com/youmark/pkcs8"
+	"golang.org/x/crypto/ssh"
+)
+
+func mergeSchemas(schemaCollections ...map[string]*schema.Resource) map[string]*schema.Resource {
```
sfc-gh-asawicki, line 23: with new provider I would like to have unit tests for all such helper methods. Let's just keep it in mind, this should not be a part of this PR but let's not forget about this.

sfc-gh-asawicki (changes requested):

internal/planmodifiers/stringplanmodifier.go
```diff
@@ -0,0 +1,35 @@
+package stringplanmodifiers
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+// useStateForUnknownModifier implements the plan modifier.
```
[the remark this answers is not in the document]
sfc-gh-swinkler: yeah, but its unused currently

internal/provider/helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	var sb strings.Builder
+	sb.WriteString(c.Sprintf("\n[ %s %s %s ]", operation, resourceName, id))
+	for _, command := range commands {
+		sb.WriteString(c.Sprintf("\n  - %s", command))
+	}
+	sb.WriteString("\n")
+	return sb.String()
+}
+
+type sensitiveAttributes struct {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: this is something we should discuss in the Monday meeting. its just a proposal for now

internal/provider/helpers.go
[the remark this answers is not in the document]
sfc-gh-swinkler: i was just trying to get a demonstration of how this might work, I don't even know if we should be doing this or not. we should discuss with the tf provider framework team when we have a meeting

internal/provider/helpers.go
sfc-gh-swinkler: no we don't have any tests for it. again, this is more of a proposal

internal/provider/helpers.go
sfc-gh-swinkler: this is a resource address. Supposed to be you have a `~/.snowflake/sensitive` file containing contents of:

```
snowflake_resource_monitor.*.credit_quota
snowflake_resource_monitor.*.frequency
```
And this would match the resource addresses included there

internal/provider/resource_monitor_resource.go
sfc-gh-swinkler: Not from the documentation I have read. This was what I was saying, when I think there is stuff we could generate.

pkg/sdk/client.go
sfc-gh-swinkler: i agree, i think there could be a better way of doing this. But this is a simple approach. We also don't enable dry run except in this one resource monitor file which isn't even used by the provider, so it is okay.

internal/provider/provider.go
sfc-gh-swinkler: The Configure() function is required as part of the interface of being a provider. You could introduce other design patterns such as builder as part of that, but you still need this function.

And you are right, it does match the current provider config. For good reason, it has to for the provider muxing to work. If they didn't match exactly then we would not be able to write resources using the plugin framework.

internal/provider/provider_helpers.go
sfc-gh-swinkler: agreed

internal/provider/provider.go
```diff
@@ -0,0 +1,818 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/provider/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/snowflakedb/gosnowflake"
+)
+
+// Ensure SnowflakeProvider satisfies various provider interfaces.
+var _ provider.Provider = &SnowflakeProvider{}
```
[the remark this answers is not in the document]
sfc-gh-swinkler: okay

internal/provider/provider.go
```diff
@@ -0,0 +1,818 @@
...
+					stringvalidator.AlsoRequires(path.MatchRoot("oauth_refresh_token"), path.MatchRoot("oauth_client_id"), path.MatchRoot("oauth_client_secret"), path.MatchRoot("oauth_endpoint")),
+				},
+				DeprecationMessage: "Use `token_accessor.0.redirect_uri` instead",
+			},
+			"browser_auth": schema.BoolAttribute{
+				Description:        "Required when `oauth_refresh_token` is used. Can also be sourced from `SNOWFLAKE_USE_BROWSER_AUTH` environment variable.",
+				Optional:           true,
+				Sensitive:          false,
+				DeprecationMessage: "Use `authenticator` instead",
+				/*Validators: []validator.Bool{
```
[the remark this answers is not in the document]
sfc-gh-swinkler: done

internal/provider/provider.go
sfc-gh-swinkler: a good idea, i also wasn't happy with the current design, but I was just trying to keep it as vanilla as possible. the current implementation matches how the tf provider docs suggest.

sfc-gh-jcieslak (approved):

pkg/sdk/client.go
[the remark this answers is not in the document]
sfc-gh-swinkler: No, it just thinks that because these are prefixed with "password".

main.go
sfc-gh-asawicki: Closing after verbal discussion

sfc-gh-asawicki (dismissed):

PR #2140, opened 2023-10-19
feat: Use task from SDK in resource and data source

sfc-gh-asawicki:
Used tasks from the SDK in resource and data source. This does not conclude the work on this topic; the following PR is needed (read below).

## Changes
* Used task from the SDK in resource and data source
  * Old implementation removed
  * Retries will be extracted in the following PR (TODO left)
  * The missing `SUSPEND_TASK_AFTER_NUM_FAILURES` attribute will be added in the following PR (TODO left)
  * Old implementation handles session parameters change badly; will be added in the following PR (TODO left)
* Added missing session parameters to unset
* Added two methods to produce SessionParameters from plain string maps
  * They will be used in other places in the following PR (TODO left)
  * They will be unit tested in the following PR (TODO left)
* Added missing capabilities of SDK (all together with definition, validations, and builders changes)
  * Alter set user_task_managed_initial_warehouse_size
  * Alter set and unset error_integration
* Introduced TaskState
* Changed type of task predecessors to list of schema object identifiers
  * Used old mapping implementation
  * Removed probably deprecated part of this copied implementation
  * Added unit test to this copied implementation
* Introduced a new method to get root tasks for task
  * Implementation copied from old code
  * Added unit tests to this copied implementation
  * Old implementation is not protected against cycles (2 TODOs left)
* Added integration tests for DAGs


pkg/resources/task.go
```diff
@@ -344,156 +258,180 @@ func ReadTask(d *schema.ResourceData, meta interface{}) error {
...
+		size, err := sdk.ToWarehouseSize(v.(string))
+		if err != nil {
+			return err
+		}
+		createRequest.WithWarehouse(sdk.NewCreateTaskWarehouseRequest().WithUserTaskManagedInitialWarehouseSize(&size))
 	}
 
 	if v, ok := d.GetOk("schedule"); ok {
-		builder.WithSchedule(v.(string))
+		createRequest.WithSchedule(sdk.Pointer(v.(string)))
```
sfc-gh-swinkler, line 289: this could be sdk.String()

pkg/resources/task.go
```diff
@@ -191,110 +142,73 @@ func Task() *schema.Resource {
...
-	}
-
-	// The "DESCRIBE TASK ..." command returns the string "null" for error_integration
-	if t.ErrorIntegration.String == "null" {
-		t.ErrorIntegration.Valid = false
-		t.ErrorIntegration.String = ""
-	}
-
-	if err := d.Set("error_integration", t.ErrorIntegration.String); err != nil {
+	if err := d.Set("error_integration", task.ErrorIntegration); err != nil {
```
sfc-gh-swinkler, line 190: is this part of the describe or the show? there was a comment previously that this was part of the describe command. Are we combining the two together?

pkg/resources/task.go
```diff
@@ -191,110 +142,73 @@ func Task() *schema.Resource {
...
-	paramRows, err := snowflake.Query(db, q)
-	if err != nil {
-		return err
-	}
-	params, err := snowflake.ScanTaskParameters(paramRows)
+	opts := &sdk.ShowParametersOptions{In: &sdk.ParametersIn{Task: taskId}}
+	params, err := client.Parameters.ShowParameters(ctx, opts)
 	if err != nil {
 		return err
 	}
```
sfc-gh-swinkler, line 214: log.Printf("[TRACE] %+v\n", param)
maybe lets not print the param

pkg/resources/task.go
```diff
@@ -191,110 +142,73 @@ func Task() *schema.Resource {
...
 		log.Printf("[DEBUG] task (%s) not found", d.Id())
 		d.SetId("")
 		return nil
 	}
-	if err != nil {
-		return err
-	}
 
-	if err := d.Set("enabled", t.IsEnabled()); err != nil {
+	if err := d.Set("enabled", task.IsStarted()); err != nil {
```
sfc-gh-swinkler, line 158: I can't find this in Snowflake documentation so I'm not sure. Should this be IsEnabled()? Or was the previous attribute poorly named?

sfc-gh-asawicki: There are two states of the task: "started" and "suspended". Actions are RESUME and SUSPEND.

I think that IsEnabled is a poor name for it, that's why I have renamed it.

pkg/resources/task.go
sfc-gh-asawicki: I have removed it because this logic is already in the SDK (mapping task row): 
```
	if r.ErrorIntegration.Valid && r.ErrorIntegration.String != "null" {
		task.ErrorIntegration = r.ErrorIntegration.String
	}
```

pkg/resources/task.go
sfc-gh-asawicki: Okay, that was part of the previous implementation - removed.

pkg/resources/task.go
sfc-gh-asawicki: Used `sdk.String`, `sdk.Int` and `sdk.Bool` instead of `sdk.Pointer` throughout this file.

pkg/sdk/tasks_impl_gen.go
```diff
@@ -218,6 +274,29 @@ func (r taskDBRow) convert() *Task {
 	return &task
 }
 
+func getPredecessors(predecessors string) ([]string, error) {
```
sfc-gh-jcieslak, line 277: I'm not sure if Trim and LastIndex functions will remove \" every time. I've wrote a simple test for the cases you mentioned and it failed for second case. Take a look

```go
func TestLastIdx(t *testing.T) {
	tcs := []string{
		"\".\"x",
		"\".x",
		"\".\".x",
	}
	for i, tc := range tcs {
		formattedName := strings.Trim(tc, "\\\"")
		idx := strings.LastIndex(formattedName, "\"") + 1
		if strings.LastIndex(formattedName, ".\"")+2 < idx {
			idx++
		}
		require.Equal(t, "x", formattedName[idx:], fmt.Sprintf("failed for %d", i))
	}
}
```

pkg/sdk/tasks_impl_gen.go
```diff
@@ -60,6 +63,45 @@ func (v *tasks) Execute(ctx context.Context, request *ExecuteTaskRequest) error
...
+			return nil, fmt.Errorf("unable to get predecessors for task %s err = %w", predecessor.FullyQualifiedName(), err)
+		}
+		rootTasks = append(rootTasks, predecessorTasks...)
+	}
+
+	// TODO [SNOW-884987]: extract unique function in our collection helper (if cycle-proof algorithm still needs it)
+	keys := make(map[string]bool)
+	uniqueRootTasks := make([]Task, 0, len(rootTasks))
+	for _, rootTask := range rootTasks {
+		if _, value := keys[rootTask.ID().FullyQualifiedName()]; !value {
```
sfc-gh-jcieslak, line 96: could you change that to `exists` or `ok` ? I was confused for sec, because the first returned value is the actual map value and the second one is existence check called "value"

pkg/sdk/tasks_def.go
```diff
@@ -114,15 +114,18 @@ var TasksDef = g.NewInterface(
 			OptionalQueryStructField(
 				"Set",
 				g.QueryStruct("TaskSet").
-					OptionalIdentifier("Warehouse", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().SQL("WAREHOUSE")).
+					OptionalIdentifier("Warehouse", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().Equals().SQL("WAREHOUSE")).
```
sfc-gh-jcieslak, line 117: we can add validation g.ValidIdentifierIfSet for Warehouse identifier

pkg/datasources/tasks.go
```diff
@@ -63,25 +63,24 @@ func Tasks() *schema.Resource {
 
 func ReadTasks(d *schema.ResourceData, meta interface{}) error {
 	db := meta.(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+	ctx := context.Background()
+
+	d.SetId("tasks_read")
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I must have copy-pasted it from somewhere, I don't think we need this. Removing.

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -192,6 +188,123 @@ func TestInt_Tasks(t *testing.T) {
...
+
+		require.Equal(t, []sdk.SchemaObjectIdentifier{}, root.Predecessors)
+
+		t1Name := random.String()
+		t1Id := sdk.NewSchemaObjectIdentifier(testDb(t).Name, testSchema(t).Name, t1Name)
+
+		request = sdk.NewCreateTaskRequest(t1Id, sql).WithAfter([]sdk.SchemaObjectIdentifier{rootId})
+		t1 := createTaskWithRequest(t, request)
+
+		require.Equal(t, []sdk.SchemaObjectIdentifier{rootId}, t1.Predecessors)
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Contains is a one-way check, which allows the collection to contain more. In other places, I do it by also checking the size and using contains (because order matters). But for single-element collections `Equal` is easier.

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -192,6 +188,123 @@ func TestInt_Tasks(t *testing.T) {
...
+	})
+
+	t.Run("create dag of tasks - multiple roots", func(t *testing.T) {
+		root1Name := random.String()
+		root1Id := sdk.NewSchemaObjectIdentifier(testDb(t).Name, testSchema(t).Name, root1Name)
+
+		request := sdk.NewCreateTaskRequest(root1Id, sql).WithSchedule(sdk.String("10 MINUTE"))
+		root1 := createTaskWithRequest(t, request)
+
+		require.Equal(t, []sdk.SchemaObjectIdentifier{}, root1.Predecessors)
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Done

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -192,6 +188,123 @@ func TestInt_Tasks(t *testing.T) {
...
+
+		root2Name := random.String()
+		root2Id := sdk.NewSchemaObjectIdentifier(testDb(t).Name, testSchema(t).Name, root2Name)
+
+		request = sdk.NewCreateTaskRequest(root2Id, sql).WithSchedule(sdk.String("10 MINUTE"))
+		root2 := createTaskWithRequest(t, request)
+
+		require.Equal(t, []sdk.SchemaObjectIdentifier{}, root2.Predecessors)
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Done

pkg/sdk/tasks_test.go
```diff
@@ -0,0 +1,108 @@
...
+			for i, predecessor := range v {
+				predecessors[i] = NewSchemaObjectIdentifier(db, sc, predecessor)
+			}
+			t.Predecessors = predecessors
+			r[k] = &t
+		}
+		return r
+	}
+
+	tests := []map[string][]string{
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I have started with struct, but it was a bit more verbose. I prefer it now because these are all just easy-to-read one-liners. But I can show you how it looked before and you will decide if we keep it or leave it.

pkg/sdk/tasks_impl_gen.go
sfc-gh-asawicki: Done

pkg/sdk/tasks_impl_gen.go
sfc-gh-asawicki: This were not the complete cases, just parts of them - how the "last" dot in the identifier is placed, depending on quotations around it (the ones that were failing - it appears that the old implementation was just bad, and still is not perfect).

I will just remove the comment because all of these three options are provided in the unit test and they are passing.

pkg/sdk/tasks_def.go
sfc-gh-asawicki: It's not needed now because an exactlyOneValueSet underneath uses the valueSet, which validates the correctness of identifiers. With such validation, we could never get to its branch to test it.

pkg/sdk/tasks_test.go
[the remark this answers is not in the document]
sfc-gh-asawicki: Discussed. We leave the map with a introduction comment.

sfc-gh-jcieslak (approved):

PR #2153, opened 2023-10-25
feat: Use tasks from the SDK followup

sfc-gh-asawicki:
Using tasks from the SDK in resource and data source was cleaned up. This is the continuation for #2140.

## Changes
 * The missing `SUSPEND_TASK_AFTER_NUM_FAILURES` attribute added
 * SessionParameters `setParam` method introduced in the previous PR used in other places
 * SessionParameters `setParam` method introduced in the previous PR unit tested
 * Changed old implementation of getting root tasks to be protected against cycles
 * Changed old implementation of handling session parameters change (it was only discovering parameters to add and to remove, not the ones that need to be changed)

## What is left
  * Retries will be extracted in the following PR (TODO left)

## Reference
- #2140


sfc-gh-jcieslak (approved):

pkg/resources/task.go
```diff
@@ -558,6 +588,20 @@ func UpdateTask(d *schema.ResourceData, meta interface{}) error {
...
+		o, n := d.GetChange("suspend_task_after_num_failures")
+		alterRequest := sdk.NewAlterTaskRequest(taskId)
+		if o.(int) > 0 && n.(int) == 0 {
+			alterRequest.WithUnset(sdk.NewTaskUnsetRequest().WithSuspendTaskAfterNumFailures(sdk.Bool(true)))
+		} else {
+			alterRequest.WithSet(sdk.NewTaskSetRequest().WithSuspendTaskAfterNumFailures(sdk.Int(n.(int))))
+		}
+		err := client.Tasks.Alter(ctx, alterRequest)
+		if err != nil {
+			return fmt.Errorf("error updating suspenf task after num failures on task %s", taskId.FullyQualifiedName())
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Done

pkg/sdk/tasks_impl_gen.go
```diff
@@ -66,40 +68,38 @@ func (v *tasks) Execute(ctx context.Context, request *ExecuteTaskRequest) error
...
-			uniqueRootTasks = append(uniqueRootTasks, rootTask)
+		predecessors := task.Predecessors
+		if len(predecessors) == 0 {
+			rootTasks = append(rootTasks, *task)
+		} else {
+			for _, p := range predecessors {
+				tasksToExamine.Push(p)
+			}
 		}
+		alreadyExaminedTasksNames = append(alreadyExaminedTasksNames, current.Name())
```
rmargarint-nydig, line 99: I think using current.Name() is an issue (Before this FullyQualifiedName was being returned in the array of task names)? After upgrading to the current release we're getting failures when creating or applying changes to DAG's of tasks related to the root task being unable to be found.

sfc-gh-asawicki: Hey @rmargarint-nydig. The `alreadyExaminedTasksNames` is built using only `Name()`s so I don't think that this is the issue.

If you are experiencing any problems with tasks, please create an issue with the description and the smallest possible sample showing the problems you are experiencing. Thanks!

PR #2113, opened 2023-10-11
feat: Migrate streams to the new sdk

sfc-gh-jcieslak:
## Test Plan

* [x] integration tests

* [x] unit tests

## References

* [CREATE STREAM](https://docs.snowflake.com/en/sql-reference/sql/create-stream)

PR #2163, opened 2023-10-30
feat: Migrate stage to the new sdk

sfc-gh-jcieslak:
Migrated stages to the new SDK + moved set comment in the FileFormat file to the right place, so now options could be reusable in other places without weird comment option.

## Test Plan
* [x] unit tests
* [x] integration tests

## References
* [Create stage](https://docs.snowflake.com/en/sql-reference/sql/create-stage)
* [Alter stage](https://docs.snowflake.com/en/sql-reference/sql/alter-stage)
* [Drop stage](https://docs.snowflake.com/en/sql-reference/sql/drop-stage)
* [Desc stage](https://docs.snowflake.com/en/sql-reference/sql/desc-stage)
* [Show stage](https://docs.snowflake.com/en/sql-reference/sql/show-stages)

## Other references
* [Internal stages](https://docs.snowflake.com/en/user-guide/data-load-local-file-system-create-stage#creating-a-named-stage)
* [Stage PUT command](https://docs.snowflake.com/en/sql-reference/sql/put)

## Things TODO in the next PR
- Introduce commands on stage like PUT, LS, COPY INTO, etc.
- Replace stage usage in test helper functions
- Add external stages when we will be able to test them


pkg/sdk/stages_def.go
```diff
@@ -0,0 +1,401 @@
...
+	OptionalAssignment("TYPE", g.KindOfTPointer[FileFormatType](), g.ParameterOptions()).
+	PredefinedQueryStructField("Options", g.KindOfTPointer[FileFormatTypeOptions](), g.ListOptions().NoComma())
+
+var stageCopyOptionsDef = g.NewQueryStruct("StageCopyOptions").
+	OptionalQueryStructField(
+		"OnError",
+		g.NewQueryStruct("StageCopyOnErrorOptions").
+			OptionalSQL("CONTINUE").
+			OptionalSQL("SKIP_FILE").
+			//OptionalSQL("SKIP_FILE_n").  // TODO templated value - not even supported by structToSQL
```
sfc-gh-jcieslak, line 88: We have a few options with this one
- enhance sql_builder to be able to make it work
- leave it as is with a comment e.g. in the readme that this should be eventually fixed
- replace the whole struct with string statement

pkg/sdk/poc/generator/list_builders.go
```diff
@@ -0,0 +1,6 @@
+package generator
+
+func (v *QueryStruct) List(name string, itemKind string, transformer FieldTransformer) *QueryStruct {
```
sfc-gh-asawicki, line 3: hmmm, in my change there is a different `List` added, so we should either merge them or change their names respectively (https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2171/files#diff-7d0e3fbb8967ad1b37eea35ab269bebbeb403363ebf75f5c43fdb10e77d7d839R65)

pkg/sdk/poc/generator/query_struct.go
```diff
@@ -32,6 +32,11 @@ func (v *QueryStruct) WithValidation(validationType ValidationType, fieldNames .
 	return v
 }
 
+func (v *QueryStruct) PredefinedQueryStructField(name string, kind string, transformer FieldTransformer) *QueryStruct {
```
sfc-gh-asawicki, line 35: Could it be used in `keyword_builders.go` for methods like `OptionalLike` and others (because now these predefined query struct fields are used there either way)

pkg/sdk/stages_gen_test.go
```diff
@@ -0,0 +1,718 @@
...
+			Type: &InternalStageEncryptionFull,
+		}
+		opts.DirectoryTableOptions = &InternalDirectoryTableOptions{
+			Enable:          Bool(true),
+			RefreshOnCreate: Bool(true),
+		}
+		opts.FileFormat = &StageFileFormat{
+			FormatName: String("format name"),
+		}
+		opts.CopyOptions = &StageCopyOptions{
```
sfc-gh-asawicki, line 52: please add more than one copy option (here and in other places), to see that we are generating multiple options correctly according to syntax

sfc-gh-asawicki (changes requested):

pkg/sdk/poc/generator/query_struct.go
sfc-gh-jcieslak: True, I'll use them there

pkg/sdk/poc/generator/list_builders.go
sfc-gh-jcieslak: I needed one that would be more flexible and accept FieldTransformer interface. so List could represent a Go slice and still be a keyword or parameter, which I think was my case, so I think we can use that one and adjust the one in your pr. wdyt ?

sfc-gh-asawicki: Yep, we can leave yours and I will rename mine to be more specific

pkg/sdk/poc/generator/query_struct.go
sfc-gh-jcieslak: Unfortunately we cannot use it inside generator package, because it will create circular dependencies

pkg/sdk/poc/generator/list_builders.go
```diff
@@ -0,0 +1,6 @@
...
+	v.fields = append(v.fields, NewField(name, KindOfSlice(itemKind), Tags(), transformer))
```
sfc-gh-asawicki, line 4: Shouldn't it be `Tags().List()` here? And maybe ListTransformer instead of FieldTransformer.

pkg/sdk/poc/generator/query_struct.go
sfc-gh-asawicki: But both are inside the generator package, right? So can you elaborate more on the problem here?

pkg/sdk/stages_def.go
sfc-gh-asawicki: For me the second option for now.

pkg/sdk/poc/generator/query_struct.go
sfc-gh-jcieslak: Nvm, I added them. I wanted to use sdk types like sdk.TagAssociation and that was causing the circular dependency, and that's why I backed off with the idea of replacing existing builders. But now, I replaced them with string types, so no circular dependency and the builder methods are shorter, so probably it was worth it even though I couldn't use sdk types.

pkg/sdk/stages_gen_test.go
sfc-gh-jcieslak: Added

pkg/sdk/poc/generator/list_builders.go
sfc-gh-jcieslak: True, I think I needed a different use case. I'm going to add Tags().List() and change the type to *ListTransformer and add other builder function in keyword_builders called NamedList, so that sql prefix could be used with list type (sql prefix is not allowed in `ddll:"list"`).

pkg/sdk/stages_dto_builders_gen.go
```diff
@@ -0,0 +1,825 @@
...
+	s.SkipFile = SkipFile
+	return s
+}
+
+func (s *StageCopyOnErrorOptionsRequest) WithAbortStatement(AbortStatement *bool) *StageCopyOnErrorOptionsRequest {
+	s.AbortStatement = AbortStatement
+	return s
+}
+
+func NewCreateOnS3StageRequest(
```
sfc-gh-swinkler, line 164: if you have `CreateInternal` then wouldnt it be more appropriate to call this `CreateExternalForS3`?

pkg/sdk/stages_dto_gen.go
```diff
@@ -0,0 +1,263 @@
...
+
+type ExternalS3StageParamsRequest struct {
+	Url                string // required
+	StorageIntegration *AccountObjectIdentifier
+	Credentials        *ExternalStageS3CredentialsRequest
+	Encryption         *ExternalStageS3EncryptionRequest
+}
+
+type ExternalStageS3CredentialsRequest struct {
+	AwsKeyId     *string
```
sfc-gh-swinkler, line 88: minor point, but Aws is an acronymn that stands for Amazon Web Services. I don't think it should ever be written as Aws. It should be AWSKeyId. If this is an intentional design decision for readability, then we should agree on it.

pkg/sdk/stages_dto_builders_gen.go
```diff
@@ -0,0 +1,825 @@
...
+	s.Temporary = Temporary
+	return s
+}
+
+func (s *CreateInternalStageRequest) WithIfNotExists(IfNotExists *bool) *CreateInternalStageRequest {
+	s.IfNotExists = IfNotExists
+	return s
+}
+
+func (s *CreateInternalStageRequest) WithEncryption(Encryption *InternalStageEncryptionRequest) *CreateInternalStageRequest {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: It is an enum type, It's in the definition file `InternalStageEncryptionOption`. It's just that the generator generates a request type for this enum, so instead of `.WithEncryption(myEnum)` it `.WithEncryption(MyEnumRequest(myEnum))`. I think we can leave it as is and this will change after dto builder refactoring. I'll add note to the readme about enums.

pkg/sdk/stages_dto_builders_gen.go
sfc-gh-jcieslak: Same as for enum types, that could be our next generator improvement. Added to readme. (previous discussion on this - https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2113#discussion_r1356777447)

pkg/sdk/stages_dto_gen.go
sfc-gh-jcieslak: Sure, I've changed it to uppercase. I've talked with @sfc-gh-asawicki and maybe we could use uppercased version, because it's used in other places like CSV, JSON and XML in the file format files.

sfc-gh-asawicki (approved):

PR #2184, opened 2023-11-09
chore: add username_password auth test

sfc-gh-swinkler:
adds a test case for username and password acceptance test, more tests to follow.

### DIFF
--- a/pkg/provider/provider_acc_test.go
+++ b/pkg/provider/provider_acc_test.go
@@ -1,0 +1,102 @@
+package provider
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-mux/tf5to6server"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+var testAccProvider *schema.Provider
+
+var testAccProtoV6ProviderFactories = map[string]func() (tfprotov6.ProviderServer, error){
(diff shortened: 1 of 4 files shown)


pkg/provider/provider_acc_test.go
```diff
@@ -0,0 +1,102 @@
...
+			},
+		},
+	})
+}
+
+func testAccCheckCurrentAccount(t *testing.T, n string) resource.TestCheckFunc {
+	t.Helper()
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[n]
+		if !ok {
```
sfc-gh-asawicki, line 80: Why not use `testify/assert` or `testify/require` instead?

pkg/provider/provider_acc_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+var testAccProvider *schema.Provider
+
+var testAccProtoV6ProviderFactories = map[string]func() (tfprotov6.ProviderServer, error){
```
sfc-gh-jcieslak, line 20: shouldn't this (+ testAccProvider variable and init function) be placed in some test_helper file? It seems like a variable that will be used in all new tests.

pkg/provider/provider_acc_test.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func init() {
+	testAccProvider = Provider()
+}
+
+func TestAcc_ProviderUsernameAndPasswordAuth(t *testing.T) {
+	user := os.Getenv("SNOWFLAKE_USER")
+	if user == "" {
+		t.Skip("SNOWFLAKE_USER must be set")
```
sfc-gh-asawicki, line 36: Skip means this test will be treated as a success (currently). Should it be skipped or failed in such a case?

sfc-gh-asawicki (changes requested):

sfc-gh-asawicki:
Superseded by #2551

PR #2186, opened 2023-11-10
fix: changing masking policy on a column requires FORCE

malduarte:
Attempting to change masking policy without FORCE would cause the following error:

003529 (42601): Specified column already attached to another masking policy. A column cannot be attached to multiple masking policies. Please drop the current association in order to attach a new masking policy.

This PR adds FORCE to the generated ALTER statement and improves error information if masking policy change fails to apply.

## Test Plan
Added TestAcc_TableWithMaskingPolicy that
- creates 2 masking policies (pol1 and pol1)
- creates single column table using masking policy pol1
- changes masking policy from pol1 to pol2 

## References
- Updating masking policy requires force: https://docs.snowflake.com/en/user-guide/security-column-intro#replace-a-masking-policy-on-a-column


sfc-gh-asawicki:
Hey @malduarte. Thanks for the contribution.

Unfortunately, we are currently changing the implementation used in resources as a part of our ongoing SDK rewrite. This means we will eliminate table implementation residing in the `snowflake` package.

However, we will add this behavior as a part of the process.

malduarte:
@sfc-gh-asawicki Is there an active branch? I'm happy to submit the MR there

sfc-gh-asawicki:
Yes, there is an open PR: #2042.

Issue #1624, opened 2023-03-16
Resource monitor timestamps are always considered to have changed, preventing second `terraform apply`


mpbarlow:
**Provider Version**

0.58.2

**Terraform Version**

1.3.9

**Describe the bug**

There are actually two issues here, though one is comparatively minor. 

The main issue:

Creating a resource monitor with either the start or end timestamp in the format `YYYY-MM-DD HH:mm` works when the monitor is first created, but on subsequent runs of `terraform apply`, those timestamps are picked up as a change from `YYYY-MM-DDTHH:mm-HH:mm`. It appears that the provider is then trying to set the timestamp but _not_ the frequency, triggering the error "Must specify frequency and start time together." The only workaround I've found for this is to taint the resource monitor each time, to force it to be recreated. 

Setting the extended `YYYY-MM-DDTHH:mm-HH:mm` timestamp for an existing monitor triggers the same error, and that format is not recognised for new monitors either. 

The second issue, which I discovered by accident, is that the above error is also displayed if creating a monitor with an empty `notify_users` attribute. This is not the end of the world, but is quite confusing.  

**Expected behavior**

- The timestamp is not considered a modification unless it has actually changed
- Modifications to the timestamp are permitted without tainting the resource, as they are currently not even if you _do_ change the the `start_timestamp`
- An empty set is permitted for `notify_users` _or_ the error message returned correctly identifies that the empty set is the problem

**Code samples and commands**

```hcl
resource "snowflake_resource_monitor" "monitor" {
  name = "MONITOR"
  credit_quota = 100
  frequency = "DAILY"
  start_timestamp = "2023-03-20 00:00"
  notify_triggers = [40]
  suspend_triggers = [50]
  suspend_immediate_triggers = [90]
  notify_users = ["johnsmith"]
}
```

```bash
# This works
terraform apply

# This does not
terraform apply
```

**Additional context**

N/A

bbelescot:
Any update on this? Impacted as well. Thank you!

curtisq:
Another ➕ for this. This makes it really annoying to manage resource monitors with this provider.

rasmus-kristensen:
I just hit this exact same issue thinking I had made a mistake.

Bob-Jia-catalina:
Hi I ran into the same issue today, Resource Monitor 'start_timestamp' is always considered 'changed' whenever 'terraform plan' or 'terraform apply'.

[...]

PR #2214, opened 2023-11-24
fix: for 1624 resource monitor timestamps are always considered to have changed

RichBarnes:
## Summary
Three changes:

1. validation logic requires both frequency and start_timestamp to be retrieved if either has been altered. If start_timestamp is set to immediately it has always changed on second terraform apply. Updated to set the haschanged boolean if either has been modified to correctly pass validation.

2. When start_timestamp is set to a specific date time, this is stored in Snowflake as a UTC date with a local offset.  This causes an incorrect date to be returned and written to the tf state file. Added code to parse returned UTC back into a local format.

3. previous suspend_trigger naming support appears to cause issues once the start_timestamp problems are fixed. Changed logic to and rather then or's which resolved the issue. 

Noticed a separate issue with the sql for trigger changes, if credit_quota hasn't been set then that line isn't included in the SQL builder which results in malformed sql and an error back from Snowflake. This should be raised as a separate fix/bug ticket.
The minor notify_users attribute behaviour should be resolved under another fix ticket.

## Test Plan

Setting start_timestamp to IMMEDIATELY or a change to start_timestamp  and running terraform apply twice reproduces the issue. Post fix, changing these values results in desired behaviour and no errors.

Ran existing tests as only logic minor changes to hasChanged logic and timestamp retrieval. 
* [x] acceptance tests
* [x] integration tests
* [x] unit tests

## References
Allow closure of [#1624 ](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1624)


sfc-gh-asawicki:
Hey @RichBarnes, thanks for the PR!

I will add review comments (and one of @sfc-gh-jcieslak @sfc-gh-swinkler will also take a look).

pkg/resources/resource_monitor.go
```diff
@@ -335,17 +335,13 @@ func UpdateResourceMonitor(d *schema.ResourceData, meta interface{}) error {
 		opts.Set.CreditQuota = sdk.Pointer(d.Get("credit_quota").(int))
 	}
 
-	if d.HasChange("frequency") {
+	if d.HasChange("frequency") || d.HasChange("start_timestamp") {
```
sfc-gh-asawicki, line 338: There is only one acceptance test for resource monitor that is currently run. It would be nice to have one showing the issue you are fixing (should fail before the fix, but should be okay after). Can you add it?

pkg/sdk/resource_monitors.go
```diff
@@ -97,7 +98,13 @@ func (row *resourceMonitorRow) convert() (*ResourceMonitor, error) {
 		resourceMonitor.Frequency = *frequency
 	}
 	if row.StartTime.Valid {
-		resourceMonitor.StartTime = row.StartTime.String
+		const YYMMDDhhmm = "2006-01-02 15:04"
```
sfc-gh-asawicki, line 101: There is a test `TestInt_ResourceMonitorAlter`#`when changing scheduling info` in `resource_monitors_integration_test.go`. It uses `ParseTimestampWithOffset` that already uses similar logic, but it's run only for test check, and not the implementation.

Maybe this should be reused/merged with your proposed solution?

sfc-gh-asawicki (changes requested):

pkg/resources/resource_monitor.go
RichBarnes: Hey, yes I'll add 👍

pkg/sdk/resource_monitors.go
RichBarnes: I'll take a look - thanks

pkg/sdk/resource_monitors.go
```diff
@@ -97,7 +98,13 @@ func (row *resourceMonitorRow) convert() (*ResourceMonitor, error) {
...
+		t, err := time.Parse(time.RFC3339, row.StartTime.String)
+		if err != nil {
+			return nil, err
+		}
+		localTime := t.Local()
+		resourceMonitor.StartTime = localTime.Format(YYMMDDhhmm)
 	}
 	if row.EndTime.Valid {
```
sfc-gh-jcieslak, line 109: The issue was also mentioning the end timestamp, so shouldn't this also change?

sfc-gh-jcieslak: The issue also mentions the `notify_users` param. Did you look into that?

RichBarnes: Good point - I'll add

pkg/resources/resource_monitor.go
RichBarnes: Hey, I have added a acceptance test for start and endtime stamps

pkg/sdk/resource_monitors.go
RichBarnes: Hey, I've moved parsers.go into sdk and updated with a merged parse timestamp function. Updated the integration tests and used the function in the resource monitor update.

pkg/resources/resource_monitor.go
```diff
@@ -1,507 +1,514 @@
...
+
+	ctx := context.Background()
+	var runSetStatement bool
+
+	opts := sdk.AlterResourceMonitorOptions{Set: &sdk.ResourceMonitorSet{}}
+
+	if d.HasChange("notify_users") {
+		runSetStatement = true
+		userNames := expandStringList(d.Get("notify_users").(*schema.Set).List())
+		users := []sdk.NotifiedUser{{}}
```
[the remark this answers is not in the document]
RichBarnes: My bad, thought I'd removed it. Done.

pkg/resources/resource_monitor.go
```diff
@@ -1,507 +1,514 @@
...
+		for _, userName := range userNames {
+			users = append(users, sdk.NotifiedUser{Name: userName})
+		}
+		users = append(users[:0], users[1:]...)
+
+		if opts.NotifyUsers == nil {
```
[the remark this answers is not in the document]
RichBarnes: agreed, removed.

pkg/resources/resource_monitor_acceptance_test.go
```diff
@@ -1,146 +1,218 @@
...
+
+func resourceMonitorConfigInitialTimestamp(accName string) string {
+	return fmt.Sprintf(`
+resource "snowflake_warehouse" "warehouse" {
+  name           = "test"
+  comment        = "foo"
+  warehouse_size = "XSMALL"
+}
+
+resource "snowflake_resource_monitor" "test" {
```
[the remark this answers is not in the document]
RichBarnes: a notify_user has to be a real user (unless there's a way to mock it), so I've added a new test TestAcc_ResourceMonitorUpdateNotifyUsers, also an empty user set is ignored when properties are returned.

pkg/resources/resource_monitor_acceptance_test.go
```diff
@@ -1,146 +1,218 @@
...
+					resource.TestCheckResourceAttr("snowflake_resource_monitor.test", "end_timestamp", "2055-01-01 12:00"),
+				),
+			},
+			{
+				Config: resourceMonitorConfigUpdatedTimestamp(name),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_resource_monitor.test", "name", name),
+					resource.TestCheckResourceAttr("snowflake_resource_monitor.test", "frequency", "WEEKLY"),
+					resource.TestCheckResourceAttr("snowflake_resource_monitor.test", "start_timestamp", "2055-01-01 12:00"),
+					resource.TestCheckResourceAttr("snowflake_resource_monitor.test", "end_timestamp", "2056-01-01 12:00"),
```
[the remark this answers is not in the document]
RichBarnes: see above.

sfc-gh-jcieslak (approved):

pkg/sdk/resource_monitors.go
sfc-gh-asawicki: Hey, did you revert it? I can't see the move of `parsers.go` currently.

pkg/resources/resource_monitor_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-asawicki: I would prefer to use `ContextFunctions.CurrentUser`, because we are now not setting tests up with SNOWFLAKE_USER, but with prepared config. But we can do this cleanup after it's merged, because it's also about the other test in this file (as you already pointed out).

pkg/sdk/resource_monitors.go
RichBarnes: Hey -my branch reverted (didn't do this intentionally), I've re-pushed all the changes, hopefully at parity now!

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (dismissed):

sfc-gh-asawicki:
@RichBarnes you have 2 approves but still 2 conflicts that must be resolved before merging. :)

RichBarnes:
> @RichBarnes you have 2 approves but still 2 conflicts that must be resolved before merging. :)

Conflicts should be fixed, had issues with my local branch/git :)

sfc-gh-asawicki (approved):

PR #2225, opened 2023-11-30
feat: Add unsafe_execute resource

sfc-gh-asawicki:
Introduce unsafe_execute experimental resource.

The resource is not recommended to be used in production environments. It consists of:
* required `execute` and `revert` - they are not validated in any way
* optional `query` that fetches results and updates `query_results`; in case of error clears `query_results`

### DIFF
--- a/pkg/sdk/testint/client_unsafe_extensions_integration_test.go
+++ b/pkg/sdk/testint/client_unsafe_extensions_integration_test.go
@@ -26,7 +26,6 @@ func TestInt_Client_UnsafeQuery(t *testing.T) {
 		assert.Equal(t, "", *row["options"])
 		assert.Equal(t, "", *row["comment"])
 		assert.Equal(t, "N", *row["is_default"])
-		assert.Equal(t, "Y", *row["is_current"])
 		assert.Nil(t, *row["budget"])
 	})
 


pkg/sdk/client_extensions_unsafe.go
```diff
@@ -0,0 +1,65 @@
+package sdk
+
+import (
+	"context"
+	"database/sql"
+)
+
+func (c *Client) ExecUnsafe(ctx context.Context, sql string) (sql.Result, error) {
```
sfc-gh-jcieslak, line 8: If that is an extension maybe we shouldn't pollute the main client with such functions and we should just create a new client with the extension we want to add, what I mean by that:
```go
type UnsafeClient struct {
	*Client
}

func NewUnsafeClientFromDb(db *sql.DB) *UnsafeClient {
	return &UnsafeClient{
		Client: NewClientFromDB(db),
	}
}

func (c *UnsafeClient) UnsafeExec(ctx context.Context, sql string) (sql.Result, error) {
        // c (UnsafeClient) can call the same functions as regular client (even private ones, but they need to be in the same package)
	return c.exec(ctx, sql)
}
```

sfc-gh-jcieslak (approved):

pkg/sdk/client_extensions_unsafe.go
sfc-gh-asawicki: Can we change it in the following PR?

Issue #2212, opened 2023-11-24
PASSWORD_MIN_AGE_DAYS and PASSWORD_HISTORY parameters availability for snowflake_password_policy resource


GomathiMa:
### Terraform CLI and Provider Versions

Terraform Snowflake provider version: 68.2

### Use Cases or Problem Statement

We are not able to utilise the PASSWORD_MIN_AGE_DAYS and PASSWORD_HISTORY parameters  in snowflake_password_policy resource for password policy creation.

### Proposal

Please add the parameters PASSWORD_MIN_AGE_DAYS and PASSWORD_HISTORY parameters  in snowflake_password_policy resource

### How much impact is this issue causing?

Medium

### Additional Information

Password acceptable criteria not being met due to this.

sfc-gh-asawicki:
Closing as done in #2231

PR #2231, opened 2023-12-04
feat: Add missing parameters to password policy resource

sfc-gh-asawicki:
- Add `PASSWORD_MIN_AGE_DAYS` and `PASSWORD_HISTORY` to `snowflake_password_policy` resource
- revert COMMENT functionality (it's working again)
- fix UNSET validations and implementation for password policy and tasks (it will be fixed for other resources in a separate issue/PR)
- move acceptance tests config to file
- solves #2212


pkg/resources/password_policy.go
```diff
@@ -90,6 +90,13 @@ var passwordPolicySchema = map[string]*schema.Schema{
 		Description:  "Specifies the minimum number of special characters the password must contain. Supported range: 0 to 256, inclusive. Default: 1",
 		ValidateFunc: validation.IntBetween(0, 256),
 	},
+	"min_age_days": {
+		Type:         schema.TypeInt,
+		Optional:     true,
+		Default:      0,
```
sfc-gh-swinkler, line 96: i don't know that we actually need to set the default. Snowflake has its own default if a value isn't set, we don't need to specify it here.

pkg/resources/password_policy_acceptance_test.go
```diff
@@ -1,49 +1,64 @@
...
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
 		CheckDestroy: nil,
 		Steps: []resource.TestStep{
 			{
-				Config: passwordPolicyConfig(accName, 10, 30, "this is a test resource", acc.TestDatabaseName, acc.TestSchemaName),
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_PasswordPolicy_basic"),
```
sfc-gh-swinkler, line 36: why not use config.TestNameDirectory()?
```
// TestNameDirectory returns the name of the test prefixed with
// "testdata".
//
// For example, given test code:
//
//	func TestExampleCloudThing_basic(t *testing.T) {
//	    resource.Test(t, resource.TestCase{
//	        Steps: []resource.TestStep{
//	            {
//	                ConfigDirectory: config.TestNameDirectory(),
//	            },
//	        },
//	    })
//	}
//
// The testing configurations will be expected in the
// testdata/TestExampleCloudThing_basic/ directory.
func TestNameDirectory() func(TestStepConfigRequest) string {
	return func(req TestStepConfigRequest) string {
		return filepath.Join("testdata", req.TestName)
	}
}
```

pkg/resources/password_policy.go
sfc-gh-asawicki: The default is set for other params in this resource. The default is the SF default from the docs. I have added it for two reasons:
- consistency with other parameters
- consistency in update implementation for this resource (as you can see, there is no set/unset, but set only; it behaves right because `d.Get()` returns the default if not set - I added test case for this because the implementation seemed strange for me at first sight)

pkg/resources/password_policy_acceptance_test.go
sfc-gh-asawicki: Done.

pkg/resources/password_policy_acceptance_test.go
```diff
@@ -1,49 +1,64 @@
...
-					Check: resource.ComposeTestCheckFunc(
-						resource.TestCheckResourceAttr("snowflake_password_policy.pa", "comment", ""),
-					),
-				},
-			*/
 			{
+				ConfigDirectory: config.TestNameDirectory(),
+				ConfigVariables: m(20, 50, ""),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_password_policy.pa", "comment", ""),
```
sfc-gh-jcieslak, line 56: maybe you should also check other params in the previous test case, e.g. to show that when the comment is set the attribute is also set

pkg/resources/password_policy_acceptance_test.go
```diff
@@ -1,49 +1,64 @@
...
 func TestAcc_PasswordPolicy(t *testing.T) {
 	accName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
+	m := func(minLength int, maxLength int, comment string) map[string]config.Variable {
+		return map[string]config.Variable{
+			"name":       config.StringVariable(accName),
+			"database":   config.StringVariable(acc.TestDatabaseName),
+			"schema":     config.StringVariable(acc.TestSchemaName),
+			"min_length": config.IntegerVariable(minLength),
+			"max_length": config.IntegerVariable(maxLength),
+			"comment":    config.StringVariable(comment),
```
sfc-gh-jcieslak, line 23: to make it more complete maybe add new fields?

pkg/resources/password_policy_acceptance_test.go
sfc-gh-asawicki: oky

pkg/resources/password_policy_acceptance_test.go
sfc-gh-asawicki: there are not all fields specified in the test (even not all previous fields), so I think it's fine to test just a few

sfc-gh-jcieslak (approved):

PR #2234, opened 2023-12-06
fix: Fix some bugs

sfc-gh-asawicki:
TODO: describe


pkg/datasources/alerts_acceptance_test.go
```diff
@@ -0,0 +1,118 @@
...
+		CheckDestroy: nil,
+		Steps: []resource.TestStep{
+			{
+				Config: alertsResourceConfig(name) + alertsDatasourceConfigNoOptionals(),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrSet("data.snowflake_alerts.test_datasource_alert", "alerts.#"),
+				),
+			},
+			{
+				Config: alertsResourceConfig(name) + alertsDatasourceConfigDbOnly(),
```
sfc-gh-swinkler, line 32: is there a reason to be having the config inline vs in the testdata folder?

pkg/resources/database_acceptance_test.go
```diff
@@ -83,6 +89,50 @@ func TestAcc_Database(t *testing.T) {
...
+func TestAcc_DatabaseRemovedOutsideOfTerraform(t *testing.T) {
+	id := generateUnsafeExecuteTestDatabaseName(t)
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: testAccCheckDatabaseExistence(t, id, false),
```
sfc-gh-swinkler, line 101: should this function be moved into the database acceptance tests? it's strange that this function is in a different file. because i imagine we will be having more of these check destroy functions in the future, and these functions should be in the file for the object. or maybe we make one generic check destroy function that can handle any type.

sfc-gh-swinkler: instead of putting this in the checkdestroy, when should probably put this in a custom check function. there is actually an example here of this exact use case https://developer.hashicorp.com/terraform/plugin/sdkv2/testing/acceptance-tests/teststep#custom-check-functions

sfc-gh-asawicki: I can move it to helper but this will result in new pumped helper file.

I want to have it in check destroy, because I check after destroy, not a custom test check which is done after step.

pkg/datasources/alerts_acceptance_test.go
sfc-gh-asawicki: @sfc-gh-jcieslak wdyt?

sfc-gh-jcieslak (approved):

sfc-gh-swinkler (dismissed):

sfc-gh-jcieslak (approved):

Issue #2248, opened 2023-12-08
Support plural ICEBERG TABLES in snowflake_grant_privileges_to_role


danu165:
### Terraform CLI and Provider Versions

CLI: 1.0.11
Provider: 0.78.0

### Use Cases or Problem Statement

`snowflake_grant_privileges_to_role` is not able to grant privileges such as `SELECT` on all future iceberg tables. Code to reproduce:
```
resource "snowflake_role" "role" {
  name       = "DUMMY_ICEBERG_ROLE"
}

resource "snowflake_grant_privileges_to_role" "iceberg_privs" {
  depends_on = [snowflake_role.role]
  privileges = ["SELECT"]
  role_name  = snowflake_role.role.name

  on_schema_object {
    future {
      object_type_plural = "ICEBERG TABLES"
      in_schema          = "MY_DB.MY_SCHEMA"
    }
  }
}
```

Error:
```
╷
│ Error: expected on_schema_object.0.future.0.object_type_plural to be one of ["ALERTS" "DYNAMIC TABLES" "EVENT TABLES" "FILE FORMATS" "FUNCTIONS" "PROCEDURES" "SECRETS" "S
EQUENCES" "PIPES" "MASKING POLICIES" "PASSWORD POLICIES" "ROW ACCESS POLICIES" "SESSION POLICIES" "TAGS" "STAGES" "STREAMS" "TABLES" "EXTERNAL TABLES" "TASKS" "VIEWS" "MATE
RIALIZED VIEWS"], got ICEBERG TABLES
│
│   with snowflake_grant_privileges_to_role.ice_privs,
│   on main.tf line 50, in resource "snowflake_grant_privileges_to_role" "ice_privs":
│   50:       object_type_plural = "ICEBERG TABLES"
│

```

### Proposal

Either support `ICEBERG TABLES` in `object_type_plural` or possibly remove the validation in general so it's easier to use new Snowflake features without contributing to provider code.

### How much impact is this issue causing?

High

### Additional Information

We are working to deliver Iceberg Tables by December 13th to our company. We would need the ability to issue these grants for Iceberg Tables.

danu165:
Note: we would also need support for grants on `EXTERNAL VOLUMES`. Additionally, `object_type` should support `ICEBERG TABLE` and  `EXTERNAL VOLUME`.

bjornm82:
How to create the external volumes in terraform? This are also missing still right?

[...]

PR #2250, opened 2023-12-11
fix: Add missing parameters

sfc-gh-jcieslak:
A fix for #2248

### DIFF
--- a/pkg/resources/grant_privileges_to_role.go
+++ b/pkg/resources/grant_privileges_to_role.go
@@ -56,7 +56,7 @@ var grantPrivilegesToRoleSchema = map[string]*schema.Schema{
 				"object_type": {
 					Type:        schema.TypeString,
 					Required:    true,
-					Description: "The object type of the account object on which privileges will be granted. Valid values are: USER | RESOURCE MONITOR | WAREHOUSE | DATABASE | INTEGRATION | FAILOVER GROUP | REPLICATION GROUP",
+					Description: "The object type of the account object on which privileges will be granted. Valid values are: USER | RESOURCE MONITOR | WAREHOUSE | DATABASE | INTEGRATION | FAILOVER GROUP | REPLICATION GROUP | EXTERNAL VOLUME",
 					ValidateFunc: validation.StringInSlice([]string{
 						"USER",
 						"RESOURCE MONITOR",
@@ -121,7 +121,7 @@ var grantPrivilegesToRoleSchema = map[string]*schema.Schema{
 				"object_type": {
 					Type:          schema.TypeString,
 					Optional:      true,
-					Description:   "The object type of the schema object on which privileges will be granted. Valid values are: ALERT | DYNAMIC TABLE | EVENT TABLE | FILE FORMAT | FUNCTION | PROCEDURE | SECRET | SEQUENCE | PIPE | MASKING POLICY | PASSWORD POLICY | ROW ACCESS POLICY | SESSION POLICY | TAG | STAGE | STREAM | TABLE | EXTERNAL TABLE | TASK | VIEW | MATERIALIZED VIEW",
+					Description:   "The object type of the schema object on which privileges will be granted. Valid values are: ALERT | DYNAMIC TABLE | EVENT TABLE | FILE FORMAT | FUNCTION | ICEBERG TABLE | PROCEDURE | SECRET | SEQUENCE | PIPE | MASKING POLICY | PASSWORD POLICY | ROW ACCESS POLICY | SESSION POLICY | TAG | STAGE | STREAM | TABLE | EXTERNAL TABLE | TASK | VIEW | MATERIALIZED VIEW",
 					RequiredWith:  []string{"on_schema_object.0.object_name"},
 					ConflictsWith: []string{"on_schema_object.0.all", "on_schema_object.0.future"},
 					ForceNew:      true,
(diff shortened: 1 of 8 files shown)


pkg/resources/grant_privileges_to_role.go
```diff
@@ -65,6 +65,7 @@ var grantPrivilegesToRoleSchema = map[string]*schema.Schema{
 						"INTEGRATION",
 						"FAILOVER GROUP",
 						"REPLICATION GROUP",
+						"EXTERNAL VOLUME",
```
sfc-gh-asawicki, line 68: You did not change the description. And they are also listed there. You will also have to regenerate docs after changing the description.

pkg/resources/grant_privileges_to_role.go
```diff
@@ -191,6 +192,7 @@ var grantPrivilegesToRoleSchema = map[string]*schema.Schema{
 									"TASKS",
 									"VIEWS",
 									"MATERIALIZED VIEWS",
+									"ICEBERG TABLES",
```
sfc-gh-asawicki, line 195: same with description here

pkg/resources/grant_privileges_to_role.go
```diff
@@ -245,6 +247,7 @@ var grantPrivilegesToRoleSchema = map[string]*schema.Schema{
...
```
sfc-gh-asawicki, line 250: same with descritpion here

pkg/resources/grant_privileges_to_role_acceptance_test.go
```diff
@@ -2,6 +2,8 @@ package resources_test
 
 import (
 	"fmt"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
```
sfc-gh-asawicki, line 6: move imports pls

pkg/sdk/grants_test.go
```diff
@@ -35,6 +35,22 @@ func TestGrantPrivilegesToAccountRole(t *testing.T) {
...
+				AllPrivileges: Bool(true),
+			},
+			on: &AccountRoleGrantOn{
+				AccountObject: &GrantOnAccountObject{
+					ExternalVolume: Pointer(NewAccountObjectIdentifier("ex volume")),
+				},
+			},
+			accountRole: NewAccountObjectIdentifier("role1"),
+		}
+		assertOptsValidAndSQLEquals(t, opts, `GRANT ALL PRIVILEGES ON EXTERNAL VOLUME "ex volume" TO ROLE "role1"`)
```
sfc-gh-asawicki, line 51: I wonder if external volume identifier follows the normal identifier rules

pkg/sdk/privileges.go
```diff
@@ -71,6 +72,9 @@ const (
 	AccountObjectPrivilegeMonitor            AccountObjectPrivilege = "MONITOR"
 	AccountObjectPrivilegeUsage              AccountObjectPrivilege = "USAGE"
 
+	// -- For EXTERNAL VOLUME
+	// AccountObjectPrivilegeUsage              AccountObjectPrivilege = "USAGE" (duplicate)
```
sfc-gh-asawicki, line 76: why this is commented out?

pkg/sdk/privileges.go
```diff
@@ -178,6 +184,15 @@ const (
 	// USAGE [ , ... ]
 	SchemaObjectPrivilegeUsage SchemaObjectPrivilege = "USAGE"
 
+	// -- For ICEBERG TABLE
+	SchemaObjectPrivilegeApplyBudget SchemaObjectPrivilege = "APPLYBUDGET"
+	//SchemaObjectPrivilegeDelete      SchemaObjectPrivilege = "DELETE" (duplicate)
```
sfc-gh-asawicki, line 189: Why these are commented out?

sfc-gh-asawicki (changes requested):

pkg/sdk/grants_test.go
sfc-gh-jcieslak: From the documentation it seems like a normal account object identifier, plus `An external volume is a named, account-level Snowflake object` - [docs](https://docs.snowflake.com/en/user-guide/tables-iceberg#external-volumes)

pkg/sdk/privileges.go
sfc-gh-jcieslak: So it seems like we're adding the privileges as they appear in the documentation, e.g. for External Volume ... For Tables ... etc. For some of them, they can duplicate (e.g. tables and iceberg tables may have "SELECT") and the practice in this file was to comment the duplicate with the comment `(duplicate)` and the end.

pkg/sdk/privileges.go
sfc-gh-jcieslak: same reason as the above

pkg/sdk/privileges.go
sfc-gh-asawicki: Oh, okay. We should deal with that. Maybe as a subtask in redesign grants? (close after reading and adding it there if you agree)

pkg/resources/grant_privileges_to_role_acceptance_test.go
```diff
@@ -861,3 +864,38 @@ func TestAcc_GrantPrivilegesToRole_onSchemaObject_futureInDatabase_externalTable
 		},
 	})
 }
+
+func TestAcc_GrantPrivilegesToRole_onSchemaObject_futureIcebergTables(t *testing.T) {
```
sfc-gh-asawicki, line 868: +1 for that test

sfc-gh-asawicki (approved):

PR #2269, opened 2023-12-14
feat: identifier validation

sfc-gh-jcieslak:
This PR introduces a function that validates identifiers and prints readable error messages that sometimes may be able to help resolve the common issue of specifying identifiers incorrectly (especially in grant resources).
The validator will be used in the follow-up pull requests.

## Test Plan

* [x] unit tests


pkg/resources/helpers.go
```diff
@@ -142,3 +146,77 @@ func IsDataType() schema.SchemaValidateFunc { //nolint:staticcheck
...
+// Typically, we expect passed identifiers to be a variation of sdk.ObjectIdentifier. To use this function, pass it as
+// a validation function on identifier field with generic parameter set to the desired sdk.ObjectIdentifier implementation.
+func IsValidIdentifier[T sdk.ObjectIdentifier]() schema.SchemaValidateDiagFunc {
+	return func(value any, path cty.Path) diag.Diagnostics {
+		var diags diag.Diagnostics
+
+		// For now, we won't support sdk.ExternalObjectIdentifiers. The reason behind it is that the functions that parse identifiers are not
+		// able to differentiate between sdk.ExternalObjectIdentifiers and sdk.DatabaseObjectIdentifier or sdk.SchemaObjectIdentifier,
+		// because sdk.ExternalObjectIdentifiers has varying parts count (2 or 3).
+		if _, ok := any(sdk.ExternalObjectIdentifier{}).(T); ok {
```
sfc-gh-asawicki, line 161: we can replace `T sdk.ObjectIdentifier` with `[T sdk.AccountObjectIdentifier | sdk.DatabaseObjectIdentifier | sdk.SchemaObjectIdentifier | sdk.TableColumnIdentifier]` and we don't have to handle this case then

pkg/resources/helpers.go
```diff
@@ -142,3 +146,77 @@ func IsDataType() schema.SchemaValidateFunc { //nolint:staticcheck
...
+			diags = append(diags, diag.Diagnostic{
+				Severity:      diag.Error,
+				Summary:       "Invalid schema identifier type",
+				Detail:        "Identifier validation is not available for sdk.ExternalObjectIdentifier type. This is a provider error please file a report: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/new/choose",
+				AttributePath: path,
+			})
+			return diags
+		}
+
+		if stringValue, ok := value.(string); ok {
```
sfc-gh-asawicki, line 171: nit-pick:
It would be a bit more readable to inverse the check here (if not string return an error - fail fast; then we won't have the nesting here - it would look more go like - the sequence of checking different errs basically).

pkg/resources/helpers.go
```diff
@@ -142,3 +146,77 @@ func IsDataType() schema.SchemaValidateFunc { //nolint:staticcheck
...
+			id, err := helpers.DecodeSnowflakeParameterID(stringValue)
+			if err != nil {
+				diags = append(diags, diag.Diagnostic{
+					Severity: diag.Error,
+					Summary:  "Unable to parse the identifier",
+					Detail: fmt.Sprintf(
+						"Unable to parse the identifier: %s. Make sure you are using the correct form of the fully qualified name for this field: %s",
```
sfc-gh-asawicki, line 178: shouldn't we also provide the error here? it would be better for us

pkg/resources/helpers.go
```diff
@@ -142,3 +146,77 @@ func IsDataType() schema.SchemaValidateFunc { //nolint:staticcheck
...
+						getExpectedIdentifierForm(&id),
+					),
+					AttributePath: path,
+				})
+			}
+		} else {
+			diags = append(diags, diag.Diagnostic{
+				Severity:      diag.Error,
+				Summary:       "Invalid schema identifier type",
+				Detail:        fmt.Sprintf("Expected schema string type, but got: %T. This is a provider error please file a report: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/new/choose", value),
```
sfc-gh-asawicki, line 205: why do we want to consider this a provider error? This just means that it is not used on string

pkg/resources/helpers.go
```diff
@@ -142,3 +146,77 @@ func IsDataType() schema.SchemaValidateFunc { //nolint:staticcheck
...
+		return diags
+	}
+}
+
+// getExpectedIdentifierForm will choose the type either from the objectIdentifier parameter if it's present. If it's not,
+// then it will create a new identifier based on the generic type parameter T, then it will return the proper structure
+// we are expecting for the given sdk.ObjectIdentifier type.
+func getExpectedIdentifierForm[T sdk.ObjectIdentifier](objectIdentifier *T) string {
+	if objectIdentifier != nil {
+		return (*objectIdentifier).Representation()
```
sfc-gh-asawicki, line 219: I would put the logic of Representation here (check other comment).

sfc-gh-asawicki (changes requested):

pkg/resources/helpers.go
sfc-gh-jcieslak: True, didn't think of that

pkg/resources/helpers.go
sfc-gh-jcieslak: Yeah, I was thinking about it, but ultimately left it as it is. I'll change it.

pkg/resources/helpers.go
sfc-gh-jcieslak: It's a provider error because this 'if' branch means that the identifier from the schema is not a string (e.g. could be int as Terraform type). Anyway, it would later fail on helpers.DecodeSnowflakeParameterID.

pkg/resources/validators_test.go
```diff
@@ -0,0 +1,189 @@
...
+			Name:     "correct schema object identifier from generic parameter",
+			Expected: "<database_name>.<schema_name>.<name>",
+			Actual:   getExpectedIdentifierRepresentationFromGeneric[sdk.SchemaObjectIdentifier](),
+		},
+		{
+			Name:     "correct table column identifier from generic parameter",
+			Expected: "<database_name>.<schema_name>.<table_name>.<column_name>",
+			Actual:   getExpectedIdentifierRepresentationFromGeneric[sdk.TableColumnIdentifier](),
+		},
+		{
```
sfc-gh-asawicki, line 163: It would be nice to split these 4 into separate one (then we can get rid of getExpectedIdentifierRepresentationFromParam in testcase def - move it to body). ALso in body we can then test two options: ptr and no ptr one.

sfc-gh-jcieslak: Done

sfc-gh-asawicki (approved):

PR #2270, opened 2023-12-14
chore: Reference IDs of share objects in share.go error messages.

originalrkk:
Without this, we get error messages like
```
Error: error reading share err = [errors.go:17] object does not exist or not authorized
```
that are nearly impossible to debug.

## Test Plan

This PR only adds extra information to error messages.


sfc-gh-asawicki:
Hey @originalrkk. Thanks for the PR.

For me, it's fine to add such additional info. Out of curiosity: did you try running the provider with TF_LOG set to DEBUG or TRACE? You get a lot more info that way.

originalrkk:
Ah, d'oh, good idea. Still think it's useful to have these since ideally we'd only have to fall back to debug for the most pathological / confusing cases.

sfc-gh-jcieslak (approved):

sfc-gh-asawicki (approved):

PR #2288, opened 2023-12-20
chore: Add short scripts used to fetch all currently opened issues

sfc-gh-asawicki:
Add short scripts used to fetch all currently opened issues (for future reference).

### DIFF
--- a/pkg/scripts/issues/README.md
+++ b/pkg/scripts/issues/README.md
@@ -1,0 +1,11 @@
+1. To use the script, generate access token here: https://github.com/settings/tokens?type=beta.
+2. To get all open issues invoke the [first script](./gh/main.go) setting `SF_TF_SCRIPT_GH_ACCESS_TOKEN`:
+```shell
+  cd gh && SF_TF_SCRIPT_GH_ACCESS_TOKEN=<YOUR_PERSONAL_ACCESS_TOKEN> go run .
+```
+3. File `issues.json` should be generated in the `gh` directory. This is the input file for the second script.
+4. To get process the issues invoke the [second script](./file/main.go):
+```shell
+  cd file && go run .
+```
+5. File `issues.csv` should be generated in the `file` directory. This is the CSV which summarizes all the issues we have.
--- a/pkg/scripts/issues/file/main.go
+++ b/pkg/scripts/issues/file/main.go
@@ -9,6 +9,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"time"
 
 	i "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/scripts/issues"
 )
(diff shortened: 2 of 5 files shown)


pkg/scripts/issues/README.md
```diff
@@ -0,0 +1,11 @@
+1. To use the script, generate access token here: https://github.com/settings/tokens?type=beta.
+2. To get all open issues invoke the [first script](./gh/main.go) setting `SF_TF_SCRIPT_GH_ACCESS_TOKEN`:
+```shell
+  cd gh && SF_TF_SCRIPT_GH_ACCESS_TOKEN=<YOUR_PERSONAL_ACCESS_TOKEN> go run .
+```
+3. File `issues.json` should be generated in the `gh` directory. This is the input file for the second script.
```
sfc-gh-swinkler, line 6: why can't steps 2/3/4 be combined into one step? Getting and processing at same time.

pkg/scripts/issues/file/main.go
```diff
@@ -0,0 +1,155 @@
...
+func getTerraformVersion(issue i.Issue) string {
+	oldRegex := regexp.MustCompile(`\*\*Terraform Version\*\*\s*([[:ascii:]]*)\s*\*\*Describe the bug\*\*`)
+	matches := oldRegex.FindStringSubmatch(issue.Body)
+	if len(matches) == 0 {
+		return "NONE"
+	} else {
+		versionRegex := regexp.MustCompile(`v?\.?(\d+\.(\d+)(.\d+)?)`)
+		vMatches := versionRegex.FindStringSubmatch(matches[1])
+		if len(vMatches) == 0 {
+			return "NONE"
```
sfc-gh-swinkler, line 103: instead of returning None, we could return the version of Terraform that was current at the time the issue was created.

pkg/scripts/issues/file/main.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+func main() {
+	issues := loadIssues()
+	processedIssues := processIssues(issues)
+	saveCsv(processedIssues)
+}
+
+func loadIssues() []i.Issue {
+	bytes, err := os.ReadFile("../gh/issues.json")
```
sfc-gh-swinkler, line 24: is this meant to load issues coming from multiple sources? if so, it seems to be strongly dependent on the github issue interface

pkg/scripts/issues/README.md
sfc-gh-asawicki: The reason was that the data does not change between runs: The result of the first script is always the same. So this way, I was able to experiment on the static data that was faster and without locking my GH account because of the API limits they have.

pkg/scripts/issues/file/main.go
sfc-gh-asawicki: Yes, I was thinking about that: to keep a static list of releases with dates or to fetch them through API also. But this way we would get the latest possible version, which is not necessarily true. Also the regexps for extracting the version are not perfect. I guess this is the addition not needed now, because I already have a version for each issue.

pkg/scripts/issues/file/main.go
sfc-gh-asawicki: Nope, only from GH

sfc-gh-jcieslak (approved):

PR #2304, opened 2023-12-22
feat: add grant role resource

sfc-gh-swinkler:
adds snowflake_grant_role resource. Should this be renamed to snowflake_grant_account_role instead?
https://docs.snowflake.com/en/sql-reference/sql/grant-role


pkg/resources/grant_role_acceptance_test.go
```diff
@@ -0,0 +1,74 @@
...
+			"role_name":        config.StringVariable(roleName),
+			"parent_role_name": config.StringVariable(parentRoleName),
+		}
+	}
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
```
sfc-gh-jcieslak, line 31: There's no check for delete

sfc-gh-jcieslak (changes requested):

sfc-gh-asawicki:
> adds snowflake_grant_role resource. Should this be renamed to snowflake_grant_account_role instead? https://docs.snowflake.com/en/sql-reference/sql/grant-role

I think this is a valid idea. We have *_role for every other planned resource. @sfc-gh-jcieslak wdyt?

pkg/resources/grant_role.go
```diff
@@ -0,0 +1,175 @@
...
+				if err := d.Set("role_name", parts[0]); err != nil {
+					return nil, err
+				}
+				switch parts[1] {
+				case "ROLE":
+					if err := d.Set("parent_role_name", parts[2]); err != nil {
+						return nil, err
+					}
+				case "USER":
+					if err := d.Set("usere_name", parts[2]); err != nil {
```
sfc-gh-asawicki, line 68: usere_name -> user_name

sfc-gh-asawicki: It was not spotted by tests, so let's add an import step to both of the tests pls.

pkg/resources/grant_role.go
```diff
@@ -0,0 +1,175 @@
...
+		d.SetId("")
+	}
+
+	return nil
+}
+
+func DeleteGrantRole(d *schema.ResourceData, meta interface{}) error {
+	db := meta.(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+	parts := strings.Split(d.Id(), helpers.IDDelimiter)
```
sfc-gh-asawicki, line 155: we could have a safety check like in the importer func here (i.e. `if len(parts) != 3)`). But this is an optional addition.

sfc-gh-asawicki (changes requested):

examples/resources/snowflake_grant_role/import.sh
```diff
@@ -0,0 +1,2 @@
+# format is role_name (string) | grantee_object_type (ROLE|USER) | grantee_name (string)
```
sfc-gh-asawicki, line 1: Hmm, the docs were not generated for this new resource (the check did not fail btw, we should check why).

sfc-gh-asawicki (changes requested):

examples/resources/snowflake_grant_role/import.sh
sfc-gh-swinkler: ran `make docs`

pkg/resources/grant_role.go
sfc-gh-swinkler: added check

pkg/resources/grant_role_acceptance_test.go
sfc-gh-swinkler: added check destroy func

sfc-gh-jcieslak (dismissed):

sfc-gh-asawicki (dismissed):

PR #2306, opened 2023-12-22
feat: grant privileges to database role resource

sfc-gh-jcieslak:
TODO


pkg/resources/grant_privileges_to_database_role_identifier_test.go
```diff
@@ -0,0 +1,404 @@
...
+					OnAllOrFuture: &BulkOperationGrantData{
+						ObjectNamePlural: "TABLES",
+						Kind:             InSchemaBulkOperationGrantKind,
+						Schema:           sdk.Pointer(sdk.NewDatabaseObjectIdentifier("database-name", "schema-name")),
+					},
+				},
+			},
+		},
+		{
+			Name:       "validation: grant database role not enough parts",
```
sfc-gh-asawicki, line 222: Can you add tests for different parts empty? e.g.:
- empty privileges
- empty bools (e.g. with grant option)
- etc

pkg/resources/grant_privileges_to_database_role.go
```diff
@@ -0,0 +1,971 @@
...
+		Type:        schema.TypeString,
+		Optional:    true,
+		Default:     "",
+		Description: "This field should not be set and its main purpose is to achieve the functionality described by always_apply field. This is value will be flipped to the opposite value on every terraform apply, thus creating a new plan that will re-apply grants.",
+	},
+	"on_database": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ForceNew:         true,
+		Description:      "The fully qualified name of the database on which privileges will be granted. If the identifier is not fully qualified (in the form of <db_name>.≤database_role_name>), the command looks for the database role in the current database for the session. All privileges are limited to the database that contains the database role, as well as other objects in the same database.",
```
sfc-gh-asawicki, line 74: is this description right?

pkg/resources/testdata/TestAcc_GrantPrivilegesToDatabaseRole_AlwaysApply/test.tf
```diff
@@ -0,0 +1,6 @@
+resource "snowflake_grant_privileges_to_database_role" "test" {
+  database_role_name = "\"${var.database}\".\"${var.name}\""
+  all_privileges     = var.all_privileges
+  on_database        = "\"${var.database}\""
```
sfc-gh-asawicki, line 4: in every example this is with `\"`. Is it possible to just use unquoted database or not? I've checked description but I don't think it's right.

pkg/resources/grant_privileges_to_database_role.go
sfc-gh-asawicki, line 74: ≤ -> < (copy-pasted from the doc)

sfc-gh-asawicki (changes requested):

pkg/resources/grant_privileges_to_database_role.go
```diff
@@ -0,0 +1,971 @@
...
+}
+
+var grantPrivilegesOnDatabaseRoleBulkOperationSchema = map[string]*schema.Schema{
+	"object_type_plural": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: "The plural object type of the schema object on which privileges will be granted. Valid values are: ALERTS | DYNAMIC TABLES | EVENT TABLES | FILE FORMATS | FUNCTIONS | PROCEDURES | SECRETS | SEQUENCES | PIPES | MASKING POLICIES | PASSWORD POLICIES | ROW ACCESS POLICIES | SESSION POLICIES | TAGS | STAGES | STREAMS | TABLES | EXTERNAL TABLES | TASKS | VIEWS | MATERIALIZED VIEWS | NETWORK RULES | PACKAGES POLICIES | ICEBERG TABLES",
+		ValidateFunc: validation.StringInSlice([]string{
+			"ALERTS",
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, I just kept what we have in other places, but I replace them with predefined enum values.

pkg/resources/grant_privileges_to_database_role.go
```diff
@@ -0,0 +1,971 @@
...
+		},
+	},
+	"with_grant_option": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     false,
+		ForceNew:    true,
+		Description: "If specified, allows the recipient role to grant the privileges to other roles.",
+	},
+	"always_apply": {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Added a warning at the top + added a point in acceptance criteria for the task about the document.

pkg/resources/grant_privileges_to_database_role.go
```diff
@@ -0,0 +1,971 @@
...
+			Summary:  "Failed to retrieve grants",
+			Detail:   fmt.Sprintf("Id: %s\nError: %s", d.Id(), err.Error()),
+		})
+	}
+
+	var privileges []string
+
+	for _, grant := range grants {
+		// Accept only DATABASE ROLEs
+		if grant.GrantTo != sdk.ObjectTypeDatabaseRole && grant.GrantedTo != sdk.ObjectTypeDatabaseRole {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Added SNOW-999930

templates/resources/grant_privileges_to_database_role.md.tmpl
```diff
@@ -0,0 +1,93 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "{{.Name}} {{.Type}} - {{.ProviderName}}"
+subcategory: ""
+description: |-
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+---
+
+# {{.Name}} ({{.Type}})
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I agree, especially in this case, where grants are a non-trivial topic and the resource implementation is fairly complex in comparison to other ones.

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
```diff
@@ -0,0 +1,648 @@
...
+		CheckDestroy: testAccCheckDatabaseRolePrivilegesRevoked,
+		Steps: []resource.TestStep{
+			{
+				PreConfig:       func() { createDatabaseRoleOutsideTerraform(t, name) },
+				ConfigDirectory: config.TestNameDirectory(),
+				ConfigVariables: configVariables,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "database_role_name", databaseRoleName),
+					resource.TestCheckResourceAttr(resourceName, "privileges.#", "3"),
+					resource.TestCheckResourceAttr(resourceName, "privileges.0", string(sdk.AccountObjectPrivilegeCreateSchema)),
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Snowflake implicitly orders them alphabetically (though it's not documented anywhere).

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
```diff
@@ -0,0 +1,648 @@
...
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "always_apply", "false"),
+					resource.TestCheckResourceAttr(resourceName, "id", fmt.Sprintf("%s|false|false|ALL|OnDatabase|%s", databaseRoleName, databaseName)),
+				),
+			},
+			{
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I just wanted to emphasize that PreApply doesn't produce a plan, but I can remove the second one (I think it's already sort of visible by PostApplyPostRefresh check).

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-asawicki: But you've also provided them alphabetically. It would be nice to have a test when we provide them in a different order. It will also allow us to see what will happen: will the plan be different for different ordering? Will it behave correctly with import?

pkg/resources/grant_privileges_to_database_role.go
sfc-gh-jcieslak: Removed 2nd and 3rd sentences. Only the info about what goes into this field should be sufficient

pkg/resources/testdata/TestAcc_GrantPrivilegesToDatabaseRole_AlwaysApply/test.tf
sfc-gh-jcieslak: you should be able to, but I did it with quotes because then it's easier to compare identifiers in acceptance test

sfc-gh-jcieslak: I just checked that it didn't matter and works as intended, so I'll leave it without quotes in this test for variety

pkg/resources/grant_privileges_to_database_role.go
sfc-gh-jcieslak: I think you already commented on this line. Removed 2nd and 3rd line.

templates/resources/grant_privileges_to_database_role.md.tmpl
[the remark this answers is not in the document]
sfc-gh-asawicki: @sfc-gh-swinkler are you okay with such an addition?

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-asawicki: agree

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
```diff
@@ -0,0 +1,806 @@
...
+			},
+		})
+		if err != nil {
+			return err
+		}
+		var grantedPrivileges []string
+		for _, grant := range grants {
+			grantedPrivileges = append(grantedPrivileges, grant.Privilege)
+		}
+		if len(grantedPrivileges) < len(privileges) {
```
sfc-gh-asawicki, line 768: is this check legit? it just checks the count of the privileges but not the types

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
```diff
@@ -0,0 +1,806 @@
...
+		grants, err := client.Grants.Show(ctx, &sdk.ShowGrantOptions{
+			To: &sdk.ShowGrantsTo{
+				DatabaseRole: databaseRoleName,
+			},
+		})
+		if err != nil {
+			return err
+		}
+		for _, grant := range grants {
+			if !slices.Contains(privileges, grant.Privilege) && grant.Privilege != "USAGE" && grant.Privilege != "OWNERSHIP" {
```
sfc-gh-asawicki, line 740: it would be nice to log here in case of usage/ownership that we do not expect the grant to be on the list. But it's just a nice-to-have, not a blocker.

sfc-gh-asawicki (changes requested):

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Added the same test with reversed privileges

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Ok, removed

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
sfc-gh-jcieslak: not sure what you mean by that, but I think I would leave it as is. We expect USAGE and OWNERSHIP to be always there that's why we need two additional cases. USAGE is something that the user could explicitly grant, so it could be on that list (but I wouldn't see the point of that). The other thing we could say about OWNERSHIP is that it's protected on the schema level, so I don't think anyone would specify that in the privileges ([]string in the function argument).

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
sfc-gh-jcieslak: True, I'll change that

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
sfc-gh-asawicki: I understand the nature of the additional cases. I only asked for a simple log for the devs, when we hit either of them (to better know what's happening in the test). :)

sfc-gh-asawicki (approved):

pkg/resources/grant_privileges_to_database_role_acceptance_test.go
sfc-gh-jcieslak: kk, added
```go
log.Printf("Skipping check for %s privilege as its one of the privileges that are implicitly granted by Snowflake", grant.Privilege)

```

pkg/resources/testdata/TestAcc_GrantPrivilegesToDatabaseRole_OnAllSchemasInDatabase/test.tf
```diff
@@ -0,0 +1,9 @@
+resource "snowflake_grant_privileges_to_database_role" "test" {
```
sfc-gh-swinkler, line 1: thinking if the folder name should be 
`testdata/resources/grant_privileges_to_database_role/complete/on_all_schemas_in_database` or something to that effect. Might be better to have more levels of folders to keep things under control.

templates/resources/grant_privileges_to_database_role.md.tmpl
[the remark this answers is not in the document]
sfc-gh-swinkler: sounds good

docs/resources/grant_privileges_to_database_role.md
```diff
@@ -0,0 +1,306 @@
...
+
+#### OnFutureSchemasInDatabase
+`terraform import "<database_role_name>|<with_grant_option>|<always_apply>|<privileges>|OnSchema|OnFutureSchemasInDatabase|<database_name>"`
+
+### OnSchemaObject
+
+On schema object contains inner types for all options.
+
+#### OnObject
+`terraform import "<database_role_name>|<with_grant_option>|<always_apply>|<privileges>|OnSchemaObject|OnObject|<object_type>|<object_name>"`
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: To be consistent I'll add it to the examples, but If you want I can make it, so every type will have firstly a structure and then an example right under it. wdyt ?

pkg/resources/grant_privileges_to_database_role.go
```diff
@@ -0,0 +1,1031 @@
...
+		Description: "If specified, allows the recipient role to grant the privileges to other roles.",
+	},
+	"always_apply": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     false,
+		Description: "If true, the resource will always produce a “plan” and on “apply” it will re-grant defined privileges. It is supposed to be used only in “grant privileges on all X’s in database / schema Y” or “grant all privileges to X” scenarios to make sure that every new object in a given database / schema is granted by the account role and every new privilege is granted to the database role. Important note: this flag is not compliant with the Terraform assumptions of the config being eventually convergent (producing an empty plan).",
+	},
+	"always_apply_trigger": {
+		Type:        schema.TypeString,
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: At first, it was a boolean, but I thought it could be acting wrongly in some scenarios. For example, the only place where this flag could be switched is in the Read operation (because it is invoked on every tf action), but then we could end up in a situation where sometimes it triggers and sometimes doesn't (mostly because Update operation calls Read at the end). With random strings, we don't have to think about it and wonder if it will be switched correctly because it will be always a different value.

pkg/resources/grant_privileges_to_database_role.go
```diff
@@ -0,0 +1,1031 @@
...
+					ForceNew:    true,
+					Description: "The object type of the schema object on which privileges will be granted. Valid values are: ALERT | DYNAMIC TABLE | EVENT TABLE | FILE FORMAT | FUNCTION | PROCEDURE | SECRET | SEQUENCE | PIPE | MASKING POLICY | PASSWORD POLICY | ROW ACCESS POLICY | SESSION POLICY | TAG | STAGE | STREAM | TABLE | EXTERNAL TABLE | TASK | VIEW | MATERIALIZED VIEW | NETWORK RULE | PACKAGES POLICY | ICEBERG TABLE",
+					RequiredWith: []string{
+						"on_schema_object.0.object_name",
+					},
+					ConflictsWith: []string{
+						"on_schema_object.0.all",
+						"on_schema_object.0.future",
+					},
+					ValidateFunc: validation.StringInSlice([]string{
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, some of the fields may require different sets of object types, but it could be also used in the snowflake_grant_privilges_to_role. I'll extract it and use it there.

pkg/resources/testdata/TestAcc_GrantPrivilegesToDatabaseRole_OnAllSchemasInDatabase/test.tf
sfc-gh-jcieslak: Sure, added the main folder for all of the tests

sfc-gh-asawicki (approved):

Issue #2162, opened 2023-10-30
Add snowflake_password_policy_attachment resource


lyschoening:
**Is your feature request related to a problem? Please describe.**

We have some users that have more stringent password policy than others, so we'd like to be able to set password policies for individual users. 

**Describe the solution you'd like**

Add a `snowflake_password_policy_attachment` to mirror `snowflake_network_policy_attachment`, but for password policies. A `set_for_account` flag would be an optional addition and allow for deprecating `snowflake_account_password_policy_attachment`.

**Describe alternatives you've considered**

Property on `snowflake_user`:  Our users are provisioned externally, so a field on `snowflake_user` would not help.

PR #2307, opened 2023-12-25
feat: add resource snowflake_user_password_policy_attachment (#2162)

raulbonet:
# Description
ref #2162 
This PR creates the new resource `snowflake_user_password_policy_attachment`. The implementation mimics the one of its brother, `snowflake_account_password_policy_attachment`


sfc-gh-asawicki:
Hey @raulbonet. Thanks for the PR.

To merge it, we require a few more things:
- (easy) generate the documentation of the new resource (run `make dev-setup && make docs`);
- (medium) add test cases or steps to the prepared test which shows update and delete behavior;
- (medium) validation in tests that the password policy is assigned to the user (consult `getRowAccessPolicyFor` and `policyReference` in `helpers_test.go`);
- (medium) implement read and import methods.

The reason for that is that publishing a partially implemented/not thoroughly tested resource will result in more issues being created. We want to avoid that.

Would you be willing to introduce these changes? Alternatively, we can merge it as a separate branch and add the missing parts ourselves (but probably not earlier than Feb-Mar).

cc: @sfc-gh-swinkler @sfc-gh-jcieslak

raulbonet:
Hello @sfc-gh-asawicki 

Thanks for coming back to me. About your comments, I am happy to do all of this myself, but I want to make sure we are in the same page. As I was mentioning in the OP:

1. The "brother" of this resource, `snowflake_account_password_policy_attachment`, does not implement (properly) the `Read()` method either (it only reads _if the network policy exists, but not if it is attached_). Is it a bug? 

2. I think there are several resources that do not implement the `Read` functionality (like a lot of the `grants` I believe). I thought precisely that you were merging stuff in iterations, but if it is not the case from now on, happy to fully develop it. 

3. The `Read()` method is not implemented because the new `sdk` lacks the implementation to read from `POLICY REFERENCES` (I think it is possibly to do it with the interfaces in `snowflake`, but I guess from the code that you are deprecating this). I am happy to implement it myself, but let's make sure we are in the same page:

    3.1. You would like me to create a a new interface in the `sdk`, let's say `policy_references.go`, which among other things, method mimics this: `https://docs.snowflake.com/en/sql-reference/functions/policy_references` ?

    3.2. If so, would you like to:
        a) Create one single interface for all kinds of policy references (masking, session, password, row access)
        b) Create a specific interface for `password` (so, `policy_references_password.go`)

sfc-gh-asawicki:
Hey @raulbonet. That's great to hear that! I've included my answers below.

> 1. The "brother" of this resource, snowflake_account_password_policy_attachment, does not implement (properly) the Read() method either (it only reads if the network policy exists, but not if it is attached). Is it a bug?

Maybe not a bug, but an imperfection. :) It should be implemented too, but this is low on our roadmap. As you may have noticed, the team changed in the past few months. The resource was introduced earlier in a rushed way. We follow a bit different development cycle now. In other words, we try to avoid introducing missing/untested functionalities and are slowly correcting the buggy/incomplete ones.

You can add it too, if you want. But please do it as a separate PR (we prefer PRs targeting one issue only).

> 2. I think there are several resources that do not implement the Read functionality (like a lot of the grants I believe). I thought precisely that you were merging stuff in iterations, but if it is not the case from now on, happy to fully develop it.

We are merging stuff in iterations, but only in certain cases. One example would be when we introduce a new resource in the SDK. It does not change the provider's behavior and can be later used in a separate PR to migrate the resource from the old snowflake one to the new SDK client.

Here, the situation is different. An entirely new resource would be introduced, so people would start using it. And that may result in more issues being reported. We want to limit possible problems by correctly preparing the resource.

> 3. The Read() method is not implemented because the new sdk lacks the implementation to read from POLICY REFERENCES (I think it is possibly to do it with the interfaces in snowflake, but I guess from the code that you are deprecating this). I am happy to implement it myself, but let's make sure we are in the same page:

It is not implemented in the SDK but was already used for testing. It is here with a TODO comment to make it a not-test-only resource: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/be01d5fdab4f2d31db9c4c849b349e657c0352c8/pkg/sdk/testint/helpers_test.go#L773.

> 3.1. You would like me to create a a new interface in the sdk, let's say policy_references.go, which among other things, method mimics this: https://docs.snowflake.com/en/sql-reference/functions/policy_references ?

Yes. But let's limit it just to POLICY REFERENCES (the ENTITY syntax; I think we don't need the POLICY NAME one).

> 3.2. If so, would you like to:
a) Create one single interface for all kinds of policy references (masking, session, password, row access)
b) Create a specific interface for password (so, policy_references_password.go)

For now, you can go with 3.2.a -> e.g. PolicyReferences#GetForEntity. You can either reuse the existing implementation (we have one like that for system functions: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/be01d5fdab4f2d31db9c4c849b349e657c0352c8/pkg/sdk/system_functions.go#L22) but preferably we would like to stick with the syntax using opts and request (like `CreateDatabaseRoleRequest` and `createDatabaseRoleOptions`). We will accept either way.

We tend to add unit tests (like `TestDatabaseRoleCreate`) and validations (like `func (opts *createDatabaseRoleOptions) validate()`). Also, we add integration tests (like in `database_role_integration_test.go`) that validate the SDK in separation from Terraform (one day, we will extract the SDK as a separate project).

If you have any more questions (or more questions will appear during the implementation), feel free to ask them!

And so you know: we are aware that we lack clear contributing guidelines. We are currently working on transparency, so we should have an updated contributing guide ready in the upcoming weeks.

raulbonet:
Hello @sfc-gh-asawicki !

I have made the changes we discussed, sorry for the delay. Can you have a look? Comments on my side:

1. I had to change the resource schema: before, it used the `password_policy` (fully qualified name), now it uses `password_policy_database`, `password_policy_schema`, `password_policy_name`. 

    Problem is:

    When Terraform detects a an update (not a replacement) in the resource `password_policy`, it is not able to detect that our new resource (`user_password_policy_attachment`) depends on it and that it also needs to be updated/replaced. My guess is that `qualified name` is a **computed field** and maybe there is a bug in Terraform or Terraform behaves differently with these fields, I don't know, can't spend more time on it.

    `terraform refresh` is able to detect the change though, but still the tests were failing for this specific scenario. 

      Is it a problem for you guys to leave it like this?


2. About delimiters in the SDK:
    Functions like Policy References `https://docs.snowflake.com/en/sql-reference/functions/policy_references` enclose the values of the arguments in single quotes, but still you need the double quotes for the identifier (like a User if it contains lowercase letters). I tried something like `ddl:"single_quotes","double_quotes"` but it did not work. I solved this by encapsulating this logic outside the SDK (so, when calling `client.PolicyReferences.GetForEntity`, I had to explicitly call the method `FullyQualifiedName`  (`userName.FullyQualifiedName()`). I think however that this logic should be encapsulated in the SDK maybe?

pkg/sdk/policy_references.go
```diff
@@ -0,0 +1,112 @@
...
+	table                   *bool                    `ddl:"keyword" sql:"TABLE"`
+	policyReferenceFunction *policyReferenceFunction `ddl:"list,parentheses,no_comma"`
+}
+
+type policyReferenceFunction struct {
+	functionFullyQualifiedName *bool                             `ddl:"keyword" sql:"SNOWFLAKE.INFORMATION_SCHEMA.POLICY_REFERENCES"`
+	arguments                  *policyReferenceFunctionArguments `ddl:"list,parentheses"`
+}
+type policyReferenceFunctionArguments struct {
+	refEntityName   *string `ddl:"parameter,single_quotes,arrow_equals" sql:"ref_entity_name"`
```
sfc-gh-asawicki, line 29: I also wonder if refEntityName should not be ObjectIdentifier and if we should not run fully qualified name on it, due to this part of the docs:

```
REF_ENTITY_NAME => 'string'
The name of the object, such as the table name, view name, external table name, or username, on which the policy is set.

The entire object name must be enclosed in single quotes.

If the object name is case-sensitive or includes any special characters or spaces, double quotes are required to process the case/characters. The double quotes must be enclosed within the single quotes (i.e. '"<table_name>"').
```

It was also implemented like that for test usage (check `getRowAccessPolicyFor` func).

pkg/resources/user_password_policy_attachment.go
```diff
@@ -0,0 +1,169 @@
...
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: "Non-qualified name of the password policy",
+	},
+}
+
+func UserPasswordPolicyAttachment() *schema.Resource {
+	return &schema.Resource{
+		Description: "Specifies the password policy to use for the current account. To set the password policy of a different account, use a provider alias.",
```
sfc-gh-asawicki, line 43: This description was copy-pasted from account password policy resource, please update it to describe that it is user.

docs/resources/user_password_policy_attachment.md
```diff
@@ -0,0 +1,27 @@
...
+page_title: "snowflake_user_password_policy_attachment Resource - terraform-provider-snowflake"
+subcategory: ""
+description: |-
+  Specifies the password policy to use for the current account. To set the password policy of a different account, use a provider alias.
+---
+
+# snowflake_user_password_policy_attachment (Resource)
+
+Specifies the password policy to use for the current account. To set the password policy of a different account, use a provider alias.
+
```
sfc-gh-asawicki, line 12: Please provide example usage. You can do it by creating a directory s`nowflake_user_password_policy_attachment` inside `examples/resources` (you can check the one for account password policy - `snowflake_account_password_policy_attachment`). You will have to re-run `make docs` or `make pre-push` after that.

sfc-gh-asawicki: The same with import example.

pkg/resources/user_password_policy_attachment.go
```diff
@@ -0,0 +1,169 @@
...
+
+		Schema: userPasswordPolicyAttachmentSchema,
+
+		Importer: &schema.ResourceImporter{
+			StateContext: func(ctx context.Context, d *schema.ResourceData, m interface{}) ([]*schema.ResourceData, error) {
+				parts := strings.Split(d.Id(), helpers.IDDelimiter)
+				if len(parts) != 4 {
+					return nil, fmt.Errorf("id should be in the format 'database|schema|password_policy|user_name|roles', but I got '%s'", d.Id())
+				}
+				passwordPolicyDatabase := sdk.NewAccountIdentifierFromFullyQualifiedName(parts[0])
```
sfc-gh-asawicki, line 57: I don't get why these identifiers are created like that here. Is there a reason for customizing the importer instead of just using ImportStatePassthroughContext?

pkg/resources/user_password_policy_attachment.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+func CreateUserPasswordPolicyAttachment(d *schema.ResourceData, meta interface{}) error {
+	db := meta.(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+	ctx := context.Background()
+
+	userName := sdk.NewAccountObjectIdentifierFromFullyQualifiedName(d.Get("user_name").(string))
+	passwordPolicyDatabase := sdk.NewAccountIdentifierFromFullyQualifiedName(d.Get("password_policy_database").(string))
```
sfc-gh-asawicki, line 85: if using three separate parts, you could just do sdk.NewSchemaObjectIdcentifier(db, schema, name)

sfc-gh-asawicki:
Hey @raulbonet. I am sorry for the delay on our part; these last two weeks were a bit crazy. I finally had a moment to look at your PR.

First, I would like to thank you one more time for your contribution and the time you spent on this change! We really appreciate it.

Now, to answer the questions:
ad 1
The reason is probably that the fully qualified name does not change if you don't change the db, schema, or name in the password policy. The underlying resource may be changed, but from the point of view, the reference is still the same, so there is no need to do anything. Look at the [replace triggered by](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#replace_triggered_by); this may be helpful. We are not doing such a split in the rest of the resources, so we should be consistent and have this as one field.

ad 2
I guess it's not easily achievable right now because sql_builder's `sqlIdentifierClause` does not accept quote modifiers (which should be relatively easy to achieve though). These builder tags are only intuitive when you are used to using them. That's why we have many unit tests verifying the generated SQL statements. I think your current workaround is acceptable.

I have browsed the implementation and left a few minor comments. Some things are blockers, though:
- no unit tests for the new PolicyReferences interface (we test the SQL statements  generation and validations there)
- no integration tests for the new PolicyReferences interface (we test compatibility with SF in them).

Because there are multiple changes to do and these tests have to be added, I see two options:
- you can add them yourself if you have time
- we can hijack the change and introduce the requested modifications ourselves (it will be implemented faster, and we should be able to squeeze it in next week).

Which option works better for you?

sfc-gh-asawicki (changes requested):

raulbonet:
Hello @sfc-gh-asawicki
Thank you for your thorough review. Before addressing your issues, quick question: is it possible that you now introduced a "Secondary Account" feature / requirement that is making my Integration tests fail? I was confused because I was quite sure my tests were successful some weeks before.

It's asking for credentials in the `./snowflake/config` directory (I have the credentials of the main account as env vars)

sfc-gh-asawicki:
Yes, we fixed the tests that must be run on multiple accounts (e.g. to test share creation). You can comment out the lines in `setup_test.go` and run the tests that require only a single account (that's majority of the tests).

pkg/resources/user_password_policy_attachment.go
sfc-gh-jcieslak: Yeah, then the Terraform will just call the Read function (ReadUserPasswordPolicyAttachment in this case) to import state.

pkg/resources/user_password_policy_attachment.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	if err := d.Set("password_policy_schema", passwordPolicySchema.Name()); err != nil {
+		return err
+	}
+	if err := d.Set("password_policy_name", passwordPolicyName.Name()); err != nil {
+		return err
+	}
+	if err := d.Set("user_name", helpers.EncodeSnowflakeID(userName)); err != nil {
+		return err
+	}
```
sfc-gh-jcieslak, line 109: You can remove all of the `d.Set` functions. They should be called in the Read function that is called as the last step of create, here: `return ReadUserPasswordPolicyAttachment(d, meta)`

pkg/resources/user_password_policy_attachment.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := d.Set("password_policy_database", sdk.NewAccountIdentifierFromFullyQualifiedName(policyReferences[0].PolicyDb).Name()); err != nil {
+		return err
+	}
+	if err := d.Set("password_policy_schema", sdk.NewAccountIdentifierFromFullyQualifiedName(policyReferences[0].PolicySchema).Name()); err != nil {
+		return err
+	}
+	if err := d.Set("password_policy_name", sdk.NewAccountIdentifierFromFullyQualifiedName(policyReferences[0].PolicyName).Name()); err != nil {
+		return err
+	}
```
sfc-gh-jcieslak, line 147: here you should add `d.Set("user_name")` to detect Snowflake changes. This information could be probably fetched with https://docs.snowflake.com/en/user-guide/admin-user-management#considerations (Tracking password policy usage). It's not in the client, so:
- leave the comment and we'll do it
- use raw SQL (client.UsafeExec or something like that) in place to get the user_name and we'll move it to the client later 
- or you can add the `ReferencesFunctions` interface to the client struct (next to the system-defined functions) and use the raw SQL there to retrieve the needed username. I guess `ConvertionFunctions` are pretty good reference.
```go
// client.go
type Client struct {
	config         *gosnowflake.Config
	db             *sqlx.DB
	sessionID      string
	accountLocator string
	dryRun         bool
	traceLogs      []string

	// System-Defined Functions
	ContextFunctions     ContextFunctions
        <here>
        // ...
}

File: pkg/resources/user_password_policy_attachment.go [hunk 8]
```diff
@@ -0,0 +1,169 @@
...
+	d.SetId(fmt.Sprintf(`%s|%s`, helpers.EncodeSnowflakeID(passwordPolicy), helpers.EncodeSnowflakeID(userName)))
+
+	return ReadUserPasswordPolicyAttachment(d, meta)
+}
+
+func ReadUserPasswordPolicyAttachment(d *schema.ResourceData, meta interface{}) error {
+	db := meta.(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+	ctx := context.Background()
+	userName := sdk.NewAccountObjectIdentifierFromFullyQualifiedName(d.Get("user_name").(string))
```
sfc-gh-jcieslak, line 119: So, to remove custom import and use ImportStatePassthroughContext you need to parse the `d.ID()` to get password policy identifier and user name. `d.Get()` cannot be used in Read operation.

pkg/resources/user_password_policy_attachment.go
```diff
@@ -0,0 +1,169 @@
...
+
+	err := client.Users.Alter(ctx, userName, &sdk.AlterUserOptions{
+		Unset: &sdk.UserUnset{
+			PasswordPolicy: sdk.Bool(true),
+		},
+	})
+	if err != nil {
+		return err
+	}
+
```
sfc-gh-jcieslak, line 167: after you drop the resource, you have to set the id to the empty string `d.SetId("")`, so the Terraform will know it was successfully removed.

pkg/resources/user_password_policy_attachment_acceptance_test.go
```diff
@@ -0,0 +1,100 @@
...
+			{
+				ResourceName:      "snowflake_user_password_policy_attachment.ppa",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func testAccCheckYourResourceDestroy(s *terraform.State) error {
```
sfc-gh-jcieslak, line 53: Change `Your` to more specific `PasswordPolicyAttachment`

pkg/resources/user_password_policy_attachment.go
sfc-gh-jcieslak: Oh ok, I saw you already did it, so you can use the function here to get the username

sfc-gh-jcieslak (changes requested):

docs/resources/user_password_policy_attachment.md
raulbonet: Done!

pkg/resources/user_password_policy_attachment.go
sfc-gh-jcieslak: I meant that if we use `ImportStatePassthroughContext` (so Read as Import), then Read cannot use `d.Get()`, because Import functionality wouldn't work. Using "Read as Import" requires parsing `d.ID()` to get all needed information for the resource and not rely on the `d.Get()` parameters which are not set yet. If we go with the custom import it can stay as is, but if we go with `ImportStatePassthroughContext` as the Import option then `userName` should be parsed from the `d.ID()`. Reference: https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2307#discussion_r1476262478

pkg/resources/user_password_policy_attachment.go
sfc-gh-jcieslak: Ok, I see. We can probably leave it as is then. cc: @sfc-gh-asawicki

pkg/resources/user_password_policy_attachment.go
sfc-gh-asawicki: The majority of the resources use the `ImportStatePassthroughContext` approach. All the ones using the complicated import are the grant ones, but they have their reasons: they are more complex resources, where more data needs to be provided to create a proper import. I prefer consistency: to have an `ImportStatePassthroughContext` here.

sfc-gh-asawicki:
@raulbonet, thanks for all the changes you made after our reviews! It's very helpful for us to receive help with the implementation.

Before we proceed, please answer my comment: https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2307#issuecomment-1924251824, the part about unit and integration tests. We can write or help you write them, but we need a consensus here because we can only merge the change with them.

raulbonet:
Hello @sfc-gh-asawicki 
Sorry, yesterday night I made the changes and did not answer your message.

1. About the computed field issue: I think I will create a parallel branch and re-test the behaviour because I think that the behaviour was incorrect, but I might be wrong: even when you changed the database or schema of the Password Policy (which triggers a replacement of the PP), this did not in turn trigger a replacement of the attachment, but I am going to double check.

2. About unit test: I can add them.

3. About integration tests: we would love to collaborate but this is a bit more delicate: unfortunately, I am seeing that the integration tests (and acceptance tests) might leave some "resource traces" in our Snowflake account (warehouses, users, etc.), particularly if they fail, so we do not feel comfortable running these in our Snowflake account. We requested an extra Snowflake Account to our Account manager but we did not get an answer; until we have this sorted out we prefer not to run these. Would you mind chiming in for the integration tests?

pkg/resources/user_password_policy_attachment.go
raulbonet: No prob! Changed!

pkg/resources/user_password_policy_attachment.go
raulbonet: Ah, I see, now I understand your comment. Thanks for the clarification :) Everything changed :)

pkg/sdk/policy_references_test.go
raulbonet, line 1: I added unit tests. Can somebody have a look?

pkg/sdk/policy_references_test.go
```diff
@@ -0,0 +1,143 @@
...
+				policyReferenceFunction: &policyReferenceFunction{
+					functionFullyQualifiedName: Bool(true),
+					arguments: &policyReferenceFunctionArguments{
+						refEntityName:   []ObjectIdentifier{userName},
+						refEntityDomain: String("user"),
+					},
+				},
+			},
+		}
+		assertOptsValidAndSQLEquals(t, opts, "SELECT * FROM TABLE (SNOWFLAKE.INFORMATION_SCHEMA.POLICY_REFERENCES (ref_entity_name => '%s', ref_entity_domain => 'user'))", strings.ReplaceAll(userName.FullyQualifiedName(), `"`, `\"`))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Seems fine, we will check the statement execution with the integration tests and change the sqlbuilder behavior or opts config if needed.

sfc-gh-asawicki (dismissed):

pkg/sdk/policy_references_test.go
sfc-gh-asawicki: They're 👍

sfc-gh-jcieslak (approved):

PR #2311, opened 2024-01-02
fix: Revert indirect dependencies

sfc-gh-asawicki:
Based on the suggestion from gosnowflake driver team, we revert aws-sdk-go-v2 to v1.17.7. We revert other dependencies too.


sfc-gh-jcieslak (approved):

sfc-gh-asawicki:
Rejecting for now

PR #2337, opened 2024-01-11
feat: add-call-with to sdk

sfc-gh-swinkler:
adds call and call-with to procedures. 

https://docs.snowflake.com/en/sql-reference/sql/call
https://docs.snowflake.com/en/sql-reference/sql/call-with

this was tricky to do, so was not added to original PR. will be useful in the future, for making a data source that could invoke the call function. similar to the AWS invoke lamba data source. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_invocation


pkg/sdk/procedures_def.go
```diff
@@ -319,6 +325,211 @@ var ProceduresDef = g.NewInterface(
...
+			"Returns",
+			procedureReturns,
+			g.KeywordOptions().SQL("RETURNS").Required(),
+		).
+		SQL("LANGUAGE SCALA").
+		TextAssignment("RUNTIME_VERSION", g.ParameterOptions().SingleQuotes().Required()).
+		ListQueryStructField(
+			"Packages",
+			procedurePackage,
+			g.ParameterOptions().MustParentheses().SQL("PACKAGES").Required(),
```
sfc-gh-asawicki, line 406: The reviewdog seems to be right, there is no such method on ParameterTransformer

pkg/sdk/procedures_impl_gen.go
```diff
@@ -97,7 +127,11 @@ func (r *CreateForJavaProcedureRequest) toOpts() *CreateForJavaProcedureOptions
 	if r.Arguments != nil {
 		s := make([]ProcedureArgument, len(r.Arguments))
 		for i, v := range r.Arguments {
-			s[i] = ProcedureArgument(v)
+			s[i] = ProcedureArgument{
```
sfc-gh-asawicki, line 130: Please change this and the lower ones back to `ProcedureArgument(v)` syntax

pkg/sdk/procedures_gen.go
```diff
@@ -247,3 +253,116 @@ type ProcedureDetail struct {
...
+// CallProcedureOptions is based on https://docs.snowflake.com/en/sql-reference/sql/call.
+type CallProcedureOptions struct {
+	call              bool                   `ddl:"static" sql:"CALL"`
+	name              SchemaObjectIdentifier `ddl:"identifier"`
+	CallArguments     []string               `ddl:"keyword,must_parentheses"`
+	ScriptingVariable *string                `ddl:"parameter,no_quotes,no_equals" sql:"INTO"`
+}
+
+// CreateAndCallForJavaProcedureOptions is based on https://docs.snowflake.com/en/sql-reference/sql/call-with#java-and-scala.
+type CreateAndCallForJavaProcedureOptions struct {
```
sfc-gh-asawicki, line 266: Looking at the CreateAndCallForX I started to wander: these are all really similar to the createForX. Should we consider merging/reusing them? I think it would be beneficial but also complicates them... I would just like to have a short discussion about this, it's not a blocking comment, eventually some TODO for the future.

pkg/sdk/procedures_validations_gen.go
```diff
@@ -172,3 +178,128 @@ func (opts *DescribeProcedureOptions) validate() error {
...
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	if !valueSet(opts.RuntimeVersion) {
+		errs = append(errs, errNotSet("CreateAndCallForJavaProcedureOptions", "RuntimeVersion"))
+	}
+	if !valueSet(opts.Handler) {
+		errs = append(errs, errNotSet("CreateAndCallForJavaProcedureOptions", "Handler"))
+	}
```
sfc-gh-jcieslak, line 203: packages are also required

pkg/sdk/procedures_validations_gen.go
```diff
@@ -172,3 +178,128 @@ func (opts *DescribeProcedureOptions) validate() error {
...
+	return JoinErrors(errs...)
+}
+
+func (opts *CreateAndCallForScalaProcedureOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	if !valueSet(opts.RuntimeVersion) {
```
sfc-gh-jcieslak, line 223: same here, packages are also required

pkg/sdk/procedures_validations_gen.go
```diff
@@ -172,3 +178,128 @@ func (opts *DescribeProcedureOptions) validate() error {
...
+	var errs []error
+	if !valueSet(opts.ProcedureDefinition) {
+		errs = append(errs, errNotSet("CreateAndCallForJavaScriptProcedureOptions", "ProcedureDefinition"))
+	}
+	if !ValidObjectIdentifier(opts.ProcedureName) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	if !ValidObjectIdentifier(opts.Name) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
```
sfc-gh-jcieslak, line 256: return should be set

pkg/sdk/testint/procedures_integration_test.go
```diff
@@ -555,3 +556,410 @@ func TestInt_OtherProcedureFunctions(t *testing.T) {
 		require.NoError(t, err)
 	})
 }
+
+func TestInt_CallProcedure(t *testing.T) {
```
sfc-gh-asawicki, line 560: We have almost the same setups for the procedures of every type. Shouldn't we just run call after creation in the tests above (to prevent setup dulications)?

sfc-gh-asawicki (changes requested):

pkg/sdk/procedures_validations_gen.go
```diff
@@ -172,3 +178,128 @@ func (opts *DescribeProcedureOptions) validate() error {
...
+func (opts *CreateAndCallForJavaScriptProcedureOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	if !valueSet(opts.ProcedureDefinition) {
+		errs = append(errs, errNotSet("CreateAndCallForJavaScriptProcedureOptions", "ProcedureDefinition"))
+	}
+	if !ValidObjectIdentifier(opts.ProcedureName) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
```
sfc-gh-jcieslak, line 252: For invalid identifiers other than "name" I would suggest using errInvalidIdentifier() function. I know this is generated, but it can only generate it if you specify validation g.ValidIdentifierIfSet (or something like that)

pkg/sdk/procedures_validations_gen.go
```diff
@@ -172,3 +178,128 @@ func (opts *DescribeProcedureOptions) validate() error {
...
+	}
+	var errs []error
+	if !valueSet(opts.RuntimeVersion) {
+		errs = append(errs, errNotSet("CreateAndCallForPythonProcedureOptions", "RuntimeVersion"))
+	}
+	if !valueSet(opts.Handler) {
+		errs = append(errs, errNotSet("CreateAndCallForPythonProcedureOptions", "Handler"))
+	}
+	if !ValidObjectIdentifier(opts.ProcedureName) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
```
sfc-gh-jcieslak, line 272: same here, errInvalidIndentifier()

pkg/sdk/procedures_validations_gen.go
```diff
@@ -172,3 +178,128 @@ func (opts *DescribeProcedureOptions) validate() error {
...
+	}
+	if !ValidObjectIdentifier(opts.Name) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	return JoinErrors(errs...)
+}
+
+func (opts *CreateAndCallForPythonProcedureOptions) validate() error {
```
sfc-gh-jcieslak, line 260: procedure definition and packages are required and should be set

pkg/sdk/procedures_validations_gen.go
```diff
@@ -172,3 +178,128 @@ func (opts *DescribeProcedureOptions) validate() error {
...
+func (opts *CreateAndCallForSQLProcedureOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	if !valueSet(opts.ProcedureDefinition) {
+		errs = append(errs, errNotSet("CreateAndCallForSQLProcedureOptions", "ProcedureDefinition"))
+	}
+	if !ValidObjectIdentifier(opts.ProcedureName) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
```
sfc-gh-jcieslak, line 294: should append errInvalidIdentifier()

pkg/sdk/procedures_def.go
sfc-gh-swinkler: needed to update branch, this was added in a previous PR

pkg/sdk/procedures_gen.go
sfc-gh-swinkler: thats a good point. i think in this case the names are clear enough to make it easy to understand, and being easy to understand is more important than having shorter code. there could be a way to merge them, but I am not sure it is worth the trouble.

pkg/sdk/procedures_validations_gen.go
sfc-gh-swinkler: fixed

pkg/sdk/procedures_validations_gen.go
sfc-gh-swinkler: fixed

pkg/sdk/procedures_validations_gen.go
sfc-gh-swinkler: fixed

pkg/sdk/testint/procedures_integration_test.go
sfc-gh-swinkler: i am not sure it is worth the trouble. the creation tests are already complicated enough, no need to put the call code in there as well.

pkg/sdk/procedures_gen.go
sfc-gh-asawicki: LEt's leave it for the future then

pkg/sdk/testint/procedures_integration_test.go
sfc-gh-asawicki: It would allow us to remove the duplicated code also and in this case I think it could be worth it. Especially because the use case we will have from Terraform side would be probable create (in resource) + call (in datasource), not createAndCall (from just one object).

But I don't want to block this change because there are other, more important topics, and this is mostly cosmetics. So closing.

sfc-gh-asawicki (approved):
I approve but please wait for @sfc-gh-jcieslak re-review also.

pkg/sdk/procedures_validations_gen.go
```diff
@@ -29,9 +29,6 @@ func (opts *CreateForJavaProcedureOptions) validate() error {
 	if !valueSet(opts.Handler) {
 		errs = append(errs, errNotSet("CreateForJavaProcedureOptions", "Handler"))
 	}
-	if !valueSet(opts.Packages) {
```
sfc-gh-jcieslak, line 26: don't remove packages, they're required - https://docs.snowflake.com/en/sql-reference/sql/create-procedure?utm_source=snowscope&utm_medium=serp&utm_term=create+procedure#java
Same for other creates

pkg/sdk/procedures_validations_gen.go
sfc-gh-swinkler: added validation

pkg/sdk/procedures_validations_gen.go
sfc-gh-swinkler: added validation

pkg/sdk/procedures_validations_gen.go
sfc-gh-swinkler: added validation

sfc-gh-jcieslak (approved):

PR #2350, opened 2024-01-15
feat: add application to sdk

sfc-gh-swinkler:
adds snowflake application to sdk 

https://docs.snowflake.com/en/sql-reference/sql/create-application


pkg/sdk/applications_def.go
```diff
@@ -0,0 +1,142 @@
+package sdk
+
+import g "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/poc/generator"
+
+//go:generate go run ./poc/main.go
+
+/*
+ * 	todo: add definition for `CREATE APPLICATION <name> FROM LISTING <listing_name> [ COMMENT = '<string_literal>' ] [ WITH TAG ( <tag_name> = '<tag_value>' [ , ... ] ) ]`
```
sfc-gh-asawicki, line 8: Why we are leaving it for later? Can we add it to the SNOW-933289 epic as a subtask if we don't need it now?

pkg/sdk/applications_validations_gen.go
```diff
@@ -0,0 +1,77 @@
...
+
+var (
+	_ validatable = new(CreateApplicationOptions)
+	_ validatable = new(DropApplicationOptions)
+	_ validatable = new(AlterApplicationOptions)
+	_ validatable = new(ShowApplicationOptions)
+	_ validatable = new(DescribeApplicationOptions)
+)
+
+func (opts *CreateApplicationOptions) validate() error {
```
sfc-gh-asawicki, line 11: according to the docs DEBUG cannot be set if not using the `USING` syntax (first one here: https://docs.snowflake.com/en/sql-reference/sql/create-application#syntax). Is this the error in the docs or should it be checked?

pkg/sdk/applications_def.go
```diff
@@ -0,0 +1,142 @@
...
+		Alter().
+		SQL("APPLICATION").
+		IfExists().
+		Name().
+		OptionalQueryStructField(
+			"Set",
+			applicationSet,
+			g.KeywordOptions().SQL("SET"),
+		).
+		OptionalSQL("UNSET COMMENT").
```
sfc-gh-asawicki, line 74: isn't it the same like with UNSET on application package (that the commas are missing from the syntax description and it could be wrapped in one object as set?)

pkg/sdk/applications_impl_gen.go
```diff
@@ -0,0 +1,180 @@
...
+				Patch:   r.UpgradeVersion.VersionAndPatch.Patch,
+			}
+		}
+	}
+	if r.UnsetReferences != nil {
+		opts.UnsetReferences = &ApplicationReferences{}
+		if r.UnsetReferences.References != nil {
+			s := make([]ApplicationReference, len(r.UnsetReferences.References))
+			for i, v := range r.UnsetReferences.References {
+				s[i] = ApplicationReference{
```
sfc-gh-asawicki, line 128: +1 for reviewdog comment here

pkg/sdk/testint/applications_gen_integration_test.go
```diff
@@ -0,0 +1,345 @@
...
+	t.Run("show application: with like", func(t *testing.T) {
+		applicationPackageName, version, patch := "hello_snowflake_package_test", "V001", 0
+		_, e := createApplicationHandle(t, applicationPackageName, version, patch, false, true, false)
+		packages, err := client.Applications.Show(ctx, sdk.NewShowApplicationRequest().WithLike(&sdk.Like{Pattern: &e.Name}))
+		require.NoError(t, err)
+		require.Equal(t, 1, len(packages))
+		require.Equal(t, *e, packages[0])
+	})
+
+	t.Run("alter application: set", func(t *testing.T) {
```
sfc-gh-asawicki, line 221: what about `SHARE_EVENTS_WITH_PROVIDER`? This one I would like to see in the tests, because it's not supported in the CREATE.

sfc-gh-asawicki (changes requested):

pkg/sdk/testint/applications_gen_integration_test.go
```diff
@@ -0,0 +1,345 @@
...
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/internal/random"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+/*
+ * todo: add integration test for `ALTER APPLICATION <name> UPGRADE`
```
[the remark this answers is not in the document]
sfc-gh-swinkler: this only works for upgrading applications which were created using listings, and we cannot create from listing at this time. there is already a ticket for this here: SNOW-1015095

pkg/sdk/applications_def.go
sfc-gh-swinkler: made Unset its own struct

pkg/sdk/applications_def.go
```diff
@@ -0,0 +1,142 @@
...
+			g.KeywordOptions().SQL("UPGRADE USING"),
+		).
+		OptionalQueryStructField(
+			"UnsetReferences",
+			applicationReferences,
+			g.KeywordOptions().SQL("UNSET REFERENCES"),
+		).
+		OptionalSetTags().
+		OptionalUnsetTags().
+		WithValidation(g.ValidIdentifier, "name").
```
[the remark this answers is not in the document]
sfc-gh-swinkler: added validation

pkg/sdk/applications_gen.go
```diff
@@ -0,0 +1,136 @@
...
+	"database/sql"
+)
+
+type Applications interface {
+	Create(ctx context.Context, request *CreateApplicationRequest) error
+	Drop(ctx context.Context, request *DropApplicationRequest) error
+	Alter(ctx context.Context, request *AlterApplicationRequest) error
+	Show(ctx context.Context, request *ShowApplicationRequest) ([]Application, error)
+	ShowByID(ctx context.Context, id AccountObjectIdentifier) (*Application, error)
+	Describe(ctx context.Context, id AccountObjectIdentifier) ([]ApplicationDetail, error)
```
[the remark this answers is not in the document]
sfc-gh-swinkler: renamed to applicationproperty

pkg/sdk/applications_validations_gen.go
sfc-gh-swinkler: added validation

pkg/sdk/testint/applications_gen_integration_test.go
```diff
@@ -0,0 +1,345 @@
...
+		return func() {
+			err := client.Applications.Drop(ctx, sdk.NewDropApplicationRequest(id))
+			if errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) {
+				return
+			}
+			require.NoError(t, err)
+		}
+	}
+
+	putOnStageHandle := func(t *testing.T, id sdk.SchemaObjectIdentifier, name string, content string) {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: created new helper function "putOnStageWithContent" for code reusue

pkg/sdk/testint/applications_gen_integration_test.go
```diff
@@ -0,0 +1,345 @@
...
+				{
+					Name:  tagTest.ID(),
+					Value: "v1",
+				},
+			})
+		err := client.Applications.Create(ctx, request)
+		require.NoError(t, err)
+		t.Cleanup(cleanupApplicationHandle(id))
+
+		e, err := client.Applications.ShowByID(ctx, id)
```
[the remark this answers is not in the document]
sfc-gh-swinkler: fixed

pkg/sdk/testint/applications_gen_integration_test.go
```diff
@@ -0,0 +1,345 @@
...
+
+		setTags := []sdk.TagAssociation{
+			{
+				Name:  tagTest.ID(),
+				Value: "v1",
+			},
+		}
+		err := client.Applications.Alter(ctx, sdk.NewAlterApplicationRequest(id).WithSetTags(setTags))
+		require.NoError(t, err)
+		assertApplication(t, id, applicationPackageName, version, patch)
```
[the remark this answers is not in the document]
sfc-gh-swinkler: tags don't currently work for applications, so added a note

pkg/sdk/applications_impl_gen.go
sfc-gh-swinkler: i think this is fixed, but do we have a make target to run linting errors created by reviewdog locally?

pkg/sdk/testint/applications_gen_integration_test.go
sfc-gh-swinkler: getting an error message: `share_events_with provider can only be set/unset if the app is created in a different account from the app package` cross account resource creation would make this a challenging test to write. i added a todo, but I don't think it adds much value for the time being.

pkg/sdk/applications_impl_gen.go
sfc-gh-asawicki: we have `make pre-push` which runs the linters

pkg/sdk/applications_def.go
sfc-gh-asawicki: Okay, and as with application package I guess that unsetting multiple is possible when you use commas? (consult with the thread https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2323#discussion_r1457115048)

pkg/sdk/testint/applications_gen_integration_test.go
sfc-gh-asawicki: Ok! Please add a subtask for SNOW-933289 and add the issue number to the comment please. We will get back to it when the time comes.

sfc-gh-asawicki (changes requested):
Please fix the build and I responded to two previous discussions.

pkg/sdk/testint/applications_gen_integration_test.go
sfc-gh-swinkler: SNOW-1016268

pkg/sdk/applications_def.go
sfc-gh-swinkler: fixed

sfc-gh-asawicki (dismissed):

pkg/sdk/testint/applications_gen_integration_test.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: So maybe we should skip it for now? It doesn't make sense if there're no asserts to check tags (+ knowing they're not supported yet)

sfc-gh-jcieslak (dismissed):
Approved, but consider skipping that tag test

PR #2357, opened 2024-01-16
feat: Add managed account to the SDK

sfc-gh-asawicki:
Add managed account to the SDK:
- used generator
- changed no quotes to single quotes for `ADMIN_NAME`

References
* [CREATE MANAGED ACCOUNT](https://docs.snowflake.com/en/sql-reference/sql/create-managed-account)
* [DROP MANAGED ACCOUNT](https://docs.snowflake.com/en/sql-reference/sql/drop-managed-account)
* [SHOW MANAGED ACCOUNTS](https://docs.snowflake.com/en/sql-reference/sql/show-managed-accounts)


pkg/sdk/managed_accounts_def.go
```diff
@@ -0,0 +1,70 @@
...
+		g.NewQueryStruct("CreateManagedAccount").
+			Create().
+			SQL("MANAGED ACCOUNT").
+			Name().
+			QueryStructField(
+				"CreateManagedAccountParams",
+				g.NewQueryStruct("CreateManagedAccountParams").
+					TextAssignment("ADMIN_NAME", g.ParameterOptions().SingleQuotes().Required()).
+					TextAssignment("ADMIN_PASSWORD", g.ParameterOptions().SingleQuotes().Required()).
+					PredefinedQueryStructField("typeProvider", "string", g.StaticOptions().SQL("TYPE = READER")).
```
sfc-gh-swinkler, line 45: should this be an enum rather than hardcoding? currently the only type supported is READER, but that could change in the future.

pkg/sdk/testint/managed_accounts_gen_integration_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	cleanupMangedAccountProvider := func(id sdk.AccountObjectIdentifier) func() {
+		return func() {
+			err := client.ManagedAccounts.Drop(ctx, sdk.NewDropManagedAccountRequest(id))
+			require.NoError(t, err)
+		}
+	}
+
+	createManagedAccountBasicRequest := func(t *testing.T) *sdk.CreateManagedAccountRequest {
```
sfc-gh-swinkler, line 38: we should have a discussion on what the structure for these kinds of test request helpers should be. i think there are a few different patterns right now. also there's the helpers_test.go file in the integration test folder, I think we should refactor that as well.

sfc-gh-swinkler (approved):

sfc-gh-jcieslak (approved):

pkg/sdk/testint/managed_accounts_gen_integration_test.go
sfc-gh-asawicki: We have already an issue for the helper test (SNOW-936093). We will have the discussion there. You are right we should unify the current approaches.

pkg/sdk/managed_accounts_def.go
sfc-gh-asawicki: It can change in the future, but it's currently hardcoded in the docs. We can change it without problems later (if needed - because it may never change).

PR #2380, opened 2024-01-19
feat: Use SDK in the storage integration

sfc-gh-jcieslak:
- Replacing snowflake package with SDK one
- Fixing some of the bugs
- Existing tests rewritten in the new framework + added new ones


pkg/resources/helpers_test.go
```diff
@@ -21,6 +22,26 @@ const (
 	onAll
 )
 
+var (
+	awsBucketUrl, awsBucketUrlIsSet = os.LookupEnv("AWS_EXTERNAL_BUCKET_URL")
```
sfc-gh-asawicki, line 26: I had a comment about naming these in other PR. It also applies here.

pkg/resources/storage_integration.go
```diff
@@ -118,284 +120,262 @@ func StorageIntegration() *schema.Resource {
...
-	if _, ok := d.GetOk("storage_aws_object_acl"); ok {
-		stmt.SetString("STORAGE_AWS_OBJECT_ACL", d.Get("storage_aws_object_acl").(string))
-	}
+	storageProvider := d.Get("storage_provider").(string)
 
-	// Now, set the storage provider
-	if err := setStorageProviderSettings(d, stmt); err != nil {
-		return err
+	switch storageProvider {
+	case "S3", "S3GOV":
```
sfc-gh-asawicki, line 157: "S3", "S3gov", "GCS", "AZURE", "S3GOV" -> these are the values available, so specifying `S3gov` with such switch will result with an error return from the default branch.

pkg/resources/storage_integration.go
```diff
@@ -118,284 +120,262 @@ func StorageIntegration() *schema.Resource {
...
-	id := d.Id()
-
-	stmt := snowflake.NewStorageIntegrationBuilder(d.Id()).Show()
-	row := snowflake.QueryRow(db, stmt)
-
-	// Some properties can come from the SHOW INTEGRATION call
+	ctx := context.Background()
+	client := sdk.NewClientFromDB(db)
+	id, ok := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
+	if !ok {
```
sfc-gh-asawicki, line 193: We are currently not doing a check like this in other places. I wonder if putting such an error is enough for the user. After all, id should be set by us, so we should have no trouble decoding it (if someone was not messing with state). We can leave it, but this will be bigger discussion with resources redesign.

sfc-gh-asawicki (approved):

pkg/resources/helpers_test.go
sfc-gh-jcieslak: I would merge it and change everything in the stages pr

sfc-gh-asawicki (approved):

PR #2405, opened 2024-01-23
feat: use roles from the SDK

sfc-gh-jcieslak:
TODO

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/datasources/account_roles.go
```diff
@@ -0,0 +1,94 @@
...
+	"database/sql"
+	"fmt"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var accountRolesSchema = map[string]*schema.Schema{
+	"pattern": {
```
sfc-gh-swinkler, line 14: so what i have done in some places is use 
```
like {
 pattern = "<pattern>"
}
```
The like doesn't really serve any purpose, but it does make it look more like the SQL. I don't have a problem either way, but we need to decide so we can be consistent in other data sources

pkg/datasources/account_roles.go
```diff
@@ -0,0 +1,94 @@
...
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Filters the command output by object name.",
+	},
```
sfc-gh-swinkler, line 18: are we not doing in_class or starts_with?

pkg/datasources/account_roles.go
```diff
@@ -0,0 +1,94 @@
...
+	"roles": {
```
sfc-gh-swinkler, line 19: another idea i had was instead of calling this roles, calling it `result` or `rows`.So instead of `data.snowflake_account_roles.roles.roles` we would have `data.snowflake_account_roles.roles.result`.

pkg/datasources/account_roles.go
```diff
@@ -0,0 +1,94 @@
...
+func ReadAccountRoles(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	db := meta.(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+
+	req := sdk.NewShowRoleRequest()
+	if pattern, ok := d.GetOk("pattern"); ok {
+		req.WithLike(sdk.NewLikeRequest(pattern.(string)))
+	}
+
+	roles, err := client.Roles.Show(ctx, req)
```
sfc-gh-swinkler, line 61: not that it applies here, but a question I would have is what if there was a describe command? Should we have a separate terraform data source for doing describe command? i.e `data.snowflake_account_role` ?  i think we do this with database and databases

pkg/datasources/account_roles_acceptance_test.go
```diff
@@ -0,0 +1,86 @@
...
+		Steps: []resource.TestStep{
+			{
+				ConfigDirectory: config.TestNameDirectory(),
+				ConfigVariables: configVariables,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.snowflake_account_roles.test", "roles.#", "2"),
+					containsAccountRole(accountRoleName1, comment),
+					containsAccountRole(accountRoleName2, comment),
+					func(state *terraform.State) error {
+						err := containsAccountRole(accountRoleName3, comment)(state)
```
sfc-gh-swinkler, line 48: isnt this just !containsAccountRole? i guess you still have error message to deal with. maybe just putting it into a function called stateNotContainsAccountRole? or adding a boolean flag to containsAccountRole for whether or not the role is expected to be present

pkg/resources/external_table.go
```diff
@@ -285,18 +285,22 @@ func UpdateExternalTable(d *schema.ResourceData, meta any) error {
 	if d.HasChange("tag") {
 		unsetTags, setTags := GetTagsDiff(d, "tag")
 
-		err := client.ExternalTables.Alter(ctx, sdk.NewAlterExternalTableRequest(id).WithUnsetTag(unsetTags))
-		if err != nil {
-			return fmt.Errorf("error setting tags on %v, err = %w", d.Id(), err)
+		if len(unsetTags) > 0 {
```
sfc-gh-swinkler, line 288: this seems unrelated and that it should be in a separate PR?

pkg/resources/account_role.go
```diff
@@ -0,0 +1,253 @@
...
+		},
+	}
+}
+
+func CreateAccountRole(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	db := meta.(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+
+	name := d.Get("name").(string)
+	id, err := helpers.DecodeSnowflakeParameterID(name)
```
sfc-gh-swinkler, line 47: why are you doing a decode here?

pkg/datasources/account_roles.go
sfc-gh-asawicki: I think this PR was not meant to improve the resource/data source. It was to use the SDK from within the existing resource. One additional thing that @sfc-gh-jcieslak did was rename the resource to be consistent (but left the improvements for the upcoming resources rework).

pkg/datasources/account_roles.go
sfc-gh-asawicki: check my other comment about this change's purpose

sfc-gh-asawicki (approved):

pkg/datasources/account_roles_acceptance_test.go
sfc-gh-jcieslak: done

pkg/resources/external_table.go
sfc-gh-jcieslak: I mean, it's a small change (just enclose within if-s). I can extract it into another pr, but that was just a thing I saw when I was looking for example uses of Tag changes.

pkg/resources/account_role.go
sfc-gh-jcieslak: true changed to new acc object identifier

pkg/datasources/account_roles.go
sfc-gh-jcieslak: I would leave it as is, but please add it to the list of topics we would like to discuss during the planning of supporting GA features.

pkg/datasources/account_roles.go
sfc-gh-jcieslak: Same as below

pkg/datasources/account_roles.go
sfc-gh-jcieslak: 👍 Removed renamed version, but left new CRUD functions

sfc-gh-swinkler (dismissed):

sfc-gh-asawicki (approved):

PR #2409, opened 2024-01-24
feat: Add API integration to the SDK

sfc-gh-asawicki:
Add API integration to the SDK


pkg/sdk/testint/api_integrations_gen_integration_test.go
```diff
@@ -0,0 +1,436 @@
...
+		request := createApiIntegrationAwsRequest(t)
+
+		request = request.
+			WithAwsApiProviderParams(request.AwsApiProviderParams.WithApiKey(sdk.String("key"))).
+			WithApiBlockedPrefixes(prefixes(awsBlockedPrefix)).
+			WithComment(sdk.String("comment"))
+
+		integration := createApiIntegrationWithRequest(t, request)
+
+		assertApiIntegration(t, integration, request.GetName(), "comment")
```
sfc-gh-jcieslak, line 136: 2 newly set params could we asserted with description properties. Probably `assertApiIntegrationProperties` function should be added to test whether newly added parameters were successfully set. Same in tests below.

sfc-gh-jcieslak: or just the same thing you've added in line:181

pkg/sdk/testint/api_integrations_gen_integration_test.go
```diff
@@ -0,0 +1,436 @@
...
+	t.Run("alter api integration: aws", func(t *testing.T) {
+		integration := createAwsApiIntegration(t)
+
+		otherRoleArn := "arn:aws:iam::000000000001:/role/other"
+		setRequest := sdk.NewAlterApiIntegrationRequest(integration.ID()).
+			WithSet(
+				sdk.NewApiIntegrationSetRequest().
+					WithAwsParams(sdk.NewSetAwsApiParamsRequest().WithApiAwsRoleArn(sdk.String(otherRoleArn)).WithApiKey(sdk.String("key"))).
+					WithEnabled(sdk.Bool(true)).
+					WithApiAllowedPrefixes(prefixes(awsBlockedPrefix)).
```
sfc-gh-jcieslak, line 171: it doesn't matter, but you mixed allowed prefixes argument with blocked prefix value (in other tests too)

sfc-gh-asawicki: It's on purpose because by default it's created with allowed, so to really alter something the blocked ones are used, maybe I can rename them to make it easier to read.

pkg/sdk/testint/api_integrations_gen_integration_test.go
sfc-gh-asawicki: I can add the assertions

sfc-gh-asawicki: Added

sfc-gh-jcieslak (approved):

PR #2419, opened 2024-01-25
feat: use sequence from sdk and add ordering attr

### LINKED ISSUE #2387: allow to specify noorder/order parameter for  sequence resource
### Terraform CLI and Provider Versions

TF 1.7.0
provider 0.84.0

With coming change in sequence behaviour https://docs.snowflake.com/en/release-notes/bcr-bundles/2024_01/bcr-1483 the support for ORDER and NOORDER parameters is missing in provider sequence resource. 

### Proposal

Add support to specify order and noorder parameters. Something like =>

resource snowflake_sequence sequence {
  database = "db"
  schema   = "test"
  name     = "seq1"
  order  = true
}

Medium

### Additional Information

_No response_

sfc-gh-swinkler:
removes old code from snowflake package, replacing with SDK package. does not change any existing functionality.

closes #2387


docs/resources/sequence.md
```diff
@@ -42,12 +42,13 @@ resource "snowflake_sequence" "test_sequence" {
 
 - `comment` (String) Specifies a comment for the sequence.
 - `increment` (Number) The amount the sequence will increase by each time it is used
+- `ordering` (String) The ordering of the sequence. Either ORDER or NOORDER. Default is ORDER.
```
sfc-gh-asawicki, line 45: Do we leave ORDER as a default in the provider? I mean, it changes in https://docs.snowflake.com/en/release-notes/bcr-bundles/2024_01/bcr-1483 to NOORDER by default.

So the current behavior would be aligned with Snowflake but it will change in the couple of days. Introducing NOORDER as a default would break existing configs (not break but cause different behavior) but it will be ready for the change later.

Maybe the one other way would be to do a small breaking change in the provider by introducing the ordering as a required param (adding it to our [migration notes](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md)) and letting users to decide themselves? WDYT @sfc-gh-swinkler @sfc-gh-jcieslak.

sfc-gh-asawicki (changes requested):

docs/resources/sequence.md
sfc-gh-jcieslak: It's a small behavior change, maybe the best would be to stay with the previous default to not break the existing configs and have a task to change it (to the manual setting or different default) as a pre-v1 change.

sfc-gh-swinkler: we can just make this field nullable. if it is null, it will not set it, and therefore use whatever behavior is default

sfc-gh-swinkler: okay so i made the change, but I'm still not sure I am happy with it. feels janky. The problem is that during the read I only want to set the attribute if it was set in the config (otherwise it will look like drift .... user applies, then will be forever stuck in a planning loop until they edit their Terraform configuration). However, during the import there is no way of knowing whether we should set the configuration code or not.How to know if value is default or not default? I think this is an important design question we should answer, as there are many places where we do similar operations. So we should all agree on a general solution

sfc-gh-swinkler: i do know in some places (e.g. warehouses) we cannot rely on the default value in the docs, either because the default is wrong for that particular account, or because it could be setting feature flags that are only available for business critical, and therefore wouldn't work in other accounts. So having a null value for an attribute is a necessity. but understanding when to set this during the read is a mystery.

sfc-gh-asawicki: Discussed during the call:
- we rollback these changes
- we stick with explicitly stating our default
- we may change it with V1

sfc-gh-asawicki (approved):

PR #2420, opened 2024-01-25
feat: Use managed account from the SDK

sfc-gh-asawicki:
Use managed account from the SDK
- acceptance test was run manually and was successful after the change (it is still waiting for SNOW-1011985)
- initial read sleep was replaced with a retry (like in account resource)


pkg/resources/managed_account.go
```diff
@@ -102,82 +96,109 @@ func ManagedAccount() *schema.Resource {
...
+	client := sdk.NewClientFromDB(db)
+
+	ctx := context.Background()
+	objectIdentifier := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
+
+	// We have to wait during the first read, since the locator takes some time to appear.
+	// This approach has a downside of not handling correctly the situation where managed account was removed externally.
+	// TODO [SNOW-1003380]: discuss it as a provider-wide topic during resources redesign.
+	var managedAccount *sdk.ManagedAccount
+	var err error
```
sfc-gh-jcieslak, line 138: why err is predefined ?

sfc-gh-jcieslak (approved):

pkg/resources/managed_account.go
sfc-gh-asawicki: because managedAccount has to be predefined and then you either:
- predefined err here
- predefined err inside retry (because you can't use `:=` syntax with predefined and not predefined var)

either way you predefine it somewhere

PR #2427, opened 2024-01-26
feat: Use stage from sdk

sfc-gh-jcieslak:
Changes:
- Partially replaced resource and data source implementation with SDK (see code comments) + replace snowflake/stages usage in streams resource
- Partially removed snowflake/stage implementation/tests (same reason as above)
- Added acc test (smoke test)


pkg/datasources/stages.go
```diff
@@ -56,42 +57,55 @@ var stagesSchema = map[string]*schema.Schema{
 
 func Stages() *schema.Resource {
 	return &schema.Resource{
-		Read:   ReadStages,
-		Schema: stagesSchema,
+		ReadContext: ReadStages,
```
sfc-gh-swinkler, line 60: i didnt know about this ReadContext function. we could actually use this for all resources, as we are using context.Background()

pkg/resources/stage.go
```diff
@@ -357,57 +297,81 @@ func UpdateStage(d *schema.ResourceData, meta interface{}) error {
...
 		if err := snowflake.Exec(db, q); err != nil {
-			return fmt.Errorf("error updating stage comment on %v", d.Id())
+			return diag.Errorf("error updating stage comment on %v", d.Id())
 		}
 	}
 
-	tagChangeErr := handleTagChanges(db, d, builder)
-	if tagChangeErr != nil {
-		return tagChangeErr
+	if d.HasChange("tag") {
```
sfc-gh-swinkler, line 328: is there a reason for moving the tag diff logic here?

pkg/datasources/stages.go
sfc-gh-jcieslak: Yeah, this is one of the points I added to our discussion on our resource-designing conventions. I'm using it because Read is deprecated and ReadContext or ReadWithoutTimeout should be used instead.

pkg/resources/stage.go
sfc-gh-jcieslak: So, tags were one of the things I could change in the Update method, so I did it. `handleTagChanges` is the old implementation using the snowflake package.

sfc-gh-asawicki (approved):

PR #2428, opened 2024-01-26
feat: Use row access policy from SDK

sfc-gh-asawicki:
- Removed old row access policy implementation from resource and datasource.
- Migrated tests and added more.
- Left better TODO for signatures diff suppression.


pkg/resources/row_access_policy.go
```diff
@@ -223,37 +155,29 @@ func ReadRowAccessPolicy(d *schema.ResourceData, meta interface{}) error {
...
-	policyName := rowAccessPolicyID.RowAccessPolicyName
-
-	builder := snowflake.RowAccessPolicy(policyName, dbName, schema)
+	ctx := context.Background()
+	client := sdk.NewClientFromDB(db)
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.SchemaObjectIdentifier)
 
 	if d.HasChange("comment") {
 		comment := d.Get("comment")
 		if c := comment.(string); c == "" {
```
sfc-gh-swinkler, line 164: is an unset comment the same as setting it to ""?

pkg/resources/row_access_policy.go
```diff
@@ -117,104 +69,84 @@ func RowAccessPolicy() *schema.Resource {
...
-	signature := d.Get("signature").(map[string]interface{})
+	id := sdk.NewSchemaObjectIdentifier(databaseName, schemaName, name)
+
+	signature := d.Get("signature").(map[string]any)
 	rowAccessExpression := d.Get("row_access_expression").(string)
 
-	builder := snowflake.RowAccessPolicy(name, database, schema)
+	args := make([]sdk.CreateRowAccessPolicyArgsRequest, 0)
+	for k, v := range signature {
+		dataType, err := sdk.ToDataType(v.(string))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: It will. That's why there is TODO with SNOW-1020074 left.

I will for now remove from here ToDataType part (so that it's not stripped) but still, describe strips this info so we will still end up with infinite plan without the diff suppression (which SNOW-1020074 is about).

pkg/resources/row_access_policy.go
sfc-gh-asawicki: That's a good question. I think this is one of the topics that we need to discuss globally with SNOW-867235. For me: if there is UNSET supported, we should use it. It may differ between different parameters though.

sfc-gh-jcieslak (approved):

PR #2429, opened 2024-01-26
feat: Use API integration from SDK

sfc-gh-asawicki:
- Removed old api integration implementation from resource.
- Migrated tests and added more.
- Added not listed in docs ALTER options (`SET AZURE_TENANT_ID` and `SET GOOGLE_AUDIENCE`) to the SDK. Added tests.


pkg/resources/api_integration.go
```diff
@@ -127,143 +130,186 @@ func APIIntegration() *schema.Resource {
...
-	s, err := snowflake.ScanAPIIntegration(row)
+	integration, err := client.ApiIntegrations.ShowByID(ctx, id)
 	if err != nil {
-		// If no such resource exists, it is not an error but rather not exist
-		if err.Error() == snowflake.ErrNoRowInRS {
-			d.SetId("")
-			return nil
-		}
-		return fmt.Errorf("could not show api integration: %w", err)
+		log.Printf("[DEBUG] api integration (%s) not found", d.Id())
```
sfc-gh-swinkler, line 221: one thing we should discuss is how if you don't have required permissions, then the show command will return error same as if the object didn't exist. if customer applies using accountadmin, then later with a less privileged role, then their terraform state file will be completely wiped out. should we check that user has sufficient privileges to do show command before setting resource id to empty? I think its not necessary, but it might be a good improvement later.

pkg/resources/api_integration.go
```diff
@@ -127,143 +130,186 @@ func APIIntegration() *schema.Resource {
...
+func toApiIntegrationEndpointPrefix(paths []string) []sdk.ApiIntegrationEndpointPrefix {
+	allowedPrefixes := make([]sdk.ApiIntegrationEndpointPrefix, len(paths))
+	for i, prefix := range paths {
+		allowedPrefixes[i] = sdk.ApiIntegrationEndpointPrefix{Path: prefix}
+	}
+	return allowedPrefixes
+}
+
 // CreateAPIIntegration implements schema.CreateFunc.
 func CreateAPIIntegration(d *schema.ResourceData, meta interface{}) error {
```
sfc-gh-swinkler, line 142: jan was using createcontext function in his stage pr, I thought that was a pretty good idea. i don't think its necessary to do in these prs now, but its something we should eventually do for all resources as part of a larger refactoring effort

sfc-gh-jcieslak (approved):

pkg/resources/api_integration.go
sfc-gh-asawicki: Added as a part of SNOW-867235.

pkg/resources/api_integration.go
sfc-gh-asawicki: That's true for all the resources, so I am adding this to SNOW-867235.

pkg/resources/api_integration_acceptance_test.go
```diff
@@ -1,107 +1,395 @@
...
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
-func TestAcc_ApiIntegration(t *testing.T) {
-	if _, ok := os.LookupEnv("SKIP_API_INTEGRATION_TESTS"); ok {
-		t.Skip("Skipping TestAcc_ApiIntegration")
-	}
+const (
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I moved them lower, I don't think that we should be moving too much to helpers without first discussing the right convention (now it's a mess and we have task to clean them up SNOW-936093). (on the other hand I don't believe someone will use such names or these constants at all).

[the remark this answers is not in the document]
sfc-gh-jcieslak: Even better 👍

sfc-gh-jcieslak (approved):

Issue #2385, opened 2024-01-22
snowflake_file_format update-in-place doesn't apply


alexmaras:
### Terraform CLI and Provider Versions

```
Terraform v1.6.5
on linux_amd64
+ provider registry.terraform.io/snowflake-labs/snowflake v0.82.0
```

### Terraform Configuration

```terraform
resource "snowflake_file_format" "csv" {
    binary_format                  = "HEX"
    compression                    = "AUTO"
    database                       = "RAW"
    date_format                    = "AUTO"
    empty_field_as_null            = true
    encoding                       = "UTF8"
    error_on_column_count_mismatch = true
    escape                         = "NONE"
    escape_unenclosed_field        = "NONE"
    field_delimiter                = ","
    field_optionally_enclosed_by   = "\""
    format_type                    = "CSV"
    id                             = "RAW|EXAMPLE_SCHEMA|PROCESS_CSV"
    name                           = "PROCESS_CSV"
    null_if                        = [
        "NULL",
    ]
    parse_header                   = false
    record_delimiter               = <<-EOT
        
    EOT
    replace_invalid_characters     = false
    schema                         = "EXAMPLE_SCHEMA"
    skip_blank_lines               = false
    skip_byte_order_mark           = true
    skip_header                    = 1
    time_format                    = "AUTO"
    timestamp_format               = "AUTO"
    trim_space                     = false
}
```

### Expected Behavior

Updating a field in the `file_format` resource and re-applying updates the file format in snowflake

### Actual Behavior

The update does not change anything in snowflake, and terraform picks up the change as needing to be applied every time

### Steps to Reproduce

1. Create a `snowflake_file_format` resource
2. `terraform apply`
3. Update the resource, change the `field_optionally_enclosed_by` field for instance (not specific to this field) to `NONE`
4. `terraform apply`

### How much impact is this issue causing?

Medium

### Additional Information

Workaround at the moment is to taint the resource in terraform and allow it to be destroyed and recreated.

sfc-gh-asawicki:
Hey @alexmaras. Thanks for reporting the issue.

Could you provide the debug logs (`TF_LOG=DEBUG`) showing that the update is not taking place?

justin-ramirez-gametime:
We are having a similar issue:
Terraform v1.7.0 on darwin_arm64
+ provider registry.terraform.io/snowflake-labs/snowflake v0.84.0

These are included on the plan every time:
[...]

PR #2436, opened 2024-01-27
FileFormat Update() bugs

raulbonet:
ref https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2385

2 bugs:

1. `d.Get("format_type")` has to be casted first to type `sdk.FileFormatType` in order to be compared with another FileFormatType, if not, the comparitson is always `False`. 

2. The attribute `comment` was not part of the Update() function


sfc-gh-asawicki (approved):

sfc-gh-asawicki:
Merged. Thanks for your contribution @raulbonet!

PR #2437, opened 2024-01-27
feat: add refresh_mode and initialize to dynamic tables

sonya:
This PR contains two commits that add `refresh_mode` and `initialize` parameters to the `snowflake_dynamic_table` resource.

`REFRESH_MODE` and `INITIALIZE` are parameters that can only be set at creation time. (I believe these were both added within the last couple months.) They are listed under [required parameters](https://docs.snowflake.com/en/sql-reference/sql/create-dynamic-table#required-parameters) in the documentation, but both have defaults. There are use cases where, for performance or other reasons, users need to be able to override the defaults.

This PR adds support for these two options by changing `refresh_mode` from a read-only parameter to a settable optional parameter, and adding `initialize` as a new parameter. Please see the individual commit messages for additional caveats and implementation notes.

I am opening this PR because we have ongoing development that is blocked without the ability to set these parameters in Terraform. I realize this repo is in the middle of a major framework update, so please let me know if there's anything I can do to help make these features available without being disruptive to the framework work.

## Test Plan
* [x] acceptance tests
* [x] testing with our existing stack. We have a stack that includes ~100 dynamic tables, and I have been running `terraform apply` with this code change in place. This should provide decent coverage of use cases.


sfc-gh-asawicki:

pkg/resources/dynamic_table.go
```diff
@@ -75,6 +75,34 @@ var dynamicTableShema = map[string]*schema.Schema{
 		Optional:    true,
 		Description: "Specifies a comment for the dynamic table.",
 	},
+	"refresh_mode": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "INCREMENTAL if the dynamic table will use incremental refreshes, or FULL if it will recompute the whole table on every refresh.",
+		ForceNew:    true,
```
sfc-gh-asawicki, line 82: With our upcoming resources redesign on the road to V1 we have a topic of force new versus create or replace. According to the docs (both for `refresh_mode` and `initialize`):
```
To modify the property, replicate the dynamic table with a CREATE OR REPLACE DYNAMIC TABLE command.
```
So this opens two possibilities:
1. We leave the `ForceNew` as is.
2. We modify the resource's logic to use `OrReplace` when refresh mode or initialize changes.

Both have their pros and cons. Because the discussion will still take place globally for every resource, I think there is no need to change it here. My preference is to use OrReplace, though.

pkg/sdk/dynamic_table.go
```diff
@@ -22,6 +22,8 @@ type createDynamicTableOptions struct {
 	dynamicTable bool                    `ddl:"static" sql:"DYNAMIC TABLE"`
 	name         SchemaObjectIdentifier  `ddl:"identifier"`
 	targetLag    TargetLag               `ddl:"parameter,no_quotes" sql:"TARGET_LAG"`
+	Initialize   *string                 `ddl:"parameter,no_quotes" sql:"INITIALIZE"`
```
sfc-gh-asawicki, line 25: Please introduce types for RefreshMode and Initialize (e.g. like `ColumnConstraintType`).

Then you can have the AllRefreshModes list (like `AllColumnConstraintTypes`) which can be used for validation (`ValidateFunc`) on the resource parameter (like in `table_constraint.go`)

pkg/resources/dynamic_table_acceptance_test.go
```diff
@@ -25,13 +25,15 @@ func TestAcc_DynamicTable_basic(t *testing.T) {
...
-			"schema":     config.StringVariable(acc.TestSchemaName),
-			"warehouse":  config.StringVariable(acc.TestWarehouseName),
-			"query":      config.StringVariable(fmt.Sprintf(`select "id" from "%v"."%v"."%v"`, acc.TestDatabaseName, acc.TestSchemaName, tableName)),
-			"comment":    config.StringVariable("Terraform acceptance test"),
-			"table_name": config.StringVariable(tableName),
+			"name":         config.StringVariable(name),
+			"database":     config.StringVariable(acc.TestDatabaseName),
+			"schema":       config.StringVariable(acc.TestSchemaName),
+			"warehouse":    config.StringVariable(acc.TestWarehouseName),
+			"initialize":   config.StringVariable("ON_SCHEDULE"),
```
sfc-gh-asawicki, line 32: These are both optional parameters. Please add the steps to show how they behave with changes (e.g. first create without them set, then set one of them, change it, set the other one).

pkg/resources/dynamic_table.go
```diff
@@ -200,6 +223,15 @@ func ReadDynamicTable(d *schema.ResourceData, meta interface{}) error {
 			return err
 		}
 	}
+	if strings.Contains(dynamicTable.Text, "initialize = 'ON_CREATE'") {
```
sfc-gh-asawicki, line 226: Please set the refresh mode too. If I understand the docs correctly (https://docs.snowflake.com/en/sql-reference/sql/create-dynamic-table#required-parameters), it should be also extracted from the `Text` and not the `RefreshMode` prop.

pkg/sdk/testint/dynamic_table_integration_test.go
```diff
@@ -54,7 +54,9 @@ func TestInt_DynamicTableCreateAndDrop(t *testing.T) {
 		}
 		query := "select id from " + tableTest.ID().FullyQualifiedName()
 		comment := random.Comment()
-		err := client.DynamicTables.Create(ctx, sdk.NewCreateDynamicTableRequest(name, testWarehouse(t).ID(), targetLag, query).WithOrReplace(true).WithComment(&comment))
+		refreshMode := "FULL"
```
sfc-gh-asawicki, line 57: It would be nice to add a separate test for the default creation (without explicit refreshMode and initialize).

Then, for both these tests we should add checks for the show command (because of the extraction from text we do on the resource level).

pkg/resources/dynamic_table.go
```diff
@@ -75,6 +75,34 @@ var dynamicTableShema = map[string]*schema.Schema{
...
+				return false
+			case oldValue == "" && newValue != "":
+				return false
+			default:
+				return true
+			}
+		},
+		DiffSuppressOnRefresh: true,
+	},
+	"initialize": {
```
sfc-gh-asawicki, line 97: Please add validation for both refresh and initialize (you can check e.g. `table_constraint.go`).

pkg/resources/dynamic_table.go
```diff
@@ -75,6 +75,34 @@ var dynamicTableShema = map[string]*schema.Schema{
 		Optional:    true,
 		Description: "Specifies a comment for the dynamic table.",
 	},
+	"refresh_mode": {
```
sfc-gh-asawicki, line 78: Another awaiting discussion topic during resources redesign is: do we set the Snowflake default or leave it without the value? Again, both have their pros and cons. For now, for the majority of cases the default is set, so let's set it here also(for both refresh mode and initialize).

pkg/resources/dynamic_table.go
```diff
@@ -75,6 +75,34 @@ var dynamicTableShema = map[string]*schema.Schema{
...
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "INCREMENTAL if the dynamic table will use incremental refreshes, or FULL if it will recompute the whole table on every refresh.",
```
sfc-gh-asawicki, line 81: LEt's allow setting every option, the default one too.

pkg/resources/dynamic_table.go
```diff
@@ -75,6 +75,34 @@ var dynamicTableShema = map[string]*schema.Schema{
...
+		ForceNew:    true,
+		DiffSuppressFunc: func(_, oldValue, newValue string, d *schema.ResourceData) bool {
```
sfc-gh-asawicki, line 83: I think these two DiffSuppressFuncs are not needed. I think they can be removed even with the default specified.

sfc-gh-asawicki (changes requested):
Hey @sonya. Thank you for the contribution one more time! Sorry for the delay in the review.

I left a few comments, please let me know if you have time to adjust the implementation to our requests. @sfc-gh-jcieslak please add more comments if I left something out.

pkg/resources/dynamic_table.go
sfc-gh-jcieslak: It was already set (line: 247)

pkg/sdk/dynamic_table.go
sonya: Will do - I actually created these types in my first iteration of working on this but removed them after I felt like they didn't actually add value or readability to the code. But I also don't know yall's conventions and figured you'd let me know in the PR review.

pkg/resources/dynamic_table.go
sonya: Thanks for this context. It was also interesting to read through this codebase and see how `create or replace` is implemented for different resources.

Since you say this will be more of a global decision, I will leave it as is to save myself the time of getting the alternative working, but I'll check back in the future to see where things ended up.

sonya:
Thanks for all the pointers and your patience with my code. I *think* I addressed all the points (but I'm not actually able to run the integration tests because they appear to require an enterprise account. The acceptance tests I can run).

In order to test this on my actual stack, I had to revert the latest commit (#a5f969c) which otherwise causes a panic with an array-out-of-bounds error when parsing existing functions. I'll open a new issue about this.

sfc-gh-jcieslak (approved):
Looks good for me 👍 Thanks again for the contribution. Still, let's wait for the @sfc-gh-asawicki re-review.

pkg/sdk/dynamic_table.go
sfc-gh-asawicki: Thanks for adding them. Please use them in multiple places:
- in `createDynamicTableOptions` (instead of `*string` for `Initialize` and `RefreshMode`)
- in acceptance tests (instead of hardcoded string values)
- in resource implementation (instead of hardcoded string values)

pkg/resources/dynamic_table_acceptance_test.go
```diff
@@ -87,6 +103,58 @@ func TestAcc_DynamicTable_basic(t *testing.T) {
...
+					}),
+				),
+			},
+			// test changing refresh_mode setting
+			{
+				ConfigDirectory: config.TestStepDirectory(),
+				ConfigVariables: variableSet4,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "name", name),
+					resource.TestCheckResourceAttr(resourceName, "database", acc.TestDatabaseName),
```
sfc-gh-asawicki, line 144: a cosmetic change for me would be not to validate every single param with every step, but only the ones that we are changing but this is not required to merge this PR

sonya:
Updated again, thanks for all your time and detailed comments @sfc-gh-asawicki!

sfc-gh-asawicki:
I have two minor issues, but I will merge the change and solve them with one of my PRs today.

For future reference, please run `make pre-push` before pushing. There are multiple checks there, one of which is doc generation (because you changed the description of the arguments, and the docs are generated from it, so there's a diff there - I forgot to mention it earlier).

Thank you for your other contribution, @sonya; we really appreciate your help in improving the provider!

sfc-gh-asawicki (approved):

Issue #2329, opened 2024-01-09
Dynamic tables are not parsing the query correctly and recreate resources on every apply


dtiesling:
### Terraform CLI and Provider Versions

terraform v1.3.9
snowflake-provider v0.82.0

### Terraform Configuration

```terraform
resource "snowflake_dynamic_table" "lease_signature" {
  database  = "DB"
  name      = "LEASE_SIGNATURE"
  schema    = "PUBLIC"
  warehouse = "COMPUTE_WH"
  target_lag {
    maximum_duration = "20 minutes"
  }
  query = "SELECT * FROM DB.PUBLIC.LEASE_SIGNATURE"
}
```

### Expected Behavior

I expect the state to show the query as "SELECT * FROM ADDITIONAL_TERMS FROM DB.PUBLIC.FOO"

### Actual Behavior

State says `query  = "_SIGNATURE\" lag = '20 minutes' refresh_mode = 'AUTO' initialize = 'ON_CREATE' warehouse = COMPUTE_WH AS SELECT * FROM DB.PUBLIC.LEASE_SIGNATURE"`

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

High

### Additional Information

The select query in the example has been altered to not expose our code. If you cannot reproduce please reach out to me and I will help reproduce.

dtiesling:
I tested terraform v1.6.6 as well. Same behavior.

sfc-gh-asawicki:
Hey, @dtiesling. Thanks for reporting the issue.

I can't reproduce it. We even have a test verifying such a simple query: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/05f96c699e4e36db4b28380d9ac3577d4f50a709/pkg/resources/dynamic_table_acceptance_test.go#L21 Maybe to reproduce it, we should know the query you are using (because as you say, you altered it not to expose your code)? Is the provided config failing for you with such an altered query?

dtiesling:
@sfc-gh-asawicki I emailed you with one of the the exact terraform resources that is getting this behavior.

dtiesling:
@sfc-gh-asawicki were you able to reproduce my issue?

sfc-gh-asawicki:
Hey @dtiesling. I did not manage to check it last week I have it planned for this one, sorry for the delay.

sonya:
[...]

PR #2438, opened 2024-01-27
fix: parse dynamic table query from DDL

sonya:
Previously we have been extracting the `query` portion of dynamic table definitions from the stored DDL in Snowflake by searching for the string position of keywords that preceded the query. However, this approach has proven inadequate as there are many ways to write working dynamic table definitions that conflict with string position assumptions.
    
Observing that definitions of views and materialized views have faced similar challenges but are currently working, this commit extends the existing ViewSelectStatementExtractor that is currently being used to extract the query definition from view and materialized view DDL statements, applying the same approach to dynamic table DDL statements.

This is an attempt to address #2329

## Test Plan
* [x] acceptance tests
* [x] testing in our existing stack. We have an existing stack with ~100 dynamic tables where many have CTEs and fail to be parsed with v0.85.0. A few contains `AS` in the name and were affected by spurious diffs with v0.78.0 (as observed in #2329). All of our dynamic tables are properly preserved with code change in this PR.

## References

Original issue: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2329
Previous attempt to fix: https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2421


sfc-gh-asawicki:

pkg/snowflake/parser_test.go
```diff
@@ -137,6 +137,69 @@ from bar;`
...
+lag = 'DOWNSTREAM'
+refresh_mode = 'AUTO'
+initialize = 'ON_CREATE'
+warehouse = COMPUTE_WH
+as
+-- comment
+select *
+from bar;`
+
+	comment := `create dynamic table foo comment = 'asdf' lag = 'DOWNSTREAM' refresh_mode = 'AUTO' initialize = 'ON_CREATE' warehouse = COMPUTE_WH as select * from bar;`
```
sfc-gh-asawicki, line 163: In every example with `comment` it is placed in the "wrong" place. If you check `createDynamicTableOptions` you can see that the `Comment` is the last thing before the `query`. Please change it here and in the `ExtractDynamicTable()`.

sfc-gh-asawicki (changes requested):

I have left one comment. I am still not convinced with this change (and the extractors logic in general) and we may still want to remove it and change the behavior with V1 (https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2329#issuecomment-1915129444). I guess we can leave it for now (after you correct this according to my comment) and get back to it with the resources redesign which is upcoming.

pkg/snowflake/parser_test.go
sonya: Thanks for this info. This appears to an inconsistency in what Snowflake does internally with the query, as fetching the query using `SHOW DYNAMIC TABLES LIKE foo` results in a query where the comment comes before all other parameters.

This could be upstream instability due to dynamic tables being a preview feature. I will make the parser handle both cases.

I totally get your hesitation with this change given the state of the framework in general, but I am also prepared to use this as a temporary bandaid that at least unblocks us and others with an urgent need to rely on Terraform to manage dynamic tables.

sonya:
Updated the parser to handle a comment as both the first and last parameter before the query

sfc-gh-asawicki:
Thanks, I will check it today!

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

sfc-gh-asawicki:
@sonya I have merged the PR, thank you for the contribution! It will be a part of 0.86.0 release.

PR #2439, opened 2024-01-27
fix: allow DT warehouse to be updated in-place

sonya:
Dynamic table warehouses can be updated in-place with an ALTER DYNAMIC TABLE statement. The terraform resource was requiring the DT to be dropped and recreated in order to change the warehouse, but this is not necessary.

This commit removes the ForceNew modifier on the "warehouse" parameter so that users can easily change the warehouse. This is a setting that users may want to easily change for cost allocation purposes.

## Test Plan
* [ ] tested by running `terraform apply` and `terraform plan` on an existing stack with warehouse changes, observed that dynamic tables were updated in place, and `ALTER DYNAMIC TABLE` was run in Snowflake

## References

Documentation for `ALTER DYNAMIC TABLE` command: https://docs.snowflake.com/en/sql-reference/sql/alter-dynamic-table


sfc-gh-asawicki:

sfc-gh-jcieslak (dismissed):

pkg/resources/dynamic_table.go
```diff
@@ -61,7 +61,6 @@ var dynamicTableShema = map[string]*schema.Schema{
 		Type:        schema.TypeString,
 		Required:    true,
 		Description: "The warehouse in which to create the dynamic table.",
-		ForceNew:    true,
```
sfc-gh-asawicki, line 64: Please add the `warehouse` param change in the `TestAcc_DynamicTable_basic` test.

sfc-gh-asawicki (changes requested):

I have just a small comment with the missing test, and we can merge it after!

sonya:
Added the field to the basic acceptance test. I'm not sure if that's sufficient to demonstrate that the DT doesn't get destroyed and recreated. I did some more work in #2437 on being able to tell if a change caused a replace (but I might be missing a much more obvious way)

sfc-gh-asawicki:
Thanks, I will check it today.

sfc-gh-jcieslak (approved):
Maybe the second warehouse could be named TestSecondaryWarehouseName, but It's just a cosmetic change and we can do it later. Thanks for the contribution!

sfc-gh-asawicki (approved):

sfc-gh-asawicki:
I have merged the change. Thank you for your contribution @sonya!

PR #2445, opened 2024-01-30
feat: Use notification integration from sdk

sfc-gh-asawicki:
Use notification integration from sdk:
- use SDK
- change resource config to work correctly
- add to migration notes


pkg/resources/notification_integration.go
```diff
@@ -141,180 +160,182 @@ func NotificationIntegration() *schema.Resource {
...
+		}
+		role, ok := d.GetOk("aws_sns_role_arn")
+		if !ok {
+			return fmt.Errorf("if you use AWS_SNS provider you must specify an aws_sns_role_arn")
+		}
+		createRequest.WithPushNotificationParams(
+			sdk.NewPushNotificationParamsRequest().WithAmazonPushParams(sdk.NewAmazonPushParamsRequest(topic.(string), role.(string))),
+		)
+	case "GCP_PUBSUB":
+		if v, ok := d.GetOk("gcp_pubsub_subscription_name"); ok {
```
sfc-gh-jcieslak, line 191: maybe this param and `gcp_pubsub_topic_name` should be conflicting fields? Now you can specify both and integration will created with `gcp_pubsub_topic_name` only because it's checked later

sfc-gh-asawicki: Good idea. I will add it in the following PR, great success?

sfc-gh-jcieslak (approved):

PR #2447, opened 2024-01-30
feat: Add grant privileges to share resource

sfc-gh-jcieslak:
Changes
* Acceptance tests for new resource
* Unified GrantsOn in GrantPrivilegesToShare + Now user can specify multiple privileges when grating/revoking privileges
* New resource for granting privileges to share
* Created a ticket for missing identifier functionality for functions
* Added documentation for the new resource

Ref: #2417


pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+
+func CreateGrantPrivilegesToShare(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	db := meta.(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+	id := createGrantPrivilegesToShareIdFromSchema(d)
+	log.Printf("[DEBUG] created identifier from schema: %s", id.String())
+
+	err := client.Grants.GrantPrivilegeToShare(ctx, getObjectPrivilegesFromSchema(d), getShareGrantOn(d), sdk.NewAccountObjectIdentifier(id.ShareName.Name()))
+	if err != nil {
+		return diag.Diagnostics{
```
sfc-gh-swinkler, line 167: you can do diag.FromError() instead. Or maybe a helper function would be more appropriate.

pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+			diag.Diagnostic{
+				Severity: diag.Error,
+				Summary:  "Failed to parse internal identifier",
+				Detail:   fmt.Sprintf("Id: %s\nError: %s", d.Id(), err.Error()),
+			},
+		}
+	}
+
+	if d.HasChange("privileges") {
+		before, after := d.GetChange("privileges")
```
sfc-gh-swinkler, line 197: not important at all, but typically this would be called "old" and "new"

pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+
+		if len(privilegesToRemove) > 0 {
+			err = client.Grants.RevokePrivilegeFromShare(
+				ctx,
+				privilegesToRemove,
+				grantOn,
+				sdk.NewAccountObjectIdentifier(id.ShareName.Name()),
+			)
+			if err != nil {
+				return diag.Diagnostics{
```
sfc-gh-swinkler, line 243: so the point of having diags is so you can append warnings and such. but here it seems you are always just returning on diagnostic. is this intentional?

pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+			continue
+		}
+		if grant.GranteeName.Name() == id.ShareName.Name() { // TODO: id.ShareName should be outside resource identifier (forgot the name)
+			if grantedOn == grant.GrantedOn {
+				privileges = append(privileges, grant.Privilege)
+			}
+		}
+	}
+
+	// It's a pseudo-role, so we have to append it whenever it's specified in the configuration
```
sfc-gh-swinkler, line 338: i dont know what this means, but it looks like a hack. i guess there's probably a good reason for it though

pkg/resources/testdata/TestAcc_GrantPrivilegesToShare/OnAllTablesInSchema/variables.tf
```diff
@@ -0,0 +1,15 @@
+variable "share_name" {
+  type = string
+}
```
sfc-gh-swinkler, line 3: so the variables.tf doesn't actually need to be in its own file. maybe we should just put everything in test.tf. what do you think?

pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+	"tag_name",
+	"view_name",
+}
+
+var grantPrivilegesToShareSchema = map[string]*schema.Schema{
+	"share_name": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: "The fully qualified name of the share on which privileges will be granted.",
```
sfc-gh-swinkler, line 30: validation?

pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+	},
+	"privileges": {
+		Type:        schema.TypeSet,
+		Required:    true,
+		Description: "The privileges to grant on the share.",
```
sfc-gh-asawicki, line 35: the privileges list changes depending on the type of the resource. Can we at least direct customer to https://docs.snowflake.com/en/sql-reference/sql/grant-privilege-share#syntax for the list of allowed privileges?

docs/resources/grant_privileges_to_share.md
```diff
@@ -0,0 +1,108 @@
...
+
+```terraform
+##################################
+### on database
+##################################
+
+resource "snowflake_grant_privileges_to_share" "example" {
+  share_name  = snowflake_share.example.name
+  privileges  = ["USAGE"]
+  on_database = snowflake_database.example.name
```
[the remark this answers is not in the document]
sfc-gh-asawicki: No other resource was designed with `on { ... }` block, it is an interesting idea, though. We should have discussed that during the design, I am not sure if we want to open this can of worms again

pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+		switch id.Kind {
+		case OnDatabaseShareGrantKind:
+			if err := d.Set("database_name", id.Identifier.Name()); err != nil {
+				return nil, err
+			}
+		case OnSchemaShareGrantKind:
+			if err := d.Set("schema_name", id.Identifier.FullyQualifiedName()); err != nil {
+				return nil, err
+			}
+		// case OnFunctionShareGrantKind:
```
sfc-gh-asawicki, line 133: Add TODO with jira number here please (it will be easier to reference it later)

pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+	for _, grant := range grants {
+		if grant.GrantedTo != sdk.ObjectTypeShare {
+			continue
+		}
+		// Only consider privileges that are already present in the ID, so we
+		// don't delete privileges managed by other resources.
+		if !slices.Contains(id.Privileges, grant.Privilege) {
+			continue
+		}
+		if grant.GranteeName.Name() == id.ShareName.Name() { // TODO: id.ShareName should be outside resource identifier (forgot the name)
```
sfc-gh-asawicki, line 331: When/in what follow-up issue we will address this TODO?

pkg/resources/grant_privileges_to_share.go
sfc-gh-asawicki: Please describe bit a bit more thoroughly @sfc-gh-jcieslak: 
- attach the privilege name to the comment
- elaborate a bit more on what a pseudo-role is (e.g. it is not returned, etc)
- add a reason for which it has to be added

pkg/resources/grant_privileges_to_share.go
sfc-gh-asawicki: It also affects all the other places in this resource where you commented out the functions-related implementation. We should either remove these or leave a comment with the ticket number on each of them.

pkg/resources/grant_privileges_to_share_acceptance_test.go
```diff
@@ -0,0 +1,494 @@
+package resources_test
+
+import (
```
sfc-gh-asawicki, line 3: nice tests! Can we get at least a few negative ones? (e.g. privilege not matching the object type, no `on` option provided)

pkg/resources/testdata/TestAcc_GrantPrivilegesToShare/OnAllTablesInSchema/variables.tf
sfc-gh-asawicki: Like in the comment https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2448#discussion_r1472122669 I would opt for not changing it on the spot in the single PR. We should aim for consistency, and currently in all other tests we have these two files separately.

pkg/sdk/grants_test.go
```diff
@@ -847,30 +847,40 @@ func TestRevokePrivilegeFromShare(t *testing.T) {
...
+				Tag: NewSchemaObjectIdentifier("database-name", "schema-name", "tag-name"),
 			},
 			from: id,
 		}
-		assertOptsValidAndSQLEquals(t, opts, "REVOKE USAGE ON ALL VIEWS IN SCHEMA %s FROM SHARE %s", otherID.FullyQualifiedName(), id.FullyQualifiedName())
-	})
+		assertOptsValidAndSQLEquals(t, opts, "REVOKE READ ON TAG  \"database-name\".\"schema-name\".\"tag-name\" FROM SHARE %s", id.FullyQualifiedName())
+	})
+
+	// TODO: This one throws an error
```
sfc-gh-asawicki, line 870: what error? It throws it on unit or integration level? When we will address it?

sfc-gh-asawicki (changes requested):

docs/resources/grant_privileges_to_share.md
```diff
@@ -0,0 +1,108 @@
...
+}
+
+## ID: "\"share_name\"|USAGE|OnDatabase|\"database_name\""
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yeah, I wanted to keep it close to the previous ID's format and have non-changing parameters to the left and conditional parameters that depend on the inner type (e.g. OnDatabase, OnSchema, etc.) to the right.

pkg/resources/grant_privileges_to_share.go
```diff
@@ -0,0 +1,474 @@
...
+	},
+	"schema_name": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ForceNew:         true,
+		Description:      "The fully qualified name of the schema on which privileges will be granted.",
+		ValidateDiagFunc: IsValidIdentifier[sdk.DatabaseObjectIdentifier](),
+		ExactlyOneOf:     grantPrivilegesToShareGrantExactlyOneOfValidation,
+	},
+	//	TODO(SNOW-1021686): Because function identifier contains arguments which are not supported right now
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yup, we decided to leave it commented for now and come back after identifiers rework

pkg/resources/grant_privileges_to_share.go
sfc-gh-jcieslak: yeah, but diag.FromError() puts everything in the summary section and I wanted to put specifics in the details. We should probably make some internal helper functions for it to have a function that can create diags with severity, summary and details as arguments.

pkg/resources/grant_privileges_to_share.go
sfc-gh-jcieslak: Yes it is, I just didn't find a case in this resource to append a diag to a slice and return it later on. Most of the diags created here are Errors that should be returned immediately. The only Warning is when show on all tables in schemas should happen, but I also returned, because for now, we don't support that case (like in the other grant resources).

pkg/resources/grant_privileges_to_share.go
sfc-gh-jcieslak: Added more detailed comment

pkg/resources/grant_privileges_to_share.go
sfc-gh-jcieslak: True, I had it at the beginning but removed it when changing share_name for some reason. Added again.

pkg/resources/grant_privileges_to_share.go
sfc-gh-jcieslak: changed

pkg/resources/grant_privileges_to_share.go
sfc-gh-jcieslak: added

pkg/resources/grant_privileges_to_share.go
sfc-gh-jcieslak: I added TODO(ticket number) in all of the places someone would need to visit in order to add function_name param

pkg/resources/grant_privileges_to_share.go
sfc-gh-jcieslak: Yeah, that my bad, for some reason I didn't see it. Maybe because earlier I had different ShareName type. I was confused by grant.GranteeName returning <account_name>.<share_name> at first, but I saw that It only does that if you call `show grants on share`, but here we call `show grants on <granted_object_type> <granted_object_id>` and it only returns <share_name>. In short it means that the TODO can be removed.

pkg/sdk/grants_test.go
sfc-gh-jcieslak: Also my mistake. This option was possible in the SDK but Snowflake is not allowing it (I mean `grant/revoke on all views`). Removed the test case.

docs/resources/grant_privileges_to_share.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Example Usage
+
+```terraform
+##################################
+### on database
+##################################
+
+resource "snowflake_grant_privileges_to_share" "example" {
+  share_name  = snowflake_share.example.name
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: changed to "to_share" and for others like this "database_name" -> "on_database"

docs/resources/grant_privileges_to_share.md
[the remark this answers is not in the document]
sfc-gh-jcieslak: changed to "to_share" and for others like this "database_name" -> "on_database"

sfc-gh-asawicki (approved):

sfc-gh-asawicki (approved):

KarthikRajashekaran:
When this will be available

sfc-gh-jcieslak:
@KarthikRajashekaran in 0.85.0, It should be available today here's the pr for it - https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2401

PR #2448, opened 2024-01-30
feat: Use materialized views and views from SDK

sfc-gh-asawicki:
- views and materialized views are similar
- set/unset tags was tested and fixed for both
- tests were fixed


pkg/datasources/materialized_views.go
```diff
@@ -58,34 +58,32 @@ func MaterializedViews() *schema.Resource {
...
-		d.SetId("")
-		return nil
-	} else if err != nil {
-		log.Printf("[DEBUG] materialized unable to parse views in schema (%s)", d.Id())
+	schemaId := sdk.NewDatabaseObjectIdentifier(databaseName, schemaName)
+	extractedMaterializedViews, err := client.MaterializedViews.Show(ctx, sdk.NewShowMaterializedViewRequest().WithIn(
+		&sdk.In{Schema: schemaId},
+	))
+	if err != nil {
+		log.Printf("[DEBUG] failed when searching materialized views in schema (%s), err = %s", schemaId.FullyQualifiedName(), err.Error())
```
sfc-gh-swinkler, line 71: does this throw an error if no materialized views are found?

pkg/resources/testdata/TestAcc_View_basic/variables.tf
```diff
@@ -0,0 +1,31 @@
+variable "name" {
+  type = string
+}
```
sfc-gh-swinkler, line 3: should the vairables also be moved into the test.tf file?

pkg/datasources/materialized_views.go
sfc-gh-asawicki: No, but it can fail for other reasons (e.g. not existing schema - added test to SDK to show that).

pkg/resources/testdata/TestAcc_View_basic/variables.tf
sfc-gh-asawicki: What do you mean? For every test written this way, we have `test.tf` with config and `variables.tf` with variables.

pkg/resources/view.go
```diff
@@ -89,291 +89,192 @@ func View() *schema.Resource {
...
 		if err != nil {
-			return fmt.Errorf("error when building sql query on %v, err = %w", d.Id(), err)
-		}
-
-		if err := snowflake.Exec(db, query); err != nil {
 			return fmt.Errorf("error when changing property on %v and performing create or replace to update view statements, err = %w", d.Id(), err)
 		}
 	}
 
 	if d.HasChange("name") {
```
sfc-gh-jcieslak, line 223: Same with this rename, maybe move it as the last update operation? Updates after rename, like comment, will throw an error

pkg/resources/view_acceptance_test.go
```diff
@@ -1,109 +1,282 @@
...
-				Config: viewConfig(accName, true, "SELECT ROLE_NAME, ROLE_OWNER FROM INFORMATION_SCHEMA.APPLICABLE_ROLES", acc.TestDatabaseName, acc.TestSchemaName),
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_View_basic"),
+				ConfigVariables: m(),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("snowflake_view.test", "name", accName),
-					resource.TestCheckResourceAttr("snowflake_view.test", "database", acc.TestDatabaseName),
-					resource.TestCheckResourceAttr("snowflake_view.test", "comment", "Terraform test resource"),
-					resource.TestCheckResourceAttr("snowflake_view.test", "copy_grants", "true"),
-					checkBool("snowflake_view.test", "is_secure", true), // this is from user_acceptance_test.go
+					resource.TestCheckResourceAttr("snowflake_view.test", "name", viewName),
```
sfc-gh-jcieslak, line 213: same case as with the materialized views, maybe add other simple params (like comment) to see if rename won't break other changes?

pkg/resources/view_acceptance_test.go
```diff
@@ -242,3 +423,31 @@ data "snowflake_grants" "grants" {
 		databaseName, schemaName,
 		databaseName, schemaName)
 }
+
+func testAccCheckViewDestroy(s *terraform.State) error {
+	db := acc.TestAccProvider.Meta().(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "snowflake_materialized_view" {
```
sfc-gh-jcieslak, line 431: should be snowflake_view

pkg/snowflake/view.go
```diff
@@ -235,27 +234,3 @@ func ScanView(row *sqlx.Row) (*View, error) {
 	err := row.StructScan(r)
 	return r, err
 }
```
sfc-gh-jcieslak, line 236: We should be able to remove this whole file, right? or Drop and other operations are used somewhere?

sfc-gh-jcieslak (changes requested):

pkg/resources/materialized_view.go
```diff
@@ -79,148 +78,95 @@ func MaterializedView() *schema.Resource {
...
-		builder.WithComment(v.(string))
+		createRequest.WithComment(sdk.String(v.(string)))
 	}
 
-	if v, ok := d.GetOk("tag"); ok {
-		tags := getTags(v)
-		builder.WithTags(tags.toSnowflakeTagValues())
+	warehouseName := d.Get("warehouse").(string)
+	// TODO [SNOW-867235]: this was the old implementation, it's left for now, we will address this with resources rework discussions
+	err := client.Sessions.UseWarehouse(ctx, sdk.NewAccountObjectIdentifier(warehouseName))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: As I have written in the comment: this is the previous behavior, there was no fallback in it, I can add it to SNOW-867235 topic.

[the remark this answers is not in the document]
sfc-gh-asawicki: Added.

pkg/resources/materialized_view.go
```diff
@@ -230,76 +176,86 @@ func ReadMaterializedView(d *schema.ResourceData, meta interface{}) error {
...
-	view := mvid.ViewName
-
-	builder := snowflake.NewMaterializedViewBuilder(view).WithDB(dbName).WithSchema(schema)
-
 	db := meta.(*sql.DB)
+	ctx := context.Background()
+	client := sdk.NewClientFromDB(db)
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.SchemaObjectIdentifier)
+
 	if d.HasChange("name") {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Added `id = newId` to prevent what you described. Also changed one prop in the rename test to check it. Good catch! I wonder how many other resources have this problem -  I will add this to one of the tasks we have with common errors in the repo.

pkg/resources/materialized_view_acceptance_test.go
```diff
@@ -1,78 +1,188 @@
...
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: testAccCheckMaterializedViewDestroy,
 		Steps: []resource.TestStep{
 			{
-				Config: materializedViewConfig(warehouseName, tableName, viewName, fmt.Sprintf("SELECT ID, DATA FROM \\\"%s\\\" WHERE ID LIKE 'foo%%';", tableName), acc.TestDatabaseName, acc.TestSchemaName),
+				Config: materializedViewConfig(warehouseName, tableName, viewName, queryEscaped, acc.TestDatabaseName, acc.TestSchemaName, "Terraform test resource", true, false),
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Check comment above.

pkg/resources/materialized_view_acceptance_test.go
```diff
@@ -91,17 +201,94 @@ resource "snowflake_table" "test" {
...
+}
+
+func alterMaterializedViewQueryExternally(t *testing.T, id sdk.SchemaObjectIdentifier, query string, warehouse string) {
+	t.Helper()
+
+	client, err := sdk.NewDefaultClient()
+	require.NoError(t, err)
+	ctx := context.Background()
+
+	err = client.Sessions.UseWarehouse(ctx, sdk.NewAccountObjectIdentifier(warehouse))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I tried without it and some tests failed without it sometimes. I left it as a discussion topic in SNOW-867235 because this is not the only resource behavinglike that.

pkg/resources/view.go
sfc-gh-asawicki: Fixed like the previous one

pkg/resources/view_acceptance_test.go
sfc-gh-asawicki: check above

pkg/resources/view_acceptance_test.go
sfc-gh-asawicki: fixed

pkg/snowflake/view.go
sfc-gh-asawicki: It was used in streams still, I though it was something bigger, but it seems I can get rid of that afterall.

sfc-gh-jcieslak (approved):

sfc-gh-jcieslak (approved):

PR #2454, opened 2024-02-02
feat: use external functions

sfc-gh-swinkler:
update external func resources to use sdk


pkg/datasources/external_functions.go
```diff
@@ -56,42 +57,59 @@ var externalFunctionsSchema = map[string]*schema.Schema{
...
 
-	currentExternalFunctions, err := snowflake.ListExternalFunctions(databaseName, schemaName, db)
-	if errors.Is(err, sql.ErrNoRows) {
-		// If not found, mark resource to be removed from state file during apply or refresh
-		log.Printf("[DEBUG] external functions in schema (%s) not found", d.Id())
-		d.SetId("")
-		return nil
-	} else if err != nil {
-		log.Printf("[DEBUG] unable to parse external functions in schema (%s)", d.Id())
+	externalFunctions, err := client.ExternalFunctions.Show(ctx, req)
```
sfc-gh-asawicki, line 72: in old implementation there was showing in schema: ```stmt := fmt.Sprintf(`SHOW EXTERNAL FUNCTIONS IN SCHEMA "%s"."%v"`, databaseName, schemaName)```, so I guess we should stick with this (ofc after making them optional it has to be a bit more sophisticated, but still, we should use them).

pkg/datasources/external_functions.go
```diff
@@ -1,24 +1,25 @@
...
 	"database": {
 		Type:        schema.TypeString,
-		Required:    true,
+		Optional:    true,
 		Description: "The database from which to return the schemas from.",
 	},
 	"schema": {
 		Type:        schema.TypeString,
-		Required:    true,
+		Optional:    true,
```
sfc-gh-asawicki, line 22: database should be required if schema is present. maybe add requiredWith?

pkg/datasources/external_functions_acceptance_test.go
```diff
@@ -1,80 +1,59 @@
...
-	externalFunctionName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
-	resource.ParallelTest(t, resource.TestCase{
-		Providers:    providers(),
+	if _, ok := os.LookupEnv("SKIP_EXTERNAL_FUNCTION_TESTS"); ok {
+		t.Skip("Skipping TestAcc_ExternalFunction")
+	}
+	accName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
+	m := func() map[string]config.Variable {
+		return map[string]config.Variable{
+			"database":                  config.StringVariable(acc.TestDatabaseName),
```
sfc-gh-asawicki, line 23: if you are making these two optional, you should add 2 more tests

pkg/resources/external_function.go
```diff
@@ -89,6 +85,7 @@ var externalFunctionSchema = map[string]*schema.Schema{
 		Optional:    true,
 		ForceNew:    true,
 		Description: "Indicates whether the function can return NULL values or must return only NON-NULL values.",
+		Default:     true,
```
sfc-gh-asawicki, line 88: Why? It seems like a breaking change. If it was missing earlier, `returnNullAllowed` was not set in the `ExternalFunctionBuilder`, so it was false by default and the resulting query had `NOT NULL`:
```
	if !fb.returnNullAllowed {
		q.WriteString(` NOT`)
	}
	q.WriteString(` NULL`)
```

pkg/resources/external_function.go
```diff
@@ -172,7 +169,6 @@ var externalFunctionSchema = map[string]*schema.Schema{
 		Type:        schema.TypeString,
 		Optional:    true,
 		Default:     "user-defined function",
-		ForceNew:    true,
```
sfc-gh-asawicki, line 175: It also changes behavior. Please at least add this to the migration guide to make users aware of the change?

pkg/resources/external_function.go
```diff
@@ -196,216 +193,196 @@ func ExternalFunction() *schema.Resource {
...
 
-// ReadExternalFunction implements schema.ReadFunc.
-func ReadExternalFunction(d *schema.ResourceData, meta interface{}) error {
+func ReadContextExternalFunction(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	db := meta.(*sql.DB)
-	externalFunctionID, err := externalFunctionIDFromString(d.Id())
-	if err != nil {
-		return err
+	client := sdk.NewClientFromDB(db)
+	parts := strings.Split(d.Id(), helpers.IDDelimiter)
```
sfc-gh-asawicki, line 326: why not `DecodeSnowflakeID()`?

pkg/resources/external_function.go
```diff
@@ -473,63 +445,115 @@ func ReadExternalFunction(d *schema.ResourceData, meta interface{}) error {
...
 
-// DeleteExternalFunction implements schema.DeleteFunc.
-func DeleteExternalFunction(d *schema.ResourceData, meta interface{}) error {
+func UpdateContextExternalFunction(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	db := meta.(*sql.DB)
-	externalFunctionID, err := externalFunctionIDFromString(d.Id())
-	if err != nil {
-		return err
+	client := sdk.NewClientFromDB(db)
+	parts := strings.Split(d.Id(), helpers.IDDelimiter)
```
sfc-gh-asawicki, line 493: same as above

pkg/resources/external_function.go
```diff
@@ -473,63 +445,115 @@ func ReadExternalFunction(d *schema.ResourceData, meta interface{}) error {
...
+		return diag.Errorf("external function ID is not valid: %s", d.Id())
 	}
+	databaseName := parts[0]
+	schemaName := parts[1]
+	name := parts[2]
+	argtypes := parts[3]
+
+	arguments := make([]sdk.DataType, 0)
+	if argtypes != "" {
+		argTypes := strings.Split(argtypes, "-")
```
sfc-gh-asawicki, line 504: this split logic is repeated, please extract it

pkg/resources/external_function_acceptance_test.go
```diff
@@ -7,126 +7,158 @@ import (
...
 
 	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
+	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
-func TestAcc_ExternalFunction(t *testing.T) {
+func TestAcc_ExternalFunction_basic(t *testing.T) {
```
sfc-gh-asawicki, line 16: can we update args and test the behavior then? Also, let's test no args at all.

sfc-gh-asawicki (changes requested):

pkg/resources/external_function_acceptance_test.go
sfc-gh-swinkler: added test with no_args. updating args would just be a force-new change, so I don't think that is something we need to test.

pkg/resources/external_function.go
sfc-gh-swinkler: fixed

pkg/resources/external_function.go
sfc-gh-swinkler: n/a anymore, there was no need to encode it in the first place. using a simple `sdk.NewSchemaObjectIdentifierFromFullyQualifiedName(d.Id())` instead

pkg/resources/external_function.go
sfc-gh-swinkler: same as above

pkg/datasources/external_functions.go
sfc-gh-swinkler: make schemaName required with databaseName

pkg/datasources/external_functions.go
sfc-gh-swinkler: done, and added to MIGRATION_GUIDE

pkg/resources/external_function.go
sfc-gh-swinkler: > "Indicates whether the function can return NULL values or must return only NON-NULL values"
if true then it can return null value
if false it can only return non-null values.
In documentation
>> Default: The default is NULL (i.e. the function can return NULL values).
So if you would like me to make this false to keep with the old behavior, I can or else I have updated the MIGRATION DOCS with the change

pkg/resources/external_function.go
sfc-gh-swinkler: added to migration guide

pkg/datasources/external_functions.go
sfc-gh-swinkler: According to documentation, this is not supported https://docs.snowflake.com/en/sql-reference/sql/show-external-functions.

pkg/datasources/external_functions_acceptance_test.go
sfc-gh-swinkler: i dont think 2 more tests are necessary, since there are only two possibilities: database + schema set, or neither set. I have added a test for the second case

sfc-gh-asawicki: and what about just a database set?

pkg/resources/external_function.go
sfc-gh-asawicki: I don't like introducing so many behavior changes in the PR which was supposed to just use the new SDK implementation, and also knowing, that we will rework every resource in the upcoming months. This means potentially having behavioral changes multiple times.

Let's leave this as a behavior change this time. Close after reading.

pkg/resources/external_function_acceptance_test.go
sfc-gh-asawicki: We want it to be force new, so the test would document the expected behavior. I will close it either way.

pkg/datasources/external_functions_acceptance_test.go
sfc-gh-asawicki: Oh, okay, we allow db only with schema, fine.

MIGRATION_GUIDE.md
```diff
@@ -14,6 +14,19 @@ It is noted as a behavior change but in some way it is not; with the previous im
 
 We will consider adding `NOT NULL` back because it can be set by `ALTER COLUMN columnX SET NOT NULL`, but first we want to revisit the whole resource design.
 
+### snowflake_external_function resource changes
+
+#### *(behavior change)* return_null_allowed default is now true
+The `return_null_allowed` attribute default value is now `true`. This is a behavior change because it was `false` before. The reason it was changed is to match the expected default value in the [documentation](https://docs.snowflake.com/en/sql-reference/sql/create-external-function#optional-parameters) `Default: The default is NULL (i.e. the function can return NULL values).`
+
+#### *(behavior change)* comment is no longer required
```
sfc-gh-asawicki, line 22: this is not a behavior change, but let's leave it here.

sfc-gh-asawicki (dismissed):

pkg/datasources/external_functions.go
```diff
@@ -56,42 +58,59 @@ var externalFunctionsSchema = map[string]*schema.Schema{
...
 
-		externalFunctions = append(externalFunctions, externalFunctionMap)
+		externalFunctionMap["comment"] = externalFunction.Description
+		externalFunctionMap["language"] = externalFunction.Language
+		externalFunctionsList = append(externalFunctionsList, externalFunctionMap)
 	}
 
-	d.SetId(fmt.Sprintf(`%v|%v`, databaseName, schemaName))
-	return d.Set("external_functions", externalFunctions)
+	d.SetId(fmt.Sprintf(`external_functions|%v|%v`, databaseName, schemaName))
```
sfc-gh-jcieslak, line 111: why the identifier is different, will it cause issues for users who already use it? Because I don't think we need this prefix

sfc-gh-swinkler: anyways i changed it back to what it was

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

PR #2455, opened 2024-02-02
chore: Grants migration guide

sfc-gh-jcieslak:
- Migration guide - step-by-step migration of grant resources
- Small change in test.tf (introduce resource dependency for proper delete order)


MIGRATION_GUIDE.md
```diff
@@ -1,9 +1,170 @@
...
-describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behaviour
+This document is meant to help you migrate your Terraform config to the new newest version. In migration guides, we will only 
+describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behavior
 across different versions.
 
+## vX.XX.X -> v0.85.0
+
+### Migration from old grant resources to new ones
+
+In recent changes, we introduced new grant resources intended to replace old grant solutions. Here's some of the useful
```
sfc-gh-asawicki, line 11: I do not have any remarks regarding the content (although I did not run them).

One general remark I have, would be to move this description to separate file - it takes much space here. I would leave the `vX.XX.X -> v0.85.0` with short description and add link to the separate description.

sfc-gh-asawicki (approved):

MIGRATION_GUIDE.md
```diff
@@ -1,9 +1,170 @@
...
+    object_type = "DATABASE"
+    object_name = snowflake_database.test.name
+  }
+}
+```
+
+#### 3.1.2. terraform import
+
+Write the `terraform import` command with the ID so that the resource will be able to parse and fill the state correctly.
+You can find import syntax in the documentation for a given resource, [here](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_privileges_to_account_role#import)
```
sfc-gh-swinkler, line 63: would also put a link to id examples: https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_privileges_to_account_role#import-examples

MIGRATION_GUIDE.md
```diff
@@ -1,9 +1,170 @@
...
+resource "snowflake_database_grant" "old_resource" {
+  depends_on = [ snowflake_database.test, snowflake_role.a, snowflake_role.b ]
+  database_name = snowflake_database.test.name
+  privilege = "USAGE"
+  roles = [ snowflake_role.a.name, snowflake_role.b.name ]
+}
+```
+
+#### 1. terraform list
+
```
sfc-gh-swinkler, line 27: i would probably put a few sentences why you are doing this. "first we need to list all the grant resources that will need to be migrated. We can do that by running the `terraform state list` command".

Tip: for larger configurations, its best to filter the results using the grep commands;

MIGRATION_GUIDE.md
```diff
@@ -1,9 +1,170 @@
...
+Run `terraform state list` to search the grants you're looking for (for larger configurations it's best to filter the results), 
+for example, `terraform state list | grep "snowflake_database_grant"`.
+
+#### 2. terraform rm
+
+Now choose which one you would to migrate next and remove it from the state with `terraform state rm <resource_address>`.
```
sfc-gh-swinkler, line 33: again, an explanation here i think is important. i certainly hesitated for a few seconds when I read that I needed to remove something from my statefile. Just something so the person doesn't worry they will lose their resource

MIGRATION_GUIDE.md
```diff
@@ -1,9 +1,170 @@
...
+  }
+}
+```
+
+#### 3.3.3. terraform plan and apply
+
+After running `terraform plan` you'll see if there are any changes we have to do before applying our generated configuration.
+If no errors are appearing you can run `terraform apply` to import state into generated configurations. 
+
+#### 3.3.4. Thoughts on configuration generation
```
sfc-gh-swinkler, line 155: would recommend changing the title here to something more clear. like "Limitations of Configuration Generation".

MIGRATION_GUIDE.md
```diff
@@ -1,9 +1,170 @@
 # Migration guide
 
-This document is meant to help you migrate your Terraform config to new newest version. In migration guides we will only 
-describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behaviour
+This document is meant to help you migrate your Terraform config to the new newest version. In migration guides, we will only
```
sfc-gh-swinkler, line 3: should we link this in the dicussions? or from the discussions?

sfc-gh-asawicki (approved):

PR #2458, opened 2024-02-02
chore: Remove unused old implementation from snowflake pkg

sfc-gh-asawicki:
- Remove unused code from snowflake package
- Migrate datasources missed earlier
- Migrate small statements still using old package


pkg/datasources/resource_monitors.go
```diff
@@ -49,8 +49,10 @@ func ResourceMonitors() *schema.Resource {
 
 func ReadResourceMonitors(d *schema.ResourceData, meta interface{}) error {
 	db := meta.(*sql.DB)
+	client := sdk.NewClientFromDB(db)
+	ctx := context.Background()
 
-	account, err := snowflake.ReadCurrentAccount(db)
+	account, err := client.ContextFunctions.Current(ctx)
```
sfc-gh-swinkler, line 55: i think this function should be named CurrentAccount()  https://docs.snowflake.com/en/sql-reference/functions/current_account. Also should this be put in a  helper struct so it can be something like client.Context.Sessions.CurrentAccount()? Or maybe just client.ContextSessions.CurrentAccount?

pkg/datasources/resource_monitors_acceptance_test.go
```diff
@@ -5,14 +5,22 @@ import (
...
 func TestAcc_ResourceMonitors(t *testing.T) {
 	resourceMonitorName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
-	resource.ParallelTest(t, resource.TestCase{
-		Providers:    providers(),
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
```
sfc-gh-swinkler, line 22: eventually we should make this a configurable variable that can be set in one place and used everywhere. will allow us to upgrade required tf version as necessary

pkg/datasources/users.go
```diff
@@ -107,49 +106,47 @@ func Users() *schema.Resource {
...
 func ReadUsers(d *schema.ResourceData, meta interface{}) error {
 	db := meta.(*sql.DB)
+	ctx := context.Background()
+	client := sdk.NewClientFromDB(db)
+
 	userPattern := d.Get("pattern").(string)
 
-	account, err := snowflake.ReadCurrentAccount(db)
-	if err != nil {
+	account, err1 := client.ContextFunctions.CurrentAccount(ctx)
```
sfc-gh-swinkler, line 114: wait we already have CurrentAccount? Then what was the previous Current()? I don't see that anywhere in here: https://docs.snowflake.com/en/sql-reference/functions-context

pkg/resources/user_public_keys.go
```diff
@@ -60,10 +62,12 @@ func UserPublicKeys() *schema.Resource {
 }
 
 func checkUserExists(db *sql.DB, name string) (bool, error) {
```
sfc-gh-swinkler, line 64: this seems like a rather pointless function, we should probably remove this entirely. But not important right now

pkg/sdk/context_functions.go
```diff
@@ -130,6 +159,21 @@ func (c *contextFunctions) CurrentUser(ctx context.Context) (string, error) {
 	return s.CurrentUser, nil
 }
 
+func (c *contextFunctions) Current(ctx context.Context) (*CurrentDetails, error) {
```
sfc-gh-swinkler, line 162: okay, i see now. Is this really necessary though? just because it was done in a similar way in the old snowflake package, doesn't mean we need to do it again here. also there are more current functions than just what is listed here.

pkg/datasources/users.go
sfc-gh-asawicki: The problem is, we've already had a family of `CurrentX()` functions. In the old snowflake package there was one method, fetching both account and region. To spare two invocation, I created a new one in the SDK, which fetches all these CurrentX params. We can name it differently if you have an idea how. CurrentAll? CurrentAccountAll? any ideas? cc: @sfc-gh-jcieslak

pkg/datasources/resource_monitors.go
sfc-gh-asawicki: We already have current account check another comment).

I don't understand the second part, though. It's already exposed in client's ContextFunctions interface.

pkg/datasources/users.go
sfc-gh-swinkler: if your concern is with multiple invocations, there is a way we can fix that by batching requests together. I don't know exactly how we would accomplish this, but it is worth thinking about.

pkg/resources/user_public_keys.go
sfc-gh-asawicki: I am not sure, but I tried not to modify the existing logic in this PR.

pkg/sdk/context_functions.go
sfc-gh-asawicki: There are multiple times when we want to fetch both account and region, for me it is better to fetch all at once and not shoot for every single param.

pkg/datasources/users.go
sfc-gh-asawicki: It was not a concern about the multiple invocations performance but the need to check the errors for each. Now, we fetch multiple values together, and I think it's easier to use it.

sfc-gh-jcieslak (approved):

pkg/datasources/resource_monitors.go
sfc-gh-asawicki: Renamed to CurrentSessionDetails (because CurrentSession was already taken too).

sfc-gh-jcieslak (approved):

PR #2462, opened 2024-02-05
feat: use funcs from sdk

sfc-gh-swinkler:
refactor funcs to use sdk


pkg/datasources/functions.go
```diff
@@ -60,43 +62,58 @@ var functionsSchema = map[string]*schema.Schema{
...
-
-	for _, function := range currentFunctions {
-		functionMap := map[string]interface{}{}
+		d.SetId("")
+		return diag.Diagnostics{
+			diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary:  fmt.Sprintf("Unable to parse functions in schema (%s)", id),
+				// TODO: link to the design decisions doc
+				Detail: "See our document on design decisions for functions: <LINK (coming soon)>",
```
sfc-gh-jcieslak, line 93: I don't think we're going to release any design decisions on functions, are we? (this looks like copy-paste error with grants)

pkg/datasources/functions.go
```diff
@@ -60,43 +62,58 @@ var functionsSchema = map[string]*schema.Schema{
...
-		functionMap["comment"] = function.Description.String
-		functionMap["argument_types"] = functionSignatureMap["argumentTypes"].([]string)
-		functionMap["return_type"] = functionSignatureMap["returnType"].(string)
-
-		functions = append(functions, functionMap)
+		entities = append(entities, m)
 	}
-
-	return d.Set("functions", functions)
+	d.SetId(fmt.Sprintf(`%v|%v`, databaseName, schemaName))
```
sfc-gh-jcieslak, line 114: helpers.EncodeSnowflakeIdentifier could be used

pkg/resources/function.go
```diff
@@ -161,402 +168,663 @@ func Function() *schema.Resource {
...
+	switch lang {
+	case "JAVA":
+		return createJavaFunction(ctx, d, meta)
+	case "JAVASCRIPT":
+		return createJavascriptFunction(ctx, d, meta)
+	case "PYTHON":
+		return createPythonFunction(ctx, d, meta)
+	case "SCALA":
+		return createScalaFunction(ctx, d, meta)
+	case "": // SQL if language is not set
```
sfc-gh-jcieslak, line 182: add "scala" as one of the allowed values in the "language" global variable used in the ValidateFunc of "language" schema field, and add another case for "SQL". If user writes `language = "SQL"` then it will throw, because it will land into default statement. Make it `case "", "SQL":`

pkg/resources/function.go
```diff
@@ -161,402 +168,663 @@ func Function() *schema.Resource {
...
 	if err != nil {
-		return err
+		// if function is not found then mark resource to be removed from state file during apply or refresh
+		d.SetId("")
+		return diag.Diagnostics{
+			diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary:  "Describe function failed.",
+				// TODO: link to the design decisions doc
+				Detail: "See our document on design decisions for functions: <LINK (coming soon)>",
```
sfc-gh-jcieslak, line 538: same here (copy-paste error ?)

pkg/resources/function.go
```diff
@@ -161,402 +168,663 @@ func Function() *schema.Resource {
...
+			if diags != nil {
+				return nil, diags
+			}
+			args = append(args, sdk.FunctionArgumentRequest{ArgName: argName, ArgDataType: argDataType})
+		}
+	}
+	return args, nil
+}
+
 type functionID struct {
```
sfc-gh-jcieslak, line 726: Is there a point in keeping this ID? we have helpers for encoding and decoding to replace such IDs. At first sight, there's no special functionality, can we replace them with our helper functions?

pkg/resources/function.go
```diff
@@ -61,9 +66,11 @@ var functionSchema = map[string]*schema.Schema{
 		Type:        schema.TypeString,
 		Description: "The return type of the function",
 		// Suppress the diff shown if the values are equal when both compared in lower case.
-		DiffSuppressFunc: DiffTypes,
-		Required:         true,
-		ForceNew:         true,
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
sfc-gh-asawicki, line 69: `DiffTypes` was doing almost the same but in a case-insensitive manner. Why we are changing that? If this is needed, it should be added to the migration guide as a behavior change.

pkg/resources/function.go
```diff
@@ -76,7 +83,7 @@ var functionSchema = map[string]*schema.Schema{
 		Type:         schema.TypeString,
 		Optional:     true,
 		ForceNew:     true,
-		ValidateFunc: validation.StringInSlice(languages, false),
+		ValidateFunc: validation.StringInSlice(languages, true),
```
sfc-gh-asawicki, line 86: the same remark as above

pkg/resources/function.go
```diff
@@ -161,402 +168,663 @@ func Function() *schema.Resource {
...
-		SchemaName:   arr[1],
-		FunctionName: arr[2],
-		ArgTypes:     strings.Split(arr[3], "-"),
-	}, nil
+		name:          sdk.NewSchemaObjectIdentifier(database, schema, name),
+		argumentTypes: argDataTypes,
+	}
+}
+
+func DecodeFunctionID(id string) (*functionID, error) {
```
sfc-gh-asawicki, line 774: shouldn't it return SchemaObjectIdentifier (using sdk.NewSchemaObjectIdentifierWithArguments(...))?

pkg/resources/function_acceptance_test.go
```diff
@@ -31,39 +31,39 @@ func TestAcc_Function(t *testing.T) {
...
-					checkBool("snowflake_function.test_funct_java", "is_secure", false), // this is from user_acceptance_test.go
+					resource.TestCheckResourceAttr("snowflake_function.test_function_java", "name", functName),
+					resource.TestCheckResourceAttr("snowflake_function.test_function_java", "comment", "Terraform acceptance test for java"),
+					resource.TestCheckResourceAttr("snowflake_function.test_function_java", "statement", expBody4),
+					resource.TestCheckResourceAttr("snowflake_function.test_function_java", "arguments.#", "1"),
+					resource.TestCheckResourceAttr("snowflake_function.test_function_java", "arguments.0.name", "ARG1"),
+					resource.TestCheckResourceAttr("snowflake_function.test_function_java", "arguments.0.type", "NUMBER"),
+					checkBool("snowflake_function.test_function_java", "is_secure", false), // this is from user_acceptance_test.go
 
 					// TODO: temporarily remove unit tests to allow for urgent release
```
sfc-gh-asawicki, line 60: should they stay commented out?

pkg/resources/function_acceptance_test.go
```diff
@@ -31,39 +31,39 @@ func TestAcc_Function(t *testing.T) {
 			{
 				Config: functionConfig(functName, acc.TestDatabaseName, acc.TestSchemaName),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("snowflake_function.test_funct", "name", functName),
-					resource.TestCheckResourceAttr("snowflake_function.test_funct", "comment", "Terraform acceptance test"),
-					resource.TestCheckResourceAttr("snowflake_function.test_funct", "statement", expBody2),
-					resource.TestCheckResourceAttr("snowflake_function.test_funct", "arguments.#", "1"),
-					resource.TestCheckResourceAttr("snowflake_function.test_funct", "arguments.0.name", "ARG1"),
-					resource.TestCheckResourceAttr("snowflake_function.test_funct", "arguments.0.type", "VARCHAR"),
+					resource.TestCheckResourceAttr("snowflake_function.test_function", "name", functName),
```
sfc-gh-asawicki, line 34: We discussed that with SDK migration we are adding the missing tests. Can you add the one for scala? Can you add update/import steps?

sfc-gh-asawicki (changes requested):

pkg/datasources/functions.go
sfc-gh-swinkler: fixed

pkg/resources/function.go
sfc-gh-swinkler: added

pkg/resources/function.go
sfc-gh-swinkler: no, it was waiting on a fix from proc with identifier helpers, but I just added here to this PR

pkg/datasources/functions.go
```diff
@@ -60,43 +62,58 @@ var functionsSchema = map[string]*schema.Schema{
...
 	databaseName := d.Get("database").(string)
 	schemaName := d.Get("schema").(string)
 
-	d.SetId("functions")
-	currentFunctions, err := snowflake.ListUserFunctions(databaseName, schemaName, db)
-	if err != nil {
-		log.Printf("[DEBUG] error listing functions: %v", err)
-		return nil
+	request := sdk.NewShowFunctionRequest()
+	if databaseName != "" {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: in this case no, since database and schema are both required. but we should fix the data source to not require these. this can come later. cleaned up logic in meantime

pkg/datasources/functions.go
```diff
@@ -60,43 +62,58 @@ var functionsSchema = map[string]*schema.Schema{
...
+		request.WithIn(&sdk.In{Database: sdk.NewAccountObjectIdentifier(databaseName)})
 	}
+	if schemaName != "" {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: same as above

pkg/resources/function.go
sfc-gh-swinkler: i mean these are both case insensitive, so there is no behavior change. we should probably move this logic to a helper function, but since we also do it this same way in many other places, I took out DiffTypes which is a rather poor name anyways.

pkg/resources/function.go
sfc-gh-swinkler: there is no behavior change, since we were already ignoring the case diff. Even the default value of languages is "SQL", which is all caps, yet the languages array has it as lower case. But if you want i can add this to the migration guide.

pkg/resources/function.go
sfc-gh-swinkler: fixed

pkg/resources/function_acceptance_test.go
sfc-gh-swinkler: removed

pkg/resources/function_acceptance_test.go
sfc-gh-swinkler: done

pkg/datasources/functions.go
sfc-gh-swinkler: removed

pkg/resources/function.go
sfc-gh-swinkler: removed

sfc-gh-asawicki (approved):

pkg/resources/function.go
sfc-gh-jcieslak: Oh, ok 👍

sfc-gh-jcieslak (approved):

PR #2466, opened 2024-02-06
feat: Use sdk in table constraint resource

sfc-gh-asawicki:
- Removed old implementation.
- Removed not working `NOT NULL` from the resource definition with proper note.


pkg/resources/table_constraint.go
```diff
@@ -307,26 +356,24 @@ func ReadTableConstraint(_ *schema.ResourceData, _ interface{}) error {
...
 
 	if d.HasChange("name") {
 		_, n := d.GetChange("name")
-		_, err := db.Exec(builder.Rename(n.(string)))
+		constraintRequest := sdk.NewTableConstraintRenameActionRequest().WithOldName(tc.name).WithNewName(n.(string))
+		alterStatement := sdk.NewAlterTableRequest(*tableIdentifier).WithConstraintAction(sdk.NewTableConstraintActionRequest().WithRename(constraintRequest))
+		err = client.Tables.Alter(ctx, alterStatement)
 		if err != nil {
-			return fmt.Errorf("error renaming table constraint %v err = %w", tc.name, err)
+			return fmt.Errorf("error renaming table constraint %v err = %w", &tc.name, err)
```
sfc-gh-jcieslak, line 376: `&tc.name` will print a pointer, right ? If so, remove `&`

sfc-gh-jcieslak (approved):

pkg/resources/table_constraint.go
sfc-gh-asawicki: True. Will change in the following change then.

PR #2477, opened 2024-02-08
feat: create a workaround for granting privileges on all pipes

sfc-gh-jcieslak:
`granting privileges on all pipes to role/database role` is not allowed in Snowflake. As it was just one case, we decided to create an internal workaround that would make `on_all pipes` work as it was supported by Snowflake. The workaround works as follows:
- check if on_all pipes are called
- fetch pipes (in database or in schema; depending on what was in the original request)
- loop through the pipes and grant/revoke privileges on them
- created integration tests for the SDK and acceptance tests for `snowflake_grant_privileges_to_role`, `snowflake_grant_privileges_to_account_role`, and `snowflake_grant_privileges_to_database_role`

> Note: grant on_future pipes works as intended

[Reference](https://docs.snowflake.com/en/sql-reference/sql/grant-privilege#required-parameters) (check the text at the bottom of <object_type>)

### DIFF
--- a/pkg/sdk/grants_impl.go
+++ b/pkg/sdk/grants_impl.go
@@ -23,7 +23,9 @@ func (v *grants) GrantPrivilegesToAccountRole(ctx context.Context, privileges *A
 	opts.accountRole = role
 	logging.DebugLogger.Printf("[DEBUG] Grant privileges to account role: opts %+v", opts)
 
-	// TODO: Describe why it's here
+	// Snowflake doesn't allow bulk operations on Pipes. Because of that, when SDK user
+	// issues "grant x on all pipes" operation, we'll go and grant specified privileges
+	// to every Pipe one by one.
 	if on != nil &&
 		on.SchemaObject != nil &&
 		on.SchemaObject.All != nil &&
@@ -64,7 +66,9 @@ func (v *grants) RevokePrivilegesFromAccountRole(ctx context.Context, privileges
 	opts.accountRole = role
 	logging.DebugLogger.Printf("[DEBUG] Revoke privileges from account role: opts %+v", opts)
 
-	// TODO: Describe why it's here
+	// Snowflake doesn't allow bulk operations on Pipes. Because of that, when SDK user
+	// issues "revoke x on all pipes" operation, we'll go and revoke specified privileges
+	// from every Pipe one by one.
 	if on != nil &&
 		on.SchemaObject != nil &&
 		on.SchemaObject.All != nil &&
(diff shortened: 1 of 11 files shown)


pkg/sdk/grants_impl.go
```diff
@@ -33,6 +65,37 @@ func (v *grants) RevokePrivilegesFromAccountRole(ctx context.Context, privileges
...
+	if on != nil &&
+		on.SchemaObject != nil &&
+		on.SchemaObject.All != nil &&
+		on.SchemaObject.All.PluralObjectType == PluralObjectTypePipes {
+		return v.runOnAllPipes(
+			ctx,
+			on.SchemaObject.All.InDatabase,
+			on.SchemaObject.All.InSchema,
+			func(pipe Pipe) error {
+				return v.client.Grants.RevokePrivilegesFromAccountRole(
```
sfc-gh-asawicki, line 81: it seems like a few things could be extracted here: condition, all parameters are the same, the only diff is `GrantPrivilegesToAccountRole` vs `RevokePrivilegesFromAccountRole`, right?

sfc-gh-asawicki: And also the following ones

pkg/sdk/testint/grants_integration_test.go
```diff
@@ -205,6 +206,83 @@ func TestInt_GrantAndRevokePrivilegesToAccountRole(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, 0, len(grants))
 	})
+
+	t.Run("grant and revoke on all pipes", func(t *testing.T) {
```
sfc-gh-asawicki, line 210: can we have a negative test that will show that multiple errors are returned?

sfc-gh-asawicki (changes requested):

pkg/sdk/grants_impl.go
sfc-gh-jcieslak: I tried to extract it, but in condition `on` is used and it's different between account and database roles, InDatabase and InSchema I was able to extract and just pass it to the method (same as previous `on` is different) and the SDK call had to be exposed, because pass-through parameters are also different between functions (like privileges, role and opts).

pkg/sdk/testint/grants_integration_test.go
sfc-gh-jcieslak: Done. This change made me think about how we produce random names for our resources and that sometimes this is the cause of non-stable tests. In this case, I saw that pipes require identifiers within a certain range of characters and all of our random ID generators were not satisfying its requirements. This pushed me to make my own simple random name generator with an a-z set of characters (other ones like String, StringN, AlphaN, AlphanumericN return unsupported characters).

sfc-gh-asawicki (approved):

PR #2485, opened 2024-02-12
chore: update password policy attachment

sfc-gh-jcieslak:
Continue the change for password policy attachment (to a user):
- minor resource changes
- minor test changes
- changed returned field types for policy_references call
- added code conventions
- added integration test


pkg/resources/user_password_policy_attachment.go
```diff
@@ -74,27 +56,25 @@ func CreateUserPasswordPolicyAttachment(d *schema.ResourceData, meta interface{}
 	if err != nil {
 		return err
 	}
-	d.SetId(fmt.Sprintf(`%s|%s`, helpers.EncodeSnowflakeID(passwordPolicy), helpers.EncodeSnowflakeID(userName)))
+
+	d.SetId(helpers.EncodeSnowflakeID(userName.FullyQualifiedName(), passwordPolicy.FullyQualifiedName()))
```
sfc-gh-asawicki, line 60: Shouldn't we use pipes in each part? It will currently fail for "a.b"."c"."d". I know it's not the nicest but it will work (and await identifiers rework).

pkg/sdk/policy_references.go
```diff
@@ -5,108 +5,106 @@ import (
...
+	RefColumnName     *string
+	RefArgColumnNames *string
+	TagDatabase       *string
+	TagSchema         *string
+	TagName           *string
 	PolicyStatus      string
 }
 
 type policyReferenceDBRow struct {
 	PolicyDb          sql.NullString `db:"POLICY_DB"`
```
sfc-gh-asawicki, line 58: can db and schema be nullable?

sfc-gh-asawicki (changes requested):

pkg/resources/user_password_policy_attachment.go
sfc-gh-jcieslak: Right now the identifier looks like this `<user_name>|<password_policy_fully_qualified_name>`, so each part is divided by a pipe.

pkg/sdk/policy_references.go
sfc-gh-jcieslak: Yes, because afaik this is a general for all policies and some of them are created on the account level, so I'm guessing db and schema are optional. Same goes for RefDatabaseName and RefSchemaName.

pkg/resources/user_password_policy_attachment.go
sfc-gh-asawicki: Hm, that's not true:

```
func TestJanek(t *testing.T) {
	t.Run("is he lying?", func(t *testing.T) {
		userName := sdk.RandomAccountObjectIdentifier()
		passwordPolicy := sdk.RandomSchemaObjectIdentifier()

		encoded := EncodeSnowflakeID(userName.FullyQualifiedName(), passwordPolicy.FullyQualifiedName())

		require.Equal(t, fmt.Sprintf("%s|%s|%s|%s", userName.Name(), passwordPolicy.DatabaseName(), passwordPolicy.SchemaName(), passwordPolicy.Name()), encoded)
	})
}
```

```
Expected :st_,1}0dV5)8|pFuY9+6?_}ai|4)qY3FLyo;I=|8*e+HQ}@7)2H
Actual   :"st_,1}0dV5)8"|"pFuY9+6?_}ai"."4)qY3FLyo;I="."8*e+HQ}@7)2H"
```

... but, it may be not a problem for this particular resource for two reasons:
1. we do not use the second part at all
2. the current logic for `EncodeSnowflakeID` is really bad for multiple params and for that reason, the double quotes are left in the encoded id, which is bad but would help with the case I mentioned: `"a.b"."c"."d"`

I would prefer having the encoded id stored correctly, without double quotes and without the problems I have mentioned, but we may revisit it during the identifiers rework.

sfc-gh-jcieslak: 👍 let's leave it as is and as you said come back to it during the identifiers rework

sfc-gh-asawicki (approved):

Issue #2481, opened 2024-02-08
Why error handling in user.go was reverted (@sfc-gh-pbosak)


israel:
Hi,

It seems that the error handling change that I have committed in the past was reverted by @sfc-gh-pbosak in commit d644b63e673223b67f7092563a0bfc8c826fc38b . Why? 
My commit hash was ea01e66797703e53c58e29d3bdb36557b22dbf79
Please revert that change (just the error handling part) since a user that does not exist is not an error/exception.
Should be
`if err != nil && snowflake.IsResourceNotExistOrNotAuthorized(err.Error(), "User") {
		// If not found, mark resource to be removed from statefile during apply or refresh
		log.Printf("[DEBUG] user (%s) not found or we are not authorized.Err:\n%s", d.Id(), err.Error())
		d.SetId("")
		return nil
	}
	if err != nil {
		return err
	}`
instead of https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/0385650f9ea7d1708b65139f2eeeacc373d5c9f8/pkg/resources/user.go#L248

Thanks

sfc-gh-jcieslak:
Hey @israel 
Yesterday we released 0.86.0 which should fix this issue, could you confirm ?

sfc-gh-jcieslak:
Hey @israel
Could you please confirm and close the ticket if the provided change fixed your issue?

sfc-gh-jcieslak:
Closing due to inactivity

PR #2486, opened 2024-02-12
fix: user error handling

sfc-gh-jcieslak:
Fix for https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2481
Adjusted error handling in the User's Read operation. Now when an error happens and we get the "object does not exist or not authorized" error then we're marking the resource as removed and return nil. Otherwise, we return the error. Also, added a test case that proves no error happens when the user is removed outside of the Terraform (it failed with the previous implementation).

### DIFF
--- a/pkg/resources/user_acceptance_test.go
+++ b/pkg/resources/user_acceptance_test.go
@@ -1,11 +1,15 @@
 package resources_test
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"strconv"
 	"strings"
 	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 
 	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/testhelpers"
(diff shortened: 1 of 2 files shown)


pkg/resources/user_acceptance_test.go
```diff
@@ -112,6 +116,47 @@ func TestAcc_User(t *testing.T) {
...
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: config,
+			},
+			{
+				PreConfig: removeUserOutsideOfTerraform(t, userName),
+				Config:    config,
```
sfc-gh-asawicki, line 140: can we have at least one check here? that state is empty

sfc-gh-asawicki (changes requested):

pkg/resources/user_acceptance_test.go
```diff
@@ -1,12 +1,18 @@
 package resources_test
 
 import (
+	"context"
+	"errors"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
```
sfc-gh-asawicki: meee

sfc-gh-asawicki (approved):

PR #2496, opened 2024-02-13
feat: supports collation of table column

YAhiru:
add the collation attribute of the table column to the table resource.

please let me know if there is any code that does not follow the convention 🙏 

## Test Plan

* [x] acceptance tests

## References

*

### DIFF
--- a/pkg/resources/table.go
+++ b/pkg/resources/table.go
@@ -139,6 +139,12 @@ var tableSchema = map[string]*schema.Schema{
 					Default:     "",
 					Description: "Masking policy to apply on column. It has to be a fully qualified name.",
 				},
+				"collate": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Default:     "",
+					Description: "Column collation, e.g. utf8",
+				},
 			},
 		},
 	},
@@ -423,6 +429,10 @@ func getTableColumnRequest(from interface{}) *sdk.TableColumnRequest {
 		request.WithMaskingPolicy(sdk.NewColumnMaskingPolicyRequest(sdk.NewSchemaObjectIdentifierFromFullyQualifiedName(maskingPolicy)))
 	}
 
+	if strings.Contains(_type, "CHAR") || _type == "STRING" || _type == "TEXT" {
+		request.WithCollate(sdk.String(c["collate"].(string)))
+	}
+
 	return request.
 		WithNotNull(sdk.Bool(!c["nullable"].(bool))).
 		WithComment(sdk.String(c["comment"].(string)))
@@ -468,6 +478,10 @@ func toColumnConfig(descriptions []sdk.TableColumnDetails) []any {
 
 		if td.Comment != nil {
 			flat["comment"] = *td.Comment
+		}
+
+		if td.Collation != nil {
+			flat["collate"] = *td.Collation
 		}
 
 		if td.PolicyName != nil {
--- a/pkg/resources/table.go
+++ b/pkg/resources/table.go
@@ -261,6 +261,7 @@ type column struct {
 	identity      *columnIdentity
 	comment       string
 	maskingPolicy string
+	collate       string
 }
 
 type columns []column
@@ -274,13 +275,14 @@ type changedColumn struct {
 	dropedDefault         bool
 	changedComment        bool
 	changedMaskingPolicy  bool
+	changedCollate        bool
 }
 
 func (c columns) getChangedColumnProperties(new columns) (changed changedColumns) {
 	changed = changedColumns{}
 	for _, cO := range c {
 		for _, cN := range new {
-			changeColumn := changedColumn{cN, false, false, false, false, false}
+			changeColumn := changedColumn{cN, false, false, false, false, false, false}
 			if cO.name == cN.name && cO.dataType != cN.dataType {
 				changeColumn.changedDataType = true
 			}
@@ -297,6 +299,10 @@ func (c columns) getChangedColumnProperties(new columns) (changed changedColumns
 
 			if cO.name == cN.name && cO.maskingPolicy != cN.maskingPolicy {
 				changeColumn.changedMaskingPolicy = true
+			}
+
+			if cO.name == cN.name && cO.collate != cN.collate {
+				changeColumn.changedCollate = true
 			}
 
 			changed = append(changed, changeColumn)
@@ -369,6 +375,7 @@ func getColumn(from interface{}) (to column) {
 		_default:      cd,
 		identity:      id,
 		comment:       c["comment"].(string),
+		collate:       c["collate"].(string),
 		maskingPolicy: c["masking_policy"].(string),
 	}
 }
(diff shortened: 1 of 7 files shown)


sfc-gh-asawicki:
Hey @YAhiru. Thank you for the pull request, we will analyze and review it this week.

pkg/resources/table.go
```diff
@@ -786,14 +807,18 @@ func UpdateTable(d *schema.ResourceData, meta interface{}) error {
...
 			err := client.Tables.Alter(ctx, sdk.NewAlterTableRequest(id).WithColumnAction(sdk.NewTableColumnActionRequest().WithAdd(addRequest)))
 			if err != nil {
 				return fmt.Errorf("error adding column: %w", err)
 			}
 		}
 		for _, cA := range changed {
-			if cA.changedDataType {
-				err := client.Tables.Alter(ctx, sdk.NewAlterTableRequest(id).WithColumnAction(sdk.NewTableColumnActionRequest().WithAlter([]sdk.TableColumnAlterActionRequest{*sdk.NewTableColumnAlterActionRequest(fmt.Sprintf("\"%s\"", cA.newColumn.name)).WithType(sdk.Pointer(sdk.DataType(cA.newColumn.dataType)))})))
+			if cA.changedDataType || cA.changedCollate {
+				err := client.Tables.Alter(ctx, sdk.NewAlterTableRequest(id).WithColumnAction(sdk.NewTableColumnActionRequest().WithAlter([]sdk.TableColumnAlterActionRequest{*sdk.NewTableColumnAlterActionRequest(fmt.Sprintf("\"%s\"", cA.newColumn.name)).WithType(sdk.Pointer(sdk.DataType(cA.newColumn.dataType))).WithCollate(sdk.String(cA.newColumn.collate))})))
```
sfc-gh-jcieslak, line 821: Afaik column collation cannot be changed (also couldn't find anything in the documentation). There's no `alter column x set collation` I tried to set data type with the same data type and different collation, but Snowflake returned `SQL compilation error: cannot change column S from type "VARCHAR(16777216) COLLATE 'utf8'" to "VARCHAR(16777216) COLLATE 'en'" because they have incompatible collations.`. Maybe I'm not aware of something, but It should also show when you add test case I proposed.

pkg/resources/table_acceptance_test.go
```diff
@@ -1228,6 +1228,114 @@ resource "snowflake_table" "test_table" {
...
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "name", accName),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "database", acc.TestDatabaseName),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "schema", acc.TestSchemaName),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "comment", "Terraform acceptance test"),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "column.#", "3"),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "column.0.name", "column1"),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "column.0.collate", "en"),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "column.1.name", "column2"),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "column.1.collate", ""),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "column.2.name", "column3"),
```
sfc-gh-jcieslak, line 1254: there should be a test case that should check the behavior of changing from one collation to another. If that returns an error in every case and that's what is expected I would add new test function that would expect this error to show it's expected result.

sfc-gh-jcieslak (changes requested):
Thanks for the contribution! I had some small remarks, but other than that everything looks good. I would like @sfc-gh-asawicki to look at the change since he has more experience with the table resource.

pkg/resources/table.go
```diff
@@ -423,6 +436,10 @@ func getTableColumnRequest(from interface{}) *sdk.TableColumnRequest {
 		request.WithMaskingPolicy(sdk.NewColumnMaskingPolicyRequest(sdk.NewSchemaObjectIdentifierFromFullyQualifiedName(maskingPolicy)))
 	}
 
+	if strings.Contains(_type, "CHAR") || _type == "STRING" || _type == "TEXT" {
```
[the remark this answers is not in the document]
YAhiru: Thank you!

I have implemented a function to determine whether a type is a string type based on the documentation.
https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2496/commits/931dc0abdfad06d9cc3bb1781b7ae353dcab0f84

pkg/resources/table_acceptance_test.go
YAhiru: Thank you!

I have added a test case to cover the scenario where changing the Collation fails.

https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2496/commits/e15b911b62a9d12af36fb3e954c791324e77c6b2

pkg/resources/table.go
YAhiru: Thank you for your feedback! I agree with you.
However, if the collate change is on terraform, I think we need to tell the user in some way that the collate change will fail.
I would think that changing a VARCHAR to a shorter length, for example, would fail as well, but I didn't think it would be a problem since it would behave the same way as this implementation. What do you think?

pkg/resources/table.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Thanks, nicely tested

pkg/resources/table.go
sfc-gh-jcieslak: Yeah, you're right. Let's leave it as is.

sfc-gh-jcieslak (approved):

sfc-gh-asawicki (approved):

pkg/resources/table.go
```diff
@@ -139,6 +139,12 @@ var tableSchema = map[string]*schema.Schema{
 					Default:     "",
 					Description: "Masking policy to apply on column. It has to be a fully qualified name.",
 				},
+				"collate": {
```
sfc-gh-asawicki, line 142: We generate the docs automatically from the schema and we push it as part of the change. You can run make-pre push to generate them. Also, it would be nice to add an example with collate to examples/resources/snowflake_table/resource.tf

sfc-gh-asawicki (changes requested):
Good change, have just one small comment. Thanks for the contribution @YAhiru. I will also run the tests on it, to validate the correctness.

pkg/resources/table.go
YAhiru: Thank you !
I fixed golangci and doc-check and added example !

https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2496/commits/9f2fb0f414887a2d437f2e1e19bb52faf0ef2e26
https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2496/commits/1ba35b07682bc2fb89b71c33eaebff9e36b8eeea

sfc-gh-asawicki:
I have merged it. Thanks for the contribution one more time @YAhiru!

YAhiru:
@sfc-gh-jcieslak @sfc-gh-asawicki 
thank you for your reviews!

Issue #2356, opened 2024-01-16
Snowflake objects created via terraform differs from "same" objects created via snowsight


Petruspop:
### Terraform CLI and Provider Versions

Terraform 1.6.6
Snowflake-Labs/snowflake 0.83.1

### Terraform Configuration

```terraform
#Set the provider
terraform {
  required_providers {
    snowflake = {
      source = "Snowflake-Labs/snowflake"
      version = "0.83.1"
    }
  }
}

# Configure the Snowflake Provider
# pwd as env var, set snowflake_password=your-password-here
provider "snowflake" {
  account = "RUTERAS-CDPPETER"
  user = "TMPTF"
  role = "ACCOUNTADMIN"
}

# Create DB with data retention set
resource "snowflake_database" "database" {
  name = "data-retention-db"
  data_retention_time_in_days = 28
  comment = "Database created in tf with data retention 28"
}

# Create schema without paramater for data retention -> should inherit
resource "snowflake_schema" "schema" {
  database = snowflake_database.database.name
  name = "tfschema"
  comment = "Schema created in tf, no data retention parameter used/set"
}

# Create table without paramater for data retention -> should inherit
resource "snowflake_table" "table" {
  database = snowflake_database.database.name
  schema = snowflake_schema.schema.name
  name = "tftable"
  comment = "Table created in tf, no value for data retention parameter"
  column {
    name = "id"
    type = "int"
  }
  column {
    name = "textfield"
    type = "text"
  }
}
```

### Expected Behavior

Data retention days on schema and table should be 28, as it should be inherited from database.

### Actual Behavior

Data retention days is defaulted to 1, both for schema and table.

### Steps to Reproduce

1. `terraform apply`
2. Check parameter on objects

### How much impact is this issue causing?

High

### Additional Information

Marked this as High as you expect you have time travel for 28 days, but in fact only have 1 day. So you risk losing data which is really dangerous.

The following script can be used in Snowsight to verify the difference. 

-- Verify TF defect
USE DATABASE "data-retention-db"; -- This was created in step 1 from tf provider

CREATE SCHEMA "uischema";

USE SCHEMA "uischema";

CREATE TABLE "uitable" (id int,textfield text) COMMENT ='Table created via snowsight, with no data retention parameter value';

SHOW PARAMETERS LIKE '%DATA_RETENTION%' IN DATABASE "data-retention-db";
--> 28 - as expected (from tf)

[...]

PR #2502, opened 2024-02-14
fix: data retention time parameters

sfc-gh-jcieslak:
Fixes: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2356
Database retention time won't be set in the Read operation every time to avoid situations where Terraform expects the user to set it every time (if it's not set then it will always produce a plan, because of the Snowflake's data retention hierarchy). 
It will only be set there only if:
- it's not set in config and schema data retention value != database retention value
- it's set in the config

- acceptance tests added

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -4,6 +4,12 @@ This document is meant to help you migrate your Terraform config to the new newe
 describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behavior
 across different versions.
 
+## v0.86.0 ➞ v0.87.0
+
+#### *(behavior change)* Schema `data_retention_days`
+To make `snowflake_schema.data_retention_days` truly optional field (previously it was producing plan every time when no value was set),
+we added `-1` as a possible value as set it as default. That got rid of the unexpected plans when no value is set and added possibility to use default value assigned by Snowflake (see [the data retention period](https://docs.snowflake.com/en/user-guide/data-time-travel#data-retention-period)).
+
 ## v0.85.0 ➞ v0.86.0
 ### snowflake_table_constraint resource changes
 
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -5,7 +5,12 @@ describe deprecations or breaking changes and help you to change your configurat
 across different versions.
 
 ## v0.86.0 ➞ v0.87.0
+### snowflake_failover_group resource changes
+#### *(bug fix)* ACCOUNT PARAMETERS is returned as PARAMETERS from SHOW FAILOVER GROUPS
+Longer context in [#2517](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2517).
+After this change, one apply may be required to update the state correctly for failover group resources using `ACCOUNT PARAMETERS`.
 
+### snowflake_schema resource changes
 #### *(behavior change)* Schema `data_retention_days`
 To make `snowflake_schema.data_retention_days` truly optional field (previously it was producing plan every time when no value was set),
 we added `-1` as a possible value as set it as default. That got rid of the unexpected plans when no value is set and added possibility to use default value assigned by Snowflake (see [the data retention period](https://docs.snowflake.com/en/user-guide/data-time-travel#data-retention-period)).
--- a/pkg/resources/schema.go
+++ b/pkg/resources/schema.go
@@ -53,7 +53,6 @@ var schemaSchema = map[string]*schema.Schema{
 	"data_retention_days": {
 		Type:         schema.TypeInt,
 		Optional:     true,
-		Default:      1,
 		Description:  "Specifies the number of days for which Time Travel actions (CLONE and UNDROP) can be performed on the schema, as well as specifying the default Time Travel retention time for all tables created in the schema.",
 		ValidateFunc: validation.IntBetween(0, 90),
 	},
--- a/pkg/resources/schema.go
+++ b/pkg/resources/schema.go
@@ -55,14 +55,6 @@ var schemaSchema = map[string]*schema.Schema{
 		Optional:     true,
 		Description:  "Specifies the number of days for which Time Travel actions (CLONE and UNDROP) can be performed on the schema, as well as specifying the default Time Travel retention time for all tables created in the schema.",
 		ValidateFunc: validation.IntBetween(0, 90),
-		//DiffSuppressOnRefresh: true,
-		//DiffSuppressFunc: func(_, oldValue, newValue string, d *schema.ResourceData) bool {
-		//	if _, ok := d.GetOk("data_retention_days"); !ok && newValue != "" {
-		//		return true
-		//	}
-		//
-		//	return false
-		//},
 	},
 	"tag": tagReferenceSchema,
 }
@@ -74,17 +66,9 @@ func Schema() *schema.Resource {
 		Read:   ReadSchema,
 		Update: UpdateSchema,
 		Delete: DeleteSchema,
-
 		Schema: schemaSchema,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
-		},
-		CustomizeDiff: func(ctx context.Context, diff *schema.ResourceDiff, i interface{}) error {
-			if diff.HasChange("data_retention_days") {
-
-			}
-
-			return nil
 		},
 	}
 }
@@ -121,7 +105,7 @@ func ReadSchema(d *schema.ResourceData, meta interface{}) error {
 	ctx := context.Background()
 	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.DatabaseObjectIdentifier)
 
-	_, err := client.Databases.ShowByID(ctx, sdk.NewAccountObjectIdentifier(id.DatabaseName()))
+	database, err := client.Databases.ShowByID(ctx, sdk.NewAccountObjectIdentifier(id.DatabaseName()))
 	if err != nil {
 		d.SetId("")
 	}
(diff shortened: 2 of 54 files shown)


pkg/resources/schema.go
```diff
@@ -67,7 +66,6 @@ func Schema() *schema.Resource {
 		Read:   ReadSchema,
 		Update: UpdateSchema,
 		Delete: DeleteSchema,
-
```
sfc-gh-asawicki, line 70: I am adding such lines and you are removing them :D we will have to pick one :D

pkg/resources/schema.go
```diff
@@ -133,10 +131,15 @@ func ReadSchema(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	if _, ok := d.GetOk("data_retention_days"); ok || (!ok && retentionTime != int64(database.RetentionTime)) {
```
sfc-gh-asawicki, line 134: `!ok` part is not needed because we get to the second operand only if ok is not true

sfc-gh-asawicki: I am not convinced why we are setting the data retention for `!ok && retentionTime != int64(database.RetentionTime)` case. It is inconsistent:
1. for the simple case where we just do not fill the config for schema and want to use the database one, the data_retention_days on schema won't be filled.
2. for the case when someone alters the schema by hand, we will set this.

pkg/resources/schema_acceptance_test.go
```diff
@@ -137,6 +140,165 @@ func TestAcc_Schema_TwoSchemasWithTheSameNameOnDifferentDatabases(t *testing.T)
 	})
 }
 
+// proves https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2356 issue is fixed.
+func TestAcc_Schema_DefaultDataRetentionTime(t *testing.T) {
+	databaseName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
+	schemaName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
+
+	configVariables := func(databaseDataRetentionTime int) config.Variables {
```
sfc-gh-asawicki, line 148: it will be more readable in the tests below if we name this one configVariablesWithoutSchemaDataRetentionTime

pkg/resources/schema_acceptance_test.go
```diff
@@ -137,6 +140,165 @@ func TestAcc_Schema_TwoSchemasWithTheSameNameOnDifferentDatabases(t *testing.T)
...
+				ConfigVariables: configVariablesWithSchemaDataRetentionTime(10, 15),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_schema.test", "data_retention_days", "15"),
+				),
+			},
+			{
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_Schema_DefaultDataRetentionTime/WithoutSchema"),
+				ConfigVariables: configVariables(10),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_schema.test", "data_retention_days", "0"),
```
sfc-gh-asawicki, line 202: hmmm, I don't like this zero here. I know that this is the default for integer, but 0 is also a valid input for the retention time for schemas https://docs.snowflake.com/en/sql-reference/sql/create-schema#optional-parameters. We should work on it with the resource redesign, please add this to our epic.

pkg/resources/schema_acceptance_test.go
```diff
@@ -137,6 +140,165 @@ func TestAcc_Schema_TwoSchemasWithTheSameNameOnDifferentDatabases(t *testing.T)
...
+				),
+			},
+			{
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_Schema_DefaultDataRetentionTime/WithSchema"),
+				ConfigVariables: configVariablesWithSchemaDataRetentionTime(10, 0),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_schema.test", "data_retention_days", "0"),
```
sfc-gh-asawicki, line 209: so here the effective value on schema retention inside SF would be still 10, right? That will prove what I have written above.

After this step I am confident, that we should add not only param checks in each step, but also a check for the schema (and even db) retention param value inside SF and check both of them.

pkg/resources/schema_acceptance_test.go
```diff
@@ -137,6 +140,165 @@ func TestAcc_Schema_TwoSchemasWithTheSameNameOnDifferentDatabases(t *testing.T)
...
+		CheckDestroy: testAccCheckSchemaDestroy,
+		Steps: []resource.TestStep{
+			{
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_Schema_DefaultDataRetentionTime/WithoutSchema"),
+				ConfigVariables: configVariables(5),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckNoResourceAttr("snowflake_schema.test", "data_retention_days"),
+				),
+			},
+			// Terraform will unset it (hierarchy default in Snowflake)
```
sfc-gh-asawicki, line 257: wait, where are the checks for this step? Please check the state one (I expect it to be filled based on the logic in the resource; and I left the comment that it should not be set in state IMO) and the db.

sfc-gh-asawicki (changes requested):

pkg/resources/schema.go
sfc-gh-jcieslak: My thinking was that: if `retentionTime == database.RetentionTime`, it means that it's the same as default so the value shouldn't be set, but when they're not the same it means there was some manual change done on the Snowflake side, and terraform should correct it to the value from the config.

pkg/resources/schema_acceptance_test.go
sfc-gh-jcieslak: Changed

pkg/resources/schema_acceptance_test.go
sfc-gh-jcieslak: Added state from Snowflake checks

pkg/resources/schema_acceptance_test.go
sfc-gh-jcieslak: added to the resource redesign list + logic changed so this 0 is checked as a "set" value and not as a "non-set" value.

pkg/resources/schema.go
sfc-gh-jcieslak: added to GA resources redesign list

pkg/resources/schema.go
```diff
@@ -53,9 +53,9 @@ var schemaSchema = map[string]*schema.Schema{
 	"data_retention_days": {
 		Type:         schema.TypeInt,
 		Optional:     true,
-		Default:      1,
+		Default:      -1,
 		Description:  "Specifies the number of days for which Time Travel actions (CLONE and UNDROP) can be performed on the schema, as well as specifying the default Time Travel retention time for all tables created in the schema.",
```
sfc-gh-asawicki, line 57: If we stick with the default set to -1:
- please add this to the description (and regen the docs)
- please add this to the migration guide

pkg/resources/schema.go
```diff
@@ -107,11 +112,6 @@ func ReadSchema(d *schema.ResourceData, meta interface{}) error {
 	ctx := context.Background()
 	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.DatabaseObjectIdentifier)
 
-	_, err := client.Databases.ShowByID(ctx, sdk.NewAccountObjectIdentifier(id.DatabaseName()))
```
sfc-gh-asawicki, line 110: please bring it back

pkg/resources/schema_acceptance_test.go
```diff
@@ -137,6 +141,218 @@ func TestAcc_Schema_TwoSchemasWithTheSameNameOnDifferentDatabases(t *testing.T)
...
+					resource.TestCheckResourceAttr("snowflake_schema.test", "data_retention_days", "-1"),
+					checkDataRetentionTime(id, 5, 5),
+				),
+			},
+			{
+				PreConfig:       setSchemaDataRetentionTime(t, id, 20),
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_Schema_DefaultDataRetentionTime/WithoutSchema"),
+				ConfigVariables: configVariablesWithoutSchemaDataRetentionTime(5),
+				Check: resource.ComposeTestCheckFunc(
+					checkDataRetentionTime(id, 5, 20),
```
sfc-gh-asawicki, line 272: I think here we should unset for schema (because there is still -1 in schema config).

sfc-gh-asawicki (changes requested):

pkg/resources/schema.go
sfc-gh-asawicki: Yes, this is good, I was not precise enough, I only meant setting it in state (added a comment in test).

MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,12 @@ This document is meant to help you migrate your Terraform config to the new newe
 describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behavior
 across different versions.
 
+## v0.86.0 ➞ v0.87.0
+
+#### *(behavior change)* Schema `data_retention_days`
```
sfc-gh-asawicki, line 9: I think you have conflicts here (I already merged one change with v0.86.0 ➞ v0.87.0 changes

sfc-gh-asawicki (approved):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Resolved

sfc-gh-asawicki (approved):

Issue #2490, opened 2024-02-12
[prerelease] panic with index out of range [1] when reading existing function


sonya:
### Terraform CLI and Provider Versions

Terraform v1.6.5
on darwin_arm64

Provider is built from current main branch (pre-0.86.0)

### Terraform Configuration

I can reproduce the problem with just this

```terraform
resource "snowflake_function" "my_function" {
  name            = "MY_FUNCTION"
  database        = "MY_DB"
  schema          = "MY_SCHEMA"
  return_type     = "VARCHAR"
  return_behavior = "IMMUTABLE"
  statement       = "SELECT PARAM"
  arguments {
    name = "PARAM"
    type = "VARCHAR"
  }
}
```

### Expected Behavior

Based on the steps to reproduce, the last step should result in `Your infrastructure matches the configuration.`

### Actual Behavior

`terraform plan` fails with the following stack trace:
```
Stack trace from the terraform-provider-snowflake plugin:

panic: runtime error: index out of range [1] with length 1

goroutine 3649 [running]:
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk.NewSchemaObjectIdentifierFromFullyQualifiedName({0x1400026c230?, 0x0?})
    ~/code/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/identifier_helpers.go:208 +0x344
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources.ReadContextFunction({0x1065eef50, 0x14000898770}, 0x1400150a780, {0x1065aad00?, 0x14000e85790?})
    ~/code/Snowflake-Labs/terraform-provider-snowflake/pkg/resources/function.go:515 +0x1f4
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0x14000cb3260, {0x1065eeea8, 0x140006d6d20}, 0xd?, {0x1065aad00, 0x14000e85790})
    ~/go/pkg/mod/github.com/hashicorp/terraform-plugin-sdk/v2@v2.31.0/helper/schema/resource.go:795 +0xe8
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).RefreshWithoutUpgrade(0x14000cb3260, {0x1065eeea8, 0x140006d6d20}, 0x1400160eea0, {0x1065aad00, 0x14000e85790})
[...]

PR #2503, opened 2024-02-14
fix: Fix functions and small other fixes

sfc-gh-asawicki:
- change parallel test in task
- run pre-push (because main was not pre-pushed: docs and reformat)
- fix materialized view datasource test
- add state upgrader to function resource's id
- add test utilizing older provider version

References: #2490


docs/resources/dynamic_table.md
```diff
@@ -42,15 +42,16 @@ resource "snowflake_dynamic_table" "dt" {
 ### Optional
 
 - `comment` (String) Specifies a comment for the dynamic table.
+- `initialize` (String) Initialize trigger for the dynamic table. Can only be set on creation.
```
sfc-gh-jcieslak, line 45: as in refresh_mode specify what are the allowed values

sfc-gh-asawicki: Done

sfc-gh-jcieslak (approved):

Issue #2506, opened 2024-02-15
Schema scan for snowflake views with same names in single database produce wrong state file and cyclic changes


JESCHO99:
### Terraform CLI and Provider Versions

Terraform CLI = 1.5.0
Snowflake Provider = 0.85.0

### Terraform Configuration

```terraform
resource "snowflake_database" "this" {
  name                        = "SCHEMA_ISSUE_TEST"
  data_retention_time_in_days = 1
}

resource "snowflake_schema" "test_az" {
  database = snowflake_database.this.name
  name     = "TEST_SCHEMA_AZ"
  data_retention_days = 1
}

resource "snowflake_schema" "test_bz" {
  database = snowflake_database.this.name
  name     = "TEST_SCHEMA_BZ"
  data_retention_days = 1
}

resource "snowflake_view" "a" {

  database = snowflake_database.this.name
  schema   = snowflake_schema.test_az.name
  name     = "TEST_TABLE"

  statement =  <<-SQL
    select 1 as A from dual
SQL

  or_replace = true
  is_secure  = false
}

resource "snowflake_view" "b" {

  database = snowflake_database.this.name
  schema   = snowflake_schema.test_bz.name
  name     = "TEST_TABLE"

  statement =  <<-SQL
    select 1 as A from dual
SQL

  or_replace = true
  is_secure  = false
}
```

### Expected Behavior

Two views are created inside Snowflake and the next Plan will show no further applies. Schema references are handled correctly even if there are views with the same name in one database in different schemas.

### Actual Behavior

![image](https://github.com/Snowflake-Labs/terraform-provider-snowflake/assets/59776328/5ac1f77a-758e-40dc-b970-48ec04687fc4)

Even if the correct schema is mentioned in the state file terraform thinks that the view "b" is in the a-schema because of a failure during the Snowflake Metadata Scan. I tested the same terraform code in version 0.84.1 and there the scan of the views is currently still correct. Here you can see the changes in the Snowflake Query History.

Correct version 0.84.1:
![MicrosoftTeams-image](https://github.com/Snowflake-Labs/terraform-provider-snowflake/assets/59776328/7861110e-b31e-4fd1-bd90-1da039af7308)

Wrong version 0.85.0:
![MicrosoftTeams-image (1)](https://github.com/Snowflake-Labs/terraform-provider-snowflake/assets/59776328/1f8bea2b-e028-46e9-88e2-428020fa0fe4)

### Steps to Reproduce

[...]

Issue #2501, opened 2024-02-14
Issue with "snowflake_notification_integration" direction param


harukitake:
### Terraform CLI and Provider Versions

```
Terraform v1.7.2
```
#### before
```terraform

resource "snowflake_notification_integration" "this" {
  name                  = "${upper(replace(var.prefix, "-", "_"))}_NOTIFICATION_INTEGRATION"
  enabled               = true
  direction             = "OUTBOUND"
  notification_provider = "AWS_SNS"
  aws_sns_topic_arn     = var.sns_topic_arn
  aws_sns_role_arn      = "arn:aws:iam::${local.account_id}:role/${local.sns_publisher_role_name}"
}
```

#### after 
```terraform
resource "snowflake_notification_integration" "this" {
  name                  = "${upper(replace(var.prefix, "-", "_"))}_NOTIFICATION_INTEGRATION"
  enabled               = true
  notification_provider = "AWS_SNS"
  aws_sns_topic_arn     = var.sns_topic_arn
  aws_sns_role_arn      = "arn:aws:iam::${local.account_id}:role/${local.sns_publisher_role_name}"
}
```

### Expected Behavior

NO changes

### Actual Behavior

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following
symbols:
  ~ update in-place
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # module.notification_integration.aws_iam_role.sns_publisher will be updated in-place
  ~ resource "aws_iam_role" "sns_publisher" {
      ~ assume_role_policy    = jsonencode(
            {
              - Statement = [
                  - {
                      - Action    = "sts:AssumeRole"
                      - Condition = {
                          - StringEquals = {
                              - "sts:ExternalId" = "EXTERNAL_ID"
                            }
                        }
                      - Effect    = "Allow"
                      - Principal = {
                          - AWS = "arn:aws:iam:aaaaaa:user/bbbbbbb"
                        }
                    },
                ]
              - Version   = "2012-10-17"
            }
        ) -> (known after apply)
        id                    = "notification-integration-id"
        name                  = "notification-integration-name"
        tags                  = {}
        # (9 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

  # module.notification_integration.snowflake_notification_integration.this must be replaced
-/+ resource "snowflake_notification_integration" "this" {
[...]

Issue #2488, opened 2024-02-12
snowflake_table "data_rentention_days" and "data_retention_time_in_days" chaos


JESCHO99:
### Terraform CLI and Provider Versions

0.85.0

### Terraform Configuration

```terraform
"data_retention_days": {
		Type:         schema.TypeInt,
		Optional:     true,
		Default:      1,
		Description:  "Specifies the retention period for the table so that Time Travel actions (SELECT, CLONE, UNDROP) can be performed on historical data in the table. Default value is 1, if you wish to inherit the parent schema setting then pass in the schema attribute to this argument.",
		ValidateFunc: validation.IntBetween(0, 90),
		Deprecated:   "Use snowflake_object_parameter instead",
		Type:          schema.TypeInt,
		Optional:      true,
		Description:   "Specifies the retention period for the table so that Time Travel actions (SELECT, CLONE, UNDROP) can be performed on historical data in the table. Default value is 1, if you wish to inherit the parent schema setting then pass in the schema attribute to this argument.",
		ValidateFunc:  validation.IntBetween(0, 90),
		Deprecated:    "Use data_retention_time_in_days attribute instead",
		ConflictsWith: []string{"data_retention_time_in_days"},
	},
	"data_retention_time_in_days": {
		Type:          schema.TypeInt,
		Optional:      true,
		Description:   "Specifies the retention period for the table so that Time Travel actions (SELECT, CLONE, UNDROP) can be performed on historical data in the table. Default value is 1, if you wish to inherit the parent schema setting then pass in the schema attribute to this argument.",
		ValidateFunc:  validation.IntBetween(0, 90),
		Deprecated:    "Use snowflake_object_parameter instead",
		ConflictsWith: []string{"data_retention_days"},
	},
```

### Expected Behavior

Both arguments for a snowflake table "data_rentention_days" and "data_retention_time_in_days" are market as deprecated and pop up as a warning every time you execute a plan. For me it looks like "data_retention_time_in_days" should be the right parameter but the depreciation warning was just merged from the previous resource and was not changed.

### Actual Behavior

You get unwanted error messages and it is unclear which of the parameters to use or in general which parameter should be used to configure the rentention time of a table.

### Steps to Reproduce

1. `terraform plan`

### How much impact is this issue causing?

Medium

sfc-gh-asawicki:
Hey @JESCHO99. Thanks for creating the issue.

[...]

PR #2507, opened 2024-02-15
fix: Fix few smaller issues

sfc-gh-asawicki:
- Removed incorrect validation on procedure resources that were causing failure of the following tests:
  - `TestAcc_Procedure_Java`
  - `TestAcc_Procedure_Scala`
- Removed incorrect deprecation from table resource (#2488)
- Fixed ShowByID for views (#2506)
- Fixed notification_integration resource warnings for deprecated parameters (#2501)
- Fixed external functions and procedure - state migrations similar to functions (#2490)

References: #2501 #2506 #2488 #2490

### DIFF
--- a/pkg/resources/external_function_state_upgraders.go
+++ b/pkg/resources/external_function_state_upgraders.go
@@ -3,7 +3,6 @@ package resources
 import (
 	"context"
 	"encoding/csv"
-	"fmt"
 	"strings"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
@@ -21,14 +20,14 @@ func parseV085ExternalFunctionId(stringID string) (*v085ExternalFunctionId, erro
 	reader.Comma = '|'
 	lines, err := reader.ReadAll()
 	if err != nil {
-		return nil, fmt.Errorf("not CSV compatible")
+		return nil, sdk.NewError("not CSV compatible")
 	}
 
 	if len(lines) != 1 {
-		return nil, fmt.Errorf("1 line at a time")
+		return nil, sdk.NewError("1 line at a time")
 	}
 	if len(lines[0]) != 4 {
-		return nil, fmt.Errorf("4 fields allowed")
+		return nil, sdk.NewError("4 fields allowed")
 	}
 
 	return &v085ExternalFunctionId{
--- a/pkg/resources/external_function_state_upgraders.go
+++ b/pkg/resources/external_function_state_upgraders.go
@@ -1,0 +1,76 @@
+package resources
+
+import (
+	"context"
+	"encoding/csv"
+	"fmt"
+	"strings"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+type v085ExternalFunctionId struct {
+	DatabaseName             string
+	SchemaName               string
+	ExternalFunctionName     string
+	ExternalFunctionArgTypes string
+}
+
+func parseV085ExternalFunctionId(stringID string) (*v085ExternalFunctionId, error) {
+	reader := csv.NewReader(strings.NewReader(stringID))
+	reader.Comma = '|'
+	lines, err := reader.ReadAll()
+	if err != nil {
+		return nil, fmt.Errorf("not CSV compatible")
+	}
+
+	if len(lines) != 1 {
+		return nil, fmt.Errorf("1 line at a time")
+	}
+	if len(lines[0]) != 4 {
+		return nil, fmt.Errorf("4 fields allowed")
+	}
+
+	return &v085ExternalFunctionId{
+		DatabaseName:             lines[0][0],
+		SchemaName:               lines[0][1],
+		ExternalFunctionName:     lines[0][2],
+		ExternalFunctionArgTypes: lines[0][3],
+	}, nil
+}
+
+func v085ExternalFunctionStateUpgrader(ctx context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	if rawState == nil {
+		return rawState, nil
+	}
+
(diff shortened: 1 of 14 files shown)


pkg/resources/external_function_state_upgraders.go
```diff
@@ -0,0 +1,78 @@
...
+
+type v085ExternalFunctionId struct {
+	DatabaseName             string
+	SchemaName               string
+	ExternalFunctionName     string
+	ExternalFunctionArgTypes string
+}
+
+func parseV085ExternalFunctionId(stringID string) (*v085ExternalFunctionId, error) {
+	reader := csv.NewReader(strings.NewReader(stringID))
```
sfc-gh-jcieslak, line 20: For curiosity, is it a different case than `parseV085FunctionId` to parse it with `csv.NewReader` rather than `strings.Split`?

pkg/resources/procedure_acceptance_test.go
```diff
@@ -177,3 +180,60 @@ func TestAcc_Procedure_complex(t *testing.T) {
...
+					"snowflake": {
+						VersionConstraint: "=0.85.0",
+						Source:            "Snowflake-Labs/snowflake",
+					},
+				},
+				Config: procedureConfig(acc.TestDatabaseName, acc.TestSchemaName, name),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "name", name),
+					resource.TestCheckResourceAttr(resourceName, "database", acc.TestDatabaseName),
+					resource.TestCheckResourceAttr(resourceName, "schema", acc.TestSchemaName),
```
sfc-gh-jcieslak, line 207: I had a hard time seeing what was checked here. I saw that ID was changed in the upgrader, so I think IDs should be checked to show it's different between versions (and to show what we're testing here). That applies to other upgrader tests where id was migrated, but not checked.

pkg/resources/procedure_state_upgraders.go
```diff
@@ -0,0 +1,62 @@
...
+
+	oldId := rawState["id"].(string)
+	parsedV085ProcedureId, err := parseV085ProcedureId(oldId)
+	if err != nil {
+		return nil, err
+	}
+
+	argDataTypes := make([]sdk.DataType, len(parsedV085ProcedureId.ArgTypes))
+	for i, argType := range parsedV085ProcedureId.ArgTypes {
+		argDataType, err := sdk.ToDataType(argType)
```
sfc-gh-jcieslak, line 51: just to make sure `sdk.ToDataType` or `sdk.Datatype(argType)`? `ToDataType` cuts part of the type

pkg/sdk/testint/views_gen_integration_test.go
```diff
@@ -488,6 +488,31 @@ func TestInt_Views(t *testing.T) {
...
+		request1 := sdk.NewCreateViewRequest(id1, sql)
+		request2 := sdk.NewCreateViewRequest(id2, sql)
+
+		view1 := createViewWithRequest(t, request1)
+		view2 := createViewWithRequest(t, request2)
+
+		assert.Equal(t, id1, view1.ID())
+		assert.Equal(t, id2, view2.ID())
+
+		for i := 0; i < 10; i++ {
```
sfc-gh-jcieslak, line 509: shouldn't it be better to just show that ShowByID for id1 returns view1 and id2 view2 instead of loop? With the previous InDatabase search, I guess this could pass depending on the generated IDs. Anyway, we should most likely check both views and maybe not necessarily in a loop.

pkg/resources/external_function_state_upgraders.go
```diff
@@ -0,0 +1,78 @@
...
+	reader.Comma = '|'
+	lines, err := reader.ReadAll()
+	if err != nil {
+		return nil, fmt.Errorf("not CSV compatible")
```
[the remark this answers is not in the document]
sfc-gh-asawicki: This is a copy-paste from the old implementation, I don't think we should change it. I will change it, though.

pkg/resources/procedure_state_upgraders.go
sfc-gh-asawicki: Yes. check `convertProcedureDataType`, I had to use the same logic as in the new implementation.

pkg/sdk/testint/views_gen_integration_test.go
sfc-gh-asawicki: Without the loop, it failed in like 10% of runs. I remember that you said one time that this is ordered alphabetically, but I did not find anything like that in the documentation. Having the loop made it fail constantly, that's also why I started to doubt the alphabetical ordering. This loop does not cost us anything, so I would leave it, unless you have more knowledge about the predictable ordering.

pkg/resources/external_function_state_upgraders.go
sfc-gh-asawicki: It is copy-pasted from the old implementation. This is the best way, to have the compatible parser. I think we should not refactor these old methods if they were used before as counterpart from the old converters.

pkg/resources/procedure_acceptance_test.go
sfc-gh-asawicki: Added.

pkg/resources/external_function_state_upgraders.go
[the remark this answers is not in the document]
sfc-gh-asawicki: Added.

pkg/sdk/testint/views_gen_integration_test.go
sfc-gh-asawicki: Changed to alphabetical identifiers and :prayge:

sfc-gh-jcieslak (approved):

Issue #2410, opened 2024-01-24
Granting a database role to a share fails with "Root object was present, but now absent."


kingalban:
### Terraform CLI and Provider Versions

Terraform v1.7.0
on linux_amd64
+ provider registry.terraform.io/hashicorp/aws v4.67.0
+ provider registry.terraform.io/hashicorp/http v3.4.0
+ provider registry.terraform.io/hashicorp/random v3.6.0
+ provider registry.terraform.io/snowflake-labs/snowflake v0.84.1

### Terraform Configuration

```terraform
resource "snowflake_database" "example_database" {
  name = "EXAMPLE"
}

resource "snowflake_database_role" "exampled_database_role" {
  database = snowflake_database.example_database.name
  name = "EXAMPLE"
}

resource "snowflake_share" "example_share" {
  name     = "EXAMPLE_SHARE"
  accounts = ["<another>.<account>"]
}

resource "snowflake_database_grant" "share_use_database" {
  database_name = snowflake_database.example_database.name
  shares = [snowflake_share.example_share.name]
}

resource "snowflake_grant_database_role" "share_use_database_role" {
  database_role_name = "${snowflake_database_role.exampled_database_role.database}.${snowflake_database_role.exampled_database_role.name}"
  share_name         = snowflake_share.example_share.name
  depends_on = [snowflake_database_grant.share_use_database] # neglecting this results in a separate error
}
```

### Expected Behavior

The share `snowflake_share.example_share` should be granted the privilege `USAGE` on the database `snowflake_database.example_database` **and** database role `snowflake_database_role.exampled_database_role`.

### Actual Behavior

`terraform apply` produces a reasonable plan, but when it goes to apply it gives this error:
~~~
╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to snowflake_grant_database_role.share_use_database_role, provider "provider[\"registry.terraform.io/snowflake-labs/snowflake\"]" produced an unexpected new value: Root object was present, but
│ now absent.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
~~~

Checking Snowflake, the share wasn't granted any privileges on the database role.

Repeating `terraform apply`, it shows that the `snowflake_grant_database_role.share_use_database_role` needs to be created. Trying to do that gives the same error.

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

High

### Additional Information

[...]

PR #2508, opened 2024-02-15
fix: SHOW GRANTS mapping for share data type

sfc-gh-jcieslak:
Fixes: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2410
The root cause of the issue was that in SHOW GRANTS conversion from row representation to SDK object representation, we had a special case for shares that would split it on dot. Share may have only one part and this change adds support for such a situation. Additionally, it was covered by an existing test that was commented out and was waiting for the `snowflake_grant_privileges_to_share` resource.

### DIFF
--- a/pkg/sdk/grants.go
+++ b/pkg/sdk/grants.go
@@ -223,8 +223,13 @@ func (row grantRow) convert() *Grant {
 	var granteeName AccountObjectIdentifier
 	if grantedTo == ObjectTypeShare {
 		parts := strings.Split(row.GranteeName, ".")
-		name := strings.Join(parts[1:], ".")
-		granteeName = NewAccountObjectIdentifier(name)
+		switch {
+		case len(parts) == 1:
+			granteeName = NewAccountObjectIdentifier(parts[0])
+		case len(parts) > 1:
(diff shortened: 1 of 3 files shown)


pkg/sdk/grants.go
```diff
@@ -223,8 +223,13 @@ func (row grantRow) convert() *Grant {
 	var granteeName AccountObjectIdentifier
 	if grantedTo == ObjectTypeShare {
 		parts := strings.Split(row.GranteeName, ".")
-		name := strings.Join(parts[1:], ".")
-		granteeName = NewAccountObjectIdentifier(name)
+		switch {
+		case len(parts) == 1:
+			granteeName = NewAccountObjectIdentifier(parts[0])
+		case len(parts) > 1:
```
sfc-gh-asawicki, line 229: let's add logic explicitely for 2 and add a default (but here I think we should discuss if we have any default for such situation - like using last part always?) or should be change the method to return error).

sfc-gh-asawicki (approved):

sfc-gh-asawicki (approved):

PR #2518, opened 2024-02-16
Update failover_group.md allowed_integration_types

gdelia-pm:
Added STORAGE INTEGRATIONS to allowed_integration_types as it's noted as supported in the below, and testing show it works.  The document also shows EXTERNAL ACCESS as an acceptable integration type but that is still in preview in GCP and not something I've tested so left it out. 

https://docs.snowflake.com/en/user-guide/account-replication-intro#replicated-objects
https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/failover_group#allowed_integration_types

*


sfc-gh-asawicki:
Hey @gdelia-pm. Thanks for the contribution!

I have a few remarks until we can merge it:
1. We do not modify the doc md files directly. We update the resource schema (in this case, description) and generate the automatically (you can do this by running `make pre-push`). This change will be rewritten next time we generate the docs, so we have to follow this.
2. I would like to see an automatic test that it works. The best way would be to modify one of the tests in `failover_group_acceptance_test.go` or `failover_groups_integration_test.go`. The first one is better because we recently unskipped these tests, and the ones in the second one are still awaiting.

Can you proceed with the requested changes, or should we queue them on our side?

gdelia-pm:
I'd prefer you queue them on your side.  I've not done any of this before so suspect me trying would take way more time /effort than y'all.

sfc-gh-asawicki:
That's fine! I will queue this on our side.

sfc-gh-jmichalak:
@gdelia-pm This was fixed in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2776

Issue #2517, opened 2024-02-16
snowflake_failover_group object types showing changes on every run


gdelia-pm:
### Terraform CLI and Provider Versions

```
 .\terraform.exe version
Terraform v1.7.3
on windows_amd64
+ provider registry.terraform.io/snowflake-labs/snowflake v0.86.0
```

### Terraform Configuration

```terraform
resource "snowflake_failover_group" "SOURCE_FAILOVER_GROUP" {
  depends_on = [snowflake_database.DB, snowflake_storage_integration.AZURE_STORAGE_INTEGRATION_DEV, snowflake_role.FUNCTIONAL_ROLES, snowflake_warehouse.WH,]

  name                      = "FAILOVER_GROUP1"
  object_types              = ["ACCOUNT PARAMETERS", "DATABASES", "INTEGRATIONS", "NETWORK POLICIES", "RESOURCE MONITORS", "ROLES", "USERS", "WAREHOUSES"]
  allowed_accounts          = ["ACCOUNT.CORE1", "ACCOUNT.CORE_DR1"]
  allowed_integration_types = ["SECURITY INTEGRATIONS", "STORAGE INTEGRATIONS"]
  allowed_databases         = toset([for db in snowflake_database.DB : db.name])

  replication_schedule {
    interval = 15 # minutes
  }
}
```

### Expected Behavior

After the initial apply no further plan should be generated.  

### Actual Behavior

On every plan, object type PARAMETERS is detected as removed, while ACCOUNT PARAMETERS is set to be added.
This seems to be a result of how snowflake stores the account parameter option, as it's noted as just parameter when running show failover groups. I think the tf provider is reading the parameters on each load and seeing that as different than the accepted object type of account parameters. 

I can work around this by creating the group with the "ACCOUNT PARAMETERS" object type, applying the config, then changing my object type to "PARAMETERS" to suppress the change, but if I have to make another change that breaks things. 

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # snowflake_failover_group.SOURCE_FAILOVER_GROUP will be updated in-place
  ~ resource "snowflake_failover_group" "SOURCE_FAILOVER_GROUP" {
        id                        = "FAILOVER_GROUP1"
        name                      = "FAILOVER_GROUP1"
      ~ object_types              = [
          - "PARAMETERS",
[...]

PR #2522, opened 2024-02-19
fix: Fix ACCOUNT PARAMETERS option failover group resource

sfc-gh-asawicki:
- `SHOW FAILOVER GROUPS` return `PARAMETERS` instead of `ACCOUNT PARAMETERS`
- acceptance tests were unskipped by adding environment variable
- SDK integration tests still stay skipped

References: #2517

### DIFF
--- a/pkg/sdk/failover_groups.go
+++ b/pkg/sdk/failover_groups.go
@@ -410,7 +410,12 @@ func (row failoverGroupDBRow) convert() *FailoverGroup {
 	ots := strings.Split(row.ObjectTypes, ",")
 	pluralObjectTypes := make([]PluralObjectType, 0, len(ots))
 	for _, ot := range ots {
-		pluralObjectTypes = append(pluralObjectTypes, PluralObjectType(strings.TrimSpace(ot)))
+		pot := PluralObjectType(strings.TrimSpace(ot))
+		if pot == PluralObjectTypeParameters {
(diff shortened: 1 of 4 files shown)


pkg/sdk/failover_groups.go
```diff
@@ -410,7 +410,12 @@ func (row failoverGroupDBRow) convert() *FailoverGroup {
 	ots := strings.Split(row.ObjectTypes, ",")
 	pluralObjectTypes := make([]PluralObjectType, 0, len(ots))
 	for _, ot := range ots {
-		pluralObjectTypes = append(pluralObjectTypes, PluralObjectType(strings.TrimSpace(ot)))
+		pot := PluralObjectType(strings.TrimSpace(ot))
+		if pot == PluralObjectTypeParameters {
+			pluralObjectTypes = append(pluralObjectTypes, PluralObjectType("ACCOUNT PARAMETERS"))
```
sfc-gh-jcieslak, line 415: maybe account parameters should be one of the enum values instead of adhoc created one?

sfc-gh-jcieslak (approved):

pkg/sdk/failover_groups.go
sfc-gh-asawicki: I was not sure about this, I couldn't find them anywhere but just in this one place in the docs. That's why I did not move them there.

PR #2530, opened 2024-02-20
fix: data retention time parameters follow-up

sfc-gh-jcieslak:
A follow-up for https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2502.
In this pr I applied the same changes to the Database and Table resource. On the note side, I left some comments in the table resource. There's still a deprecated data retention parameter that I would remove because it wouldn't be used with the new implementation that always expects a value (-1 if not set). We can add a note in the migration note and change the description of the deprecated parameter stating that it doesn't have any effect on the table configuration. Also in the database tests I'm changing account-level data retention time which I think shouldn't (but may) affect other tests. I would leave it anyway, because on delete we set back the default value and crashes may occur, but very rarely (I think).

## Test Plan

* [x] acceptance tests


pkg/resources/testdata/TestAcc_Database_DefaultDataRetentionTime/WithDatabase/test.tf
```diff
@@ -0,0 +1,4 @@
+resource "snowflake_database" "test" {
```
sfc-gh-asawicki, line 1: is the name of this file okay? Shouldn't it be: withRetentionSet or smth like that? (the same with others)

pkg/resources/helpers_test.go
```diff
@@ -548,3 +548,23 @@ func queriedPrivilegesContainAtLeast(query func(client *sdk.Client, ctx context.
 		return nil
 	}
 }
+
+func updateAccountParameter(t *testing.T, client *sdk.Client, parameter sdk.AccountParameter, temporarily bool, newValue string) func() {
```
sfc-gh-asawicki, line 552: I don't like that this function puts so many cleanup's on the stack. There is much bigger change that one of them will fail. We base it also on the LIFO behavior of the t.Cleanup (which may never change but can).

As I wrote to you earlier, I think that the `updateAccountParameterTemporarily` is sufficient here. The difference is, that you should invoke the `t.Cleanup` just in the first step modifying the param and ignore it in the subsequent steps, i.e.:
```
PreConfig: func() {
		revertParameter := updateAccountParameterTemporarily(t, client, sdk.AccountParameterDataRetentionTimeInDays, "1")
		t.Cleanup(revertParameter)
}
```

```
PreConfig: func() {
		_ := updateAccountParameterTemporarily(t, client, sdk.AccountParameterDataRetentionTimeInDays, "1")
		}
```

I think we should limit the number of additional helper functions and not add the ones similar to the already existing ones.

pkg/resources/table_acceptance_test.go
```diff
@@ -1475,6 +1480,158 @@ func TestAcc_Table_MaskingPolicy(t *testing.T) {
...
+					resource.TestCheckResourceAttr("snowflake_table.test", "data_retention_time_in_days", "3"),
+					checkSchemaAndTableDataRetentionTime(id, 10, 3),
+				),
+			},
+		},
+	})
+}
+
+// proves https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2356 issue is fixed.
+func TestAcc_Table_DefaultDataRetentionTime_SetOutsideOfTerraform(t *testing.T) {
```
sfc-gh-asawicki, line 1573: because table is at the bottom of this hierarchy, can we have a test where we clean the above objects (i.e. we clean schema - it goes to db, we later clean db - it goes to account). Just for the sake of showing the behavior in the most complex scenario. (not needed but nice)

sfc-gh-asawicki (changes requested):

docs/resources/table.md
```diff
@@ -96,7 +96,7 @@ resource "snowflake_table" "table" {
...
 - `comment` (String) Specifies a comment for the table.
 - `data_retention_days` (Number, Deprecated) Specifies the retention period for the table so that Time Travel actions (SELECT, CLONE, UNDROP) can be performed on historical data in the table. Default value is 1, if you wish to inherit the parent schema setting then pass in the schema attribute to this argument.
-- `data_retention_time_in_days` (Number) Specifies the retention period for the table so that Time Travel actions (SELECT, CLONE, UNDROP) can be performed on historical data in the table. Default value is 1, if you wish to inherit the parent schema setting then pass in the schema attribute to this argument.
+- `data_retention_time_in_days` (Number) Number of days for which Snowflake retains historical data for performing Time Travel actions (SELECT, CLONE, UNDROP) on the object. A value of 0 effectively disables Time Travel for the specified table. Default value for this field is set to -1, which is a fallback to use Snowflake default. For more information, see Understanding & Using Time Travel.
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

docs/resources/table.md
```diff
@@ -96,7 +96,7 @@ resource "snowflake_table" "table" {
 - `cluster_by` (List of String) A list of one or more table columns/expressions to be used as clustering key(s) for the table
 - `comment` (String) Specifies a comment for the table.
 - `data_retention_days` (Number, Deprecated) Specifies the retention period for the table so that Time Travel actions (SELECT, CLONE, UNDROP) can be performed on historical data in the table. Default value is 1, if you wish to inherit the parent schema setting then pass in the schema attribute to this argument.
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

pkg/resources/helpers_test.go
sfc-gh-jcieslak: I added a `temporarily` parameter earlier but for some reason didn't use it. I think I'll change it to return cleanup which feels more consistent with what we have in integration tests.

pkg/resources/table_acceptance_test.go
sfc-gh-asawicki: I just asked for one additional test with the bigger hierarchy, not to change the existing one. But I guess it's also fine to leave it as is.

MIGRATION_GUIDE.md
```diff
@@ -10,11 +10,15 @@ across different versions.
 Longer context in [#2517](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2517).
 After this change, one apply may be required to update the state correctly for failover group resources using `ACCOUNT PARAMETERS`.
 
-### snowflake_schema resource changes
-#### *(behavior change)* Schema `data_retention_days`
-To make `snowflake_schema.data_retention_days` truly optional field (previously it was producing plan every time when no value was set),
+### snowflake_database, snowflake_schema, and snowflake_table resource changes
+#### *(behavior change)* Database `data_retention_time_in_days` + Schema `data_retention_days` + Table `data_retention_time_in_days`
+To make data retention fields truly optional (previously they were producing plan every time when no value was set),
 we added `-1` as a possible value as set it as default. That got rid of the unexpected plans when no value is set and added possibility to use default value assigned by Snowflake (see [the data retention period](https://docs.snowflake.com/en/user-guide/data-time-travel#data-retention-period)).
```
sfc-gh-asawicki, line 16: should it be `and set it as default` instead of `as set it as default.`?

I also had a thought recently, that we could add a issue link to give more context to the migrations.

pkg/resources/helpers_test.go
sfc-gh-asawicki: Okay, I meant to reuse the same method or at least copy the existing one (because it is currently not importable), so that we do not have too many methods doing the exact same thing. Please just leave a TODO comment with SNOW-936093 and add it to the description of the issue (to merge this two eventually).

pkg/resources/table_acceptance_test.go
```diff
@@ -1475,6 +1480,175 @@ func TestAcc_Table_MaskingPolicy(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: testAccCheckTableDestroy,
+		Steps: []resource.TestStep{
+			{
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_Table_DefaultDataRetentionTime/WithDatabaseDataRetentionSet"),
```
sfc-gh-asawicki, line 1520: one last thought... We should check setting explicit -1 (to -1 and from -1)

sfc-gh-asawicki (approved):
Let's apply requested changes in the following PR (will be easier to review).

sfc-gh-asawicki (approved):

Issue #1926, opened 2023-07-05
snowflake_tag_association resource unexpected behavior and error on Snowflake column object with "." in name 


sbhattacharjeeClearcover:
**Provider Version**

v0.64.0

**Terraform Version**

v1.1.6

**Describe the bug**

When trying to create a snowflake_tag_association resource for a Snowflake column name with "." (e.g. **"test.column1")** in the "name" section of "object_identifier" in the resource, an error is seen as shown below:

│ Error: error associating tag to object: ["<database>"."<schema>"."<table>.'**test.column1**'"] with command: [ALTER TABLE "<database>"."<schema>"."<table>" MODIFY COLUMN "**'test**" SET TAG "<database>"."<schema>"."<tag_name>" = '<tag_value>'], tag_id [<database>|<schema>|<tag_name>]
│ 
│   with snowflake_tag_association.<resource_name>,
│   on tag_associate.tf line 1, in resource "snowflake_tag_association" "<resource_name>":
│    1: resource "snowflake_tag_association" "<resource_name>" {
│ 

The ALTER TABLE command being executed when the snowflake_tag_association resource is applied, does not consider the entire Snowflake column name, but stops before "." character. For example "**test.column1**" is taken as just "**test**" in the alter query

**Expected behavior**

The ALTER table query should be as follows:
`ALTER TABLE "<database>"."<schema>"."<table>" MODIFY COLUMN "**test.column1**" SET TAG "<database>"."<schema>"."<tag_name>" = '<tag_value>'], tag_id [<database>|<schema>|<tag_name>]`

**Code samples and commands**

The snowflake_tag_association sample being used is:

resource "snowflake_tag_association" "tag_column_association" {
    provider           = snowflake.sysadmin
    tag_id             = snowflake_tag.<tag_resource_name>.id
    object_type        = "COLUMN"
    tag_value          = "<tag_value>"
    object_identifier {
        database         = "test_db"
        schema           = "test_schema"
        name             = "test_table.'test.column1'"
    }
}

**Additional context**

Other variations of the column name in the object_identifier were also tested as listed below but the error wasn't resolved:

- "test_table.\"test.column1\""
- "test_table.\'test.column1\'"
- "'test_table'.'test.column1'"
- ""'test_table'.'test.column1'""

sfc-gh-swinkler:
Thank you for reporting this issue. I agree it is a problem, and we will look into it.

VickyWinner:
[...]

Issue #1910, opened 2023-06-26
After using snowflake_tag_association to set tag on account, subsequent terraform plan/apply fail with "error listing tag associations".


pstDnb:
**Provider Version**

0.64.0

**Terraform Version**

1.3.9

**Describe the bug**

The first attempt to set tags on the account using `snowflake_tag_association` succeeds. However, afterwards any attempt to run `terraform plan` or `apply` fail with `error listing tag associations`, because the `GET_TAG` function is invoked using account name (e.g  `SANDPIT01_AWS_EW1_DEV` instead of account locator (e.g. `KP44023`).

Example of failing SQL query:
```
SELECT SYSTEM$GET_TAG('"LIBRARY_FEATURE_PAALDEV"."TAGS"."ACCOUNT_TYPE"', '"SANDPIT01_AWS_EW1_DEV"', 'ACCOUNT') TAG_VALUE WHERE TAG_VALUE IS NOT NULL;
```
The above query succeeds if `SANDPIT01_AWS_EW1_DEV `is replaced with `KP44023`.

It is not an option to use account locator in `object_identifier`, since the SQL generated to set the tag explicitly mentions the account:
```
ALTER ACCOUNT "SANDPIT01_AWS_EW1_DEV" SET TAG "LIBRARY_FEATURE_PAALDEV"."TAGS"."ACCOUNT_GROUP" = 'SANDPIT01';
```
(The Snowflake documentation specifies a syntax without account name for 'ALTER ACCOUNT SET TAG', and even though it works
with account name, the same query fails when replacing account name with account locator.)

**Expected behavior**

When listing tag associations on the account (during `plan` or `apply`), the provider should use a query that correctly retrieves the tag.

**Code samples and commands**

Please add code examples and commands that were run to cause the problem.

**Additional context**

sfc-gh-swinkler:
@pstDnb thank you for reporting this issue. We will look into it and get back to you.

jia-ying-lin:
We have same error using snowflake_tag_association to set tag on VIEW. It's working fine on TABLE.

sfc-gh-asawicki:
Hey @jia-ying-lin, @pstDnb. We will address this issue as part of https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#preparing-essential-ga-objects-for-the-provider-v1.

sfc-gh-jmichalak:
Hi @pstDnb  👋

[...]

Issue #1909, opened 2023-06-26
When a snowflake_tag_association has multiple object_identifier blocks, only the first block is applied.


pstDnb:
**Provider Version**

0.64.0

**Terraform Version**

1.3.9

**Describe the bug**

When trying to use `snowflake_tag_association` to set a tag value on multiple objects of the same type, the tag is only set for the first object listed. This behavior was observed for the object types `WAREHOUSE` and `ROLE`, but the problem is probably the same for other object types.

**Expected behavior**

The tag value should be set for all objects listed in `object_identifier` blocks under `snowflake_tag_association`.

**Code samples and commands**

In the output from terraform plan I can see that the following `WAREHOUSE` objects will be tagged with `DOMAIN_PREFIX = 'DPINFRA'`:
 
 ```
# module.main_aws_ew1_dev.snowflake_tag_association.tag_associations["----.----.ADMIN_WH_PROD-ADMIN_WH_NONPROD-SHARED_XSMALL_WH-SHARED_SMALL_WH-SHARED_MEDIUM_WH.WAREHOUSE.LIBRARY.TAGS.DOMAIN_PREFIX"] will be created
  + resource "snowflake_tag_association" "tag_associations" {
      + id              = (known after apply)
      + object_type     = "WAREHOUSE"
      + skip_validation = true
      + tag_id          = "LIBRARY.TAGS.DOMAIN_PREFIX"
      + tag_value       = "DPINFRA"
      + object_identifier {
          + name = "ADMIN_WH_PROD"
        }
      + object_identifier {
          + name = "ADMIN_WH_NONPROD"
        }
      + object_identifier {
          + name = "SHARED_XSMALL_WH"
        }
      + object_identifier {
          + name = "SHARED_SMALL_WH"
        }
      + object_identifier {
          + name = "SHARED_MEDIUM_WH"
        }
    }
```

In the output from `terraform apply`, the `snowflake_tag_association` seems to be successfully applied:

```
module.main_aws_ew1_dev.snowflake_tag_association.tag_associations["----.----.ADMIN_WH_PROD-ADMIN_WH_NONPROD-SHARED_XSMALL_WH-SHARED_SMALL_WH-SHARED_MEDIUM_WH.WAREHOUSE.LIBRARY.TAGS.DOMAIN_PREFIX"]: Creating...
[...]

Issue #1202, opened 2022-09-05
snowflake_tag_association does not support fully qualified object names


ogr003:
**Provider Version**

0.43.0

**Terraform Version**

1.2.1

**Describe the bug**

The snowflake_tag_association resource provides limited utility for objects that are not at account-level, as it does not support using fully qualified names. For instance, for a table named `DB.SCHEMA.TABLE_NAME` it is only possible to pass `TABLE_NAME` as identifier, which will only work if you have the correct context. This is impractical when using a single provider to configure several databases or schemas.

**Expected behavior**

The resource should support passing database, schema, table (only supported for column) and name as separate attributes, or parse the attributes similarly to what is done for the `tag_id` attribute.

**Code samples and commands**

```
# TABLE
resource "snowflake_tag_association" "tag_associations" {
    object_name     = "DB.SCHEMA.TABLE_NAME"
    object_type     = "TABLE"
    tag_id          = "TAG_DB.TAG_SCHEMA.TAG_NAME"
    tag_value       = "TEST"
}
```

**Additional context**

The first statement above executes `ALTER TABLE "DB.SCHEMA.TABLE_NAME" SET TAG "TAG_DB"."TAG_SCHEMA"."TAG_NAME" = "TEST"` which would be correct if the table was named `DB.SCHEMA.TABLE_NAME`, but makes it impossible to add a tag for a table named `TABLE_NAME` in the `DB.SCHEMA` schema without specifying database and schema in the provider.

PR #2534, opened 2024-02-21
fix: tag identifiers problems

sfc-gh-swinkler:
fixes:
#1926 
#1910
#1909 
#1202


pkg/resources/tag_association_acceptance_test.go
```diff
@@ -149,3 +246,131 @@ resource "snowflake_tag_association" "columnTag" {
...
+		name = "${snowflake_table.test_table.name}"
+	}
+	object_type = "TABLE"
+	tag_id      = snowflake_tag.test_tag.id
+	tag_value   = "v1"
+}
+`, tagName, tableName, databaseName, schemaName)
+}
+
+func tagAssociationConfigAccountIssues1910(tagName, accountName string, databaseName string, schemaName string) string {
```
sfc-gh-jcieslak, line 349: As far as I remember ORGADMIN role is required for this (?). It's kind of connected to SNOW-1002025, maybe there should be another ticket or some reference created so we don't forget about this.

pkg/resources/tag_masking_policy_association.go
```diff
@@ -102,98 +85,70 @@ func TagMaskingPolicyAssociation() *schema.Resource {
...
 		}
 		defer func() {
-			err := client.Warehouses.Drop(ctx, tempWarehouseID, nil)
-			if err != nil {
+			if err := client.Warehouses.Drop(ctx, wid, nil); err != nil {
 				log.Printf("[WARN] error cleaning up temp warehouse %v", err)
 			}
-			err = client.Sessions.UseWarehouse(ctx, sdk.NewAccountObjectIdentifier(originalWarehouse))
-			if err != nil {
+			if err := client.Sessions.UseWarehouse(ctx, sdk.NewAccountObjectIdentifier(warehouse)); err != nil {
```
sfc-gh-jcieslak, line 140: this will always throw an error because this will only happen when `warehouse == ""`

pkg/resources/tag_acceptance_test.go
```diff
@@ -1,45 +1,70 @@
...
 			{
-				Config: tagConfig(accName, acc.TestDatabaseName, acc.TestSchemaName),
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_Tag/basic"),
+				ConfigVariables: m(),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "name", name),
+					resource.TestCheckResourceAttr(resourceName, "database", acc.TestDatabaseName),
+					resource.TestCheckResourceAttr(resourceName, "schema", acc.TestSchemaName),
+					resource.TestCheckResourceAttr(resourceName, "allowed_values.#", "2"),
+					resource.TestCheckResourceAttr(resourceName, "allowed_values.0", "alv1"),
```
sfc-gh-jcieslak, line 43: forgot second value

pkg/sdk/tags_validations.go
```diff
@@ -144,3 +141,25 @@ func (opts *undropTagOptions) validate() error {
 	}
 	return errors.Join(errs...)
 }
+
+func (opts *setTagOptions) validate() error {
+	if opts == nil {
+		return errors.Join(ErrNilOptions)
+	}
+	var errs []error
```
sfc-gh-jcieslak, line 149: do we want to validate what objectTypes can be called with this function or we let it fail on Snowflake level?

sfc-gh-jcieslak (changes requested):

pkg/resources/tag_association.go
```diff
@@ -101,132 +103,166 @@ func TagAssociation() *schema.Resource {
...
-	if d.HasChange("skip_validation") {
-		o, n := d.GetChange("skip_validation")
-		log.Printf("[DEBUG] skip_validation changed from %v to %v", o, n)
+	tid, ids, ot := tagIdentifierAndObjectIdentifier(d)
+	for _, oid := range ids {
+		tagValue, err := client.SystemFunctions.GetTag(ctx, tid, oid, ot)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set("tag_value", tagValue); err != nil {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: it wasn't meant to support a list of more than one. i was actually thinking to deprecate object_identifiers field for this very reason.

pkg/resources/tag.go
```diff
@@ -150,146 +91,111 @@ func Tag() *schema.Resource {
...
-
-	t, err := snowflake.ScanTag(row)
-	if errors.Is(err, sql.ErrNoRows) {
-		// If not found, mark resource to be removed from state file during apply or refresh
-		log.Printf("[DEBUG] tag (%s) not found", d.Id())
-		d.SetId("")
-		return nil
-	}
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.SchemaObjectIdentifier)
+	request := sdk.NewShowTagRequest().WithIn(&sdk.In{Schema: sdk.NewDatabaseObjectIdentifier(id.DatabaseName(), id.SchemaName())}).WithLike(id.Name())
```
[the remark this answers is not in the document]
sfc-gh-swinkler: fixed

pkg/resources/tag.go
```diff
@@ -150,146 +91,111 @@ func Tag() *schema.Resource {
...
-			q := builder.RemoveAllowedValues()
-
-			if err := snowflake.Exec(db, q); err != nil {
-				return fmt.Errorf("error removing ALLOWED_VALUES for tag %v", tag)
+		v, ok := d.GetOk("allowed_values")
+		if ok {
+			allowedValues := expandAllowedValues(v)
+			// unset the allowed values
+			unset := sdk.NewTagUnsetRequest().WithAllowedValues(true)
+			if err := client.Tags.Alter(ctx, sdk.NewAlterTagRequest(id).WithUnset(unset)); err != nil {
```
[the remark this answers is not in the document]
sfc-gh-swinkler: i agree, it makes sense to do it this way. its safer and doesn't change the intended behavior.

pkg/resources/tag_association_acceptance_test.go
```diff
@@ -73,6 +74,102 @@ func TestAcc_TagAssociationColumn(t *testing.T) {
...
+		Providers:    acc.TestAccProviders(),
+		PreCheck:     func() { acc.TestAccPreCheck(t) },
+		CheckDestroy: nil,
+		Steps: []resource.TestStep{
+			{
+				Config: tagAssociationConfigColumnIssues1909(tagName, table1Name, table2Name, acc.TestDatabaseName, acc.TestSchemaName, columnName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_tag_association.tag_column_association", "object_type", "COLUMN"),
+					resource.TestCheckResourceAttr("snowflake_tag_association.tag_column_association", "tag_id", fmt.Sprintf("%s|%s|%s", acc.TestDatabaseName, acc.TestSchemaName, tagName)),
+					resource.TestCheckResourceAttr("snowflake_tag_association.tag_column_association", "tag_value", "v1"),
```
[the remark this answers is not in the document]
sfc-gh-swinkler: good idea. done

pkg/resources/tag_association_acceptance_test.go
```diff
@@ -149,3 +246,131 @@ resource "snowflake_tag_association" "columnTag" {
...
+	database            = "%[3]v"
+	schema              = "%[4]v"
+	column {
+		name    = "test_column"
+		type    = "VARIANT"
+	}
+}
+resource "snowflake_tag_association" "tag_column_association" {
+	object_identifier {
+		name = "${snowflake_table.test_table.name}"
```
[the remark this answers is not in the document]
sfc-gh-swinkler: right, as per discussion above, object_name was marked as un-deprecated for this very reason

pkg/resources/tag_association_acceptance_test.go
sfc-gh-swinkler: added a ticket

pkg/resources/tag_masking_policy_association.go
sfc-gh-swinkler: removed. when wh is dropped, it will remove current warehouse anyways

pkg/resources/tag_acceptance_test.go
sfc-gh-swinkler: fixed

pkg/sdk/tags_validations.go
sfc-gh-swinkler: i think its fine to let it fail on Snowflake level.

MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,13 @@ This document is meant to help you migrate your Terraform config to the new newe
 describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behavior
 across different versions.
 
+## v0.87.0 ➞ v0.88.0
+### snowflake_tag_association resource changes
+#### *(behavior change)* `object_name`
+No longer deprecated.
+#### *(behavior change)* `object_identifier`
+No longer required.
```
sfc-gh-jcieslak, line 12: Could you elaborate a bit? Why we decided to make this change and what should users use after this change, `object_name` or `object_identifier`? Also, If `object_name` is no longer deprecated then users should be able to use it. Right now, I don't think it's used anywhere else than in schema.

sfc-gh-swinkler: reverted change. This PR already has a lot of changes, so probably best not to add even more changes. But I think that the object_identifier struct was not a good idea, and should never have been implemented in the first place. We should only allow for fully-qualified-names as we do for many other resources. However, this kind of change will impact other resources as well which should be changed at the same time.

pkg/resources/tag_masking_policy_association_acceptance_test.go
```diff
@@ -29,6 +31,68 @@ func TestAcc_TagMaskingPolicyAssociation(t *testing.T) {
 	})
 }
 
+func TestAcc_TagMaskingPolicyAssociation1(t *testing.T) {
```
sfc-gh-asawicki, line 34: nit-pick: `TestAcc_TagMaskingPolicyAssociation1` -> this test deserves a better name than just an ordinal

pkg/sdk/system_functions.go
```diff
@@ -7,6 +7,7 @@ import (
 
 type SystemFunctions interface {
 	GetTag(ctx context.Context, tagID ObjectIdentifier, objectID ObjectIdentifier, objectType ObjectType) (string, error)
+	GetTags(ctx context.Context, tagID ObjectIdentifier, objectID ObjectIdentifier, objectType ObjectType) ([]string, error)
```
sfc-gh-asawicki, line 10: please add integration tests in `system_functions_integration_test.go`

pkg/sdk/system_functions.go
```diff
@@ -26,3 +27,18 @@ func (c *systemFunctions) GetTag(ctx context.Context, tagID ObjectIdentifier, ob
 	}
 	return s.Tag, nil
 }
+
+func (c *systemFunctions) GetTags(ctx context.Context, tagID ObjectIdentifier, objectID ObjectIdentifier, objectType ObjectType) ([]string, error) {
+	res := []struct {
+		Tag string `db:"TAG"`
+	}{}
+	sql := fmt.Sprintf(`SELECT SYSTEM$GET_TAG('%s', '%s', '%s') TAG_VALUE WHERE TAG_VALUE IS NOT NULL`, tagID.FullyQualifiedName(), objectID.FullyQualifiedName(), objectType)
```
sfc-gh-asawicki, line 35: Does this method even work? `Tag string db:"TAG"` does not match `TAG_VALUE`. And also, how does it differ from the one above? Is it possible to set multiple and check them like this?

pkg/resources/tag_association.go
```diff
@@ -101,132 +104,166 @@ func TagAssociation() *schema.Resource {
...
+		return true
+	default:
+		return false
 	}
+}
 
-	skipValidate := d.Get("skip_validation").(bool)
-	if !skipValidate {
-		log.Println("[DEBUG] validating tag creation")
+func tagIdentifierAndObjectIdentifier(d *schema.ResourceData) (sdk.SchemaObjectIdentifier, []sdk.ObjectIdentifier, sdk.ObjectType) {
```
sfc-gh-asawicki, line 116: This method is complicated and have multiple possible outputs. Please add unit test and validate different possibilities.

pkg/resources/testdata/TestAcc_TagAssociation/issue1202/main.tf
```diff
@@ -0,0 +1,22 @@
...
+    name = "test_column"
+    type = "VARIANT"
+  }
+}
+
+resource "snowflake_tag_association" "test" {
+  object_name = "\"${var.database}\".\"${var.schema}\".\"${var.table_name}\""
+  object_type = "TABLE"
+  tag_id      = snowflake_tag.test.id
+  tag_value   = "v1"
```
sfc-gh-asawicki, line 21: `depends_on` is missing (we require `snowflake_table` to be created first)

pkg/resources/tag_association_acceptance_test.go
```diff
@@ -1,151 +1,283 @@
...
 
-resource "snowflake_tag_association" "columnTag" {
-	object_identifier {
-		database            = "%[3]v"
-		schema              = "%[4]v"
-		name     = "${snowflake_table.test_table.name}.${snowflake_table.test_table.column[0].name}"
+func TestAcc_TagAssociationAccountIssues1910(t *testing.T) {
+	// todo: use role with ORGADMIN in CI (SNOW-1165821)
+	// SNOWFLAKE_TEST_ACCOUNT_CREATE must be set to 1 to run this test
+	if _, ok := os.LookupEnv("SNOWFLAKE_TEST_ACCOUNT_CREATE"); !ok {
```
sfc-gh-asawicki, line 210: please use new `_ = testenvs.GetOrSkipTest(t, testenvs.TestAccountCreate)`

pkg/resources/tag_association_acceptance_test.go
```diff
@@ -1,151 +1,283 @@
...
+					resource.TestCheckResourceAttr(resourceName, "tag_id", fmt.Sprintf("%s|%s|%s", acc.TestDatabaseName, acc.TestSchemaName, tagName)),
+					resource.TestCheckResourceAttr(resourceName, "tag_value", "v1"),
+				),
+			},
+		},
+	})
 }
-`, n1, n2, databaseName, schemaName)
+
+func TestAcc_TagAssociationIssue1926(t *testing.T) {
```
sfc-gh-asawicki, line 246: How it will behave if we set the table name to "table.test" and column to "column" or column to "test.column"?

sfc-gh-asawicki (changes requested):

pkg/snowflake/tag.go
```diff
@@ -5,170 +5,33 @@ import (
...
 
 // TagBuilder abstracts the creation of SQL queries for a Snowflake tag.
 type TagBuilder struct {
-	name                 string
-	db                   string
-	schema               string
-	comment              string
-	allowedValues        string
-	maskingPolicyBuilder *MaskingPolicyBuilder
+	tagId           sdk.SchemaObjectIdentifier
```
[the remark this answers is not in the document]
sfc-gh-swinkler: SNOW-1205579
created a jira ticket for this work. basically, the scope of this PR already is so big that I had to draw the line in the sand somewhere. it already expanded beyond just tag associations to also tags in general., the masking policy tag association has nothing to do with any of the issues addressed in this PR. it wouldn't be too difficult to refactor it, but also I wasn't sure where to put it. Hence, leaving it for another day.

pkg/sdk/tags_impl.go
```diff
@@ -33,7 +33,10 @@ func (v *tags) Show(ctx context.Context, request *ShowTagRequest) ([]Tag, error)
 }
 
 func (v *tags) ShowByID(ctx context.Context, id SchemaObjectIdentifier) (*Tag, error) {
-	request := NewShowTagRequest().WithLike(id.Name())
+	request := NewShowTagRequest().WithIn(&In{
+		Schema: NewDatabaseObjectIdentifier(id.DatabaseName(), id.SchemaName()),
```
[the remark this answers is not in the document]
sfc-gh-swinkler: is there an ask here?

pkg/sdk/tags_test.go
```diff
@@ -322,3 +297,97 @@ func TestTagAlter(t *testing.T) {
...
+				Name:  NewAccountObjectIdentifier("tag1"),
+				Value: "value1",
+			},
+		}
+		assertOptsValidAndSQLEquals(t, opts, `ALTER %s %s SET TAG "tag1" = 'value1'`, opts.objectType, id.FullyQualifiedName())
+	})
+
+	t.Run("set with column", func(t *testing.T) {
+		table, column := NewSchemaObjectIdentifier("db1", "schema1", "table1"), "column1"
+		objectName := NewObjectIdentifierFromFullyQualifiedName(fmt.Sprintf("%s.%s.%s.%s", table.DatabaseName(), table.SchemaName(), table.Name(), column))
```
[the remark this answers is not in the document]
sfc-gh-swinkler: its really very useful in this case, as we need to do:

`ALTER TABLE "db1"."schema1"."table1" MODIFY COLUMN "column1" SET TAG "tag1" = 'value1'`

which as you can see, does not follow the table column identifier fully qualified name. I added it anyways, because its not a lot of work

pkg/sdk/tags_test.go
```diff
@@ -322,3 +297,97 @@ func TestTagAlter(t *testing.T) {
...
+		request := SetTagRequest{
```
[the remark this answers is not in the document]
sfc-gh-swinkler: okay fixed

pkg/resources/tag_masking_policy_association_acceptance_test.go
sfc-gh-swinkler: renamed

pkg/sdk/system_functions.go
sfc-gh-swinkler: copied the logic as it was in the Snowflake package. i wasn't sure if the len of tags was necessary for retry logic, such as in the case where the tag association hasn't been created, but no error. I tried using the GetTag by itself, and it seems to work fine. so I have removed the GetTags code

pkg/sdk/system_functions.go
sfc-gh-swinkler: see below

pkg/resources/testdata/TestAcc_TagAssociation/issue1202/main.tf
sfc-gh-swinkler: name     = snowflake_table.test.name
this is already an implicit dependency. actually you do not need to do depends_on

pkg/resources/tag_association_acceptance_test.go
sfc-gh-swinkler: okay

pkg/resources/tag_association_acceptance_test.go
sfc-gh-swinkler: Get a lot of issues with Table resource. It simply cannot handle identifiers or columns with these names. Has nothing to do with tag or tag association. First I just added the 

```
query ALTER TABLE "terraform_test_database"."terraform_test_schema"."table" MODIFY COLUMN "test.test.column" SET TAG "terraform_test_database"."terraform_test_schema"."tag-XLKK" = 'v1' err 002003 (42S02): SQL compilation error:
Table '"terraform_test_database"."terraform_test_schema"."table"' does not exist or not authorized.
```

and also:
```
2024/03/04 22:54:18 [DEBUG] sql-conn-exec: [query ALTER TABLE "terraform_test_database"."terraform_test_schema"."table.test" DROP COLUMN "test.column" err 001432 (42601): SQL compilation error:
cannot drop all the columns of a table duration 109.416409ms args {}] ()
```

pkg/resources/tag_association.go
sfc-gh-swinkler: added unit tests

pkg/sdk/tags_impl.go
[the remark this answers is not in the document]
sfc-gh-asawicki: No, I like that fix :)

pkg/resources/testdata/TestAcc_TagAssociation/issue1202/main.tf
sfc-gh-asawicki: yep, but it weren't there, so any dependency was indeed needed

pkg/resources/testdata/TestAcc_TagAssociation/issue1202/main.tf
```diff
@@ -0,0 +1,31 @@
...
+  schema   = var.schema
+  column {
+    name = "test_column"
+    type = "VARIANT"
+  }
+}
+
+resource "snowflake_tag_association" "test" {
+  // we need to set the object_identifier to avoid the following error:
+  // provider_test.go:17: err: resource snowflake_tag_association: object_identifier: Optional or Required must be set, not both
```
sfc-gh-asawicki, line 19: wasn't that error the result of incorrect schema where you had both Optional and Required? https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2534#discussion_r1508623139

sfc-gh-swinkler: yes, but also @sfc-gh-jcieslak wanted to not make any breaking changes by having object_identifier be optional (I think it should be deprecated actually). Also there is the fact that object_name wasn't even being used internally. Since no customers are using this, i have reverted to previous behavior

sfc-gh-asawicki (approved):

Issue #2533, opened 2024-02-20
External Volume - Terraform not reading privileges properly


caseywhoop:
### Terraform CLI and Provider Versions
```bash
Terraform v1.7.3
on darwin_arm64
+ provider registry.terraform.io/snowflake-labs/snowflake v0.85.0
```

### Terraform Configuration

```terraform
resource "snowflake_grant_privileges_to_account_role" "grant_external_volume_iceberg_prod_external_volume" {
  privileges = ["USAGE"]
  account_role_name  = "ACCOUNTING_ROLE"

  on_account_object {
    object_type = "EXTERNAL VOLUME"
    object_name = "ICEBERG_PROD_EXTERNAL_VOLUME"
  }
}
```

### Expected Behavior

1 - We expect there to be no diff after applying changes, confirming resource is properly configured in state file, and running a terraform plan.

2 - We expect terraform to delete the resource after removing the code from the repo and re-applying.

### Actual Behavior

1.
Running `terraform plan` still shows:

```sh
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.roles.module.roles.module.accounting_role.snowflake_grant_privileges_to_account_role.grant_external_volume_iceberg_prod_external_volume will be updated in-place
  ~ resource "snowflake_grant_privileges_to_account_role" "grant_external_volume_iceberg_prod_external_volume" {
        id                = "\"ACCOUNTING_ROLE\"|false|false|USAGE|OnAccountObject|EXTERNAL VOLUME|\"ICEBERG_PROD_EXTERNAL_VOLUME\""
      ~ privileges        = [
          + "USAGE",
        ]
        # (5 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

Re-applying this will not fix resolve the diff; It will always try to re-add `"USAGE"`

2.
Removing the resource from the repo shows the incorrect changes to be applied (note, `USAGE` privilege is missing:
```bash
Terraform will perform the following actions:

  # module.roles.module.roles.module.accounting_role.snowflake_grant_privileges_to_account_role.grant_external_volume_iceberg_prod_external_volume will be destroyed
  # (because snowflake_grant_privileges_to_account_role.grant_external_volume_iceberg_prod_external_volume is not in configuration)
[...]

PR #2538, opened 2024-02-21
fix: grants on external volumes

sfc-gh-jcieslak:
Fixes: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2533
Terraform couldn't read privileges for External volumes, because Snowflake returns a shorter `VOLUME` name where we expect `EXTERNAL VOLUME` to be returned. The proposed fix relies on replacing `EXTERNAL VOLUME` with `VOLUME` in the prepare read request function so that the Read operation will be untouched and will work for external volumes.

## Test Plan

* [x] acceptance tests to show the fix works (didn't pass without the fix)

## Other
Wrote to the docs team to add this case to the SHOW GRANTS page.

**Update**: added missing privileges (CREATE MODEL needed for https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2563)


pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -815,6 +815,11 @@ func prepareShowGrantsRequestForAccountRole(id GrantPrivilegesToAccountRoleId) (
 	case OnAccountObjectAccountRoleGrantKind:
 		data := id.Data.(*OnAccountObjectGrantData)
 		grantedOn = data.ObjectType
+
+		// For EXTERNAL VOLUME, we have to change it to VOLUME because that's what Snowflake returns
```
sfc-gh-asawicki, line 819: We have recently fixed similar issue on the SDK level: #2522. Shouldn't we also fix it here on the SDK level (by modifying `func (row grantRow) convert() *Grant {` function)?

pkg/resources/grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -1065,3 +1110,28 @@ func queriedAccountRolePrivilegesContainAtLeast(roleName sdk.AccountObjectIdenti
 		})
 	}, roleName, privileges...)
 }
+
+func createExternalVolumeTemporarily(t *testing.T, externalVolumeName string) {
```
sfc-gh-asawicki, line 1114: Why the `temporarily`?

sfc-gh-asawicki (changes requested):

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jcieslak: hmmm, true. Done

pkg/resources/grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: removed `temporarily` and returned cleanup, so it's consistent with `testint` package utils

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-asawicki: I don't understand why :(

sfc-gh-jcieslak: I think I was thinking about mapping in a wrong way that's why I thought it couldn't be done, but with mapping VOLUME to EXTERNAL VOLUME can be done in SDK. Moved to the SDK.

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

Issue #2519, opened 2024-02-17
COPY GRANTS option fails on new views


scastria:
### Terraform CLI and Provider Versions

TF v1.5.2
Snowflake TF 0.80.0

### Terraform Configuration

```terraform
resource "snowflake_view" "CustomerView" {
  provider = snowflake.READER_ACCOUNTADMIN
  for_each = local.customer_views
  database = snowflake_database.CustomerDatabase[split(".", each.key)[0]].name
  schema = snowflake_schema.CustomerSchema[join(".", [split(".", each.key)[0], split(".", each.key)[1]])].name
  name = split(".", each.key)[2]
  statement = <<-SQL
    -- Force a change to the view so that it is recompiled ${timestamp()}
    SELECT * FROM DB_SHARE.${split(".", each.key)[1]}.${split(".", each.key)[2]}
  SQL
  is_secure = true
  copy_grants = true
}
```

### Expected Behavior

View is created successfully

### Actual Behavior

verbose TF_LOG shows that the CREATE VIEW statement includes the COPY GRANTS clause since copy_grants = true above.  However, since this is a brand new view and NOT a CREATE OR REPLACE, COPY GRANTS is invalid and you get a SQL error of: "Invalid operation COPY GRANTS without specifying source object".  Therefore, if the Snowflake provider knows this is a brand new view, then it should omit COPY GRANTS even if copy_grants = true so that valid SQL is produced.

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

Medium

sfc-gh-asawicki:
Also, I see you are using "." splitting in your configurations. Snowflake identifiers are a bit more complex; it's possible to have the dot character as part of the identifier if you have a fully qualified part (wrapped in double quotes). We have the identifiers rework incoming. We want to address the issues our users have with identifiers.

This is just a FYI because these identifiers can really hurt sometimes.

sfc-gh-asawicki:
Hey @scastria . We have released the fix as part of [v0.87.0](https://github.com/Snowflake-Labs/terraform-provider-snowflake/releases/tag/v0.87.0) release. Please follow the [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/v0.87.0/MIGRATION_GUIDE.md#v0860--v0870) during the update. Please confirm that the issue is resolved in the newest version. Thanks!

sfc-gh-asawicki:
[...]

Issue #2495, opened 2024-02-13
Issue with "cluster_by" in snowflake_table when written to state file


JESCHO99:
### Terraform CLI and Provider Versions

0.85.0

### Terraform Configuration

```terraform
resource "snowflake_table" "test" {

  database            = "TEST"
  schema              = "TEST"
  name                = "TEST"
  data_retention_time_in_days = 1
  change_tracking     = false

  cluster_by = [
    "date_trunc('month', LAST_LOAD_TIME)",
    "STATUS"
  ]

  column {
    name     = "LAST_LOAD_TIME"
    type     = "TIMESTAMP_LTZ(6)"
    nullable = true
  }

  column {
    name     = "STATUS"
    type     = "VARCHAR(16777216)"
    nullable = true
  }

}
```

### Expected Behavior

Expression in the cluster_by parameter is written at once in the state file and is not splitted into several part at every "," in the list.

### Actual Behavior

The expression is split on every "," which leads to a situation where the table plans changes for the cluster by column in every plan which is executed
![image](https://github.com/Snowflake-Labs/terraform-provider-snowflake/assets/59776328/bfbdbdab-0dec-4447-87b3-66a37210ac1d)

### Steps to Reproduce

1. terraform apply
2. terraform apply

### How much impact is this issue causing?

High

sfc-gh-asawicki:
Hey @JESCHO99. Thanks for creating the issue.

We are aware of that problem, and we have it on our roadmap. I hope we will address it soon.

sfc-gh-asawicki:
Hey @JESCHO99. We have released the fix as part of [v0.87.0](https://github.com/Snowflake-Labs/terraform-provider-snowflake/releases/tag/v0.87.0) release. Please follow the [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/v0.87.0/MIGRATION_GUIDE.md#v0860--v0870) during the update. Please confirm that the issue is resolved in the newest version. Thanks!

sfc-gh-jmichalak:
Closing due to inactivity.

Issue #2110, opened 2023-10-10
Parsing error of cluster_by value containing an expression


buwi:
**Provider Version**

0.72.0

**Terraform Version**

v1.5.2

**Describe the bug**

I am using Terragrunt to manage state in AWS S3. It appears, the Snowflake provider is parsing the cluster_by value in the wrong place when it contains an expression. 

For a table definition:

```
resource "snowflake_table" "sample-table" {
  database        = snowflake_schema.mydb-my_schema.database
  schema          = snowflake_schema.mydb-my_schema.name
  name              = local.sample-table_name
  cluster_by      = ["date_trunc('HOUR',TIMESTAMP_HR)","CLIENT_ID"]
  comment        = "..."
  change_tracking = false

  column {
    name     = "CLIENT_ID"
    type     = "VARCHAR(16777216)"
  }
  
  column {
    name     = "TIMESTAMP_HR"
    type     = "TIMESTAMP_TZ(9)"
  }
  
...
```

The initial run of terraform created the table successfully, with the correct cluster_by definition. 

`
CLUSTERING_KEY: LINEAR(date_trunc('HOUR', TIMESTAMP_HR), CLIENT_ID)
`

But the state file contained the cluster_by as follows:
```
            "cluster_by": [
              "date_trunc('HOUR'",
              "TIMESTAMP_HR)",
              "CLIENT_ID"
            ],
```

On a subsequent run of Terraform, the cluster_by was identified as a change.

```
  
  Terraform will perform the following actions:

  # snowflake_table.sample-table will be updated in-place
  ~ resource "snowflake_table" "sample-table" {
      ~ cluster_by      = [
          - "date_trunc('HOUR'",
          - "TIMESTAMP_HR)",
          + "date_trunc('HOUR',TIMESTAMP_HR)",
            "CLIENT_ID",
        ]
        id              = "MYDB|MY_SCHEMA|MY_TABLE"
        name            = "MY_TABLE"
        # (6 unchanged attributes hidden)

        # (36 unchanged blocks hidden)
    }
```

**Expected behavior**

The state file should store the cluster_by definition as follows.

```
            "cluster_by": [
              "date_trunc('HOUR',TIMESTAMP_HR)",
              "CLIENT_ID"
            ],
```

**Code samples and commands**

**Additional context**

```
sample-table_name  ="MY_TABLE"
```

meinczinger:
[...]

PR #2552, opened 2024-02-23
fix: Fix various small problems

sfc-gh-asawicki:
- Fixed clustering for table_resource (clustering with nested functions are parsed correctly now)
- Adjusted view resource (`copy_grants` require `or_replace`) and corrected the documentation

References: #2110 #2495 #2519

### DIFF
--- a/pkg/resources/table_acceptance_test.go
+++ b/pkg/resources/table_acceptance_test.go
@@ -1537,3 +1537,50 @@ func testAccCheckTableDestroy(s *terraform.State) error {
 	}
 	return nil
 }
+
+// proves issues https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2110 and https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2495
+func TestAcc_Table_ClusterBy(t *testing.T) {
+	accName := strings.ToUpper(acctest.RandStringFromCharSet(10, acctest.CharSetAlpha))
+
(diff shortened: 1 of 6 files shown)


pkg/resources/table_acceptance_test.go
```diff
@@ -1537,3 +1537,50 @@ func testAccCheckTableDestroy(s *terraform.State) error {
...
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "cluster_by.#", "2"),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "cluster_by.0", "date_trunc('month', LAST_LOAD_TIME)"),
+					resource.TestCheckResourceAttr("snowflake_table.test_table", "cluster_by.1", "COL1"),
+				),
+			},
+		},
+	})
+}
+
+func tableConfigWithComplexClusterBy(name string, databaseName string, schemaName string) string {
```
sfc-gh-swinkler, line 1568: should this be moved into testdata directory?

sfc-gh-swinkler (approved):

sfc-gh-jcieslak (approved):

pkg/resources/table_acceptance_test.go
sfc-gh-asawicki: We did not agree to automatically move every config to a file - it is not necessarily more readable. Currently, we accept both options (they are both not ideal), but we have ideas for the better, third option (maybe coming soon).

Issue #2560, opened 2024-02-26
pipe must be replaced without any config change 


toni-moreno:
### Terraform CLI and Provider Versions

Terraform v1.5.0
on linux_amd64
+ provider registry.terraform.io/snowflake-labs/snowflake v0.86.0

### Terraform Configuration

```terraform
locals {

pipe_map = {
 "MY_DB.MY_SCHEMA.MY_PIPE_1" = {
    "auto_ingest" = true
    "copy_statement" = <<-EOT
    copy into MY_DB.MY_SCHEMA.MY_TABLE_1
    from @MY_DB.PUBLIC.EXTERNAL_STAGE
    PATTERN = 'some_directory/some_file.csv'
    file_format = (format_name = 'MY_SCHEMA.PUBLIC.CRM_CSV')
    
    EOT
    "db" = "MY_DB"
    "integration" = "MY_AZURE_NOTIF_INT"
    "pipe" = "MY_PIPE"
    "schema" = "MY_SCHEMA"
    "schema_id" = "MY_DB.MY_SCHEMA"
  }
}
}

resource "snowflake_pipe" "pipe" {
  for_each = local.pipe_map

  depends_on = [snowflake_table.table]

  provider = snowflake.SYSADMIN

  database       = snowflake_database.db[each.value.db].name
  schema         = each.value.schema == "PUBLIC" ? "PUBLIC" : snowflake_schema.schema[each.value.schema_id].name
  name           = each.value.pipe
  copy_statement = each.value.copy_statement
  auto_ingest    = each.value.auto_ingest
  integration    = each.value.integration

}
```

### Expected Behavior

One created if no changes in the config files, pipes never should be recreated.

### Actual Behavior

after created when I exec again (without any config change) "terraform apply"   this message is shown.

```bash
  # module.xxxx.snowflake_pipe.pipe["MY_DB.MY_SCHEMA.MY_PIPE_1"] must be replaced
-/+ resource "snowflake_pipe" "pipe" {
      ~ copy_statement       = <<-EOT # forces replacement
     copy into MY_DB.MY_SCHEMA.MY_TABLE_1
    from @MY_DB.PUBLIC.EXTERNAL_STAGE
    PATTERN = 'some_directory/some_file.csv'
    file_format = (format_name = 'MY_SCHEMA.PUBLIC.CRM_CSV')
        EOT
[...]

Issue #2544, opened 2024-02-22
Snowflake Failover Group Replication Schedule updates fail


gdelia-pm:
### Terraform CLI and Provider Versions

```
> .\terraform.exe version
Terraform v1.7.3
on windows_amd64
+ provider registry.terraform.io/snowflake-labs/snowflake v0.86.0

Your version of Terraform is out of date! The latest version
is 1.7.4. You can update by downloading from https://www.terraform.io/downloads.html
```

### Terraform Configuration

```terraform
resource "snowflake_failover_group" "SOURCE_FAILOVER_GROUP" {
  depends_on = [
    snowflake_database.DB,
    snowflake_role.FUNCTIONAL_ROLES,
    snowflake_warehouse.WH,
  ]

  name = "FAILOVER_GROUP1"
  object_types              = ["ACCOUNT PARAMETERS", "DATABASES", "INTEGRATIONS", "NETWORK POLICIES", "RESOURCE MONITORS", "ROLES", "USERS", "WAREHOUSES"]
  allowed_accounts          = ["TEST.CORE1", "TEST.CORE_DR1"]
  allowed_integration_types = ["SECURITY INTEGRATIONS", "STORAGE INTEGRATIONS"]
  allowed_databases         = toset([for db in snowflake_database.DB : db.name])

  replication_schedule {
    interval = 60 # minutes
  }
}

resource "snowflake_failover_group" "TARGET_FAILOVER_GROUP" {
  depends_on = [snowflake_failover_group.SOURCE_FAILOVER_GROUP]
  provider   = snowflake.dr
  name       = snowflake_failover_group.SOURCE_FAILOVER_GROUP.name
  from_replica {
    organization_name   = "TEST"
    source_account_name = "CORE1"
    name                = snowflake_failover_group.SOURCE_FAILOVER_GROUP.name
  }
}
```

### Expected Behavior

On initial creation the interval is set correct.  If I try to change the interval it fails
```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # snowflake_failover_group.SOURCE_FAILOVER_GROUP will be updated in-place
  ~ resource "snowflake_failover_group" "SOURCE_FAILOVER_GROUP" {
        id                        = "FAILOVER_GROUP1"
        name                      = "FAILOVER_GROUP1"
      ~ object_types              = [
          - "PARAMETERS",
[...]

PR #2562, opened 2024-02-27
fix: Fix failover group alter syntax and suppression for pipe statement

sfc-gh-asawicki:
- Run two separate alters in failover group resource
- Handle `.` in failover ids
- Trim leading and trailing space for pipe statement diff suppression
- Unskip part of the integration tests for failover groups

References: #2544 #2560

### DIFF
--- a/pkg/sdk/testint/failover_groups_integration_test.go
+++ b/pkg/sdk/testint/failover_groups_integration_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvs"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/internal/random"
 	"github.com/avast/retry-go"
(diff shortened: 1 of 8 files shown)


pkg/sdk/testint/failover_groups_integration_test.go
```diff
@@ -104,8 +138,83 @@ func TestInt_FailoverGroupsCreate(t *testing.T) {
...
+			sdk.IntegrationTypeAPIIntegrations,
+		}
+		newReplicationSchedule := "20 MINUTE"
+
+		// does not work together:
+		opts := &sdk.AlterSourceFailoverGroupOptions{
+			Set: &sdk.FailoverGroupSet{
+				ObjectTypes:             newObjectTypes,
+				AllowedIntegrationTypes: newAllowedIntegrationTypes,
+				ReplicationSchedule:     sdk.String(newReplicationSchedule),
```
sfc-gh-jcieslak, line 189: If that's not allowed, then maybe it should be moved to sdk.FailoverGroupSet validation (?)

sfc-gh-asawicki: Until we get the response about the syntax I would leave like this. If we get the confirmation, we should change the SDK implementation (so like you suggested validation, but also the generation objects, etc)

sfc-gh-jcieslak (approved):

PR #2565, opened 2024-02-27
docs: Add example tag association column

koonagi:
This PR adds an example of how to associate tags with columns. 

Previously, there was no example provided for this particular use case, which made it difficult for users to understand how to apply tags to columns. 
This addition aims to fill that gap and provide clear guidance on the process.

## Test Plan
* 

## References

*


sfc-gh-jcieslak (approved):
Hey 👋 Thanks for the contribution :)

sfc-gh-asawicki (approved):

PR #2575, opened 2024-02-29
fix: file_format validate() for field_optionally_enclosed_by

raulbonet:
ref #2385 

As pointed out by https://github.com/alexmaras , the `validate()` function does not properly validate the use-case where field_optionally_enclosed_by is None.


raulbonet:
@sfc-gh-asawicki 
I wonder if the `validation()` function should go as far as validating potential values that Snowflake will already validate when the query is executed. Not sure if it would be best just to remove all together this part of the validate. If tomorrow Snowflake adds other accepted values to this list, it would manually need to be modified.

pkg/sdk/file_format.go
```diff
@@ -572,7 +572,7 @@ func (opts *FileFormatTypeOptions) validate() error {
 		return fmt.Errorf("IgnoreUTF8Errors and ReplaceInvalidCharacters cannot be set simultaneously")
 	}
 
-	validEnclosedBy := []string{"null", "'", `"`}
+	validEnclosedBy := []string{"NONE", "'", `"`}
```
sfc-gh-asawicki, line 575: It would be nice to see this value used in one of the SDK integration tests (`file_format_integration_test.go`).

sfc-gh-asawicki:
> @sfc-gh-asawicki I wonder if the `validation()` function should go as far as validating potential values that Snowflake will already validate when the query is executed. Not sure if it would be best just to remove all together this part of the validate. If tomorrow Snowflake adds other accepted values to this list, it would manually need to be modified.

We currently have this discussion when approaching resources redesign. Our SDK-level validations or sometimes to strict and it opens us to the problem you mentioned. We will probably soften the validations on the general level but there is no decision yet.

sfc-gh-asawicki:
Thanks for the PR @raulbonet. I left one comment.

pkg/sdk/file_format.go
raulbonet: added integration test

sfc-gh-asawicki (approved):

sfc-gh-asawicki:
Thanks for the contribution @raulbonet! 🎉

PR #2581, opened 2024-03-01
chore: Add deprecated resources/datasources to the main documentation page

sfc-gh-asawicki:
Add a list of deprecated resources to the main documentation page:
- links are not ideal (they will break if we introduce subcategories)
- added additional checks and dependencies in make pre-push
- list has to be sorted in an alphabetic order (so that generation is deterministic)

Really hacky but we generate separate MD files that are included in our docs and can be included in our GH readme.


docs/index.md
```diff
@@ -226,3 +226,39 @@ The Snowflake provider will use the following order of precedence when determini
...
+- [snowflake_task_grant](./resources/task_grant) - use [snowflake_grant_privileges_to_account_role](./resources/grant_privileges_to_account_role) instead
+- [snowflake_unsafe_execute](./resources/unsafe_execute)
+- [snowflake_user_grant](./resources/user_grant) - use [snowflake_grant_privileges_to_account_role](./resources/grant_privileges_to_account_role) instead
+- [snowflake_view_grant](./resources/view_grant) - use [snowflake_grant_privileges_to_account_role](./resources/grant_privileges_to_account_role) instead
+- [snowflake_warehouse_grant](./resources/warehouse_grant) - use [snowflake_grant_privileges_to_account_role](./resources/grant_privileges_to_account_role) instead
+
+
+## Currently deprecated datasources
+
+- [snowflake_role](./resources/role) - use [snowflake_roles](./resources/roles) instead
```
sfc-gh-jcieslak, line 263: Is ./resources correct or it should be ./data-sources?

pkg/internal/tools/doc-gen-helper/main.go
```diff
@@ -0,0 +1,104 @@
+//aaago:build exclude
```
sfc-gh-jcieslak, line 1: shouldn't this be //go:build exclude 😅 ?

pkg/provider/doc_helpers.go
```diff
@@ -0,0 +1,24 @@
+package provider
```
sfc-gh-jcieslak, line 1: I think this should be in `internal/provider` package (as long as it will be still usable from the doc generator)

pkg/internal/tools/doc-gen-helper/main.go
sfc-gh-asawicki: should, the syntax is not highlighted correctly then, so for the time-being I changed it, forgot to remove it :D

docs/index.md
sfc-gh-asawicki: hmmm, it's probably wrong in this case, I will check

pkg/provider/doc_helpers.go
sfc-gh-asawicki: I'll try and we'll see

sfc-gh-asawicki: Done (moved to internal/provider/docs)

docs/index.md
sfc-gh-asawicki: Fixed

pkg/internal/tools/doc-gen-helper/main.go
sfc-gh-asawicki: Removed

sfc-gh-jcieslak (approved):

Issue #2582, opened 2024-03-02
Cannot delete snowflake_network_policy_attachment resource


brandocomando:
### Terraform CLI and Provider Versions

resource "snowflake_network_policy" "policy" {
  name    = "policy"
  comment = "A policy."

  allowed_ip_list = ["192.168.0.100/24"]
  blocked_ip_list = ["192.168.0.101"]
}

resource "snowflake_network_policy_attachment" "attach" {
  network_policy_name = "policy"
  set_for_account     = true
}

### Terraform Configuration

```terraform
Terraform v1.5.5
on darwin_amd64
+ provider registry.terraform.io/hashicorp/aws v5.37.0
+ provider registry.terraform.io/hashicorp/random v3.6.0
+ provider registry.terraform.io/hashicorp/tls v4.0.5
+ provider registry.terraform.io/snowflake-labs/snowflake v0.86.0

Your version of Terraform is out of date! The latest version
is 1.7.4. You can update by downloading from https://www.terraform.io/downloads.html
```

### Expected Behavior

tf destroy should remove the policy attachment

### Actual Behavior

│ Error: error deleting attachment for network policy policy err = error unsetting network policy policy on account err = 001003 (42000): SQL compilation error:
│ syntax error line 1 at position 20 unexpected ''NETWORK_POLICY''.
│ 

### Steps to Reproduce

1. `terraform apply`
2. `terraform destroy`

### How much impact is this issue causing?

Medium

### Additional Information

Seems like a simple sql error. Not sure what the provider is trying to do, but 
ALTER ACCOUNT UNSET NETWORK_POLICY seems to work for me. as accountadmin

and yes I am running the tf as accountadmin

raulbonet:
We are about to undergo a migration of network policies and I realised we have the same problem. The issue is that in file `sdk/parameters.go`, `ObjectParameterUnset` is tagged with `ddl:"keyword,single_quotes"` but the single_quotes should not be there I believe; I can take over and create a PR for it.

sfc-gh-asawicki:
Hey @brandocomando. Can you bump to the newest provider version and check if the problem persists after change #2584?

kvillar:
I had this issue as well. Bumping to the latest provider version resolved it. Thanks!

sfc-gh-asawicki:
Closing this issue then. @brandocomando please file a new issue if the problem persists in the newest provider version.

PR #2584, opened 2024-03-03
fix: unset network policy should not have single quotes

raulbonet:
ref #2582 

When unsetting, `NETWORK_POLICY` should not be enclosed between single quotes. This was not detected in the acceptance tests because these only covered user attachments; I added a test for an account attachment.


pkg/resources/network_policy_attachment_acceptance_test.go
```diff
@@ -49,6 +50,20 @@ func TestAcc_NetworkPolicyAttachment(t *testing.T) {
 			},
 		},
 	})
+	resource.ParallelTest(t, resource.TestCase{
```
sfc-gh-asawicki, line 53: I have two remarks here:
1. We move away from Parallel tests globally. They were introduced by our predecessors without the proper setup. Please change to Test.
2. Please extract this to separate `func TestAcc_...`.

pkg/resources/network_policy_attachment_acceptance_test.go
```diff
@@ -49,6 +50,20 @@ func TestAcc_NetworkPolicyAttachment(t *testing.T) {
...
+		Providers:    acc.TestAccProviders(),
```
sfc-gh-asawicki, line 54: We use now:
```
ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
```
and additionally 
```
		},
```
Please add them

pkg/resources/network_policy_attachment_acceptance_test.go
```diff
@@ -49,6 +50,20 @@ func TestAcc_NetworkPolicyAttachment(t *testing.T) {
...
+		PreCheck:     func() { acc.TestAccPreCheck(t) },
+		CheckDestroy: nil,
+		Steps: []resource.TestStep{
+			{
+				Config: networkPolicyAttachmentConfigAccount(policyNameAccount),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_network_policy_attachment.test", "network_policy_name", policyNameAccount),
+					resource.TestCheckResourceAttr("snowflake_network_policy_attachment.test", "set_for_account", "true"),
```
sfc-gh-asawicki, line 62: nit-picking (just a comment, not required change):
this test initially fails on destroy (before the fix) but this is implicit, the reader is not aware of the use case being tested. It would be more readable to either:
- add CheckDestroy explicitly requiring network policy attachment to be unset or...
- add a second test step with just a network policy and no snowflake_network_policy_attachment

That way our intent is clear.

sfc-gh-asawicki (changes requested):
Hey @raulbonet. Thanks for another contribution! I have just left cosmetic remarks, I hope that's okay. :)

sfc-gh-asawicki:
Hey @raulbonet. Did you have a chance to introduce the changes after the review? Maybe we can help you and hijack the PR? cc: @sfc-gh-jcieslak

raulbonet:
Hello @sfc-gh-asawicki 

Sorry for the delay, I got pulled into another project. 

I can do this tomorrow :)

pkg/resources/network_policy_attachment_acceptance_test.go
raulbonet: Done!

pkg/resources/network_policy_attachment_acceptance_test.go
raulbonet: Done!

pkg/resources/network_policy_attachment_acceptance_test.go
raulbonet: I created a CheckDestroy function but sorry I am not sure I understand what the problem was. Probably it's just that it was a while ago and I don't remember the context. 
Is it OK now?

sfc-gh-asawicki (dismissed):

raulbonet:
Hi @sfc-gh-asawicki 
I think the linter was complaining, committed again!

sfc-gh-asawicki (approved):

sfc-gh-asawicki:
@raulbonet merged, thank you for another contribution! :)

Issue #2366, opened 2024-01-17
GRANT IMPORTED PRIVILEGES TO ROLE introduced permadiff


jrobison-sb:
### Terraform CLI and Provider Versions

```
Terraform v1.6.6
on darwin_arm64
+ provider registry.terraform.io/snowflake-labs/snowflake v0.73.0
```

### Terraform Configuration

```terraform
resource "snowflake_database_grant" "staff_imported_priviliges_on_snowflake_database" {
  # `GRANT IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE TO ROLE STAFF_READ_WRITE`
  # This allows staff to query billing metrics in the snowflake database.
  # https://docs.snowflake.com/en/sql-reference/account-usage#enabling-the-snowflake-database-usage-for-other-roles
  database_name = "SNOWFLAKE"
  privilege     = "IMPORTED PRIVILEGES"
  roles = [
    snowflake_role.staff_read_write.name,
    snowflake_role.staff_read_only.name,
  ]
  with_grant_option = false
}
```
```


### Expected Behavior

I expect to be able to apply this with the usual terraform workflow:
1. `terraform apply` # This should apply cleanly, which it does
2. `terraform plan`  # This should report no diffs, because I already applied the grants

### Actual Behavior

Results in a permadiff:
1. `terraform apply` # This should apply cleanly, which it does
2. `terraform plan`  # This reports a permadiff

### Steps to Reproduce

Use the above HCL code, and then:

1. `terraform apply`
2. `terraform plan` # permadiff seen here


### How much impact is this issue causing?

Low

### Logs

_No response_

### Additional Information

_No response_

sfc-gh-jcieslak:
Hey @jrobison-sb 
The fix for that should be introduced soon. We are aware of the issue with IMPORTED PRIVILEGES. We have to handle it differently than other privileges as this privilege is a "pseudo-role" and it cannot be read from Snowflake.

sfc-gh-jcieslak:
Hey 👋, the fix for imported privileges has been merged in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2471 and will be available in the next release. As soon as the latest version appears (next week), please let us know if it fixes your bug and the issue can be closed.

chrisweis:
[...]

Issue #1998, opened 2023-08-07
Imported privileges on SNOWFLAKE database not registered in state


dlouseiro:
**Provider Version**

0.69.0

**Terraform Version**

1.5.4

**Describe the bug**

The `GRANT IMPORTED PRIVILEGES ON DATABASE SNOWFLAKE TO ROLE X`  is not properly registered in the state file, which causes every `terraform plan` to consider these privileges as non-existent privileges every time.

**Expected behavior**

For these privileges to be registered in the state file.

**Code samples and commands**

```
resource "snowflake_grant_privileges_to_role" "snowflake_imported_privileges_usage_monitoring" {
  privileges = ["IMPORTED PRIVILEGES"]
  role_name  = snowflake_role.usage_monitoring.name
  on_account_object {
    object_type = "DATABASE"
    object_name = "SNOWFLAKE"
  }
}
```

dstuck:
I'm seeing this in provider version 0.68.0 as well

chrisweis:
So happy to see this issue raised again, it's been driving me nuts.  I've needed to manually comment/uncomment this line in my snowflake_database_grant resource to work around it:

```
  lifecycle {
    ignore_changes = all
  }
```

Related: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1573

attrivivekFF:
We had been facing the same issue, wherein snowflake.account_usage grant would show up on each terraform plan. We would ignore it and apply the changes and nothing would change. But from past few days apply has been failing for this resource with insufficient privileges error. Can you please help to check this?

wh-k-klemens:
We are facing the same issue with a `USAGE` grant on version `0.70.0` of the provider. It is not properly recoded in the state. For example applying following resource: 
```
resource "snowflake_grant_privileges_to_role" "<some_grant_name>" {
  for_each   = var.schemas
  privileges = ["USAGE"]
  role_name  = snowflake_role.<some-role>.name
  on_schema {
    schema_name = "${snowflake_database.<some-db>.name}.${snowflake_schema.<some-schema>[each.key].name}"
  }
}
```
It will still yield the following changes when using plan:
```
# snowflake_grant_privileges_to_role.<some_grant_name>["<some-key>"] will be updated in-place
~ resource "snowflake_grant_privileges_to_role" "<some_grant_name>" {
      id                = "<generated-id>"
    ~ privileges        = [
        + "USAGE",
      ]
[...]

PR #2596, opened 2024-03-06
fix: support snowflake application as database

sfc-gh-jcieslak:
Fix for: #1998 and #2366

Changes
- Because it's the default database it fulfills our `if` checks for futures, so added a check for `name != SNOWFLAKE`
- The second check was added to make it possible to grant privileges on applications by setting object_type to DATABASE
- Those cases were added to the documentation for the `snowflake_grant_privilege_to_account_role` resource
- Added acceptance test that would check if SNOWFLAKE database could be made with `object_type = DATABASE`
- I also added a follow-up ticket to add additional tests for applications when they are available.

### DIFF
--- a/docs/resources/task_grant.md
+++ b/docs/resources/task_grant.md
@@ -53,6 +53,6 @@ resource "snowflake_task_grant" "grant" {
 Import is supported using the following syntax:
 
 ```shell
-# format is database_name|schema_name|task_name|privilege|with_grant_option|on_future|on_all|roles"
-terraform import snowflake_task_grant.example "MY_DATABASE|MY_SCHEMA|MY_TASK|OPERATE|false|false|false|role1,role2"
+# format is database_name|schema_name|task_name|privilege|with_grant_option|on_future|roles"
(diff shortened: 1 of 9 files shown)


docs/resources/task_grant.md
```diff
@@ -53,6 +53,6 @@ resource "snowflake_task_grant" "grant" {
 Import is supported using the following syntax:
 
 ```shell
-# format is database_name|schema_name|task_name|privilege|with_grant_option|on_future|on_all|roles"
-terraform import snowflake_task_grant.example "MY_DATABASE|MY_SCHEMA|MY_TASK|OPERATE|false|false|false|role1,role2"
+# format is database_name|schema_name|task_name|privilege|with_grant_option|on_future|roles"
```
sfc-gh-asawicki, line 56: are you removing the changes from #2592 ?

sfc-gh-asawicki (changes requested):

docs/resources/task_grant.md
sfc-gh-jcieslak: Didn't notice it was previously added only to .md file so It was re-generated to the wrong format. Fixed import.sh file, so it generated as it should be.

sfc-gh-asawicki (approved):

PR #2603, opened 2024-03-07
chore: Fix env variables for tests

sfc-gh-asawicki:
- Introduce acceptance test context
- Expose default config in integration tests
- Remove not needed test envs
- Expose envs to test jobs (because of [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsenv) and [that](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow))
- Skip the tests requiring business critical account

### DIFF
--- a/pkg/provider/provider_acceptance_test.go
+++ b/pkg/provider/provider_acceptance_test.go
@@ -17,6 +17,8 @@ import (
 )
 
 func TestAcc_Provider_configHierarchy(t *testing.T) {
+	t.Setenv(string(testenvs.ConfigureClientOnce), "")
+
 	user := acc.DefaultConfig(t).User
 	pass := acc.DefaultConfig(t).Password
 	account := acc.DefaultConfig(t).Account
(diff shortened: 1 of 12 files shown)


pkg/provider/provider_acceptance_test.go
```diff
@@ -119,6 +121,37 @@ func TestAcc_Provider_configHierarchy(t *testing.T) {
 	})
 }
 
+func TestAcc_Provider_(t *testing.T) {
```
sfc-gh-jcieslak, line 124: what does this test prove? I think there's a missing test title (function name) and I don't get what it's testing. Could you briefly describe?

sfc-gh-asawicki: I forgot to add the title 🤦 

It tests what we discussed on Slack, so the situation when setting the client once is ignored for one test and potentially it can leave the client for the next tests in a bad shape.

sfc-gh-asawicki: Added helping comments

sfc-gh-jcieslak (approved):

PR #2604, opened 2024-03-07
feat: Add snowflake grant ownership resource

sfc-gh-jcieslak:
TODO

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+			},
+		},
+	}
+}
+
+func GrantOwnership() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: CreateGrantOwnership,
+		DeleteContext: DeleteGrantOwnership,
+		ReadContext:   ReadGrantOwnership,
```
sfc-gh-asawicki, line 212: let's leave a comment that update is not needed because everything is forceNew. Also, should we have (is there already in the SDK) a no-op function we can just reference in such cases?

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+	},
+}
+
+func grantOwnershipBulkOperationSchema(branchName string) map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"object_type_plural": {
+			Type:         schema.TypeString,
+			Required:     true,
+			ForceNew:     true,
+			Description:  "Specifies the type of object in plural form on which you are transferring ownership. Available values are: AGGREGATION POLICIES | ALERTS | AUTHENTICATION POLICIES | COMPUTE POOLS | DATABASES | DATABASE ROLES | DYNAMIC TABLES | EVENT TABLES | EXTERNAL TABLES | EXTERNAL VOLUMES | FAILOVER GROUPS | FILE FORMATS | FUNCTIONS | HYBRID TABLES | ICEBERG TABLES | IMAGE REPOSITORIES | INTEGRATIONS | MATERIALIZED VIEWS | NETWORK POLICIES | NETWORK RULES | PACKAGES POLICIES | PIPES | PROCEDURES | MASKING POLICIES | PASSWORD POLICIES | PROJECTION POLICIES | REPLICATION GROUPS | ROLES | ROW ACCESS POLICIES | SCHEMAS | SESSION POLICIES | SECRETS | SEQUENCES | STAGES | STREAMS | TABLES | TAGS | TASKS | USERS | VIEWS | WAREHOUSES",
```
sfc-gh-asawicki, line 180: same as above

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+			}
+		}
+	}
+
+	d.SetId("")
+
+	return nil
+}
+
+func ReadGrantOwnership(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
```
sfc-gh-asawicki, line 379: so are there cases where we should setId to `""` in read like in the normal resources?

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+	}
+
+	return ownershipGrantTo
+}
+
+func getOwnershipGrantOpts(id *GrantOwnershipId) *sdk.GrantOwnershipOptions {
+	if id != nil && id.OutboundPrivilegesBehavior != nil {
+		outboundPrivileges := id.OutboundPrivilegesBehavior.ToOwnershipCurrentGrantsOutboundPrivileges()
+		if outboundPrivileges != nil {
+			return &sdk.GrantOwnershipOptions{
```
sfc-gh-asawicki, line 564: it would be easier to have just one return in this method, so:
1. create options at the start
2. set outbound privileges inside if
3. return at the end

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+				Summary:  "Couldn't find OWNERSHIP privilege on target object",
+				Detail:   fmt.Sprintf("Id: %s", d.Id()),
+			},
+		}
+	}
+
+	return nil
+}
+
+func getOnObjectIdentifier(objectType sdk.ObjectType, objectName string) (sdk.ObjectIdentifier, error) {
```
sfc-gh-asawicki, line 453: this looks lika a general use validation method that could be inside the SDK (so each object has its dedicated type of identifier). Please leave a TODO with identifiers rework issue number and add reference to this method inside the issue's description, so that we don't forget about it.

pkg/resources/grant_ownership_test.go
```diff
@@ -0,0 +1,518 @@
...
+		{
+			Name:       "validation - unsupported type",
+			ObjectType: sdk.ObjectTypeShare,
+			ObjectName: "some_share",
+			Error:      "object_type SHARE is not supported",
+		},
+		{
+			Name:       "validation - invalid account object identifier",
+			ObjectType: sdk.ObjectTypeDatabase,
+			ObjectName: "test_database.test_schema",
```
sfc-gh-asawicki, line 70: We recently had the discussion about the validations for account object identifiers that concluded that it should be less strict for account identifiers, right? What were the arrangements there? Is it a topic that we will revisit with identifiers rework or...?

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+
+	if grantOn.Future != nil {
+		// TODO (SNOW-1182623): Still waiting for the response on the behavior/SQL syntax we should use here
+	} else {
+		accountRoleName, err := client.ContextFunctions.CurrentRole(ctx)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		err = client.Grants.GrantOwnership( // TODO: Should we always set outbound privileges to COPY in delete operation or set it to the config value?
```
sfc-gh-asawicki, line 354: this is similar question like the one we had in copyGrants in other resources, right?

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+			ctx,
+			grantOn,
+			sdk.OwnershipGrantTo{
+				AccountRoleName: sdk.Pointer(sdk.NewAccountObjectIdentifier(accountRoleName)), // TODO: What if current role is database role (not a common but possible case)
```
sfc-gh-asawicki, line 358: is it possible really? the use role is listed under account-level role management: https://docs.snowflake.com/en/sql-reference/sql/use-role. Are there any docs for use database role/ use role <database role>? If not, I think it deserves an internal question first.

pkg/resources/grant_ownership.go
sfc-gh-asawicki, line 354: how we know that this operation will succeed? What are the requirements for the tf role used for this to succeed? Can we explicitly state it at least in the docs (an ideally later maybe even validate this first)?

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+				ownershipFound = true
+			}
+		case ToDatabaseGrantOwnershipTargetRoleKind:
+			if grant.GranteeName.Name() == id.DatabaseRoleName.Name() {
+				ownershipFound = true
+			}
+		}
+	}
+
+	if !ownershipFound {
```
sfc-gh-asawicki, line 440: so should it be an error? in other (normal) resources, if we don't find the object we mark it as removed based on the error. So why is the no ownership an error. In the following example:
1. config with ownership
2. manual ownership transfer
3. terraform plan

the result would be an error. Should it be an error or should we try to set the ownership with tf?

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+	onObjectName := on["object_name"].(string)
+	onAll := on["all"].([]any)
+	onFuture := on["future"].([]any)
+
+	switch {
+	case len(onObjectType) > 0 && len(onObjectName) > 0:
+		objectType := sdk.ObjectType(strings.ToUpper(onObjectType))
+		objectName, err := getOnObjectIdentifier(objectType, onObjectName)
+		if err != nil {
+			return ownershipGrantOn, err
```
sfc-gh-asawicki, line 531: not `nil, err` to be more explicit? (even though underneath there is a nil)

pkg/resources/grant_ownership_test.go
```diff
@@ -0,0 +1,518 @@
...
+			config := make(map[string]any)
+			if tt.AccountRole != nil {
+				config["account_role_name"] = *tt.AccountRole
+			}
+			if tt.DatabaseRole != nil {
+				config["database_role_name"] = *tt.DatabaseRole
+			}
+			d := schema.TestResourceDataRaw(t, grantOwnershipSchema, config)
+
+			defer func() {
```
sfc-gh-asawicki, line 456: parametrized tests are nice but if the setup becomes like this, I think it's more readable (and less error-prone) to setup to set of tests: one for valid and second for invalid configs. In this case for invalid ones we could just use assert/require.Panics(). Also, there is clear separation between db and account roles, so I think that splitting it into 3 different sets would be best, the ost readable option

pkg/resources/grant_ownership_acceptance_test.go
```diff
@@ -0,0 +1,491 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
+	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
```
sfc-gh-asawicki, line 20: please add a few negative tests with bad configs.

Also, the test if something happens outside of terraform (but I think such a test will be added in the followup PR with more sophisticated edge cases).

sfc-gh-asawicki (changes requested):

pkg/resources/grant_ownership.go
```diff
@@ -0,0 +1,668 @@
...
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/logging"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+var validGrantOwnershipObjectTypes = []sdk.ObjectType{
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: One question, after moving it to the `sdk`, should it be used as it is right now (resource-level validation), used to validate it on `sdk` level, or both `sdk` and resource level validation?

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: I will try to set it to the NoOp function, but some time ago I had with difficulties with it, so nil != NoOp. Regardless, I'll add a note as to why there's no update.

pkg/resources/grant_ownership.go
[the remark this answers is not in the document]
sfc-gh-asawicki: I think for now it's fine to have it on resource level only (we will think about duplicating it while extracting the SDK).

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: Kind of, yeah. When someone defines this resource he/she will be thinking about `on create` effects setting COPY or REVOKE accordingly, but the question is: how do we behave when the resource is being removed? Should we transfer the ownership back with the same option as it was transferred in the Create or similarly in the case of the `view` resource it should be always internally set to COPY `on delete`?

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: Hmm, I thought it's possible, but apparently it's not. I'll remove the comment.

templates/resources/grant_ownership.md.tmpl
```diff
@@ -0,0 +1,92 @@
...
+where:
+- role_type - string - type of granted role (either ToAccountRole or ToDatabaseRole)
+- role_name - string - fully qualified identifier for either account role or database role (depending on the role_type)
+- outbound_privileges_behavior - string - behavior specified for existing roles (can be either COPY or REVOKE)
+- grant_type - enum
+- grant_data - data dependent on grant_type
+
+It has varying number of parts, depending on grant_type. All the possible types are:
+
+### OnObject
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Moved imports to imports.sh

pkg/sdk/object_types.go
```diff
@@ -15,54 +15,59 @@ type Object struct {
...
+	ObjectTypeApplicationPackage   ObjectType = "APPLICATION PACKAGE"
+	ObjectTypeApplicationRole      ObjectType = "APPLICATION ROLE"
+	ObjectTypeStreamlit            ObjectType = "STREAMLIT"
+	ObjectTypeColumn               ObjectType = "COLUMN"
+	ObjectTypeIcebergTable         ObjectType = "ICEBERG TABLE"
+	ObjectTypeExternalVolume       ObjectType = "EXTERNAL VOLUME"
+	ObjectTypeNetworkRule          ObjectType = "NETWORK RULE"
+	ObjectTypePackagesPolicy       ObjectType = "PACKAGES POLICY"
+	ObjectTypeComputePool          ObjectType = "COMPUTE POOL"
+	ObjectTypeAggregationPolicy    ObjectType = "AGGREGATION POLICY"
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: SNOW-1230423 created to address this.

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: Added link at the end.

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: As discussed, every time where we cannot find the ownership on the target object

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: Do you mean what would happen in cases where, e.g. the terraform role is not set to accountadmin?

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: As discussed, we should set id to an empty string, and in the create operation we should try to forcefully take ownership and then grant it (or just grant it straight to the target role). Maybe to make it simpler I'll add it to the list of things I'll do in the next pr (should be added to the pr desc).

pkg/resources/grant_ownership_test.go
sfc-gh-jcieslak: Divided based on valid / invalid and role type

pkg/resources/grant_ownership_acceptance_test.go
sfc-gh-jcieslak: Yeah, added to the list in the pr description to overall test outside of terraform interactions.

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: Noop cannot be set, because all of the fields are marked as ForceNew. I just left the comment why there's no Update.

pkg/resources/grant_ownership_test.go
sfc-gh-jcieslak: Yeah, that's my bad. This should be possible, so I'll change it and I'll add TODO to address this when we'll be working on identifiers.

pkg/resources/grant_ownership_test.go
```diff
@@ -429,47 +455,33 @@ func TestGetOwnershipGrantTo(t *testing.T) {
...
 		})
 	}
 }
 
+func TestInvalidDatabaseRoleGetOwnershipGrantTo(t *testing.T) {
+	d := schema.TestResourceDataRaw(t, grantOwnershipSchema, map[string]any{
+		"database_role_name": "account_role_name",
+	})
+
+	defer func() {
```
sfc-gh-asawicki, line 477: could be done with assert.Panics as suggested, you can change in the follow-up PR.

sfc-gh-asawicki (dismissed):

pkg/resources/grant_ownership_test.go
sfc-gh-jcieslak: Changed

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

Issue #2459, opened 2024-02-02
grant_privileges_to_role doesn't recognize changes to `with_grant_option` that were made outside of Terraform


joey-squid:
### Terraform CLI and Provider Versions

Terraform v1.6.2
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v5.33.0
+ provider registry.terraform.io/snowflake-labs/snowflake v0.84.0

### Terraform Configuration

```terraform
resource "snowflake_grant_privileges_to_role" "grant_metrics_table_usage" {
  privileges = ["SELECT", "INSERT", "UPDATE", "DELETE"]
  role_name  = snowflake_role.metrics_role.name
  on_schema_object {
    object_name = "\"${snowflake_database.metrics_db.name}\".\"PUBLIC\".\"METRICS\""
    object_type = "TABLE"
  }
  with_grant_option = true
}
```

### Expected Behavior

Those permissions, which currently have their `grant_option` set to `false` (as reported in `show grants to role`), should change to have their grants set to `true`.

### Actual Behavior

`terraform plan` shows
```
      ~ privileges        = [
          + "DELETE",
          + "INSERT",
          + "SELECT",
          + "UPDATE",
        ]
```
and `terraform apply` doesn't actually change anything.

### Steps to Reproduce

`terraform apply` or `terraform plan`

### How much impact is this issue causing?

Low

### Additional Information

I took a gander at the code to see if there was anything I was doing obviously wrong, and it seems that the `UpdateGrantPrivilegesToRole` function never calls `d.Get("with_grant_option").(bool)` (the `CreateGrantPrivilegesToRole` function does), so I'm fairly certain this is just a bug in the provider. I don't have the bandwidth to fix it but I do have the bandwidth to test a fix if someone else gets there.

sfc-gh-jcieslak:
Hey @joey-squid 👋 
[...]

PR #2608, opened 2024-03-08
fix: with_grant_option diff drift

sfc-gh-jcieslak:
Fixes: #2459

The issue was related to the fact that the Read operation is only concerned about privileges (with_grant_option is also taken into consideration, but it's never set). Given this configuration:
```terraform
resource "snowflake_grant_privileges_to_account_role" "test" {
  account_role_name = "TEST_ROLE"
  privileges = ["TRUNCATE"]
  on_schema_object {
    object_type = "TABLE"
    object_name = "TEST_DATABASE.TEST_SCHEMA.TEST_TABLE"
  }
  with_grant_option = true
}
```
after apply we run the following commands by hand
```sql
revoke truncate on table test_table from role test_role;
grant truncate on table test_table to role test_role; -- notice we don't add "with grant option" which our resource should detect
```
Now, when we run `plan` or `apply` our resource is seeing a drift (the "TRUNCATE" privilege is not present, because `with_grant_option` is not matching) and tries to run the Update operation (unsuccessfully; 1. because of the "sdk.GrantPrivOptions" not set 2. because of the incorrect logic). When there're already existing grants there are two ways to update `with_grant_option` which depends on what is set in Snowflake. It's better to show it with SQLs, so:

```sql
-- imagine this is ran by Snowflake Terraform Plugin (with_grant_option is set to true in the config)
grant truncate on table test_table to role test_role with grant option;

-- this is ran by hand in the worksheet
revoke truncate on table test_table from role test_role;
grant truncate on table test_table to role test_role; 

-- now update tries to run the following
grant truncate on table test_table to role test_role with grant option; -- this will successfully update with_grant_option to 'true'
```

```sql
-- imagine this is ran by Snowflake Terraform Plugin (with_grant_option is set to false in the config)
grant truncate on table test_table to role test_role;

-- this is ran by hand in the worksheet
revoke truncate on table test_table from role test_role;
grant truncate on table test_table to role test_role with grant option; 

-- now update tries to run the following
grant truncate on table test_table to role test_role; -- this won't update the with_grant_option to false because Snowflake is not updating the value when the option is already set to true (you have to revoke it first)
```

The fix I opted to is to:
- when with_grant_option is set to true in the config
  - proceed as it was (but now set option struct correctly with with_grant_option set to true)
- when with_grant_option is set to false in the config
  - firstly revoke privileges we would like to add (just in case this issue happens; it won't fail even if the grant does not exist)
  - then proceed as it was (grant privileges we would like to add) 

todo other grant privileges to database role

## Test Plan

* [x] Acceptance tests that prove the issue has been fixed for every privilege-granting resource

### DIFF
--- a/pkg/resources/grant_privileges_to_account_role.go
+++ b/pkg/resources/grant_privileges_to_account_role.go
@@ -510,7 +510,24 @@ func UpdateGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceD
 
 			if len(privilegesToAdd) > 0 {
 				logging.DebugLogger.Printf("[DEBUG] Granting privileges: %v", privilegesToAdd)
-				if !id.WithGrantOption {
+				if id.WithGrantOption {
+					err = client.Grants.GrantPrivilegesToAccountRole(
+						ctx,
+						getAccountRolePrivileges(
+							false,
+							privilegesToAdd,
+							id.Kind == OnAccountAccountRoleGrantKind,
+							id.Kind == OnAccountObjectAccountRoleGrantKind,
+							id.Kind == OnSchemaAccountRoleGrantKind,
+							id.Kind == OnSchemaObjectAccountRoleGrantKind,
+						),
+						grantOn,
+						id.RoleName,
+						&sdk.GrantPrivilegesToAccountRoleOptions{
+							WithGrantOption: sdk.Bool(true),
+						},
+					)
+				} else {
 					err = client.Grants.RevokePrivilegesFromAccountRole(
 						ctx,
 						getAccountRolePrivileges(
@@ -548,23 +565,6 @@ func UpdateGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceD
 						grantOn,
 						id.RoleName,
 						new(sdk.GrantPrivilegesToAccountRoleOptions),
-					)
-				} else {
-					err = client.Grants.GrantPrivilegesToAccountRole(
-						ctx,
-						getAccountRolePrivileges(
-							false,
-							privilegesToAdd,
-							id.Kind == OnAccountAccountRoleGrantKind,
-							id.Kind == OnAccountObjectAccountRoleGrantKind,
-							id.Kind == OnSchemaAccountRoleGrantKind,
-							id.Kind == OnSchemaObjectAccountRoleGrantKind,
-						),
-						grantOn,
-						id.RoleName,
-						&sdk.GrantPrivilegesToAccountRoleOptions{
-							WithGrantOption: sdk.Bool(true),
-						},
 					)
 				}
 				if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -510,20 +514,32 @@ func UpdateGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceD
...
+				privilegesToGrant := getAccountRolePrivileges(
+					false,
+					privilegesToAdd,
+					id.Kind == OnAccountAccountRoleGrantKind,
+					id.Kind == OnAccountObjectAccountRoleGrantKind,
+					id.Kind == OnSchemaAccountRoleGrantKind,
+					id.Kind == OnSchemaObjectAccountRoleGrantKind,
 				)
+
+				if id.WithGrantOption {
```
sfc-gh-asawicki, line 526: nit-pick:
we could do it a bit less repetitive by:
```
if !id.WithGrantOption {
// revoke here
}
// just one grant privileges parametrized with `id.WithGrantOption`
err = client.Grants.GrantPrivilegesToAccountRole(ctx, privilegesToGrant, grantOn, id.RoleName, &sdk.GrantPrivilegesToAccountRoleOptions{WithGrantOption: sdk.Bool(id.WithGrantOption)}))
```
less repetitive and no nesting

sfc-gh-asawicki (dismissed):

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jcieslak: True, changing

pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -512,20 +516,28 @@ func UpdateGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceD
...
+							diag.Diagnostic{
+								Severity: diag.Error,
+								Summary:  "Failed to revoke privileges to add",
+								Detail:   fmt.Sprintf("Id: %s\nPrivileges to add: %v\nError: %s", d.Id(), privilegesToAdd, err.Error()),
+							},
+						}
+					}
+				}
+
+				err = client.Grants.GrantPrivilegesToAccountRole(ctx, privilegesToGrant, grantOn, id.RoleName, new(sdk.GrantPrivilegesToAccountRoleOptions))
```
sfc-gh-asawicki, line 540: shouldn't there be `&sdk.GrantPrivilegesToAccountRoleOptions{WithGrantOption: sdk.Bool(id.WithGrantOption)}` instead of `new(sdk.GrantPrivilegesToAccountRoleOptions)`?

sfc-gh-jcieslak: Tru, changed.

sfc-gh-asawicki (approved):

Issue #2549, opened 2024-02-23
snowflake_grant_privileges_to_account_role to support OWNERSHIP role


AndrewKlimovski:
### Terraform CLI and Provider Versions

v1.5.4

### Use Cases or Problem Statement

Usage of `snowflake_<resource>_grant` is being deprecated in favour of `snowflake_grant_privileges_to_account_role` as per the deprecation warning.

```
Warning: Deprecated Resource
│ 
│   with snowflake_database_grant.grant_db_dev_data_lake,
│   on database.tf line 15, in resource "snowflake_database_grant" "grant_db_dev_data_lake":
│   15: resource "snowflake_database_grant" "grant_db_dev_data_lake" {
│ 
│ This resource is deprecated and will be removed in a future major version release. Please use snowflake_grant_privileges_to_account_role instead.
```

Example `snowflake_database_grant` accepts the privilege `OWNERSHIP` however trying to move to the proposed new function throws the error:

```
Error: Unsupported privilege 'OWNERSHIP'
|
|
|
Granting ownership is only allowed in dedicated resources (snowflake_user_ownership_grant, snowflake_role_ownership_grant)
```

### Proposal

Either the function `snowflake_grant_privileges_to_account_role` supports the 'OWNERSHIP' role or examples are give of how to port from the deprecated `snowflake_<resource>_grant` function to the new function

### How much impact is this issue causing?

Medium

lachniej:
The same thing is happening with the snowflake_grant_priviliges_to_database_role on attempted OWNERSHIP grants, could we add similar modification of that role as well?

AndrewKlimovski:
@sfc-gh-jcieslak to be clear there are a number of functions that being deprecated and all have the same behaviour, I've just listed one of them in the example and @lachniej has called out another. Would be amazing to support all deprecated functions

sfc-gh-jcieslak:
Hey @AndrewKlimovski @lachniej 👋 
We wanted to create another resource that will be specialized in granting ownership as already mentioned `snowflake_grant_priviliges_to_database_role` and `snowflake_grant_priviliges_to_account_role` are already complex resources. Additionally granting ownership comes with many edge cases that would add to the complexity and maintainability of those resources. That said, we'll add a new resource that will be used only for ownership transfer operations. I should start the work on the implementation this week.

AndrewKlimovski:
Thanks for the update @sfc-gh-jcieslak 

[...]

Issue #2199, opened 2023-11-17
snowflake resources import - snowflake_sequence_grant


pogorzem92:
**Provider Version**

0.67

**Terraform Version**

0.13.7

**Describe the bug**
I got error during importing resources to terraform:

Error: unexpected format of ID ("DATABASE|SCHEMA||OWNERSHIP|false|true|role"), expected database_name|schema_name|sequence_name|privilege|with_grant_option|on_future|roles

**Expected behavior**

resource should be imported

**Code samples and commands**

terraform import --var-file=dev.tfvars module.snowflake_bi_stack.module.schemas.snowflake_sequence_grant.o_sequence[\"NAME\"] "DATABASE|SCHEMA||OWNERSHIP|false|true|role"

**Additional context**

import didn't work correctly and error message suggest that code it fine
I tried add one field in ID and it worked - it looks like code needed additional 'on_all' field in ID - added fight before roles name - documentation and ERROR on console doesn't mention it

terraform import --var-file=dev.tfvars module.snowflake_bi_stack.module.schemas.snowflake_sequence_grant.o_sequence[\"NAME\"] "DATABASE|SCHEMA||OWNERSHIP||false|true|role"

I got similar errors with other resources: snowflake_task_grant and snowflake_stream_grant - probably issue is the same

sfc-gh-jcieslak:
Hey 👋
Closing, as the issue was about the deprecated/not allowed functionality. Recently, we released [a new grant resource](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_ownership) which is capable of granting ownership. Please, give it a try. If there will be any issues with it, create another GitHub issue. Also, please check our [technical documentation section](https://github.com/Snowflake-Labs/terraform-provider-snowflake/tree/main/docs/technical-documentation) where you can find a migration guide that can help you with upgrading to the latest grant resources and our newly added design decision doc (regarding new grant resources).

Issue #2084, opened 2023-09-29
OWNERSHIP can only be transferred error on destruction of snowflake_grant_privileges_to_role resource


liamjamesfoley:
**Provider Version**
0.71

**Terraform Version**
1.5.2

**Describe the bug**
https://docs.snowflake.com/en/sql-reference/sql/grant-ownership
Destruction of `snowflake_grant_privileges_to_role` produces that a query that will always fail.

<img width="692" alt="image" src="https://github.com/Snowflake-Labs/terraform-provider-snowflake/assets/4086743/2098de34-b902-4dfb-a095-7db85bdcfd2c">

**Expected behavior**
This is happening in a managed access schema module I'm working on, so TBH, I don't care as long as the resource destruction happens without using `terraform state rm` b/c I'm also destroying the schema, but ownership should probably go to `ACCOUNTADMIN`?

Since `snowflake_schema_grant` is being deprecated the new resource should have something similar to [revert_ownership_to_role_name](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/schema_grant#revert_ownership_to_role_name).
https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/schema_grant.go#L106-L114

**Code samples and commands**

Resource:
```
resource "snowflake_grant_privileges_to_role" "owner_schema_grant" {
  # `GRANT OWNERSHIP ON SCHEMA ${schema_identifier} TO ROLE ${owner_role};`,
  role_name  = snowflake_role.owner_role.name
  privileges = ["OWNERSHIP"]
  on_schema {
    schema_name = "${var.db_name}.${snowflake_schema.schema.name}"
  }
  with_grant_option = true
}
```

imre-kerr-sb1:
We also have issues with ownership grants. For now we are using the following workaround:

```terraform
data "snowflake_current_role" "current" {}

resource "snowsql_exec" "owner_schema_grant" {
  name = "owner_schema_grant"

  create {
    statements           = "GRANT OWNERSHIP ON SCHEMA ${schema_identifier} TO ROLE ${snowflake_role.owner_role.name};"
    number_of_statements = 1
  }

  read {
    statements           = "SHOW GRANTS ON SCHEMA ${schema_identifier};"
    number_of_statements = 1
  }

  delete {
[...]

Issue #1942, opened 2023-07-10
Changing snowflake_grant_privileges_to_role resource involving ownership fails due to dependent grants


dmfay:
Provider 0.68.1

Here's my `snowflake_grant_privileges_to_role` resource:

```tf
resource "snowflake_grant_privileges_to_role" "admin_schema" {
  for_each          = toset(["PUBLIC", snowflake_schema.raw.name, snowflake_schema.security.name])
  role_name         = local.role_admin
  privileges        = ["OWNERSHIP", "CREATE TABLE", "USAGE"]
  with_grant_option = false

  on_schema {
    # has to be qualified
    schema_name = "${snowflakedb.db.name}.${each.value}"
  }
}
```

This applied successfully once, but then I needed to modify the schema list in the `for_each` (specifically, I hardcoded `PUBLIC` as you see here [instead of defining it as a resource](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/306)). The plan shows all additions even though each grant already exists and is already managed by terraform. For example, here's one of the three schemata referenced in the resource declaration:

```
  # module.snowflakedb.snowflake_grant_privileges_to_role.admin_schema["PUBLIC"] will be created
+ resource "snowflake_grant_privileges_to_role" "admin_schema" {
+       all_privileges    = false
+       id                = (known after apply)
+       on_account        = false
+       privileges        = [
+           "CREATE TABLE",
+           "OWNERSHIP",
+           "USAGE",
        ]
+       role_name         = "ADMIN_PRODUCTION"
+       with_grant_option = false

+       on_schema {
+           schema_name = "PROD.PUBLIC"
        }
    }
```

Application then fails due to dependent grant errors even though ownership has not actually been changed:

```
╷
│ Error: error granting privileges to account role: 003036 (23001): SQL execution error: Dependent grant of privilege 'CREATE TABLE' on securable 'PROD.RAW' to role 'ADMIN_PRODUCTION' exists.  It must be revoked first.  More than one dependent grant may exist: use 'SHOW GRANTS' command to view them.  To revoke all dependent grants while transferring object ownership, use convenience command 'GRANT OWNERSHIP ON <target_objects> TO <target_role> REVOKE CURRENT GRANTS'.
│ 
│   with module.snowflakedb.snowflake_grant_privileges_to_role.admin_schema["PUBLIC"],
[...]

PR #2628, opened 2024-03-18
feat: Grant ownership follow up

sfc-gh-jcieslak:
A follow-up for #2604. 

Done in this pr:
- Add setId("") in Read (when ownership is not found on the target object) and forcefully grant ownership in Create (this was already present, but added test cases for it).
- Edge cases
  - Granting `ON PIPE` and `ON ALL PIPES` is handled (pipes are paused before and resumed after ownership transfer)

Full list of things that still need to be done:
- Deprecation messages
- More documentation (explain how grant_ownership resource handles edge cases) and examples that would show simple usage, edge cases, cases where the resource may cause trouble
- Referring to https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2604#discussion_r1519628668, test different cases where the Delete operation may struggle with
- Test outside of Terraform interactions to see how it behaves in different situations
  - A test where used role is not privileged enough to transfer ownership
  - Also cases within Terraform to see how grant_ownership will act with other grant resources within certain configurations
- Edge cases
  - Granting `ON TASK`
  - Use `VIEW` when granting on `MATERIALIZED VIEW`
  - Granting `ON EXTERNAL TABLES`

## References
[GRANT OWNERSHIP](https://docs.snowflake.com/en/sql-reference/sql/grant-ownership)

## Mentioned in
A list of issues requesting this resource: #2549 #2199 #2084 #1942 #1875


pkg/resources/grant_ownership_acceptance_test.go
```diff
@@ -577,6 +579,281 @@ func TestAcc_GrantOwnership_InvalidConfiguration_MultipleTargets(t *testing.T) {
...
+				ConfigVariables:   configVariables,
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAcc_GrantOwnership_ForceOwnershipTransferOnCreate(t *testing.T) {
```
sfc-gh-asawicki, line 639: I don't understand what do we check in this test.
1. We create a role.
2. We create a Database and assign the role to it.
3. That was all preconfig. Then we have tf run which does not have to do anything as far as i understand this.

So what is checked exactly? Or what am I missing?

pkg/resources/grant_ownership_acceptance_test.go
```diff
@@ -577,6 +579,281 @@ func TestAcc_GrantOwnership_InvalidConfiguration_MultipleTargets(t *testing.T) {
...
+					resource.TestCheckResourceAttr(resourceName, "on.0.object_name", pipeFullyQualifiedName),
+					resource.TestCheckResourceAttr(resourceName, "id", fmt.Sprintf("ToAccountRole|%s||OnObject|PIPE|%s", accountRoleFullyQualifiedName, pipeFullyQualifiedName)),
+					checkResourceOwnershipIsGranted(&sdk.ShowGrantOptions{
+						On: &sdk.ShowGrantsOn{
+							Object: &sdk.Object{
+								ObjectType: sdk.ObjectTypePipe,
+								Name:       sdk.NewSchemaObjectIdentifierFromFullyQualifiedName(pipeFullyQualifiedName),
+							},
+						},
+						// TODO(): Fix this identifier
```
sfc-gh-asawicki, line 724: add issue number?

pkg/resources/grant_ownership_acceptance_test.go
```diff
@@ -577,6 +579,281 @@ func TestAcc_GrantOwnership_InvalidConfiguration_MultipleTargets(t *testing.T) {
...
+					}, sdk.ObjectTypePipe, accountRoleName, fmt.Sprintf("%s\".\"%s\".%s", acc.TestDatabaseName, acc.TestSchemaName, pipeName)),
+				),
+			},
+			{
+				ConfigDirectory:   acc.ConfigurationDirectory("TestAcc_GrantOwnership/OnPipe"),
```
sfc-gh-asawicki, line 729: do we need import step?

pkg/resources/grant_ownership_acceptance_test.go
```diff
@@ -577,6 +579,281 @@ func TestAcc_GrantOwnership_InvalidConfiguration_MultipleTargets(t *testing.T) {
...
+					checkResourceOwnershipIsGranted(&sdk.ShowGrantOptions{
+						To: &sdk.ShowGrantsTo{
+							Role: sdk.NewAccountObjectIdentifier(accountRoleName),
+						},
+						// TODO: Fix this identifier
+					}, sdk.ObjectTypePipe, accountRoleName, fmt.Sprintf("%s\".\"%s\".%s", acc.TestDatabaseName, acc.TestSchemaName, pipeName), fmt.Sprintf("%s\".\"%s\".%s", acc.TestDatabaseName, acc.TestSchemaName, secondPipeName)),
+				),
+			},
+			{
+				ConfigDirectory:   acc.ConfigurationDirectory("TestAcc_GrantOwnership/OnAllPipes"),
```
sfc-gh-asawicki, line 781: do we need the import step?

pkg/resources/grant_ownership_acceptance_test.go
```diff
@@ -577,6 +579,281 @@ func TestAcc_GrantOwnership_InvalidConfiguration_MultipleTargets(t *testing.T) {
...
+		new(sdk.GrantOwnershipOptions),
+	)
+	assert.NoError(t, err)
+
+	return func() {
+		assert.NoError(t, client.Databases.Drop(ctx, databaseId, &sdk.DropDatabaseOptions{}))
+	}
+}
+
+func createAccountRole(t *testing.T, name string) func() {
```
sfc-gh-asawicki, line 820: There is already `createAccountRoleOutsideTerraform` method in `grant_privileges_to_account_role_acceptance_test.go`, let's reuse it for now

sfc-gh-asawicki (changes requested):

docs/resources/grant_ownership.md
```diff
@@ -0,0 +1,64 @@
+---
+page_title: "snowflake_grant_ownership Resource - terraform-provider-snowflake"
+subcategory: ""
+description: |-
+  
+---
+
+# snowflake_grant_ownership (Resource)
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yeah, I'll remove grant_ownership from the list of resources, and it'll remove itself after `make pre-push`

pkg/sdk/grants_impl.go
```diff
@@ -226,6 +256,138 @@ func (v *grants) Show(ctx context.Context, opts *ShowGrantOptions) ([]Grant, err
 	return resultList, nil
 }
 
+// TODO (remove ?): it was only for me to grasp what has to be done in this function, but may also leave it to guide others
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Adjusted list slightly to make it render properly

pkg/sdk/grants_impl.go
```diff
@@ -226,6 +256,138 @@ func (v *grants) Show(ctx context.Context, opts *ShowGrantOptions) ([]Grant, err
...
+
+		if err := revokeOperate(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (v *grants) grantOperateOnPipeTemporarily(ctx context.Context, pipeId SchemaObjectIdentifier, currentRole AccountObjectIdentifier) (func() error, error) {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: But still there's a case where:
1. Pipe (is running) and the current role we're using is e.g. accountadmin
2. We pause the pipe (right now, it doesn't matter how)
3. We grant ownership on pipe to role X
4. After granting ownership to role X we don't have permission on this pipe (even as accountadmin)
5. To resume the pipe we can either try to use `USE ROLE` on the role X (also hacky because it has to be granted to the user; we would have to enforce that or grant it ourselves) or grant OPERATE privilege on the current role for the time of resuming (like it is right now).

I think other options would require some additional work from the user pov. We can also just leave the pipe paused and figure something out from the perspective of pipe resource or document that pipes need manual work after ownership transfer.

pkg/resources/grant_ownership_acceptance_test.go
```diff
@@ -577,6 +579,281 @@ func TestAcc_GrantOwnership_InvalidConfiguration_MultipleTargets(t *testing.T) {
...
+				ConfigVariables: configVariables,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "account_role_name", accountRoleName),
+					resource.TestCheckResourceAttr(resourceName, "on.0.object_type", "DATABASE"),
+					resource.TestCheckResourceAttr(resourceName, "on.0.object_name", databaseName),
+					resource.TestCheckResourceAttr(resourceName, "id", fmt.Sprintf("ToAccountRole|%s||OnObject|DATABASE|%s", accountRoleFullyQualifiedName, databaseFullyQualifiedName)),
+				),
+			},
+			{
+				ConfigDirectory:   acc.ConfigurationDirectory("TestAcc_GrantOwnership/ForceOwnershipTransferOnCreate"),
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Same as above, I'll remove it.

pkg/resources/grant_ownership_acceptance_test.go
sfc-gh-jcieslak: It is supposed to show that we can transfer ownership of the objects owned by other roles than the current one (in the Create operation). Maybe this test was too much, because tbh without this working we wouldn't be able to retrieve the ownership back in the Delete operation. I can remove it.

pkg/resources/grant_ownership_acceptance_test.go
sfc-gh-jcieslak: Same as above, I'll remove it.

pkg/resources/grant_ownership_acceptance_test.go
sfc-gh-jcieslak: Same as above, I'll remove it.

pkg/resources/grant_ownership_acceptance_test.go
sfc-gh-jcieslak: Removed `createAccountRole` and used `createAccountRoleOutsideTerraform`. I was missing the register cleanup function added recently by you.

pkg/sdk/grants_impl.go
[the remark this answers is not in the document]
sfc-gh-asawicki: I think throwing the error there informing the user that manual work is required is the right option:
- it is not hacky
- nothing happens behind the curtains
- we knew and agreed that such cases will have to be described in the docs, so users are aware that for ownership grants manual work is expected

pkg/resources/grant_ownership_acceptance_test.go
sfc-gh-asawicki: I still don't get it. As I understand the setup (and the steps I put above), nothing happens in the create, because it is already set up correctly in the preconfig, right? So, I still don't know what this verifies.

If the delete needs to be verified, we should be explicit about it inside the test.

sfc-gh-jcieslak: My bad, the test was wrong and I didn't see it. I changed it, so it creates one role with database owned by that role, and the second role which is used in Terraform and on Create operation "steals" ownership from the first role. It was to show what happens if we would like to transfer ownership of the resources we don't own in the first place.

pkg/sdk/grants_impl.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Changed to a non-hacky version, with as much automation as possible (it depends on the user and what privileges are granted to the pipe).

pkg/resources/grant_ownership_acceptance_test.go
sfc-gh-asawicki: okay, now it makes sense, thanks!

pkg/sdk/testint/grants_integration_test.go
```diff
@@ -935,41 +1059,246 @@ func TestInt_GrantOwnership(t *testing.T) {
...
+
+		// Use a previously prepared role to create a pipe and grant MONITOR + OPERATE to the previously used role (ACCOUNTADMIN).
+		usePreviousRole := useRole(t, client, pipeRole.Name)
 
-		copyStatement := createPipeCopyStatement(t, table, stage)
+		pipe, pipeCleanup := createPipe(t, client, testDb(t), testSchema(t), random.AlphaN(20), copyStatement)
+		t.Cleanup(func() {
+			usePreviousRole = useRole(t, client, pipeRole.Name)
+			pipeCleanup()
+			usePreviousRole()
```
sfc-gh-asawicki, line 1227: what if pipeCleanup fails? we never get to usePReviousROle, should we defer it right after `usePreviousRole = useRole(t, client, pipeRole.Name)`? The same goes for other tests too.

sfc-gh-asawicki (changes requested):

docs/resources/grant_ownership.md
```diff
@@ -0,0 +1,64 @@
+---
```
sfc-gh-asawicki, line 1: one more thing: we should document the behavior for pipes, so that our users know how to use them. I think it would be best to add it already.

sfc-gh-jcieslak: Added a section in the `/tmp/grant_ownership.md.tmpl`

sfc-gh-asawicki (approved):

PR #2650, opened 2024-03-25
feat: Introduce shared function to suspend task roots

sfc-gh-jcieslak:
Refactor previously planned function for task suspension, which is also necessary to push the grant ownership resource forward.

## Test Plan

* [x] acceptance tests - tested on exiting tests (had to adjust new suspension code to mimic what was previously, otherwise I got a Snowflake error indicating that the root tasks were not suspended and the logic was wrong).
* [x] integration test to test new SDK function

## Reference
[CREATE TASK](https://docs.snowflake.com/en/sql-reference/sql/create-task)

### DIFF
--- a/pkg/resources/task.go
+++ b/pkg/resources/task.go
@@ -353,7 +353,6 @@ func CreateTask(d *schema.ResourceData, meta interface{}) (returnedErr error) {
 			if suspendErrs != nil {
 				return suspendErrs
 			}
-			// This can affect the read?
 			defer func() {
 				resumeErr := resumeSuspended()
 				if returnedErr != nil {
@@ -415,7 +414,6 @@ func UpdateTask(d *schema.ResourceData, meta interface{}) (returnedErr error) {
 	if err != nil {
 		return err
 	}
-	// TODO: This can affect the read?
 	defer func() {
 		resumeErr := resumeSuspended()
 		if returnedErr != nil {
@@ -666,7 +664,6 @@ func UpdateTask(d *schema.ResourceData, meta interface{}) (returnedErr error) {
 		}
 	}
 
-	// TODO Desc
 	enabled := d.Get("enabled").(bool)
 	if enabled {
 		if waitForTaskStart(ctx, client, taskId) != nil {
@@ -674,12 +671,10 @@ func UpdateTask(d *schema.ResourceData, meta interface{}) (returnedErr error) {
 		}
 	} else {
 		if err := client.Tasks.Alter(ctx, sdk.NewAlterTaskRequest(taskId).WithSuspend(sdk.Bool(true))); err != nil {
-			log.Printf("[WARN] failed to suspend task %s", taskId.FullyQualifiedName())
-			// TODO: Should it return anything or error ?????
-			// Previously it returned, but it's saying it's a warning
-			// return fmt.Errorf("[WARN] failed to suspend task %s", taskId.FullyQualifiedName())
+			return fmt.Errorf("failed to suspend task %s", taskId.FullyQualifiedName())
 		}
 	}
+
 	return ReadTask(d, meta)
 }
 
--- a/pkg/resources/task.go
+++ b/pkg/resources/task.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"slices"
@@ -284,7 +285,7 @@ func ReadTask(d *schema.ResourceData, meta interface{}) error {
 }
 
 // CreateTask implements schema.CreateFunc.
-func CreateTask(d *schema.ResourceData, meta interface{}) error {
+func CreateTask(d *schema.ResourceData, meta interface{}) (returnedErr error) {
 	client := meta.(*provider.Context).Client
 	ctx := context.Background()
 
@@ -348,24 +349,19 @@ func CreateTask(d *schema.ResourceData, meta interface{}) error {
 		precedingTasks := make([]sdk.SchemaObjectIdentifier, 0)
 		for _, dep := range after {
 			precedingTaskId := sdk.NewSchemaObjectIdentifier(databaseName, schemaName, dep)
-			rootTasks, err := sdk.GetRootTasks(client.Tasks, ctx, precedingTaskId)
-			if err != nil {
-				return err
+			resumeSuspended, suspendErrs := client.Tasks.TemporarilySuspendRootTasks(ctx, precedingTaskId, taskId)
+			if suspendErrs != nil {
+				return suspendErrs
 			}
-			for _, rootTask := range rootTasks {
-				// if a root task is started, then it needs to be suspended before the child tasks can be created
-				if rootTask.IsStarted() {
-					err := suspendTask(ctx, client, rootTask.ID())
-					if err != nil {
-						return err
-					}
-
-					// resume the task after modifications are complete as long as it is not a standalone task
-					if !(rootTask.Name == name) {
-						defer func(identifier sdk.SchemaObjectIdentifier) { _ = resumeTask(ctx, client, identifier) }(rootTask.ID())
-					}
+			// This can affect the read?
+			defer func() {
+				resumeErr := resumeSuspended()
+				if returnedErr != nil {
+					returnedErr = errors.Join(returnedErr, resumeErr)
+				} else {
+					returnedErr = resumeErr
 				}
-			}
+			}()
 			precedingTasks = append(precedingTasks, precedingTaskId)
 		}
 		createRequest.WithAfter(precedingTasks)
@@ -392,7 +388,7 @@ func CreateTask(d *schema.ResourceData, meta interface{}) error {
 }
 
 func waitForTaskStart(ctx context.Context, client *sdk.Client, id sdk.SchemaObjectIdentifier) error {
-	err := resumeTask(ctx, client, id)
+	err := client.Tasks.Alter(ctx, sdk.NewAlterTaskRequest(id).WithResume(sdk.Bool(true)))
 	if err != nil {
 		return fmt.Errorf("error starting task %s err = %w", id.FullyQualifiedName(), err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/resources/task.go
```diff
@@ -284,7 +285,7 @@ func ReadTask(d *schema.ResourceData, meta interface{}) error {
 }
 
 // CreateTask implements schema.CreateFunc.
-func CreateTask(d *schema.ResourceData, meta interface{}) error {
+func CreateTask(d *schema.ResourceData, meta interface{}) (returnedErr error) {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: The test I mentioned:
```
func TestNamedReturn(t *testing.T) {
	g := func(e error) error {
		if e != nil {
			return e
		}
		return nil
	}

	def := func(e error) error {
		if e != nil {
			return e
		}
		return nil
	}

	f := func(mainErr error, anotherErr error, deferredErr error) (returnedErr error) {
		defer func() {
			returnedErr = errors.Join(returnedErr, def(deferredErr))
		}()
		if mainErr != nil {
			return mainErr
		}
		return g(anotherErr)
	}

	me := errors.New("main function error")
	ae := errors.New("another function error")
	de := errors.New("deferred function error")

	_, _ = ae, de

	t.Run("no errors", func(t *testing.T) {
		returnedErr := f(nil, nil, nil)

		require.Nil(t, returnedErr)
	})

	t.Run("error in main func, no errors in deferred", func(t *testing.T) {
		returnedErr := f(me, nil, nil)

		// they are not equal, because the packaging takes place in defer
		require.NotEqual(t, returnedErr, me)
		require.Contains(t, returnedErr.Error(), me.Error())
		require.NotContains(t, returnedErr.Error(), ae.Error())
		require.NotContains(t, returnedErr.Error(), de.Error())
	})

	t.Run("error in another func, no errors in deferred", func(t *testing.T) {
		returnedErr := f(nil, ae, nil)

		// they are not equal, because the packaging takes place in defer
		require.NotEqual(t, returnedErr, ae)
		require.NotContains(t, returnedErr.Error(), me.Error())
		require.Contains(t, returnedErr.Error(), ae.Error())
		require.NotContains(t, returnedErr.Error(), de.Error())
	})

	t.Run("no errors in main and in another, error in deferred", func(t *testing.T) {
		returnedErr := f(nil, nil, de)

		require.NotEqual(t, returnedErr, de)
		require.NotContains(t, returnedErr.Error(), me.Error())
		require.NotContains(t, returnedErr.Error(), ae.Error())
		require.Contains(t, returnedErr.Error(), de.Error())
	})

	t.Run("error in main func, error in deferred", func(t *testing.T) {
		returnedErr := f(me, nil, de)

		require.NotEqual(t, returnedErr, me)
		require.NotEqual(t, returnedErr, de)
		require.Contains(t, returnedErr.Error(), me.Error())
		require.NotContains(t, returnedErr.Error(), ae.Error())
		require.Contains(t, returnedErr.Error(), de.Error())
	})

	t.Run("error in another func, error in deferred", func(t *testing.T) {
		returnedErr := f(nil, ae, de)

		require.NotEqual(t, returnedErr, ae)
		require.NotEqual(t, returnedErr, de)
		require.NotContains(t, returnedErr.Error(), me.Error())
		require.Contains(t, returnedErr.Error(), ae.Error())
		require.Contains(t, returnedErr.Error(), de.Error())
	})
}

```

pkg/resources/task.go
```diff
@@ -348,24 +349,11 @@ func CreateTask(d *schema.ResourceData, meta interface{}) error {
 		precedingTasks := make([]sdk.SchemaObjectIdentifier, 0)
 		for _, dep := range after {
 			precedingTaskId := sdk.NewSchemaObjectIdentifier(databaseName, schemaName, dep)
-			rootTasks, err := sdk.GetRootTasks(client.Tasks, ctx, precedingTaskId)
+			resumeSuspended, err := client.Tasks.TemporarilySuspendRootTasks(ctx, precedingTaskId, taskId)
+			defer func() { returnedErr = errors.Join(returnedErr, resumeSuspended()) }()
```
sfc-gh-asawicki, line 353: first defer or first check for err and return? (in update there is the same ordering but different one is in delete)

pkg/sdk/tasks_gen.go
```diff
@@ -14,6 +14,7 @@ type Tasks interface {
 	ShowByID(ctx context.Context, id SchemaObjectIdentifier) (*Task, error)
 	Describe(ctx context.Context, id SchemaObjectIdentifier) (*Task, error)
 	Execute(ctx context.Context, request *ExecuteTaskRequest) error
+	TemporarilySuspendRootTasks(ctx context.Context, depId SchemaObjectIdentifier, id SchemaObjectIdentifier) (func() error, error)
```
sfc-gh-asawicki, line 17: I think that a more clear API would be to have:
- method SuspendRootTasks, with taskId instead of depId, and returning the IDs of suspended tasks
- method ResumeTasks, which accepts the list of suspended tasks

Afterall, the behavior will be almost the same, but more explicit and readable for the user.

pkg/sdk/tasks_impl_gen.go
```diff
@@ -65,6 +67,51 @@ func (v *tasks) Execute(ctx context.Context, request *ExecuteTaskRequest) error
...
+// check if the task is started. If it is, then suspend it and add to the list of tasks to resume only if the root task name
+// is not that same as taskId name.
+//
+// Returns:
+// A callback function to resume all the suspended root tasks.
+// An error joined from all the suspend calls, nil if no error was returned during by task suspending calls.
+func (v *tasks) TemporarilySuspendRootTasks(ctx context.Context, depId SchemaObjectIdentifier, taskId SchemaObjectIdentifier) (func() error, error) {
+	rootTasks, err := GetRootTasks(v.client.Tasks, ctx, depId)
+	if err != nil {
+		return func() error { return nil }, err
```
sfc-gh-asawicki, line 80: bleh

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -592,4 +592,30 @@ func TestInt_Tasks(t *testing.T) {
 		err := client.Tasks.Execute(ctx, executeRequest)
 		require.NoError(t, err)
 	})
+
+	t.Run("temporarily suspend root tasks", func(t *testing.T) {
```
sfc-gh-asawicki, line 596: Okay, let's create more complicated tests please:
- more roots (started/not started - it should be possible to have only one started but let's try)
- deeper tree

I started to wonder if the current implementation of the resource is the correct one. Check the description for `RESUME | SUSPEND` in https://docs.snowflake.com/en/sql-reference/sql/alter-task#parameters: `Before resuming the root task of your Task Graph, resume all child tasks. To recursively resume the root task’s child tasks, use [SYSTEM$TASK_DEPENDENTS_ENABLE](https://docs.snowflake.com/en/sql-reference/functions/system_task_dependents_enable).`. We don't want to resume all, but we should resume some, right? Maybe let's experiment a bit.

pkg/resources/task.go
```diff
@@ -348,24 +349,11 @@ func CreateTask(d *schema.ResourceData, meta interface{}) error {
...
 			}
-			for _, rootTask := range rootTasks {
-				// if a root task is started, then it needs to be suspended before the child tasks can be created
-				if rootTask.IsStarted() {
-					err := suspendTask(ctx, client, rootTask.ID())
-					if err != nil {
-						return err
-					}
-
-					// resume the task after modifications are complete as long as it is not a standalone task
```
sfc-gh-asawicki, line 363: I don't get that part TBH. Why this is compared with the name of the task? (I ask it there because you have moved this logic to the SDK too).

pkg/sdk/tasks_gen.go
sfc-gh-asawicki: I meant more that depId is a enigmatic name ;) I am okay with having two ids, but let's name them better.

pkg/resources/task.go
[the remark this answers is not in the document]
sfc-gh-asawicki: Yes, let's do it like this at least for now, but add this as a consideration for the follow-up ticket (I mean handling the resume errors).

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-asawicki: Please add my above comment to the follow-up ticket too.

pkg/sdk/tasks_impl_gen.go
```diff
@@ -65,6 +67,48 @@ func (v *tasks) Execute(ctx context.Context, request *ExecuteTaskRequest) error
 	return validateAndExec(v.client, ctx, opts)
 }
 
+// TODO(SNOW-1277135): See if depId is necessary or could be removed
```
sfc-gh-asawicki, line 70: depId is not there anymore, but we cal live with this (correct in in some other PR please)

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -592,4 +592,135 @@ func TestInt_Tasks(t *testing.T) {
...
+		rootTask := createTaskWithRequest(t, sdk.NewCreateTaskRequest(rootTaskId, sql).WithSchedule(sdk.String("60 minutes")))
+
+		middleTaskId := sdk.NewSchemaObjectIdentifier(testDb(t).Name, testSchema(t).Name, random.String())
+		middleTask := createTaskWithRequest(t, sdk.NewCreateTaskRequest(middleTaskId, sql).WithAfter([]sdk.SchemaObjectIdentifier{rootTask.ID()}))
+
+		childTaskId := sdk.NewSchemaObjectIdentifier(testDb(t).Name, testSchema(t).Name, random.String())
+		childTask := createTaskWithRequest(t, sdk.NewCreateTaskRequest(childTaskId, sql).WithAfter([]sdk.SchemaObjectIdentifier{middleTask.ID()}))
+
+		require.NoError(t, client.Tasks.Alter(ctx, sdk.NewAlterTaskRequest(childTask.ID()).WithResume(sdk.Bool(true))))
+		t.Cleanup(func() {
```
sfc-gh-asawicki, line 692: Why do we need to resume them in cleanup?

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -592,4 +592,135 @@ func TestInt_Tasks(t *testing.T) {
...
+			require.NoError(t, client.Tasks.Alter(ctx, sdk.NewAlterTaskRequest(childTask.ID()).WithSuspend(sdk.Bool(true))))
+		})
+
+		require.NoError(t, client.Tasks.Alter(ctx, sdk.NewAlterTaskRequest(rootTask.ID()).WithResume(sdk.Bool(true))))
+		t.Cleanup(func() {
+			require.NoError(t, client.Tasks.Alter(ctx, sdk.NewAlterTaskRequest(rootTask.ID()).WithSuspend(sdk.Bool(true))))
+		})
+
+		tasksToResume, err := client.Tasks.SuspendRootTasks(ctx, middleTask.ID(), middleTask.ID())
```
sfc-gh-asawicki, line 701: is this invocation right? I guess we are invoking this function with two different task ids and that's why we need both?

**EDIT**: okay, I see that for update we invoke it with the same id... so yes, I think that what you wrote about checking the necessity for two of them is valid in a follow-up task.

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -592,4 +592,135 @@ func TestInt_Tasks(t *testing.T) {
...
+		require.NoError(t, err)
+		require.NotEmpty(t, tasksToResume)
+
+		rootTaskStatus, err := client.Tasks.ShowByID(ctx, rootTask.ID())
+		require.NoError(t, err)
+		require.Equal(t, sdk.TaskStateSuspended, rootTaskStatus.State)
+
+		childTaskStatus, err := client.Tasks.ShowByID(ctx, childTask.ID())
+		require.NoError(t, err)
+		require.Equal(t, sdk.TaskStateStarted, childTaskStatus.State)
```
sfc-gh-asawicki, line 711: hmmm, this is interesting that the child task is not suspended automatically when the root task is. But I suppose this is how SF works.

sfc-gh-asawicki (approved):

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: Yup yup

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: Yeah, It's confusing that sometimes it was with the same id and sometimes we need to pass two different ones. For now, I implemented it in a way it was working before, but in the follow-up task I created, this should be analyzed (to answer questions like why we need 2 parameters? If not, can we remove one of them and stick with only one?).

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: It's necessary because dropping tasks in the cleanup is modifying the task graph. To resolve those errors we can either modify the task setup first to create tasks and then alter them to create connections (creating them in a different order will make t.Cleanup's run in the right order, but we would have to add alters to add "after's" to create graph) or leave it as is and create them in the "wrong" order with after's specified during create and pausing in t.Cleanup's to not error during calling DROP TASK X.

pkg/sdk/tasks_impl_gen.go
sfc-gh-jcieslak: Will do in the next pr.

sfc-gh-asawicki (approved):

Issue #2653, opened 2024-03-26
Add S3_STAGE_VPCE_DNS_NAME session parameter


sfc-gh-asawicki:
### Terraform CLI and Provider Versions

0.87.2

### Use Cases or Problem Statement

S3_STAGE_VPCE_DNS_NAME is missing (https://docs.snowflake.com/en/sql-reference/parameters#s3-stage-vpce-dns-name).

### Proposal

Add it.

### How much impact is this issue causing?

Low

Issue #2652, opened 2024-03-26
Error: MinClusterCount must be less than or equal to MaxClusterCount - When increasing min_cluster_count param of warehouse


wpl-sf-nbirch:
### Terraform CLI and Provider Versions

0.87.2

### Terraform Configuration

```terraform
Increase min_cluster_count from 1 to 2.

Original:
resource "snowflake_warehouse" "reporting_wh" {
  name                                = "REPORTING_WH"
  auto_resume                         = true
  auto_suspend                        = 600
  comment                             = "Reporting Warehouse"
  enable_query_acceleration           = false
  initially_suspended                 = true
  max_cluster_count                   = 10
  min_cluster_count                   = 1
  scaling_policy                      = "STANDARD"
  warehouse_size                      = "SMALL"
  warehouse_type                      = "STANDARD"
  statement_timeout_in_seconds        = 7200
  max_concurrency_level               = 16
}

Updated:
resource "snowflake_warehouse" "reporting_wh" {
  name                                = "REPORTING_WH"
  auto_resume                         = true
  auto_suspend                        = 600
  comment                             = "Reporting Warehouse"
  enable_query_acceleration           = false
  initially_suspended                 = true
  max_cluster_count                   = 10
  min_cluster_count                   = 2
  scaling_policy                      = "STANDARD"
  warehouse_size                      = "SMALL"
  warehouse_type                      = "STANDARD"
  statement_timeout_in_seconds        = 7200
  max_concurrency_level               = 16
}
```

### Expected Behavior

The minimum cluster count should increase to 2 nodes and allow auto-scale up to 10 nodes.

### Actual Behavior

**Error: MinClusterCount must be less than or equal to MaxClusterCount**
with snowflake_warehouse.reporting_wh
on snowflake_warehouses.tf line 125, in resource "snowflake_warehouse" "reporting_wh":
resource "snowflake_warehouse" "reporting_wh" {

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

High

wpl-sf-nbirch:
It seems setting the min_cluster_count and max_cluster_count to the same value 2 works, and then increasing the max_cluster_count back to 10 works, as a two step process.

sfc-gh-asawicki:
Hey @wpl-sf-nbirch. Thanks for reporting the issue.

[...]

Issue #2642, opened 2024-03-21
snowflake_procedure execute_as=CALLER update does not work on existing resource


krishreddy56:
### Terraform CLI and Provider Versions

Terraform: 1.7.2
Provider: 0.87

### Terraform Configuration

```terraform
resource "snowflake_procedure" "test_procedure" {
  database = "TEST"
  schema   = "TEST_SCHEMA
  name     = "TEST_PROCEDURE"
  arguments {
    name = "TEST_DATE"
    type = "DATE"
  }
  arguments {
    name = "NUMBER_OF_DAYS"
    type = "NUMBER"
  }
  language    = "SQL"
  comment     = "sample procedure"
  return_type = "TABLE (ROWS_UNLOADED NUMBER, INPUT_BYTES NUMBER, OUTPUT_BYTES NUMBER)"
  execute_as = "CALLER"
  statement = templatefile("${path.module}/test.sql", {
    bucket_name              = "test-bucket",
    storage_integration_name = "TEST_STORAGE_INTEGRATION,
  })
}
```

### Expected Behavior

I have a snowflake procedure created with `execute_as=OWNER` through terraform. Modifying the resource to update `execute_as=CALLER `should just allow the procedure to be updated. Tried modifying the procedure in the snowflake console and it works.

### Actual Behavior

terraform apply fails with below error

│ Error: 001003 (42000): SQL compilation error:
│ syntax error line 1 at position 93 unexpected 'CALLER'.

### Steps to Reproduce

1. Run`terraform apply` with execute_as=OWNER
2. Procedure is created
3. Update the resource with execute_as=CALLER to modify the procedure
4. Run `terraform apply`and it fails

### How much impact is this issue causing?

High

### Additional Information

We had to destroy the procedure and then recreate it using `execute_as=CALLER` as a workaround

sfc-gh-asawicki:
Hey @krishreddy56. Thanks for reaching out to us.

Please run the terraform with `TF_LOG=DEBUG` flag enabled. You should be able to see the ALTER statement run for execute_as update (it should be close above to the error you have shared).

krishreddy56:
Re ran the terraform apply by enabling the debug log. I see below in the logs

```
2024-03-21T14:30:07.082-0600 [INFO]  Starting apply for module.sql_compliance.snowflake_procedure.test_procedure
[...]

Issue #2606, opened 2024-03-07
snowflake_schema errors when comment is changed to null (empty string)


Bryan-Meier:
### Terraform CLI and Provider Versions

Terraform v1.5.5
snowflake-labs/snowflake v0.87.0

### Terraform Configuration

```terraform
resource "snowflake_schema" "sf_schema" {
  for_each = local.database_schemas

  database            = each.value.database_name
  name                = each.value.schema_name
  comment             = each.value.schema_comment
  is_transient        = each.value.is_transient
  is_managed          = each.value.is_managed
  data_retention_days = each.value.data_retention_days

  depends_on = [
    snowflake_database.sf_database
  ]
}
```

### Expected Behavior

When a schema has a comment that was set previously, the comment should be able to be removed from the schema in a later apply since comments are not a requirement by Snowflake. This used to work in a previous version of the provider. I would expect that when a comment is changed from an existing comment to an empty string, the apply should run without error.

### Actual Behavior

If a comment exists on the schema from a previous apply and in a subsequent apply you try to remove it by setting the comment value to an empty string, it's throwing the following error:

│ Error: error updating schema comment on SOME_DB|SOME_SCHEMA err = [schemas.go:195] at least one of SchemaSet fields [DataRetentionTimeInDays MaxDataExtensionTimeInDays DefaultDDLCollation Comment] must be set
│ 
│   with snowflake_schema.sf_schema["SOME_DB|SOME_SCHEMA"],
│   on main.tf line 143, in resource "snowflake_schema" "sf_schema":
│  143: resource "snowflake_schema" "sf_schema" {

### Steps to Reproduce

1. Set a comment on a schema
2. Apply the comment
3. Remove the comment on the same schema
4. Apply the removal

### How much impact is this issue causing?

High

### Additional Information

When we went to upgrade from v0.71.0 to 0.87.0 and were met with error above. We are basically making sure we don't remove any comments at this point.

sfc-gh-asawicki:
Hey @Bryan-Meier. Thanks for reporting the issue.

This is an easy one, we should be able to add it to one of the other changes. We should have a fix with the next release.

sfc-gh-asawicki:
Hey @Bryan-Meier. Fix released in v0.88.0.

Please check and close the issue if the problem is solved for you.

Bryan-Meier:
I will check it out. Thank you!

Bryan-Meier:
[...]

PR #2654, opened 2024-03-26
fix: Fix issues 2606 2642 2652 2653

sfc-gh-asawicki:
- Fix unsetting comment on schema (references #2606)
- Fix changing the execute as on procedure (references #2642)
- Fix setting only min cluster count on warehouse (references #2652)
- Add missing S3_STAGE_VPCE_DNS_NAME session parameter (references #2653)

### DIFF
--- a/pkg/sdk/parameters_impl.go
+++ b/pkg/sdk/parameters_impl.go
@@ -236,6 +236,8 @@ func (sessionParametersUnset *SessionParametersUnset) setParam(parameter Session
 		sessionParametersUnset.QuotedIdentifiersIgnoreCase = Bool(true)
 	case SessionParameterRowsPerResultset:
 		sessionParametersUnset.RowsPerResultset = Bool(true)
+	case SessionParameterS3StageVpceDnsName:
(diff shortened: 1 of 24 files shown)


pkg/sdk/parameters_impl.go
```diff
@@ -107,6 +107,8 @@ func (sessionParameters *SessionParameters) setParam(parameter SessionParameter,
 			return fmt.Errorf("ROWS_PER_RESULTSET session parameter is an integer, got %v", value)
 		}
 		sessionParameters.RowsPerResultset = Pointer(v)
+	case SessionParameterS3StageVpceDnsName:
```
sfc-gh-jcieslak, line 110: Shouldn't this be also updated?
```go
func (sessionParametersUnset *SessionParametersUnset) setParam(parameter SessionParameter) error {
```

sfc-gh-asawicki: Yep, good catch!

sfc-gh-jcieslak (approved):

Issue #2640, opened 2024-03-21
snowflake_view crashes on import of existing view


lastlee:
### Terraform CLI and Provider Versions

Terraform 1.5.7
Snowflake 0.87.2

### Terraform Configuration

```terraform
resource "snowflake_view" "secure_view" {
  database = snowflake_database.subscription.name
  schema   = snowflake_schema.subscription_classification.name
  name     = var.classification_table
  comment  = "Classification View of the union of classification tables"

  statement  = <<-SQL
            select * from AB1_SUBSCRIPTION.CLASSIFICATION.CLASSIFICATION
              union
            select * from AB2_SUBSCRIPTION.CLASSIFICATION.CLASSIFICATION
SQL
  or_replace = true
  is_secure  = true
}
```

### Expected Behavior

The import of `'SUBSCRIPTION|CLASSIFICATION|CLASSIFICATION'` should import successfully

### Actual Behavior

```
snowflake_view.secure_view: Importing from ID "SUBSCRIPTION|CLASSIFICATION|CLASSIFICATION"...
snowflake_view.secure_view: Import prepared!
  Prepared snowflake_view for import
snowflake_view.secure_view: Refreshing state... [id=SUBSCRIPTION|CLASSIFICATION|CLASSIFICATION]
╷
│ Error: Plugin did not respond
│
│ The plugin encountered an error, and failed to respond to the
│ plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more
│ details.
╵


Stack trace from the terraform-provider-snowflake_v0.87.2 plugin:

panic: runtime error: index out of range [0] with length 0

goroutine 83 [running]:
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/snowflake.(*ViewSelectStatementExtractor).consumeToken(0xc000edf068, {0x208e28a, 0x6})
        github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/snowflake/parser.go:129 +0x14f
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/snowflake.(*ViewSelectStatementExtractor).Extract(0xc000edf068)
        github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/snowflake/parser.go:30 +0x92
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources.ReadView(0xc0005fed00, {0x1cc21a0?, 0xc000288a20?})
[...]

PR #2655, opened 2024-03-26
fix: Fix handling secure views read and import

sfc-gh-asawicki:
To import the secure view the role owning the view is needed. As https://docs.snowflake.com/en/sql-reference/sql/create-view#usage-notes state:
> By design, the [SHOW VIEWS](https://docs.snowflake.com/en/sql-reference/sql/show-views) command does not provide information about secure views. To view information about a secure view, you must use the [VIEWS](https://docs.snowflake.com/en/sql-reference/info-schema/views) view in the Information Schema and you must use the role that owns the view.

This change:
- tests and proves #2640 
- checks the parser for view using union statement (that was the first idea for the source of error)
- handles the empty `text` from `SHOW VIEWS` command gracefully (no panic like before)
- adds a proper description with documentation link to the `snowflake_view` documentation

References: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2640#issuecomment-2020739114

### DIFF
--- a/pkg/resources/view_acceptance_test.go
+++ b/pkg/resources/view_acceptance_test.go
@@ -7,16 +7,16 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
-
 	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
 
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/resources/view_acceptance_test.go
```diff
@@ -415,6 +415,54 @@ func TestAcc_View_copyGrants(t *testing.T) {
...
+				),
+			},
+			// try to import secure view without being its owner (proves https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2640)
+			{
+				PreConfig: func() {
+					createAccountRoleOutsideTerraform(t, roleName)
+					registerAccountRoleCleanup(t, roleName)
+					alterViewOwnershipExternally(t, viewName, roleName)
+				},
+				ResourceName: "snowflake_view.test",
```
sfc-gh-jcieslak, line 448: For curiosity, Config can be only passed in the first step (and skipped in the others) or it should be also defined in the 2nd and 3rd step?

sfc-gh-asawicki: The 2nd and 3rd steps are import testing steps, they do not require config because all they need is resource id:
```
	// ImportState, if true, will test the functionality of ImportState
	// by importing the resource with ResourceName (must be set) and the
	// ID of that resource.
	ImportState bool

	// ImportStateId is the ID to perform an ImportState operation with.
	// This is optional. If it isn't set, then the resource ID is automatically
	// determined by inspecting the state for ResourceName's ID.
	ImportStateId string
```
the second step is an err one, but the third verifies correctness with the previous state of the resource:
```
	// ImportStateVerify, if true, will also check that the state values
	// that are finally put into the state after import match for all the
	// IDs returned by the Import.  Note that this checks for strict equality
	// and does not respect DiffSuppressFunc or CustomizeDiff.
	//
	// By default, the prior resource state and import resource state are
	// matched by the "id" attribute. If the "id" attribute is not implemented
	// or another attribute more uniquely identifies the resource, set the
	// ImportStateVerifyIdentifierAttribute field to adjust the attribute for
	// matching.
	//
	// If certain attributes cannot be correctly imported, set the
	// ImportStateVerifyIgnore field.
	ImportStateVerify bool
```

sfc-gh-jcieslak: Hmm, true. Oky

PR #2660, opened 2024-03-28
chore: Update grant examples in all resources

sfc-gh-asawicki:
- Removed all deprecated examples (using deprecated `_grant` resources) from the non-deprecated resources/datasources (only one).
- Removed using deprecated `_grant` and `_grants` resources from resource and datasource acceptance tests (only from the non-deprecated resources/datasources).
- Added migration guide to the readme.


examples/resources/snowflake_stage/resource.tf
```diff
@@ -6,11 +6,3 @@ resource "snowflake_stage" "example_stage" {
   schema      = "EXAMPLE_SCHEMA"
   credentials = "AWS_KEY_ID='${var.example_aws_key_id}' AWS_SECRET_KEY='${var.example_aws_secret_key}'"
 }
-
-resource "snowflake_stage_grant" "grant_example_stage" {
```
sfc-gh-jcieslak, line 10: Should we remove this ownership grant or add todo for SNOW-991423 (grant ownership implementation)?

pkg/resources/file_format_acceptance_test.go
```diff
@@ -70,12 +70,10 @@ func TestAcc_FileFormatCSV(t *testing.T) {
...
-				ResourceName:      "snowflake_file_format.test",
-				ImportState:       true,
-				ImportStateVerify: true,
-				ImportStateVerifyIgnore: []string{
-					"enable_multiple_grants", // feature flag attribute not defined in Snowflake, can't be imported
-				},
+				ResourceName:            "snowflake_file_format.test",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{},
```
sfc-gh-jcieslak, line 76: ImportStateVerifyIgnore can be fully removed

examples/resources/snowflake_stage/resource.tf
sfc-gh-asawicki: No other resource have such an example, so my choice was to remove and not do a follo-up just for this one

PR #2664, opened 2024-03-29
fix: Adjust dynamic tables after BCR-1543

sfc-gh-jcieslak:
[BCR-1543](https://docs.snowflake.com/en/release-notes/bcr-bundles/2024_02/bcr-1543#return-value-behavior-for-the-show-dynamic-table-function) introduced changes to the dynamic table output for `scheduling_state` column.

## Test Plan

* [x] Test that was failing before, now passes with newly added `ACTIVE` scheduling state


sfc-gh-asawicki (approved):
Docs were not regenerated after description change. I will merge it and fix after.

PR #2665, opened 2024-03-29
fix: grant read operation

sfc-gh-jcieslak:
This pr resolves an issue that was happening when a granted resource was removed by hand. This touches all the `grant privilege to` resources (others too, but they somewhat already resolved this by marking the resource as removed; the other resources are only missing nested `if` check that would indicate for which error we're making the resources as removed. Right now, it's happening for every kind of error in the `SHOW` command). I adjusted those resource to:
1. Check if the granted resource is present (e.g. grant x to database role -> in Read operation check if database role is present).
2. Check the error type in the `SHOW GRANTS`, so we'll only mark the resource as removed only for `object does not exist or not authorized` type of errors.

## Test Plan

* [x] acceptance tests that show resources are marked as removed without dependent resources (grant privileges to share have only one test because Snowflake is preventing you from removing dependent objects like databases when they're granted to share)


pkg/resources/grant_privileges_to_share_acceptance_test.go
```diff
@@ -521,6 +521,45 @@ func TestAcc_GrantPrivilegesToShare_NoOnOption(t *testing.T) {
...
+				},
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_GrantPrivilegesToShare/OnCustomShare"),
+				ConfigVariables: configVariables,
+			},
+			{
+				PreConfig:       func() { shareCleanup() },
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_GrantPrivilegesToShare/OnCustomShare"),
+				ConfigVariables: configVariables,
+				// The error occurs in the Create operation, indicating the Read operation removed the resource from the state in the previous step.
+				ExpectError: regexp.MustCompile("An error occurred when granting privileges to share"),
```
sfc-gh-asawicki, line 557: hmm, haven't we discussed that this could work like e.g. schema resource (check before granting, setID "" if object does not exist)? Or maybe it was other topic?

sfc-gh-asawicki (commented):
Just one question, valid for all the resources covered in this PR.

pkg/resources/grant_privileges_to_share_acceptance_test.go
sfc-gh-jcieslak: Hmm, I'm pretty certain we wanted to do it in this way, where the check is in the Read operation and not the Create operation (schema also does that). I'm not sure if such a check is necessary for the Create, because it will fail anyway and won't create anything. Read matters for grant resources because without an additional check there, we will create a plan and fail at the Update operation which creates an infinite plan.

sfc-gh-asawicki: Followed-up privately. Discussions resulted in: please add similar tests for schema to confirm that the behavior is the same.

sfc-gh-asawicki (approved):

PR #2671, opened 2024-04-03
chore: with_grant_option diff drift follow-up

sfc-gh-jcieslak:
A follow-up for https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2608
Changes:
- Added `GrantOptionFor` option during revoke, so only `with_grant_option` will be removed instead of the whole grant, making the process less destructive

Tests:
- Tests added in the previous change


sfc-gh-jcieslak:
Duplicate

PR #2687, opened 2024-04-08
fix: do not drop dynamic tables when query changes

sonya:
Hi, it's me again. I'm proposing a "fix", but it may be considered a behavior change as the `or_replace` attribute on `snowflake_dynamic_table` becomes a no-op. 

## Summary of changes

This change allows dynamic tables with changed queries to be updated in-place.

Previously, there was a `or_replace` attribute on the snowflake_dynamic_table resource that was meant to specify whether `OR REPLACE` is meant to be used if the table already exists. However, this attribute was confusing to use because it often resulted in spurious changes in terraform plans after applying.

A separate issue with dynamic tables was that we have been dropping and recreating dynamic tables when the `query` or `refresh_mode` properties changed, when we could take a less destructive action by calling `CREATE OR REPLACE`. The presence of `or_replace` as a configured attribute made this more complicated.

This commit removes the ability for users to set `or_replace` on dynamic table configurations, and instead will always use `CREATE OR REPLACE` for dynamic table creation. With this simplification, we can then check for changes to the `query` and `refresh_mode` arguments when updating the dynamic table, and either use `CREATE OR REPLACE` or `ALTER DYNAMIC TABLE` depending on which attributes have changed. This allows users to update the query and refresh mode without causing temporary unavailability.

## Test Plan

I have tested this on our actual deployed stack and can confirm it executes a single `create or replace` statement instead of a `drop` and a `create`.

For acceptance tests, I was thinking of what acceptance test to write but it seems like the existing acceptance tests already test everything I want to test (the only behavioral difference between using `create or replace` instead of drop-then-create is that the former has less downtime)

## References
* I vaguely recall there was a thread on one of my PRs where we discussed `create or replace` vs. `ForceNew` and decided to proceed with `ForceNew` at the time. I can't find the conversation anymore, but after using the provider in production for awhile we are convinced `create or replace` is the superior option.

### DIFF
--- a/docs/resources/dynamic_table.md
+++ b/docs/resources/dynamic_table.md
@@ -42,7 +42,7 @@ resource "snowflake_dynamic_table" "dt" {
 
 - `comment` (String) Specifies a comment for the dynamic table.
 - `initialize` (String) Initialize trigger for the dynamic table. Can only be set on creation. Available options are ON_CREATE and ON_SCHEDULE.
-- `or_replace` (Boolean) Specifies whether to replace the dynamic table if it already exists.
+- `or_replace` (Boolean) This argument is deprecated and setting it has no effect. All dynamic tables are created with `CREATE OR REPLACE`
 - `refresh_mode` (String) INCREMENTAL to use incremental refreshes, FULL to recompute the whole table on every refresh, or AUTO to let Snowflake decide.
 
 ### Read-Only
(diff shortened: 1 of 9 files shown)


sfc-gh-asawicki:

We have just recently (yesterday) discussed inside the team the general approaches we want to take with all the resources on our route to V1. ForceNew vs OR REPLACE was one of them. Currently, the outcome of these discussions for this particular topic (and the second with OR REPLACE is used with COPY GRANTS) is as follows:
- Remove OR REPLACE from ale resource schemas.
- Stick with forceNew for the V1.
- Consider using OR REPLACE for the forceNew parameters for every resource after V1.

We have considered using OR REPLACE automatically, to have the atomicity but we've chosen to stick with the explicit drop+recreate visible on the terraform plan for now.

sfc-gh-asawicki:
Yes, we stick with drop + create for the time-being, thanks for understanding. :)

PR #2690, opened 2024-04-10
feat: add snowflake_grant_application_role resource

sfc-gh-swinkler:
Adds a new resource for grant_application_role. Also adds some stuff to SDK to make it work. Relies on an app package to exist, which may not exist in the current account. No application resource currently exists to use.

## Test Plan

* [x] acceptance tests

* [ ] …

## References

https://docs.snowflake.com/en/sql-reference/sql/grant-application-role


pkg/resources/grant_application_role_acceptance_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_GrantApplicationRole_accountRole(t *testing.T) {
```
sfc-gh-asawicki, line 20: both tests failed in the current build

pkg/resources/grant_application_role_acceptance_test.go
```diff
@@ -0,0 +1,133 @@
...
+	applicationName := "int_test_app"
```
sfc-gh-asawicki, line 21: This app is not created anywhere, so this test is missing the setup? Or am I missing something?

pkg/sdk/testint/application_roles_gen_integration_test.go
```diff
@@ -118,42 +185,4 @@ func TestInt_ApplicationRoles(t *testing.T) {
 		require.Equal(t, sdk.ObjectTypeRole, grants[0].GrantedTo)
 		require.Equal(t, sdk.NewAccountObjectIdentifier(appName), grants[0].GrantedBy)
 	})
-
-	t.Run("show grants to application", func(t *testing.T) {
```
sfc-gh-asawicki, line 122: why this test was removed?

sfc-gh-asawicki (changes requested):

pkg/resources/grant_application_role.go
```diff
@@ -0,0 +1,199 @@
...
+				if err := d.Set("name", parts[0]); err != nil {
+					return nil, err
+				}
+				switch parts[1] {
+				case "ROLE":
+					if err := d.Set("parent_role_name", strings.Trim(parts[2], "\"")); err != nil {
+						return nil, err
+					}
+				case "APPLICATION":
+					if err := d.Set("application_name", strings.Trim(parts[2], "\"")); err != nil {
```
sfc-gh-jcieslak, line 69: I'm not sure if calling `strings.Trim` is right, when e.g. in the Create we're saving the id with `FullyQualifiedName`. That makes it inconsistent with what is saved to the state. There were some issues created about it, so maybe we should just always save it with quotes?

pkg/resources/grant_application_role.go
```diff
@@ -0,0 +1,199 @@
...
+	applicationRoleIdentifier := sdk.NewDatabaseObjectIdentifierFromFullyQualifiedName(applicationRoleName)
+	objectType := parts[1]
+	targetIdentifier := parts[2]
+	ctx := context.Background()
+	grants, err := client.Grants.Show(ctx, &sdk.ShowGrantOptions{
+		Of: &sdk.ShowGrantsOf{
+			ApplicationRole: applicationRoleIdentifier,
+		},
+	})
+	if err != nil {
```
sfc-gh-jcieslak, line 121: We recently did a change to detect missing dependent resources. Right now with SHOW GRANTS you can only detect if the application role is missing, but can't really tell if the parent_role or application is missing. Could you please adjust it to the recent change where in Read we check the existence of other objects (parent_role or application in this case). You can check the code below (it's from the snowflake_grant_privileges_to_account_role):
```go
	if _, err := client.Roles.ShowByID(ctx, sdk.NewShowByIdRoleRequest(id.RoleName)); err != nil && err.Error() == "object does not exist" {
		d.SetId("")
		return diag.Diagnostics{
			diag.Diagnostic{
				Severity: diag.Warning,
				Summary:  "Failed to retrieve account role. Marking the resource as removed.",
				Detail:   fmt.Sprintf("Id: %s", d.Id()),
			},
		}
	}

	logging.DebugLogger.Printf("[DEBUG] About to show grants")
	grants, err := client.Grants.Show(ctx, opts)
	if err != nil {
		if errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) {
			d.SetId("")
			return diag.Diagnostics{
				diag.Diagnostic{
					Severity: diag.Warning,
					Summary:  "Failed to retrieve grants. Target object not found. Marking the resource as removed.",
					Detail:   fmt.Sprintf("Id: %s", d.Id()),
				},
			}
		}
		return diag.Diagnostics{
			diag.Diagnostic{
				Severity: diag.Error,
				Summary:  "Failed to retrieve grants",
				Detail:   fmt.Sprintf("Id: %s\nError: %s", d.Id(), err),
			},
		}
	}
```
Also, see that we brought back the error comparison in the show grants error handling to SetId("") only if the sdk.ErrObjectNotExistOrAuthorized is returned.

pkg/resources/grant_application_role_acceptance_test.go
```diff
@@ -0,0 +1,133 @@
...
+			Of: &sdk.ShowGrantsOf{
+				ApplicationRole: sdk.NewDatabaseObjectIdentifierFromFullyQualifiedName(applicationRoleName),
+			},
+		})
+		if err != nil {
+			continue
+		}
+		for _, grant := range grants {
+			if grant.GrantedTo == sdk.ObjectType(objectType) {
+				if grant.GranteeName.FullyQualifiedName() == parentRoleName {
```
sfc-gh-jcieslak, line 126: Won't this be a problem? `parentRoleName` is saved with quotes trimmed and saved to the state and now it's compared to the identifier with quotes (`FullyQualifiedName`)

sfc-gh-jcieslak (changes requested):

pkg/resources/grant_application_role.go
```diff
@@ -0,0 +1,199 @@
...
+
+var grantApplicationRoleSchema = map[string]*schema.Schema{
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      "Specifies the identifier for the application role to grant.",
+		ForceNew:         true,
+		ValidateDiagFunc: IsValidIdentifier[sdk.DatabaseObjectIdentifier](),
+	},
+	"parent_role_name": {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Thanks 👍

pkg/resources/grant_application_role.go
```diff
@@ -79,54 +86,91 @@ func GrantApplicationRole() *schema.Resource {
...
 		d.SetId("")
-		return nil
+		return diag.Diagnostics{
+			diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary:  "Failed to retrieve application role. Marking the resource as removed.",
+				Detail:   fmt.Sprintf("Id: %s", d.Id()),
+			},
+		}
 	}
```
sfc-gh-jcieslak, line 168: shouldn't we also check the err even if it's not the `ErrObjectNotExistOrAuthorized`? Right now, we discard the error, but maybe in that case we should return an error with `err` as additional context.

pkg/resources/grant_application_role.go
```diff
@@ -39,38 +40,44 @@ var grantApplicationRoleSchema = map[string]*schema.Schema{
...
+					if err := d.Set("parent_account_role_name", p); err != nil {
 						return nil, err
 					}
 				case "APPLICATION":
-					if err := d.Set("application_name", strings.Trim(parts[2], "\"")); err != nil {
+					p := parts[2]
+					if strings.HasPrefix(p, "\"\"") && strings.HasSuffix(p, "\"\"") {
+						p = strings.Trim(p, "\"")
+					}
+					if err := d.Set("application_name", p); err != nil {
```
sfc-gh-jcieslak, line 76: Not sure why now the account role is different than the application. I thought the quoted identifier would be better for being consistent with other id's, so maybe doing this would be enough: `if err := d.Set("application_name", sdk.NewAccountObjectIdentifier(parts[2]).FullyQualifiedName()); err != nil {` in both cases.

pkg/sdk/testint/application_roles_gen_integration_test.go
sfc-gh-jmichalak: Reverted.

pkg/resources/grant_application_role_acceptance_test.go
sfc-gh-jmichalak: They're created in Terraform config in https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/grant-app-role/pkg/resources/testdata/TestAcc_GrantApplicationRole/account_role/test.tf and https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/grant-app-role/pkg/resources/testdata/TestAcc_GrantApplicationRole/application/test.tf for each test case.

pkg/resources/grant_application_role.go
sfc-gh-jmichalak: Fixed.

pkg/resources/grant_application_role.go
sfc-gh-jmichalak: Trim removed.

pkg/resources/grant_application_role_acceptance_test.go
sfc-gh-jmichalak: `parentRoleName` is not trimmed now, so I think this should be OK.

pkg/resources/grant_application_role.go
sfc-gh-jmichalak: Fixed.

pkg/resources/grant_application_role.go
```diff
@@ -127,7 +121,7 @@ func ReadContextGrantApplicationRole(ctx context.Context, d *schema.ResourceData
 	switch objectType {
 	case sdk.ObjectTypeRole:
 		{
-			if _, err := client.Roles.ShowByID(ctx, sdk.NewShowByIdRoleRequest(sdk.NewAccountObjectIdentifierFromFullyQualifiedName(targetIdentifier))); err != nil && errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) {
+			if _, err := client.Roles.ShowByID(ctx, sdk.NewAccountObjectIdentifierFromFullyQualifiedName(targetIdentifier)); err != nil && errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) {
```
sfc-gh-jcieslak, line 124: the error check should be `errors.Is(err, sdk.ErrObjectNotFound)` (also in line 136; cannot comment there ;_;)

pkg/resources/testdata/TestAcc_GrantApplicationRole/application/test.tf
```diff
@@ -1,3 +1,8 @@
+provider "snowflake" {
+  profile   = "default"
+  warehouse = "\"${var.warehouse}\""
+}
+
```
sfc-gh-jcieslak, line 5: Remove the providers from tests as they are not needed

sfc-gh-jcieslak (changes requested):

pkg/resources/grant_application_role.go
sfc-gh-jmichalak: Fixed in both lines.

pkg/resources/testdata/TestAcc_GrantApplicationRole/application/test.tf
sfc-gh-jmichalak: Removed. But without this custom config, these tests are failing on my machine. Let's see what happens on CI.

sfc-gh-jmichalak: You're right, it works now. Also, I fixed application deps.

sfc-gh-jcieslak (dismissed):

pkg/resources/grant_application_role_acceptance_test.go
sfc-gh-asawicki: They weren't before, closing.

v1-preparations/REMAINING_GA_OBJECTS.MD
```diff
@@ -36,7 +36,7 @@ Known issues lists open issues touching the given object. Note that some of thes
...
+| PIPE                        |   ❌    | [#2075](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2075), [#1781](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1781), [#1707](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1707), [#1478](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1478), [#533](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/533), [#2785](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2785)                                                                                                                                                                                                                                                                  |
```
sfc-gh-asawicki, line 39: just FYI, all these issues were ordered in a desc order ;) but do not change it if there are no other comments

pkg/sdk/testint/application_roles_gen_integration_test.go
```diff
@@ -78,6 +84,63 @@ func TestInt_ApplicationRoles(t *testing.T) {
...
+		// grant the application role to the role
+		kindOfRole := sdk.NewKindOfRoleRequest().WithRoleName(sdk.Pointer(role.ID()))
+		gr := sdk.NewGrantApplicationRoleRequest(id).WithTo(*kindOfRole)
+		err := client.ApplicationRoles.Grant(ctx, gr)
+		require.NoError(t, err)
+
+		grants, err := client.Grants.Show(ctx, &sdk.ShowGrantOptions{To: &sdk.ShowGrantsTo{Role: role.ID()}})
+		require.NoError(t, err)
+		require.NotEmpty(t, grants)
+		require.NotEmpty(t, grants[0].CreatedOn)
```
sfc-gh-asawicki, line 101: these 4 lines are repeated in the next test, could we extract the assertion?

pkg/sdk/config.go
```diff
@@ -69,6 +69,9 @@ func MergeConfig(baseConfig *gosnowflake.Config, mergeConfig *gosnowflake.Config
 	if baseConfig.Host == "" {
 		baseConfig.Host = mergeConfig.Host
 	}
+	if baseConfig.Warehouse == "" {
```
sfc-gh-asawicki, line 72: it is a behavior change and as @sfc-gh-jcieslak tells me you have reverted  some changes introduced because of some local setup problems. So I guess it can be reverted too, right?

pkg/resources/testdata/TestAcc_GrantApplicationRole/account_role/variables.tf
```diff
@@ -0,0 +1,19 @@
...
+
+variable "schema_name" {
+  type = string
+}
+
+variable "application_name" {
+  type = string
+}
+
+variable "random_name" {
```
sfc-gh-asawicki, line 17: it is used for stage and application package, let's be explicit by introducing stage_name and application_package_name instead

pkg/resources/grant_application_role_acceptance_test.go
```diff
@@ -0,0 +1,142 @@
...
+				ConfigVariables:          m(),
+				ResourceName:             resourceName,
+				ImportState:              true,
+				ImportStateVerify:        true,
+			},
+		},
+	})
+}
+
+func testAccCheckGrantApplicationRoleDestroy(s *terraform.State) error {
```
sfc-gh-asawicki, line 113: we have a separate place for the checkDestroy functions now (pkg/acceptance/check_destroy.go)

pkg/resources/testdata/TestAcc_GrantApplicationRole/application/manifest.yml
```diff
@@ -0,0 +1,5 @@
+manifest_version: 1 # required
```
sfc-gh-asawicki, line 1: is separate manifest and setup.sql needed? (we already have them in tests; check comment in test.tf file)

pkg/resources/testdata/TestAcc_GrantApplicationRole/application/test.tf
```diff
@@ -0,0 +1,52 @@
...
+  schema   = var.schema_name
+}
+
+locals {
+  stage_identifier            = "\"${var.database_name}\".\"${var.schema_name}\".\"${snowflake_stage.stage.name}\""
+  file_prefix                 = abspath(path.root)
+  application_role_identifier = "\"${var.application_name}\".\"app_role_1\""
+}
+
+resource "snowflake_unsafe_execute" "put_manifest" {
```
sfc-gh-asawicki, line 13: these were done before it was easily accessible from helpers, so we can set it up without unsafe execute calls

sfc-gh-asawicki (changes requested):

pkg/sdk/testint/application_roles_gen_integration_test.go
sfc-gh-jmichalak: Done.

pkg/resources/testdata/TestAcc_GrantApplicationRole/account_role/variables.tf
sfc-gh-jmichalak: Fixed.

pkg/resources/grant_application_role_acceptance_test.go
sfc-gh-jmichalak: Moved.

pkg/resources/testdata/TestAcc_GrantApplicationRole/application/manifest.yml
sfc-gh-jmichalak: Merged into one with accountRole.

pkg/resources/testdata/TestAcc_GrantApplicationRole/application/test.tf
sfc-gh-jmichalak: Done.

pkg/resources/grant_application_role_acceptance_test.go
```diff
@@ -1,44 +1,58 @@
...
+
+	acc.TestClient().Stage.PutOnStage(t, stage.ID(), "TestAcc_GrantApplicationRole/manifest.yml")
+	acc.TestClient().Stage.PutOnStage(t, stage.ID(), "TestAcc_GrantApplicationRole/setup.sql")
+
+	applicationPackage, cleanupApplicationPackage := acc.TestClient().ApplicationPackage.CreateApplicationPackage(t)
+	t.Cleanup(cleanupApplicationPackage)
+
+	acc.TestClient().ApplicationPackage.AddApplicationPackageVersion(t, applicationPackage.ID(), stage.ID(), "v1")
+
+	application, cleanupApplication := acc.TestClient().Application.CreateApplicationWithID(t, sdk.NewAccountObjectIdentifier(name), applicationPackage.ID(), "v1")
```
sfc-gh-asawicki, line 29: why do we need a certain name here? We want to steer away from that and leave the name/id generation to helpers whenever needed. It's possible to set it up that way here (application is returned so its id can be used to set up variables for the tf config file)

pkg/resources/grant_application_role_acceptance_test.go
```diff
@@ -1,44 +1,58 @@
...
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 
 	"github.com/hashicorp/terraform-plugin-testing/config"
-	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
-	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
+func createApp(t *testing.T, name string) *sdk.Application {
```
sfc-gh-asawicki, line 15: this is a nice helper that should be reusable, please leave a comment to move it to helpers (we still need to figure out how we want to serve such complex helpers because know there are only single object methods there). Please create a subtask in SNOW-933289 with a brief description (and this discussion linked) and add the issue number to the TODO comment left here

pkg/acceptance/helpers/application_client.go
```diff
@@ -25,10 +25,14 @@ func (c *ApplicationClient) client() sdk.Applications {
 }
 
 func (c *ApplicationClient) CreateApplication(t *testing.T, packageId sdk.AccountObjectIdentifier, version string) (*sdk.Application, func()) {
+	t.Helper()
+	return c.CreateApplicationWithID(t, c.ids.RandomAccountObjectIdentifier(), packageId, version)
+}
+
+func (c *ApplicationClient) CreateApplicationWithID(t *testing.T, id sdk.AccountObjectIdentifier, packageId sdk.AccountObjectIdentifier, version string) (*sdk.Application, func()) {
```
sfc-gh-asawicki, line 32: this shouldn't be needed

sfc-gh-asawicki (changes requested):

pkg/acceptance/helpers/application_client.go
sfc-gh-jmichalak: Removed.

pkg/resources/grant_application_role_acceptance_test.go
sfc-gh-jmichalak: Got it, fixed.

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

Issue #2700, opened 2024-04-11
`snowflake_grant_ownership` fails to operate on database roles


sgrzemski:
### Terraform CLI and Provider Versions

terraform cli:
```bash
Terraform v1.6.0
on darwin_arm64
```

provider versions: `0.88.0`

### Terraform Configuration

```terraform
resource "snowflake_grant_ownership" "ownership" {
  provider = snowflake.SYSADMIN_ER

  for_each = { for key, role in snowflake_database_role.this : key => role }

  account_role_name   = data.snowflake_current_role.USERADMIN_ER.name
  outbound_privileges = "COPY"

  on {
    object_type = "DATABASE ROLE"
    object_name = "\"${each.value.database}\".\"${each.value.name}\""
  }
}
```

### Expected Behavior

Ownership of a database role is properly changed.

### Actual Behavior

The ownership of a database role is in fact changed, but then the provider fails to verify the ownership:

```bash
module.existing_raw_schemas["REDACTED"].snowflake_grant_ownership.read_ownership: Creating...
╷
│ Warning: Applied changes may be incomplete
│ 
│ The plan was created with the -target option in effect, so some changes requested in the configuration may have been ignored and the output values may not be fully
│ updated. Run the following command to verify that no other changes are pending:
│     terraform plan
│       
│ Note that the -target option is not suitable for routine use, and is provided only for exceptional situations such as recovering from errors or mistakes, or when
│ Terraform specifically suggests to use it as part of an error message.
╵
╷
│ Warning: Couldn't find OWNERSHIP privilege on the target object. Marking the resource as removed.
│ 
│   with module.existing_raw_schemas["REDACTED"].snowflake_grant_ownership.read_ownership,
│   on .terraform/modules/existing_raw_schemas/readonly_privilege_grants.tf line 1, in resource "snowflake_grant_ownership" "read_ownership":
│    1: resource "snowflake_grant_ownership" "read_ownership" {
│ 
│ Id: 
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.existing_raw_schemas["REDACTED"].snowflake_grant_ownership.read_ownership, provider
│ "provider[\"registry.terraform.io/snowflake-labs/snowflake\"].SYSADMIN_ER" produced an unexpected new value: Root object was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

[...]

PR #2703, opened 2024-04-12
fix: granting ownership on database roles

sfc-gh-jcieslak:
A fix for #2700. Changed expected granted_on when it's equal to the database role. Acceptance test added.


sfc-gh-asawicki (approved):

sfc-gh-jcieslak:
> @sfc-gh-jcieslak @sfc-gh-asawicki is there a chance we can get this released soon?

The fix is merged and the next version will be most likely released tomorrow.

PR #2710, opened 2024-04-15
chore: Initialize the SDK client in fewer places

sfc-gh-asawicki:
Result of SNOW-935947.

- the SDK default client was already setup earlier in the acceptance tests context; used in all valid places
- added secondary client setup to acceptance tests context and used in valid places

### DIFF
--- a/pkg/acceptance/testing.go
+++ b/pkg/acceptance/testing.go
@@ -62,11 +62,22 @@ func init() {
 		log.Panicf("Cannot instantiate new client, err: %v", err)
 	}
 	atc.client = client
+
+	cfg, err := sdk.ProfileConfig(testprofiles.Secondary)
+	if err != nil {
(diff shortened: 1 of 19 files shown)


pkg/acceptance/testing.go
```diff
@@ -144,6 +155,11 @@ func Client(t *testing.T) *sdk.Client {
 	return atc.client
 }
 
+func SecondaryClient(t *testing.T) *sdk.Client {
+	t.Helper()
+	return atc.client
```
sfc-gh-jcieslak, line 160: secondaryClient*

sfc-gh-jcieslak (changes requested):

pkg/acceptance/testing.go
sfc-gh-asawicki: 🤦 fixed

sfc-gh-jcieslak (approved):

PR #2743, opened 2024-04-24
fix: go casting errors + additional fixes

sfc-gh-jcieslak:
Done in this pr:
- Fixed a few Go casting errors. (actually just two in `external_oauth_integraiton`)
- Other casts like identifiers were adjusted by adding schema validation where known object type's is expected (e.g. `table_id`). Also introduced parsing function that will fail in Create operation (just in case). 
- **Question:** Other castings were not adjusted as they were casting the `d.Id()` in Update/Read/Delete operations and every single of them was correct (I was hesitating because those castings were not in Create, but can replace old parsing function the new one: `id := helpers.DecodeSnowflakeID(d.Id()).(sdk.SomeIdentifier)` would translate to `id, err := helpers.SafelyDecodeSnowflakeID[sdk.SomeIdentifier](d.Id())`. Should we do it or before identifiers rework just leave it as is ?
- When going through the resources and data sources also fixed.
  - Failover groups and unsetting `replication_schedule`.
  - Table constraint and required field being `optional` (actually no behavior change, because without the `references` block it would crash anyway, but now the error makes more sense with the `required` flag).
  - Resource monitor (followed IDE hint and added nil check as the value could be not set in the previous lines).

## Test Plan

* [x] acceptance tests added for failover groups and table constraint
* [x] unit and integration also added for failover groups and new unset option
* [x] units for new "decoding" helper function


sfc-gh-jmichalak (dismissed):

docs/resources/table_constraint.md
```diff
@@ -125,12 +125,15 @@ resource "snowflake_table_constraint" "unique" {
 <a id="nestedblock--foreign_key_properties"></a>
 ### Nested Schema for `foreign_key_properties`
 
+Required:
+
+- `references` (Block List, Min: 1, Max: 1) The table and columns that the foreign key references. Not applicable for primary/unique keys (see [below for nested schema](#nestedblock--foreign_key_properties--references))
```
sfc-gh-asawicki, line 130: can we also adjust the description then? If this is nested in foreign key properties, than it does not have to be explicitly not applicable for primary/unique

pkg/resources/failover_group_acceptance_test.go
```diff
@@ -181,6 +181,18 @@ func TestAcc_FailoverGroupInterval(t *testing.T) {
 					resource.TestCheckResourceAttr("snowflake_failover_group.fg", "replication_schedule.0.interval", "10"),
 				),
 			},
+			// Remove replication schedule
```
sfc-gh-asawicki, line 184: can we not make it the last step before the import? (separate test or different ordering) We are now testing import without replication schedule

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

docs/resources/table_constraint.md
sfc-gh-jcieslak: 👍 adjusted

pkg/resources/failover_group_acceptance_test.go
sfc-gh-jcieslak: 👍 Reordered

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (commented):
Had to remove schema validators, because of the way we are now encoding/decoding identifiers. Added a note for myself to add it as identifiers rework item.

MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,28 @@ This document is meant to help you migrate your Terraform config to the new newe
 describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behavior
 across different versions.
 
+## v0.89.0 ➞ v0.90.0
+### tag_masking_policy_association resource changes
```
sfc-gh-asawicki, line 8: You have conflicts here

MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,28 @@ This document is meant to help you migrate your Terraform config to the new newe
...
+Now the `tag_masking_policy_association` resource will only accept fully qualified names separated by dot `.` instead of pipe `|`.
```
sfc-gh-asawicki, line 9: I think we should add a state upgrader to make it easy for the users (this is not much work and can be done in a separate PR - but before the release). WDYT?

sfc-gh-asawicki (approved):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Resolved

MIGRATION_GUIDE.md
sfc-gh-jcieslak: True, will work on it and push the next pr

sfc-gh-asawicki (approved):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: I think I'll leave this as it is, would have to write a few state upgrades and make changes to how IDs are saved/retrieved also in a few places, which makes me feel something could go wrong there. Also, forcing the migration to a more complex ID, but also a more correct one would most likely result in no migration after identifiers rework (because it's already a valid identifier. If we add some helpers field like `qualified_name` it will depend on the user when they would like to upgrade).

Issue #2482, opened 2024-02-09
Add support for network rules


Relativity74205:
### Terraform CLI and Provider Versions

Terraform v1.5.7
Providor v0.85.0

### Use Cases or Problem Statement

The old way of assigning IPs directly to a network policy is deprecated by Snowflake. The new way is to define network rules (see e.g. here https://docs.snowflake.com/en/sql-reference/sql/create-network-rule) and to assign them to network policies (see https://docs.snowflake.com/en/sql-reference/sql/alter-network-policy).

### Proposal

Add a snowflake_network_rule resource and modify snowflake_network_policy resource.

### How much impact is this issue causing?

Medium

sfc-gh-asawicki:
Hey @Relativity74205. Thanks for reaching out to us.

We have adding the missing GA features on our roadmap: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#our-roadmap. We'll make sure to take care of it then.

Relativity74205:
@sfc-gh-asawicki I would like to create a PR, if you accept contributions at this stage of your roadmap. I have already made some contributions in the past.

sfc-gh-asawicki:
@Relativity74205, we are always open to contributions. Some topics need to be addressed first:
- We do not have up-to-date contribution guidelines, PR requirements, etc. We want to update them and plan to do it in the next couple of weeks.
- We have recently moved to the SDK implementation. We have lots of conventions there that are not documented. Furthermore, we have written a generator to help us with the implementation. It is not that intuitive, though. This means we would prefer adding the SDK implementation ourselves - it will be much better from the timewise perspective. This is the first part needed to introduce the new resource.
- The second part is to add the resource itself, together with the acceptance tests. We design the new resources before implementation to know which direction we are heading. We have yet to have such a design, but it may not be that big of an issue.

From our side, the best course of action would be:
1. We will implement the SDK part in the upcoming weeks.
2. You can implement the resource part using the prepared SDK. We will go ahead and summarize the guidelines beforehand.
3. We can have the design discussions on the PR review level and adjust the resource later.

How does it sound?

Relativity74205:
Hi @sfc-gh-asawicki,
sure, that sounds good. Do you have any issues I can track so I will get notified automatically, when the SDK is finished?

sfc-gh-asawicki:
There is no issue on GH for this yet. I will create one, and I will let you know its number.

[...]

PR #2746, opened 2024-04-25
feat: add network rules

Relativity74205:
resolves https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2482

- adds network rules
- adds network rule lists to network policies
- fixes network policies

## Test Plan

* [x] acceptance tests

* [ ] …

## References

*


pkg/resources/network_policy.go
```diff
@@ -20,19 +22,31 @@ var networkPolicySchema = map[string]*schema.Schema{
...
+		Elem:        &schema.Schema{Type: schema.TypeString},
+		Optional:    true,
+		Description: "Specifies a list of fully qualified network rules that contain the network identifiers that are allowed access to Snowflake.",
+	},
+	"blocked_network_rule_list": {
+		Type:        schema.TypeSet,
+		Elem:        &schema.Schema{Type: schema.TypeString},
+		Optional:    true,
+		Description: "Specifies a list of fully qualified network rules that contain the network identifiers that are denied access to Snowflake.",
+	},
```
Relativity74205, line 36: I am not sure, what the best way to solve a problem here. Network rules are schema level objects, thus it is necessary to pass a list of fully qualified names as parameters. The problem is the validation. There is a IsValidIdentifier function, which does such a check, however, it is only defined for collections. In addition, the entire framework does not to use ValidateDiagFunc functions for collections (not implemented).

We could leave this at it is, however, the provider crashes with an error, which is not easily understood, when someone passes e.g. "foo" as a network rule name. Therefore, there should be a form of validation.

I am thinking about adding a custom CustomizeDiff function into the schema.Resource struct, which does the validation. For this I would extract the main validation logic from the IsValidIdentifier function into a function, which can be used outside the ValidateDiagFunc  mechanism and use it to check very entry of the network rule list.

Does this work for you or would you prefer another solution?

sfc-gh-jcieslak: Hey 👋 
I think the proposed CustomizeDiff would work fine, but please add an acceptance test showing it throws an expected error whenever the incorrect identifier is passed.

sfc-gh-jmichalak: Hi, thanks for you contribution :) What do you mean by `There is a IsValidIdentifier function, which does such a check, however, it is only defined for collections`? I can see that it supports singular ID types. I think you can use `parseNetworkRulesList` which essentially converts `schema.TypeSet` -> `[]sdk.SchemaObjectIdentifier` an call `IsValidIdentifier` on each element in the returned slice, right? Correct me if I'm missing something. cc @sfc-gh-jcieslak

Relativity74205: @sfc-gh-jmichalak 
Regarding `There is a IsValidIdentifier function, which does such a check, however, it is only defined for collections`: There was an error on my side, I actually meant `There is a IsValidIdentifier function, which does such a check, however, it is only defined for single elements and not for collections`. Sorry for the confusion made.

@sfc-gh-jcieslak 
You're suggestion worked perfectly! Then I think I am finally finished with the actual implementation. I will let you know when I have added the tests and finalized the docs.

Relativity74205: @sfc-gh-jcieslak @sfc-gh-jmichalak 
I have added tests and docs, I hope that I haven't overlooked something. A review of the PR would be greatly appreciated :)

pkg/resources/network_policy_acceptance_test.go
```diff
@@ -92,12 +142,89 @@ resource "snowflake_network_policy" "test" {
...
+
+resource "snowflake_network_policy" "test" {
+	name            = "%v"
+	comment         = "%v"
+	allowed_ip_list = ["192.168.0.100/24"]
+	blocked_ip_list = ["192.168.0.101"]
+	allowed_network_rule_list = [snowflake_network_rule.test1.qualified_name]
+	blocked_network_rule_list = [snowflake_network_rule.test2.qualified_name]
+}
+`, nameRule1, database, schema, networkPolicyComment, nameRule2, database, schema, networkPolicyComment, name, networkPolicyComment)
```
sfc-gh-jmichalak, line 206: You can use https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes to simplify formatting.

pkg/resources/network_policy.go
```diff
@@ -109,111 +138,260 @@ func ReadNetworkPolicy(d *schema.ResourceData, meta interface{}) error {
...
 			if err != nil {
-				return fmt.Errorf("error updating comment for network policy %v err = %w", name, err)
+				return getUpdateContextDiag("updating comment", name, err)
+			}
+		}
+	}
+
+	if d.HasChange("allowed_network_rule_list") {
+		oldList, newList := d.GetChange("allowed_network_rule_list")
+		addedNetworkRuleIdentifiers, removedNetworkRuleIdentifiers := getAddedAndRemovedIdentifiers(oldList, newList)
```
sfc-gh-jmichalak, line 223: Why not just use SET with newList?

sfc-gh-jmichalak (changes requested):
Overall looks good :) Please also add `import.sh` and `resource.tf` to `examples/resources/snowflake_network_role` and regenerate docs

sfc-gh-jcieslak (commented):
Hey 👋 I had a few small comments, but other than that very good pr 👍

pkg/resources/network_policy_acceptance_test.go
Relativity74205: Didn't know this. Thanks. Fixed.

pkg/resources/network_policy.go
Relativity74205: Because I have overlooked this option and concentrated on add/remove network rule from list. Thanks for the hint. Fixed.

pkg/resources/network_rule.go
```diff
@@ -0,0 +1,219 @@
...
+func ReadContextNetworkRule(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	diags := diag.Diagnostics{}
+	client := meta.(*provider.Context).Client
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.SchemaObjectIdentifier)
+
+	networkRule, err := client.NetworkRules.ShowByID(ctx, id)
+	if networkRule == nil || err != nil {
+		// If not found, mark resource to be removed from state file during apply or refresh
+		log.Printf("[DEBUG] network rule (%s) not found", d.Id())
+		d.SetId("")
```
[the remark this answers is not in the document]
Relativity74205: Done

examples/resources/snowflake_network_rule/import.sh
```diff
@@ -0,0 +1 @@
+terraform import snowflake_network_rule.example rulename
```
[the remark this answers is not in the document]
Relativity74205: Fixed for `network_rules`. However, what is the error with `network_policies`? Policies are accout level objects in contrast to network rules, which are schema level objects.

pkg/resources/network_rule.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return ReadContextNetworkRule(ctx, d, meta)
+}
+
+func DeleteContextNetworkRule(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	name := d.Id()
+	client := meta.(*provider.Context).Client
+	id := helpers.DecodeSnowflakeID(name).(sdk.SchemaObjectIdentifier)
+
+	if err := client.NetworkRules.Drop(ctx, sdk.NewDropNetworkRuleRequest(id)); err != nil {
```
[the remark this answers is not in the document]
Relativity74205: Done.

pkg/resources/network_policy_acceptance_test.go
```diff
@@ -33,6 +36,8 @@ func TestAcc_NetworkPolicy(t *testing.T) {
 					resource.TestCheckResourceAttr("snowflake_network_policy.test", "comment", networkPolicyComment),
 					resource.TestCheckResourceAttr("snowflake_network_policy.test", "allowed_ip_list.#", "2"),
 					resource.TestCheckResourceAttr("snowflake_network_policy.test", "blocked_ip_list.#", "0"),
+					resource.TestCheckResourceAttr("snowflake_network_policy.test", "allowed_network_rule_list.#", "0"),
+					resource.TestCheckResourceAttr("snowflake_network_policy.test", "blocked_network_rule_list.#", "0"),
```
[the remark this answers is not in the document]
Relativity74205: I think that this tests are already doing it (line 66-82). Or have I understood something wrong?

```
			{
				Config: networkPolicyConfigIPsAndRules(name, nameRule1, nameRule2, acc.TestDatabaseName, acc.TestSchemaName),
					resource.TestCheckResourceAttr("snowflake_network_policy.test", "name", name),
					resource.TestCheckResourceAttr("snowflake_network_policy.test", "comment", networkPolicyComment),
					resource.TestCheckResourceAttr("snowflake_network_policy.test", "allowed_ip_list.#", "1"),
					resource.TestCheckResourceAttr("snowflake_network_policy.test", "blocked_ip_list.#", "1"),
					resource.TestCheckResourceAttr("snowflake_network_policy.test", "allowed_network_rule_list.#", "1"),
					resource.TestCheckResourceAttr("snowflake_network_policy.test", "blocked_network_rule_list.#", "1"),
				),
			},
			// IMPORT - all fields are non-empty
			{
				ResourceName:      "snowflake_network_policy.test",
				ImportState:       true,
				ImportStateVerify: true,
			},
```

docs/resources/network_policy.md
```diff
@@ -12,12 +12,21 @@ description: |-
...
+  value_list = ["192.168.0.100/24", "29.254.123.20"]
+}
+
 resource "snowflake_network_policy" "policy" {
   name    = "policy"
   comment = "A policy."
 
-  allowed_ip_list = ["192.168.0.100/24"]
-  blocked_ip_list = ["192.168.0.101"]
+  allowed_network_rule_list = [snowflake_network_rule.rule.qualified_name]
```
sfc-gh-asawicki, line 29: It would be nice to have both examples in place (the previous one with allowed/disallowed IPs and the new one with the network rule list).

pkg/resources/network_policy.go
```diff
@@ -2,8 +2,10 @@ package resources
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
-	"log"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
```
sfc-gh-asawicki, line 8: when we run the linter it should be marked as nongofumpted (please run make pre-push).

sfc-gh-asawicki (approved):

examples/resources/snowflake_network_rule/import.sh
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yeah, for some reason I think I thought it was a schema object 🤔 .

pkg/resources/network_policy_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: I meant that maybe we should have a test that would check the create operation with all the optional fields set. On 66-68, I believe this will be an update operation (using alters) which is a different part of the code. But since we'll be touching this codebase I think we can leave it to ourselves.

pkg/resources/network_rule.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Thank You, the only thing I would change is the error type (that's on me, because the example said `sdk.ErrObjectNotExistOrAuthorized`), because in this case we should compare against `sdk.ErrObjectNotFound` instead.

sfc-gh-jcieslak (approved):

docs/resources/network_policy.md
Relativity74205: Sure, done.

pkg/resources/network_policy.go
Relativity74205: Sorry, forgot about this. done.

pkg/resources/network_rule.go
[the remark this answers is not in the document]
Relativity74205: Sure, done.

However, I have questions: What is the exact difference between the two error types and most of all, how to recognize which to use?

Relativity74205:
@sfc-gh-asawicki @sfc-gh-jcieslak @sfc-gh-jmichalak 
I think I have covered all yours comments. Is there anything else? If not, I would appreciate it, if we could finally merge this one :)

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

pkg/resources/network_rule.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: `ErrObjectNotExistOrAuthorized` is a Snowflake error, so this one could be returned by any operation, and `ErrObjectNotFound` is an SDK error returned from ShowByID operations. It indicates that we called SHOW and filtered its output and didn't find the expected object (you can see ShowByID implementation and that it returns an error inside collections.FindOne function).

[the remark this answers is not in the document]
Relativity74205: I understand. Thanks for the explanation :)

sfc-gh-asawicki:
Merged! Thanks for your contribution @Relativity74205!

PR #2747, opened 2024-04-25
chore: Replace id generation in integration and acceptance tests part1

sfc-gh-asawicki:
In order to move to the unified ID generation (connected with the changes in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2737) to stabilize parallel builds, the following changes were introduced inside this PR:
- IdsGenerator was introduced in TestClient. It allows us to get the ids of the default objects and build the new identifiers with the desired suffix (indicating the current build). Most of the helpers were moved to use this generator to get old and new ids. There are still TODOs left there (there are helpers that get the name of the object from the caller - we have to ensure that the names are generated the same way, or they have a good reason to be different). Ids generator is exposed so that both integration and acceptance tests can use it.
- Changed the generation method of resource/datasource names to the exposed method described below in all the acceptance tests.
- Fixed the random.go file in the SDK (so that this one is used only in the SDK unit tests - read the content of the file below, it had the long comment describing the problems with this file):
  - renamed to random_test.go
  - unexported all the methods
  - switch to appropriate methods in integration/acceptance tests
- Next part of the change is to check all identifier creations throughout helpers/acceptance tests/integration tests and switch to a proper generator. This is not finished. The following methods were attacked (the rest will be tackled in the next PR):
  - NewExternalObjectIdentifier
  - NewExternalObjectIdentifierFromFullyQualifiedName
  - NewAccountIdentifier
  - NewAccountIdentifierFromAccountLocator (moved to ids generator)
  - NewAccountIdentifierFromFullyQualifiedName (2 usages left in business critical skipped tests)
  - NewAccountObjectIdentifier (testAccCheckDatabaseExistence left with todo, grantOwnershipToTheCurrentRole, testAccCheckSharePrivilegesRevoked, sdk.NewAccountObjectIdentifier("ACCOUNTADMIN"), sdk.NewAccountObjectIdentifier("ORGADMIN"), sdk.NewAccountObjectIdentifier("NOT_EXISTING_ACCOUNT"), sdk.NewAccountObjectIdentifier("does_not_exist"), current role/user, sdk.NewAccountObjectIdentifier("<DB>"))
  - NewAccountObjectIdentifierFromFullyQualifiedName
- other smaller changes:
  - removed warehouse 2 in acceptance tests for now
  - fixed `TestAcc_TaskOwnershipGrant_onFuture` test

Left to be done regarding NewAccountObjectIdentifier:
- `alterMaterializedViewQueryExternally`
- `checkDatabaseAndSchemaDataRetentionTime`
- `checkDatabaseSchemaAndTableDataRetentionTime`
- `unsafe_execute_acceptance_test.go`
- `alterViewOwnershipExternally`
- check `app_role_1`
- check hardcoded `filter_by_role`, `stproc1`, `sp_pi`, `filterByRole`, `records`
- `CreateDatabaseWithName` (and other similar helpers)
- check `setup_test.go`
- ask @sfc-gh-jcieslak about `sdk.NewAccountObjectIdentifier("S3_STORAGE_INTEGRATION")`, `sdk.NewAccountObjectIdentifier("GCP_STORAGE_INTEGRATION")`, `sdk.NewAccountObjectIdentifier("AZURE_STORAGE_INTEGRATION")`
- extract common helpers for `sdk.NewAccountObjectIdentifier("does_not_exist"` and similar


pkg/sdk/testint/schemas_integration_test.go
```diff
@@ -81,16 +81,16 @@ func TestInt_SchemasCreate(t *testing.T) {
 	})
 
 	t.Run("with tags", func(t *testing.T) {
-		tagName := random.String()
-		tagID := sdk.NewAccountObjectIdentifier(tagName)
+		tagID := testClientHelper().Ids.RandomAccountObjectIdentifier()
+		tagName := tagID.Name()
 		_, err := client.ExecForTests(ctx, fmt.Sprintf(`CREATE TAG "%s"`, tagName))
```
sfc-gh-jcieslak, line 86: It's interesting that it works, because as far as I remember tags are objects create on schema. I guess, some database + schema are taken from session context. Maybe we should mark it as something to do in tags refactoring ?

sfc-gh-jcieslak (approved):

pkg/sdk/testint/schemas_integration_test.go
sfc-gh-asawicki: It works, because our test clients have db and schema context (and we are reusing db and schema in the majority of tests). I will use the create tag from helper, because that one uses the fully qualified name, so we will be safer in case we change the client context approach (in one of the next PRs, we still have to target all ExecForTests invocations and replace the with helper methods, but it was supposed to be don iteratively)

sfc-gh-jcieslak (approved):

PR #2748, opened 2024-04-25
chore: Adjust issue templates

sfc-gh-jcieslak:
## Questions
- Q: Are existing fields fine? - is it too similar to the bug one?
- Q: Is field validation correct? - although some forms are the same as in the bug one, they lack validation, because it depends on the type of issue as they can be *general*.
- Q: Do we add deprecated object types or stick to the non-deprecated ones? - right now I only specified non-deprecated ones

## Changes
- **New issue template for general usage questions**
- **Adjusted templates**
  - Bug
  - Docs
  - Feature
- **Added top-level `config.yml` to show additional links and disable the possibility of creating blank issues.**

## Possible improvements
- Right now, it's not possible to add conditional labeling based on e.g. chosen category from the dropdown. Afaik there's already an existing solution based on gh actions: https://github.com/marketplace/actions/advanced-issue-labeler.

## TODO
- Add missing category and object labels in the repository
- Link FAQ in a few places once it will be available


.github/ISSUE_TEMPLATE/01-bug.yml
```diff
@@ -0,0 +1,266 @@
...
+      required: true
+
+  - id: object_type
+    type: dropdown
+    attributes:
+      label: Object type(s)
+      description: Choose object type(s) causing the issue (if any). Specify object types and add **multiple** if you suspect more objects are affected. Use **all** with caution!
+      multiple: true
+      options:
+        - all
```
sfc-gh-asawicki, line 97: I have a few remarks/thoughts/questions here:
1. Can we make this field required conditionally for resource/datasource category? (I dunno if this is possible in GH)
2. Let's get rid of all and multiple. There are multiple reasons. First of all, people will abuse it. Secondly, it's better to have at least one concrete resource (and we may ask to pitck one resource in cases that more or all are affected and state the rest in the additional comment section or just in description).
3. Can we somehow reuse the list of objects between the templates? It would be better to change this list just in one place (or even generate it later based on the code 😮), than to keep it in multiple places.
4.

sfc-gh-jcieslak: 1. I'm pretty sure that's not possible, it can be either `ture` or `false` (templates are pretty "static"). Do we want to stick with optional or required then ?
2. True, I'll remove all and multiple.
3. Well, we could generate templates from Go templates based on our current schema, but I'm not sure If we want to go with this now. Other than that I don't think there's another easy option.

sfc-gh-asawicki: 1. oky
2. oky
3. oky, I will put this as a fun exercise for one of the Fridays

.github/ISSUE_TEMPLATE/01-bug.yml
```diff
@@ -0,0 +1,264 @@
...
+        - category:grants
+        - category:other
+    validations:
+      required: true
+
+  - id: object_type
+    type: dropdown
+    attributes:
+      label: Object type(s)
+      description: Choose object type(s) causing the issue (if any). Specify object types and add **multiple** if you suspect more objects are affected. Use **all** with caution!
```
sfc-gh-asawicki, line 94: Please adjust the descriptions after removing all and multiple :)

sfc-gh-asawicki (changes requested):

.github/ISSUE_TEMPLATE/01-bug.yml
sfc-gh-jcieslak: 🤦 adjusted

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #2753, opened 2024-04-29
chore: Update the contribution guidelines (and small fixes)

sfc-gh-asawicki:
The main part of the change is refreshing the outdated contribution guidelines. Note, that part on advanced debugging was just improved with the building instructions and will be validated separately (because it was updated not that long ago). Keep in mind, that this can be an iterative process, we can still improve them but we should have a good, refreshed based already.

Additionally, the TODO comments were updated and enriched with the proper ticket references.


CONTRIBUTING.md
```diff
@@ -1,73 +1,143 @@
...
-role='ACCOUNTADMIN'
+account = "<your account in form of organisation-account_name>"
+user = "<your user>"
+password = "<your password>"
+role = "<your role>"
+host="<host of your account, e.g. organisation-account_name.snowflakecomputing.com>"
 ```
 
-**Note: PRs for new resources will not be accepted without passing acceptance tests.**
+To be able to run all the tests you additionally need two more profiles `[secondary_test_account]` and `[incorrect_test_profile]`:
```
sfc-gh-jcieslak, line 57: Maybe it would be worth mentioning that some of the tests require the privileged role (preferably the`ACCOUNTADMIN`) and certain Snowflake Edition (preferably the highest available).

CONTRIBUTING.md
```diff
@@ -1,73 +1,143 @@
...
+```
+
+We are aware that not everyone has access two multiple accounts, so the majority of tests can be run using just one account. The tests setup however, requires both profiles (`default` and `secondary_test_account`) to be present. You can use the same details for `secondary_test_account` as in the `default` one, if you don't plan to run tests requiring multiple accounts.
+
+To run the tests we have three different commands:
+- `make test` run unit and integration tests
+- `make test-acceptance` run acceptance tests
+- `make test-integration` run integration tests
+
+You can run the particular tests form inside your chosen IDE but remember that you have to set `TF_ACC=1` environment variable to run any acceptance tests (the above commands set it for you).
```
sfc-gh-jcieslak, line 82: also having the `TF_LOG=DEBUG` set up helps with making sure that the right SQLs are being run.

CONTRIBUTING.md
sfc-gh-asawicki: Good point, I will add it.

CONTRIBUTING.md
sfc-gh-asawicki: Good point, I will add this too.

README.md
```diff
@@ -96,4 +97,10 @@ By default, the underlying driver is set to error level logging. It can be chang
...
 ## Contributing
 
-Cf. [Contributing](./CONTRIBUTING.md).
+Check [Contributing](./CONTRIBUTING.md).
+
+## Releases
+
+Releases will be performed as needed, typically once every 2 weeks.
+
+Releases are published to [the terraform registry](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest). Each change has its own release notes (e.g. https://github.com/Snowflake-Labs/terraform-provider-snowflake/releases/tag/v0.89.0) and migration guide if needed (e.g. https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v0880--v0890).
```
sfc-gh-jmichalak, line 106: Links shouldn't point to specific versions which will be quickly obsolete. E.g. https://github.com/Snowflake-Labs/terraform-provider-snowflake/releases, https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md

sfc-gh-jmichalak (approved):

sfc-gh-jcieslak (approved):

README.md
sfc-gh-asawicki: Why not? This are just examples, and we won't remove either of them.

PR #2759, opened 2024-04-30
fix: UNSET and empty SET in network policies

sfc-gh-jmichalak:
Removed UnsetComment field. This is now nested as Unset.Comment with unsetting other fields.

## Test Plan

* [x] acceptance tests

* [x] integration tests
* [x] unit tests
## References

https://docs.snowflake.com/en/sql-reference/sql/alter-network-policy -- UNSET should be documented.


pkg/resources/network_policy.go
```diff
@@ -181,7 +182,7 @@ func UpdateNetworkPolicy(d *schema.ResourceData, meta interface{}) error {
 		for i, v := range newIps {
 			ipRequests[i] = *sdk.NewIPRequest(v)
 		}
-		setReq := sdk.NewNetworkPolicySetRequest().WithBlockedIpList(ipRequests)
```
sfc-gh-asawicki, line 184: Please add a test in `network_policy_acceptance_test.go` to check the removal.

sfc-gh-jmichalak: Added.

pkg/resources/network_policy_acceptance_test.go
```diff
@@ -45,6 +45,16 @@ func TestAcc_NetworkPolicy(t *testing.T) {
 					resource.TestCheckResourceAttr("snowflake_network_policy.test", "blocked_ip_list.#", "1"),
 				),
 			},
+			// CHANGE PROPERTIES - empty ip lists
```
sfc-gh-asawicki, line 48: I have written a similar comment to one of the recent @sfc-gh-jcieslak changes: https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2743#discussion_r1582723559. Making the lists empty in this step will make the import step test incomplete/different. Ideally we should test all following situations:
- importing with "things" in every list and field
- importing a smaller subset
- creating object with lists filled
- updating object - setting lists to empty
- updating object - adding/removing to/from lists

So I am not sure if there is a nice ordering so that we can test it in just one test in different steps but pleasenot remove before impot :)

sfc-gh-jmichalak: Fixed, I added import check after setting non-empty lists.

sfc-gh-asawicki (approved):

PR #2762, opened 2024-04-30
chore: Add scripts to close issues from Pre Snowflake bucket

sfc-gh-asawicki:
Add scripts to close issues from Pre Snowflake bucket (for future reference; they were already run)


pkg/scripts/issues/README.md
```diff
@@ -9,3 +10,20 @@
...
+# Closing old issues (regarding https://github.com/Snowflake-Labs/terraform-provider-snowflake/discussions/2755)
+1. To use the script, generate access token here: https://github.com/settings/tokens?type=beta.
+2. First get all open issues by invoking:
+```shell
+  cd gh && SF_TF_SCRIPT_GH_ACCESS_TOKEN=<YOUR_PERSONAL_ACCESS_TOKEN> go run .
+```
+3. File `issues.json` should be generated in the `gh` directory. This is the input file for the second script.
+4. To filter only closeable issues invoke [this script](./filter-closeable-old-issues/main.go):
+```shell
+  cd file && go run .
```
sfc-gh-jcieslak, line 23: I think that should be `cd filter-closeable-old-issues` + running this script has one additional step before running which is preparing pre-snowflake-issues.csv

sfc-gh-asawicki: It was done once and they are committed to the repository (I will add the information about this, because additional step is not necessary but it will be more clear).

sfc-gh-jcieslak (approved):

Issue #2704, opened 2024-04-12
While upgrading terraform-snowflake provider from 0.80.0 to 0.87.0, the snowflake function grant fails


GomathiMa:
### Terraform CLI and Provider Versions

Terraform Version: 1.5.4 and snowflake provider version : 0.87.0

### Terraform Configuration

```terraform
resource "snowflake_grant_privileges_to_account_role" “xx_grant” {
  privileges        = ["USAGE"]
  account_role_name = “xx_role”
  on_schema_object {
    object_type = "FUNCTION"
    object_name = "\"${snowflake_database.xx_db.name}\”.\”XX_SCHEMA\”.\”XX_FUNCTION\”(OBJECT)"
  }
  depends_on = [xx_function]
}
```

### Expected Behavior

Grants to be successful

### Actual Behavior

│ Error: Unable to parse the identifier

│ 

│ Unable to parse the identifier:

│ "XX_DB”.”XX_SCHEMA.”XX”_FUNCTION(OBJECT). Make sure

│ you are using the correct form of the fully qualified name for this field:

│ <database_name>.<schema_name>.<name>.

│ Original Error: unable to read identifier:

│ "XX_DB”.”XX”_SCHEMA.”XX”_FUNCTION(OBJECT), err =

│ parse error on line 1, column 56: extraneous or missing " in quoted-field

╵

### Steps to Reproduce

1. `terraform apply`

### How much impact is this issue causing?

High

### Additional Information

Request to provide an example of how the snowflake function grant script should be while using the provider 0.87.0

sfc-gh-jcieslak:
Hey,
This is related to the way identifiers for functions and procedures are represented. For now, it's possible to grant on function, but it's not possible to quote the function name (otherwise the internal identifier parser fails). See the below example:
```terraform
resource "snowflake_function" "test" {
  name            = "test_function_name"
  schema          = snowflake_schema.test.name
  database        = snowflake_database.test.name
  language        = "python"
  runtime_version = "3.8"
  return_type     = "NUMBER(38,0)"
  statement       = "def add_two(i): return i + 2"
  handler         = "add_two"
  arguments {
    name = "i"
    type = "number"
  }
}

resource "snowflake_grant_privileges_to_account_role" "test" {
  account_role_name = "TEST_ROLE"
  privileges = [ "USAGE" ]
  on_schema_object {
    object_type = "FUNCTION"
[...]

PR #2768, opened 2024-05-02
fix: support procedure grant_ownership

YAhiru:
I'm not good at English, so please feel free to point out any code comments or function names that are unnatural :pray:.

- Fixed problem with grant_ownership resource not working for procedures.
  - This was caused by a lack of support for the fully qalified name signature part.
- Fixed incorrect behavior of FullyQualifiedName method of SchemaObjectIdentifier for signatures with no arguments
  - before: `db.schema.some_function()` -> `"db"."schema"."some_function"`
  - after: `db.schema.some_function()` -> `"db"."schema"."some_function"()`
- DecodeSnowflakeParameterID now also works with procedure and UDF identifiers
  - However, it still does not support strings with return values or named arguments
    - ex: `db.schema."func(ARG1 VARCHAR):VARCHAR(16777216)"`

## Test Plan

* [x] acceptance tests

* [ ] …

## References

*

### DIFF
--- a/pkg/resources/grant_ownership_acceptance_test.go
+++ b/pkg/resources/grant_ownership_acceptance_test.go
@@ -317,6 +317,146 @@ func TestAcc_GrantOwnership_OnObject_Table_ToDatabaseRole(t *testing.T) {
 	})
 }
 
+func TestAcc_GrantOwnership_OnObject_Procedure_ToAccountRole(t *testing.T) {
+	databaseId := acc.TestClient().Ids.RandomAccountObjectIdentifier()
+	databaseName := databaseId.Name()
+	schemaName := acc.TestClient().Ids.Alpha()
+	withArgProcedureName := sdk.NewSchemaObjectIdentifierWithArguments(databaseName, schemaName, acc.TestClient().Ids.Alpha(), []sdk.DataType{sdk.DataTypeVARCHAR})
+	withoutArgProcedureName := sdk.NewSchemaObjectIdentifierWithArguments(databaseName, schemaName, acc.TestClient().Ids.Alpha(), []sdk.DataType{})
+
+	accountRoleId := acc.TestClient().Ids.RandomAccountObjectIdentifier()
+	accountRoleName := accountRoleId.Name()
+	accountRoleFullyQualifiedName := accountRoleId.FullyQualifiedName()
+
+	configVariables := config.Variables{
+		"account_role_name":          config.StringVariable(accountRoleName),
+		"database_name":              config.StringVariable(databaseName),
+		"schema_name":                config.StringVariable(schemaName),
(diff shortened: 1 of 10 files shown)


pkg/resources/grant_ownership_acceptance_test.go
```diff
@@ -317,6 +317,146 @@ func TestAcc_GrantOwnership_OnObject_Table_ToDatabaseRole(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(withArgResourceName, "account_role_name", accountRoleName),
+					resource.TestCheckResourceAttr(withArgResourceName, "on.0.object_type", "PROCEDURE"),
+					resource.TestCheckResourceAttr(withArgResourceName, "on.0.object_name", withArgProcedureName.FullyQualifiedName()),
+					resource.TestCheckResourceAttr(withArgResourceName, "id", fmt.Sprintf("ToAccountRole|%s||OnObject|PROCEDURE|%s", accountRoleFullyQualifiedName, withArgProcedureName.FullyQualifiedName())),
+					checkResourceOwnershipIsGranted(&sdk.ShowGrantOptions{
+						To: &sdk.ShowGrantsTo{
+							Role: accountRoleId,
+						},
+					}, sdk.ObjectTypeProcedure, accountRoleName, fmt.Sprintf("%s.%s.\"%s(ARG1 VARCHAR):VARCHAR(16777216)", databaseName, schemaName, withArgProcedureName.Name())),
```
YAhiru, line 360: The last character should end with "\", but it is trimmed somewhere, so it is written this way.
It would be better to modify it so that it can be asserted with the correct value, but I did not modify it because I did not want to expand the scope as much as possible.

sfc-gh-asawicki:
Hey @YAhiru. Thanks for creating a PR!

We will check it at the start of the following week. We also have to check if this does not interfere with our upcoming https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#identifiers-rework. We'll stay in touch.

sfc-gh-jcieslak:
Hey @YAhiru 👋 ,
Thank you for creating this PR, but unfortunately, this sort of collides with our [identifier rework](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#identifiers-rework) planned for this quarter. I'm afraid that after merging this PR, we would still have to make additional changes after the rework making everyone migrate twice. That being said, It would be better to close this PR as this problem should be addressed during this quarter. For now, please try [this workaround](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2704). I highly suggest creating a GitHub issue before creating a PR to discuss the changes with the team to limit the wasted work 😢, but again Thank you for your time 🙏.

YAhiru:
okay!
Thank you for your kindness :)

PR #2816, opened 2024-05-17
chore: accept non-pointer values in the generated builder methods

sfc-gh-jcieslak:
Accept non-pointer values in the generated builder methods. Now the nested requests have to be explicitly unreferenced, but two additional cases could be added to resolve this "issue". There's only a special case for strings which shouldn't be set whenever they're optional and the input value is equal to an empty string, because: 
1. that would be a rare case to have an empty string in Snowflake.
2. empty strings are treated as non-set values in Terraform.

The improved generator was tested and used as an example on the external table.

### DIFF
--- a/pkg/sdk/dto-builder-generator/main.go
+++ b/pkg/sdk/dto-builder-generator/main.go
@@ -207,11 +207,6 @@ func (gen *Generator) generateBuilderMethods(d *structDef) {
 		gen.printf("func (s *%s) With%s(%s %s) *%s {\n", d.name, toTitle(field.name), field.name, strings.TrimLeft(field.typeString, "*"), d.name)
 
 		switch {
-		case field.typeString == "*string":
-			// When the expected type is optional string (*string), we want to avoid assigning empty strings, because:
-			// 1. It's a rare (or non-existent) case to set something to an empty string in Snowflake (most of the cases are resolved by explicit UNSET commands).
-			// 2. Empty strings are treated as non-set values in Terraform, so no value set and empty string would have the same effect.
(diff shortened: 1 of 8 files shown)


pkg/sdk/dto-builder-generator/main.go
```diff
@@ -204,8 +204,20 @@ func (gen *Generator) generateBuilderMethods(d *structDef) {
 	}
 
 	for _, field := range optionalFields {
-		gen.printf("func (s *%s) With%s(%s %s) *%s {\n", d.name, toTitle(field.name), field.name, field.typeString, d.name)
-		gen.printf("s.%s = %s\n", field.name, field.name)
+		gen.printf("func (s *%s) With%s(%s %s) *%s {\n", d.name, toTitle(field.name), field.name, strings.TrimLeft(field.typeString, "*"), d.name)
+
+		switch {
+		case field.typeString == "*string":
```
sfc-gh-asawicki, line 210: I am not sure about this logic. It is a bit misleading for the caller. Also, usually we invoke it after doing the tf check if the value is ok - it's not ok for empty string, so I think it would used used hardly ever.

sfc-gh-asawicki (changes requested):

pkg/sdk/dto-builder-generator/main.go
sfc-gh-jcieslak: Removed this case, re-generated builder methods for the external table, and adjusted the resource so It will only set those fields when they're set.

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #2820, opened 2024-05-17
chore: Improve generator template organization

sfc-gh-jcieslak:
Changes done:
- Templates were moved to files, so that
  - We get proper highlighting from IDE
  - We get help with model type and fields or methods inside it + ability to go to definition and more (https://blog.jetbrains.com/go/2018/12/14/go-templates-made-easy/)
  - Reorganized into smaller chunks, so it should be easier to navigate (main templates are in the `templates` directory, and helper templates used inside the main ones are kept inside the `subtemplates` directory).
- In the `templates.go` now all of the templates are loaded by embed, except the ones that use other templates which have to be initiated in `func init()`. 
- All of the testing was made on the `managed account` resource and it shouldn't collide with #2816
- Also, there were made small adjustments like 
  - Checking for operation name inside unit tests so we won't make invalid Show test
  - Mapping adjusted to the changes made in #2816


sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):
Hard to validate the changes with such a split but I like it. Approve.

PR #2822, opened 2024-05-20
chore: Make ShowByID configurable in generator

sfc-gh-jmichalak:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*

### DIFF
--- a/pkg/sdk/poc/generator/config.go
+++ b/pkg/sdk/poc/generator/config.go
@@ -1,0 +1,26 @@
+package generator
+
+type Config struct {
+	Values map[string]any
+}
+
+func NewConfig() *Config {
+	return &Config{
+		Values: make(map[string]any),
+	}
+}
+
+type Pattern struct {
+	Field   string
+	Pattern string
+}
+
+func (c *Config) WithLike(patterns []Pattern) *Config {
+	c.Values["Like"] = patterns
+	return c
+}
+
+func (c *Config) WithIn(patterns []Pattern) *Config {
+	c.Values["In"] = patterns
+	return c
+}
--- a/pkg/sdk/poc/generator/plain_struct.go
+++ b/pkg/sdk/poc/generator/plain_struct.go
@@ -53,6 +53,10 @@ func (v *plainStruct) OptionalNumber(dbName string) *plainStruct {
 	return v.Field(dbName, "*int")
 }
 
+func (v *plainStruct) PlainStructField(name string, plainStruct *plainStruct) *plainStruct {
+	return v.Field(name, plainStruct.IntoField().Kind)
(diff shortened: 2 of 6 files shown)


pkg/sdk/views_def.go
```diff
@@ -201,7 +201,11 @@ var ViewsDef = g.NewInterface(
 			OptionalStartsWith().
 			OptionalLimit(),
 	).
-	ShowByIdOperation().
+	ShowByIdOperation(
```
sfc-gh-jcieslak, line 204: I feel like an additional layer of abstraction would be better. I think receiving an enum for common presets would be sufficient (like in the case of DescribeOperation), e.g.
```go
// enum names could be better, but that's what I imagined it like
ShowByIdOperation(g.ShowByIdLikeFiltering)
ShowByIdOperation(g.ShowByIdInDatabaseWithLikeFiltering)
ShowByIdOperation(g.ShowByIdInSchemaWithLikeFiltering)
```

pkg/sdk/poc/generator/config.go
```diff
@@ -0,0 +1,26 @@
+package generator
+
+type Config struct {
```
sfc-gh-jcieslak, line 3: Config is very generic, maybe something like `ShowByIdConfig` would be better. Oh ok, I see it's made to be more generic, but as I mentioned this could be replaced most likely with enum value on Operation struct level like in the case of Describe's different "modes". wdyt?

pkg/sdk/poc/generator/plain_struct.go
```diff
@@ -53,6 +53,10 @@ func (v *plainStruct) OptionalNumber(dbName string) *plainStruct {
 	return v.Field(dbName, "*int")
 }
 
+func (v *plainStruct) PlainStructField(name string, plainStruct *plainStruct) *plainStruct {
```
sfc-gh-jcieslak, line 56: It's not used anywhere (or I missed something), do you think it's needed?

sfc-gh-jmichalak: I used it earlier, forgot to remove.

pkg/sdk/views_def.go
sfc-gh-asawicki: This is exactly what I wanted to point out. I would even say that it would be possible to base the `In` choice on the identifier type (because for schema objects with search in schema, for db objects we search in db, etc.). But the one suggested by @sfc-gh-jcieslak should be enough.

sfc-gh-asawicki (changes requested):
I am not leaving more implementation comments because the whole approach should be changed according to one of the above comments

PR #2833, opened 2024-05-23
chore: Random ids rework part3

sfc-gh-asawicki:
TODO


pkg/sdk/random_test.go
```diff
@@ -4,10 +4,19 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers/random"
 )
 
+var (
+	invalidAccountObjectIdentifier = NewAccountObjectIdentifier(random.StringN(256))
+	invalidSchemaObjectIdentifier  = NewSchemaObjectIdentifier(random.StringN(255), random.StringN(255), random.StringN(255))
```
sfc-gh-jcieslak, line 9: I get why the upper one is invalid, but I'm not sure why this one is invalid 😕

pkg/sdk/testint/external_functions_integration_test.go
```diff
@@ -125,6 +124,7 @@ func TestInt_ExternalFunctions(t *testing.T) {
 
 	t.Run("alter external function: set api integration", func(t *testing.T) {
 		e := createExternalFunction(t)
+		e.ID()
```
sfc-gh-jcieslak, line 127: I'm guessing this should be assigned to `id` below

pkg/sdk/testint/tags_integration_test.go
```diff
@@ -16,9 +16,7 @@ func TestInt_Tags(t *testing.T) {
 	client := testClient(t)
 	ctx := context.Background()
 
-	databaseTest, databaseCleanup := testClientHelper().Database.CreateDatabase(t)
-	t.Cleanup(databaseCleanup)
-	schemaTest, schemaCleanup := testClientHelper().Schema.CreateSchemaInDatabase(t, databaseTest.ID())
+	schema, schemaCleanup := testClientHelper().Schema.CreateSchema(t)
```
sfc-gh-jcieslak, line 19: I think tags could also be tested on the predefined schema, but I guess that can be adjusted later on during the refactor for tags.

pkg/sdk/random_test.go
sfc-gh-asawicki: It's too long :D (it was just moved): https://docs.snowflake.com/en/sql-reference/identifiers-syntax

pkg/sdk/testint/tags_integration_test.go
sfc-gh-asawicki: Yep, it's not the scope for this change (I didn't want to think about it for every resource's test in this change)

pkg/sdk/testint/external_functions_integration_test.go
sfc-gh-asawicki: oh, I missed that one. Actually, no, it should be just removed. I left `NewSchemaObjectIdentifierWithArguments` for the identifiers rework (because of the arguments part and the fact that `NewSchemaObjectIdentifierWithArguments` is not equal to `NewSchemaObjectIdentifier` - because of the lacking arguments underneath).

I will remove this e.ID() in the following PR, it's not hurting

sfc-gh-asawicki: (I have added this topic to the identifiers rework)

sfc-gh-jmichalak (approved):

pkg/sdk/random_test.go
sfc-gh-jcieslak: Hmm, I thought 255 is a capacity for a "part", so individually database, schema, and schema object name are valid.

sfc-gh-jcieslak (approved):

PR #2843, opened 2024-05-29
Do not escape table column name.

pbsf:
Snowflake allows a table column to have the char `'` in its name (e.g. `my'column`). The terraform-provider currently escapes it, making creating such a column name impossible (e.g. it will create `my\'column` instead).

This PR removes the escaping logic, which shouldn't be needed since the column name is wrapped around double quotes.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

* Issue: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2844

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,23 +1,23 @@
 builds:
-- env:
-  - >-
-    {{- if or (eq .Os "windows") (eq .Os "darwin") -}}
-      CGO_ENABLED=1
-    {{- else -}}
-      CGO_ENABLED=0
-    {{- end -}}
-  goos:
-    - windows
-    - linux
-    - darwin
-    - freebsd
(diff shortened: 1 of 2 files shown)


sfc-gh-asawicki:
Hey @pbsf. Thanks for the contribution.

Because of what I wrote in the comment: https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2844#issuecomment-2141386050, we won't introduce any more non-crucial changes to the resource before the redesign.

A few comments about submitting the contributions in the future:
1. Please familiarize yourself with our [contribution guidelines](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/CONTRIBUTING.md).
2. Please test the change according to the aforementioned guidelines.

PR #2846, opened 2024-05-31
feat: SCIM integration v1 readiness

sfc-gh-jmichalak:
- Added missing fields
- Renamed fields to match docs
- Migrated to Context functions
- Added acceptance tests
- Fixed handling empty comments in SDK - changed UNSET to SET with empty value
## Test Plan

* [x] acceptance tests

## References

https://docs.snowflake.com/en/sql-reference/sql/create-security-integration-scim


pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
-	row := snowflake.QueryRow(db, stmt)
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
 
-	// Some properties can come from the SHOW INTEGRATION call
-
-	s, err := snowflake.ScanScimIntegration(row)
+	integration, err := client.SecurityIntegrations.ShowByID(ctx, id)
 	if err != nil {
-		return fmt.Errorf("could not show security integration")
+		log.Printf("[DEBUG] notification integration (%s) not found", d.Id())
```
sfc-gh-jcieslak, line 124: Check for `errors.Is(sdk.ErrObjectNotFound, err)` and only then remove the resources `d.SetId("")`

pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
-	// Note: category must be Security or something is broken
-	if c := s.Category.String; c != "SECURITY" {
-		return fmt.Errorf("expected %v to be an Security integration, got %v", id, c)
+	if c := integration.Category; c != "SECURITY" {
+		return diag.FromErr(fmt.Errorf("expected %v to be a SECURITY integration, got %v", id, c))
 	}
 
-	if err := d.Set("scim_client", strings.TrimPrefix(s.IntegrationType.String, "SCIM - ")); err != nil {
-		return err
+	if err := d.Set("name", integration.Name); err != nil {
```
sfc-gh-jcieslak, line 133: Recently I was wondering if we want to set the `name` field. To read this integration we searched by `name`, so It should be the same and the output from Snowflake may differ in casing/quoting which may result in plan being produced for `name`.

docs/resources/scim_integration.md
```diff
@@ -25,13 +28,16 @@ resource "snowflake_scim_integration" "aad" {
 
 ### Required
 
-- `name` (String) Specifies the name of the SCIM integration. This name follows the rules for Object Identifiers. The name should be unique among security integrations in your account.
-- `provisioner_role` (String) Specify the SCIM role in Snowflake that owns any users and roles that are imported from the identity provider into Snowflake using SCIM.
+- `enabled` (Boolean) Specify whether the security integration is enabled.
```
sfc-gh-asawicki, line 31: what was the behavior before? was it enabled by default or disabled by default?

Also, this is a breaking change (the old config won't work); please add this to the migration guide.

docs/resources/scim_integration.md
```diff
@@ -12,11 +12,14 @@ description: |-
 ## Example Usage
 
 ```terraform
-resource "snowflake_scim_integration" "aad" {
-  name             = "AAD_PROVISIONING"
-  network_policy   = "AAD_NETWORK_POLICY"
-  provisioner_role = "AAD_PROVISIONER"
-  scim_client      = "AZURE"
+resource "snowflake_scim_integration" "test" {
```
sfc-gh-asawicki, line 15: let's have a minimal example and all fields example

pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
-
-	stmt := snowflake.NewSCIMIntegrationBuilder(id).Alter()
-
-	var runSetStatement bool
-
-	if d.HasChange("scim_client") {
-		runSetStatement = true
-		stmt.SetString(`SCIM_CLIENT`, d.Get("scim_client").(string))
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
+	var runSet, runUnset bool
```
sfc-gh-jcieslak, line 196: You can do as in e.g. secondary_database, that instead of setting flags you can compare `set` with `sdk.NewScimIntegrationSetRequest()` to see if any of the fields has been set.

docs/resources/scim_integration.md
```diff
@@ -25,13 +28,16 @@ resource "snowflake_scim_integration" "aad" {
...
+- `enabled` (Boolean) Specify whether the security integration is enabled.
+- `name` (String) String that specifies the identifier (i.e. name) for the integration; must be unique in your account.
+- `run_as_role` (String) Specify the SCIM role in Snowflake that owns any users and roles that are imported from the identity provider into Snowflake using SCIM.
 - `scim_client` (String) Specifies the client type for the scim integration
 
 ### Optional
 
-- `network_policy` (String) Specifies an existing network policy active for your account. The network policy restricts the list of user IP addresses when exchanging an authorization code for an access or refresh token and when using a refresh token to obtain a new access token. If this parameter is not set, the network policy for the account (if any) is used instead.
+- `comment` (String) Specifies a comment for the integration.
+- `network_policy` (String) Specifies an existing network policy that controls SCIM network traffic.
```
sfc-gh-asawicki, line 39: let's keep track of the newly added fields (and other changes), so that we can communicate them more easily (with release notes or maybe we'll come up with some other way).

For now, please create a new MD file in `v1-preparations`, e.g. changes-before-v1 (or a better name), please put an explanation sentence (that this is a changelog in resources and datasources as part of the https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#preparing-essential-ga-objects-for-the-provider-v1) and v0.91.0-v0.92.0 changes (for now with your resource section only; @sfc-gh-jcieslak will add database and I will add warehouse).

pkg/helpers/helpers.go
```diff
@@ -167,3 +167,14 @@ func Retry(attempts int, sleepDuration time.Duration, f func() (error, bool)) er
 	}
 	return fmt.Errorf("giving up after %v attempts", attempts)
 }
+
+// MatchStringsRegex returns a regex matching every string in stringSlice
+func MatchStringsRegex(stringSlice []string) *regexp.Regexp {
```
sfc-gh-asawicki, line 172: 1. Please unit-test it.
2. Let's move it to acceptance package (for now).
3. is `fmt.Println(a)` still needed? If so, let's add a DEBUG log instead.
4. escapedMessages -> escapedParts
5. stringSlice -> parts
6. MatchStringsRegex -> MatchAllStrings?
7. Order matter currently, right? Let's be explicit about this MatchAllStrings -> MatchAllStringsInOrder or smth

pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
+			runSet = true
+			set.WithNetworkPolicy(networkPolicyID)
 		} else {
-			runSetStatement = true
-			stmt.SetString(`NETWORK_POLICY`, d.Get("network_policy").(string))
+			runUnset = true
+			unset.WithNetworkPolicy(true)
 		}
 	}
+	if d.GetRawConfig().AsValueMap()["sync_password"].IsNull() {
```
sfc-gh-jcieslak, line 212: Why RawConfig? I suspect this may be connected to the fact the `sync_password` change is not being detected on update "normally" (It could be resolved by writing customdiff to "refresh" the value).

pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
+		unset.WithSyncPassword(true)
+	} else if d.HasChange("sync_password") {
+		runSet = true
+		set.WithSyncPassword(d.Get("sync_password").(bool))
+	}
 
-	if runSetStatement {
-		if err := snowflake.Exec(db, stmt.Statement()); err != nil {
-			return fmt.Errorf("error updating security integration")
+	if d.HasChange("comment") {
```
sfc-gh-jcieslak, line 220: Not that crucial, but we should agree on what should happen when the comment is empty (`unset` or `set` empty string?).

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -1,79 +1,180 @@
...
 
-func TestAcc_ScimIntegration(t *testing.T) {
-	scimIntName := acc.TestClient().Ids.Alpha()
-	scimProvisionerRole := "AAD_PROVISIONER"
-	scimNetworkPolicy := acc.TestClient().Ids.Alpha()
+func TestAcc_ScimIntegration_basic(t *testing.T) {
+	networkPolicy, networkPolicyCleanup := acc.TestClient().NetworkPolicy.CreateNetworkPolicy(t)
+	t.Cleanup(networkPolicyCleanup)
 
+	role, roleCleanup := acc.TestClient().Role.CreateRoleWithRequest(t, sdk.NewCreateRoleRequest(snowflakeroles.GenericScimProvisioner).WithOrReplace(true))
```
sfc-gh-jcieslak, line 20: Hmm 🤔 , I thought these are predefined like SECURITYADMIN, ACCOUNTADMIN, etc.

pkg/sdk/testint/security_integrations_gen_integration_test.go
```diff
@@ -1285,10 +1285,19 @@ func TestInt_SecurityIntegrations(t *testing.T) {
 		err = client.SecurityIntegrations.AlterScim(ctx, unsetRequest)
 		require.NoError(t, err)
 
+		// check setting empty comment because of lacking UNSET COMMENT
```
sfc-gh-asawicki, line 1288: I don't like this logic, as stated in your doc-discuss comment and DOC-2249, if it works for other integrations, it is most probably a SF bug.

Let's leave it like this, but let's create a task to remember about this, so that we can address it when it's fixed. Let's add its number to a TODO comment (and add reference to DOC-2249 in it)

pkg/sdk/security_integrations_validations_gen.go
```diff
@@ -98,9 +98,6 @@ func (opts *CreateOauthForPartnerApplicationsSecurityIntegrationOptions) validat
 	if everyValueSet(opts.OrReplace, opts.IfNotExists) {
 		errs = append(errs, errOneOf("CreateOauthForPartnerApplicationsSecurityIntegrationOptions", "OrReplace", "IfNotExists"))
 	}
-	if opts.OauthClient == OauthSecurityIntegrationClientLooker && opts.OauthRedirectUri == nil {
```
sfc-gh-asawicki, line 101: why this was removed?

pkg/sdk/security_integrations_gen.go
```diff
@@ -487,14 +487,17 @@ type ScimIntegrationSet struct {
 	Enabled       *bool                    `ddl:"parameter" sql:"ENABLED"`
 	NetworkPolicy *AccountObjectIdentifier `ddl:"identifier,equals" sql:"NETWORK_POLICY"`
 	SyncPassword  *bool                    `ddl:"parameter" sql:"SYNC_PASSWORD"`
-	Comment       *string                  `ddl:"parameter,single_quotes" sql:"COMMENT"`
+	Comment       *StringAllowEmpty        `ddl:"parameter" sql:"COMMENT"`
+}
+
+type StringAllowEmpty struct {
```
sfc-gh-asawicki, line 493: it seems like something that should be part of commons; it would be nice to add a comment documenting its purpose

pkg/sdk/security_integrations_dto_gen.go
```diff
@@ -414,14 +414,17 @@ type ScimIntegrationSetRequest struct {
 	Enabled       *bool
 	NetworkPolicy *AccountObjectIdentifier
 	SyncPassword  *bool
-	Comment       *string
+	Comment       *StringAllowEmptyRequest
+}
+
+type StringAllowEmptyRequest struct {
```
sfc-gh-asawicki, line 420: commons

pkg/sdk/security_integrations_def.go
```diff
@@ -307,15 +308,14 @@ var scimIntegrationSetDef = g.NewQueryStruct("ScimIntegrationSet").
 	OptionalBooleanAssignment("ENABLED", g.ParameterOptions()).
 	OptionalIdentifier("NetworkPolicy", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().Equals().SQL("NETWORK_POLICY")).
 	OptionalBooleanAssignment("SYNC_PASSWORD", g.ParameterOptions()).
-	OptionalComment().
+	OptionalQueryStructField("Comment", stringAllowEmptyDef, g.ParameterOptions().SQL("COMMENT")).
```
sfc-gh-jcieslak, line 311: Not sure about this change, what is it for?

sfc-gh-jcieslak: Ah, I remember, set comment wasn't produced for empty comment, right? In that case you can just compare comment field contents like so `d.Get("comment").(string) == "")` and if it's true you can call unset otherwise set.

pkg/sdk/security_integrations_validations_gen.go
sfc-gh-jcieslak, line 101: Why it's removed? It seems correct

pkg/resources/scim_integration.go
```diff
@@ -17,40 +20,49 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
+	},
+	"enabled": {
+		Type:        schema.TypeBool,
+		Required:    true,
+		Description: "Specify whether the security integration is enabled.",
 	},
 	"scim_client": {
 		Type:        schema.TypeString,
 		Required:    true,
+		ForceNew:    true,
```
sfc-gh-asawicki, line 33: this is a breaking change too

pkg/resources/scim_integration.go
```diff
@@ -17,40 +20,49 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
+		ValidateFunc: validation.StringInSlice(sdk.AsStringList([]sdk.ScimSecurityIntegrationScimClientOption{
+			sdk.ScimSecurityIntegrationScimClientOkta, sdk.ScimSecurityIntegrationScimClientAzure, sdk.ScimSecurityIntegrationScimClientGeneric,
+		}), true),
+		DiffSuppressFunc: ignoreTrimSpaceSuppressFunc,
 	},
-	"provisioner_role": {
+	"run_as_role": {
 		Type:        schema.TypeString,
 		Required:    true,
+		ForceNew:    true,
```
sfc-gh-asawicki, line 43: bc

pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
+	scimClient := d.Get("scim_client").(string)
+	runAsRole := d.Get("run_as_role").(string)
 
-	stmt := snowflake.NewSCIMIntegrationBuilder(name).Create()
-
-	// Set required fields
-	stmt.SetRaw(`TYPE=SCIM`)
-	stmt.SetString(`SCIM_CLIENT`, d.Get("scim_client").(string))
-	stmt.SetString(`RUN_AS_ROLE`, d.Get("provisioner_role").(string))
+	req := sdk.NewCreateScimSecurityIntegrationRequest(id, enabled, sdk.ScimSecurityIntegrationScimClientOption(scimClient), sdk.ScimSecurityIntegrationRunAsRoleOption(runAsRole))
```
sfc-gh-asawicki, line 98: Instead of sdk.ScimSecurityIntegrationScimClientOption(scimClient) We should have a `ToScimClient` method on the SDK level (now casing may be different, and it may cause some hard-to-spot errors in the future)

pkg/resources/scim_integration.go
sfc-gh-asawicki, line 124: shouldn't we have a check for errors.Is(err, sdk.ErrObjectNotFound)  first and differentiate between different errors here?

pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
-		return fmt.Errorf("could not show security integration")
+		log.Printf("[DEBUG] notification integration (%s) not found", d.Id())
+		d.SetId("")
+		return diag.FromErr(err)
 	}
 
-	// Note: category must be Security or something is broken
-	if c := s.Category.String; c != "SECURITY" {
-		return fmt.Errorf("expected %v to be an Security integration, got %v", id, c)
+	if c := integration.Category; c != "SECURITY" {
```
sfc-gh-asawicki, line 129: let's introduce a const for this instead of a plain string (on SDK side)

pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
-	if err != nil {
-		return fmt.Errorf("could not describe security integration")
+	if err := d.Set("enabled", integration.Enabled); err != nil {
+		return diag.FromErr(err)
 	}
-	defer rows.Close()
-	for rows.Next() {
-		if err := rows.Scan(&k, &pType, &v, &unused); err != nil {
-			return fmt.Errorf("unable to parse security integration rows")
+	typeParts := strings.Split(integration.IntegrationType, "-")
```
sfc-gh-asawicki, line 148: this logic should happen on the SDK side

pkg/resources/scim_integration.go
```diff
@@ -74,140 +86,170 @@ func SCIMIntegration() *schema.Resource {
...
+	for _, property := range integrationProperties {
+		name := property.Name
+		value := property.Value
+		switch name {
+		case "ENABLED", "COMMENT":
+			// We set this using the SHOW INTEGRATION call so let's ignore it here
 		case "NETWORK_POLICY":
-			if err := d.Set("network_policy", v.(string)); err != nil {
-				return fmt.Errorf("unable to set network policy for security integration")
+			if err := d.Set("network_policy", value); err != nil {
```
sfc-gh-asawicki, line 173: network policy is an identifier, should we do some preprocessing before putting to state? (i.e. what if it is returned with quotes from SF?)

pkg/resources/scim_integration.go
```diff
@@ -17,40 +20,49 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
-			}
-			return normalize(old) == normalize(new)
-		},
+		ValidateFunc: validation.StringInSlice(sdk.AsStringList([]sdk.ScimSecurityIntegrationScimClientOption{
+			sdk.ScimSecurityIntegrationScimClientOkta, sdk.ScimSecurityIntegrationScimClientAzure, sdk.ScimSecurityIntegrationScimClientGeneric,
+		}), true),
+		DiffSuppressFunc: ignoreTrimSpaceSuppressFunc,
 	},
-	"provisioner_role": {
+	"run_as_role": {
```
sfc-gh-asawicki, line 40: is a role also an account object identifier?

pkg/resources/scim_integration.go
sfc-gh-asawicki, line 196: please check e.g. secondary_database, we introduced a new, simpler way to handle update sets/unsets

pkg/resources/scim_integration.go
```diff
@@ -17,40 +20,49 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
-		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-			normalize := func(s string) string {
-				return strings.ToUpper(strings.ReplaceAll(s, "-", ""))
-			}
-			return normalize(old) == normalize(new)
-		},
+		ValidateFunc: validation.StringInSlice(sdk.AsStringList([]sdk.ScimSecurityIntegrationRunAsRoleOption{
+			sdk.ScimSecurityIntegrationRunAsRoleOktaProvisioner, sdk.ScimSecurityIntegrationRunAsRoleAadProvisioner, sdk.ScimSecurityIntegrationRunAsRoleGenericScimProvisioner,
+		}), true),
+		DiffSuppressFunc: ignoreTrimSpaceSuppressFunc,
```
sfc-gh-asawicki, line 48: casing in the previous suppression mattered

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -1,79 +1,180 @@
 package resources_test
 
 import (
-	"fmt"
 	"testing"
```
sfc-gh-asawicki, line 4: please add a test for migration from v091 to v092 (resource created in 091, provisioner_name -> run_as_role); the test is necessary so we know, what to recommend to our users (because currently the migration notes are not guiding with this change); maybe some state migratory should be introduced, based on the outcome of the test, we can discuss when you have it

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-asawicki, line 20: I thought these roles are preexisting, because you have added them as constants, I think that we should not replace the if they are preexisting.

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -1,79 +1,180 @@
...
 		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
 			tfversion.RequireAbove(tfversion.Version1_5_0),
 		},
-		CheckDestroy: nil,
 		Steps: []resource.TestStep{
 			{
-				Config: scimIntegrationConfigAzure(scimIntName, scimProvisionerRole, scimNetworkPolicy),
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_ScimIntegration/basic"),
+				ConfigVariables: m(false, sdk.ScimSecurityIntegrationScimClientGeneric, role, false),
+				Check: resource.ComposeTestCheckFunc(
```
sfc-gh-asawicki, line 52: let's check the computed fields too

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -1,79 +1,180 @@
...
-		depends_on = [
-			snowflake_grant_privileges_to_account_role.azure_grants,
-			snowflake_grant_account_role.azure
-		]
+func TestAcc_ScimIntegration_invalid(t *testing.T) {
+	id := acc.TestClient().Ids.RandomAccountObjectIdentifier()
+	m := func() map[string]config.Variable {
+		return map[string]config.Variable{
+			"name":                config.StringVariable(id.Name()),
+			"enabled":             config.BoolVariable(false),
```
sfc-gh-asawicki, line 155: let's not put enabled (another error should pop out)

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -1,79 +1,180 @@
...
+			"scim_client":         config.StringVariable("invalid"),
+			"sync_password":       config.BoolVariable(false),
+			"network_policy_name": config.StringVariable("invalid"),
```
sfc-gh-asawicki, line 158: let's put incorrect identifier for another error

sfc-gh-asawicki (changes requested):

pkg/resources/scim_integration.go
```diff
@@ -17,40 +20,49 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
 	"network_policy": {
+		Type:             schema.TypeString,
+		ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
+		Optional:         true,
+		Description:      "Specifies an existing network policy that controls SCIM network traffic.",
+	},
+	"sync_password": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Computed:    true,
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Added.

pkg/resources/scim_integration.go
```diff
@@ -17,40 +20,49 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
+			sdk.ScimSecurityIntegrationScimClientOkta, sdk.ScimSecurityIntegrationScimClientAzure, sdk.ScimSecurityIntegrationScimClientGeneric,
+		}), true),
+		DiffSuppressFunc: ignoreTrimSpaceSuppressFunc,
 	},
-	"provisioner_role": {
+	"run_as_role": {
 		Type:        schema.TypeString,
 		Required:    true,
+		ForceNew:    true,
 		Description: "Specify the SCIM role in Snowflake that owns any users and roles that are imported from the identity provider into Snowflake using SCIM.",
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration.go
```diff
@@ -17,40 +20,49 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
+	"enabled": {
+		Type:        schema.TypeBool,
+		Required:    true,
+		Description: "Specify whether the security integration is enabled.",
 	},
 	"scim_client": {
 		Type:        schema.TypeString,
 		Required:    true,
+		ForceNew:    true,
 		Description: "Specifies the client type for the scim integration",
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: I think in our guidelines we agreed to UNSET in such cases. However, in this case, UNSET COMMENT is not working, that's why I'm using SET.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: They aren't, that's why they need to be provided here.

pkg/sdk/security_integrations_def.go
sfc-gh-jmichalak: As I wrote above, UNSET COMMENT is not working for this resource, that's why we need to support SET COMMENT=''. Without this change this SQL is not generated.

pkg/sdk/security_integrations_validations_gen.go
sfc-gh-jmichalak: Just generator things :( Fixed.

docs/resources/scim_integration.md
sfc-gh-jmichalak: It was disabled by default. Added to migration guide.

docs/resources/scim_integration.md
sfc-gh-jmichalak: Done.

docs/resources/scim_integration.md
sfc-gh-jmichalak: Done.

pkg/sdk/security_integrations_validations_gen.go
sfc-gh-jmichalak: Just generator things :( Fixed.

pkg/sdk/security_integrations_gen.go
sfc-gh-jmichalak: Moved to `common_types.go` and added a comment.

pkg/sdk/security_integrations_dto_gen.go
sfc-gh-jmichalak: Removed.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Added to migration guide.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Done, also for `runAsRole`.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Moved.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Yes, but it's an enum of 3 roles below, that's why I decided not to validate as AccountObjectIdentifier.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: They are not guaranteed by snowflake and they may not exist in a deployment. That's why we need to provide them. They are added as constants because role names are listed in enum (check [docs](https://docs.snowflake.com/en/sql-reference/sql/create-security-integration-scim)). So, if a user declares an integration with one of these roles, they need to exist in a deployment.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: The only check that was missing was TestCheckResourceAttrSet with run_as_role - added.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: Done in a separate test case because enum validations errors don't appear together with it.

pkg/sdk/testint/security_integrations_gen_integration_test.go
sfc-gh-jmichalak: Done.

pkg/helpers/helpers.go
sfc-gh-jmichalak: Right, the order matters here. Added unit tests and changed some names.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Fixed. According to docs: 
> The values OKTA_PROVISIONER, AAD_PROVISIONER, and GENERIC_SCIM_PROVISIONER are case-sensitive and must always be capitalized.

pkg/resources/scim_integration.go
sfc-gh-jcieslak: Yesterday we talked about `name`s (again ;_;) and shouldn't we actually have them in in Read, so they'll be read during import?

docs/resources/scim_integration.md
sfc-gh-asawicki: The migration guide's purpose is to guide the user through the migration. It's not enough to write: `Field `enabled` is now required.` We have to be explicit about what the user has to do, what was the previous behavior, etc. In this case, if the previous behavior was disabled by default, then we should add something like: `Previously the default behavior was disabled, so please add enabled: false if you want to mimic the old one.` (or something better).

The alternative would be to consider setting a default in this case (it would be an exception to our baseline that we usually steer away from defaults). But maybe, in this case, it would be okay because it is not based on any Snowflake default underneath but would be a default in our provider. Even in this case, a short sentence for the user would be appreciated.

@sfc-gh-jmichalak @sfc-gh-jcieslak what do you think?

pkg/acceptance/helpers/common_test.go
```diff
@@ -0,0 +1,46 @@
...
+		"empty text": {
+			parts: []string{"a", "b"},
+			text:  "",
+		},
+		"matching non empty": {
+			parts:     []string{"a", "b"},
+			text:      "xyaxyb",
+			wantMatch: true,
+		},
+		"not matching": {
```
sfc-gh-asawicki, line 34: nit: this is partial match, we can be explicit with it

pkg/acceptance/helpers/common_test.go
```diff
@@ -0,0 +1,46 @@
+package helpers
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMatchAllStringsInOrder(t *testing.T) {
+	testCases := map[string]struct {
```
sfc-gh-asawicki, line 10: Please add tests for:
- different order of parts (to validate if order really matter)
- fancy case like "abba" with []string{"abb", "bba"} - this should not match given the current implementation, right? but from the method signature it could be expected to

pkg/acceptance/helpers/common.go
```diff
@@ -19,3 +24,23 @@ func EnsureQuotedIdentifiersIgnoreCaseIsSetToFalse(client *sdk.Client, ctx conte
...
+func MatchAllStringsInOrder(parts []string) *regexp.Regexp {
+	escapedParts := make([]string, len(parts))
+	for i := range parts {
+		escapedParts[i] = regexp.QuoteMeta(parts[i])
+	}
+	return regexp.MustCompile(strings.Join(escapedParts, "((.|\n)*)"))
+}
+
+// AssertErrorContainsPartsFunc returns a function asserting error message contains each string in parts
+func AssertErrorContainsPartsFunc(t *testing.T, parts []string) resource.ErrorCheckFunc {
```
sfc-gh-asawicki, line 38: so why do we need a second method doing almost the same thing as the one above?

And also: this one should also be tested, but because assertions are a bit trickier to test (but we have examples in our repository), it does not have to be a part of this PR.

pkg/sdk/testint/security_integrations_gen_integration_test.go
```diff
@@ -1286,10 +1286,11 @@ func TestInt_SecurityIntegrations(t *testing.T) {
 		require.NoError(t, err)
 
 		// check setting empty comment because of lacking UNSET COMMENT
+		// TODO(SNOW-1461780): change this in UNSET
```
sfc-gh-asawicki, line 1289: nit pick: `change this to UNSET` or `replace this with UNSET`

pkg/resources/scim_integration.go
sfc-gh-asawicki: Shouldn't it be like in e.g. secondary database:
```
	if err != nil {
		if errors.Is(err, sdk.ErrObjectNotFound) {
			d.SetId("")
			return diag.Diagnostics{
				diag.Diagnostic{
					Severity: diag.Warning,
					Summary:  "Failed to query secondary database. Marking the resource as removed.",
					Detail:   fmt.Sprintf("DatabaseName: %s, Err: %s", secondaryDatabaseId.FullyQualifiedName(), err),
				},
			}
		}
		return diag.FromErr(err)
	}
```

?
	
Now the behavior is slightly different.

pkg/sdk/security_integrations_gen.go
```diff
@@ -553,3 +559,11 @@ type SecurityIntegration struct {
 	Comment         string
 	CreatedOn       time.Time
 }
+
+func (s *SecurityIntegration) SubType() (string, error) {
```
sfc-gh-asawicki, line 563: nit-pick: unit tests would be nice

pkg/resources/scim_integration.go
sfc-gh-asawicki: But his means, it is put to state with double quotes. So if the user uses it without them in his config, this will cause a perma diff, right?

sfc-gh-asawicki: Oh, okay, you have diff suppression for it. Let's leave it like this then. We will revisit it with identifiers rework

pkg/resources/scim_integration.go
sfc-gh-asawicki: I was not clear with the previous comment:
1. Yes, they must be capitalized and we handle it on the SDK level.
2. We agreed that we allow users to put e.g. lower cased values (this was the previous behavior).
3. With the old implementation (before your changes) it was allowed, and diff suppression handled also lowercase correctly (and my comment was only about that using `ignoreTrimSpaceSuppressFunc` breaks it)

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-asawicki: Okay, in such a case, I think that:
1. We should not clean them up as part of the test.
2. We should pre-create them on our environments.
3. We should add a check similar to the one added recently for Quoted identifiers ignore case param (check `EnsureQuotedIdentifiersIgnoreCaseIsSetToFalse`)
4. We should assume in our tests that they exist and be explicit with it in our docs.
5. It would be interesting to check what happens if such a role does not exist (I mean to have acceptance test for it):
  - is it nicely handled on SF side?
  - do we pass these error nicely to the user? if not, should we handle it with a check on the provider side?
  - if we want to test it, let's maybe create only one of these 3 roles, assert that second does not exist, and have a test for the non-existing one?

@sfc-gh-jcieslak I would like your opinion here too.

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -170,11 +176,102 @@ func TestAcc_ScimIntegration_invalid(t *testing.T) {
...
+}
+`
+	return fmt.Sprintf(s, name, sdk.ScimSecurityIntegrationScimClientGeneric, roleName)
+}
+
+func scimIntegrationv092(name, roleName string) string {
+	s := `
+resource "snowflake_scim_integration" "test" {
+	name             = "%s"
+	enabled          = true
```
sfc-gh-asawicki, line 271: how is this empty plan if the enabled was changed to true? In other comment you said that it was disabled by default so we should see a change. So:
1. Please change this test to enabled false.
2. Add a separate migration one, where enabled is set to true and we expect a change.

pkg/resources/scim_integration.go
```diff
@@ -71,37 +75,72 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
+				Version: 0,
+				// setting type to cty.EmptyObject is a bit hacky here but following https://developer.hashicorp.com/terraform/plugin/framework/migrating/resources/state-upgrade#sdkv2-1 would require lots of repetitive code; this should work with cty.EmptyObject
+				Type:    cty.EmptyObject,
+				Upgrade: v091ScimIntegrationStateUpgrader,
+			},
+		},
+	}
+}
+
+func v091ScimIntegrationStateUpgrader(ctx context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
```
sfc-gh-asawicki, line 110: let's move it to a separate file like with other upgraders

pkg/resources/scim_integration.go
```diff
@@ -71,37 +75,72 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
+					res[i] = props[i]
+				}
+				return res, err
+			}),
+		),
 		Schema: scimIntegrationSchema,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
+		SchemaVersion: 1,
```
sfc-gh-asawicki, line 98: Please add it as first prop of Resource and add interline after

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -130,18 +135,19 @@ func TestAcc_ScimIntegration_complete(t *testing.T) {
...
-				ConfigVariables:   m(),
-				ResourceName:      "snowflake_scim_integration.test",
-				ImportState:       true,
-				ImportStateVerify: true,
+				ConfigDirectory:         acc.ConfigurationDirectory("TestAcc_ScimIntegration/complete"),
+				ConfigVariables:         m(),
+				ResourceName:            "snowflake_scim_integration.test",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"name"},
```
sfc-gh-asawicki, line 150: Okay, so this is a downside of not setting the name. So maybe we should just set it? What were the downsides again?

@sfc-gh-jmichalak @sfc-gh-jcieslak

MIGRATION_GUIDE.md
```diff
@@ -20,6 +20,14 @@ Its purpose was also to divide the resources by their specific purpose rather th
 
 Renamed field `provisioner_role` to `run_as_role` to align with Snowflake docs.
```
sfc-gh-asawicki, line 21: Let's add, that config should be renamed and the state would be migrated automatically

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-asawicki: ok, with identifiers rework then (cc: @sfc-gh-jcieslak )

pkg/resources/custom_diffs.go
```diff
@@ -47,3 +47,36 @@ func NestedValueComputedIf(key string, showParam func(client *sdk.Client) (*sdk.
 		return param.Value != valueToString(stateValue[0].(map[string]any)["value"])
 	})
 }
+
+type Property interface {
+	GetName() string
+	GetDefault() string
+}
+
+func BoolComputedIf(key, property string, describe func(client *sdk.Client, id sdk.AccountObjectIdentifier) ([]Property, error)) schema.CustomizeDiffFunc {
```
sfc-gh-asawicki, line 56: Hmm, let's leave it like this. I think we will change it slightly in a moment after my proposal for warehouses. I think that this logic handles the external change incorrectly (but I see no test for it). Let's leave it for now.

sfc-gh-asawicki (changes requested):

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jcieslak: I answered in the thread above that maybe we should set it, so it's read during the import, but I'm not sure about exact upsides/downsides of reading it in Read operation.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Yeah, imports aren't working correctly because of that. Reverted this and removed skip check.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Fixed.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Reverted to old implementation.

pkg/acceptance/helpers/common_test.go
sfc-gh-jmichalak: Fixed.

pkg/acceptance/helpers/common_test.go
sfc-gh-jmichalak: Done. Renamed the function.

pkg/acceptance/helpers/common.go
sfc-gh-jmichalak: Because for some errors, e.g. missing required fields in config, the order of errors is not deterministic. We get error message as a simple `error` so the alternative is to parse it and sort messages and I'm not a fan of such implementation. Also, function signatures are different because they are used as `resource.TestStep.ExpectError` and `resource.TestCase.ErrorCheck` respectively.

I can add tests in the next PR.

pkg/sdk/testint/security_integrations_gen_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/security_integrations_gen.go
sfc-gh-jmichalak: Added.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: Sorry, I got confused with defaults for `ENABLED` because `property_default` is `false`, but when you run a create without `ENABLED` (it's actually possible, contrary to the docs), the `property_value` is `true`.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: Reverted setting name in Read. Removed ImportStateVerifyIgnore.

MIGRATION_GUIDE.md

pkg/resources/custom_diffs.go
```diff
@@ -47,3 +47,36 @@ func NestedValueComputedIf(key string, showParam func(client *sdk.Client) (*sdk.
...
+	return customdiff.ComputedIf(key, func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
+		configValue := d.GetRawConfig().AsValueMap()[key]
+		if !configValue.IsNull() {
+			return false
+		}
+
+		client := meta.(*provider.Context).Client
+
+		props, err := describe(client, sdk.NewAccountObjectIdentifier(d.Id()))
```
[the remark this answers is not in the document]

pkg/resources/scim_integration.go
```diff
@@ -71,37 +75,72 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
 		DeleteContext: DeleteContextSCIMIntegration,
-
+		CustomizeDiff: customdiff.All(
+			BoolComputedIf("sync_password", "SYNC_PASSWORD", func(client *sdk.Client, id sdk.AccountObjectIdentifier) ([]Property, error) {
+				props, err := client.SecurityIntegrations.Describe(context.Background(), id)
+				res := make([]Property, len(props))
+				for i := range props {
+					res[i] = props[i]
+				}
+				return res, err
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Hmm I was filtering them in BoolComputedIf with a different approach before, but I think it should be done here. Changed.

pkg/resources/scim_integration.go
```diff
@@ -17,40 +23,50 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
-				return strings.ToUpper(strings.ReplaceAll(s, "-", ""))
-			}
-			return normalize(old) == normalize(new)
+	"scim_client": {
+		Type:         schema.TypeString,
+		Required:     true,
+		ForceNew:     true,
+		Description:  fmt.Sprintf("Specifies the client type for the scim integration. Valid options are: %v.", sdk.AsStringList(sdk.AllScimSecurityIntegrationScimClients)),
+		ValidateFunc: validation.StringInSlice(sdk.AsStringList(sdk.AllScimSecurityIntegrationScimClients), true),
+		DiffSuppressFunc: func(_, old, new string, _ *schema.ResourceData) bool {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Done.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jcieslak: Yeah, It's confusing to have predefined values that you have to create on your own, but we have to deal with it somehow. Having it pre-created to be able to use them without remembering to create them before would be nice addition. I agree with all the points.

docs/resources/scim_integration.md
sfc-gh-jcieslak: I think we can describe the change more clearly in the migration guide and leave it without a default (?). I think it makes it more readable from the user's point of view to see that the migration is not abled rather than have it hidden away.

pkg/resources/custom_diffs.go
sfc-gh-jmichalak: Okay.

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: Done, also removed creating these roles from integration tests. Note that  `AadProvisioner` is not provisioned because it's not used in tests.

docs/resources/scim_integration.md
sfc-gh-jmichalak: It turned out that this field is actually optional, so I changed SDK and marked `enabled` as optional. But the behavior of this field is odd: property_default is false, but when you run a create without ENABLED, the property_value is true. When you do an UNSET, the `property_value` is false. I think we should state this in field description or introduce some custom logic to handle this case. What do you think?

pkg/acceptance/helpers/common.go
```diff
@@ -19,3 +25,59 @@ func EnsureQuotedIdentifiersIgnoreCaseIsSetToFalse(client *sdk.Client, ctx conte
 	}
 	return nil
 }
+
+func EnsureScimProvisionerRolesExist(client *sdk.Client, ctx context.Context) error {
+	log.Printf("[DEBUG] Making sure Scim Provisioner roles exist")
+	roleIDs := []sdk.AccountObjectIdentifier{snowflakeroles.GenericScimProvisioner, snowflakeroles.OktaProvisioner}
```
sfc-gh-jcieslak, line 31: What about `AAD_PROVISIONER`?

pkg/acceptance/helpers/common.go
```diff
@@ -19,3 +25,59 @@ func EnsureQuotedIdentifiersIgnoreCaseIsSetToFalse(client *sdk.Client, ctx conte
...
+		}
+		grants, err := client.Grants.Show(ctx, &sdk.ShowGrantOptions{
+			Of: &sdk.ShowGrantsOf{
+				Role: roleID,
+			},
+		})
+		if err != nil {
+			return err
+		}
+		if !hasGranteeName(grants, currentRoleID) {
```
sfc-gh-jcieslak, line 49: Should we also check for those roles to be granted to the current user? Otherwise, we won't be able to call `use role` on them, was it intentional or not that useful to do for those roles?

pkg/resources/scim_integration.go
```diff
@@ -24,34 +27,45 @@ var scimIntegrationSchema = map[string]*schema.Schema{
...
 	"enabled": {
 		Type:        schema.TypeBool,
-		Required:    true,
+		Optional:    true,
 		Description: "Specify whether the security integration is enabled.",
+		DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+			if d.GetRawConfig().AsValueMap()["enabled"].IsNull() {
+				return true
+			}
+			return oldValue == newValue
```
sfc-gh-jcieslak, line 36: Hmm, I think in this case the diff shouldn't be produced, but maybe I'm wrong.

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -1,38 +1,34 @@
...
 	t.Cleanup(networkPolicyCleanup)
-
-	role, roleCleanup := acc.TestClient().Role.CreateRoleWithRequest(t, sdk.NewCreateRoleRequest(snowflakeroles.GenericScimProvisioner).WithOrReplace(true))
-	t.Cleanup(roleCleanup)
-	acc.TestClient().Role.GrantRoleToCurrentRole(t, role.ID())
-	role2, role2Cleanup := acc.TestClient().Role.CreateRoleWithRequest(t, sdk.NewCreateRoleRequest(snowflakeroles.OktaProvisioner).WithOrReplace(true))
-	t.Cleanup(role2Cleanup)
-	acc.TestClient().Role.GrantRoleToCurrentRole(t, role2.ID())
-
+	role, role2 := snowflakeroles.GenericScimProvisioner, snowflakeroles.OktaProvisioner
```
sfc-gh-jcieslak, line 22: What about aad one ?

docs/resources/scim_integration.md
sfc-gh-jcieslak: Yeah, maybe document it and later on resolve this with security integration guys or doc discuss team that would follow up with them (because the documentation is probably not including that either).

pkg/acceptance/helpers/common.go
sfc-gh-jmichalak: I think this check is sufficient - we don't call `use role` for these roles.

pkg/acceptance/helpers/common.go
sfc-gh-jmichalak: It's not used in security integration tests.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: Fixed.

docs/resources/scim_integration.md
sfc-gh-jmichalak: Marked field `enabled` as required. Updated tests and added info to migration guide.

pkg/resources/scim_integration.go
sfc-gh-asawicki: Hmm. Okay, but the old implementation (with `return strings.ToUpper(strings.ReplaceAll(s, "-", ""))` is kinda fishy, is it necessary?

Let's talk about this and solve this in the next PR.

pkg/acceptance/helpers/common.go
sfc-gh-asawicki: Okay , but why:
- do we need `ErrorCheck` versus `ExpectError`? Why can't we just use expect error too here?
- do we care about the ordering of errors at all then? Maybe one assertion not caring about the ordering would be enough?

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -232,21 +221,21 @@ func TestAcc_ScimIntegration_migrateFromVersion091(t *testing.T) {
 						Source:            "Snowflake-Labs/snowflake",
 					},
 				},
-				Config: scimIntegrationv091(id.Name(), role.Name),
+				Config: scimIntegrationv091(id.Name(), role.Name()),
```
sfc-gh-asawicki, line 224: names changed (we already have v0.92)

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -213,10 +204,8 @@ func TestAcc_ScimIntegration_InvalidIncomplete(t *testing.T) {
 }
 
 func TestAcc_ScimIntegration_migrateFromVersion091(t *testing.T) {
```
sfc-gh-asawicki, line 206: let's add one more test for migration (with setting enabled to false after and expecting changes) - it may be added in the next PR

sfc-gh-asawicki (approved):

sfc-gh-jmichalak:
Merging, unresolved comments will be addressed/discussed in the next PR.

PR #2868, opened 2024-06-13
feat: SAML2 integration v1 readiness

sfc-gh-jmichalak:
- Introduce `snowflake_saml2_resource` (to be consistent with docs and sql) and deprecate `snowflake_saml_resource`
- Fix migration guide
- Fix checking errors for invalid terraform configurations
- Added one more test for scim migrator
- Altered SDK:
    - improved enums
    - marked some fields as optional
    - introduced enums for some fields (they were not here before because of confusing docs, added to docs improvements list)
    - added new issues to objects lists

## Test Plan

* [x] acceptance tests

* [ ] …

## References

https://docs.snowflake.com/en/sql-reference/sql/create-security-integration-saml2


pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+
+	if d.HasChange("comment") {
+		if v := d.Get("comment").(string); len(v) > 0 {
+			set.WithComment(v)
+		} else {
+			unset.WithComment(true)
+		}
+	}
+	if !reflect.DeepEqual(*set, sdk.Saml2IntegrationSetRequest{}) {
```
sfc-gh-jcieslak, line 513: here you're doing DeepEqual and in the second one, simple equality check, is there any reason for this check to differ from the other one?

sfc-gh-asawicki: I guess that the reason are underlying lists, so equality operator cannot be used, right?

examples/resources/snowflake_saml2_integration/resource.tf
```diff
@@ -0,0 +1,29 @@
...
+  saml2_requested_nameid_format       = "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"
+  saml2_sign_request                  = true
+  saml2_snowflake_acs_url             = "example.snowflakecomputing.com/fed/login"
+  saml2_snowflake_issuer_url          = "example.snowflakecomputing.com/fed/login"
+  saml2_snowflake_x509_cert           = "MIICYzCCAcygAwIBAgIBADANBgkqhkiG9w0BAQUFADAuMQswCQYDVQQGEwJVUzEMMAoGA1UEChMDSUJNMREwDwYDVQQLEwhMb2NhbCBDQTAeFw05OTEyMjIwNTAwMDBaFw0wMDEyMjMwNDU5NTlaMC4xCzAJBgNVBAYTAlVTMQwwCgYDVQQKEwNJQk0xETAPBgNVBAsTCExvY2FsIENBMIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQD2bZEo7xGaX2/0GHkrNFZvlxBou9v1Jmt/PDiTMPve8r9FeJAQ0QdvFST/0JPQYD20rH0bimdDLgNdNynmyRoS2S/IInfpmf69iyc2G0TPyRvmHIiOZbdCd+YBHQi1adkj17NDcWj6S14tVurFX73zx0sNoMS79q3tuXKrDsxeuwIDAQABo4GQMIGNMEsGCVUdDwGG+EIBDQQ+EzxHZW5lcmF0ZWQgYnkgdGhlIFNlY3VyZVdheSBTZWN1cml0eSBTZXJ2ZXIgZm9yIE9TLzM5MCAoUkFDRikwDgYDVR0PAQH/BAQDAgAGMA8GA1UdEwEB/wQFMAMBAf8wHQYDVR0OBBYEFJ3+ocRyCTJw067dLSwr/nalx6YMMA0GCSqGSIb3DQEBBQUAA4GBAMaQzt+zaj1GU77yzlr8iiMBXgdQrwsZZWJo5exnAucJAEYQZmOfyLiMD6oYq+ZnfvM0n8G/Y79q8nhwvuxpYOnRSAXFp6xSkrIOeZtJMY1h00LKp/JX3Ng1svZ2agE126JHsQ0bhzN5TKsYfbwfTwfjdWAGy6Vf1nYi/rO+ryMO"
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I agree, but what about `The Base64 encoded IdP signing certificate on a single line without the leading -----BEGIN CERTIFICATE----- and ending -----END CERTIFICATE----- markers.` from the docs?

pkg/acceptance/helpers/common.go
```diff
@@ -81,3 +71,12 @@ func AssertErrorContainsPartsFunc(t *testing.T, parts []string) resource.ErrorCh
 		return nil
 	}
 }
+
+// EnumToTerraformStringList converts a list of enum values to the format represented in Terraform errors.
```
sfc-gh-asawicki, line 75: We already have:
```
func AsStringList[T ~string](input []T) []string {
	output := make([]string, len(input))
	for i, element := range input {
		output[i] = string(element)
	}
	return output
}
```
so please handle only the format part here.

sfc-gh-asawicki: Also, for me, it's a bit too much to extract it as function, because it only follows the error format in `func StringInSlice`, right?

pkg/resources/custom_diffs.go
```diff
@@ -67,6 +67,29 @@ func BoolComputedIf(key string, getDefault func(client *sdk.Client, id sdk.Accou
 	})
 }
 
+// SetEmptyForceNewIfChange sets a ForceNew for a list field which was set to an empty value.
+func SetEmptyForceNewIfChange[T any](key string) schema.CustomizeDiffFunc {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: And please adjust the name to state clearly that it is for lists

pkg/resources/custom_diffs.go
```diff
@@ -67,6 +67,29 @@ func BoolComputedIf(key string, getDefault func(client *sdk.Client, id sdk.Accou
...
+	return customdiff.ForceNewIfChange(key, func(ctx context.Context, oldValue, newValue, meta any) bool {
+		oldList, newList := oldValue.([]T), newValue.([]T)
+		return len(oldList) > 0 && len(newList) == 0
+	})
+}
+
+// SetEmptyForceNewIfChangeString sets a ForceNew for a string field which was set to an empty value.
+func SetEmptyForceNewIfChangeString(key string) schema.CustomizeDiffFunc {
```
sfc-gh-asawicki, line 79: the name is misleading

pkg/resources/custom_diffs.go
```diff
@@ -67,6 +67,29 @@ func BoolComputedIf(key string, getDefault func(client *sdk.Client, id sdk.Accou
...
+	return customdiff.ForceNewIfChange(key, func(ctx context.Context, oldValue, newValue, meta any) bool {
+		oldString, newString := oldValue.(string), newValue.(string)
+		return len(oldString) > 0 && len(newString) == 0
+	})
+}
+
+// SetForceNewIfNull sets a ForceNew for key missing in RawConfig.
+func SetForceNewIfNull(key string) schema.CustomizeDiffFunc {
+	return customdiff.ForceNewIf(key, func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) bool {
+		return d.GetRawConfig().AsValueMap()[key].IsNull()
```
sfc-gh-asawicki, line 89: hmmmm, I don't think this is good. It will trigger every time, right (even with no change: empty -> empty).

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+		},
+		Optional:    true,
+		Description: "A list of regular expressions that email addresses are matched against to authenticate with a SAML2 security integration.",
+	},
+	"comment": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies a comment for the integration.",
+	},
+	"created_on": {
```
sfc-gh-asawicki, line 149: wrap the SHOW outpu into show_output map please

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+
+	if !d.GetRawConfig().AsValueMap()["enabled"].IsNull() {
+		req.WithEnabled(d.Get("enabled").(bool))
+	}
+
+	if v, ok := d.GetOk("saml2_sp_initiated_login_page_label"); ok {
+		req.WithSaml2SpInitiatedLoginPageLabel(v.(string))
+	}
+
+	req.WithSaml2EnableSpInitiated(d.Get("saml2_enable_sp_initiated").(bool))
```
sfc-gh-asawicki, line 197: it is optional, so why we don't check it? (the same with some others)

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+)
+
+var saml2IntegrationSchema = map[string]*schema.Schema{
+	"name": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: "Specifies the name of the SAML2 integration. This name follows the rules for Object Identifiers. The name should be unique among security integrations in your account.",
+	},
+	"enabled": {
```
sfc-gh-asawicki, line 28: why it is optional?

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+
+	return nil
+}
+
+func UpdateContextSAMLIntegration(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
+	set, unset := sdk.NewSaml2IntegrationSetRequest(), sdk.NewSaml2IntegrationUnsetRequest()
+
+	if d.HasChange("enabled") {
```
sfc-gh-asawicki, line 409: why not unset? (if this is the buggy one, we have to be explicit with it)

And this is for all others too.

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+		set.WithEnabled(d.Get("enabled").(bool))
+	}
+
+	if d.HasChange("saml2_issuer") {
```
sfc-gh-asawicki, line 413: is empty value correct here? (the same with few others)

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -164,14 +164,14 @@ func TestAcc_ScimIntegration_invalid(t *testing.T) {
 		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
 			tfversion.RequireAbove(tfversion.Version1_5_0),
 		},
+		ErrorCheck: helpers.AssertErrorContainsPartsFunc(t, []string{
```
sfc-gh-asawicki, line 167: hmm, the usual way we follow is using ExpectError, not the error check (I think I have already written it in the previous PR)

pkg/sdk/security_integrations_def.go
```diff
@@ -48,13 +48,47 @@ const (
...
 
+var AllOauthSecurityIntegrationUseSecondaryRoles = []OauthSecurityIntegrationUseSecondaryRolesOption{
+	OauthSecurityIntegrationUseSecondaryRolesImplicit,
+	OauthSecurityIntegrationUseSecondaryRolesNone,
+}
+
+func ToOauthSecurityIntegrationUseSecondaryRolesOption(s string) (OauthSecurityIntegrationUseSecondaryRolesOption, error) {
+	s = strings.ToUpper(s)
+	switch s {
+	case "IMPLICIT":
```
sfc-gh-asawicki, line 59: you can use the values as switch cases like: `string(OauthSecurityIntegrationUseSecondaryRolesImplicit)`

pkg/sdk/security_integrations_def.go
```diff
@@ -63,6 +97,97 @@ const (
...
+		return Saml2SecurityIntegrationSaml2ProviderCustom, nil
+	default:
+		return "", fmt.Errorf("invalid Saml2SecurityIntegrationSaml2ProviderOption: %s", s)
+	}
+}
+
+type Saml2SecurityIntegrationSaml2RequestedNameidFormatOption string
+
+const (
+	Saml2SecurityIntegrationSaml2RequestedNameidFormatUnspecified                Saml2SecurityIntegrationSaml2RequestedNameidFormatOption = "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"
```
sfc-gh-asawicki, line 151: wwow, do we really need those?

pkg/resources/testdata/TestAcc_Saml2Integration/complete/test.tf
```diff
@@ -0,0 +1,21 @@
...
+  saml2_enable_sp_initiated      = var.saml2_enable_sp_initiated
+  saml2_force_authn              = var.saml2_force_authn
+  saml2_issuer                   = var.saml2_issuer
+  saml2_post_logout_redirect_url = var.saml2_post_logout_redirect_url
+  saml2_provider                 = var.saml2_provider
+  saml2_requested_nameid_format  = var.saml2_requested_nameid_format
+  saml2_sign_request             = var.saml2_sign_request
+  saml2_snowflake_acs_url        = var.saml2_snowflake_acs_url
+  saml2_snowflake_issuer_url     = var.saml2_snowflake_issuer_url
+  # TODO(SNOW-1479617): set saml2_snowflake_x509_cert
```
sfc-gh-asawicki, line 16: let's leave such TODOs not in the tf file, but in the test/implementation (like we do with others)

pkg/resources/saml2_integration_acceptance_test.go
```diff
@@ -0,0 +1,292 @@
+package resources_test
+
```
sfc-gh-asawicki, line 2: there are no tests for external changes behavior (with/without defaults), there are no tests with zero values. You can take something from database/warehouse.

sfc-gh-asawicki (changes requested):
Gave a few comments but I am worried because the schema and handling in create/update are inconsistent with database/warehouse (and I think that some of the use cases won't work correctly with the current setup). Let's talk.

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+		Computed:    true,
+		Description: "Signature methods used for this integration.",
+	},
+	"allowed_user_domains": {
+		Type: schema.TypeList,
+		Elem: &schema.Schema{
+			Type: schema.TypeString,
+		},
+		Optional:    true,
+		Description: "A list of email domains that can authenticate with a SAML2 security integration.",
```
sfc-gh-jcieslak, line 134: Could you also describe conditional forceNew cases in the description to avoid confusion?

examples/resources/snowflake_saml2_integration/import.sh
```diff
@@ -0,0 +1 @@
+terraform import snowflake_saml2_integration.example name
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: This is just an example, but I see your point  - done.

pkg/resources/scim_integration_acceptance_test.go
```diff
@@ -253,14 +294,14 @@ resource "snowflake_scim_integration" "test" {
 	return fmt.Sprintf(s, name, sdk.ScimSecurityIntegrationScimClientGeneric, roleName)
 }
 
-func scimIntegrationv092(name, roleName string) string {
+func scimIntegrationv093(name, roleName string, enabled bool) string {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Because this was implemented before v0.92 release and the versions here weren't adjusted. In other places as well.

pkg/resources/custom_diffs.go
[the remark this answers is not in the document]

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+	},
+	"saml2_sp_initiated_login_page_label": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "The string containing the label to display after the Log In With button on the login page.",
+	},
+	"saml2_enable_sp_initiated": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Description: "The Boolean indicating if the Log In With button will be shown on the login page. TRUE: displays the Log in WIth button on the login page.  FALSE: does not display the Log in With button on the login page.",
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Fixed.

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,543 @@
...
+		case "SAML2_DIGEST_METHODS_USED":
+			if err := d.Set("saml2_digest_methods_used", value); err != nil {
+				return diag.FromErr(err)
+			}
+		case "SAML2_SIGNATURE_METHODS_USED":
+			if err := d.Set("saml2_signature_methods_used", value); err != nil {
+				return diag.FromErr(err)
+			}
+		case "ALLOWED_USER_DOMAINS":
+			value = strings.TrimLeft(value, "[")
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Wdym? This isn't a valid JSON, e.g. `[a.com, b.com]`.

pkg/resources/saml2_integration.go
sfc-gh-jmichalak: That's correct - DeepEqual is used because set structure contains lists. For consistency we could always use `DeepEqual`, wdyt?

pkg/acceptance/helpers/common.go
sfc-gh-jmichalak: Removed.

pkg/resources/custom_diffs.go
sfc-gh-jmichalak: Changed.

pkg/resources/custom_diffs.go
sfc-gh-jmichalak: This is not needed - removed.

pkg/resources/saml2_integration.go
sfc-gh-jmichalak: Fixed.

pkg/resources/saml2_integration.go
sfc-gh-jmichalak: Because you can create an integration without setting this field.

pkg/resources/saml2_integration.go
sfc-gh-jmichalak: No, but it's a required field, so it should never be empty, right?

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jmichalak: Okay, but I noticed that order of some errors is not deterministic, so it's better to check it here instead of using a regex.

pkg/sdk/security_integrations_def.go
sfc-gh-jmichalak: Done.

pkg/sdk/security_integrations_def.go
sfc-gh-jmichalak: Yes because you can't set this field to an arbitrary value - check [docs](https://docs.snowflake.com/en/sql-reference/sql/create-security-integration-saml2#optional-parameters).

pkg/resources/testdata/TestAcc_Saml2Integration/complete/test.tf
sfc-gh-jmichalak: Done.

pkg/resources/saml2_integration.go
sfc-gh-jmichalak: Done.

pkg/resources/saml2_integration.go
sfc-gh-jcieslak: Done

pkg/resources/saml2_integration.go
sfc-gh-asawicki: @sfc-gh-jcieslak what about this?

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-asawicki: So usually it's better to have a configuration for test which will result in just one error; you don't have the order problem then (check warekouse acceptance test)

pkg/resources/saml2_integration_acceptance_test.go
sfc-gh-asawicki: @sfc-gh-jcieslak

pkg/resources/saml2_integration.go
sfc-gh-jcieslak: Unset doesn't work so in the case of unset I set it to the Snowflake default value

pkg/resources/saml2_integration_acceptance_test.go
sfc-gh-jcieslak: missed that one somehow, I'll add some

pkg/resources/saml2_integration.go
```diff
@@ -180,54 +201,184 @@ func ImportSaml2Integration(ctx context.Context, d *schema.ResourceData, meta an
 
 	integration, err := client.SecurityIntegrations.ShowByID(ctx, id)
 	if err != nil {
-		if errors.Is(err, sdk.ErrObjectNotFound) {
```
sfc-gh-jmichalak: Why is it removed?

pkg/resources/saml2_integration.go
```diff
@@ -241,21 +392,28 @@ func CreateContextSAML2Integration(ctx context.Context, d *schema.ResourceData,
...
 
-	if v, ok := d.GetOk("saml2_snowflake_x509_cert"); ok {
-		req.WithSaml2SnowflakeX509Cert(v.(string))
+	if v := d.Get("saml2_snowflake_x509_cert").(string); v != SnowflakeDefaultStringValuePlaceholder {
+		req.WithSaml2SnowflakeX509Cert(v)
 	}
 
-	if !d.GetRawConfig().AsValueMap()["saml2_sign_request"].IsNull() {
-		req.WithSaml2SignRequest(d.Get("saml2_sign_request").(bool))
+	if v := d.Get("saml2_sign_request").(string); v != "unknown" {
```
sfc-gh-jmichalak, line 405: Why is it sometimes "unknown" and sometimes SnowflakeDefaultStringValuePlaceholder? What's the difference?

pkg/resources/saml2_integration.go
```diff
@@ -413,20 +765,36 @@ func UpdateContextSAML2Integration(ctx context.Context, d *schema.ResourceData,
...
+				return diag.FromErr(err)
+			}
+			set.WithSaml2EnableSpInitiated(parsed)
+		} else {
+			set.WithSaml2EnableSpInitiated(false)
+		}
 	}
 
 	if d.HasChange("saml2_snowflake_x509_cert") {
+		// TODO: ForceNew on empty set
```
sfc-gh-jmichalak: Please link to SNOW-1515781

pkg/schemas/saml_security_integration.go
```diff
@@ -0,0 +1,53 @@
+package schemas
+
+import (
```
sfc-gh-jmichalak, line 3: Nit: file should be named `saml2_security_integration.go`

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,891 @@
...
+				describeMapping{"saml2_snowflake_x509_cert", "saml2_snowflake_x509_cert", snowflakeX509Cert.Value, snowflakeX509Cert.Value, nil},
+			); err != nil {
+				return diag.FromErr(err)
+			}
+		}
+
+		// These are all identity sets, needed for the case where:
+		// - previous config was empty (therefore Snowflake defaults had been used)
+		// - new config have the same values that are already in SF
+		if !d.GetRawConfig().IsNull() {
```
sfc-gh-asawicki, line 657: this can be also made a bit nicer after #2900

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,891 @@
...
+			}
+			set.WithSaml2EnableSpInitiated(parsed)
+		} else {
+			// UNSET is not implemented
+			set.WithSaml2EnableSpInitiated(false)
+		}
+	}
+
+	if d.HasChange("saml2_snowflake_x509_cert") {
+		// UNSET is not implemented (conditional ForceNew on unset)
```
sfc-gh-asawicki, line 770: it would be nice to have one issue for all unsets not working for security integrations (so we have context and so that we can come back)

sfc-gh-asawicki (changes requested):

pkg/resources/saml2_integration.go
sfc-gh-jcieslak: Hmm, isn't it that if we throw the error here it won't import this object (and the setId("") is obsolete)?

pkg/schemas/saml_security_integration.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func DescribeSaml2IntegrationToSchema(props []sdk.SecurityIntegrationProperty) map[string]any {
+	propsSchema := make(map[string]any)
+	for _, property := range props {
+		switch property.Name {
+		case "ENABLED", "COMMENT", "SAML2_ISSUER", "SAML2_SSO_URL", "SAML2_PROVIDER", "SAML2_X509_CERT", "SAML2_SP_INITIATED_LOGIN_PAGE_LABEL", "SAML2_SNOWFLAKE_X509_CERT",
+			"SAML2_REQUESTED_NAMEID_FORMAT", "SAML2_POST_LOGOUT_REDIRECT_URL", "SAML2_SNOWFLAKE_ISSUER_URL", "SAML2_SNOWFLAKE_ACS_URL", "SAML2_SNOWFLAKE_METADATA",
+			"SAML2_DIGEST_METHODS_USED", "SAML2_SIGNATURE_METHODS_USED", "SAML2_ENABLE_SP_INITIATED", "SAML2_SIGN_REQUEST", "SAML2_FORCE_AUTHN", "ALLOWED_USER_DOMAINS", "ALLOWED_EMAIL_PATTERNS":
+			propsSchema[strings.ToLower(property.Name)] = []map[string]any{
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yeah, I wasn't rebased, so I didn't saw it ;_;

pkg/resources/helpers.go
```diff
@@ -141,55 +141,25 @@ func GetPropertyAsPointer[T any](d *schema.ResourceData, property string) *T {
...
-
-	typedNestedValue, ok := typedNestedMap[nestedValueKey].(T)
-	if !ok {
-		return nil, fmt.Errorf("nested property %s.%s is not of type %T, got: %T", propertyKey, nestedValueKey, *new(T), typedNestedMap[nestedValueKey])
+// ParseCommaSeparatedStringArray can be used to parse Snowflake output containing a list in the format of "[item1, item2, ...]",
+// the assumptions are that:
+// 1. The list is enclosed by [] brackets, and they shouldn't be a part of any item's value
+// 2. Items are separated by commas, and they shouldn't be a part of any item's value
+// 3. Items can have as many spaces in between, but after separation they will be trimmed and shouldn't be a part of any item's value
+func ParseCommaSeparatedStringArray(value string) []string {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yeah, I was thinking about it, but ultimately just left it here. I'll move it to sdk/parsers.go

pkg/resources/saml2_integration.go
```diff
@@ -0,0 +1,891 @@
...
+		Type:        schema.TypeString,
+		Required:    true,
+		Description: "The string containing the IdP SSO URL, where the user should be redirected by Snowflake (the Service Provider) with a SAML AuthnRequest message.",
+	},
+	"saml2_provider": {
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      fmt.Sprintf("The string describing the IdP. Valid options are: %v.", sdk.AllSaml2SecurityIntegrationSaml2Providers),
+		ValidateDiagFunc: sdkValidation(sdk.ToSaml2SecurityIntegrationSaml2ProviderOption),
+		DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, I don't think we need it.
cc: @sfc-gh-jmichalak (thoughts?)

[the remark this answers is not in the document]
sfc-gh-jmichalak: Well, I initially left it to be compatible. Potentially we can delete `strings.ReplaceAll(s, "-", "")` (breaking change) and just ignore string case.

pkg/resources/saml2_integration.go
sfc-gh-jcieslak: Added todos

pkg/resources/scim_integration_acceptance_test.go
sfc-gh-jcieslak: I'll split this into two

sfc-gh-jmichalak (commented):
lgtm

pkg/resources/saml2_integration.go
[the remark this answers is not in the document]
sfc-gh-asawicki: if this is new object and this is not needed now (so we have no backward compatibility reasons), let's just remove it

sfc-gh-asawicki (dismissed):

pkg/resources/custom_diffs_test.go
```diff
@@ -301,8 +301,8 @@ func TestForceNewIfChangeToEmptySet(t *testing.T) {
 		}, {
 			name: "non-empty to empty",
 			stateValue: map[string]string{
-				"value.#": "1",
-				"value.0": "foo",
+				"value.#":          "1",
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Addressed in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2912

pkg/resources/saml2_integration.go
sfc-gh-jcieslak: Addressed in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2912

Issue #2236, opened 2023-12-06
snowflake_table_column_masking_policy_application keeps getting recreated at every plan


patcky:
### Terraform CLI and Provider Versions

Terraform version: 1.2
Provider version: was initially using 0.67.0 but also tested with the most recent version (0.77.0 atm) to see if the bug was fixed, but it remained the same.

### Terraform Configuration

```terraform
# _database/_modules/_masking_policy/masking_policy.tf

resource "snowflake_masking_policy" "policy" {
  name     = "MY_APP_${upper(var.schema)}_STRING_MASK"
  database = var.database
  schema   = var.schema
  signature {
    column {
      name = "VAL"
      type = "VARCHAR"
    }
  }
  return_data_type = "VARCHAR(16777216)"
  masking_expression = trimspace(<<-EOT
  case
    when current_role() in (
      '${join("',\n    '", var.sensitive_select_roles)}'
    ) then val
    else sha2(val)) # This was changed but it doesn't matter because it's working properly
  end
  EOT
  )
}

resource "snowflake_masking_policy_grant" "ownership" {
  database_name       = var.database
  schema_name         = var.schema
  masking_policy_name = snowflake_masking_policy.policy.name
  privilege           = "OWNERSHIP"
  roles               = var.masking_admins
}

output "masking_policy" {
  value = snowflake_masking_policy.policy.qualified_name
}

# _database/schema/schema.tf

resource "snowflake_schema" "my_schema" {
  database     = var.database_name
  name         = "MY_SCHEMA"
  comment      = "Schema for xxx"
  is_transient = var.transient
}

...

module "masking_policies" {
  source                 = "../../_modules/_masking_policy"
  database               = snowflake_schema.my_schema.database
  schema                 = snowflake_schema.my_schema.name
  masking_admins         = var.masking_admins
  sensitive_select_roles = var.sensitive_select_roles
}

...

module "my_app_tables" {
  source         = "../../_modules/_table"
  database       = snowflake_schema.games.database
  schema         = snowflake_schema.games.name
  masking_policy = module.masking_policies.masking_policy
  
  # This is to create a table for each namespace passed to the schema
  for_each = {
    for n in var.namespaces : n.namespace => n
  }

  namespace      = each.value.namespace
  selector_roles = each.value.selector_roles
}

# _database/_modules/_table/table.tf

resource "snowflake_table" "my_table" {
  database = local.database
  schema   = local.schema
  name     = local.namespace

[...]

PR #2930, opened 2024-07-11
feat: Streamlit v1 readiness

sfc-gh-jmichalak:
This PR adds streamlits datasource and streamlit resource. SDK is adjusted. 

## Test Plan

* [x] acceptance tests

## References

https://docs.snowflake.com/en/sql-reference/sql/create-streamlit


pkg/datasources/streamlits.go
```diff
@@ -0,0 +1,106 @@
...
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     true,
+		Description: "Runs DESC STREAMLIT for each streamlit returned by SHOW STREAMLITS. The output of describe is saved to the description field. By default this value is set to true.",
+	},
+	"like": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Filters the output with **case-insensitive** pattern, with support for SQL wildcard characters (`%` and `_`).",
+	},
```
sfc-gh-jcieslak, line 25: What about `in` and `limit from` filtering options?

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_Streamlits(t *testing.T) {
+	databaseId := acc.TestClient().Ids.DatabaseId()
+	schemaId := acc.TestClient().Ids.SchemaId()
+	id := acc.TestClient().Ids.RandomSchemaObjectIdentifierInSchema(schemaId)
+	ctx := context.Background()
+	acc.TestAccPreCheck(t)
```
sfc-gh-jcieslak, line 26: this should be in the resource.Test (check other tests);

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,189 @@
...
+	t.Cleanup(stageCleanup)
+	err := acc.Client(t).Sessions.UseSchema(ctx, schemaId)
+	require.NoError(t, err)
+	// warehouse is needed because default warehouse uses lowercase, and it fails in snowflake.
+	warehouse, warehouseCleanup := acc.TestClient().Warehouse.CreateWarehouse(t)
+	t.Cleanup(warehouseCleanup)
+	integrationId := acc.TestClient().Ids.RandomSchemaObjectIdentifier()
+	networkRule, networkRuleCleanup := acc.TestClient().NetworkRule.CreateNetworkRule(t)
+	t.Cleanup(networkRuleCleanup)
+	_, err = acc.Client(t).ExecForTests(ctx, fmt.Sprintf(`CREATE EXTERNAL ACCESS INTEGRATION %s ALLOWED_NETWORK_RULES = (%s) ENABLED = TRUE`, integrationId.Name(), networkRule.ID().Name()))
```
sfc-gh-jcieslak, line 37: maybe better to extract it into TestClient rather than copy paste in multiple tests

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,189 @@
...
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		_, err = acc.Client(t).ExecForTests(ctx, fmt.Sprintf(`DROP EXTERNAL ACCESS INTEGRATION %s`, integrationId.Name()))
```
sfc-gh-jcieslak, line 40: drop if exists

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_Streamlits(t *testing.T) {
+	databaseId := acc.TestClient().Ids.DatabaseId()
+	schemaId := acc.TestClient().Ids.SchemaId()
+	id := acc.TestClient().Ids.RandomSchemaObjectIdentifierInSchema(schemaId)
```
sfc-gh-jcieslak, line 24: Use Ids. RandomSchemaObjectIdentifier()

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,189 @@
...
+	ctx := context.Background()
+	acc.TestAccPreCheck(t)
+	stage, stageCleanup := acc.TestClient().Stage.CreateStageInSchema(t, schemaId)
+	t.Cleanup(stageCleanup)
+	err := acc.Client(t).Sessions.UseSchema(ctx, schemaId)
```
sfc-gh-jcieslak, line 29: Is this needed?

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,189 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("data.snowflake_streamlits.test", "streamlits.#", "1"),
+
+					resource.TestCheckResourceAttr("data.snowflake_streamlits.test", "streamlits.0.show_output.#", "1"),
+					resource.TestCheckResourceAttrSet("data.snowflake_streamlits.test", "streamlits.0.show_output.0.created_on"),
+					resource.TestCheckResourceAttr("data.snowflake_streamlits.test", "streamlits.0.show_output.0.name", id.Name()),
+					resource.TestCheckResourceAttr("data.snowflake_streamlits.test", "streamlits.0.show_output.0.database_name", databaseId.Name()),
+					resource.TestCheckResourceAttr("data.snowflake_streamlits.test", "streamlits.0.show_output.0.schema_name", schemaId.Name()),
+					resource.TestCheckResourceAttr("data.snowflake_streamlits.test", "streamlits.0.show_output.0.title", "foo"),
+					resource.TestCheckResourceAttr("data.snowflake_streamlits.test", "streamlits.0.show_output.0.owner", snowflakeroles.Accountadmin.Name()),
```
sfc-gh-jcieslak, line 76: nit: maybe use current role instead

pkg/helpers/helpers.go
```diff
@@ -160,3 +161,21 @@ func ConcatSlices[T any](slices ...[]T) []T {
...
+	})
+	if err != nil {
+		return sdk.SchemaObjectIdentifier{}, "", err
+	}
+	if len(parts) < 3 {
+		return sdk.SchemaObjectIdentifier{}, "", fmt.Errorf("expected 3 parts, got ")
+	}
+	parts[2] = strings.Join(parts[2:], ".")
+	lastParts := strings.Split(parts[2], "/")
+	return sdk.NewSchemaObjectIdentifier(parts[0], parts[1], lastParts[0]), strings.Join(lastParts[1:], "/"), nil
```
sfc-gh-jcieslak, line 180: I kind of don't get this part fully (please add tests)

pkg/resources/streamlit.go
```diff
@@ -0,0 +1,408 @@
...
+var streamlitSchema = map[string]*schema.Schema{
+	"name": {
+		Type:        schema.TypeString,
+		Required:    true,
+		Description: "String that specifies the identifier (i.e. name) for the streamlit; must be unique in your account.",
+	},
+	"database": {
+		Type:        schema.TypeString,
+		Required:    true,
+		Description: "The database in which to create the Cortex search service.",
```
sfc-gh-jcieslak, line 27: fix description

pkg/resources/streamlit.go
```diff
@@ -0,0 +1,408 @@
...
+		Type:             schema.TypeString,
+		ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
+		Optional:         true,
+		Description:      "Specifies the warehouse where SQL queries issued by the Streamlit application are run.",
+		DiffSuppressFunc: SuppressIfAny(suppressIdentifierQuoting, IgnoreChangeToCurrentSnowflakeValueInOutput(ShowOutputAttributeName, "query_warehouse")),
+	},
+	"external_access_integrations": {
+		Type: schema.TypeSet,
+		Elem: &schema.Schema{
+			Type: schema.TypeString,
```
sfc-gh-jcieslak, line 65: You can add ValidateDiagFunc: IsValidIdentifier here

pkg/resources/streamlit.go
```diff
@@ -0,0 +1,408 @@
...
+		},
+		Optional:         true,
```
sfc-gh-jcieslak, line 67: This should have diff suppression added in network policy to avoid infinite plans that can occur with quoting differences like in (https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2236#issuecomment-2220003509).

pkg/resources/streamlit.go
```diff
@@ -0,0 +1,408 @@
...
+		if err := d.Set("stage", stageId.FullyQualifiedName()); err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set("directory_location", location); err != nil {
+			return diag.FromErr(err)
+		}
+		if err := d.Set("main_file", streamlitDetails.MainFile); err != nil {
+			return diag.FromErr(err)
+		}
+		if withExternalChangesMarking {
```
sfc-gh-jcieslak, line 265: I'm not sure, but this resource looks like it could be done without this `if` and additional stuff (just set everything in read). I had similar case in Roles and NetworkPolicies. You can check them and decide (It would simplify the code).

pkg/resources/streamlit.go
```diff
@@ -0,0 +1,408 @@
...
+func CreateContextStreamlit(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+	databaseName := d.Get("database").(string)
+	schemaName := d.Get("schema").(string)
+	name := d.Get("name").(string)
+	id := sdk.NewSchemaObjectIdentifier(databaseName, schemaName, name)
+	stageId := sdk.NewSchemaObjectIdentifierFromFullyQualifiedName(d.Get("stage").(string))
+	rootLocation := fmt.Sprintf("@%s", stageId.FullyQualifiedName())
+	if v, ok := d.GetOk("directory_location"); ok {
+		rootLocation = fmt.Sprintf("%s/%s", rootLocation, v)
```
sfc-gh-jcieslak, line 179: would it better to use `path.Join()`?

pkg/resources/streamlit_acceptance_test.go
```diff
@@ -0,0 +1,428 @@
...
+					resource.TestCheckResourceAttr("snowflake_streamlit.test", "stage", stage.ID().FullyQualifiedName()),
+					resource.TestCheckResourceAttr("snowflake_streamlit.test", "directory_location", ""),
+					resource.TestCheckResourceAttr("snowflake_streamlit.test", "main_file", "foo")),
+			},
+		},
+	})
+}
+
+func TestAcc_Streamlit_complete(t *testing.T) {
+	databaseId := acc.TestClient().Ids.DatabaseId()
```
sfc-gh-jcieslak, line 251: This setup is repeated throught the tests, but its cluttered. Maybe it could be cleaned up or at least organized in more redable way (?)

pkg/resources/streamlit_acceptance_test.go
```diff
@@ -0,0 +1,428 @@
...
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: acc.CheckDestroy(t, resources.NetworkPolicy),
+		Steps: []resource.TestStep{
+			{
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_Streamlit/basic"),
+				ConfigVariables: m(id.Name()),
```
sfc-gh-jcieslak, line 374: Could you add also a change of one param at the same time (e.g. comment) to show the rename is not influencing other alters.

pkg/resources/diff_suppressions.go
```diff
@@ -30,17 +30,17 @@ func IgnoreAfterCreation(_, _, _ string, d *schema.ResourceData) bool {
 	return d.Id() != ""
 }
 
-func IgnoreChangeToCurrentSnowflakeValueInShow(keyInShowOutput string) schema.SchemaDiffSuppressFunc {
+func IgnoreChangeToCurrentSnowflakeValueInOutput(attrName, keyInShowOutput string) schema.SchemaDiffSuppressFunc {
```
sfc-gh-jcieslak, line 33: I think this and the function below should be renamed. This one is used for plain values and the function below for "describe properties" (could be also used for parameters I guess), so the names should be something like Ignore...Plain[or Flat]ValueInOutput and the function below Ignore...ParameterValueInOutput or something similar (names can be changed).

pkg/schemas/streamlit.go
```diff
@@ -0,0 +1,84 @@
...
+		Type:     schema.TypeString,
+		Computed: true,
+	},
+	"user_packages": {
+		Type:     schema.TypeList,
+		Elem:     &schema.Schema{Type: schema.TypeString},
+		Computed: true,
+	},
+	"import_urls": {
+		Type:     schema.TypeList,
```
sfc-gh-jcieslak, line 47: Are those (this field and the one above) fields Lists or should be Sets?

sfc-gh-jcieslak (changes requested):

pkg/datasources/streamlits.go
sfc-gh-jmichalak: Added.

pkg/datasources/streamlits_acceptance_test.go
sfc-gh-jmichalak: This needs to be called earlier because there are resources being created in the provided schema.

pkg/datasources/streamlits_acceptance_test.go
sfc-gh-jmichalak: Done. I wasn't sure if we want to do this for resources missing from SDK.

pkg/datasources/streamlits_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/datasources/streamlits_acceptance_test.go
sfc-gh-jmichalak: Yes, otherwise reference to external access integration fails for some reason. Added a comment.

pkg/resources/streamlit.go
sfc-gh-jmichalak: Done.

pkg/resources/streamlit.go
sfc-gh-jmichalak: Done.

pkg/resources/streamlit_acceptance_test.go
sfc-gh-jmichalak: Cleaned up a bit.

pkg/resources/streamlit_acceptance_test.go
sfc-gh-jmichalak: Done with `main_file`.

pkg/schemas/streamlit.go
sfc-gh-jmichalak: Changed to sets.

pkg/resources/diff_suppressions.go
sfc-gh-jmichalak: Changed the first one to suggested by you, but changed the second one to `IgnoreChangeToCurrentSnowflakeListValueInDescribe` because it's used when field is `DescribePropertyListSchema`.

pkg/resources/streamlit.go
sfc-gh-jmichalak: Done.

pkg/helpers/helpers.go
sfc-gh-jmichalak: Added.

MIGRATION_GUIDE.md
```diff
@@ -21,6 +21,35 @@ Changes:
...
+- added `show_output` field that holds the response from SHOW ROLES. Remember that the field will be only recomputed if one of the fields (`name` or `comment`) are changed.
+
+### *(breaking change)* refactored snowflake_roles data source
+
+Changes:
+- New `in_class` filtering option to filter out roles by class name, e.g. `in_class = "SNOWFLAKE.CORE.BUDGET"`
+- `pattern` was renamed to `like`
+- output of SHOW is enclosed in `show_output`, so before, e.g. `roles.0.comment` is now `roles.0.show_output.0.comment`
+
+## v0.93.0 ➞ v0.94.0
```
sfc-gh-asawicki, line 39: there is already v0.93.0 ➞ v0.94.0 above, also, snowflake account role seems to be listed twice (apparently merge was unsuccessful)

docs/data-sources/streamlits.md
```diff
@@ -0,0 +1,177 @@
+---
+page_title: "snowflake_streamlits Data Source - terraform-provider-snowflake"
+subcategory: ""
+description: |-
+  Datasource used to get details of filtered streamlits. Filtering is aligned with the current possibilities for SHOW STREAMLITS https://docs.snowflake.com/en/sql-reference/sql/show-streamlits query (only like is supported). The results of SHOW and DESCRIBE are encapsulated in one output collection streamlits.
+---
+
+!> **V1 release candidate** This resource was reworked and is a release candidate for the V1. We do not expect significant changes in it before the V1. We will welcome any feedback and adjust the resource if needed. Any errors reported will be resolved with a higher priority. We encourage checking this resource out before the V1 release. Please follow the [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v0920--v0930) to use it.
```
sfc-gh-asawicki, line 8: migration guide should be updated to 0.93-0.94

docs/resources/streamlit.md
```diff
@@ -0,0 +1,106 @@
+---
+page_title: "snowflake_streamlit Resource - terraform-provider-snowflake"
+subcategory: ""
+description: |-
+  Resource used to manage streamlits objects. For more information, check streamlit documentation https://docs.snowflake.com/en/sql-reference/commands-streamlit.
+---
+
+!> **V1 release candidate** This resource was reworked and is a release candidate for the V1. We do not expect significant changes in it before the V1. We will welcome any feedback and adjust the resource if needed. Any errors reported will be resolved with a higher priority. We encourage checking this resource out before the V1 release. Please follow the [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v0920--v0930) to use it.
```
sfc-gh-asawicki, line 8: migration guide should be updated to 0.93-0.94

docs/resources/streamlit.md
```diff
@@ -0,0 +1,106 @@
...
+## Example Usage
+
+```terraform
+# basic resource
+resource "snowflake_streamlit" "streamlit" {
+  database  = "database"
+  schema    = "schema"
+  name      = "streamlit"
+  stage     = "streamlit_db.streamlit_schema.streamlit_stage"
+  main_file = "/streamlit_main.py"
```
sfc-gh-asawicki, line 23: do we always have to start the path with a `/` ? what will happen if this leading slash is not provided? Do we fail? Should we add something in the docs? or should we add it if absent (and suppress a change between with and without)?

pkg/acceptance/helpers/external_access_integration_client.go
```diff
@@ -0,0 +1,46 @@
...
+		context: context,
+		ids:     idsGenerator,
+	}
+}
+
+func (c *ExternalAccessIntegrationClient) client() *sdk.Client {
+	return c.context.client
+}
+
+func (c *ExternalAccessIntegrationClient) CreateExternalAccessIntegration(t *testing.T, networkRuleId sdk.SchemaObjectIdentifier) (sdk.SchemaObjectIdentifier, func()) {
```
sfc-gh-asawicki, line 28: nit: In cases like this, when we "mock" the implementation, please leave a TODO with a proper issue number to replace it in the future. In this case SNOW-1325215.

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_Streamlits(t *testing.T) {
+	databaseId := acc.TestClient().Ids.DatabaseId()
+	schemaId := acc.TestClient().Ids.SchemaId()
+	id := acc.TestClient().Ids.RandomSchemaObjectIdentifier()
+	acc.TestAccPreCheck(t)
+	stage, stageCleanup := acc.TestClient().Stage.CreateStageInSchema(t, schemaId)
```
sfc-gh-asawicki, line 26: if you use the default schema, you don't have to specify it, just use `CreateStage`

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,238 @@
...
+	t.Cleanup(stageCleanup)
+	// use schema is needed because otherwise reference to external access integration fails.
+	err := acc.Client(t).Sessions.UseSchema(context.Background(), schemaId)
```
sfc-gh-asawicki, line 29: it should not be needed because we create it, so it is used by default.

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,238 @@
...
+	require.NoError(t, err)
+	// warehouse is needed because default warehouse uses lowercase, and it fails in snowflake.
```
sfc-gh-asawicki, line 31: is there an sdk integration test showing the faulty behavior? Are we sure that this is not the issue with the SDK implementation? If we are sure, was it raised with the proper team?

pkg/datasources/streamlits_acceptance_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+func TestAcc_Streamlits_Filtering(t *testing.T) {
+	prefix := random.AlphaN(4)
+	idOne := acc.TestClient().Ids.RandomAccountObjectIdentifierWithPrefix(prefix)
+	idTwo := acc.TestClient().Ids.RandomAccountObjectIdentifierWithPrefix(prefix)
+	idThree := acc.TestClient().Ids.RandomAccountObjectIdentifier()
+	databaseId := acc.TestClient().Ids.DatabaseId()
+	schemaId := acc.TestClient().Ids.SchemaId()
+	acc.TestAccPreCheck(t)
+	stage, stageCleanup := acc.TestClient().Stage.CreateStageInSchema(t, schemaId)
```
sfc-gh-asawicki, line 124: same as above

pkg/datasources/streamlits_acceptance_test.go
sfc-gh-asawicki: yeah, this is messy, but we know that it needs a better setup (like with integration tests); let's leave it like this for now

pkg/helpers/helpers.go
```diff
@@ -160,3 +161,21 @@ func ConcatSlices[T any](slices ...[]T) []T {
 	}
 	return tmp
 }
+
+// TODO(SNOW-1058419): address during identifiers rework
```
sfc-gh-asawicki, line 165: I think this is not the right issue to be linked here.

pkg/helpers/helpers_test.go
```diff
@@ -221,3 +222,75 @@ func Test_DecodeSnowflakeAccountIdentifier(t *testing.T) {
 		require.ErrorContains(t, err, fmt.Sprintf("unable to read identifier: %s", id))
 	})
 }
+
+func TestParseRootLocation(t *testing.T) {
+	tests := []struct {
+		name     string
+		location string
+		id       string
```
sfc-gh-asawicki, line 230: id -> expectedId
path -> expectedPath
(we should have a clear separation between the setup attrs and the expected output)

pkg/helpers/helpers_test.go
```diff
@@ -221,3 +222,75 @@ func Test_DecodeSnowflakeAccountIdentifier(t *testing.T) {
...
+		{
+			name:     "invalid location",
+			location: `@foo`,
+			wantErr:  true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotId, gotPath, err := ParseRootLocation(tt.location)
+			if (err != nil) != tt.wantErr {
```
sfc-gh-asawicki, line 288: why mix t.Errorf with asserts below? How about:
```
if tt.wantErr {
 assert.Error(t, err)
 // optionally also
 assert.ErrorContains(...)
} else {
  assert.NoError()
  assert.Equal(t, tt.id, gotId.FullyQualifiedName())
  assert.Equal(t, tt.path, gotPath)
}
```

pkg/helpers/helpers_test.go
```diff
@@ -221,3 +222,75 @@ func Test_DecodeSnowflakeAccountIdentifier(t *testing.T) {
...
 	})
 }
+
+func TestParseRootLocation(t *testing.T) {
+	tests := []struct {
+		name     string
+		location string
+		id       string
+		path     string
+		wantErr  bool
```
sfc-gh-asawicki, line 232: do we want to maybe expect a specific errors for different caseS?

pkg/helpers/helpers_test.go
```diff
@@ -221,3 +222,75 @@ func Test_DecodeSnowflakeAccountIdentifier(t *testing.T) {
...
+		},
+		{
+			name:     "unquoted with path with dots",
+			location: `@a.b.c/foo.d`,
+			id:       `"a"."b"."c"`,
+			path:     `foo.d`,
+		},
+		{
+			name:     "quoted with path with dots",
+			location: `@"a"."b"."c"/foo.d`,
```
sfc-gh-asawicki, line 275: maybe add tests with more slashes, there is a logic around them

pkg/resources/api_authentication_integration_common.go
```diff
@@ -42,7 +42,7 @@ var apiAuthCommonSchema = map[string]*schema.Schema{
 		Type:             schema.TypeString,
 		Optional:         true,
 		ValidateDiagFunc: sdkValidation(sdk.ToApiAuthenticationSecurityIntegrationOauthClientAuthMethodOption),
-		DiffSuppressFunc: SuppressIfAny(NormalizeAndCompare(sdk.ToApiAuthenticationSecurityIntegrationOauthClientAuthMethodOption), IgnoreChangeToCurrentSnowflakeValueInDescribe("oauth_client_auth_method")),
+		DiffSuppressFunc: SuppressIfAny(NormalizeAndCompare(sdk.ToApiAuthenticationSecurityIntegrationOauthClientAuthMethodOption), IgnoreChangeToCurrentSnowflakeListValueInDescribe("oauth_client_auth_method")),
```
sfc-gh-asawicki, line 45: why these changes? these are not list fields

pkg/resources/diff_suppressions.go
```diff
@@ -30,18 +30,18 @@ func IgnoreAfterCreation(_, _, _ string, d *schema.ResourceData) bool {
 	return d.Id() != ""
 }
 
-func IgnoreChangeToCurrentSnowflakeValueInShow(keyInShowOutput string) schema.SchemaDiffSuppressFunc {
+func IgnoreChangeToCurrentSnowflakePlainValueInOutput(attrName, keyInOutput string) schema.SchemaDiffSuppressFunc {
```
sfc-gh-asawicki, line 33: Instead of renmae and changing every invocation, please extract a common part, so that, usually, we can just call a named func, which better explains its purpose.

pkg/resources/streamlit.go
```diff
@@ -0,0 +1,403 @@
...
+		Type:             schema.TypeString,
+		Optional:         true,
+		Description:      "Specifies the full path to the named stage containing the Streamlit Python files, media files, and the environment.yml file.",
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakePlainValueInOutput(DescribeOutputAttributeName, "root_location"),
+	},
+	"main_file": {
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      "Specifies the filename of the Streamlit Python application. This filename is relative to the value of `root_location`",
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakePlainValueInOutput(DescribeOutputAttributeName, "main_file"),
```
sfc-gh-asawicki, line 54: -> IgnoreChangeToCurrentSnowflakeValueInDescribe

pkg/resources/streamlit.go
```diff
@@ -0,0 +1,403 @@
...
+						Detail:   fmt.Sprintf("Streamlit name: %s, Err: %s", id.FullyQualifiedName(), err),
+					},
+				}
+			}
+			return diag.FromErr(err)
+		}
+
+		streamlitDetails, err := client.Streamlits.Describe(ctx, id)
+		if err != nil {
+			if errors.Is(err, sdk.ErrObjectNotFound) {
```
sfc-gh-asawicki, line 235: not sure if we should mark it as removed here, after all, the show above was successful (and I guess that in such cases we just return error always)

pkg/resources/streamlit.go
```diff
@@ -0,0 +1,403 @@
...
+				return diag.Diagnostics{
+					diag.Diagnostic{
+						Severity: diag.Warning,
+						Summary:  "Failed to query streamlit properties. Marking the resource as removed.",
+						Detail:   fmt.Sprintf("Streamlit name: %s, Err: %s", id.FullyQualifiedName(), err),
+					},
+				}
+			}
+			return diag.FromErr(err)
+		}
```
sfc-gh-asawicki, line 246: where is the part with withExternalChangesMarking?
 - are there no underlying snowflake defaults we have to reset to?
 - in such case, do we need this param in the read func?

pkg/resources/streamlit_acceptance_test.go
```diff
@@ -0,0 +1,430 @@
...
+
+	databaseId := acc.TestClient().Ids.DatabaseId()
+	schemaId := acc.TestClient().Ids.SchemaId()
+	id := acc.TestClient().Ids.RandomSchemaObjectIdentifier()
+
+	// use schema is needed because otherwise reference to external access integration fails.
+	err := acc.Client(t).Sessions.UseSchema(context.Background(), schemaId)
+	require.NoError(t, err)
+
+	stage, stageCleanup := acc.TestClient().Stage.CreateStageInSchema(t, schemaId)
```
sfc-gh-asawicki, line 34: same comment as in datasource tests

pkg/resources/streamlit_acceptance_test.go
```diff
@@ -0,0 +1,430 @@
...
+	t.Cleanup(stageCleanup)
+	// warehouse is needed because default warehouse uses lowercase, and it fails in snowflake.
+	warehouse, warehouseCleanup := acc.TestClient().Warehouse.CreateWarehouse(t)
```
sfc-gh-asawicki, line 37: same comment as in datasource tests

pkg/resources/streamlit_acceptance_test.go
```diff
@@ -0,0 +1,430 @@
...
+					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "query_warehouse", warehouse.ID().Name()),
+					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "external_access_integrations.#", "1"),
+					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "external_access_integrations.0", externalAccessIntegrationId.Name()),
+					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "title", "foo"),
+					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "comment", "foo")),
+			},
+			// change externally
+			{
+				PreConfig: func() {
+					acc.TestClient().Streamlit.Update(t, sdk.NewAlterStreamlitRequest(id).WithSet(
```
sfc-gh-asawicki, line 185: let's add one more test with external change to a non-required attr please

sfc-gh-asawicki (changes requested):
Please split the next PRs into parts (at least SDK, resource, datasource)

MIGRATION_GUIDE.md

docs/data-sources/streamlits.md
sfc-gh-jmichalak: Fixed.

docs/resources/streamlit.md
sfc-gh-jmichalak: Fixed.

pkg/acceptance/helpers/external_access_integration_client.go
sfc-gh-jmichalak: Done.

pkg/datasources/streamlits_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/helpers/helpers.go
sfc-gh-jmichalak: Fixed.

pkg/helpers/helpers_test.go
sfc-gh-jmichalak: Done.

pkg/helpers/helpers_test.go
sfc-gh-jmichalak: Fixed.

pkg/helpers/helpers_test.go
sfc-gh-jmichalak: Done.

pkg/helpers/helpers_test.go
sfc-gh-jmichalak: Done.

pkg/resources/streamlit_acceptance_test.go
sfc-gh-jmichalak: Fixed.

pkg/datasources/streamlits_acceptance_test.go
sfc-gh-jmichalak: Right, removed.

pkg/datasources/streamlits_acceptance_test.go
sfc-gh-jmichalak: Added tests, created a jira and liked in relevant places. Also, asked about this on slack.

pkg/resources/api_authentication_integration_common.go
sfc-gh-jmichalak: This is just a rename - describe returns a list of values here. Added comments for explanation.

pkg/resources/diff_suppressions.go
sfc-gh-jmichalak: Done.

pkg/resources/streamlit.go
sfc-gh-jmichalak: Done, but changed `Ignore...` funcs a bit.

pkg/resources/streamlit.go
sfc-gh-jmichalak: The resource may be deleted between these two queries. We handle it similarly in security integrations.

pkg/resources/streamlit.go
sfc-gh-jmichalak: That's correct - removed.

pkg/resources/streamlit_acceptance_test.go
sfc-gh-jmichalak: Same response.

pkg/resources/streamlit_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/streamlit.go
sfc-gh-asawicki: It is highly unlikely.

Also, just returning error and rerunning plan/apply will result in a proper removal. I think we should not handle it that way neither here nor in security integrations (warehouses and databases are not handling it that way).

sfc-gh-asawicki (approved):
Approving but please check the two failing tests

pkg/resources/streamlit.go
sfc-gh-jcieslak: Makes sense, we can only check SHOW (the first query). Additionally, I don't think we would ever enter this if statement because Describe is not calling `collections.Find`. Instead, this should use `sdk.ErrObjectNotExistOrAuthorized`. Anyway, this can be removed.

sfc-gh-jmichalak: Removed.

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

PR #2936, opened 2024-07-12
feat: Add missing session parameters

sfc-gh-asawicki:
Add missing session parameters (before redesigning user resource)


pkg/sdk/parameters_impl.go
```diff
@@ -20,115 +20,88 @@ func GetSessionParametersFrom(params map[string]any) (*SessionParameters, error)
...
+		err = setIntegerValue(parameter, value, &sessionParameters.JSONIndent)
 	case SessionParameterLockTimeout:
-		v, err := strconv.Atoi(value)
-		if err != nil {
-			return fmt.Errorf("LOCK_TIMEOUT session parameter is an integer, got %v", value)
-		}
-		sessionParameters.LockTimeout = Pointer(v)
+		err = setIntegerValue(parameter, value, &sessionParameters.LockTimeout)
+	case SessionParameterLogLevel:
+		sessionParameters.LogLevel = Pointer(LogLevel(value))
```
sfc-gh-jcieslak, line 78: For enum do we want to use ToX(...) functions for enums? If yes, then ToLogLevel is available.

pkg/sdk/parameters_impl.go
```diff
@@ -147,43 +120,43 @@ func (sessionParameters *SessionParameters) setParam(parameter SessionParameter,
 		sessionParameters.TimeInputFormat = &value
 	case SessionParameterTimeOutputFormat:
 		sessionParameters.TimeOutputFormat = &value
+	case SessionParameterTraceLevel:
+		sessionParameters.TraceLevel = Pointer(TraceLevel(value))
```
sfc-gh-jcieslak, line 124: Same as with LogLevel and other enums

pkg/sdk/parameters_impl.go
```diff
@@ -199,85 +172,123 @@ func GetSessionParametersUnsetFrom(params map[string]any) (*SessionParametersUns
...
 	case SessionParameterWeekOfYearPolicy:
-		sessionParametersUnset.WeekOfYearPolicy = Bool(true)
+		unsetField = &sessionParametersUnset.WeekOfYearPolicy
 	case SessionParameterWeekStart:
-		sessionParametersUnset.WeekStart = Bool(true)
+		unsetField = &sessionParametersUnset.WeekStart
 	default:
 		return fmt.Errorf("%s session parameter is not supported", string(parameter))
 	}
+	*unsetField = Bool(true)
```
sfc-gh-jcieslak, line 292: Had to pause for a minute to figure out you're assigning a setter that you set at the end 😅

pkg/sdk/parameters_impl.go
sfc-gh-asawicki: I was thinking about this and was worried about possible breaking changes. Using a bad value should end with an error either way, but sometimes, values that we do not accept are accepted in Snowflake (so this is a part I was worried about). These parameters are used in multiple resources, so it's a bit of a waste to now try to announce this potential change in each of them. We can for sure do it with the redesign, because we are listing the possible values then.

Would you be okay with the TODO for session_parameters issue?

pkg/sdk/parameters_impl.go
sfc-gh-asawicki: As above

pkg/sdk/parameters_impl.go
sfc-gh-jcieslak: Yeah, the TODO would be ok. I was thinking about something similar to a configuration list (just a slice of structs) that would generate appropriate mappings (for a future improvement ofc) etc where it could look similar to
```go
var parameters = {
  "param_name": {Levels: []Level{Account, Session, User}, Type: TypeInt, ....}
}
```

sfc-gh-jcieslak (approved):

pkg/sdk/parameters_impl.go
```diff
@@ -199,85 +173,123 @@ func GetSessionParametersUnsetFrom(params map[string]any) (*SessionParametersUns
 }
 
 func (sessionParametersUnset *SessionParametersUnset) setParam(parameter SessionParameter) error {
```
sfc-gh-jmichalak, line 175: Nit: I know this is not a part of this PR, but shouldn't the name be `unsetParam`?

sfc-gh-jmichalak (commented):
What about other session parameters like `CLIENT_ENABLE_LOG_INFO_STATEMENT_PARAMETERS`, `ENABLE_UNHANDLED_EXCEPTIONS_REPORTING`, `JDBC_ENABLE_PUT_GET`, `JS_TREAT_INTEGER_AS_BIGINT`?

pkg/sdk/parameters_impl.go
sfc-gh-asawicki: hmmm, I think both are valid: it's setting the given param on the unset object, so this is setting a param after all :)

PR #2945, opened 2024-07-17
feat: Schema SDK upgrade

sfc-gh-jmichalak:
This change upgrades SDK: adds missing columns and parameters, adds parameter handling in pkg/schemas, options parsing, missing validations, adds more unit and integration tests.

## Test Plan

* [x] unit tests

* [x] integration tests
## References

https://docs.snowflake.com/en/sql-reference/sql/create-schema


pkg/sdk/testint/schemas_integration_test.go
```diff
@@ -156,25 +281,100 @@ func TestInt_SchemasAlter(t *testing.T) {
...
+				QuotedIdentifiersIgnoreCase:             sdk.Bool(true),
+				EnableConsoleOutput:                     sdk.Bool(true),
+				PipeExecutionPaused:                     sdk.Bool(true),
+			},
+		})
 		require.NoError(t, err)
-		assert.Equal(t, comment, *s.Comment)
+
+		params = testClientHelper().Parameter.ShowSchemaParameters(t, schemaTest.ID())
+		assertParameterEqualsToDefaultValue(t, params, sdk.ObjectParameterDataRetentionTimeInDays)
```
sfc-gh-asawicki, line 361: please add this also to the create minimal test

pkg/sdk/testint/schemas_integration_test.go
```diff
@@ -13,11 +15,22 @@ func TestInt_SchemasCreate(t *testing.T) {
 	client := testClient(t)
 	ctx := testContext(t)
 
-	// new schema created on purpose
-	schema, cleanupSchema := testClientHelper().Schema.CreateSchema(t)
-	t.Cleanup(cleanupSchema)
+	t.Run("minimal", func(t *testing.T) {
```
sfc-gh-asawicki, line 18: nit: we had earlier the arrangement, that we put all the integration tests in one function instead of TestCreate, TestAlter etc

pkg/sdk/testint/schemas_integration_test.go
```diff
@@ -156,25 +281,100 @@ func TestInt_SchemasAlter(t *testing.T) {
 		assert.Equal(t, table.Name, schemaDetails[0].Name)
 	})
 
-	t.Run("set", func(t *testing.T) {
```
sfc-gh-asawicki, line 159: do I see correctly that the previous test for set was replaced with the parameters one? The only one non-parameter attribute is comment, right? Even if this is the only one, let's test set + unset for it

sfc-gh-asawicki (changes requested):

pkg/sdk/testint/schemas_integration_test.go
```diff
@@ -103,12 +118,122 @@ func TestInt_SchemasCreate(t *testing.T) {
...
+		t.Helper()
+		assert.Equal(t, expected, helpers.FindParameter(t, params, parameterName).Value)
+	}
+
+	assertParameterEqualsToDefaultValue := func(t *testing.T, params []*sdk.Parameter, parameterName sdk.ObjectParameter) {
+		t.Helper()
+		param, err := collections.FindOne(params, func(param *sdk.Parameter) bool { return param.Key == string(parameterName) })
+		assert.NoError(t, err)
+		assert.NotNil(t, param)
+		if param != nil && (*param).Level == "" {
```
sfc-gh-asawicki, line 232: one more comment here: if we go to else, nothing is validated in this assertion - it is not good, right?

pkg/sdk/testint/schemas_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/schemas_integration_test.go
sfc-gh-jmichalak: Changed.

pkg/sdk/testint/schemas_integration_test.go
sfc-gh-jmichalak: Added.

pkg/sdk/testint/schemas_integration_test.go
sfc-gh-jmichalak: Dropped this condition.

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (dismissed):

PR #2953, opened 2024-07-22
fix: ATTRIBUTE set(string) parsing for cortex search service

chris-codaio:
Fixes the issue pointed out in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2860#issuecomment-2243785375

Verified in my local deployment

PTAL @sfc-gh-asawicki @sfc-gh-mloring


sfc-gh-asawicki:
Hey @chris-codaio. Thanks for the contribution!

Normally, I would ask to provide an acceptance test proving the error before and the fix after, but because our acceptance test is currently skipped (I have just checked that the error blocking us from moving on with it was fixed on Snowflake side, so I will unskip it today), I will merge it after just a local testing I did. Thanks again :)

chris-codaio:
Thank you @sfc-gh-asawicki !

PR #2955, opened 2024-07-24
feat: Rework schema resource

sfc-gh-jmichalak:
Rework schema resource:
- add missing fields
- add acceptance tests
- add entry to migration guide
- add schema upgrader with acc test

## Test Plan

* [x] acceptance tests

## References

https://docs.snowflake.com/en/sql-reference/sql/create-schema


docs/resources/schema.md
```diff
@@ -12,14 +12,38 @@ description: |-
...
+
+# resource with all fields set
+resource "snowflake_schema" "schema" {
+  name                = "schema_name"
+  database            = "database_name"
+  with_managed_access = true
+  is_transient        = true
+  comment             = "my schema"
+
+  data_retention_time_in_days                   = 10
```
sfc-gh-jcieslak, line 29: I believe this is an invalid config because the data retention time can only be 0 or 1 for transient objects. Make it not transient or make sure the values are correct.

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
 		Description: "The database in which to create the schema.",
 		ForceNew:    true,
 	},
-	"comment": {
-		Type:        schema.TypeString,
+	"with_managed_access": {
+		Type:        schema.TypeBool,
 		Optional:    true,
-		Description: "Specifies a comment for the schema.",
+		Description: "Specifies a schema as transient. Transient schemas do not have a Fail-safe period so they do not incur additional storage costs once they leave Time Travel; however, this means they are also not protected by Fail-safe in the event of a data loss.",
```
sfc-gh-jcieslak, line 39: I think it's copy-pasted description from transient.

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.DatabaseObjectIdentifier)
+	warehouseParameters, err := client.Parameters.ShowParameters(ctx, &sdk.ShowParametersOptions{
+		In: &sdk.ParametersIn{
+			Schema: id,
+		},
+	})
+	if err != nil {
+		return nil, err
 	}
+	return warehouseParameters, nil
```
sfc-gh-jcieslak, line 159: change the name

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
+		QuotedIdentifiersIgnoreCase:             quotedIdentifiersIgnoreCase,
+		EnableConsoleOutput:                     enableConsoleOutput,
+		PipeExecutionPaused:                     GetConfigPropertyAsPointerAllowingZeroValue[bool](d, "pipe_execution_paused"),
+		Comment:                                 GetConfigPropertyAsPointerAllowingZeroValue[string](d, "comment"),
 	}
-
-	err := client.Schemas.Create(ctx, sdk.NewDatabaseObjectIdentifier(database, name), createReq)
-	if err != nil {
-		return fmt.Errorf("error creating schema %v err = %w", name, err)
+	if strings.EqualFold(strings.TrimSpace(name), "PUBLIC") {
```
sfc-gh-jcieslak, line 211: I think there's also a schema named local mentioned [here](https://docs.snowflake.com/en/sql-reference/sql/show-schemas#usage-notes), but I'm not sure if you can `alter` or `create or alter` it.

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
-func ReadSchema(d *schema.ResourceData, meta interface{}) error {
+func ReadContextSchema(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*provider.Context).Client
-	ctx := context.Background()
 	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.DatabaseObjectIdentifier)
 
-	database, err := client.Databases.ShowByID(ctx, id.DatabaseId())
+	_, err := client.Databases.ShowByID(ctx, id.DatabaseId())
 	if err != nil {
 		d.SetId("")
```
sfc-gh-jcieslak, line 228: I'm wondering if we ever need this? And if so maybe it would better to return the error here and not hope that it will be returned when calling show schemas. The worst case would be if we would get error here, but not in the show schemas, then it would be really painful to debug without any log or warning added to Diagnostics.

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
 	if err != nil {
-		log.Printf("[DEBUG] schema (%s) not found", d.Id())
-		d.SetId("")
-		return nil
+		if errors.Is(err, sdk.ErrObjectNotFound) {
+			d.SetId("")
+			return diag.Diagnostics{
+				diag.Diagnostic{
+					Severity: diag.Warning,
+					Summary:  "Failed to query secondary database. Marking the resource as removed.",
```
sfc-gh-jcieslak, line 238: wrong error message (also in Details)

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
 		Description: "The database in which to create the schema.",
 		ForceNew:    true,
```
sfc-gh-jcieslak, line 34: Imo there should be some diff suppression on `name` and `database` in case we get strange quoting from Snowflake.

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
+	}
+	value, err := strconv.ParseBool((*pipeExecutionPaused).Value)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if err := d.Set("pipe_execution_paused", value); err != nil {
+		return diag.FromErr(err)
 	}
 
+	if err = d.Set(ShowOutputAttributeName, []map[string]any{schemas.SchemaToSchema(schema)}); err != nil {
```
sfc-gh-jcieslak, line 291: Are we intentionally not adding the Describe output to the schema?

MIGRATION_GUIDE.md
```diff
@@ -6,6 +6,11 @@ across different versions.
 
 ## v0.93.0 ➞ v0.94.0
 
+### *(breaking change)* refactored snowflake_schema resource
+
+Changes:
+- renamed `is_managed` to `with_managed_access`
```
sfc-gh-jcieslak, line 12: maybe it would be worth to list added things to the schema (e.g. outputs from queries or OptionalComputed params that ppl can now specify).

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
-			}
-		}
-	}
-
-	if d.HasChange("is_managed") {
-		managed := d.Get("is_managed")
+	if d.HasChange("with_managed_access") {
 		var err error
-		if managed.(bool) {
+		if _, ok := d.GetOk("with_managed_access"); ok {
```
sfc-gh-jcieslak, line 319: Shouldn't this be used with the `d.Get` and compared with the BooleanTrue value?

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
 
-	return ReadSchema(d, meta)
+	return ReadContextSchema(ctx, d, meta)
 }
 
-// DeleteSchema implements schema.DeleteFunc.
-func DeleteSchema(d *schema.ResourceData, meta interface{}) error {
-	client := meta.(*provider.Context).Client
-	ctx := context.Background()
+func HandleSchemaParametersChanges(d *schema.ResourceData, set *sdk.SchemaSet, unset *sdk.SchemaUnset) diag.Diagnostics {
```
sfc-gh-jcieslak, line 369: You were reusing stuff from the database, maybe the handling method could also be reused here?

pkg/resources/schema_acceptance_test.go
```diff
@@ -18,49 +19,361 @@ import (
...
+	)
 
 	resource.Test(t, resource.TestCase{
 		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
 		PreCheck:                 func() { acc.TestAccPreCheck(t) },
 		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
 			tfversion.RequireAbove(tfversion.Version1_5_0),
 		},
-		CheckDestroy: acc.CheckDestroy(t, resources.Schema),
+		CheckDestroy: acc.CheckDestroy(t, resources.Database),
```
sfc-gh-jcieslak, line 97: Why database?

pkg/resources/schema_acceptance_test.go
```diff
@@ -192,55 +547,55 @@ func TestAcc_Schema_DefaultDataRetentionTime(t *testing.T) {
 				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_Schema_DefaultDataRetentionTime/WithoutDataRetentionSet"),
 				ConfigVariables: configVariablesWithoutSchemaDataRetentionTime(5),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr("snowflake_schema.test", "data_retention_days", r.IntDefaultString),
+					resource.TestCheckResourceAttr("snowflake_schema.test", "data_retention_time_in_days", r.IntDefaultString),
```
sfc-gh-jcieslak, line 550: this rename is a breaking change and should be communicated in the migration guide

sfc-gh-jcieslak: Also should be handled in the state upgrader

sfc-gh-jcieslak (changes requested):

pkg/resources/schema.go
sfc-gh-jmichalak: Why? This is a regular Bool, not BooleanString.

pkg/resources/schema_acceptance_test.go
sfc-gh-jmichalak: Fixed.

pkg/resources/schema.go
sfc-gh-jmichalak: Fixed.

pkg/resources/schema.go
sfc-gh-jmichalak: Done.

docs/resources/schema.md
```diff
@@ -33,28 +57,277 @@ resource "snowflake_schema" "schema" {
...
-- `is_managed` (Boolean) Specifies a managed schema. Managed access schemas centralize privilege management with the schema owner.
-- `is_transient` (Boolean) Specifies a schema as transient. Transient schemas do not have a Fail-safe period so they do not incur additional storage costs once they leave Time Travel; however, this means they are also not protected by Fail-safe in the event of a data loss.
-- `tag` (Block List, Deprecated) Definitions of a tag to associate with the resource. (see [below for nested schema](#nestedblock--tag))
+- `data_retention_time_in_days` (Number) Specifies the number of days for which Time Travel actions (CLONE and UNDROP) can be performed on the database, as well as specifying the default Time Travel retention time for all schemas created in the database. For more details, see [Understanding & Using Time Travel](https://docs.snowflake.com/en/user-guide/data-time-travel).
```
sfc-gh-asawicki, line 62: rename should be listed in the migration guide (and added to the upgrader)

docs/resources/schema.md
```diff
@@ -33,28 +57,277 @@ resource "snowflake_schema" "schema" {
...
+- `log_level` (String) Specifies the severity level of messages that should be ingested and made available in the active event table. Valid options are: [TRACE DEBUG INFO WARN ERROR FATAL OFF]. Messages at the specified level (and at more severe levels) are ingested. For more information, see [LOG_LEVEL](https://docs.snowflake.com/en/sql-reference/parameters.html#label-log-level).
+- `max_data_extension_time_in_days` (Number) Object parameter that specifies the maximum number of days for which Snowflake can extend the data retention period for tables in the database to prevent streams on the tables from becoming stale. For a detailed description of this parameter, see [MAX_DATA_EXTENSION_TIME_IN_DAYS](https://docs.snowflake.com/en/sql-reference/parameters.html#label-max-data-extension-time-in-days).
+- `pipe_execution_paused` (Boolean) Specifies whether to pause a running pipe, primarily in preparation for transferring ownership of the pipe to a different role.
+- `quoted_identifiers_ignore_case` (Boolean) If true, the case of quoted identifiers is ignored.
```
sfc-gh-asawicki, line 70: it would be nice to give the links to the docs to all the Snowflake parameters

docs/resources/schema.md
```diff
@@ -33,28 +57,277 @@ resource "snowflake_schema" "schema" {
...
+- `default_ddl_collation` (String) Specifies a default collation specification for all schemas and tables added to the database. It can be overridden on schema or table level. For more information, see [collation specification](https://docs.snowflake.com/en/sql-reference/collation#label-collation-specification).
+- `enable_console_output` (Boolean) If true, enables stdout/stderr fast path logging for anonymous stored procedures.
+- `external_volume` (String) The database parameter that specifies the default external volume to use for Iceberg tables.
+- `is_transient` (Boolean) Specifies the schema as transient. Transient schemas do not have a Fail-safe period so they do not incur additional storage costs once they leave Time Travel; however, this means they are also not protected by Fail-safe in the event of a data loss.
```
sfc-gh-asawicki, line 66: shouldn't all the boolean attributes (NOT sf params) be our boolean strings?

pkg/resources/schema.go
sfc-gh-jmichalak: Well, this function uses different set/unset structs, so this would require extracting common parameters in general. I think this requires a separate discussion on our approach here.

pkg/resources/schema.go
sfc-gh-asawicki: Sometimes sf objects behave poorly when show is invoked and one of the first parts points to a non existing object. This is an interesting problem that could be solved solely on the SDK level for each object, but I think we could leave this discussion for the object renaming (cause it touches the removal topic too).

pkg/resources/schema.go
sfc-gh-jmichalak: It behaves like a normal schema and doesn't require any specific handling.

pkg/resources/schema_acceptance_test.go
```diff
@@ -31,36 +97,283 @@ func TestAcc_Schema(t *testing.T) {
...
 		Steps: []resource.TestStep{
 			{
-				ConfigDirectory: config.TestNameDirectory(),
-				ConfigVariables: map[string]config.Variable{
-					"name":     config.StringVariable(name),
-					"database": config.StringVariable(acc.TestDatabaseName),
-					"comment":  config.StringVariable(comment),
+				PreConfig: func() {
+					params := acc.TestClient().Parameter.ShowAccountParameters(t)
+					*accountDataRetentionTimeInDays = helpers.FindParameter(t, params, sdk.AccountParameterDataRetentionTimeInDays).Value
```
sfc-gh-asawicki, line 102: maybe try using the new object parameters assertions here? 🙏

pkg/resources/schema_acceptance_test.go
```diff
@@ -31,36 +97,283 @@ func TestAcc_Schema(t *testing.T) {
...
 			},
-			// UPDATE COMMENT (proves issue https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2606)
+			// import - without optionals
 			{
-				ConfigDirectory: config.TestNameDirectory(),
-				ConfigVariables: map[string]config.Variable{
-					"name":     config.StringVariable(name),
-					"database": config.StringVariable(acc.TestDatabaseName),
-					"comment":  config.StringVariable(""),
+				ConfigDirectory:         acc.ConfigurationDirectory("TestAcc_Schema/basic"),
```
sfc-gh-asawicki, line 149: does import care about the cfg dir and vars?

pkg/resources/schema_acceptance_test.go
```diff
@@ -424,3 +779,66 @@ func checkDatabaseAndSchemaDataRetentionTime(t *testing.T, schemaId sdk.Database
 		return nil
 	}
 }
+
+func TestAcc_Schema_migrateFromVersion093(t *testing.T) {
```
sfc-gh-asawicki, line 783: -let's migrate public schema too (just to see what happens)
-let'smigrate the one with tags too
-and the one without with_managed_access
-and maybe let's check how the newly added sf params behave after update

docs/resources/schema.md
sfc-gh-jmichalak: Why? In Snowflake, this is not a boolean field, but an option during create.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: Done.

docs/resources/schema.md
sfc-gh-jmichalak: Right, fixed.

docs/resources/schema.md
sfc-gh-jmichalak: Done.

docs/resources/schema.md
```diff
@@ -33,28 +57,277 @@ resource "snowflake_schema" "schema" {
 
 ### Optional
 
+- `catalog` (String) The database parameter that specifies the default catalog to use for Iceberg tables.
 - `comment` (String) Specifies a comment for the schema.
-- `data_retention_days` (Number) Specifies the number of days for which Time Travel actions (CLONE and UNDROP) can be performed on the schema, as well as specifying the default Time Travel retention time for all tables created in the schema. Default value for this field is set to -1, which is a fallback to use Snowflake default.
-- `is_managed` (Boolean) Specifies a managed schema. Managed access schemas centralize privilege management with the schema owner.
-- `is_transient` (Boolean) Specifies a schema as transient. Transient schemas do not have a Fail-safe period so they do not incur additional storage costs once they leave Time Travel; however, this means they are also not protected by Fail-safe in the event of a data loss.
-- `tag` (Block List, Deprecated) Definitions of a tag to associate with the resource. (see [below for nested schema](#nestedblock--tag))
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Done.

docs/resources/schema.md
sfc-gh-jmichalak: Done.

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
+		QuotedIdentifiersIgnoreCase:             quotedIdentifiersIgnoreCase,
+		EnableConsoleOutput:                     enableConsoleOutput,
+		PipeExecutionPaused:                     GetConfigPropertyAsPointerAllowingZeroValue[bool](d, "pipe_execution_paused"),
+		Comment:                                 GetConfigPropertyAsPointerAllowingZeroValue[string](d, "comment"),
 	}
-
-	err := client.Schemas.Create(ctx, sdk.NewDatabaseObjectIdentifier(database, name), createReq)
-	if err != nil {
-		return fmt.Errorf("error creating schema %v err = %w", name, err)
+	if strings.EqualFold(strings.TrimSpace(name), "PUBLIC") {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Added.

pkg/resources/schema_acceptance_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers"
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: 🤷 Moved to the group below.

pkg/resources/schema_acceptance_test.go
sfc-gh-jmichalak: When I remove them, I get `Error: No value for required variable`

pkg/resources/schema_acceptance_test.go
sfc-gh-jmichalak: If we have time before release, then sure.

docs/resources/schema.md
sfc-gh-asawicki: Because currently the proper behavior of this field assumes that SF default is also fault but it will fail with the default change. It won't with the boolstring.

Also, consistency in the provider.

pkg/resources/schema_acceptance_test.go
sfc-gh-asawicki: okay, but these are not used for the import tests AFAIK

pkg/resources/schema.go
```diff
@@ -29,268 +33,380 @@ var schemaSchema = map[string]*schema.Schema{
...
+				strings.ToLower(string(sdk.ObjectParameterUserTaskManagedInitialWarehouseSize)),
+				strings.ToLower(string(sdk.ObjectParameterUserTaskTimeoutMs)),
+				strings.ToLower(string(sdk.ObjectParameterUserTaskMinimumTriggerIntervalInSeconds)),
+				strings.ToLower(string(sdk.ObjectParameterQuotedIdentifiersIgnoreCase)),
+				strings.ToLower(string(sdk.ObjectParameterEnableConsoleOutput)),
+				strings.ToLower(string(sdk.ObjectParameterPipeExecutionPaused)),
+			),
+			ParametersCustomDiff(
+				schemaParametersProvider,
+				parameter{sdk.AccountParameterDataRetentionTimeInDays, valueTypeInt, sdk.ParameterTypeSchema},
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: They were unnecessary, removed (the only different param is PipeExecutionPaused).

pkg/resources/schema.go
sfc-gh-jmichalak: Added.

pkg/resources/schema.go
sfc-gh-jmichalak: Changed to BooleanString - done.

docs/resources/schema.md
sfc-gh-jmichalak: Changed to BooleanString (except the new bool parameter, but in databases there are also bool parameters).

pkg/resources/schema_acceptance_test.go
sfc-gh-jmichalak: Done without the last one.

sfc-gh-asawicki (dismissed):

pkg/resources/schema.go
```diff
@@ -145,18 +165,45 @@ func Schema() *schema.Resource {
...
+
+	if err := d.Set("database", s.DatabaseName); err != nil {
+		return nil, err
+	}
+
+	if err := d.Set("comment", s.Comment); err != nil {
+		return nil, err
+	}
+
+	if err := d.Set("is_transient", fmt.Sprintf("%t", s.IsTransient())); err != nil {
```
sfc-gh-jcieslak, line 189: I think we already have mapping methods for 3 value bools

pkg/resources/schema.go
```diff
@@ -208,94 +253,139 @@ func CreateContextSchema(ctx context.Context, d *schema.ResourceData, meta any)
...
-	}
-	if err := d.Set("pipe_execution_paused", value); err != nil {
-		return diag.FromErr(err)
-	}
+		if withExternalChangesMarking {
+			if err = handleExternalChangesToObjectInShow(d,
+				showMapping{"options", "is_transient", schema.IsTransient(), fmt.Sprintf("%t", schema.IsTransient()), func(x any) any {
+					return slices.Contains(sdk.ParseCommaSeparatedStringArray(x.(string), false), "TRANSIENT")
+				}},
+				showMapping{"options", "with_managed_access", schema.IsManagedAccess(), fmt.Sprintf("%t", schema.IsManagedAccess()), func(x any) any {
```
sfc-gh-jcieslak, line 361: same here, predefined mapping functions can be used

pkg/resources/schema.go
```diff
@@ -208,94 +253,139 @@ func CreateContextSchema(ctx context.Context, d *schema.ResourceData, meta any)
...
+		if err != nil {
+			log.Printf("[DEBUG] database %s for schema %s not found", id.DatabaseId().Name(), id.Name())
 			d.SetId("")
-			return diag.Diagnostics{
-				diag.Diagnostic{
-					Severity: diag.Warning,
-					Summary:  "Failed to query secondary database. Marking the resource as removed.",
-					Detail:   fmt.Sprintf("DatabaseName: %s, Err: %s", id.FullyQualifiedName(), err),
-				},
+			return nil
```
sfc-gh-jcieslak, line 296: If we return here I think it would be better to also return Diag warning as in the case below to inform user about failed database query and that we're marking resource as removed. Otherwise, user will get meaningless `produced an unexpected new value: Root │ object was present, but now absent.` error.

pkg/resources/schema.go
```diff
@@ -79,16 +101,14 @@ func Schema() *schema.Resource {
 		SchemaVersion: 1,
 
 		CreateContext: CreateContextSchema,
-		ReadContext:   ReadContextSchema,
+		ReadContext:   ReadContextSchema(true),
 		UpdateContext: UpdateContextSchema,
 		DeleteContext: DeleteContextSchema,
+		Description:   "Resource used to manage schema objects. For more information, check [schema documentation](https://docs.snowflake.com/en/sql-reference/sql/create-schema).",
 
 		CustomizeDiff: customdiff.All(
```
sfc-gh-jcieslak, line 109: I'm wondering if we should somehow detect change to recompute the Describe output? Because right now I think it will be always the same and never change (because of lacking custom diff to recompute). Btw, you could at least add ComputedIfAnyAttributeChanged for Describe output with "name" attrib.

pkg/resources/schema.go
```diff
@@ -315,18 +405,30 @@ func UpdateContextSchema(ctx context.Context, d *schema.ResourceData, meta any)
...
+				})
+			}
+			if err != nil {
+				return diag.FromErr(fmt.Errorf("error handling with_managed_access on %v err = %w", d.Id(), err))
+			}
 		}
-		if err != nil {
-			return diag.FromErr(fmt.Errorf("error changing management state on %v err = %w", d.Id(), err))
+		// managed access can not be UNSET to a default value
+		if err := client.Schemas.Alter(ctx, id, &sdk.AlterSchemaOptions{
```
sfc-gh-jcieslak, line 428: I don't get this part; we always disable managed access even if we enabled it a second ago in line 416. I think this can cause infinite plan for with_managed_access field.

MIGRATION_GUIDE.md
```diff
@@ -6,6 +6,49 @@ across different versions.
...
+### *(breaking change)* refactored snowflake_schema resource
+
+Renamed fields:
+- renamed `is_managed` to `with_managed_access`
+- renamed `data_retention_days` to `data_retention_time_in_days`
+
+Please rename these fields in your configuration files. State will be migrated automatically.
+
+Removed fields:
+- `tag`
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Added.

pkg/resources/diff_suppressions.go
```diff
@@ -24,12 +24,61 @@ func NormalizeAndCompare[T comparable](normalize func(string) (T, error)) schema
 	}
 }
 
+func NormalizeAndCompareIdentifiers() schema.SchemaDiffSuppressFunc {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Changed to suppressIdentifierQuoting and removed NormalizeAndCompareIdentifiers.

pkg/resources/schema.go
sfc-gh-jmichalak: Done.

pkg/resources/schema.go
sfc-gh-jmichalak: Done.

pkg/resources/schema.go
sfc-gh-jmichalak: Done.

pkg/resources/schema.go
sfc-gh-jmichalak: Well, in description we only have a list of objects inside the schema, and we're not specifying any inside terraform config. Added ComputedIfAnyAttributeChanged for Describe output with "name", though.

pkg/resources/schema.go
sfc-gh-jmichalak: Fixed.

sfc-gh-jcieslak (dismissed):

sfc-gh-asawicki (dismissed):

sfc-gh-jcieslak (dismissed):

sfc-gh-jcieslak (approved):

PR #2957, opened 2024-07-24
feat: Add identifier parsers

sfc-gh-jcieslak:
Changes:
- Add and test identifier parsers for object and account identifiers with different separators (only dot was added as identifier parser, description provided below).

The encoder/parser for resource identifiers was in the end implemented as simple string join/split by pipe, because of the following:
- Identifiers that have only one part that is represented by the `sdk.ObjectIdentifier` should be encoded as `FullyQualifiedName` and parsed with one of the newly introduced parsers. This should improve a bit the support for tools that generate import commands (or import blocks) because the resource identifier is the same as the Snowflake one.
- More complicated identifiers having more parts should use the provided encoder/parser to create or parse resource identifiers into parts. This could cause problems for things that may contain pipe, but we should generally avoid it or document it (can be done in the usage part of identifier rework). Also, CSV reader/writer cannot be used for resource identifiers, because this would make identifiers more complex. For example, a fully qualified identifier with quotes would have to be imported as `terraform import snowflake_table.test_table '"""database_name"".""schema_name"".table_name"""'`. To have this as an identifier would also mean creating a state upgrader for every resource (or general one that could be reused) that would transform the identifiers into CSV-compliant ones. 

Notes: 
- Implementation with the use of `identifeirParsingFunc` was added because initially, I had two parsers (one with dot as csv.Comma and the other one with pipe) and later on removed the one with the pipe as csv.Comma (I can merge the unexported parsers with exported ones or leave it as is if we would like to introduce another internal parser).
- ExternalObjectIdentifier and AccountIdentifier parsers were created with certain assumptions (described in code).

## Test Plan

* [x] unit tests

## References

* SNOW-1495051


pkg/helpers/identifier_string_parser_test.go
```diff
@@ -91,4 +94,251 @@ func Test_ParseIdentifierString(t *testing.T) {
...
+		{Input: []string{"", "", ""}, Expected: `||`},
+		{Input: []string{"a", "b", "c"}, Expected: `a|b|c`},
+		// If one of the parts contains a separator sign (pipe in this case),
+		// we can end up with more parts than we started with.
+		{Input: []string{"a", "b", "c|d"}, Expected: `a|b|c|d`, ExpectedAfterDecoding: []string{"a", "b", "c", "d"}},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(fmt.Sprintf(`Encoding and parsing %s resource identifier`, testCase.Input), func(t *testing.T) {
+			encodedIdentifier := EncodeResourceIdentifier(testCase.Input...)
```
sfc-gh-jmichalak, line 333: How about we use real sdk.*Identifier types instead of []string? Potentially, we could have helpers to extract `parts` for these types so they can be easily used here.

sfc-gh-jmichalak (changes requested):

pkg/helpers/identifier_string_parser.go
```diff
@@ -25,8 +24,126 @@ func parseIdentifierStringWithOpts(identifier string, opts func(*csv.Reader)) ([
...
+		return sdk.DatabaseObjectIdentifier{}, fmt.Errorf(`unexpected number of parts %d in identifier %s, expected 2 in a form of "<database_name>%c<database_object_name>"`, len(parts), identifier, delimiter)
+	}
+	return sdk.NewDatabaseObjectIdentifier(parts[0], parts[1]), nil
+}
+
+func ParseDatabaseObjectIdentifier(identifier string) (sdk.DatabaseObjectIdentifier, error) {
+	return parseDatabaseObjectIdentifier(identifier, ParseIdentifierString, IdDelimiter)
+}
+
+func parseSchemaObjectIdentifier(identifier string, parser identifierParsingFunc, delimiter rune) (sdk.SchemaObjectIdentifier, error) {
```
sfc-gh-asawicki, line 83: delimiter is used here only to build the error and not to pass it to parser, so I don't understand why we pass it (or why we don't use it).

pkg/helpers/identifier_string_parser.go
```diff
@@ -25,8 +24,126 @@ func parseIdentifierStringWithOpts(identifier string, opts func(*csv.Reader)) ([
...
+		return make([]string, 0)
+	}
+	return strings.Split(identifier, string(ResourceIdDelimiter))
+}
+
+func EncodeResourceIdentifier(parts ...string) string {
+	return strings.Join(parts, string(ResourceIdDelimiter))
+}
+
+func parseAccountObjectIdentifier(identifier string, parser identifierParsingFunc) (sdk.AccountObjectIdentifier, error) {
```
sfc-gh-asawicki, line 53: all this unexported functions are almost the same, can't we just have a parametrized func like:
`func parseIdentifier[I sdk.ObjectIdentifier](identifier string, parser (however it's always ParseIdentifierString currently , so maybe we can inline it), ex[pectedParts int, expectedFormat string(like "<account_object_name>"))`? it will be much less code and will be much more readable, because now one may wonder what happens in these functions if each of them is rewritten

pkg/helpers/identifier_string_parser_test.go
```diff
@@ -91,4 +94,251 @@ func Test_ParseIdentifierString(t *testing.T) {
...
+
+func Test_ParseAccountObjectIdentifier(t *testing.T) {
+	testCases := []struct {
+		Input    string
+		Expected sdk.AccountObjectIdentifier
+		Error    string
+	}{
+		{Input: ``, Error: "incompatible identifier: "},
+		{Input: "a\nb", Error: "incompatible identifier: a\nb"},
+		{Input: `a"b`, Error: "unable to read identifier: a\"b, err = parse error on line 1, column 2: bare \" in non-quoted-field"},
```
sfc-gh-asawicki, line 127: is snowflake escaping such `"` or do we have to do this on the SDK level?

pkg/helpers/identifier_string_parser_test.go
```diff
@@ -91,4 +94,251 @@ func Test_ParseIdentifierString(t *testing.T) {
...
+		{Input: `abc.cde`, Error: `unexpected number of parts 2 in identifier abc.cde, expected 1 in a form of "<account_object_name>"`},
+		{Input: `""`, Expected: sdk.NewAccountObjectIdentifier(``)},
+		{Input: `""""`, Expected: sdk.NewAccountObjectIdentifier(`"`)},
+		{Input: `abc`, Expected: sdk.NewAccountObjectIdentifier(`abc`)},
+		{Input: `"abc"`, Expected: sdk.NewAccountObjectIdentifier(`abc`)},
+		{Input: `"ab.c"`, Expected: sdk.NewAccountObjectIdentifier(`ab.c`)},
+		{Input: `"a""bc"`, Expected: sdk.NewAccountObjectIdentifier(`a"bc`)},
```
sfc-gh-asawicki, line 134: I would ask about the uppercase but this is of course pointless in the unit tests because they will work the same. However, it would be great to have integration tests for these parsers, i.e. let's take a few object that have id fields in show/describe, create the object with one of the above id, and check if we can parse it from the response correctly (and there the casing may matter)

pkg/helpers/identifier_string_parser_test.go
```diff
@@ -91,4 +94,251 @@ func Test_ParseIdentifierString(t *testing.T) {
...
+	}
+
+	for _, testCase := range testCases {
```
sfc-gh-asawicki, line 137: this test is the same, let's just reuse one (either by having one test func, or having this test func extracted)

pkg/helpers/identifier_string_parser.go
```diff
@@ -25,8 +24,126 @@ func parseIdentifierStringWithOpts(identifier string, opts func(*csv.Reader)) ([
...
+	if err != nil {
+		return sdk.SchemaObjectIdentifier{}, err
+	}
+	if len(parts) != 3 {
+		return sdk.SchemaObjectIdentifier{}, fmt.Errorf(`unexpected number of parts %[1]d in identifier %[2]s, expected 3 in a form of "<database_name>%[3]c<schema_name>%[3]c<schema_object_name>"`, len(parts), identifier, delimiter)
+	}
+	return sdk.NewSchemaObjectIdentifier(parts[0], parts[1], parts[2]), nil
+}
+
+func ParseSchemaObjectIdentifier(identifier string) (sdk.SchemaObjectIdentifier, error) {
```
sfc-gh-asawicki, line 94: To better grasp the impact of the change I would like to see the example usage (in at least one resource; any of the ParseX functions, or example for all of them). Also, the encode one.

It will help to understand what we are replacing and how it looks after the transition

pkg/helpers/identifier_string_parser.go
```diff
@@ -4,17 +4,16 @@ import (
 	"encoding/csv"
 	"fmt"
 	"strings"
+
```
sfc-gh-asawicki, line 7: I wonder where this whole parsing logic should reside, and I think that:
- parsing fits best in the SDK package (because it's SF logic, and nothing tf-specific)
- encoding should reside in tf specific package, because it it where we will use it (and this is nothing to do with the SDK)

sfc-gh-asawicki (changes requested):

pkg/helpers/identifier_string_parser.go
```diff
@@ -25,8 +24,126 @@ func parseIdentifierStringWithOpts(identifier string, opts func(*csv.Reader)) ([
...
+	if lines == nil {
+		return make([]string, 0), nil
+	}
 	if len(lines) != 1 {
 		return nil, fmt.Errorf("incompatible identifier: %s", identifier)
 	}
 	return lines[0], nil
 }
+
+func ParseIdentifierString(identifier string) ([]string, error) {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: unexported

pkg/helpers/identifier_string_parser.go
```diff
@@ -25,8 +24,126 @@ func parseIdentifierStringWithOpts(identifier string, opts func(*csv.Reader)) ([
...
+		return sdk.TableColumnIdentifier{}, fmt.Errorf(`unexpected number of parts %[1]d in identifier %[2]s, expected 4 in a form of "<database_name>%[3]c<schema_name>%[3]c<table_name>%[3]c<table_column_name>"`, len(parts), identifier, delimiter)
+	}
+	return sdk.NewTableColumnIdentifier(parts[0], parts[1], parts[2], parts[3]), nil
+}
+
+func ParseTableColumnIdentifier(identifier string) (sdk.TableColumnIdentifier, error) {
+	return parseTableColumnObjectIdentifier(identifier, ParseIdentifierString, IdDelimiter)
+}
+
+func parseAccountIdentifier(identifier string, parser identifierParsingFunc, delimiter rune) (sdk.AccountIdentifier, error) {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: In this case, nothing much, I guess. Those parsers are going to replace the old ones so in this case It'll be replaced.

pkg/helpers/identifier_string_parser_test.go
```diff
@@ -91,4 +94,251 @@ func Test_ParseIdentifierString(t *testing.T) {
...
+	})
+}
+
+func Test_ParseAccountObjectIdentifier(t *testing.T) {
+	testCases := []struct {
+		Input    string
+		Expected sdk.AccountObjectIdentifier
+		Error    string
+	}{
+		{Input: ``, Error: "incompatible identifier: "},
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I don't get it. It uses ParseIdentifierString, so this one has the same capabilities as ParseIdentifierString. It will error on \`\`, but won't on \`""\` (just as ParseIdentifierString)

pkg/helpers/identifier_string_parser.go
sfc-gh-jcieslak: Applied the suggestion below (delimiter is not needed now)

pkg/helpers/identifier_string_parser.go
sfc-gh-jcieslak: Right, changed

pkg/helpers/identifier_string_parser_test.go
sfc-gh-jcieslak: Not sure If I understand correctly, but `a"b` is not allowed (in Snowflake and CSV parser). It would be legal in Snowflake and CSV parser if it would be enclosed by double quotes, and the quote inside should be escaped with another quote like so `"a""b"` (which results in `a"b` being saved in SF).

sfc-gh-asawicki: yes, and in such a case, what will the Snowflake return: escaped or unescaped quote inside?

pkg/helpers/identifier_string_parser_test.go
sfc-gh-jcieslak: That made me think about how we actually want to save our IDs in the state. I think for IDs represented by only Snowflake ID we could do just `d.SetId(id.FullyQualifiedName())` and `id, err := ParseXIdentifier(d.Id)`. For more complex identifiers (like grant ones) we could use this one as `EncodeResourceIdentifier(id.FullyQualifiedName(), "Type", "some other data")` and `parts := ParseResourceIdentifier; do the stuff with parts like id parsing from the first part`. With that in mind, there's no need for proposed logic (or maybe I misunderstood it).

pkg/helpers/identifier_string_parser_test.go
sfc-gh-jcieslak: Merged into one test func

pkg/helpers/identifier_string_parser.go
sfc-gh-jcieslak: Yeah, I forgot to add this to the pr notes, but I just wasn't sure where this should go. Moved to appropriate packages.

pkg/resources/shared_database_acceptance_test.go
```diff
@@ -25,6 +25,13 @@ func TestAcc_CreateSharedDatabase_Basic(t *testing.T) {
 	newId := acc.TestClient().Ids.RandomAccountObjectIdentifier()
 	newComment := random.Comment()
 
+	t.Cleanup(func() {
+		// TODO(SNOW-1562172: Create a better solution for this type of situations
+		// It's needed, because we have to wait for Snowflake to "register" the cleanup after this test.
+		// Without it, the next test would fail.
+		time.Sleep(time.Second * 60)
```
sfc-gh-jcieslak, line 32: It ended up as a pretty big number (I tried a bunch of lower ones), lmk what do you think we should do about it (or just leave it as is for now)

pkg/helpers/identifier_string_parser_test.go
sfc-gh-jcieslak: I checked, and unfortunately, Snowflake will return this as unescaped, which will result in an error from the parser. Realistically speaking, I doubt anyone would like to put double quotes as part of their identifier, but I guess if Snowflake allows it, we should too 🙁. I checked further on grants how it will return multiple parts, and this time, it returned it as escaped (database and schema named as a"b` resulted in `"a" b." "a" b"` 👍). That makes me wonder if we should try and encode single-column identifiers with csv writer 🤔, but again, that's only to cover this case. Should I document it and leave it for now or do we want to have an implementation and example of this somewhere in code? e.g. database `<name>`

pkg/helpers/identifier_string_parser.go
sfc-gh-jcieslak: Used encode/parse of the resource identifier in a few places, and used mostly parser for external object identifier in shared+secondary databases (also revealing possible root causes of their flakiness 😞).

pkg/helpers/identifier_string_parser_test.go
sfc-gh-jcieslak: Added integration tests, but more to document the Snowflake identifier behavior for certain scenarios.

pkg/helpers/identifier_string_parser_test.go
sfc-gh-jcieslak: As discussed, added a validation for now that would return an error whenever double quote is in the identifier.

pkg/helpers/resource_identifier_helpers.go
```diff
@@ -0,0 +1,18 @@
+package helpers
```
sfc-gh-jmichalak, line 1: Nit `helpers` is not needed in filename, because it resides in `helpers` package anyway.

sfc-gh-jmichalak (dismissed):

pkg/helpers/identifier_string_parser_test.go
sfc-gh-asawicki: Did you do a follow-up issue for this?

pkg/resources/shared_database_acceptance_test.go
sfc-gh-jmichalak: BTW there are also functions like [require.Eventually](https://pkg.go.dev/github.com/stretchr/testify/require#Eventually).

sfc-gh-jcieslak: Managed to change it with the use of require.Eventually and periodic checking with 1 minute timeout

pkg/resources/secondary_database_acceptance_test.go
```diff
@@ -23,7 +24,13 @@ func TestAcc_CreateSecondaryDatabase_Basic(t *testing.T) {
 	_, externalPrimaryId, primaryDatabaseCleanup := acc.SecondaryTestClient().Database.CreatePrimaryDatabase(t, []sdk.AccountIdentifier{
 		acc.TestClient().Account.GetAccountIdentifier(t),
 	})
-	t.Cleanup(primaryDatabaseCleanup)
+	t.Cleanup(func() {
+		// TODO(SNOW-1562172): Create a better solution for this type of situations
+		// Have to wait; otherwise the secondary database removal can be not registered yet,
+		// resulting in an error in the cleanup below.
+		time.Sleep(time.Second)
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Changed to require.Eventually

pkg/helpers/resource_identifier_helpers.go
sfc-gh-jcieslak: Changed

pkg/sdk/identifier_helpers.go
```diff
@@ -183,9 +183,6 @@ func (i AccountObjectIdentifier) Name() string {
 }
 
 func (i AccountObjectIdentifier) FullyQualifiedName() string {
-	if i.name == "" {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yeah, I also had some concerns about it and should've left it, reverted.

pkg/helpers/identifier_string_parser_test.go
sfc-gh-jcieslak: SNOW-1571674 added as in-code comment

sfc-gh-jmichalak (approved):

Issue #2960, opened 2024-07-25
[Feature]: Add "CREATE NOTEBOOK" privilege to the list of allowed plural object types for all and future lists in snowflake_grant_privileges_to_database_role resource.


anita0918:
### Use Cases or Problem Statement

All the schemas in the firm are managed schemas and are controlled by a terraform role. To create the Notebooks within underlying managed schema, user needs to apply CREATE NOTEBOOK on the schema which is not possible using current versions of terraform modules as notebook isn't included in the list of allowed plural object types.

[Nested Schema for on_schema_object](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_privileges_to_database_role#nested-schema-for-on_schema_object)

 The object type of the schema object on which privileges will be granted. Valid values are: AGGREGATION POLICY | ALERT | AUTHENTICATION POLICY | CORTEX SEARCH SERVICE | DATA METRIC FUNCTION | DYNAMIC TABLE | EVENT TABLE | EXTERNAL TABLE | FILE FORMAT | FUNCTION | GIT REPOSITORY | HYBRID TABLE | IMAGE REPOSITORY | ICEBERG TABLE | MASKING POLICY | MATERIALIZED VIEW | MODEL | NETWORK RULE | PACKAGES POLICY | PASSWORD POLICY | PIPE | PROCEDURE | PROJECTION POLICY | ROW ACCESS POLICY | SECRET | SERVICE | SESSION POLICY | SEQUENCE | STAGE | STREAM | TABLE | TAG | TASK | VIEW | STREAMLIT

### Category

category:grants

### Object type(s)

resource:database_role

### How much impact is this issue causing?

Medium

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jcieslak:
Hey @anita0918 👋 
I'll create a pr today and It'll be a part of the next release.

sfc-gh-jcieslak:
Hey @anita0918 
The privilege was added in the latest provider version (v0.94.0). Could you please upgrade and confirm it works on your end?

PR #2961, opened 2024-07-25
fix: external function header parsing and add missing privileges

sfc-gh-jcieslak:
Changes:
- Add CREATE NOTEBOOK privilege for on all/future grants (and other missing privileges based on the [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/sql/grant-privilege)): #2960
- Fix header parsing in external function


pkg/resources/external_function_acceptance_test.go
```diff
@@ -391,6 +391,53 @@ func TestAcc_ExternalFunction_issue2528(t *testing.T) {
 	})
 }
 
+// Proves that header parsing handles values wrapped in curly braces, e.g. `value = "{1}"`
+func TestAcc_ExternalFunction_HeaderParsing(t *testing.T) {
```
sfc-gh-asawicki, line 395: is this a result of a particular GH issue? If so, let's at least link it in the PR description

sfc-gh-asawicki (approved):

pkg/resources/external_function_acceptance_test.go
sfc-gh-jcieslak: It was an internal request (no GH issue was created).

sfc-gh-jmichalak (approved):

PR #2968, opened 2024-07-30
feat: Add user parameters to resource

sfc-gh-asawicki:
This is the first of the next few changes being a part of the user resource rework. It includes:
- adding user parameters to the user resource schema
- adding the `parameters` collection to user schema
- testing the behavior of parameters on user
- a bunch of TODOs to move parameters definitions was left for the next PRs (this one is already too big)
- missing assertions and config builders generated

Checks for setting `TIMEZONE` may fail (skip added to the test messing with our account; it was cleaned with `ALTER ACCOUNT UNSET TIMEZONE` but still this test from other branches may interfere).

Additionally:
- logic to handle parameters has been slightly improved and made a bit more generic (with changes to databases and schemas; function to read database parameters was replaced with a more generic solution fixing SNOW-1480106)
- one of the account parameters was messing with the change, so the test was temporarily skipped
- new github issues added to the known issues list

It does not include (it will be handled in the following PRs):
- checking quotes for `DATE_INPUT_FORMAT` and `DATE_OUTPUT_FORMAT`
- checking if zero is allowed for `CLIENT_MEMORY_LIMIT`
- handling `snowflake_network_policy_attachment`
- handling `snowflake_user_public_keys`
- setting diff suppressions and validations for the parameters (it will be handled in the next PR)
- moving the parameters logic fully to the SDK
- reworking other attributes on the user resource
- adding more acceptance tests


docs/resources/user.md
```diff
@@ -43,26 +43,879 @@ resource "snowflake_user" "user" {
 
 ### Optional
 
+- `abort_detached_query` (Boolean) Specifies the action that Snowflake performs for in-progress queries if connectivity is lost due to abrupt termination of a session (e.g. network outage, browser termination, service interruption). For more information, check [ABORT_DETACHED_QUERY docs](https://docs.snowflake.com/en/sql-reference/parameters#abort-detached-query).
```
sfc-gh-jcieslak, line 46: Add custom template for user with the V1 candidate message

pkg/resources/user.go
```diff
@@ -126,23 +128,47 @@ var userSchema = map[string]*schema.Schema{
...
...s.ToLower(string(sdk.UserParameterTimestampLtzOutputFormat)), strings.ToLower(string(sdk.UserParameterTimestampNtzOutputFormat)), strings.ToLower(string(sdk.UserParameterTimestampOutputFormat)), strings.ToLower(string(sdk.UserParameterTimestampTypeMapping)), strings.ToLower(string(sdk.UserParameterTimestampTzOutputFormat)), strings.ToLower(string(sdk.UserParameterTimezone)), strings.ToLower(string(sdk.UserParameterTimeInputFormat)), strings.ToLower(string(sdk.UserParameterTimeOutputFormat)), strings.ToLower(string(sdk.UserParameterTraceLevel)), strings.ToLower(string(sdk.UserParameterTransactionAbortOnError)), strings.ToLower(string(sdk.UserParameterTransactionDefaultIsolationLevel)), strings.ToLower(string(sdk.UserParameterTwoDigitCenturyStart)), strings.ToLower(string(sdk.UserParameterUnsupportedDdlAction)), strings.ToLower(string(sdk.UserParameterUseCachedResult)), strings.ToLower(string(sdk.UserParameterWeekOfYearPolicy)), strings.ToLower(string(sdk.UserParameterWeekStart)), strings.ToLower(string(sdk.UserParameterEnableUnredactedQuerySyntaxError)), strings.ToLower(string(sdk.UserParameterNetworkPolicy)), strings.ToLower(string(sdk.UserParameterPreventUnloadToInternalStages))),
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Yes.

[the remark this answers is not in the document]
sfc-gh-jmichalak: Can this be reused/moved to pkg/schemas/user_parameters.go?

pkg/resources/user_parameters_test.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_UserParametersSchema(t *testing.T) {
+	t.Run("description references parameter docs correctly", func(t *testing.T) {
+		require.True(t, strings.HasSuffix(userParametersSchema["abort_detached_query"].Description, "For more information, check [ABORT_DETACHED_QUERY docs](https://docs.snowflake.com/en/sql-reference/parameters#abort-detached-query)."))
```
sfc-gh-jmichalak, line 12: Do we really need that? It's almost the same like pkg/resources/resource_parameters_commons_test.go

pkg/resources/user_parameters.go
```diff
@@ -0,0 +1,399 @@
...
+	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
+	userParameters, err := client.Parameters.ShowParameters(ctx, &sdk.ShowParametersOptions{
+		In: &sdk.ParametersIn{
+			User: id,
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+	return userParameters, nil
```
sfc-gh-jmichalak, line 179:
```suggestion
	return client.Parameters.ShowParameters(ctx, &sdk.ShowParametersOptions{
		In: &sdk.ParametersIn{
			User: id,
		},
	})
```
BTW, I think these parameter providers could be in SDK - even as a `Users` method. WDYT?

sfc-gh-jmichalak (changes requested):

pkg/resources/user.go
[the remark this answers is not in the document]
sfc-gh-asawicki: Yes, they will be fully moved to the SDK (as stated in the PR description and the TODO in `user_parameters.go` - `// TODO [SNOW-1348101][next PR]: move to the SDK`).

pkg/resources/user_parameters.go
sfc-gh-asawicki: Yep, we could add them (however, for the generated objects we will have to modify the generator). I will add a TODO for the generator and I will add such method to Users, Databases, and Schemas.

pkg/resources/user_parameters_test.go
sfc-gh-asawicki: Later we won't, but currently these tests are different:
- this one validates that the info about param docs is added
- the second one you mention validates ho the message is constructed.

You are right they are pretty similar and the second one could be removed. They don't hurt us much now and I would like to have later a generic solution transforming the SDK definition into tf schema, then we will probably change this test a bit.

WDYT?

pkg/resources/user.go
```diff
@@ -213,83 +242,119 @@ func CreateUser(d *schema.ResourceData, meta interface{}) error {
...
+		client := meta.(*provider.Context).Client
+		// We use User.Describe instead of User.Show because the "SHOW USERS ..." command
+		// requires the "MANAGE GRANTS" global privilege
+		id := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
+		user, err := client.Users.Describe(ctx, id)
+		if err != nil {
+			if errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) {
+				log.Printf("[DEBUG] user (%s) not found or we are not authorized. Err: %s", d.Id(), err)
+				d.SetId("")
+				return nil
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Added

docs/resources/user.md
sfc-gh-asawicki: It's not yet the rework of the whole user resource (this implementation was there before).

Changed nevertheless (had to leave the condition to avoid different behavior before the redesign).

pkg/resources/user.go
```diff
@@ -213,83 +242,119 @@ func CreateUser(d *schema.ResourceData, meta interface{}) error {
...
+		if err := setBoolProperty(d, "disabled", user.Disabled); err != nil {
+			return diag.FromErr(err)
+		}
+		if err := setStringProperty(d, "default_role", user.DefaultRole); err != nil {
+			return diag.FromErr(err)
+		}
+
+		var defaultSecondaryRoles []string
+		if user.DefaultSecondaryRoles != nil && len(user.DefaultSecondaryRoles.Value) > 0 {
+			defaultRoles, _ := strings.CutPrefix(user.DefaultSecondaryRoles.Value, "[\"")
```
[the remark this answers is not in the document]
sfc-gh-asawicki: It's not yet the rework of the whole user resource (this implementation was there before).

Changed nevertheless (had to leave the condition to avoid different behavior before the redesign).

pkg/resources/user_parameters.go
sfc-gh-asawicki: Done (for warehouses too)

sfc-gh-jcieslak (dismissed):

sfc-gh-jcieslak (dismissed):

templates/resources/user.md.tmpl
```diff
@@ -0,0 +1,32 @@
...
+subcategory: ""
+description: |-
+{{ if gt (len (split .Description "<deprecation>")) 1 -}}
+{{ index (split .Description "<deprecation>") 1 | plainmarkdown | trimspace | prefixlines "  " }}
+{{- else -}}
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+{{- end }}
+---
+
+!> **V1 release candidate** This resource was reworked and is a release candidate for the V1. We do not expect significant changes in it before the V1. We will welcome any feedback and adjust the resource if needed. Any errors reported will be resolved with a higher priority. We encourage checking this resource out before the V1 release. Please follow the [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v0920--v0930) to use it.
```
sfc-gh-jmichalak, line 12: Wrong version?

sfc-gh-jmichalak (approved):

templates/resources/user.md.tmpl
sfc-gh-asawicki: Ah yeah, I will change in the next PR, good catch

sfc-gh-asawicki: Corrected now

pkg/sdk/parsers_test.go
```diff
@@ -61,11 +61,41 @@ func TestParseCommaSeparatedStringArray(t *testing.T) {
...
 		},
+		{
+			Name:       "list without brackets - with double quotes",
+			Value:      `"one","two","three"`,
+			TrimQuotes: true,
+			Result:     []string{"one", "two", "three"},
+		},
+		{
+			Name:       "list with brackets - with double quotes",
+			Value:      `"one","two","three"`,
```
sfc-gh-jcieslak, line 77: This one should have brackets (same in the next test case)

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (approved):

pkg/sdk/parsers_test.go
sfc-gh-asawicki: I will change the name of the test in the follow-up PR

Issue #2826, opened 2024-05-20
[Bug]: Cannot configure PUBLIC schema of new database


wesleyhillyext:
### Terraform CLI Version

1.7.4

### Terraform Provider Version

0.88.0

### Terraform Configuration

```terraform
resource "snowflake_database" "my_database" {
  name = "my_database"
}

resource "snowflake_schema" "my_database_public" {
  database = snowflake_database.my_database.name
  name     = "PUBLIC"

  is_managed          = true
  data_retention_days = 30
}
```

### Category

category:resource

### Expected Behavior

When applying these resources for the first time, the database is created by the `snowflake_database` resource. Then the `PUBLIC` schema for that database is created by the  `snowflake_schema` resource.

### Actual Behavior

The database is created by the `snowflake_database` resource. The `snowflake_schema` resource then fails to apply when it runs `CREATE SCHEMA PUBLIC`, which fails because the schema already exists, since in Snowflake a `CREATE DATABASE` automatically creates a `PUBLIC` schema with default properties.

### Steps to Reproduce

1. `terraform apply` the example configuration when the resources don't already exist in the TF state or in Snowflake.

### How much impact is this issue causing?

Medium

### Additional Information

This reopens #306 with some additional thinking.

This has been a thorn in our side when creating new databases, as we either want to have an explicit `snowflake_schema` block for the `PUBLIC` schema, or in some cases don't want a `PUBLIC` schema for a specific database.

**Proposed solution**: After considerable thinking on this issue, I've come to believe there's only one workable solution: add a boolean option to the `snowflake_database` resource which, if enabled, causes it to execute `DROP SCHEMA PUBLIC` if it executes a `CREATE DATABASE`.

Alternatives considered and dismissed:

1. Apply once, manually drop the `PUBLIC` schema when the `snowflake_schema` resource fails, then apply again.
- This is our current workaround, but the manual step prevents deployment of Snowflake terraform through a CI pipeline.
2. Allow configuration of the `PUBLIC` schema in the `snowflake_database` resource.
- This doesn't address the case when you don't want a `PUBLIC` schema.
- Would mean that different schemas would be configured different ways, which is undesireable.
3. Use `import` blocks to automatically import the `PUBLIC` schema.
- This doesn't address the case when you don't want a `PUBLIC` schema.
- The `import` would have to be executed after the `snowflake_database` resource is created, and it's not clear that Terraform allows specifying this kind of dependency.
[...]

PR #2973, opened 2024-08-01
fix: Use ALTER for managing PUBLIC schemas that exist

sfc-gh-jmichalak:
- during create operation, detect if PUBLIC schema has already been created, and instead of using OR REPLACE, redirect the flow to `UpdateContextSchema` (note that `name` change needs to be excluded to avoid unnecessary rename)
- add a check to acceptance tests to ensure the schema hasn't been dropped 
- describe new behavior in the migration guide for v0.94.1 and add a note for v0.94.0

## Test Plan

* [x] acceptance tests

## References

References https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2826, specifically comment https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2826#issuecomment-2258982446


MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,11 @@ This document is meant to help you migrate your Terraform config to the new newe
 describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behavior
 across different versions.
 
+## v0.94.0 ➞ v0.94.1
+### *(breaking change)* changes in snowflake_schema
```
sfc-gh-asawicki, line 8: It is not a breaking change IMO

MIGRATION_GUIDE.md
```diff
@@ -44,7 +49,7 @@ New fields:
 - added `describe_output` field that holds the response from DESCRIBE SCHEMA. Note that one needs to grant sufficient privileges e.g. with [grant_ownership](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_ownership) on all objects in the schema. Otherwise, this field is not filled.
 - added `parameters` field that holds the response from SHOW PARAMETERS IN SCHEMA.
 
-We allow creating and managing `PUBLIC` schemas now. When the name of the schema is `PUBLIC`, it's created with `OR_REPLACE`. We've decided this based on [#2826](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2826).
+We allow creating and managing `PUBLIC` schemas now. When the name of the schema is `PUBLIC`, it's created with `OR_REPLACE`. Please be careful with this operation, because you may experience data loss. The alternative is to import `PUBLIC` schema manually and then manage it with Terraform. We've decided this based on [#2826](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2826).
```
sfc-gh-asawicki, line 52: Please elaborate a bit more on the data loss (i.e. REPLACE does DROP underneath, tf plan will just list this as create, link to time-travel)

sfc-gh-asawicki: Also, did you validate that importing works in the older/currentversion?

pkg/resources/schema.go
```diff
@@ -213,73 +213,90 @@ func CreateContextSchema(ctx context.Context, d *schema.ResourceData, meta any)
...
-		userTaskMinimumTriggerIntervalInSeconds,
-		quotedIdentifiersIgnoreCase,
-		enableConsoleOutput,
-		err := GetAllDatabaseParameters(d)
-	if err != nil {
-		return diag.FromErr(err)
-	}
+	create := true
 
-	opts := &sdk.CreateSchemaOptions{
```
sfc-gh-asawicki, line 237: oh noooo, this will conflict with my change :(

pkg/resources/schema.go
```diff
@@ -213,73 +213,90 @@ func CreateContextSchema(ctx context.Context, d *schema.ResourceData, meta any)
...
-		userTaskManagedInitialWarehouseSize,
-		userTaskTimeoutMs,
-		userTaskMinimumTriggerIntervalInSeconds,
-		quotedIdentifiersIgnoreCase,
-		enableConsoleOutput,
-		err := GetAllDatabaseParameters(d)
-	if err != nil {
-		return diag.FromErr(err)
-	}
+	create := true
```
sfc-gh-asawicki, line 216: TBH I would prefer an explicit separation here, i.e.:
- if name is not PUBLIC -> run create
- otherwise if does not exist -> run create
- otherwise -> run update

pkg/resources/schema_acceptance_test.go
```diff
@@ -501,6 +502,28 @@ func TestAcc_Schema_ManagePublic(t *testing.T) {
 					resource.TestCheckResourceAttr("snowflake_schema.test", "pipe_execution_paused", "false"),
 				),
 			},
+			// make sure the schema wasn't dropped
```
sfc-gh-jcieslak, line 505: I think the steps before this one refer to the 0.94.0 version

sfc-gh-asawicki, line 505: why we calidate it on the import step and not on the first step after bumping the version? Also, shouldn't we check 0.94.0 -> 0.94.1 transition?

sfc-gh-asawicki (changes requested):

pkg/resources/schema_acceptance_test.go
sfc-gh-jmichalak: Changed - now tests are separated for v0.94.0 and v0.94.1.

pkg/resources/schema.go
sfc-gh-jmichalak: Simplified a bit.

pkg/resources/schema.go
sfc-gh-jmichalak: :(

MIGRATION_GUIDE.md
sfc-gh-jmichalak: IMO it is - we're changing current behavior so that objects are not dropped.

pkg/resources/schema_acceptance_test.go
sfc-gh-jmichalak: Specified 0.94 for the steps above.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: Added more info to the doc. Regarding import, I tested it manually and it works, but I can't setup automatic tests to import in v0.93.0. As we discussed, this may be because of separate state for versioned providers in the tests. Will create a follow-up ticket for this.

pkg/resources/schema_acceptance_test.go
```diff
@@ -470,27 +470,39 @@ func TestAcc_Schema_ManagePublic(t *testing.T) {
...
+					// In v0.94 `CREATE OR REPLACE` was called, so we should see a DROP event.
+					schemas := acc.TestClient().Schema.ShowWithOptions(t, &sdk.ShowSchemaOptions{
+						History: sdk.Pointer(true),
+						Like: &sdk.Like{
+							Pattern: sdk.String(schemaId.Name()),
+						},
+					})
+					require.Len(t, schemas, 2)
+					require.Zero(t, schemas[0].DroppedOn)
+					require.NotZero(t, schemas[1].DroppedOn)
```
sfc-gh-jcieslak, line 503: nit: Not ideal, because we rely on Snowflake's output ordering

sfc-gh-jcieslak (approved):

MIGRATION_GUIDE.md
sfc-gh-asawicki: And how this is breaking anything from the user's perspective? And if you want it to be a breaking, you cannot release it as bugfix. I would say, this is an implementation detail in this case.

pkg/resources/schema.go
sfc-gh-asawicki: It's not a separation like I asked for. Now the execution path is still not clear.

Let's leave it like this, and will follow up and do it in the following PRs. I will keep it open for the time-being.

MIGRATION_GUIDE.md

pkg/resources/schema_acceptance_test.go
sfc-gh-jmichalak: Added sorting.

pkg/resources/schema.go
```diff
@@ -213,88 +213,81 @@ func CreateContextSchema(ctx context.Context, d *schema.ResourceData, meta any)
...
-			traceLevel,
-			suspendTaskAfterNumFailures,
-			taskAutoRetryAttempts,
-			userTaskManagedInitialWarehouseSize,
-			userTaskTimeoutMs,
-			userTaskMinimumTriggerIntervalInSeconds,
-			quotedIdentifiersIgnoreCase,
-			enableConsoleOutput,
-			err := GetAllDatabaseParameters(d)
+	// there is no PUBLIC schema, so we continue with creating
```
sfc-gh-asawicki, line 227: nit: or the name is completely different than PUBLIC, we will handle this in the next PR

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

PR #2979, opened 2024-08-05
feat: Add identifier with arguments

sfc-gh-jcieslak:
## Changes
- Add a new identifier type that supports arguments
- Adjust the function's SDK/resource/data source implementation and tests, so that it uses the new identifier

note: Create signature wasn't changed, because it contains more info like argument name or default value.

## TODOs
- Remove additional validation from ParseIdentifierString
- Another function identifier representation to parse (SHOW GRANTS)

## Next pr(s)
- Use in procedure and external function (SDK/resources/data sources)
- Remove arguments from SchemaObjectIdentifier (and all helper functions regarding SchemaObjectIdentifier with arguments)


pkg/sdk/data_types.go
```diff
@@ -89,6 +89,11 @@ func ToDataType(s string) (DataType, error) {
 		return DataTypeTime, nil
 	}
 
+	vectorSynonyms := []string{"VECTOR"}
```
sfc-gh-asawicki, line 92: there is a data_types_test.go and as I see it was not added there.

pkg/schemas/gen/main/main.go
```diff
@@ -8,6 +8,8 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/genhelpers"
```
sfc-gh-asawicki, line 11: why this import is added here?

pkg/sdk/poc/main.go
```diff
@@ -10,6 +10,8 @@ import (
 	"os"
 	"strings"
 
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/genhelpers"
```
sfc-gh-asawicki, line 13: why this import is added here?

pkg/resources/external_function_acceptance_test.go
```diff
@@ -2,16 +2,14 @@ package resources_test
...
-
 	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
-
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+	"testing"
```
sfc-gh-asawicki, line 12: this should be raised as not gofumpted (this import should be with other internal imports)

pkg/resources/external_function_acceptance_test.go
```diff
@@ -482,80 +480,80 @@ resource "snowflake_external_function" "f" {
...
-  schema   = "%[4]s"
-  arg {
-    name = "SNS_NOTIF"
-    type = "OBJECT"
-  }
-  return_type = "VARIANT"
-  return_behavior = "VOLATILE"
-  api_integration = snowflake_api_integration.test_api_int.name
-  url_of_proxy_and_resource = "https://123456.execute-api.us-west-2.amazonaws.com/prod/test_func"
+ depends_on = [snowflake_schema.s2]
```
sfc-gh-asawicki, line 510: was this whitespace change necessary?

pkg/sdk/functions_impl_gen.go
```diff
@@ -371,14 +373,23 @@ func (r functionRow) convert() *Function {
...
+		ArgumentsRaw:       r.Arguments,
 		Description:        r.Description,
 		CatalogName:        r.CatalogName,
 		IsTableFunction:    r.IsTableFunction == "Y",
 		ValidForClustering: r.ValidForClustering == "Y",
 		IsExternalFunction: r.IsExternalFunction == "Y",
 		Language:           r.Language,
 	}
+	arguments := strings.TrimLeft(r.Arguments, r.Name)
+	returnIndex := strings.Index(arguments, ") RETURN ")
```
sfc-gh-asawicki, line 385: does all functions have return? (asking, I don't know)

Also, what about the whitespace - can it be different?

pkg/sdk/identifier_helpers.go
```diff
@@ -333,6 +334,79 @@ func (i SchemaObjectIdentifier) ArgumentsSignature() string {
...
+		name:              strings.Trim(name, `"`),
+		argumentDataTypes: argumentDataTypes,
+	}
+}
+
+func NewSchemaObjectIdentifierWithArgumentsInSchema(schemaId DatabaseObjectIdentifier, name string, argumentDataTypes ...DataType) SchemaObjectIdentifierWithArguments {
+	return NewSchemaObjectIdentifierWithArguments(schemaId.DatabaseName(), schemaId.Name(), name, argumentDataTypes...)
+}
+
+func NewSchemaObjectIdentifierWithArgumentsFromFullyQualifiedName(fullyQualifiedName string) (SchemaObjectIdentifierWithArguments, error) {
```
sfc-gh-asawicki, line 357: THis one is a general question for the identifiers rework and the family of `NewXxxIdentifierFromFullyQualifiedName` -> should we have the if we prepared `ParseXxxIdentifier` helper functions?

pkg/resources/function_acceptance_test.go
```diff
@@ -231,6 +236,40 @@ func TestAcc_Function_migrateFromVersion085(t *testing.T) {
 	})
 }
 
+func TestAcc_Function_Version0941_EnsureSmoothResourceIdMigration(t *testing.T) {
```
sfc-gh-asawicki, line 239: 0950?

pkg/sdk/identifier_helpers_test.go
```diff
@@ -82,3 +83,42 @@ func TestDatabaseObjectIdentifier(t *testing.T) {
...
+		RawInput string
+		Input    SchemaObjectIdentifierWithArguments
+		Error    string
+	}{
+		{Input: NewSchemaObjectIdentifierWithArguments(`abc`, `def`, `ghi`, DataTypeFloat, DataTypeNumber, DataTypeTimestampTZ)},
+		{Input: NewSchemaObjectIdentifierWithArguments(`abc`, `def`, `ghi`, DataTypeFloat, "VECTOR(INT, 20)")},
+		{Input: NewSchemaObjectIdentifierWithArguments(`abc`, `def`, `ghi`, "VECTOR(INT, 20)", DataTypeFloat)},
+		{Input: NewSchemaObjectIdentifierWithArguments(`abc`, `def`, `ghi`, DataTypeFloat, "VECTOR(INT, 20)", "VECTOR(INT, 10)")},
+		{Input: NewSchemaObjectIdentifierWithArguments(`abc`, `def`, `ghi`, DataTypeTime, "VECTOR(INT, 20)", "VECTOR(FLOAT, 10)", DataTypeFloat)},
+		// TODO(): Won't work, because of the assumption that identifiers are not containing '(' and ')' parentheses
```
sfc-gh-asawicki, line 98: what about htis TODO?

pkg/sdk/identifier_helpers_test.go
```diff
@@ -82,3 +83,42 @@ func TestDatabaseObjectIdentifier(t *testing.T) {
...
+		{Input: NewSchemaObjectIdentifierWithArguments(`abc`, `def`, `ghi`)},
+		{Input: NewSchemaObjectIdentifierWithArguments(`abc`, `def`, `ghi`), RawInput: `abc.def.ghi()`},
+		{Input: NewSchemaObjectIdentifierWithArguments(`abc`, `def`, `ghi`, DataTypeFloat, "VECTOR(INT, 20)"), RawInput: `abc.def.ghi(FLOAT, VECTOR(INT, 20))`},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(fmt.Sprintf("processing %s", testCase.Input.FullyQualifiedName()), func(t *testing.T) {
+			var id SchemaObjectIdentifierWithArguments
+			var err error
+			if testCase.RawInput != "" {
```
sfc-gh-asawicki, line 110: This comment I have already left several times: it's better not to have too much conditional logic in the parametrized tests, it's error-prone and harder to read. It's better to split into several parametrized ones, here e.g. to input, raw input, and invalid.

pkg/sdk/identifier_parsers_test.go
```diff
@@ -250,3 +271,31 @@ func Test_ParseObjectIdentifierString(t *testing.T) {
...
 }
+
+func Test_ParseFunctionArgumentsFromString(t *testing.T) {
+	testCases := []struct {
+		Arguments string
+		Expected  []DataType
+		Error     string
+	}{
+		{Arguments: `()`, Expected: []DataType{}},
+		{Arguments: `(FLOAT, NUMBER, TIME)`, Expected: []DataType{DataTypeFloat, DataTypeNumber, DataTypeTime}},
```
sfc-gh-asawicki, line 282: please add a case when e.g. VARCHAR(200) is used

pkg/sdk/identifier_parsers_test.go
```diff
@@ -250,3 +271,31 @@ func Test_ParseObjectIdentifierString(t *testing.T) {
...
+		{Arguments: `FLOAT, NUMBER, TIME`, Expected: []DataType{DataTypeFloat, DataTypeNumber, DataTypeTime}},
+		{Arguments: `(FLOAT, NUMBER, VECTOR(FLOAT, 20))`, Expected: []DataType{DataTypeFloat, DataTypeNumber, DataType("VECTOR(FLOAT, 20)")}},
+		{Arguments: `FLOAT, NUMBER, VECTOR(FLOAT, 20)`, Expected: []DataType{DataTypeFloat, DataTypeNumber, DataType("VECTOR(FLOAT, 20)")}},
+		{Arguments: `(VECTOR(FLOAT, 10), NUMBER, VECTOR(FLOAT, 20))`, Expected: []DataType{DataType("VECTOR(FLOAT, 10)"), DataTypeNumber, DataType("VECTOR(FLOAT, 20)")}},
+		{Arguments: `VECTOR(FLOAT, 10)| NUMBER, VECTOR(FLOAT, 20)`, Error: "expected a comma delimited string but found |"},
```
sfc-gh-asawicki, line 287: please add a case with incorrect VECTOR, like VECTOR(VARCHAR, 5), and/or VECTOR(5, FLOAT)

pkg/sdk/testint/functions_integration_test.go
```diff
@@ -236,46 +235,45 @@ func TestInt_OtherFunctions(t *testing.T) {
...
-	}
-
 	t.Run("alter function: rename", func(t *testing.T) {
 		f := createFunctionForSQLHandle(t, false, true)
 
 		id := f.ID()
-		nid := testClientHelper().Ids.RandomSchemaObjectIdentifier()
-		err := client.Functions.Alter(ctx, defaultAlterRequest(id).WithRenameTo(&nid))
+		nid := testClientHelper().Ids.RandomSchemaObjectIdentifierWithArguments()
+		err := client.Functions.Alter(ctx, func(id sdk.SchemaObjectIdentifierWithArguments) *sdk.AlterFunctionRequest {
```
sfc-gh-asawicki, line 270: reviewdog

pkg/sdk/testint/functions_integration_test.go
```diff
@@ -291,7 +289,9 @@ func TestInt_OtherFunctions(t *testing.T) {
 		f := createFunctionForSQLHandle(t, true, true)
 
 		id := f.ID()
-		err := client.Functions.Alter(ctx, defaultAlterRequest(id).WithSetLogLevel(sdk.String("DEBUG")))
+		err := client.Functions.Alter(ctx, func(id sdk.SchemaObjectIdentifierWithArguments) *sdk.AlterFunctionRequest {
```
sfc-gh-asawicki, line 292: reviewdog

pkg/sdk/testint/functions_integration_test.go
```diff
@@ -300,7 +300,9 @@ func TestInt_OtherFunctions(t *testing.T) {
 		f := createFunctionForSQLHandle(t, true, true)
 
 		id := f.ID()
-		err := client.Functions.Alter(ctx, defaultAlterRequest(id).WithUnsetLogLevel(sdk.Bool(true)))
+		err := client.Functions.Alter(ctx, func(id sdk.SchemaObjectIdentifierWithArguments) *sdk.AlterFunctionRequest {
```
sfc-gh-asawicki, line 303: reviewdog

pkg/sdk/testint/functions_integration_test.go
```diff
@@ -457,47 +471,109 @@ func TestInt_FunctionsShowByID(t *testing.T) {
...
 		e2, err := client.Functions.ShowByID(ctx, id2)
 		require.NoError(t, err)
-		require.Equal(t, id2, e2.ID())
+
+		e2Id := e2.ID()
+		require.NoError(t, err)
+		require.Equal(t, id2, e2Id)
+	})
+
+	t.Run("function returns non detailed data types of arguments", func(t *testing.T) {
```
sfc-gh-asawicki, line 524: I really like this test :)

sfc-gh-asawicki (changes requested):

pkg/sdk/data_types.go
sfc-gh-jcieslak: True, I missed that somehow. Adding test cases.

pkg/schemas/gen/main/main.go
sfc-gh-jcieslak: As I remember the build was failing because of it. It's used in the main function to call `NewGenerator`.

pkg/sdk/poc/main.go
sfc-gh-jcieslak: Same reason as above (it's used at the bottom of this file).

pkg/resources/external_function_acceptance_test.go
sfc-gh-jcieslak: I'm not sure why that happened, but I'm guessing goland or some go formatter thought it's better that way 🤷

pkg/resources/external_function_acceptance_test.go
sfc-gh-jcieslak: I'm pretty sure I ran `pre-push` 🤦

pkg/sdk/functions_impl_gen.go
sfc-gh-jcieslak: Yes, functions always have to return something

pkg/sdk/identifier_helpers_test.go
sfc-gh-jcieslak: Forgot to add ticket number, for now It's another limitation that we have to define to move forward, but eventually get back to it (imo it's a bigger analysis, discussion, change to think if csv parser is enough for all Snowflake identifiers and their different versions returned by SHOW or DESC commands).

pkg/sdk/identifier_helpers.go
sfc-gh-jcieslak: Not fully understand 😅, but yes there should be ParseXxxIdentifier for the new Identifier and it should be used in this function. I'll write it in the follow up pr.

pkg/sdk/identifier_parsers.go
```diff
@@ -37,6 +38,10 @@ func ParseIdentifierString(identifier string) ([]string, error) {
 		if strings.Contains(part, `"`) {
 			return nil, fmt.Errorf(`unable to parse identifier: %s, currently identifiers containing double quotes are not supported in the provider`, identifier)
 		}
+		// TODO(SNOW-1571674): Remove the validation
+		if strings.ContainsAny(part, `()`) {
+			return nil, fmt.Errorf(`unable to parse identifier: %s, currently identifiers containing opening and closing parentheses '()' are not supported in the provider`, identifier)
+		}
```
sfc-gh-jmichalak, line 44: Do we have documented such behaviors? I think it would be nice to keep them in one place that users can easily access.

pkg/sdk/identifier_parsers.go
```diff
@@ -140,3 +145,50 @@ func ParseExternalObjectIdentifier(identifier string) (ExternalObjectIdentifier,
...
+
+	for stringBuffer.Len() > 0 {
+		// We use another buffer to peek into next data type (needed for vector parsing)
+		peekDataType, _ := bytes.NewBufferString(stringBuffer.String()).ReadString(',')
+		peekDataType = strings.TrimSpace(peekDataType)
+
+		switch {
+		// For now, only vectors need special parsing behavior
+		case strings.HasPrefix(peekDataType, "VECTOR"):
+			vectorDataType, _ := stringBuffer.ReadString(')')
```
sfc-gh-jmichalak, line 170: Why is this error not handled?

sfc-gh-jcieslak: As documentation says, the only error we can get here is EOF which also happens when the whole buffer is read, so let's say we have `(abc)` and you want to call `stringBuffer.ReadString(')')` it will return EOF because the function reached the end of buffer. That's why I decided to skip them, because there's no other error than EOF and it happens in cases where I don't expect it to. I'll leave additional comments for context.

pkg/sdk/identifier_parsers.go
sfc-gh-jcieslak: True, I'll create a ticket for that to sum up the identifiers rework in our changes before v1 GH document.

sfc-gh-jcieslak: Update: created SNOW-1616523

pkg/sdk/identifier_parsers.go
sfc-gh-jcieslak: Actually, I was wrong the error is only returned if the returned string is not ending with the expected delimiter. I'll add err checks and test cases

pkg/sdk/identifier_helpers.go
sfc-gh-jcieslak: Used ParseXxxIdentifier and removed NewXxxIdentifierFromFullyQualifiedName for the new identifier

pkg/sdk/identifier_helpers_test.go
sfc-gh-jcieslak: Moved tests to `identifier_parsers_test.go` (as we're using ParseXxxIdentifier now) and split this test into two

pkg/sdk/identifier_parsers_test.go
sfc-gh-jcieslak: Added additional validation for vectors and added test cases for them

pkg/sdk/identifier_parsers_test.go
sfc-gh-jcieslak: Added test cases (also added cases like `Number(10, 2)` and explained it's not supported for now as it's not needed for any use case)

sfc-gh-jmichalak (dismissed):

pkg/sdk/identifier_helpers_test.go
sfc-gh-jmichalak: SNOW-1571674

sfc-gh-jmichalak (approved):

Issue #2978, opened 2024-08-05
[Bug]: snowflake_database with the from_database won't migrate automatically for 0.93


kamilamarcinekpgs:
### Terraform CLI Version

1.2.6

### Terraform Provider Version

0.93

### Terraform Configuration

```terraform
resource "snowflake_database" "test" {
  name = "VERY_IMPORTANT_DB"
  data_retention_time_in_days = 7
  from_database = "THE_MODEL_DB"
}
```

### Category

category:resource

### Object type(s)

resource:database

### Expected Behavior

I'm trying to upgrade from 0.92 to 0.93.

I have the old `snowflake_database resource`, with the from_database parameter set.

The migration guide says: 

> If you had from_database set, it should migrate automatically.

### Actual Behavior

I'm not sure what it means, that it should migrate automatically. 
If I just change the version, it says "**Unsupported argument**". 
If I change the version and remove the `from_database param`, it says "**failed to upgrade the state with database created from database**".

What should I do to achieve the "automatic upgrade" mentioned in the migration guide?

### Steps to Reproduce

.

### How much impact is this issue causing?

Medium

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jcieslak:
Hey @kamilamarcinekpgs 
That's a documentation error; I'll fix it. As `from_database` is database cloning, which is not currently supported, the only way is to remove the database from the state `terraform state rm` -> upgrade to the newer version -> import the database.

sfc-gh-jcieslak:
Closing due to long inactivity.

PR #2981, opened 2024-08-06
fix: database tests and introduce a new parameter

sfc-gh-jcieslak:
## Changes
- #2978 (adjust migration guide)
- Fix database tests (remove tests for state upgraders, because they were incorrect and document step-by-step process for manual testing; I went through all of it)
- Introduce a new parameter to address #2826 and test it
  - added a drop with retries just to be sure, but I don't think it should ever fail other than some kind of strange networking error between database create and drop schema
  - I opted to drop the schema after setting id, so if we manage to create a database, but not drop the schema, the provider will print out a warning message that the public schema couldn't be dropped and it should be done manually. The other approaches I considered:
    - Do it before setting ID, but that would mean the Terraform running create again and failing on CREATE DATABASE (because it's already there).
    - Do it like right now, but make sure the public schema is gone in the alter, but I didn't go for that, because it could lead to further complications with (what is considered "after create"; after successful create? how would we keep the state of create status? etc.).


pkg/resources/database.go
```diff
@@ -22,6 +25,11 @@ var databaseSchema = map[string]*schema.Schema{
 		Required:    true,
 		Description: "Specifies the identifier for the database; must be unique for your account. As a best practice for [Database Replication and Failover](https://docs.snowflake.com/en/user-guide/db-replication-intro), it is recommended to give each secondary database the same name as its primary database. This practice supports referencing fully-qualified objects (i.e. '<db>.<schema>.<object>') by other objects in the same database, such as querying a fully-qualified table name in a view. If a secondary database has a different name from the primary database, then these object references would break in the secondary database.",
 	},
+	"drop_public_schema_on_creation": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Description: "Specifies whether to drop public schema on creation or not. Modifying the parameter after database is already created won't have any effect.",
```
sfc-gh-jmichalak, line 31: Can we validate that if this field is changed, an error/warning is returned saying that we don't support it? We can also use IgnoreAfterCreation, like in warehouses.

pkg/resources/database_acceptance_test.go
```diff
@@ -1160,233 +1161,232 @@ resource "snowflake_database" "test" {
...
-				Config:                   databaseStateUpgraderFromShareNewAfterUpgrade(id),
-				ExpectError:              regexp.MustCompile("failed to upgrade the state with database created from share, please use snowflake_shared_database or deprecated snowflake_database_old instead"),
-			},
-			{
-				ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
-				Config:                   databaseStateUpgraderFromShareNew(id, shareExternalId),
-				ResourceName:             "snowflake_shared_database.test",
-				ImportStateId:            id.FullyQualifiedName(),
-				ImportState:              true,
+				Config: databaseWithPublicSchemaConfig(id, true),
```
sfc-gh-jmichalak, line 1184: Nit: add a check that plan is noop.

pkg/resources/database_acceptance_test.go
```diff
@@ -1160,233 +1161,232 @@ resource "snowflake_database" "test" {
...
+	return func(state *terraform.State) error {
+		if !slices.ContainsFunc(acc.TestClient().Database.Describe(t, id).Rows, func(row sdk.DatabaseDetailsRow) bool { return row.Name == "PUBLIC" && row.Kind == "SCHEMA" }) {
+			return fmt.Errorf("expected database %s to contain public schema", id.FullyQualifiedName())
+		}
+		return nil
+	}
 }
 
-func databaseStateUpgraderFromReplicaNewAfterUpgrade(id sdk.AccountObjectIdentifier) string {
+func databaseWithPublicSchemaConfig(id sdk.AccountObjectIdentifier, withPublicSchema bool) string {
```
sfc-gh-jmichalak, line 1244:
```suggestion
func databaseWithDropPublicSchemaConfig(id sdk.AccountObjectIdentifier, withDropPublicSchema bool) string {
```

pkg/resources/database.go
```diff
@@ -22,6 +25,11 @@ var databaseSchema = map[string]*schema.Schema{
 		Required:    true,
 		Description: "Specifies the identifier for the database; must be unique for your account. As a best practice for [Database Replication and Failover](https://docs.snowflake.com/en/user-guide/db-replication-intro), it is recommended to give each secondary database the same name as its primary database. This practice supports referencing fully-qualified objects (i.e. '<db>.<schema>.<object>') by other objects in the same database, such as querying a fully-qualified table name in a view. If a secondary database has a different name from the primary database, then these object references would break in the secondary database.",
 	},
+	"drop_public_schema_on_creation": {
```
sfc-gh-asawicki, line 28: let's use `DiffSuppressFunc: IgnoreAfterCreation,`

pkg/resources/database.go
sfc-gh-jcieslak: Used `IgnoreAfterCreation` and validated in tests with noop assert

pkg/resources/database_acceptance_test.go
```diff
@@ -1160,233 +1161,232 @@ resource "snowflake_database" "test" {
...
 	})
 }
 
-func databaseStateUpgraderFromReplicaOld(id sdk.AccountObjectIdentifier, primaryDatabaseId sdk.ExternalObjectIdentifier) string {
-	return fmt.Sprintf(`
-resource "snowflake_database" "test" {
-	name = "%s"
-	data_retention_time_in_days = 0 # to avoid in-place update to -1
-	from_replica = %s
+func doesNotContainPublicSchema(t *testing.T, id sdk.AccountObjectIdentifier) resource.TestCheckFunc {
```
sfc-gh-asawicki, line 1234: We are moving away from these assertions and move to the new ones. I just say it to remind you this. Please keep that in mind in other PRs. 

Here, please at least move these two funcs to `pkg/acceptance/snowflakechecks/database.go` so that we have all the ones that will be replaced in one place

pkg/resources/database_acceptance_test.go
```diff
@@ -1160,233 +1161,232 @@ resource "snowflake_database" "test" {
...
-`, id.Name())
+`, id.Name(), strconv.FormatBool(!withPublicSchema))
 }
 
-func databaseStateUpgraderFromReplicaNew(id sdk.AccountObjectIdentifier, primaryDatabaseId sdk.ExternalObjectIdentifier) string {
-	return fmt.Sprintf(`
-resource "snowflake_secondary_database" "test" {
-	name = "%s"
-	as_replica_of = %s
+/*
```
sfc-gh-asawicki, line 1263: Thanks for addding the manual steps that currently can't be run using this testing framework.

I think that we should no have the directly as comment in the acceptance test but maybe rather let's add manual tests directory (or something similar), let's add a short readme why we do this, and there let's add them as separate files.

WDYT @sfc-gh-jcieslak @sfc-gh-jmichalak ?

sfc-gh-asawicki (changes requested):

pkg/resources/database_acceptance_test.go
sfc-gh-jmichalak: Good idea, maybe some of them could be automatic after upgrading to plugin framework.

sfc-gh-jmichalak (approved):

pkg/resources/database.go
sfc-gh-jcieslak: Yup, used it Kuba's comment

pkg/resources/database_acceptance_test.go
sfc-gh-jcieslak: Yeah, I tried to use the new ones, but here I'm asserting describe output that's not supported in the new assertions (also checked and it's documented in "known limitations").

sfc-gh-jmichalak (approved):

pkg/resources/database_acceptance_test.go
sfc-gh-jcieslak: Added

PR #2996, opened 2024-08-14
chore: Apply identifier conventions

sfc-gh-jcieslak:
## Changes
- Setting resource Id with the new conventions
- Added and used state upgrader for multi-part resource ids
- Using identifier parsers instead of old methods
- Suppressing quoting on identifier-part or whole identifier fields
- Added and used new ComputeIfChanged that takes suppress diff function into account
- Tested smooth migration to the new version with the new resource id
- Tested to see if the quote suppression works correctly

## Next pr
Conventions applied to grant resrources


pkg/resources/account_role.go
```diff
@@ -17,8 +16,9 @@ import (
 
 var accountRoleSchema = map[string]*schema.Schema{
 	"name": {
-		Type:     schema.TypeString,
-		Required: true,
+		Type:             schema.TypeString,
+		Required:         true,
+		DiffSuppressFunc: suppressIdentifierQuoting,
 		// TODO(SNOW-999049): Uncomment once better identifier validation will be implemented
```
sfc-gh-asawicki, line 22: this still points to the identifiers rework epic and we have already a follow-up for this validation. Will all the deprecated TODO links be replaced in this or separate change?

pkg/resources/account_role_acceptance_test.go
```diff
@@ -208,3 +208,75 @@ resource "snowflake_account_role" "role" {
...
+				Config:             accountRoleBasicConfig(quotedId, comment),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_account_role.role", "name", id.Name()),
+					resource.TestCheckResourceAttr("snowflake_account_role.role", "id", id.Name()),
+				),
+			},
+			{
+				ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+				Config:                   accountRoleBasicConfig(quotedId, comment),
+				Check: resource.ComposeAggregateTestCheckFunc(
```
sfc-gh-asawicki, line 275: can we check what actions are happening here? We expect no-op, right? (and this comment is valid for all acceptance tests added)

pkg/resources/custom_diffs.go
```diff
@@ -98,6 +98,21 @@ func ComputedIfAnyAttributeChanged(key string, changedAttributeKeys ...string) s
 	})
 }
 
+func ComputedIfAnyAttributeChangedWithSuppressDiff(key string, suppressDiffFunc schema.SchemaDiffSuppressFunc, changedAttributeKeys ...string) schema.CustomizeDiffFunc {
```
sfc-gh-asawicki, line 101: (just comment, we will address this in a separate PR) the interface of this method is a bit strange, I would expect that we can pass key + its suppress diff func. Alternatively, we could just pass the keys and check their suppress diff in schema - which would allow us adding this logic to the existing `ComputedIfAnyAttributeChanged` func -> if key has suppress dif, run it, if not, just check. Please link this comment to the created issue and also add a TODO with the issue number here.

pkg/resources/schema_acceptance_test.go
```diff
@@ -170,10 +171,10 @@ func TestAcc_Schema_basic(t *testing.T) {
 				ResourceName:    "snowflake_schema.test",
 				ImportState:     true,
 				ImportStateCheck: importchecks.ComposeAggregateImportStateCheck(
-					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "name", id.Name()),
-					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "database", databaseId.Name()),
-					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "with_managed_access", "false"),
-					importchecks.TestCheckResourceAttrInstanceState(helpers.EncodeSnowflakeID(id), "is_transient", "false"),
+					importchecks.TestCheckResourceAttrInstanceState(id.FullyQualifiedName(), "name", id.Name()),
```
sfc-gh-asawicki, line 174: just a thought here: I wonder if we really make the right move with such a change. We could just handle the encode for 1-part identifiers the right way and then from the resource logic it would always be the same: encode - decode. Now we have some resources directly with FQN and some with encoding/decoding logic whereas it would be sufficient to hide this logic inside encode/decode. Let's chat on Monday.

pkg/sdk/databases.go
```diff
@@ -90,7 +90,14 @@ type databaseRow struct {
 func (row databaseRow) convert() *Database {
 	database := &Database{
 		CreatedOn: row.CreatedOn,
-		Name:      row.Name,
+	}
+	if row.Name != "" {
```
sfc-gh-asawicki, line 94: I am not sure if I like this logic. Previously, we would always have the Name, now, we can end up with an empty identifier which is a big change. Maybe we should address SNOW-1561641 first? Let's chat on Monday.

pkg/sdk/testint/client_unsafe_extensions_integration_test.go
```diff
@@ -13,13 +13,13 @@ func TestInt_Client_UnsafeQuery(t *testing.T) {
...
 	t.Run("test show databases", func(t *testing.T) {
-		sql := fmt.Sprintf("SHOW DATABASES LIKE '%%%s%%'", testDb(t).Name)
+		sql := fmt.Sprintf("SHOW DATABASES LIKE '%%%s%%'", testDb(t).Name.Name())
 		results, err := client.QueryUnsafe(ctx, sql)
 		require.NoError(t, err)
 
 		assert.Len(t, results, 1)
 		row := results[0]
-		assert.Equal(t, testDb(t).Name, *row["name"])
+		assert.Equal(t, testDb(t).Name.Name(), *row["name"])
```
sfc-gh-asawicki, line 22: Maybe instead of having `Name.Name()` which is ugly we should have ID().Name() ? (in all other places too)

sfc-gh-asawicki (changes requested):

pkg/resources/account_role.go
sfc-gh-jcieslak: I'll create a separate pr to point every SNOW-999049 todo to the right ticket nr.

pkg/resources/account_role_acceptance_test.go
sfc-gh-jcieslak: Yup, added planchecks

pkg/resources/custom_diffs.go
sfc-gh-jcieslak: Updated issue and linked this comment \+ added TODO comment

pkg/resources/schema_acceptance_test.go
sfc-gh-jcieslak: used `helpers.EncodeResourceIdentifier`

pkg/sdk/databases.go
sfc-gh-jcieslak: Reverted

pkg/sdk/testint/client_unsafe_extensions_integration_test.go
sfc-gh-jcieslak: Mostly reverted, but when left I used `ID()`

pkg/resources/custom_diffs.go
```diff
@@ -98,6 +98,22 @@ func ComputedIfAnyAttributeChanged(key string, changedAttributeKeys ...string) s
...
 
+// TODO(SNOW-1629468): Adjust the function to make it more flexible
+func ComputedIfAnyAttributeChangedWithSuppressDiff(key string, suppressDiffFunc schema.SchemaDiffSuppressFunc, changedAttributeKeys ...string) schema.CustomizeDiffFunc {
+	return customdiff.ComputedIf(key, func(ctx context.Context, diff *schema.ResourceDiff, meta interface{}) bool {
+		for _, changedKey := range changedAttributeKeys {
+			if diff.HasChange(changedKey) {
+				oldValue, newValue := diff.GetChange(changedKey)
+				if !suppressDiffFunc(key, oldValue.(string), newValue.(string), nil) {
+					log.Printf("[DEBUG] ComputedIfAnyAttributeChangedWithSuppressDiff: changed key: %s", changedKey)
+					return true
```
sfc-gh-jmichalak, line 109: Nit: is there a reason we return early here? Why don't we log all diffs, like in ComputedIfAnyAttributeChanged?

pkg/resources/diff_suppressions.go
```diff
@@ -185,3 +187,19 @@ func IgnoreValuesFromSetIfParamSet(key, param string, values []string) schema.Sc
 		return slices.Contains(values, old)
 	}
 }
+
+func suppressIdentifierQuoting(_, oldValue, newValue string, _ *schema.ResourceData) bool {
+	if oldValue == "" || newValue == "" {
+		return false
+	} else {
+		oldId, err := sdk.ParseIdentifierString(oldValue)
```
sfc-gh-jmichalak, line 195: Nit: this doesn't need to be in else block.

pkg/resources/database_acceptance_test.go
```diff
@@ -1112,8 +1124,13 @@ func TestAcc_Database_WithReplication(t *testing.T) {
...
 			{
-				ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
-				Config:                   databaseStateUpgraderWithReplicationNew(id, secondaryAccountIdentifier),
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"snowflake": {
+						VersionConstraint: "=0.94.1",
+						Source:            "Snowflake-Labs/snowflake",
+					},
+				},
+				Config: databaseStateUpgraderWithReplicationNew(id, secondaryAccountIdentifier),
```
sfc-gh-jmichalak, line 1133: Why are these tests pinned to a version?

pkg/resources/schema_acceptance_test.go
```diff
@@ -1044,3 +1045,93 @@ resource "snowflake_schema" "test" {
...
+				Config:                   schemaBasicConfig(id.DatabaseName(), id.Name()),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_schema.test", "id", id.FullyQualifiedName()),
+				),
+			},
+		},
+	})
+}
+
+func schemaBasicConfig(databaseName string, name string) string {
```
sfc-gh-jmichalak, line 1082: We already have `schemav094`.

pkg/resources/shared_database_acceptance_test.go
```diff
@@ -318,3 +320,129 @@ func createShareableDatabase(t *testing.T) sdk.ExternalObjectIdentifier {
...
+	require.Eventually(t, func() bool {
+		database, err := acc.TestClient().Database.Show(t, testId)
+		if err != nil {
+			return false
+		}
+		// Origin is returned as "<revoked>" in those cases, because it's not valid sdk.ExternalObjectIdentifier parser sets it as nil.
+		// Once it turns into valid sdk.ExternalObjectIdentifier, we're ready to proceed with the actual test.
+		return database.Origin != nil
+	}, time.Minute, time.Second*6)
+	acc.TestClient().Database.DropDatabaseFunc(t, testId)()
```
sfc-gh-jmichalak, line 407: We use this code in multiple places and it can be reused.

pkg/resources/user_parameters.go
```diff
@@ -165,7 +167,9 @@ func init() {
 }
 
 func userParametersProvider(ctx context.Context, d ResourceIdProvider, meta any) ([]*sdk.Parameter, error) {
-	return parametersProvider(ctx, d, meta.(*provider.Context), userParametersProviderFunc)
+	return parametersProvider(ctx, d, meta.(*provider.Context), userParametersProviderFunc, func(s string) (sdk.AccountObjectIdentifier, error) {
+		return helpers.DecodeSnowflakeID(s).(sdk.AccountObjectIdentifier), nil
```
sfc-gh-jmichalak, line 171: Why not ParseAccountObjectIdentifier?

pkg/helpers/resource_identifier_test.go
```diff
@@ -37,3 +37,48 @@ func Test_Encoding_And_Parsing_Of_ResourceIdentifier(t *testing.T) {
...
+func Test_Encoding_And_Parsing_Of_ResourceIdentifier_Identifier(t *testing.T) {
+	testCases := []struct {
+		Input                 []sdk.ObjectIdentifier
+		Expected              string
+		ExpectedAfterDecoding []string
+	}{
+		{Input: []sdk.ObjectIdentifier{sdk.NewAccountObjectIdentifier("a"), sdk.NewAccountObjectIdentifier("b")}, Expected: `"a"|"b"`},
+		{Input: []sdk.ObjectIdentifier{sdk.NewDatabaseObjectIdentifier("a", "b"), sdk.NewDatabaseObjectIdentifier("b", "c")}, Expected: `"a"."b"|"b"."c"`},
+		{Input: []sdk.ObjectIdentifier{sdk.NewSchemaObjectIdentifier("a", "b", "c"), sdk.NewSchemaObjectIdentifier("c", "b", "a")}, Expected: `"a"."b"."c"|"c"."b"."a"`},
+		{Input: []sdk.ObjectIdentifier{sdk.NewSchemaObjectIdentifierWithArguments("a", "b", "c", sdk.DataTypeFloat), sdk.NewSchemaObjectIdentifierWithArguments("c", "b", "a", sdk.DataTypeInt)}, Expected: `"a"."b"."c"(FLOAT)|"c"."b"."a"(INT)`},
```
sfc-gh-jmichalak, line 50: Please add a function without arguments

pkg/helpers/resource_identifier_test.go
```diff
@@ -37,3 +37,48 @@ func Test_Encoding_And_Parsing_Of_ResourceIdentifier(t *testing.T) {
...
+				assert.Equal(t, testCase.Expected, encodedIdentifier)
+			case []sdk.TableColumnIdentifier:
+				encodedIdentifier := EncodeResourceIdentifier(typedInput...)
+				assert.Equal(t, testCase.Expected, encodedIdentifier)
+			case []sdk.ExternalObjectIdentifier:
+				encodedIdentifier := EncodeResourceIdentifier(typedInput...)
+				assert.Equal(t, testCase.Expected, encodedIdentifier)
+			case []sdk.AccountIdentifier:
+				encodedIdentifier := EncodeResourceIdentifier(typedInput...)
+				assert.Equal(t, testCase.Expected, encodedIdentifier)
```
sfc-gh-jmichalak, line 80: Please a default case here with a warning or something

sfc-gh-jmichalak (changes requested):

pkg/resources/custom_diffs.go
sfc-gh-jcieslak: I mean, I just returned early, because other changes don't matter. If one thing changed, we want to recompute anyway, so there's no point of checking other keys. I guess this should be sorted out in the upcoming change where we want to change this function (or/and merge with the other one).

pkg/resources/diff_suppressions.go
sfc-gh-jcieslak: Removed

pkg/resources/database_acceptance_test.go
sfc-gh-jcieslak: Right now, no. Changed to the latest version.

pkg/resources/schema_acceptance_test.go
sfc-gh-jcieslak: I would leave it, ultimately we want to use the new approach and this name makes more sense. Let's leave both.

pkg/resources/shared_database_acceptance_test.go
sfc-gh-jcieslak: Let's leave it for SNOW-1562172

pkg/resources/user_parameters.go
sfc-gh-jcieslak: True, changed

pkg/helpers/resource_identifier_test.go
sfc-gh-jcieslak: Added

pkg/helpers/resource_identifier_test.go
sfc-gh-jcieslak: Good call, added

sfc-gh-jmichalak (approved):

Issue #2970, opened 2024-07-30
[Bug]: Failed apply to a snowflake_user password will still update the state of the user's password


jsnb-devoted:
### Terraform CLI Version

1.8.1

### Terraform Provider Version

0.92.0

### Terraform Configuration

```terraform
resource "snowflake_user" "service_account" {
  for_each          = local.service_accounts
  name              = each.key
  login_name        = each.value.login_name
  default_role      = each.value.role
  default_namespace = each.value.namespace
  default_warehouse = each.value.warehouse
  comment           = each.value.comment
  password          = try(random_password.snowflake_service_account_passwords[each.key].result, try(jsondecode(data.aws_secretsmanager_secret_version.service_account_creds[each.key].secret_string)["values"]["password"], null))
  rsa_public_key = try(
        trimspace(tls_private_key.snowflake_service_account_key[each.key].public_key_pem), null
  )
  must_change_password = false
}
```

### Category

category:resource

### Object type(s)

resource:user

### Expected Behavior

I had a plan that was meant to update a password and add a public key:
```
  # module.account.snowflake_user.service_account["MY_USER"] will be updated in-place
~ resource "snowflake_user" "service_account" {
        id                      = "MY_USER"
        name                    = (sensitive value)
      ~ password                = (sensitive value)
      + rsa_public_key          = (known after apply)
        # (13 unchanged attributes hidden)
    }
```
That failed because the public key had the header and footer:
```
╷
│ Error: 003065 (42601): SQL execution error:
│ New public key rejected by current policy. Reason: 'Invalid public key'
│ 
```
I checked the query history log and found the failed `ALTER USER` statement as expected however the state must have indicated that the password _had_ been updated when in fact it was still the old password:
```
  # module.account.snowflake_user.service_account["MY_USER"] will be updated in-place
~ resource "snowflake_user" "service_account" {
        id                      = "MY_USER"
        name                    = (sensitive value)
      ~ rsa_public_key          = <<-EOT
            -----BEGIN PUBLIC KEY-----
            ...
            -----END PUBLIC KEY-----
        EOT
        # (14 unchanged attributes hidden)
    }
```

### Actual Behavior

If an apply for the `snowflake_user` resource then I would expect TF to not update the state such that it produces the same diff on the subsequent plan. 

### Steps to Reproduce

[...]

Issue #2942, opened 2024-07-17
[Bug]: snowflake_user: default_role attribute breaks the apply


denzhel:
### Terraform CLI Version

1.5.5

### Terraform Provider Version

0.92.0

### Terraform Configuration

```terraform
resource "snowflake_user" "service_account" {
  depends_on = [snowflake_role.service_account_role, random_password.service_account]
  # parameters
  name                 = "SA-TEST"
  login_name           = "SA-TEST"
  password             = random_password.service_account[each.key].result
  default_role         = "SA-TEST"
  must_change_password = true
}
```

### Category

category:resource

### Object type(s)

resource:user

### Expected Behavior

Users to be created with a default_role that contains a hyphen as we created up until the upgrade from `0.56.1` to `0.92.0`

### Actual Behavior

```
╷
│ Error: 001003 (42000): SQL compilation error:
│ syntax error line 1 at position 179 unexpected '-'.
│ syntax error line 1 at position 179 unexpected '-'.
│ syntax error line 1 at position 184 unexpected '<EOF>'.
│ 
│   with module.snowflake.snowflake_user.service_account["SA-TEST"],
│   on modules/snowflake/main.tf line 376, in resource "snowflake_user" "service_account":
│  376: resource "snowflake_user" "service_account" {
│ 
╵
```
Query from Snowflake's UI:
```
CREATE USER "SA-TEST" PASSWORD = '☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺☺' LOGIN_NAME = 'SA-TEST' DISPLAY_NAME = 'SA-TEST' DEFAULT_ROLE = SA-TEST
```

### Steps to Reproduce

1. Create a new snowflake_user resource and add a default_role attribute that contains a hyphen, e.g `"SA-TEST"`
2. Apply
3. See how it fails
4. Remove the default_role field and the apply will be successful 

### How much impact is this issue causing?

High

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
Hey @denzhel . Thanks for reaching out to us.

[...]

Issue #2938, opened 2024-07-15
snowflake_user_parameter resource is missing


simonepm:
### Use Cases or Problem Statement

Hello,

we have snowflake_account_parameter and snowflake_object_parameter.

I see we do not have a way to set parameters at user level e.g. CLIENT_SESSION_KEEP_ALIVE true for Kafka connector users.

I was not able to find it in the roadmap as well, feel free to close this issue if is going to be implemented.

Many thanks!

### Category

category:resource

### Object type(s)

resource:account_parameter, resource:object_parameter, resource:user

### How much impact is this issue causing?

Low

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
Hi @simonepm 👋 We're adding missing session parameters to our in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2936. We're currently redesigning user resource, so this should be implemented in the next release. cc @sfc-gh-asawicki

simonepm:
Unfortunately session_parameter does not cover this use case in particular, but glad to hear that user resource is getting a redesign, thank you!

sfc-gh-asawicki:
Hey @simonepm. I am currently redesigning the user resource. It will support ~60 user parameters (including the requested `CLIENT_SESSION_KEEP_ALIVE`). We will publish the full list in our [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#migration-guide) when it's ready.

sfc-gh-asawicki:
Done in #2968 (not released yet).

sfc-gh-asawicki:
Hey @simonepm.
We have just released v0.95.0 of the provider. It contains a reworked `snowflake_user` resource. Please consult the [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v094x--v0950).

sfc-gh-asawicki:
Closing the issue due to inactivity. Please create a new one if the issue persists in the newest version of the provider.

Issue #2902, opened 2024-07-01
[Bug]: Error when using data source snowflake_users


calleo:
### Terraform CLI Version

1.6.6

### Terraform Provider Version

0.92.0

### Terraform Configuration

```terraform
data "snowflake_users" "all_users" {
  pattern = "%MATSMART%"
}
```

### Category

category:data_source

### Object type(s)

data_source:users

### Expected Behavior

I expect the data source to return a list of users, since the there are users matching the pattern.

### Actual Behavior

`null` is returned.

### Steps to Reproduce

Snowflake version: 8.23.5

1. Create a user with the name "FOO.BAR@MATSMART.SE"
2. Use the resource as specified above, it should return the user you just created

### How much impact is this issue causing?

Medium

### Logs

https://gist.github.com/calleo/9b37c61b37eb138c16518bbc19c50d86

### Additional Information

Looks like a problem with parsing the output of `SHOW USERS`. I have tried executing the exact same SQL statement, with the same user and role, and can see that it actually returns users.

My initial thought was that the `disabled` column contained `null` values, but I can also see `true`/`false`.

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
Hey @calleo . Thanks for reaching out to us.

The user datasource will be redesigned as part of https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#preparing-essential-ga-objects-for-the-provider-v1. I will link this issue to https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/v1-preparations/ESSENTIAL_GA_OBJECTS.MD. As a workaround, you can use https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/unsafe_execute with `query` parameter.

sebasziegler:
I am having the same issue at the moment, this used to work up to one month ago from the date, I guess there were some changes on the API in Snowflake side

sfc-gh-asawicki:
Closing the issue due to inactivity. Please create a new one if the issue persists in the newest version of the provider.

PR #3013, opened 2024-08-23
fix: Fix known user resource issues

sfc-gh-asawicki:
Fix known user resource-connected issues:
- Change the sensitiveness of name and login_name (References: #2662 #2668)
- Handle "null" properly for the nullable bool text attributes in user (References: #2817)
- Fix diff suppression for default_x in user resource (References: #2836)
- Update the migration guide (References #2938 #2942)
- Fix incorrect state after failed to alter (References #2970)
- Confirm the problem with the computed disabled attribute (References #1572)
- Confirm that the problem with the null-out password was already solved (References #1535)
- Add TODO to handle days to expiry in user (References #1155)

The next 2 PRs will contain:
- adjusting user resource to our rework conventions (also adding additional fields and handling #1155 and #1572)
- adjusting user datasource (will handle #2902)

User rework will not include handling new types of users (service, legacy service); this will be done a bit later.


MIGRATION_GUIDE.md
```diff
@@ -142,6 +142,26 @@ The following set of [parameters](https://docs.snowflake.com/en/sql-reference/pa
...
+
+According to https://docs.snowflake.com/en/sql-reference/functions/all_user_names#usage-notes, `NAME`s are not considered sensitive data and `LOGIN_NAME`s are. Previous versions of the provider had this the other way around. In this version, `name` attribute was unmarked as sensitive, whereas `login_name` was marked as sensitive. This may break your configuration if you were using `login_name`s before e.g. in a `for_each` loop.
+
+Connected issues: [#2662](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2662), [#2668](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2668).
+
+### *(bugfix)* Correctly handle `default_warehouse`, `default_namespace`, and `default_role`
+
+During the [identifiers rework](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#identifiers-rework), we generalized how we compute the differences correctly for the identifier fields (TODO link to doc?). Proper suppressor was applied to `default_warehouse`, `default_namespace`, and `default_role`. Also, all these three attributes were corrected (e.g. handling spaces/hyphens in names).
```
sfc-gh-jcieslak, line 155: doc is merged so you can reference it here

pkg/resources/user.go
```diff
@@ -20,14 +20,14 @@ var userSchema = map[string]*schema.Schema{
 	"name": {
 		Type:        schema.TypeString,
 		Required:    true,
-		Sensitive:   true,
+		Sensitive:   false,
```
sfc-gh-jcieslak, line 23: nit: we could just remove it

pkg/resources/user_acceptance_test.go
```diff
@@ -572,3 +577,219 @@ func TestAcc_User_AllParameters(t *testing.T) {
 		},
 	})
 }
+
+func TestAcc_User_issue2836(t *testing.T) {
+	userId := acc.TestClient().Ids.RandomAccountObjectIdentifier()
+	defaultRole := "SOME ROLE WITH SPACE case sensitive"
```
sfc-gh-jcieslak, line 583: Isn't it better to create random identifier with this as prefix? just to be sure this test won't fail once and then start failing with "this role already exists".

pkg/sdk/users.go
```diff
@@ -144,6 +146,18 @@ func (row userDBRow) convert() *User {
 	return user
 }
 
+func handleNullableBoolString(nullableBoolString sql.NullString, field *bool) {
```
sfc-gh-jcieslak, line 149: Maybe move to some kind of helper file in sdk package

MIGRATION_GUIDE.md
```diff
@@ -142,6 +142,26 @@ The following set of [parameters](https://docs.snowflake.com/en/sql-reference/pa
...
+
+Connected issues: [#2836](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2836), [#2942](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2942)
+
+### *(bugfix)* Correctly handle failed update
+
+Not every attribute can be updated in the state during read (like `password` in the `snowflake_user` resource). In situations where update fails, we may end up with an incorrect state (read more in https://github.com/hashicorp/terraform-plugin-sdk/issues/476). There is a deprecated option to fallback to the previous state in such cases which fixed this kind of situations for `snowflake_user` resource.
```
sfc-gh-jmichalak, line 161: I would be more explicit here and say: We use a deprecated method from the plugin SDK, and now, for partially failed updates, we preserve all of the resource's previous state.

MIGRATION_GUIDE.md
sfc-gh-asawicki: Thanks! Referenced.

pkg/resources/user.go
sfc-gh-asawicki: True. Removed.

pkg/resources/user_acceptance_test.go
sfc-gh-asawicki: No role is created in this test because default_x is not validated with CREATE/ALTER USER (check the docs here: https://docs.snowflake.com/en/sql-reference/sql/create-user#optional-object-properties-objectproperties, e.g. `In addition, the CREATE USER operation does not verify that the role exists.`).

MIGRATION_GUIDE.md
sfc-gh-asawicki: Good suggestion, adjusted.

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (dismissed):

sfc-gh-jcieslak (approved):

sfc-gh-jcieslak (approved):

Issue #3007, opened 2024-08-20
[Bug]: Table Column Updates are not recognized by Terraform Provider


JESCHO99:
### Terraform CLI Version

1.5.7

### Terraform Provider Version

0.94.1

### Terraform Configuration

```terraform
terraform {
  required_version = "~> 1.5.0"
  required_providers {
    snowflake = {
      source  = "Snowflake-Labs/snowflake"
      version = "~> 0.94.0"
    }
  }
}

provider "snowflake" {
  role = "SYSADMIN"
}

resource "snowflake_database" "this" {
  name = "TEST_DATABASE"
}

resource "snowflake_schema" "this" {
  database                    = snowflake_database.this.name
  name                        = "TEST_SCHEMA"
  is_transient                = "false"
  with_managed_access         = "false"
  data_retention_time_in_days = 1
}

resource "snowflake_table" "table_1" {
  database                    = snowflake_database.this.name
  schema                      = snowflake_schema.this.name
  name                        = "TEST_TABLE_V1"
  data_retention_time_in_days = 1
  change_tracking             = false

  column {
    name = "ID"
    type = "NUMBER"
  }

  column {
    name = "NAME"
    type = "VARCHAR(3)"
  }
}

resource "snowflake_table" "table_2" {
  database                    = snowflake_database.this.name
  schema                      = snowflake_schema.this.name
  name                        = "TEST_TABLE_V2"
  data_retention_time_in_days = 1
  change_tracking             = false

  column {
    name = "ID"
    type = "NUMBER(11,2)"
  }

  column {
    name = "NAME"
    type = "VARCHAR(255)"
  }
}
```

### Category

category:resource

### Object type(s)

resource:table

### Expected Behavior

The code provided above could be applied to a Snowflake Account without any problem. If you change the data type of a column for example change the "type" of the column "NAME" in the snowflake table "TEST_TABLE_V2" to "VARCHAR(3)" the next terraform plan should recognize this change and provide a plan how this change could be applied to the backend.

```json
resource "snowflake_table" "table_2" {
  database                    = snowflake_database.this.name
  schema                      = snowflake_schema.this.name
  name                        = "TEST_TABLE_V2"
  data_retention_time_in_days = 1
  change_tracking             = false

  column {
    name = "ID"
    type = "NUMBER(11,2)"
  }

  column {
    name = "NAME"
    type = "VARCHAR(3)"
  }
}
```

### Actual Behavior

[...]

Issue #2972, opened 2024-08-01
[Bug]: Not possible to rename stage - Running into a deadlock


JESCHO99:
### Terraform CLI Version

1.5.7

### Terraform Provider Version

0.92.0

### Terraform Configuration

```terraform
provider "snowflake" {
  role  = "SYSADMIN"
}

terraform {
  required_version = "~> 1.5.0"
  required_providers {
    snowflake = {
      source  = "Snowflake-Labs/snowflake"
      version = "~> 0.92.0"
    }
  }
}

resource "snowflake_database" "this" {
  name = "LOCAL_TEST_DB"
}

resource "snowflake_schema" "this" {
  name     = "TESTING"
  database = snowflake_database.this.name
}

resource "snowflake_table" "table_v1" {
  database                    = snowflake_database.this.name
  schema                      = snowflake_schema.this.name
  name                        = "TEST_V1"
  change_tracking             = false

  column {
    name     = "ID"
    type     = "INTEGER"
  }

  column {
    name = "NAME"
    type = "VARCHAR(255)"
  }
}

resource "snowflake_table" "table_v2" {
  database                    = snowflake_database.this.name
  schema                      = snowflake_schema.this.name
  name                        = "TEST_V2"
  change_tracking             = false

  column {
    name     = "ID"
    type     = "INTEGER"
  }

  column {
    name = "NAME"
    type = "VARCHAR(255)"
  }
}

resource "snowflake_stage" "stage_v1" {
  name        = "TEST_V1"
  database    = snowflake_database.this.name
  schema      = snowflake_schema.this.name
}

resource "snowflake_stage" "stage_v2" {
  name        = "TEST_V2"
  database    = snowflake_database.this.name
  schema      = snowflake_schema.this.name
}
```

### Category

category:resource

### Object type(s)

resource:stage

### Expected Behavior

When the provided terraform code is applied in Snowflake the Database, Schema, two tables and two named stages can be deployed successful. Everything is working as expected so far. Due to manual changes on Snowflake someone executed a rename command on one of the named stages: `alter stage LOCAL_TEST_DB.TESTING.TEST_V1 rename to POST_TEST_V1;`

When executing a terraform plan now in Terraform the expected behaviour would be that terraform recognizes that the defined stage no longer exists and will show a create stage in the produced plan. When this plan is executed terraform will create a new stage having the "old" name. In Snowflake there will be the renamed "old" stage and the newly created stage with the "old" name.

[...]

PR #3020, opened 2024-08-28
fix: Fix issues 2972 and 3007

sfc-gh-asawicki:
Address two issues: #2972 and #3007:
- Correctly handle renamed/deleted stage
- Handle data type diff suppression better for text and number in the table resource

References: #2972 #3007

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -90,6 +90,30 @@ Some of the resources are excluded from this change:
 #### *(breaking change)* removed `qualified_name` from `snowflake_masking_policy`, `snowflake_network_rule`, `snowflake_password_policy` and `snowflake_table`
 Because of introducing a new `fully_qualified_name` field for all of the resources, `qualified_name` was removed from `snowflake_masking_policy`, `snowflake_network_rule`,  `snowflake_password_policy` and `snowflake_table`. Please adjust your configurations. State is automatically migrated.
 
+### snowflake_stage resource changes
+
+#### *(bugfix)* Correctly handle renamed/deleted stage
+
+Correctly handle the situation when stage was rename/deleted externally (earlier it resulted in a permanent loop). No action is required on the user's side.
+
+Connected issues: [#2972](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2972) 
+
+### snowflake_table resource changes
+
+#### *(bugfix)* Handle data type diff suppression better for text and number
+
+Data types are not entirely correctly handled inside the provider (read more e.g. in [#2735](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2735)). It will be still improved with the upcoming function, procedure, and table rework. Currently, diff suppression was fixed for text and number data types in the table resource with the following assumptions/limitations:
+- for numbers the default precision is 38 and the default scale is 0 (following the [docs](https://docs.snowflake.com/en/sql-reference/data-types-numeric#number))
+- for number types the following types are treated as synonyms: `NUMBER`, `DECIMAL`, `NUMERIC`, `INT`, `INTEGER`, `BIGINT`, `SMALLINT`, `TINYINT`, `BYTEINT`
+- for text the default length is 16777216 (following the [docs](https://docs.snowflake.com/en/sql-reference/data-types-text#varchar))
+- for text types the following types are treated as synonyms: `VARCHAR`, `CHAR`, `CHARACTER`, `STRING`, `TEXT`
+- whitespace and casing is ignored
+- if the type arguments cannot be parsed the defaults are used and therefore diff may be suppressed unexpectedly (please report such cases)
+
+No action is required on the user's side.
+
+Connected issues: [#3007](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/3007)
+
 ### snowflake_user resource changes
 
 #### *(breaking change)* user parameters added to snowflake_user resource
--- a/pkg/resources/helpers_test.go
+++ b/pkg/resources/helpers_test.go
@@ -3,6 +3,7 @@ package resources_test
 import (
 	"context"
 	"fmt"
+	"github.com/stretchr/testify/require"
 	"slices"
 	"testing"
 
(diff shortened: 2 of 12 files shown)


pkg/sdk/data_types.go
```diff
@@ -112,3 +118,59 @@ func IsStringType(_type string) bool {
 		strings.HasPrefix(t, "NVARCHAR") ||
 		strings.HasPrefix(t, "NCHAR")
 }
+
+// ParseNumberDataTypeRaw extracts precision and scale from the raw number data type input.
+// It returns defaults if it can't parse arguments, data type is different, or no arguments were provided.
+// TODO [SNOW-1348103 or SNOW-1348106]: visit with functions and procedures rework
+func ParseNumberDataTypeRaw(rawDataType string) (int, int) {
+	r := util.TrimAllPrefixes(strings.TrimSpace(strings.ToUpper(rawDataType)), DataTypeNumberSynonyms...)
```
sfc-gh-jcieslak, line 126: I don't get why this part is needed here `strings.TrimSpace(strings.ToUpper(rawDataType))` and that it works 🤔, because having string `NUMBER(30)` and trim prefix `NUMBER(30)` we should end up with `r` equal to an empty string or am I missing something.

sfc-gh-asawicki: We are trimming a prefix without parentheses, so e.g. `number(30)` becomes `(30)` and we process the remaining part to extract attribute values.
Trimming spaces was added to avoid stupid whitespace problems e.g. ` number ( 30 ) `.

sfc-gh-jcieslak (approved):

pkg/sdk/data_types.go
```diff
@@ -112,3 +118,59 @@ func IsStringType(_type string) bool {
...
+		case 2:
+			precision, err1 := strconv.Atoi(strings.TrimSpace(parts[0]))
+			scale, err2 := strconv.Atoi(strings.TrimSpace(parts[1]))
+			if err1 == nil && err2 == nil {
+				return precision, scale
+			} else {
+				logging.DebugLogger.Printf(`[DEBUG] Could not parse number precision "%s" or scale "%s", errs: %v, %v`, parts[0], parts[1], err1, err2)
+			}
+		default:
+			logging.DebugLogger.Printf("[DEBUG] Unexpected length of number arguments")
```
sfc-gh-jmichalak, line 147: Nit: I would add more description, at least a function name because this is too generic. In other lines as well.

pkg/resources/table_acceptance_test.go
```diff
@@ -2090,3 +2092,185 @@ resource "snowflake_table" "test_table" {
 }
 `, name, databaseName, schemaName)
 }
+
+func TestAcc_Table_issue3007_textColumn(t *testing.T) {
+	databaseName := acc.TestClient().Ids.Alpha()
+	schemaName := acc.TestClient().Ids.Alpha()
+	name := acc.TestClient().Ids.Alpha()
```
sfc-gh-jmichalak, line 2099: Use `RandomSchemaIdentifier`

pkg/resources/table_acceptance_test.go
```diff
@@ -2090,3 +2092,185 @@ resource "snowflake_table" "test_table" {
...
+	return fmt.Sprintf(`
+resource "snowflake_database" "test_database" {
+    name = "%[2]s"
+}
+
+resource "snowflake_schema" "test_schema" {
+    depends_on = [snowflake_database.test_database]
+    name = "%[3]s"
+    database = "%[2]s"
+}
```
sfc-gh-jmichalak, line 2256: Why not default schema?

pkg/resources/helpers_test.go
```diff
@@ -258,3 +259,147 @@ func TestListDiff(t *testing.T) {
...
+		{
+			name:     "different data type",
+			old:      string(sdk.DataTypeVARCHAR),
+			new:      string(sdk.DataTypeNumber),
+			expected: false,
+		},
+		{
+			name:     "same number data type without arguments",
+			old:      string(sdk.DataTypeNumber),
+			new:      string(sdk.DataTypeNumber),
```
sfc-gh-jmichalak, line 279: I think this case (old==new) is handled by Terraform and diff suppress isn't executed.

MIGRATION_GUIDE.md
```diff
@@ -90,6 +90,30 @@ Some of the resources are excluded from this change:
 #### *(breaking change)* removed `qualified_name` from `snowflake_masking_policy`, `snowflake_network_rule`, `snowflake_password_policy` and `snowflake_table`
 Because of introducing a new `fully_qualified_name` field for all of the resources, `qualified_name` was removed from `snowflake_masking_policy`, `snowflake_network_rule`,  `snowflake_password_policy` and `snowflake_table`. Please adjust your configurations. State is automatically migrated.
 
+### snowflake_stage resource changes
+
+#### *(bugfix)* Correctly handle renamed/deleted stage
+
+Correctly handle the situation when stage was rename/deleted externally (earlier it resulted in a permanent loop). No action is required on the user's side.
```
sfc-gh-jmichalak, line 97: Hmm, I'm not sure if a bugfix that changes nothing on the user side should belong to a migration guide.

sfc-gh-jmichalak (changes requested):

pkg/sdk/data_types.go
sfc-gh-asawicki: Added a function name temporarily.

FYI, we have a plan to rework logging inside the provider and the SDK as part of SNOW-1325306. The idea was, to have a logger that adds the code place on its own, so such descriptions won't be necessary and will need to be altered to reduce the noise.

pkg/resources/table_acceptance_test.go
sfc-gh-asawicki: Changed

pkg/resources/table_acceptance_test.go
sfc-gh-asawicki: It was a copy-pasted config; adjusted and used default schema and the db

pkg/resources/helpers_test.go
sfc-gh-asawicki: You are most probably right but it does not hurt us to handle this case properly too, right?

sfc-gh-jmichalak (approved):

sfc-gh-jcieslak (approved):

PR #3023, opened 2024-08-28
feat: View rework part 3

sfc-gh-jmichalak:
- add columns
- adjust query parsing
- adjust data source
- test columns in basic flow
- check `using` field in `masking_policy` - can there be more than 1 element? (yes there can be more)
- handle data metric function's schedule status

## Test Plan
* [x] acceptance tests (resource and data source)

## References
[CREATE VIEW](https://docs.snowflake.com/en/sql-reference/sql/create-view)


pkg/resources/view_acceptance_test.go
```diff
@@ -56,27 +59,52 @@ func TestAcc_View_basic(t *testing.T) {
 	t.Cleanup(tableCleanup)
 	statement := fmt.Sprintf("SELECT id, foo FROM %s", table.ID().FullyQualifiedName())
 	otherStatement := fmt.Sprintf("SELECT foo, id FROM %s", table.ID().FullyQualifiedName())
-	comment := "Terraform test resource'"
+	comment := random.Comment()
 
-	viewModel := model.View("test", id.DatabaseName(), id.Name(), id.SchemaName(), statement)
-	viewModelWithDependency := model.View("test", id.DatabaseName(), id.Name(), id.SchemaName(), statement)
+	// generators currently don't handle lists, so use the old way
```
sfc-gh-asawicki, line 64: nit: lists of objects, flat lists like lists of strings are handled

pkg/resources/view_acceptance_test.go
```diff
@@ -89,24 +117,32 @@ func TestAcc_View_basic(t *testing.T) {
...
-				Config:       accconfig.FromModel(t, viewModel),
-				ResourceName: "snowflake_view.test",
-				ImportState:  true,
-				ImportStateCheck: assert.AssertThatImport(t, assert.CheckImport(importchecks.TestCheckResourceAttrInstanceState(resourceId, "name", id.Name())),
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_View/basic"),
+				ConfigVariables: basicView(statement),
+				ResourceName:    "snowflake_view.test",
+				ImportState:     true,
+				ImportStateCheck: assert.AssertThatImport(t,
+					assert.CheckImport(importchecks.TestCheckResourceAttrInstanceState(resourceId, "name", id.Name())),
```
sfc-gh-asawicki, line 138: nit: this check is also below: `HasNameString(id.Name())`

pkg/resources/view_acceptance_test.go
```diff
@@ -89,24 +117,32 @@ func TestAcc_View_basic(t *testing.T) {
...
+					assert.CheckImport(importchecks.TestCheckResourceAttrInstanceState(resourceId, "column.#", "2")),
```
sfc-gh-asawicki, line 139: nit: you can add to ext a method `HasColumnLength(length int)` instead of repeating the old ugly way :)

pkg/resources/view.go
```diff
@@ -684,6 +746,53 @@ func handleDataMetricFunctions(ctx context.Context, client *sdk.Client, id sdk.S
 	})
 }
 
+func handleColumns(d *schema.ResourceData, columns []sdk.ViewDetails, policyRefs []sdk.PolicyReference) error {
```
sfc-gh-asawicki, line 749: so much logic here, could we add a unit test for it? I know that ResouceData is limited but for the test we could provide something else implementing d.Set() and check the interations

pkg/resources/view.go
```diff
@@ -452,48 +466,77 @@ func CreateView(orReplace bool) schema.CreateContextFunc {
...
+				}
+			}
 		}
 
 		return ReadView(false)(ctx, d, meta)
 	}
 }
 
-func extractPolicyWithColumns(v any, columnsKey string) (sdk.SchemaObjectIdentifier, []sdk.Column, error) {
+func extractColumns(v any) ([]sdk.ViewColumnRequest, error) {
```
sfc-gh-asawicki, line 504: can we add a unit test for it?

pkg/resources/view.go
```diff
@@ -502,6 +545,23 @@ func extractPolicyWithColumns(v any, columnsKey string) (sdk.SchemaObjectIdentif
 	return id, columns, nil
 }
 
+func extractPolicyWithColumnsList(v any, columnsKey string) (sdk.SchemaObjectIdentifier, []sdk.Column, error) {
```
sfc-gh-asawicki, line 548: can we add a unit test for it?

pkg/resources/view_acceptance_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"regexp"
 	"testing"
 
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers/random"
```
sfc-gh-asawicki, line 8: do we have a test for external column changes?
and enough tests for manipulation of columns?

sfc-gh-asawicki (changes requested):

pkg/resources/view_acceptance_test.go
sfc-gh-jcieslak: Changed comment

pkg/resources/view_acceptance_test.go
sfc-gh-jcieslak: True, removed this one.

pkg/resources/view_acceptance_test.go
sfc-gh-jcieslak: Added

pkg/resources/view_acceptance_test.go
sfc-gh-jcieslak: I'll add an acceptance test with internal and external changes to columns to ensure we know they behave correctly.

pkg/resources/view.go
sfc-gh-jcieslak: Don

pkg/resources/view.go
sfc-gh-jcieslak: Added

pkg/acceptance/bettertestspoc/assert/objectassert/view_snowflake_ext.go
```diff
@@ -28,3 +32,71 @@ func (v *ViewAssert) HasNonEmptyText() *ViewAssert {
...
+}
+
+func (v *ViewAssert) HasNoProjectionPolicyReferences() *ViewAssert {
+	return v.hasNoPolicyReference(sdk.PolicyKindProjectionPolicy)
+}
+
+func (v *ViewAssert) hasNoPolicyReference(kind sdk.PolicyKind) *ViewAssert {
+	v.AddAssertion(func(t *testing.T, o *sdk.View) error {
+		t.Helper()
+		refs, err := acc.TestClient().PolicyReferences.GetPolicyReferences(t, o.ID(), sdk.ObjectTypeView)
```
sfc-gh-asawicki, line 55: we should not add here acc.TestClient() because they can be invoked both from acc and integration tests levels, please for now parameterize invocation with a proper test client

pkg/resources/view_test.go
```diff
@@ -0,0 +1,355 @@
...
+			},
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("handle columns(%d): %v - %v", i, tc.InputColumns, tc.InputPolicyReferences), func(t *testing.T) {
+			valueSetter := newTestResourceValueSetter()
+			err := handleColumns(valueSetter, tc.InputColumns, tc.InputPolicyReferences)
+
+			if tc.Error != "" {
```
sfc-gh-asawicki, line 197: there is no test case with Error,please remove it (or add err case if needed)

sfc-gh-asawicki (changes requested):

pkg/resources/view_test.go
sfc-gh-jcieslak: Removed

pkg/acceptance/bettertestspoc/assert/objectassert/view_snowflake_ext.go
sfc-gh-jcieslak: Don

sfc-gh-asawicki (approved):

PR #3032, opened 2024-09-02
fix: Fix custom diffs for fields with diff supression

sfc-gh-asawicki:
Fix custom diffs for fields with diff impression:
- existing `ComputedIfAnyAttributeChanged` merged with `ComputedIfAnyAttributeChangedWithSuppressDiff`
- changed the way how the suppress diff is invoked from within `ComputedIfAnyAttributeChanged` (ResourceData and string values for old and new)
- fix the failing tests (behavior was fixed)
- add test to validate the behavior in warehouse


pkg/internal/provider/sdkv2enhancements/resource_data.go
```diff
@@ -0,0 +1,36 @@
...
+// CreateResourceDataFromResourceDiff allows to create schema.ResourceData out of schema.ResourceDiff.
+// Unfortunately, schema.resourceDiffer is an unexported interface, and functions like schema.SchemaDiffSuppressFunc do not use the interface but a concrete implementation.
+// One use case in which we needed to have schema.ResourceData from schema.ResourceDiff was to run schema.SchemaDiffSuppressFunc from inside schema.CustomizeDiffFunc.
+// This implementation uses:
+// - schema.InternalMap that exposes hidden schema.schemaMap (a wrapper over map[string]*schema.Schema)
+// - (m schemaMap) Data method allowing to create schema.ResourceData from terraform.InstanceState and terraform.InstanceDiff
+// - terraform.InstanceState and terraform.InstanceDiff are unexported in schema.ResourceDiff, so we get them using reflection
+func CreateResourceDataFromResourceDiff(resourceSchema schema.InternalMap, diff *schema.ResourceDiff) (*schema.ResourceData, bool) {
+	unexportedState := reflect.ValueOf(diff).Elem().FieldByName("state")
+	stateFromResourceDiff := reflect.NewAt(unexportedState.Type(), unsafe.Pointer(unexportedState.UnsafeAddr())).Elem().Interface()
```
sfc-gh-jcieslak, line 20: Possibly `reviewdog` may disappear with `unexportedState.Addr().UnsafePointer()`, but I'm not sure if that leads to the same result (I looks like it).

pkg/resources/custom_diffs.go
```diff
@@ -83,33 +84,38 @@ func ForceNewIfChangeToEmptyString(key string) schema.CustomizeDiffFunc {
 	})
 }
 
-func ComputedIfAnyAttributeChanged(key string, changedAttributeKeys ...string) schema.CustomizeDiffFunc {
+// ComputedIfAnyAttributeChanged marks the given fields as computed if any of the listed fields changes.
+// It takes field-level diffSuppress into consideration based on the schema passed.
+// If the field is not found in the given schema, it continues without error.
+func ComputedIfAnyAttributeChanged(resourceSchema map[string]*schema.Schema, key string, changedAttributeKeys ...string) schema.CustomizeDiffFunc {
```
sfc-gh-jcieslak, line 90: Hmm I think it would be worth to mention that right now the function can take only top-level keys (previously I think it could take nested keys, but now it's not possible with `resourceSchema`).

pkg/resources/custom_diffs.go
```diff
@@ -83,33 +84,38 @@ func ForceNewIfChangeToEmptyString(key string) schema.CustomizeDiffFunc {
...
+						resourceData, resourceDataOk := sdkv2enhancements.CreateResourceDataFromResourceDiff(resourceSchema, diff)
+						if !resourceDataOk {
+							log.Printf("[DEBUG] ComputedIfAnyAttributeChanged: did not create resource data correctly, skipping\n")
+							continue
+						}
+						if !diffSuppressFunc(key, fmt.Sprintf("%v", oldValue), fmt.Sprintf("%v", newValue), resourceData) {
+							log.Printf("[DEBUG] ComputedIfAnyAttributeChanged: key %s was changed and the diff is not suppressed", changedKey)
+							result = true
+						} else {
+							log.Printf("[DEBUG] ComputedIfAnyAttributeChanged: key %s was changed but the diff is suppresed", changedKey)
```
sfc-gh-jcieslak, line 109: *suppressed

sfc-gh-jcieslak (approved):

pkg/internal/provider/sdkv2enhancements/resource_data.go
sfc-gh-asawicki: Done in a separate PR

pkg/resources/custom_diffs.go
sfc-gh-asawicki: Added in the next PR.

pkg/resources/custom_diffs.go
sfc-gh-asawicki: Fixed in a separate

PR #3034, opened 2024-09-03
chore: User improvements

sfc-gh-jcieslak:
## Changes
- Added basic DiffSuppression/Validations for user parameters that are string enums.
- Added basic flow for display_name | login_name fields (should I create a ticket and include its number for the 2nd step? incorrect values after rename).
- Added integration test that shows which fields are visible when querying a given user with different query commands.
- Added HasMfa to describe and added new asserts in users acceptance test to show it's there and it's getting mapped (the only way to get it `true` is to set MFA manually so I can only assert it's `false`).

## Doing right now
- Testing manually default_namespace / default_role


pkg/resources/user_acceptance_test.go
```diff
@@ -1102,3 +1102,108 @@ func TestAcc_User_handleChangesToDefaultSecondaryRoles(t *testing.T) {
...
+	userModelWithBoth := model.User("w", newId.Name()).WithLoginName("login_name").WithDisplayName("display_name")
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		PreCheck:     func() { acc.TestAccPreCheck(t) },
+		CheckDestroy: acc.CheckDestroy(t, resources.User),
+		Steps: []resource.TestStep{
```
sfc-gh-asawicki, line 1158: please add steps with external changes

sfc-gh-asawicki (approved):

PR #3035, opened 2024-09-03
chore: Improve user test and add manual test for user default database and role

sfc-gh-jcieslak:
## Changes
- Address https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3034#discussion_r1742132935
- Add a manual test that shows the behavior of the default namespace and default role for a user when logging in with different conditions.

### DIFF
--- a/pkg/manual_tests/user_default_database_and_role/test.md
+++ b/pkg/manual_tests/user_default_database_and_role/test.md
@@ -7,6 +7,7 @@ The cases are covering the situations where:
 - Database and role are granted to the user, but the casing is not matched.
 - Database and role are granted to the user and the casing is matching.
 - Only role is granted to the user and the casing is matching.
+- Database and role are created with lowercase, granted to the user.
 
 ## Setup
 
@@ -26,9 +27,13 @@ The only thing that won't change between the tests is role and database, so we c
 ```sql
 CREATE ROLE TEST_ROLE;
 CREATE DATABASE TEST_DATABASE;
+CREATE ROLE "test_role";
+CREATE DATABASE "test_database";
 ```
 
-> Note: We have to use TYPE = LEGACY_SERVICE to be able to use 
+> Note: We have to use `TYPE = LEGACY_SERVICE` to be able to quickly log in using `login_name` + `password`.
+
+> Note: It doesn't matter if you use single or double quotes for user properties (they result in the same behavior).
 
 ### 1. Either database or role is not present.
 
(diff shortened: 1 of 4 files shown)


pkg/manual_tests/user_default_database_and_role/test.md
```diff
@@ -0,0 +1,137 @@
...
+### 3. Database and role are granted to the user, but the casing is not matched.
+
+1. Replace the user with existing default database and role, but with lowercase names.
+
+```sql
+CREATE OR REPLACE USER TEST_USER
+    TYPE = LEGACY_SERVICE
+    LOGIN_NAME = 'login_name'
+    PASSWORD = 'password'
+    DEFAULT_NAMESPACE = 'test_database'
```
sfc-gh-asawicki, line 72: Is our SDK using single or double quotes here? (I am almost sure that the double ones - maybe this is not a difference in behavior but wanted to be sure)

pkg/manual_tests/user_default_database_and_role/test.md
```diff
@@ -0,0 +1,137 @@
...
+2. Grant the role to the user and grant usage on the database to the role.
+
+```sql
+GRANT ROLE TEST_ROLE TO USER TEST_USER;
+GRANT USAGE ON DATABASE TEST_DATABASE TO ROLE TEST_ROLE;
+```
+
+3. Log into the user (logs in successfully and the database and role are selected in the context).
+
+### 5. Only role is granted to the user and the casing is matching.
```
sfc-gh-asawicki, line 107: lets do three more where:
1. role matches and is granted
2. database access is given to the role
3. database in create user is lowercased (versus the uppercased created one)

second and third:
the same as the previous one but let's create lowercased db and:
- use lowercase in create user (second test)
- use lowercase in create user (third test)

sfc-gh-asawicki (changes requested):

pkg/manual_tests/user_default_database_and_role/test.md
sfc-gh-jcieslak: I tested it in a few ways and it didn't make any difference. I'll leave it as is and add note at the top that the quoting doesn't matter in this case.

pkg/manual_tests/user_default_database_and_role/test.md
sfc-gh-jcieslak: I guess the first one proposed is already test nr 3. I'll add other ones with lowercase role and database.

sfc-gh-jcieslak: Also added summary that summarizes the behavior of those two parameters, so it's clearer what the outcomes are.

sfc-gh-asawicki (approved):

PR #3036, opened 2024-09-03
feat: Conclude user rework

sfc-gh-asawicki:
- Add missing descriptions to public keys resource
- Add missing descriptions to the migration guide
- Add missing references to issues
- Add diff suppressors and tests
- Fix review comments from #3032 
- Add missing descriptions for special fields
- Add missing notes to the user resource


pkg/resources/user.go
```diff
@@ -158,7 +158,7 @@ var userSchema = map[string]*schema.Schema{
 		Type:             schema.TypeString,
 		Optional:         true,
 		ValidateDiagFunc: validateBooleanString,
-		Description:      booleanStringFieldDescription("Allows enabling or disabling [multi-factor authentication](https://docs.snowflake.com/en/user-guide/security-mfa)."),
+		Description:      booleanStringFieldDescription("Allows enabling or disabling [multi-factor authentication](https://docs.snowflake.com/en/user-guide/security-mfa). It cannot be read from Snowflake, so it has to be marked as tainted manually after external changes."),
```
sfc-gh-jcieslak, line 161: We also have `externalChangesNotDetectedFieldDescription` (also in other fields), but because it's already wrapped in `booleanStringFieldDescription` I guess we can leave it as is. We can think later about other solution (e.g. have a function that would receive description as first arg and modifiers afterwards such as `booleanStringFieldDescription` or `externalChangesNotDetectedFieldDescription`).

sfc-gh-jcieslak (approved):

pkg/resources/user.go
sfc-gh-asawicki: TBH, we can currently just wrap one in the other (does not look bad). I also used externalChangesNotDetectedFieldDescription for other 3 fields like suggested

sfc-gh-jcieslak (approved):

Issue #3038, opened 2024-09-03
[Bug]: 2024_07 BCR breaks snowflake_user.foo.default_secondary_roles when unset


jrobison-sb:
### Terraform CLI Version

1.9.1

### Terraform Provider Version

0.92.0

### Terraform Configuration

```terraform
resource "snowflake_user" "dev3229" {
  name         = "DEV3229"
  login_name   = "DEV3229"
  comment      = "A user of snowflake."
  disabled     = false
  must_change_password = false

  default_secondary_roles = null
}
```

### Category

category:resource

### Object type(s)

`snowflake_user`

### Expected Behavior

`default_secondary_roles` should be null, and prior to the [2024_07 behavior change](https://docs.snowflake.com/en/release-notes/bcr-bundles/2024_07/bcr-1692), this worked as expected. Additionally, after applying the above infrastructure, a subsequent `terraform plan` should plan cleanly with no diffs reported, which also worked as expected prior to the behavior change.

### Actual Behavior

It's no longer possible to set this at `null`, or set it to an empty list, without plan diffs.

### Steps to Reproduce

0. Apply the [2024_07 behavior change](https://docs.snowflake.com/en/release-notes/bcr-bundles/2024_07/bcr-1692).
1. Apply the above HCL which uses `default_secondary_roles = null` (or which doesn't specify `default_secondary_roles` at all). This will apply cleanly.
2. Run a `terraform plan` and you'll see diffs like this:

```
  # module.sandbox3_application.snowflake_user.dev3229 will be updated in-place
  ~ resource "snowflake_user" "dev3229" {
      ~ default_secondary_roles = [
          - "ALL",
        ]
        id                      = "DEV3229"
        name                    = (sensitive value)
        # (12 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

3. Hmm, that's weird, try applying again maybe?
4. Now it gets even weirder because now the diff looks like this on every subsequent plan:

```
  # module.sandbox3_application.snowflake_user.dev3229 will be updated in-place
  ~ resource "snowflake_user" "dev3229" {
      ~ default_secondary_roles = [
          - jsonencode([]),
        ]
        id                      = "DEV3229"
        name                    = (sensitive value)
        # (12 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

5. Try setting it to an empty list instead of `null`. Apply again.
[...]

PR #3040, opened 2024-09-04
fix: Fix default secondary roles after bcr 2024 07

sfc-gh-asawicki:
- fix warehouse resume
- change default secondary roles handling
- address BCR 2024_07 for default secondary roles

References: #3038


pkg/sdk/testint/system_functions_integration_test.go
```diff
@@ -173,3 +173,19 @@ func TestInt_PipeForceResume(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, sdk.RunningPipeExecutionState, pipeExecutionState)
 }
+
+// TODO [SNOW-1650249]: add positive tests for bundle enablement (add SYSTEM$SHOW_ACTIVE_BEHAVIOR_CHANGE_BUNDLES() and use it to always pick a bundle that can be enabled/disabled)
```
sfc-gh-jcieslak, line 177: Should we add bcr checks in our checks before integration / acceptance tests so we will always start with the expected set of features enabled / disabled ? Not in this change though.

sfc-gh-jcieslak (approved):

pkg/sdk/testint/system_functions_integration_test.go
sfc-gh-asawicki: Yup, it has some challenges but added to the issue.

Issue #1990, opened 2023-08-01
New resource monitor SDK unexpected behavior


dlouseiro:
**Provider Version**

0.69.0

**Terraform Version**

1.5.4

**Describe the bug**

After upgrading the provider to version `0.69.0` (where the new resource monitor SDK starts being used), my `terraform plan` fails with `Error: object does not exist or not authorized` when the object exists.

**Expected behavior**

For the `terraform plan` not to fail.

**Code samples and commands**

Resource monitor definition:
```
resource "snowflake_resource_monitor" "monthly" {
  name            = "MONTHLY"
  credit_quota    = 12000
  notify_triggers = [100]
  set_for_account = true
}
```

dlouseiro:
Update on this topic: Removing the resource from the state and re-importing solves the issue!

sfc-gh-jcieslak:
Hey 👋
I'm currently working on the resource monitor as part of [preparing GA objects for V1](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#preparing-essential-ga-objects-for-the-provider-v1), so a lot may change, and the improvements are coming. I'll let you know when the updated resource is available. I may know where the issue lies; it shouldn't happen in the new resource monitor.

sfc-gh-jcieslak:
**Update**: I just tested it, and this issue won't occur in the upcoming version of the provider.

sfc-gh-jcieslak:
Hey 👋
The new and refactored resource monitor was released yesterday in version 0.96.0 of the provider. Please migrate with [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v0950--v0960) and let me know if the issue has been resolved so we could close the ticket. Thanks in advance 🙏.

Issue #1832, opened 2023-05-25
Altering Resource Monitors doesnt work as expected


curtisq:
**Provider Version**

The provider version you are using.
0.55.1

**Terraform Version**

The version of Terraform you were using when the bug was encountered.

1.0.11

**Describe the bug**

Changing only the triggers of a resource monitor throws a syntax error

For example take the following monitor:
```
resource "snowflake_resource_monitor" "some_monitor" {
  name         = "some_monitor"
  warehouses   = [snowflake_warehouse.some_warehouse.name]
  credit_quota = 10

  frequency       = "WEEKLY"
  start_timestamp = "2023-02-02T00:00:00Z"

  notify_triggers            = [90]
  suspend_triggers           = [100]
  suspend_immediate_triggers = [105]

  notify_users = [snowflake_user.some_user.name]
}
```

If I were to say change a trigger line to `notify_triggers            = [95]` and change nothing else then the provider produces the following SQL:
```
ALTER RESOURCE MONITOR "some_monitor" 
SET  
TRIGGERS ON 100 PERCENT DO SUSPEND 
    ON 105 PERCENT DO SUSPEND_IMMEDIATE 
    ON 95 PERCENT DO NOTIFY
```
This fails with 
```
syntax error line 3 at position 24 unexpected 'DO'.
syntax error line 4 at position 4 unexpected 'ON'.
...
```

**Expected behavior**

Based on the [docs](https://docs.snowflake.com/en/sql-reference/sql/alter-resource-monitor) you must put something (not triggers or notify users) in the SET clause. The following SQL will work:

```
ALTER RESOURCE MONITOR "some_monitor" 
SET  CREDIT_QUOTA=10
TRIGGERS ON 100 PERCENT DO SUSPEND 
    ON 105 PERCENT DO SUSPEND_IMMEDIATE 
    ON 90 PERCENT DO NOTIFY
```

edit: The following will also work

```
ALTER RESOURCE MONITOR "some_monitor" 
TRIGGERS ON 100 PERCENT DO SUSPEND 
    ON 105 PERCENT DO SUSPEND_IMMEDIATE 
    ON 90 PERCENT DO NOTIFY
```

I would expect the provider to produce this SQL

**Code samples and commands**

Code samples are in the description already

**Additional context**

I assume the same issue exists for changing only the notify users but have not tested this.

sfc-gh-jcieslak:
Hey 👋 
[...]

Issue #1821, opened 2023-05-22
Terraform shows diff in start_timestamp of Snowflake's resource monitor  just after apply


LucasMMota:
**Provider Version**

provider registry.terraform.io/snowflake-labs/snowflake v0.63.0

**Terraform Version**

Terraform v1.4.5

**Describe the bug**

The start date of resource monitor cannot be set as saved in the database (e.g.: "2023-04-24T03:00:00Z"). Although if we use the expected format by the provider, the plan presents a difference just after the apply.

**Expected behavior**

No diff after the apply. Or provider accepts the format stored in DB with the timezone (e.g.: "2023-04-24T03:00:00Z")

**Code samples and commands**

The resource monitor object:

```
 {
    name            = "RM_WH_REPORTING"
    credit_quota    = 30
    frequency       = "DAILY"
    start_timestamp = "2023-05-23"
    notify_triggers = [80, 90]
    suspend_immediate_trigger = 100
    notify_users    = ["SNOWFLAKE_ALERTS_SLACK_INTEGRATION"]
    grant_privileges_by_role = {
      TRANSFORMER : ["MONITOR"]
    }
  }
```

The plan:

```terraform
 # module.resource_monitor["RM_WH_REPORTING"].snowflake_resource_monitor.monitor will be created
  + resource "snowflake_resource_monitor" "monitor" {
      + credit_quota              = 30
      + frequency                 = "DAILY"
      + id                        = (known after apply)
      + name                      = "RM_WH_REPORTING"
      + notify_triggers           = [
          + 80,
          + 90,
        ]
      + notify_users              = [
          + "SNOWFLAKE_ALERTS_SLACK_INTEGRATION",
        ]
      + set_for_account           = false
      + start_timestamp           = "2023-05-23"
      + suspend_immediate_trigger = 100
    }
```

The terraform plan **after applying**:

```terraform
  # module.resource_monitor["RM_WH_REPORTING"].snowflake_resource_monitor.monitor will be updated in-place
  ~ resource "snowflake_resource_monitor" "monitor" {
        id                        = "RM_WH_REPORTING"
        name                      = "RM_WH_REPORTING"
      ~ start_timestamp           = "2023-05-23T00:00:00Z" -> "2023-05-23"
```

LucasMMota:
also if you try to not use state time (to use the immediately option) this error will throw:
[...]

Issue #1754, opened 2023-04-25
Resource monitor creation fails, yet objects are created.


rasmus-kristensen:
**Provider Version**
0.62.0

**Terraform Version**

1.4.5

**Describe the bug**

When attempting to create new resource monitor in snowflake the "terraform init" runs successfully and as expected however the apply fails, yet the objects are still created in snowflake.
When attempting to terraform apply again it now complains that the objects already exists and are not found in the state.

![image](https://user-images.githubusercontent.com/106582326/234255330-c5d265b0-3baa-4405-906d-867fc588bda2.png)

```
│ Error: Plugin did not respond
│ 
│   with module.snowflake_monitor["MRCH"].snowflake_resource_monitor.monitor,
│   on resource-monitor/main.tf line 17, in resource "snowflake_resource_monitor" "monitor":
│   17: resource "snowflake_resource_monitor" "monitor" {
│ 
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ReadResource call. The plugin logs may contain more
│ details.
╵

Stack trace from the terraform-provider-snowflake_v0.62.0 plugin:

panic: runtime error: index out of range [0] with length 0

goroutine 37 [running]:
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources.ReadResourceMonitor(0x140008b6380, {0x102069960?, 0x140008b8b60})
        github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources/resource_monitor.go:322 +0x9cc
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0x102094140?, {0x102094140?, 0x1400084f560?}, 0xd?, {0x102069960?, 0x140008b8b60?})
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:712 +0x134
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).RefreshWithoutUpgrade(0x140005bb880, {0x102094140, 0x1400084f560}, 0x140009e6410, {0x102069960, 0x140008b8b60})
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.26.1/helper/schema/resource.go:1015 +0x468
[...]

PR #3052, opened 2024-09-09
feat: Resource monitor v1 readiness

sfc-gh-jcieslak:
## Changes
- Add ValuePresent assert to our custom assertions
- Add ToConfigValues function for every model
- Update Resource Monitor SDK + unit and integration tests
- Update Resource Monitor Resource + acc tests
- Handle issues connected to the resource monitor (mostly timestamp format difference causing infinite plan or only trigger updates causing SQL compilation error):
  -  #1500 
  - #1624 
  - #1716 
  - #1754 
  - #1821 
  - #1832 
  - #1990

## Next pr
- Adjust examples and update migration notes
- Data source (impl, tests, examples, migration notes)

## References
* [CREATE RESOURCE MONITOR](https://docs.snowflake.com/en/sql-reference/sql/create-resource-monitor)


pkg/acceptance/bettertestspoc/config/model/database_role_model_gen.go
```diff
@@ -43,6 +46,22 @@ func DatabaseRoleWithDefaultMeta(
 	return d
 }
 
+func (r *DatabaseRoleModel) ToConfigVariables() tfconfig.Variables {
```
sfc-gh-asawicki, line 49: We should not generate this method every time. Let's just have a generic function expecting the ResourceModel

pkg/sdk/testint/resource_monitors_integration_test.go
```diff
@@ -270,30 +319,23 @@ func TestInt_ResourceMonitorAlter(t *testing.T) {
 		newTriggers = append(newTriggers, sdk.TriggerDefinition{Threshold: 30, TriggerAction: sdk.TriggerActionNotify})
 
 		creditQuota := 100
-		alterOptions := &sdk.AlterResourceMonitorOptions{
+		err := client.ResourceMonitors.Alter(ctx, resourceMonitor.ID(), &sdk.AlterResourceMonitorOptions{
 			Set: &sdk.ResourceMonitorSet{
 				CreditQuota: &creditQuota,
 				NotifyUsers: &sdk.NotifyUsers{
-					Users: []sdk.NotifiedUser{{Name: "ARTUR_SAWICKI"}},
+					Users: []sdk.NotifiedUser{{Name: sdk.NewAccountObjectIdentifier("JAN_CIESLAK")}},
```
sfc-gh-asawicki, line 326: xD

pkg/sdk/resource_monitors_test.go
```diff
@@ -156,3 +180,30 @@ func TestResourceMonitorShow(t *testing.T) {
...
+func TestExtractTriggerInts(t *testing.T) {
+	testCases := []struct {
+		Input    sql.NullString
+		Expected []int
+		Error    string
+	}{
+		{Input: sql.NullString{String: "51%,63%,123%", Valid: true}, Expected: []int{51, 63, 123}},
+		{Input: sql.NullString{String: "51%,63%", Valid: true}, Expected: []int{51, 63}},
+		{Input: sql.NullString{String: "51%", Valid: true}, Expected: []int{51}},
+		{Input: sql.NullString{String: "", Valid: false}, Expected: []int{}},
```
sfc-gh-asawicki, line 193: ony othe invalid options?

pkg/resources/resource_monitor.go
```diff
@@ -2,504 +2,450 @@ package resources
...
+			Type:             schema.TypeInt,
+			ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(1)),
+		},
 	},
-	"set_for_account": {
-		Type:        schema.TypeBool,
-		Optional:    true,
-		Description: "Specifies whether the resource monitor should be applied globally to your Snowflake account (defaults to false).",
-		Default:     false,
+	"suspend_trigger": {
```
sfc-gh-asawicki, line 73: should these two be also added IgnoreChangeToCurrentSnowflakeValueInShow ?

pkg/resources/resource_monitor.go
```diff
@@ -2,504 +2,450 @@ package resources
...
-
-		userNames := expandStringList(d.Get("notify_users").(*schema.Set).List())
-		users := []sdk.NotifiedUser{}
-		for _, name := range userNames {
-			users = append(users, sdk.NotifiedUser{Name: name})
-		}
-		set.NotifyUsers = &sdk.NotifyUsers{
-			Users: users,
+		if v, ok := d.GetOk("end_timestamp"); ok {
+			runSetStatement = true
```
sfc-gh-asawicki, line 338: runSetStatement and runUnsetStatement are left on purpose? (we are not using them any more in the reworked resources)

pkg/resources/resource_monitor.go
```diff
@@ -2,504 +2,450 @@ package resources
...
-	if suspendTrigger != nil {
-		triggers = append(triggers, *suspendTrigger)
-	}
-	var suspendImmediateTrigger *sdk.TriggerDefinition
 
-	if v, ok := d.GetOk("suspend_immediate_trigger"); ok {
-		suspendImmediateTrigger = &sdk.TriggerDefinition{
-			Threshold:     v.(int),
-			TriggerAction: sdk.TriggerActionSuspendImmediate,
+		if err := client.ResourceMonitors.Alter(ctx, id, &opts); err != nil {
```
sfc-gh-asawicki, line 421: dont we need Partial in both of them? (there are some values we don't read from SF, so I guess there may be a need for it)

pkg/acceptance/bettertestspoc/config/model/database_role_model_gen.go
sfc-gh-jcieslak: 👍 Placed it next to FromModel function in config package

pkg/sdk/testint/resource_monitors_integration_test.go
sfc-gh-jcieslak: hihi

pkg/resources/resource_monitor.go
```diff
@@ -2,504 +2,450 @@ package resources
...
-				Set: &sdk.WarehouseSet{
-					ResourceMonitor: objectIdentifier,
+		if len(triggers) > 0 {
+			opts.Triggers = triggers
+		} else {
+			return diag.Diagnostics{
+				diag.Diagnostic{
+					Severity: diag.Error,
+					Summary:  "Failed to update resource monitor.",
+					Detail:   "Due to Snowflake limitations triggers cannot be completely removed form resource monitor after having at least 1 trigger. The only way it to re-create resource monitor without any triggers specified.",
```
sfc-gh-jmichalak, line 398: Add this to the docs as well

pkg/sdk/testint/resource_monitors_integration_test.go
sfc-gh-asawicki: Yes, that's the reason my name was there in the first place :D (check https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/78ecb62935c1a40611418e59ed5f0d644cf85f03/pkg/resources/email_notification_integration_acceptance_test.go#L14)

pkg/sdk/resource_monitors.go
```diff
@@ -89,70 +99,65 @@ func (row *resourceMonitorRow) convert() (*ResourceMonitor, error) {
 		resourceMonitor.RemainingCredits = remainingCredits
 	}
 
+	if row.Level.Valid {
+		switch row.Level.String {
+		case "ACCOUNT":
+			resourceMonitor.Level = Pointer(ResourceMonitorLevelAccount)
+		case "WAREHOUSE":
+			resourceMonitor.Level = Pointer(ResourceMonitorLevelWarehouse)
```
sfc-gh-jmichalak, line 107: This could be extracted

pkg/sdk/resource_monitors_test.go
sfc-gh-jcieslak: Added and adjusted implementation a bit

pkg/resources/resource_monitor.go
sfc-gh-jcieslak: True, I removed them

sfc-gh-jmichalak (changes requested):

framework/provider/resource_monitor_resource.go
```diff
@@ -600,13 +600,11 @@ func (r *ResourceMonitorResource) read(ctx context.Context, data *resourceMonito
 
 	data.CreditQuota = types.Float64Value(resourceMonitor.CreditQuota)
 	data.Frequency = types.StringValue(string(resourceMonitor.Frequency))
-	switch resourceMonitor.Level {
+	switch *resourceMonitor.Level {
 	case sdk.ResourceMonitorLevelAccount:
 		data.Level = types.StringValue("ACCOUNT")
 	case sdk.ResourceMonitorLevelWarehouse:
 		data.Level = types.StringValue("WAREHOUSE")
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I didn't pay much attention to this file, because it's an old POC for the provider to use framework, we don't really use this code, so I just wanted to get rid of errors produced by `pre-push`.

pkg/sdk/resource_monitors.go
sfc-gh-jcieslak: Done

pkg/resources/resource_monitor.go
sfc-gh-jcieslak: I touch the documentation in the next pr, I'll address this comment there.

pkg/resources/resource_monitor.go
sfc-gh-jcieslak: DM'd you but most likely, yes. I'll add them.

pkg/resources/resource_monitor.go
sfc-gh-jcieslak: True, I also added test with external changes, but I'm not sure if that showed IgnoreChangeToCurrentSnowflakeValueInShow worked for those fields (I don't think so).

pkg/resources/resource_monitor.go
sfc-gh-jcieslak: Added and wrote the test showing it would produce strange plans after failure without it.

sfc-gh-jmichalak (dismissed):

pkg/resources/resource_monitor_acceptance_test.go
```diff
@@ -338,6 +344,154 @@ func TestAcc_ResourceMonitor_Updates(t *testing.T) {
...
+		PreCheck:                 func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: acc.CheckDestroy(t, resources.ResourceMonitor),
+		Steps: []resource.TestStep{
+			{
+				Config: config.FromModel(t, configModelEverythingSet),
+			},
+			// Update externally, but match the updated configuration (expected updates to the same values)
```
sfc-gh-asawicki, line 389: hmmm, but why we expect changes if we have ignorechange to current snowflake value? (e.g. check user_acceptance_test `CHANGE PROP TO THE CURRENT SNOWFLAKE VALUE` in `TestAcc_User_BasicFlows`). Isn't the triggers collection causing the update here? Can we check what happens when we only alter the basic attributes?

sfc-gh-asawicki (changes requested):

pkg/resources/resource_monitor_acceptance_test.go
sfc-gh-jcieslak: From the printed plan I saw that `notify_triggers` and `notify_users` are computed (because we have no diff suppression for them) causing `show_output` to have changes. But even when I changed them to be the same in step 1 and 2 (and didn't change them externally) they were still computed causing `show_output` to change and produce the plan. Ultimately, I wanted to put ExpectNoPlan() plan check, but couldn't because of that (I think).

sfc-gh-asawicki (approved):

PR #3063, opened 2024-09-11
Row access policy v1

sfc-gh-jmichalak:
- add show and desc output
- rename fields
- change signature to arguments
- implement renaming
- fix permadiff on body
- adjust identifiers handling
- adjust examples
- gen resource asserts (config builders are not working because we have a required list argument)
- improve handling data types
- move parsing signature to sdk
- support proper casing in arg names

## Test Plan

* [x] acceptance tests

* [ ] …

## References

https://docs.snowflake.com/en/sql-reference/sql/create-row-access-policy

## TODO (next PR)
- rework data source


MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,39 @@ This document is meant to help you migrate your Terraform config to the new newe
...
+#### *(breaking change)* Adjusted behavior of arguments/signature
+Now, arguments are stored as a list, instead of a map. Please adjust that in your configs. State is migrated automatically. Also, this means that order of the items matters and may be adjusted.
+
+Argument names are now case sensitive.
+
+#### *(breaking change)* Adjusted behavior on changing name
+Previously, after changing `name` field, the resource was recreated. Now, the object is renamed with `RENAME TO`.
+
+#### *(breaking change)* Mitigating permadiff on `body`
+Previously, `body` of a policy was compared as a raw string. This led to permament diff because of leading newlines (see https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2053).
```
sfc-gh-asawicki, line 28: Please add to the PR description all issues that are referenced in this change.

MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,39 @@ This document is meant to help you migrate your Terraform config to the new newe
...
+
+#### *(breaking change)* Adjusted behavior of arguments/signature
+Now, arguments are stored as a list, instead of a map. Please adjust that in your configs. State is migrated automatically. Also, this means that order of the items matters and may be adjusted.
+
+Argument names are now case sensitive.
+
+#### *(breaking change)* Adjusted behavior on changing name
+Previously, after changing `name` field, the resource was recreated. Now, the object is renamed with `RENAME TO`.
+
+#### *(breaking change)* Mitigating permadiff on `body`
```
sfc-gh-asawicki, line 27: is it a breaking change or a bugfix?

docs/resources/row_access_policy.md
```diff
@@ -16,14 +18,14 @@ resource "snowflake_row_access_policy" "example_row_access_policy" {
   name     = "EXAMPLE_ROW_ACCESS_POLICY"
   database = "EXAMPLE_DB"
   schema   = "EXAMPLE_SCHEMA"
-  signature = {
-    A = "VARCHAR",
-    B = "VARCHAR"
+  argument {
```
sfc-gh-asawicki, line 21: please add an example with multiple arguments

MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,39 @@ This document is meant to help you migrate your Terraform config to the new newe
...
+Previously, `body` of a policy was compared as a raw string. This led to permament diff because of leading newlines (see https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2053).
+
+Now, similarly to handling statements in other resources, we replace blank characters with a space. The provider can cause false positives in cases where a change in case or run of whitespace is semantically significant.
+
+#### *(breaking change)* Identifiers related changes
```
sfc-gh-asawicki, line 32: let's mention that it changes importing (dots instead of pipes)

pkg/acceptance/bettertestspoc/config/model/view_model_gen.go
```diff
@@ -12,6 +12,7 @@ import (
 type ViewModel struct {
 	AggregationPolicy  tfconfig.Variable `json:"aggregation_policy,omitempty"`
 	ChangeTracking     tfconfig.Variable `json:"change_tracking,omitempty"`
+	Column             tfconfig.Variable `json:"column,omitempty"`
```
sfc-gh-asawicki, line 15: Please generate the model despite the fact that the list is required and even use it by utilizing the `ConfigVariablesFromModel`

pkg/sdk/row_access_policies_gen.go
```diff
@@ -110,3 +111,21 @@ type RowAccessPolicyDescription struct {
 	ReturnType string
 	Body       string
 }
+
+// TODO(SNOW-1596962): Fully support VECTOR data type
+func (d *RowAccessPolicyDescription) Arguments() []map[string]any {
```
sfc-gh-jcieslak, line 116: We had so many parsing functions for arguments, don't we have one that could be used here.

pkg/sdk/row_access_policies_gen_test.go
```diff
@@ -49,7 +53,7 @@ func TestRowAccessPolicies_Create(t *testing.T) {
 
 	t.Run("one parameter", func(t *testing.T) {
 		opts := defaultOpts()
-		assertOptsValidAndSQLEquals(t, opts, "CREATE ROW ACCESS POLICY %s AS (n VARCHAR) RETURNS BOOLEAN -> true", id.FullyQualifiedName())
+		assertOptsValidAndSQLEquals(t, opts, `CREATE ROW ACCESS POLICY %s AS ("n" VARCHAR) RETURNS BOOLEAN -> true`, id.FullyQualifiedName())
```
sfc-gh-jcieslak, line 56: That also breaking change that could be mentioned in the migration guide. Now, we make arguments case sensitive, so it's different on Snowflake side (they have to reference them with quotes). Btw that doesn't need a state upgrader that would convert names to uppercase in the newer version ?

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -32,54 +42,393 @@ func TestAcc_RowAccessPolicy(t *testing.T) {
 			{
 				ConfigDirectory: config.TestStepDirectory(),
 				ConfigVariables: m(),
```
sfc-gh-jcieslak, line 44: You can use `ConfigVariablesFromModel` function.

pkg/sdk/row_access_policies_gen.go
```diff
@@ -29,7 +30,7 @@ type CreateRowAccessPolicyOptions struct {
 }
 
 type CreateRowAccessPolicyArgs struct {
-	Name string   `ddl:"keyword,no_quotes"`
+	Name string   `ddl:"keyword,double_quotes"`
```
sfc-gh-asawicki, line 33: won't this be a breaking change for the users? I mean, it is added in the migration guide, that now it will be case-sensitive but what will happen with the policies created before? They are all uppercase now in Snowflake, right? Shouldn't we then advise to change all the names in configs to uppercase because of this change? Can we change what happens: if we do, and if we don't?

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -32,54 +42,393 @@ func TestAcc_RowAccessPolicy(t *testing.T) {
...
-					resource.TestCheckResourceAttr("snowflake_row_access_policy.test", "signature.N", "VARCHAR"),
-					resource.TestCheckResourceAttr("snowflake_row_access_policy.test", "signature.V", "VARCHAR"),
+				Check: assert.AssertThat(t, resourceassert.RowAccessPolicyResource(t, resourceName).
+					HasNameString(id.Name()).
+					HasDatabaseString(id.DatabaseName()).
+					HasSchemaString(id.SchemaName()).
+					HasFullyQualifiedNameString(id.FullyQualifiedName()).
+					HasCommentString("Terraform acceptance test").
+					HasBodyString("case when current_role() in ('ANALYST') then true else false end"),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.#", "2")),
```
sfc-gh-jcieslak, line 52: Hmm that way is easier, but maybe having resourceassert extended and having those asserts in assert builder would be better ? wdyt

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -32,54 +42,393 @@ func TestAcc_RowAccessPolicy(t *testing.T) {
...
+						HasCreatedOnNotEmpty().
+						HasDatabaseName(id.DatabaseName()).
+						HasKind(string(sdk.PolicyKindRowAccessPolicy)).
+						HasName(id.Name()).
+						HasOptions("").
+						HasOwner(snowflakeroles.Accountadmin.Name()).
+						HasOwnerRoleType("ROLE").
+						HasSchemaName(id.SchemaName()).
+						HasComment("Terraform acceptance test"),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "describe_output.0.body", "case when current_role() in ('ANALYST') then true else false end")),
```
sfc-gh-jcieslak, line 67: I think there was example in user on creating a simple ext to assert show output. I guess it could be easily altered to make it work with desc output.

pkg/sdk/row_access_policies_gen.go
```diff
@@ -110,3 +111,21 @@ type RowAccessPolicyDescription struct {
...
+	// Format in database is `(column <data_type>)`
+	plainSignature := strings.ReplaceAll(d.Signature, "(", "")
+	plainSignature = strings.ReplaceAll(plainSignature, ")", "")
+	signatureParts := strings.Split(plainSignature, ", ")
+	arguments := make([]map[string]any, len(signatureParts))
+
+	for i, e := range signatureParts {
+		parts := strings.Split(e, " ")
+		arguments[i] = map[string]any{
+			"name": strings.Join(parts[:len(parts)-1], " "),
```
sfc-gh-asawicki, line 126: is there some spoecific use case in which we can get list longer than two elements here? If not, can we assert that it should be 2-elements long?

pkg/sdk/row_access_policies_gen_test.go
```diff
@@ -243,3 +247,55 @@ func TestRowAccessPolicies_Describe(t *testing.T) {
 		assertOptsValidAndSQLEquals(t, opts, "DESCRIBE ROW ACCESS POLICY %s", id.FullyQualifiedName())
 	})
 }
+
+func TestRowAccessPolicyDescription_Arguments(t *testing.T) {
```
sfc-gh-asawicki, line 251: can we get a signature like (A NUMBER(32, 4)) ? If so, then the implementation needs to be adjusted.

pkg/sdk/row_access_policies_gen_test.go
```diff
@@ -243,3 +247,55 @@ func TestRowAccessPolicies_Describe(t *testing.T) {
...
+				},
+				{
+					"name": "B",
+					"type": "BOOLEAN",
+				},
+			},
+		},
+		{
+			name:      "signature with complex name",
+			signature: "(a B VARCHAR)",
```
sfc-gh-asawicki, line 283: is this a valid name?

pkg/sdk/row_access_policies_gen.go
sfc-gh-asawicki, line 116: for the logic like this, it would be great to add an integration test with different arguments combinations (different data types, especially with number parameters) and assert what is returned in signature (that way we will know when Snowflake changes something)

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -32,54 +42,393 @@ func TestAcc_RowAccessPolicy(t *testing.T) {
...
+	body := "case when current_role() in ('ANALYST') then true else false end"
+	m := config.Variables{
+		"name":     config.StringVariable(id.Name()),
+		"database": config.StringVariable(id.DatabaseName()),
+		"schema":   config.StringVariable(id.SchemaName()),
+		"arguments": config.SetVariable(
+			config.MapVariable(map[string]config.Variable{
+				"name": config.StringVariable("A"),
+				"type": config.StringVariable("VARCHAR"),
+			}),
```
sfc-gh-jcieslak, line 388: Could you add more arguments and maybe an argument with lowercase letters to prove my point with arg names when upgrading?

sfc-gh-jcieslak (changes requested):

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-asawicki: +1

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-asawicki: +1

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-asawicki: it's a bit different in user because what was added was to assert showoutput inside a datasource. But. There is another option of adding the describe to the resourceshowoutputassert by adding it to the `objectassertgen.GetSdkObjectDetails`, generating, and adding a proper constructors as ext. But I don't know if this is worth a hassle now.

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -32,54 +42,393 @@ func TestAcc_RowAccessPolicy(t *testing.T) {
...
+					HasCommentString("Terraform acceptance test - changed comment").
+					HasBodyString("case when current_role() in ('ANALYST') then false else true end"),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.#", "2")),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.0.name", "N")),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.0.type", "VARCHAR")),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.1.name", "V")),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.1.type", "VARCHAR")),
 				),
 			},
 			// change signature
```
sfc-gh-asawicki, line 91: please change the singature externally as a separate step (in a few ways); I think that it would be best to have a separate test case for just signature modifications: in config and external

pkg/resources/row_access_policy.go
```diff
@@ -4,88 +4,147 @@ import (
...
+		DeleteContext: DeleteRowAccessPolicy,
 
 		Schema: rowAccessPolicySchema,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
+
+		CustomizeDiff: customdiff.All(
+			ComputedIfAnyAttributeChanged(rowAccessPolicySchema, ShowOutputAttributeName, "comment"),
+			ComputedIfAnyAttributeChanged(rowAccessPolicySchema, DescribeOutputAttributeName, "body"),
```
sfc-gh-asawicki, line 114: what about the signature?

pkg/resources/row_access_policy.go
```diff
@@ -4,88 +4,147 @@ import (
...
+					Type:        schema.TypeString,
+					Required:    true,
+					Description: "The argument name",
+					ForceNew:    true,
+				},
+				// TODO(SNOW-1596962): Fully support VECTOR data type sdk.ParseFunctionArgumentsFromString could be a base for another function that takes argument names into consideration.
+				"type": {
+					Type:     schema.TypeString,
+					Required: true,
+					DiffSuppressFunc: func(key, oldValue, newValue string, _ *schema.ResourceData) bool {
```
sfc-gh-asawicki, line 54: why not just `DiffSuppressFunc: NormalizeAndCompare(sdk.ToDataType)`?

pkg/resources/row_access_policy.go
```diff
@@ -118,101 +178,112 @@ func ReadRowAccessPolicy(d *schema.ResourceData, meta interface{}) error {
 		return nil
 	}
 	if err := d.Set(FullyQualifiedNameAttributeName, id.FullyQualifiedName()); err != nil {
-		return err
+		return diag.FromErr(err)
 	}
 
 	if err := d.Set("name", rowAccessPolicy.Name); err != nil {
```
sfc-gh-asawicki, line 184: why do we set the required fields (name, schema and db)?

sfc-gh-asawicki (changes requested):

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-jmichalak: I left this because I assume we'd like to have it in a separate pkg, like `resourcedescribeoutputassert`.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: A breaking change because now spacing can affect body when spaces are semantically significant. Before this wasn't the case. Also, permadiff wasn't happening for all queries, so some users could see a regression here.

pkg/sdk/row_access_policies_gen_test.go
sfc-gh-jmichalak: No.

pkg/resources/row_access_policy.go
sfc-gh-jmichalak: Removed and added in the import function.

pkg/resources/row_access_policy.go
sfc-gh-jmichalak: Added.

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/row_access_policies_gen.go
sfc-gh-jmichalak: Done.

pkg/sdk/row_access_policies_gen.go
sfc-gh-jmichalak: Right, they are upper case now. Added a suggestion to the migration guide.

pkg/acceptance/bettertestspoc/config/model/view_model_gen.go
sfc-gh-jmichalak: Done.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: Done.

docs/resources/row_access_policy.md
sfc-gh-jmichalak: Done.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: Done.

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/row_access_policies_gen_test.go
sfc-gh-jmichalak: It's mentioned in the migration guide. Adjusted the state upgrader and test.

pkg/sdk/row_access_policies_gen.go
sfc-gh-jmichalak: But `ParseFunctionArgumentsFromString` doesn't return arg names. Since we already have a follow-up pointed in the comment, I'll leave this for now. This signature has an odd format, I will report this internally.

pkg/resources/row_access_policy.go
```diff
@@ -4,88 +4,147 @@ import (
...
 
 		Schema: rowAccessPolicySchema,
 		Importer: &schema.ResourceImporter{
 			StateContext: schema.ImportStatePassthroughContext,
 		},
+
+		CustomizeDiff: customdiff.All(
+			ComputedIfAnyAttributeChanged(rowAccessPolicySchema, ShowOutputAttributeName, "comment"),
+			ComputedIfAnyAttributeChanged(rowAccessPolicySchema, DescribeOutputAttributeName, "body"),
+			ComputedIfAnyAttributeChanged(rowAccessPolicySchema, FullyQualifiedNameAttributeName, "name"),
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Done.

examples/resources/snowflake_row_access_policy/resource.tf
```diff
@@ -2,9 +2,10 @@ resource "snowflake_row_access_policy" "example_row_access_policy" {
   name     = "EXAMPLE_ROW_ACCESS_POLICY"
   database = "EXAMPLE_DB"
   schema   = "EXAMPLE_SCHEMA"
-  signature = {
-    A = "VARCHAR",
-    B = "VARCHAR"
+  argument {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Added.

examples/resources/snowflake_row_access_policy/import.sh
```diff
@@ -1,2 +1 @@
-# format is database name | schema name | policy name
-terraform import snowflake_row_access_policy.example 'dbName|schemaName|policyName'
+terraform import snowflake_row_access_policy.example '"<database_name>"."<schema_name>"."<view_name>"'
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Fixed.

pkg/acceptance/bettertestspoc/config/model/row_access_policy_model_ext.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	tfconfig "github.com/hashicorp/terraform-plugin-testing/config"
+)
+
+func (r *RowAccessPolicyModel) WithArgument(argument []sdk.RowAccessPolicyArgument) *RowAccessPolicyModel {
+	maps := make([]config.Variable, len(argument))
+	for i, v := range argument {
+		maps[i] = config.MapVariable(map[string]config.Variable{
```
sfc-gh-jcieslak, line 12: Shouldn't this be ObjectVariable instead?

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -40,20 +46,25 @@ func TestAcc_RowAccessPolicy(t *testing.T) {
...
 					HasFullyQualifiedNameString(id.FullyQualifiedName()).
 					HasCommentString("Terraform acceptance test").
-					HasBodyString("case when current_role() in ('ANALYST') then true else false end"),
-					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.#", "2")),
-					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.0.name", "N")),
-					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.0.type", "VARCHAR")),
-					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.1.name", "V")),
-					assert.Check(resource.TestCheckResourceAttr(resourceName, "argument.1.type", "VARCHAR")),
+					HasBodyString(body).
+					HasArguments([]sdk.RowAccessPolicyArgument{
```
sfc-gh-jcieslak, line 58: you can create a variable and reuse it in this test

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -119,17 +177,54 @@ func TestAcc_RowAccessPolicy(t *testing.T) {
...
+							Type: string(sdk.DataTypeBoolean),
+						},
+						{
+							Name: "X",
+							Type: string(sdk.DataTypeTimestampNTZ),
+						},
+					}),
+				),
+			},
+			// unset comment
```
sfc-gh-jcieslak, line 193: move this comment to the step below

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -404,29 +475,48 @@ func TestAcc_view_migrateFromVersion_0_95_0(t *testing.T) {
...
+				},
+				ExpectNonEmptyPlan: true,
 				Check: assert.AssertThat(t, resourceassert.RowAccessPolicyResource(t, resourceName).
 					HasNameString(id.Name()).
 					HasDatabaseString(id.DatabaseName()).
 					HasSchemaString(id.SchemaName()).
 					HasFullyQualifiedNameString(id.FullyQualifiedName()),
 					assert.Check(resource.TestCheckResourceAttr(resourceName, "row_access_expression", body)),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "signature.A", "VARCHAR")),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "signature.B", "VARCHAR")),
```
sfc-gh-jcieslak, line 492: Why it's lower case in config, here it's with upper case and in the latest version it's with lower case? In the state upgrader you upper case'd it, so it should be `B` right ?

pkg/resources/testdata/TestAcc_RowAccessPolicy/basic/variables.tf
```diff
@@ -10,7 +10,7 @@ variable "schema" {
   type = string
 }
 
-variable "arguments" {
+variable "argument" {
   type = set(map(string))
```
sfc-gh-jcieslak, line 14: not that important, but if you would change MapVariable to ObjectVariable when building model this should be
```terraform
variable "argument" {
  type = set(object({
    name = string
    type = string
  }))
}
```
and test.tf's should be adjusted to use object with fields instead of map

pkg/sdk/row_access_policies_gen.go
sfc-gh-jcieslak: Yeah, but maybe we should add a flag to `ParseFunctionArgumentsFromString` to check for argument names or not. Maybe create a ticket in the identifier rework follow-up and reference it here, so we won't forget about it.

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-jmichalak: Done.

sfc-gh-jcieslak (approved):

pkg/sdk/row_access_policies_gen.go
sfc-gh-asawicki: Please also add that they will be migrated to the uppercase in state automatically

pkg/sdk/testint/row_access_policies_gen_integration_test.go
```diff
@@ -350,3 +350,60 @@ func TestInt_RowAccessPoliciesShowByID(t *testing.T) {
 		require.Equal(t, id2, e2.ID())
 	})
 }
+
+func TestInt_RowAccessPoliciesDescribe(t *testing.T) {
+	client := testClient(t)
+	ctx := testContext(t)
+
+	cleanupRowAccessPolicyHandle := func(id sdk.SchemaObjectIdentifier) func() {
```
sfc-gh-asawicki, line 358: use DropRowAccessPolicyFunc here

pkg/sdk/testint/row_access_policies_gen_integration_test.go
```diff
@@ -350,3 +350,60 @@ func TestInt_RowAccessPoliciesShowByID(t *testing.T) {
...
+		return func() {
+			err := client.RowAccessPolicies.Drop(ctx, sdk.NewDropRowAccessPolicyRequest(id))
+			if errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) {
+				return
+			}
+			require.NoError(t, err)
+		}
+	}
+
+	createRowAccessPolicyHandle := func(t *testing.T, id sdk.SchemaObjectIdentifier, args []sdk.CreateRowAccessPolicyArgsRequest) {
```
sfc-gh-asawicki, line 368: we can use the creation from the test client heper here

pkg/sdk/testint/row_access_policies_gen_integration_test.go
```diff
@@ -350,3 +350,60 @@ func TestInt_RowAccessPoliciesShowByID(t *testing.T) {
...
+
+		err := client.RowAccessPolicies.Create(ctx, sdk.NewCreateRowAccessPolicyRequest(id, args, "true"))
+		require.NoError(t, err)
+		t.Cleanup(cleanupRowAccessPolicyHandle(id))
+	}
+
+	t.Run("describe", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+
+		arg1 := sdk.NewCreateRowAccessPolicyArgsRequest(random.AlphaN(5), sdk.DataTypeVARCHAR)
```
sfc-gh-asawicki, line 379: please add moar (check e.g. `function returns non detailed data types of arguments` test)

pkg/resources/row_access_policy_acceptance_test.go
```diff
@@ -23,13 +24,29 @@ func TestAcc_RowAccessPolicy(t *testing.T) {
...
+		{
+			Name: "N",
+			Type: string(sdk.DataTypeVARCHAR),
+		},
+		{
+			Name: "V",
+			Type: string(sdk.DataTypeVARCHAR),
+		},
+	}
+	changedArgument := []sdk.RowAccessPolicyArgument{
```
sfc-gh-asawicki, line 39: how is it changed comparing to the argument above?

sfc-gh-asawicki (changes requested):

pkg/resources/row_access_policy_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/row_access_policies_gen_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/row_access_policies_gen_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/row_access_policies_gen_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/row_access_policies_gen.go
sfc-gh-jmichalak: Done.

sfc-gh-asawicki (approved):

PR #3078, opened 2024-09-16
feat: Masking policy resource v1

sfc-gh-jmichalak:
- SDK
  - improve validations
  - add tests
  - improve options parsing (also, get rid of a 3rd party library)
  - add a TODO to reuse parsing from row access policies after merging https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3079
- resource
  - use new id handling
  - rename/remove some fields to be consistent with other resources
  - better ACC tests
-  misc
  - adjust helper clients
  - generate config and assertion builders
  - adjust docs

## Test Plan

* [x] acceptance tests
* [x] integration tests
* [x] unit tests 

## References

https://docs.snowflake.com/en/sql-reference/sql/create-masking-policy

## TODO
- add tests for issues
- rework data source


pkg/resources/masking_policy.go
```diff
@@ -2,282 +2,339 @@ package resources
...
-			return old != new
-		},
-	},
 	"name": {
-		Type:        schema.TypeString,
-		Required:    true,
-		Description: "Specifies the identifier for the masking policy; must be unique for the database and schema in which the masking policy is created.",
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      blocklistedCharactersFieldDescription("Specifies the identifier for the masking policy; must be unique for the database and schema in which the masking policy is created."),
```
sfc-gh-jcieslak, line 25: blocklisted characters description could be also used on database and schema

pkg/resources/masking_policy.go
```diff
@@ -2,282 +2,339 @@ package resources
...
+
+	// set optionals
 	opts := &sdk.CreateMaskingPolicyOptions{}
 	if comment, ok := d.Get("comment").(string); ok {
 		opts.Comment = sdk.String(comment)
 	}
-	if exemptOtherPolicies := d.Get("exempt_other_policies").(bool); exemptOtherPolicies {
-		opts.ExemptOtherPolicies = sdk.Bool(exemptOtherPolicies)
+	if v := d.Get("exempt_other_policies").(string); v != BooleanDefault {
+		parsed, err := strconv.ParseBool(v)
```
sfc-gh-jcieslak, line 221: we have custom function for bool value parsing

pkg/resources/masking_policy.go
```diff
@@ -2,282 +2,339 @@ package resources
...
+		client := meta.(*provider.Context).Client
+		id, err := sdk.ParseSchemaObjectIdentifier(d.Id())
+		if err != nil {
+			return diag.FromErr(err)
+		}
 
-	if err := d.Set("name", maskingPolicy.Name); err != nil {
-		return err
-	}
+		maskingPolicy, err := client.MaskingPolicies.ShowByID(ctx, id)
```
sfc-gh-jcieslak, line 246: Could you check ShowByID implementation if it's using collections.FindFirst? I think in main it's not using it and the `if errors.Is(err, sdk.ErrObjectNotFound)` branch will be never entered.

pkg/resources/masking_policy_acceptance_test.go
```diff
@@ -216,3 +369,212 @@ func TestAcc_MaskingPolicy_migrateFromVersion_0_94_1(t *testing.T) {
...
+					},
+				},
+				Config: maskingPolicy_v0_95_0(id, body, comment),
+				Check: assert.AssertThat(t, resourceassert.MaskingPolicyResource(t, resourceName).
+					HasNameString(id.Name()).
+					HasDatabaseString(id.DatabaseName()).
+					HasSchemaString(id.SchemaName()).
+					HasCommentString(comment).
+					HasFullyQualifiedNameString(id.FullyQualifiedName()),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "masking_expression", body)),
```
sfc-gh-jcieslak, line 585: add one assert on "id" field to make sure it changed

pkg/schemas/masking_policy.go
```diff
@@ -0,0 +1,65 @@
...
+	signatureElem := make([]map[string]any, len(details.Signature))
+	for i, v := range details.Signature {
+		signatureElem[i] = map[string]any{
+			"name": v.Name,
+			"type": string(v.Type),
+		}
+	}
+	return map[string]any{
+		"name":        details.Name,
+		"signature":   signatureElem,
```
sfc-gh-jcieslak, line 50: You can use here the function below to map signature

pkg/sdk/masking_policy.go
```diff
@@ -255,9 +280,11 @@ type maskingPolicyDBRow struct {
 }
 
 func (row maskingPolicyDBRow) convert() *MaskingPolicy {
-	exemptOtherPolicies, err := jsonparser.GetBoolean([]byte(row.Options), "EXEMPT_OTHER_POLICIES")
+	options, err := ParseMaskingPolicyOptions(row.Options)
 	if err != nil {
-		exemptOtherPolicies = false
+		log.Printf("converting masking policy row: error unmarshaling options: %v", err)
+		log.Printf("setting exempt_other_policies = false")
```
sfc-gh-jcieslak, line 286: prefix with `[DEBUG]`

pkg/sdk/masking_policy.go
```diff
@@ -346,6 +373,7 @@ func (row maskingPolicyDetailsRow) toMaskingPolicyDetails() *MaskingPolicyDetail
 		ReturnType: dataType,
 		Body:       row.Body,
 	}
+	// TODO (after merged changes in row access policies) use/merge with parsing in row access policies
```
sfc-gh-jcieslak, line 376: Leaving a comment as a reminder (will resolve after the upcoming change)

MIGRATION_GUIDE.md
```diff
@@ -5,6 +5,40 @@ describe deprecations or breaking changes and help you to change your configurat
...
+
+### snowflake_masking_policy resource changes
+New fields:
+  - `show_output` field that holds the response from SHOW MASKING POLICIES.
+  - `describe_output` field that holds the response from DESCRIBE MASKING POLICY.
+
+#### *(breaking change)* Renamed fields in snowflake_masking_policy resource
+Renamed fields:
+  - `masking_expression` to `body`
+  - `signature` to `arguments`
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Yes, it's also mentioned below.

pkg/sdk/masking_policy.go
sfc-gh-jmichalak: Done.

pkg/schemas/masking_policy.go
sfc-gh-jmichalak: Done.

pkg/resources/masking_policy_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/masking_policy.go
sfc-gh-jmichalak: Good point, I changed the sdk implementation to use collections.FindFirst.

pkg/resources/masking_policy.go
sfc-gh-jmichalak: Changed.

pkg/resources/masking_policy.go
sfc-gh-jmichalak: Done.

MIGRATION_GUIDE.md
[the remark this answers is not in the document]
sfc-gh-jcieslak: Still, it suggests here to rename signature to arguments. Just to be clear for the customers add simple example of adjusting columns like here https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#behavior-change-resource-renamed-snowflake_database---snowflake_database_old.

pkg/sdk/masking_policy.go
sfc-gh-jmichalak: Ok, done. I adjusted this also for row access policies.

sfc-gh-jcieslak (approved):

sfc-gh-jcieslak (approved):

pkg/resources/masking_policy_acceptance_test.go
```diff
@@ -2,79 +2,300 @@ package resources_test
...
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "describe_output.0.signature.1.name", "B")),
+					assert.Check(resource.TestCheckResourceAttr(resourceName, "describe_output.0.signature.1.type", string(sdk.DataTypeVARCHAR))),
 				),
 			},
-			// rename + change comment
+			// change fields
 			{
-				Config: maskingPolicyConfig(newId.Name(), comment2, acc.TestDatabaseName, acc.TestSchemaName),
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_MaskingPolicy/complete"),
+				ConfigVariables: tfconfig.ConfigVariablesFromModel(t, policyModel.WithBody(bodyWithBooleanReturnType).WithReturnDataType(string(sdk.DataTypeBoolean)).WithArgument(argumentWithChangedFirstArgumentType).WithComment("Terraform acceptance test - changed comment").WithExemptOtherPolicies(r.BooleanFalse)),
```
sfc-gh-asawicki, line 123: The way that policyModel is reused here is interesting/unsafe - that was not the initial intention to reuse it this way. It bases on the order of execution and is a bit misleading (e.g. in step 4 and step 5 we want to validate only external change but we use "different" config variables - different only in the written way, underneath it's the same pointer, with fields already set in the previous step).

Maybe the design of model is a bit flawed and we should not use the pointer but the model directly, so that we always get the copied struct. Either way, I am raising it as a topic to discuss internally.

Issue #3073, opened 2024-09-13
[Bug]: column block-list is changing in every plan (upgrade to v0.95.0 not possible


JESCHO99:
### Terraform CLI Version

1.9.1

### Terraform Provider Version

v0.95.0

### Terraform Configuration

```terraform
terraform {
  required_version = "~> 1.9.0"
  required_providers {
    snowflake = {
      source  = "Snowflake-Labs/snowflake"
      version = "~> 0.95.0"
    }
  }
}

provider "snowflake" {
  role = "SYSADMIN"
}

# DATABASE
resource "snowflake_database" "this" {
  name = "TEST_DATABASE"
}

# SCHEMA
resource "snowflake_schema" "this" {
  database                    = snowflake_database.this.name
  name                        = "TEST_SCHEMA"
  is_transient                = "false"
  #with_managed_access         = "false"
  data_retention_time_in_days  = 1
}

# TABLES
resource "snowflake_table" "table_1" {
  database                    = snowflake_database.this.name
  schema                      = snowflake_schema.this.name
  name                        = "TEST_TABLE_V1"
  data_retention_time_in_days = 1
  change_tracking             = false

  column {
    name = "ID"
    type = "NUMBER(11,2)"
    nullable = false
  }

  column {
    name = "NAME"
    type = "VARCHAR(30)"
    nullable = true
  }
}

resource "snowflake_view" "this" {

  database = snowflake_database.this.name
  schema   = snowflake_schema.this.name
  name     = "TEST_VIEW"

  statement = <<-SQL
    SELECT ID, NAME FROM TEST_DATABASE.TEST_SCHEMA.TEST_TABLE_V1;
  SQL

  is_secure  = false

  depends_on = [
    snowflake_database.this,
    snowflake_schema.this,
    snowflake_table.table_1
  ]
}
```

### Category

category:resource

### Object type(s)

data_source:views

### Expected Behavior

The following terraform configurations should be able to execute a terraform apply without any error and after a successful apply of this code there should not be any open changes left when a new terraform plan is executed after the successful apply

### Actual Behavior

After a successful apply when a terraform plan is executed there is always a change for the view object for the column list. Even is the column list is not touched in the code this behaviour leads to an update/replacement of all views in every terraform plan executed.

![image](https://github.com/user-attachments/assets/a8db4159-baec-4056-acf1-c0559e7b43a9)
[...]

Issue #2994, opened 2024-08-13
[Feature]: Improve documentation for snowflake_grant_privileges_to_account_role


bdumford-OM1:
### Use Cases or Problem Statement

This is not a huge deal, but could the documentation for [snowflake_grant_privileges_to_account_role](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_privileges_to_account_role) be improved?  Specifically, there needs to be better documentation for how to migrate from the deprecated [snowflake_procedure_grant](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/0.92.0/docs/resources/procedure_grant) resource to `snowflake_grant_privileges_to_account_role`.

When migrating from the `snowflake_procedure_grant` resource, it was not at all obvious as to where to put the [argument_data_types](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/0.92.0/docs/resources/procedure_grant#argument_data_types).  After a lot of searching, I found the answer in issue #2375 and on the [CREATING_ISSUES](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/CREATING_ISSUES.md#granting-on-functions-or-procedures) page.  Since there is a solution available, it would make sense to document it in the appropriate places.

### Category

category:other

### Proposal

An example for migrating from the `snowflake_procedure_grant` should be added to the [MIGRATION_GUIDE](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md).  Additionally, an example like the one [here](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2375#issuecomment-1915553094) should be added to the Terraform registry [page](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_privileges_to_account_role) for `snowflake_grant_privileges_to_account_role`.

### How much impact is this issue causing?

Low

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
Hi @bdumford-OM1 👋 Thanks for the feedback. These things can be added to our docs. Currently we're reworking essential objects on our road to v1. We welcome community contributions based on our [contribution guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/CONTRIBUTING.md).

[...]

PR #3079, opened 2024-09-16
fix: Fix views permadiff

sfc-gh-jmichalak:
- fix permadiff when columns are not specified
- adjust row access policy signature in describe_output and move parsing to sdk
- adjust the identifier doc - add procedure id 

## Test Plan

* [x] acceptance tests

* [ ] …

## References

https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/3073
#2994


pkg/resources/view_acceptance_test.go
```diff
@@ -865,6 +864,51 @@ func TestAcc_View_Rename(t *testing.T) {
...
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_View/basic"),
+				ConfigVariables: tfconfig.ConfigVariablesFromModel(t, viewModelWithColumns),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("snowflake_view.test", plancheck.ResourceActionNoop),
+					},
+				},
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_view.test", "name", id.Name()),
+				),
```
sfc-gh-jcieslak, line 906: Could you add one step showing that going back to empty columns should also result in noop ?

sfc-gh-jmichalak: Done.

pkg/resources/view.go
```diff
@@ -194,7 +194,8 @@ var viewSchema = map[string]*schema.Schema{
 				},
 			},
 		},
-		Description: "If you want to change the name of a column or add a comment to a column in the new view, include a column list that specifies the column names and (if needed) comments about the columns. (You do not need to specify the data types of the columns.)",
+		Description:      "If you want to change the name of a column or add a comment to a column in the new view, include a column list that specifies the column names and (if needed) comments about the columns. You do not need to specify the data types of the columns. If this field is not specified, columns are inferred from the `statement` field by Snowflake.",
+		DiffSuppressFunc: IgnoreNewEmptyList([]string{"column_name"}),
```
sfc-gh-jcieslak, line 198: nit: whenever a list is needed to be passed, imo, it's better to have `...string` instead of `[]string`. Not a big deal, so we can leave it as is. It's more of a preference thing.

sfc-gh-jmichalak: I agree, but it's very inconsistent across the codebase. But changed to variadic arg.

pkg/resources/diff_suppressions.go
```diff
@@ -221,3 +221,20 @@ func suppressIdentifierQuoting(_, oldValue, newValue string, _ *schema.ResourceD
...
+		if len(parts) < 2 {
+			log.Printf("[DEBUG] invalid resource key: %s", parts)
+			return false
+		}
+		// key is element count
+		if parts[1] == "#" && new == "0" {
+			return true
+		}
+		// key is one of the ignored subfields
+		return len(parts) > 2 && slices.Contains(subfields, parts[2]) && new == ""
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Well, it's the same condition in this case, but I changed it. Ignoring an empty list also means we need to ignore changes on _some_ subfields. Also, I rephrased the function doc.

sfc-gh-jcieslak (approved):

PR #3086, opened 2024-09-18
feat: Tasks v1 readiness - SDK part

sfc-gh-jcieslak:
## Changes
- Generated assertions for task
- Added a few missing fields in Create, Alter, and outputs
- Added CreateOrAlter support
- Adjusted unit and integration tests
- Adjusted validation generator a bit (It was panic-ing before for that case; it was never used before)
- Adjusted resource and datasource to use new SDK

## References
* [CREATE TASK](https://docs.snowflake.com/en/sql-reference/sql/create-task)


pkg/sdk/tasks_def.go
```diff
@@ -1,52 +1,130 @@
...
+	case TaskStateStarted, TaskStateSuspended:
+		return taskState, nil
+	default:
+		return "", fmt.Errorf("unknown task state: %s", s)
+	}
+}
+
+type TaskRelationsRepresentation struct {
+	Predecessors  []string `json:"Predecessors"`
+	FinalizerTask string   `json:"FinalizerTask"`
```
sfc-gh-jmichalak, line 31: Nit: If this JSON uses upper case names, these tags can be dropped.

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -12,324 +14,328 @@ import (
...
+		assertTask(t, task, id, "")
 	})
 
 	t.Run("create task: with initial warehouse", func(t *testing.T) {
-		request := createTaskBasicRequest(t).
-			WithWarehouse(sdk.NewCreateTaskWarehouseRequest().WithUserTaskManagedInitialWarehouseSize(sdk.Pointer(sdk.WarehouseSizeXSmall)))
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
 
-		task := createTaskWithRequest(t, request)
+		err := testClient(t).Tasks.Create(ctx, sdk.NewCreateTaskRequest(id, sql).WithWarehouse(*sdk.NewCreateTaskWarehouseRequest().WithUserTaskManagedInitialWarehouseSize(sdk.WarehouseSizeXSmall)))
```
sfc-gh-jmichalak, line 130: This warehouse size param is not checked.

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -12,324 +14,328 @@ import (
...
+			HasTaskRelations(sdk.TaskRelations{
+				Predecessors:  []sdk.SchemaObjectIdentifier{},
+				FinalizerTask: &finalizerId,
+			}),
+		)
+	})
 
-		require.Equal(t, []sdk.SchemaObjectIdentifier{rootId}, t1.Predecessors)
+	t.Run("create dag of tasks", func(t *testing.T) {
+		rootId := testClientHelper().Ids.RandomSchemaObjectIdentifier()
```
sfc-gh-jmichalak, line 209: Nit: it would be great to provide a simple visual representation of the graph.

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -338,120 +344,180 @@ func TestInt_Tasks(t *testing.T) {
...
+		err := client.Tasks.Clone(ctx, sdk.NewCloneTaskRequest(id, sourceTask.ID()))
 		require.NoError(t, err)
-		t.Cleanup(cleanupTaskProvider(id))
+		t.Cleanup(testClientHelper().Task.DropTaskFunc(t, id))
 
 		task, err := client.Tasks.ShowByID(ctx, id)
 		require.NoError(t, err)
 
-		assertTask(t, task, request.GetName())
+		assertTask(t, task, id, testClientHelper().Ids.WarehouseId().Name())
```
sfc-gh-jmichalak, line 358: It's not clear why we check warehouse id here. Shouldn't we just compare if some values are identical in these 2 tasks?

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -338,120 +344,180 @@ func TestInt_Tasks(t *testing.T) {
...
-		id := request.GetName()
-
-		err := client.Tasks.Create(ctx, request)
+	t.Run("create or alter: complete", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+		err := client.Tasks.CreateOrAlter(ctx, sdk.NewCreateOrAlterTaskRequest(id, sql).
+			WithWarehouse(*sdk.NewCreateTaskWarehouseRequest().WithWarehouse(testClientHelper().Ids.WarehouseId())).
+			WithSchedule("10 MINUTES").
+			WithConfig(`$${"output_dir": "/temp/test_directory/", "learning_rate": 0.1}$$`).
+			WithAllowOverlappingExecution(true).
```
sfc-gh-jmichalak, line 367: I've now noticed that ERROR_INTEGRATION is missing from multiple places.

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -338,120 +344,180 @@ func TestInt_Tasks(t *testing.T) {
...
+		err := client.Tasks.Drop(ctx, sdk.NewDropTaskRequest(NonExistingSchemaObjectIdentifier))
+		assert.ErrorIs(t, err, sdk.ErrObjectNotExistOrAuthorized)
+	})
+
+	t.Run("alter task: set value and unset value", func(t *testing.T) {
+		task, taskCleanup := testClientHelper().Task.Create(t)
+		t.Cleanup(taskCleanup)
+
+		err := client.Tasks.Alter(ctx, sdk.NewAlterTaskRequest(task.ID()).WithSet(*sdk.NewTaskSetRequest().
+			// TODO: Cannot set warehouse due to Snowflake error
```
sfc-gh-jmichalak, line 433: What about this todo? Do we have a ticket?

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -338,120 +344,180 @@ func TestInt_Tasks(t *testing.T) {
...
+			// WithWarehouse(testClientHelper().Ids.WarehouseId()).
+			WithSchedule("10 MINUTE").
+			WithConfig(`$${"output_dir": "/temp/test_directory/", "learning_rate": 0.1}$$`).
+			WithAllowOverlappingExecution(true).
+			WithUserTaskTimeoutMs(1000).
+			WithSuspendTaskAfterNumFailures(100).
```
sfc-gh-jmichalak, line 439: What about session parameters?

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -338,120 +344,180 @@ func TestInt_Tasks(t *testing.T) {
...
+			WithTaskAutoRetryAttempts(true).
+			WithUserTaskMinimumTriggerIntervalInSeconds(true),
+		))
+		require.NoError(t, err)
+
+		assertions.AssertThat(t, objectassert.Task(t, task.ID()).
+			HasSchedule("").
+			HasConfig("").
+			HasAllowOverlappingExecution(false).
+			HasComment(""),
```
sfc-gh-jmichalak, line 471: Some checks are missing, after SET too.

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -565,16 +653,18 @@ func TestInt_Tasks(t *testing.T) {
...
 		returnedTask, err := client.Tasks.Describe(ctx, task.ID())
 		require.NoError(t, err)
 
-		assertTask(t, returnedTask, task.ID())
+		assertTask(t, returnedTask, task.ID(), testClientHelper().Ids.WarehouseId().Name())
 	})
 
 	t.Run("execute task: default", func(t *testing.T) {
-		task := createTask(t)
+		task, taskCleanup := testClientHelper().Task.Create(t)
```
sfc-gh-jmichalak, line 666: What about execute retry last?

pkg/sdk/testint/tasks_gen_integration_test.go
```diff
@@ -610,36 +701,39 @@ func TestInt_Tasks(t *testing.T) {
 
 	t.Run("resume root tasks within a graph containing more than one root task", func(t *testing.T) {
```
sfc-gh-jmichalak, line 702: Nit: add in the description that it's supposed to fail.

sfc-gh-jmichalak (changes requested):

pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_ext.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		return errors.Join(errs...)
+	})
+	return w
+}
+
+func (t *TaskAssert) HasTaskRelations(expected sdk.TaskRelations) *TaskAssert {
+	t.AddAssertion(func(t *testing.T, o *sdk.Task) error {
+		t.Helper()
+		if slices.EqualFunc(o.TaskRelations.Predecessors, expected.Predecessors, func(id sdk.SchemaObjectIdentifier, id2 sdk.SchemaObjectIdentifier) bool {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Doesn't really matter, only we are about that expected ids being there.

pkg/sdk/tasks_gen_test.go
```diff
@@ -54,39 +54,117 @@ func TestTasks_Create(t *testing.T) {
...
+		opts.TaskAutoRetryAttempts = Int(10)
+		opts.Tag = []TagAssociation{{
+			Name:  tagId,
+			Value: "v1",
+		}}
+		opts.UserTaskMinimumTriggerIntervalInSeconds = Int(10)
+		opts.After = []SchemaObjectIdentifier{otherTaskId}
+		opts.When = String(`SYSTEM$STREAM_HAS_DATA('MYSTREAM')`)
+
+		assertOptsValidAndSQLEquals(t, opts, "CREATE OR REPLACE TASK %s WAREHOUSE = %s SCHEDULE = '10 MINUTE' CONFIG = $${\"output_dir\": \"/temp/test_directory/\", \"learning_rate\": 0.1}$$ ALLOW_OVERLAPPING_EXECUTION = true JSON_INDENT = 10, LOCK_TIMEOUT = 5 USER_TASK_TIMEOUT_MS = 5 SUSPEND_TASK_AFTER_NUM_FAILURES = 6 ERROR_INTEGRATION = some_error_integration COMMENT = 'some comment' FINALIZE = %s TASK_AUTO_RETRY_ATTEMPTS = 10 TAG (%s = 'v1') USER_TASK_MINIMUM_TRIGGER_INTERVAL_IN_SECONDS = 10 AFTER %s WHEN SYSTEM$STREAM_HAS_DATA('MYSTREAM') AS SELECT CURRENT_TIMESTAMP", id.FullyQualifiedName(), warehouseId.FullyQualifiedName(), finalizerId.FullyQualifiedName(), tagId.FullyQualifiedName(), otherTaskId.FullyQualifiedName())
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: It's part of the session parameters. It doesn't really matter if it's there or not (I think, maybe I put it there on purpose, but I don't remember now). Ok, now I remember. It has to be there for ALTER UNSET, so to make it the same in other operations I left it this way.

pkg/sdk/tasks_def.go
sfc-gh-jcieslak: Yeah, I know, but I guess this makes it more clear what the struct is used for and it was just two additional tags.

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: Because it can't be checked. I left this test because it showed that we can create tasks with this value set, but the value, sadly, cannot be checked.

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: True, added comment above.

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: Hmm true, refactored the test case a bit

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: True, added integration to some of the tests and changed SDK a bit to make it clear it's notification integration (there's no such thing as error integration).

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: Update, it's in session parameters, added a check

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: I'll add task ticket number for now. I'll add it to our list, it should be a ticket pointing to Snowflake Task fix that is not yet created.

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: Added todo for now, I forgot about them and they turned out to be pretty big, I'll tackle them fully in the next pr.

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: True, didn't notice, added

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jcieslak: Added

pkg/sdk/testint/tasks_gen_integration_test.go
sfc-gh-jmichalak: OK. In the docs we have ERROR_INTEGRATION, so maybe it's worth reporting on docs discuss channel.

sfc-gh-jmichalak (dismissed):

pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_ext.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: Yes, but they can be returned in a different order. Please use ElementsMatch.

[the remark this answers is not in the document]
sfc-gh-jcieslak: Not sure how testify package works with our package, so I rewrote checks to be similar to the assert above (it asserts predecessors, but in the Predecessors field).

sfc-gh-jmichalak (approved):

pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_ext.go
```diff
@@ -0,0 +1,74 @@
...
+		t.Helper()
+		if o.Id == "" {
+			return fmt.Errorf("expected id not empty; got: %v", o.CreatedOn)
+		}
+		return nil
+	})
+	return t
+}
+
+func (t *TaskAssert) HasPredecessors(ids ...sdk.SchemaObjectIdentifier) *TaskAssert {
```
sfc-gh-asawicki, line 35: nit pick: in such assertions (when we require these two lists to match exactly without taking order into account - but this does not matter) it is useful to not only list the missing ids but also list the unexpected ones (and this is how it is usually implemented in the assertion libraries)

pkg/resources/task.go
```diff
@@ -182,7 +182,7 @@ func ReadTask(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	if err := d.Set("enabled", task.IsStarted()); err != nil {
+	if err := d.Set("enabled", task.State == sdk.TaskStateStarted); err != nil {
```
sfc-gh-asawicki, line 185: nit: why we get rid of the helper func from SDK?

sfc-gh-asawicki, line 185: do we want to set enabled if this is not set in the config? (maybe it is already a topic for the next change but had to ask; we are going away from the defaults for such attributes; it's also similar in a way to the warehouse dilemma we had for the state that can be changes externally - let's at least discuss it)

pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_ext.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done, same for HasPredecessors

pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_ext.go
sfc-gh-jcieslak: Changed the assertions, and added InAnyOrder suffix to this function (on failure all items are printed, so It will be known what was an unexpected item).

pkg/resources/task.go
sfc-gh-jcieslak: [Brought back in](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3170)

sfc-gh-jcieslak: Don't remember where, but I wrote somewhere that it's very hard or impossible to act on the default `enabled` value (not set in the config). That's why I made it required, so certain operations are easier because we know what state are we aiming with. This was truly painful during the update where suspend/resume logic is messy because of the Snowflake limitations. Maybe it will be possible to turn it into a tri-valued bool, but for now, for simplicity's sake, I would stay with the required regular bool.

PR #3109, opened 2024-10-01
feat: Stream on table resource

sfc-gh-jmichalak:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/acceptance/bettertestspoc/assert/objectassert/stream_snowflake_ext.go
```diff
@@ -38,3 +38,50 @@ func (s *StreamAssert) HasStageName(expected string) *StreamAssert {
...
+	s.AddAssertion(func(t *testing.T, o *sdk.Stream) error {
+		t.Helper()
+		if o.BaseTables == nil {
+			return fmt.Errorf("expected base tables to have value; got: nil")
+		}
+		if len(o.BaseTables) != len(expected) {
+			return fmt.Errorf("expected base tables length: %v; got: %v", len(expected), len(o.BaseTables))
+		}
+		for i := range o.BaseTables {
+			if o.BaseTables[i].FullyQualifiedName() != expected[i].FullyQualifiedName() {
```
sfc-gh-jcieslak, line 66: Shouldn't it be contains instead of relying on order?

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/stream_show_output_ext.go
```diff
@@ -0,0 +1,29 @@
...
+
+func (s *StreamShowOutputAssert) HasStaleAfterNotEmpty() *StreamShowOutputAssert {
+	s.AddAssertion(assert.ResourceShowOutputValuePresent("stale_after"))
+	return s
+}
+
+func (s *StreamShowOutputAssert) HasBaseTables(ids []sdk.SchemaObjectIdentifier) *StreamShowOutputAssert {
+	s.AddAssertion(assert.ResourceShowOutputValueSet("base_tables.#", strconv.FormatInt(int64(len(ids)), 10)))
+	for i := range ids {
+		s.AddAssertion(assert.ResourceShowOutputValueSet(fmt.Sprintf("base_tables.%d", i), ids[i].FullyQualifiedName()))
```
sfc-gh-jcieslak, line 26: Imo we should add a new assert that would assert length, then items, and it should be reliant on order of saved items. I created a ticket for it and added an item to the list of assert improvements.

pkg/resources/stream_on_table.go
```diff
@@ -0,0 +1,307 @@
...
+	if err := d.Set("name", id.Name()); err != nil {
+		return nil, err
+	}
+	if err := d.Set("database", id.DatabaseName()); err != nil {
+		return nil, err
+	}
+	if err := d.Set("schema", id.SchemaName()); err != nil {
+		return nil, err
+	}
+	if err := d.Set("append_only", booleanStringFromBool(*v.Mode == sdk.StreamModeAppendOnly)); err != nil {
```
sfc-gh-jcieslak, line 142: nit: It's repeated a few times and could be `v.IsAppendOnly()`

pkg/resources/stream_on_table.go
```diff
@@ -0,0 +1,307 @@
...
+			return diag.FromErr(err)
+		}
+
+		req := sdk.NewCreateOnTableStreamRequest(id, tableId)
+		if orReplace {
+			req.WithOrReplace(true)
+		}
+
+		if v := d.Get("copy_grants"); v.(bool) {
+			req.WithCopyGrants(true).WithOrReplace(true)
```
sfc-gh-jcieslak, line 168: is OrReplace needed here? It allows for possible overriding of already existing stream, without it the task will use OrReplace only when modified attribute cannot be altered.

pkg/resources/stream_on_table_acceptance_test.go
```diff
@@ -0,0 +1,564 @@
...
+	r "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	pluginconfig "github.com/hashicorp/terraform-plugin-testing/config"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_StreamOnTable_Basic(t *testing.T) {
+	acc.TestAccPreCheck(t)
```
sfc-gh-jcieslak, line 27: Add checks for EnableAcceptance env variable

pkg/resources/stream_on_table_acceptance_test.go
```diff
@@ -0,0 +1,564 @@
...
+						HasSchemaString(id.SchemaName()).
+						HasFullyQualifiedNameString(id.FullyQualifiedName()).
+						HasAppendOnlyString(r.BooleanFalse).
+						HasTableString(table.ID().FullyQualifiedName()),
+				),
+			},
+			// set all fields
+			{
+				ConfigDirectory: acc.ConfigurationDirectory("TestAcc_StreamOnTable/at"),
+				ConfigVariables: tfconfig.ConfigVariablesFromModel(t, model.WithCopyGrants(true).WithComment("foo").WithAppendOnly(r.BooleanTrue).WithShowInitialRows(r.BooleanTrue).WithAtValue(pluginconfig.MapVariable(map[string]pluginconfig.Variable{
```
sfc-gh-jcieslak, line 103: hmm, that's my opinion, but I think it would be better to have all the models at the top. Yes, you have to scroll, but it's easier to read and compare them when they're next to each other.

pkg/resources/stream_on_table_acceptance_test.go
```diff
@@ -0,0 +1,564 @@
...
+						}
+						return nil
+					})),
+				),
+			},
+		},
+	})
+}
+
+// There is no way to check at/before fields in show and describe. That's why in we try creating with these values, but do not assert them.
```
sfc-gh-jcieslak, line 321: There's something missing in the second sentence? 😅

pkg/resources/stream_on_table_acceptance_test.go
```diff
@@ -0,0 +1,564 @@
...
+	resourceName := "snowflake_stream_on_table.test"
+
+	table, cleanupTable := acc.TestClient().Table.CreateWithChangeTracking(t)
+	t.Cleanup(cleanupTable)
+	acc.TestClient().Table.InsertInt(t, table.ID())
+
+	lastQueryId := acc.TestClient().Context.LastQueryId(t)
+
+	model := model.StreamOnTable("test", id.DatabaseName(), id.Name(), id.SchemaName(), table.ID().FullyQualifiedName()).
+		WithComment("foo").WithAppendOnly(r.BooleanTrue).WithShowInitialRows(r.BooleanTrue).WithCopyGrants(true)
```
sfc-gh-jcieslak, line 335: Just a stylistic comment, I think it's more readable when every WithX has its own line

pkg/resources/testdata/TestAcc_StreamOnTable/at/test.tf
```diff
@@ -0,0 +1,19 @@
...
+  database = var.database
+  schema   = var.schema
+
+  copy_grants       = true
+  table             = var.table
+  append_only       = var.append_only
+  show_initial_rows = "true"
+
+  at {
+    timestamp = try(var.at["timestamp"], null)
```
sfc-gh-jcieslak, line 12: Uuu noice!

pkg/schemas/common_types.go
```diff
@@ -27,3 +27,75 @@ var FullyQualifiedNameSchema = &schema.Schema{
...
+			},
+			"statement": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Specifies the query ID of a statement to use as the reference point for Time Travel. This parameter supports any statement of one of the following types: DML (e.g. INSERT, UPDATE, DELETE), TCL (BEGIN, COMMIT transaction), SELECT.",
+				ExactlyOneOf: []string{"at.0.timestamp", "at.0.offset", "at.0.statement", "at.0.stream"},
+			},
+			"stream": {
+				Type:         schema.TypeString,
+				Optional:     true,
```
sfc-gh-jcieslak, line 58: Can identifier validation and diff suppress on quotes be added here or it was intentionally not added ?

pkg/schemas/stream.go
```diff
@@ -0,0 +1,132 @@
...
+			streamSchema["table_name"] = tableId.FullyQualifiedName()
+		}
+	}
+	if stream.SourceType != nil {
+		streamSchema["source_type"] = stream.SourceType
+	}
+	if stream.BaseTables != nil {
+		idsFQN := make([]string, len(stream.BaseTables))
+		for i := range idsFQN {
+			idsFQN[i] = stream.BaseTables[i].FullyQualifiedName()
```
sfc-gh-jcieslak, line 107: nit: you could use collections.Map here

pkg/schemas/stream_gen.go
```diff
@@ -3,11 +3,14 @@
...
 
 import (
+	"log"
+
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 // ShowStreamSchema represents output of SHOW query for the single Stream.
+// TODO: merge with desc schema
```
sfc-gh-jcieslak, line 13: next pr ?

pkg/sdk/parsers.go
```diff
@@ -37,3 +37,18 @@ func ParseCommaSeparatedStringArray(value string, trimQuotes bool) []string {
 	}
 	return trimmedListItems
 }
+
+// ParseCommaSeparatedSchemaObjectIdentifierArray can be used to parse Snowflake output containing a list of schema-level object identifiers
+// in the format of ["db".SCHEMA."name", "db"."schema2"."name2", ...],
+func ParseCommaSeparatedSchemaObjectIdentifierArray(value string) ([]SchemaObjectIdentifier, error) {
```
sfc-gh-jcieslak, line 43: nit: could be written in a more generic way for us to extend the functionality for other id types

pkg/sdk/parsers_test.go
```diff
@@ -104,3 +105,96 @@ func TestParseCommaSeparatedStringArray(t *testing.T) {
...
+		Value string
+		Error string
+	}{
+		{
+			Name:  "bad quotes",
+			Value: `["a]`,
+			Error: "unable to read identifier: \"a, err = parse error on line 1, column 3: extraneous or missing \" in quoted-field",
+		},
+		{
+			Name:  "missing parts",
```
sfc-gh-jcieslak, line 183: you can also add too much parts case

pkg/sdk/streams_def.go
```diff
@@ -1,9 +1,54 @@
...
+
+const (
+	StreamSourceTypeTable         StreamSourceType = "Table"
+	StreamSourceTypeExternalTable StreamSourceType = "External Table"
+	StreamSourceTypeView          StreamSourceType = "View"
+	StreamSourceTypeStage         StreamSourceType = "Stage"
+)
+
+func ToStreamSourceType(s string) (StreamSourceType, error) {
+	switch streamSourceType := StreamSourceType(s); streamSourceType {
```
sfc-gh-jcieslak, line 22: Do we want to leave the values this way or actually modify them a bit, so they're consistent with other constant values (e.g. TABLE, EXTERNAL TABLE, VIEW, STAGE)? Are uppercased values possible to use? 🤔 For example, StreamMode, a few lines below, has a more standard way of defining constants. Also is this approach of parsing is case sensitive or not in configuration? Because I think we should be also consistent with that and make it sense insensitive.

sfc-gh-jmichalak: Well, this function is not used to parse config, but Snowflake responses that are capitalized. That's why initially I made enum capitalized, but not upper case. But I'll change this to be consistent.

pkg/sdk/parsers.go
sfc-gh-jmichalak: It could, but it's not needed for now.

pkg/resources/stream_on_table_acceptance_test.go
sfc-gh-jmichalak: Yeah, as we discussed today, I'll move this to top.

pkg/resources/stream_on_table.go
sfc-gh-jmichalak: It's needed because otherwise when you have COPY GRANTS without OR REPLACE, Snowflake fails.

pkg/schemas/stream_gen.go
sfc-gh-jmichalak: Nope, I think we'll leave this in case of some changes on the Snowflake side. I don't think we have coupled structs anywhere.

pkg/acceptance/bettertestspoc/assert/objectassert/stream_snowflake_ext.go
sfc-gh-jmichalak: Done.

pkg/resources/stream_on_table.go
sfc-gh-jmichalak: Done.

pkg/resources/stream_on_table_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/stream_on_table_acceptance_test.go
sfc-gh-jmichalak: Fixed.

pkg/resources/stream_on_table_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/schemas/common_types.go
sfc-gh-jmichalak: Added.

pkg/schemas/stream.go
sfc-gh-jmichalak: Done.

pkg/sdk/parsers_test.go
sfc-gh-jmichalak: Done.

pkg/schemas/stream_gen.go
sfc-gh-asawicki: yeah, let;s not merge these structs

MIGRATION_GUIDE.md
```diff
@@ -4,6 +4,15 @@ This document is meant to help you migrate your Terraform config to the new newe
 describe deprecations or breaking changes and help you to change your configuration to keep the same (or similar) behavior
 across different versions.
 
+## v0.96.0 ➞ v0.97.0
+
+### *(new feature)* snowflake_stream_on_table resource
+
+To enhance clarity and functionality, the new resource `snowflake_stream_on_table` has been introduced to replace the previous `snowflake_stream`. Recognizing that the old resource carried multiple responsibilities within a single entity, we opted to divide it into more specialized resources.
```
sfc-gh-asawicki, line 11: are there any differences between old resource and new resource/resources (besides of course being dedicated to the given object)? If so, it would be nice to guide our user what they should change if they already have `snowflake_stream` and would like to make `snowflake_stream_on_table`  out of it (e.g. import syntax changes, right?).

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/stream_show_output_ext.go
sfc-gh-asawicki: let's add it to the readme?

pkg/resources/stream_on_table.go
sfc-gh-asawicki: You are right about the need for orReplace needed with copy grants but I am with @sfc-gh-jcieslak here: we should steer the copy grants logic, and use it only when we need to replace the stream and not by default in create

pkg/resources/stream_on_table.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+
+		if v := d.Get("append_only").(string); v != BooleanDefault {
```
sfc-gh-asawicki, line 171: Please check new `booleanStringAttributeCreate` helper func and others (check user resource); they allow writing shorter create/update/read.

pkg/schemas/stream_gen.go
```diff
@@ -86,13 +91,22 @@ func StreamToSchema(stream *sdk.Stream) map[string]any {
 		streamSchema["comment"] = stream.Comment
 	}
 	if stream.TableName != nil {
-		streamSchema["table_name"] = stream.TableName
+		tableId, err := sdk.ParseSchemaObjectIdentifier(*stream.TableName)
```
sfc-gh-asawicki, line 94: question: were this and the next added manually or is our generator already that smart? (if manually, please leave a comment so that we don't overwrite it

pkg/sdk/streams_def.go
```diff
@@ -1,9 +1,54 @@
...
+type StreamSourceType string
+
+const (
+	StreamSourceTypeTable         StreamSourceType = "TABLE"
+	StreamSourceTypeExternalTable StreamSourceType = "EXTERNAL TABLE"
+	StreamSourceTypeView          StreamSourceType = "VIEW"
+	StreamSourceTypeStage         StreamSourceType = "STAGE"
+)
+
+func ToStreamSourceType(s string) (StreamSourceType, error) {
```
sfc-gh-asawicki, line 21: nit: we usually test these funcs.

pkg/resources/stream_on_table_acceptance_test.go
```diff
@@ -0,0 +1,617 @@
...
+	_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)
+	acc.TestAccPreCheck(t)
+	id := acc.TestClient().Ids.RandomSchemaObjectIdentifier()
+	resourceId := helpers.EncodeResourceIdentifier(id)
+	resourceName := "snowflake_stream_on_table.test"
+
+	table, cleanupTable := acc.TestClient().Table.CreateWithChangeTracking(t)
+	t.Cleanup(cleanupTable)
+
+	baseModel := func() *model.StreamOnTableModel {
```
sfc-gh-asawicki, line 37: nit: this could be added as Base(id, table_id) func in `_ext.go` file for model

sfc-gh-asawicki (changes requested):

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/stream_show_output_ext.go
sfc-gh-jcieslak: I added it in my pr

MIGRATION_GUIDE.md
sfc-gh-jmichalak: It's really straightforward to migrate to the new resource—just change the resource to stream_on_table and move the table ID properly—so I've decided not to include this in the migration guide.

sfc-gh-asawicki: But even such a sentence is helpful if someone does not know what are the changes

pkg/resources/stream_on_table.go
sfc-gh-jmichalak: We don't use OR REPLACE by default. It's used when the stream is already in the state, and detect changes on fields that can't be updated, because ForceNew drops the grants. In the second case, when it's not in the state, and the user wants COPY GRANTS to preserve old grants.

pkg/resources/stream_on_table.go
sfc-gh-jmichalak: Done.

pkg/schemas/stream_gen.go
sfc-gh-jmichalak: It was changed manually, I added a comment. Note that we don't mark the manual changes after generation.

pkg/sdk/streams_def.go
sfc-gh-jmichalak: I don't think we usually do, but I added tests here.

pkg/resources/stream_on_table_acceptance_test.go
sfc-gh-jmichalak: Yes, but I'll leave this for now.

pkg/resources/stream_on_table.go
sfc-gh-asawicki: I disagree with the second case. IMO a new create should be done without copy grants and without or replace. Thge assumption should be, that if we are creating a new object, the copy grants do not matter. Copy grants should be used only in the first scenario you mentioned (so the object is already managed and we discover changes).

pkg/resources/stream_common.go
```diff
@@ -58,3 +58,79 @@ func DeleteStreamContext(ctx context.Context, d *schema.ResourceData, meta any)
...
+	ConflictsWith: []string{"before"},
+}
+
+var beforeSchema = &schema.Schema{
+	Type:        schema.TypeList,
+	Optional:    true,
+	MaxItems:    1,
+	Description: externalChangesNotDetectedFieldDescription("This field specifies that the request refers to a point immediately preceding the specified parameter. This point in time is just before the statement, identified by its query ID, is completed.  Due to Snowflake limitations, the provider does not detect external changes on this field."),
+	Elem: &schema.Resource{
+		Schema: map[string]*schema.Schema{
```
sfc-gh-jcieslak, line 106: Tbh I'm not sure if there are any differences, but you could go full recycling mode and define subfields as one variable and use it in both schemas, but this kind of duplication is not that bad.

pkg/resources/stream_on_table.go
```diff
@@ -162,26 +162,24 @@ func CreateStreamOnTable(orReplace bool) schema.CreateContextFunc {
 		req := sdk.NewCreateOnTableStreamRequest(id, tableId)
 		if orReplace {
 			req.WithOrReplace(true)
+		} else if d.Get("copy_grants").(bool) {
+			return diag.Diagnostics{
+				diag.Diagnostic{
+					Severity: diag.Error,
+					Summary:  "COPY GRANTS cannot be used without OR REPLACE. If you are creating this object, first import it, so that it's managed by the provider, and try again.",
```
sfc-gh-jcieslak, line 169: I guess it's already in other thread, but I think we could have one pass in Create that doesn't task copy_grants into account, but later modifications will be counted. Importing to use copy_grants is cumbersome (I know I can set it to false and then to true when it's in state, but not everybody will understand that).

pkg/schemas/stream.go
```diff
@@ -102,11 +103,9 @@ func StreamDescriptionToSchema(stream sdk.Stream) map[string]any {
 		streamSchema["source_type"] = stream.SourceType
 	}
 	if stream.BaseTables != nil {
-		idsFQN := make([]string, len(stream.BaseTables))
-		for i := range idsFQN {
-			idsFQN[i] = stream.BaseTables[i].FullyQualifiedName()
-		}
-		streamSchema["base_tables"] = idsFQN
+		streamSchema["base_tables"] = collections.Map(stream.BaseTables, func(id sdk.SchemaObjectIdentifier) string {
```
sfc-gh-jcieslak, line 106: func could be changed to `sdk.SchemaObjectIdentifeir.FullyQualifiedName` 😎

pkg/resources/stream_common.go
sfc-gh-jmichalak: Note that these fields have different ExactlyOneOf, so this needs to be parameterized. We'd only save a couple of lines, which is not much.

pkg/resources/stream_on_table.go
sfc-gh-jmichalak: You can also leave it empty, which I think most of the users will do. Besides, that's the default workflow here - when you have an existing object outside Terraform, you should import it first.

pkg/schemas/stream.go
sfc-gh-jmichalak: Done.

pkg/resources/stream_common.go
sfc-gh-jcieslak: Yeah I didn't notice 👍

pkg/resources/stream_on_table.go
sfc-gh-jcieslak: Yeah, but it's not intuitive, because this parameter blocks Create operation suggesting import when it could be just created without that parameter and then set to true (if you know how the resource works underneath). Even if the resource already exists and someone has copy_grants set to true he'll get error from Snowflake saying 'object already exists', then the user may consider importing or removing the object manually and creating it through Terraform (with either copy_grants turned on or off without any blocking). That's how I see it.

sfc-gh-jmichalak: OK, changed it, but it's still not intuitive to me.

pkg/resources/stream_on_table.go
sfc-gh-jmichalak: Changed again.

docs/resources/stream_on_table.md
```diff
@@ -63,7 +63,7 @@ resource "snowflake_stream_on_table" "stream" {
...
 - `before` (Block List, Max: 1) This field specifies that the request refers to a point immediately preceding the specified parameter. This point in time is just before the statement, identified by its query ID, is completed.  Due to Snowflake limitations, the provider does not detect external changes on this field. External changes for this field won't be detected. In case you want to apply external changes, you can re-create the resource manually using "terraform taint". (see [below for nested schema](#nestedblock--before))
 - `comment` (String) Specifies a comment for the stream.
-- `copy_grants` (Boolean) Retains the access permissions from the original stream when a new stream is created using the OR REPLACE clause. Use only if the resource is already managed by Terraform.
+- `copy_grants` (Boolean) Retains the access permissions from the original stream when a new stream is created using the OR REPLACE clause. Use only if the resource is already managed by Terraform. Otherwise, this field is skipped.
```
sfc-gh-jcieslak, line 66: I would change the description of this field to something similar to:
`Retains the access permissions from the original stream when a new stream is created using the OR REPLACE clause that is sometimes used when updating the stream. The value won't have any effect when creating a new stream.`

sfc-gh-jmichalak: Ok, I'll rephrase this in the next pr.

sfc-gh-jcieslak (approved):

pkg/sdk/streams_def.go
sfc-gh-asawicki: We do, if you know about some places when we don't, then we should also add them there.

pkg/resources/stream_on_table_acceptance_test.go
sfc-gh-asawicki: why? For consistency we should have a way of defining a default/common base for model.

sfc-gh-asawicki (approved):

pkg/schemas/stream_gen.go
sfc-gh-asawicki: Close after reading.

pkg/resources/stream_on_table_acceptance_test.go
sfc-gh-jmichalak: Ok, I'll do this in the next pr.

PR #3112, opened 2024-10-03
chore: Bump golang-ci lint to 1.61

sfc-gh-asawicki:
To be able to run successfully with go 1.23.2.


sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (approved):
I noticed that we also have a linter version in tools/go.mod. I suppose it's not used now, and can be removed?

PR #3113, opened 2024-10-03
feat: Task resource v1 readiness

sfc-gh-jcieslak:
## Changes
- Mostly done task resource refactor
- Sweepers test fixed (they always added errors even when they're nil)
- Minor changes in SDK (find out the bug with alter task and right now we're only supporting upper-cased warehouse names; added that to the field description)
- Acceptance tests
- The enabled field is not a three-valued boolean because it created certain issues when updating the task. The task status is very important for their management and working, so I decided to have them as regular boolean values that are either true or false, nothing in between.

## Next prs
- Apply comments from https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3086
- Data source
- Move some of the logic to SDK
- Refactor SDK suspend root logic for tasks (and overall suspend logic in SDK/resource)
- Examples, documentation, and migration guide
- More test cases for complicated DAG structures to show the resource can handle more complex structures
- Analyze non-deterministic test cases
- Refactor task resuming in task resource (most likely with the use of defer) because currently, there may be cases that error can cause tasks to be not resumed.
- Test and see how the DAG of tasks could be owned by a role other than the one created with Terraform (also with less privileges, only to run the task).
- Make config without $$ escapes needed
- Check if more helper functions from user resource could be used in task resource.
- Check in one acceptance test why finalizer task in show_output is not set (is that Snowflake or mapping error).
- More tests for calling (`as` field) - https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3113#discussion_r1789900590
- Re-generate and list all the issues with asserts and models
- check test tasks_gen_integration_test.go:937 (and see why it's non deterministic).
- Support session paramters

## References
* [CREATE TASK](https://docs.snowflake.com/en/sql-reference/sql/create-task)


pkg/sdk/tasks_impl_gen.go
```diff
@@ -332,6 +351,8 @@ func (r taskDBRow) convert() *Task {
 			}
 		}
 		task.Predecessors = ids
+	} else {
+		task.Predecessors = make([]SchemaObjectIdentifier, 0)
```
sfc-gh-jmichalak, line 355: Just curious, why can't we leave this as `nil`?

pkg/schemas/task_parameters.go
```diff
@@ -0,0 +1,91 @@
...
+		ShowTaskParametersSchema[strings.ToLower(string(param))] = ParameterListSchema
+	}
+}
+
+func TaskParametersToSchema(parameters []*sdk.Parameter) map[string]any {
+	taskParametersValue := make(map[string]any)
+	for _, param := range parameters {
+		if slices.Contains(taskParameters, sdk.TaskParameter(param.Key)) {
+			taskParametersValue[strings.ToLower(param.Key)] = []map[string]any{ParameterToSchema(param)}
+		}
```
sfc-gh-jmichalak, line 88: I think we should log, if the given param is not in the list of taskParameters.

pkg/schemas/task_gen.go
```diff
@@ -102,19 +129,39 @@ func TaskToSchema(task *sdk.Task) map[string]any {
...
+	}
 	taskSchema["last_committed_on"] = task.LastCommittedOn
 	taskSchema["last_suspended_on"] = task.LastSuspendedOn
 	taskSchema["owner_role_type"] = task.OwnerRoleType
 	taskSchema["config"] = task.Config
 	taskSchema["budget"] = task.Budget
+	taskSchema["last_suspended_reason"] = task.LastSuspendedReason
+	finalizer := ""
+	if task.TaskRelations.FinalizerTask != nil {
+		finalizer = task.TaskRelations.FinalizerTask.FullyQualifiedName()
```
sfc-gh-jmichalak, line 152: Hmm why is this handled in a different way than warehouse?

pkg/resources/user_parameters.go
```diff
@@ -77,11 +77,12 @@ var (
 )
 
 type parameterDef[T ~string] struct {
```
sfc-gh-jmichalak, line 79: It's not a part of this PR, but we should move parameterDef to a common file. It's also used in schemas and databases.

pkg/resources/task_parameters.go
```diff
@@ -0,0 +1,412 @@
...
+		{Name: sdk.TaskParameterUnsupportedDdlAction, Type: schema.TypeString, Description: "Determines if an unsupported (i.e. non-default) value specified for a constraint property returns an error."},
+		{Name: sdk.TaskParameterUseCachedResult, Type: schema.TypeBool, Description: "Specifies whether to reuse persisted query results, if available, when a matching query is submitted."},
+		{Name: sdk.TaskParameterWeekOfYearPolicy, Type: schema.TypeInt, Description: "Specifies how the weeks in a given year are computed. `0`: The semantics used are equivalent to the ISO semantics, in which a week belongs to a given year if at least 4 days of that week are in that year. `1`: January 1 is included in the first week of the year and December 31 is included in the last week of the year."},
+		{Name: sdk.TaskParameterWeekStart, Type: schema.TypeInt, Description: "Specifies the first day of the week (used by week-related date functions). `0`: Legacy Snowflake behavior is used (i.e. ISO-like semantics). `1` (Monday) to `7` (Sunday): All the week-related functions use weeks that start on the specified day of the week."},
```
sfc-gh-jmichalak, line 144: No validator? In other params as well.

pkg/resources/task.go
```diff
@@ -2,709 +2,648 @@ package resources
...
+		Description:      blocklistedCharactersFieldDescription("The schema in which to create the task."),
+	},
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		DiffSuppressFunc: suppressIdentifierQuoting,
+		Description:      blocklistedCharactersFieldDescription("Specifies the identifier for the task; must be unique for the database and schema in which the task is created."),
+	},
+	"enabled": {
```
sfc-gh-jmichalak, line 47: I think we could call it `started` or `running`. I can see these phrases in the docs, instead of `enabled`.

pkg/resources/task.go
```diff
@@ -2,709 +2,648 @@ package resources
...
 	},
 	"schedule": {
-		Type:          schema.TypeString,
-		Optional:      true,
-		Description:   "The schedule for periodically running the task. This can be a cron or interval in minutes. (Conflict with after)",
-		ConflictsWith: []string{"after"},
+		Type:             schema.TypeString,
+		Optional:         true,
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("schedule"),
+		Description:      "The schedule for periodically running the task. This can be a cron or interval in minutes. (Conflicts with finalize and after)",
```
sfc-gh-jmichalak, line 71: Did you consider splitting cron and interval? I remember that in views we have it separated, and I wonder if we could be consistent here.

pkg/resources/task.go
```diff
@@ -2,709 +2,648 @@ package resources
...
+		ConflictsWith: []string{"schedule", "finalize"},
 	},
 	"when": {
-		Type:        schema.TypeString,
-		Optional:    true,
-		Description: "Specifies a Boolean SQL expression; multiple conditions joined with AND/OR are supported.",
+		Type:             schema.TypeString,
+		Optional:         true,
+		DiffSuppressFunc: SuppressIfAny(DiffSuppressStatement, IgnoreChangeToCurrentSnowflakeValueInShow("condition")),
+		Description:      "Specifies a Boolean SQL expression; multiple conditions joined with AND/OR are supported.",
```
sfc-gh-jmichalak, line 132: I think this description is incomplete: It's not clear what `when` it's used for.

pkg/resources/task.go
```diff
@@ -2,709 +2,648 @@ package resources
...
-				tasksToResume, err := client.Tasks.SuspendRootTasks(ctx, depId, taskId)
-				defer func() {
-					if err := client.Tasks.ResumeTasks(ctx, tasksToResume); err != nil {
-						log.Printf("[WARN] failed to resume tasks: %s", err)
-					}
-				}()
-				if err != nil {
-					return err
+		} else {
+			if task.TaskRelations.FinalizedRootTask == nil {
```
sfc-gh-jmichalak, line 423: I'm not sure if we should fail here. A user declaratively says in the config that the finalizer must be unset, and I think it's not important that it was unset before.

pkg/resources/task.go
```diff
@@ -2,709 +2,648 @@ package resources
...
+	}
 
-	taskId := helpers.DecodeSnowflakeID(d.Id()).(sdk.SchemaObjectIdentifier)
+	task, err := client.Tasks.ShowByID(ctx, id)
+	if err != nil {
+		return diag.FromErr(err)
+	}
 
-	tasksToResume, err := client.Tasks.SuspendRootTasks(ctx, taskId, taskId)
-	defer func() {
```
sfc-gh-jmichalak, line 420: Why is this defer removed? In case of an error, we leave the previously suspended tasks as not resumed.

pkg/resources/task_acceptance_test.go
```diff
@@ -1,564 +1,970 @@
...
-    user_task_managed_initial_warehouse_size = "SMALL"
-}
-`
-	return fmt.Sprintf(s, name, databaseName, schemaName)
+	rootTaskModel := model.TaskWithId("main_task", rootTaskId, true, statement).
+		WithSchedule(schedule).
+		WithSuspendTaskAfterNumFailures(2)
+	childTaskModel := model.TaskWithId("second_task", finalizerTaskId, true, statement).
+		WithFinalize(rootTaskId.FullyQualifiedName())
+	childTaskModel.SetDependsOn([]string{rootTaskModel.ResourceReference()})
```
sfc-gh-jmichalak, line 868: I don't think DependsOn is needed here because we reference rootTask by finalizer field.

pkg/resources/task_acceptance_test.go
```diff
@@ -1,564 +1,970 @@
...
+			{
+				Config: config.FromModel(t, rootTaskModel) + config.FromModel(t, childTaskModel),
+				Check: assert.AssertThat(t,
+					resourceassert.TaskResource(t, rootTaskModel.ResourceReference()).
+						HasScheduleString(schedule).
+						HasEnabledString(r.BooleanTrue).
+						HasSuspendTaskAfterNumFailuresString("2"),
+					resourceshowoutputassert.TaskShowOutput(t, rootTaskModel.ResourceReference()).
+						HasSchedule(schedule).
+						// HasTaskRelations(sdk.TaskRelations{FinalizerTask: &finalizerTaskId}).
```
sfc-gh-jmichalak, line 912: What about these comments?

pkg/resources/task_acceptance_test.go
```diff
@@ -1,564 +1,970 @@
...
+				),
+			},
+			// Change tasks in DAG to standalone tasks (disabled to check if resuming/suspending works correctly)
+			{
+				Config: config.FromModel(t, firstTaskStandaloneModelDisabled) + config.FromModel(t, secondTaskStandaloneModelDisabled),
+				Check: assert.AssertThat(t,
+					resourceassert.TaskResource(t, firstTaskStandaloneModelDisabled.ResourceReference()).
+						HasScheduleString(schedule).
+						HasEnabledString(r.BooleanFalse).
+						HasSuspendTaskAfterNumFailuresString("10"),
```
sfc-gh-jmichalak, line 928: We should also check that finalized/finalizer fields are empty.

pkg/resources/task_acceptance_test.go
```diff
@@ -1,564 +1,970 @@
...
+	acc.TestAccPreCheck(t)
+
+	currentRole := acc.TestClient().Context.CurrentRole(t)
+
+	id := acc.TestClient().Ids.RandomSchemaObjectIdentifier()
+	statement := "SELECT 1"
+	basicConfigModel := model.TaskWithId("test", id, false, statement)
+
+	// New warehouse created, because the common one has lower-case letters that won't work
+	warehouse, warehouseCleanup := acc.TestClient().Warehouse.CreateWarehouse(t)
```
sfc-gh-jmichalak, line 238: We already have SNOW-1541938 for streamlits, we should create a similar ticket for this.

pkg/resources/task_acceptance_test.go
```diff
@@ -1,564 +1,970 @@
...
-			Enabled:           true,
-			Schedule:          "15 MINUTE",
-			UserTaskTimeoutMs: 900000,
+	taskConfig := `$${"output_dir": "/temp/test_directory/", "learning_rate": 0.1}$$`
+	// We have to do three $ at the beginning because Terraform will remove one $.
+	// It's because `${` is a special pattern, and it's escaped by `$${`.
+	expectedTaskConfig := strings.ReplaceAll(taskConfig, "$", "")
+	taskConfigVariableValue := "$" + taskConfig
+	comment := random.Comment()
+	condition := `SYSTEM$STREAM_HAS_DATA('MYSTREAM')`
```
sfc-gh-jmichalak, line 250: I think we should have some test cases for different spaces and new lines for fields with SQL statements.

pkg/resources/task_acceptance_test.go
```diff
@@ -1,564 +1,970 @@
...
+						HasCondition(condition).
+						HasAllowOverlappingExecution(true).
+						HasErrorIntegration(errorNotificationIntegration.ID().Name()).
+						HasLastCommittedOnNotEmpty().
+						HasLastSuspendedOn("").
+						HasOwnerRoleType("ROLE").
+						HasConfig(expectedTaskConfig).
+						HasBudget("").
+						HasTaskRelations(sdk.TaskRelations{}),
+				),
```
sfc-gh-jmichalak, line 353: Pls add a test case with external changes.

pkg/resources/user_parameters.go
sfc-gh-asawicki: We already have SNOW-1645342

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/task_resource_parameters_ext.go
```diff
@@ -0,0 +1,69 @@
...
+		HasTimestampTypeMapping(sdk.TimestampTypeMappingNtz).
+		HasTimestampTzOutputFormat("").
+		HasTimezone("America/Los_Angeles").
+		HasTimeInputFormat("AUTO").
+		HasTimeOutputFormat("HH24:MI:SS").
+		HasTraceLevel(sdk.TraceLevelOff).
+		HasTransactionAbortOnError(false).
+		HasTransactionDefaultIsolationLevel(sdk.TransactionDefaultIsolationLevelReadCommitted).
+		HasTwoDigitCenturyStart(1970).
+		HasUnsupportedDdlAction(sdk.UnsupportedDDLAction(strings.ToLower(string(sdk.UnsupportedDDLActionIgnore)))).
```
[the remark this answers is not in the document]
sfc-gh-asawicki: probably because the same reason as here: https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/560ab6b444ac461684b0d25c68bcb2d55201aba8/pkg/acceptance/bettertestspoc/assert/objectparametersassert/gen/object_parameters_def.go#L89 (Snowflake lowercase default for one value and uppercase for other)

pkg/resources/task.go
sfc-gh-asawicki: +1

pkg/resources/task.go
sfc-gh-asawicki: +1

pkg/resources/task_acceptance_test.go
sfc-gh-asawicki: +1

pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_gen.go
```diff
@@ -107,11 +107,17 @@ func (t *TaskAssert) HasComment(expected string) *TaskAssert {
 	return t
 }
 
-func (t *TaskAssert) HasWarehouse(expected string) *TaskAssert {
+func (t *TaskAssert) HasWarehouse(expected *sdk.AccountObjectIdentifier) *TaskAssert {
```
sfc-gh-asawicki, line 110: We should not modify the generated assertions: if you need a better one (because this one is bad for some reason), then please document it and for now add other assertion in the _ext/go file/

pkg/acceptance/bettertestspoc/assert/objectparametersassert/task_parameters_snowflake_gen.go
```diff
@@ -1027,7 +1027,7 @@ func (t *TaskParametersAssert) HasDefaultTaskAutoRetryAttemptsValueExplicit() *T
 }
 
 func (t *TaskParametersAssert) HasDefaultUserTaskManagedInitialWarehouseSizeValueExplicit() *TaskParametersAssert {
-	return t.HasUserTaskManagedInitialWarehouseSize("Medium")
+	return t.HasUserTaskManagedInitialWarehouseSize(sdk.WarehouseSize("Medium"))
```
sfc-gh-asawicki, line 1030: is this change generated or changed manually?

pkg/acceptance/bettertestspoc/assert/resourceassert/task_resource_ext.go
```diff
@@ -0,0 +1,17 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"strconv"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+)
+
+func (t *TaskResourceAssert) HasAfterIds(ids ...sdk.SchemaObjectIdentifier) *TaskResourceAssert {
+	t.AddAssertion(assert.ValueSet("after.#", strconv.FormatInt(int64(len(ids)), 10)))
+	for i, id := range ids {
+		t.AddAssertion(assert.ValueSet(fmt.Sprintf("after.%d", i), id.FullyQualifiedName()))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: SHouldn't because the assertion does not have InOrder in name or because Snowflake ordering for alter in non-deterministic?

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/task_resource_parameters_ext.go
```diff
@@ -0,0 +1,69 @@
...
+	"strings"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func (u *TaskResourceParametersAssert) HasAllDefaults() *TaskResourceParametersAssert {
+	return u.
+		HasSuspendTaskAfterNumFailures(10).
+		HasTaskAutoRetryAttempts(0).
+		HasUserTaskManagedInitialWarehouseSize("Medium").
```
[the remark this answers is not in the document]
sfc-gh-asawicki: +1

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/task_resource_parameters_ext.go
```diff
@@ -0,0 +1,69 @@
...
+		HasClientMemoryLimit(1536).
+		HasClientMetadataRequestUseConnectionCtx(false).
+		HasClientPrefetchThreads(4).
+		HasClientResultChunkSize(160).
+		HasClientResultColumnCaseInsensitive(false).
+		HasClientSessionKeepAlive(false).
+		HasClientSessionKeepAliveHeartbeatFrequency(3600).
+		HasClientTimestampTypeMapping(sdk.ClientTimestampTypeMappingLtz).
+		HasDateInputFormat("AUTO").
+		HasDateOutputFormat("YYYY-MM-DD").
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Description for date output format form snowflake docs: "Any valid, supported date format", so we cant do enum here. The parameters used for task are similar to the ones used in user, and for users I did enums for all the fields that it made sense for me.

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/task_show_output_ext.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (t *TaskShowOutputAssert) HasLastSuspendedOnNotEmpty() *TaskShowOutputAssert {
+	t.AddAssertion(assert.ResourceShowOutputValuePresent("last_suspended_on"))
+	return t
+}
+
+func (t *TaskShowOutputAssert) HasPredecessors(predecessors ...sdk.SchemaObjectIdentifier) *TaskShowOutputAssert {
+	t.AddAssertion(assert.ResourceShowOutputValueSet("predecessors.#", strconv.Itoa(len(predecessors))))
+	for i, predecessor := range predecessors {
+		t.AddAssertion(assert.ResourceShowOutputValueSet(fmt.Sprintf("predecessors.%d", i), predecessor.FullyQualifiedName()))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: as above

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/task_show_output_gen.go
```diff
@@ -0,0 +1,158 @@
...
+	t.AddAssertion(assert.ResourceShowOutputValueSet("config", expected))
+	return t
+}
+
+func (t *TaskShowOutputAssert) HasBudget(expected string) *TaskShowOutputAssert {
+	t.AddAssertion(assert.ResourceShowOutputValueSet("budget", expected))
+	return t
+}
+
+func (t *TaskShowOutputAssert) HasTaskRelations(expected sdk.TaskRelations) *TaskShowOutputAssert {
```
sfc-gh-asawicki, line 140: is this method generated? if not, it should go to the ext file

pkg/internal/collections/collection_helpers_test.go
```diff
@@ -1,6 +1,9 @@
 package collections
 
 import (
+	"errors"
+	"fmt"
+	"github.com/stretchr/testify/assert"
```
sfc-gh-asawicki, line 6: golangci

pkg/schemas/task_gen.go
```diff
@@ -102,19 +129,39 @@ func TaskToSchema(task *sdk.Task) map[string]any {
...
+		taskSchema["error_integration"] = task.ErrorIntegration.Name()
+	}
 	taskSchema["last_committed_on"] = task.LastCommittedOn
 	taskSchema["last_suspended_on"] = task.LastSuspendedOn
 	taskSchema["owner_role_type"] = task.OwnerRoleType
 	taskSchema["config"] = task.Config
 	taskSchema["budget"] = task.Budget
+	taskSchema["last_suspended_reason"] = task.LastSuspendedReason
+	finalizer := ""
+	if task.TaskRelations.FinalizerTask != nil {
```
sfc-gh-asawicki, line 151: if this file was edited, then please leave a comment because it will change with regeneration and document what we lack in the current generator

pkg/resources/task.go
```diff
@@ -2,709 +2,648 @@ package resources
...
-			return err
-		}
-		createRequest.WithSessionParameters(*sessionParameters)
+	if v, ok := d.GetOk("config"); ok {
+		req.WithConfig(v.(string))
 	}
 
-	if v, ok := d.GetOk("user_task_timeout_ms"); ok {
-		createRequest.WithUserTaskTimeoutMs(v.(int))
+	if v := d.Get("allow_overlapping_execution").(string); v != BooleanDefault {
```
sfc-gh-asawicki, line 234: did you consider using the helper methods here (check the user resource) - it makes some of this code more compact

pkg/resources/task_acceptance_test.go
```diff
@@ -1,564 +1,970 @@
...
-			Schema:            acc.TestSchemaName,
-			SQL:               "SELECT *",
-			When:              "FALSE",
-			Enabled:           true,
-			Schedule:          "15 MINUTE",
-			UserTaskTimeoutMs: 900000,
+	taskConfig := `$${"output_dir": "/temp/test_directory/", "learning_rate": 0.1}$$`
+	// We have to do three $ at the beginning because Terraform will remove one $.
+	// It's because `${` is a special pattern, and it's escaped by `$${`.
+	expectedTaskConfig := strings.ReplaceAll(taskConfig, "$", "")
```
sfc-gh-asawicki, line 247: if there is some repeated escaping logic for tasks configs please extract it either as a function here or in some test helper

pkg/resources/task_acceptance_test.go
sfc-gh-asawicki: +1

sfc-gh-asawicki (changes requested):

pkg/acceptance/bettertestspoc/assert/resourceassert/task_resource_ext.go
```diff
@@ -0,0 +1,17 @@
+package resourceassert
+
+import (
+	"fmt"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: yes, I wrote that I'm waiting for the pr with upgrader libraries. It's merged now, so I'll run pre-push

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/task_resource_parameters_ext.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: But this is actually "Medium" not "MEDIUM". I don't know why, but the that's the default value for this parameter.

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/task_resource_parameters_ext.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Don't remember, but @sfc-gh-asawicki is most likely right and that's some kind of workaround (same as with "Medium" case).

pkg/schemas/task_parameters.go
sfc-gh-jcieslak: Yeah, but currently, we don't do it for any other parameters schema, and I think even if we log it, it wouldn't really see it until someone gets an error because of it. Plus, there are a lot of parameters that are in the tasks that are not handled here. That would mean for every read for every task we would log at least 60 parameters (same for users).

pkg/schemas/task_gen.go
sfc-gh-jcieslak: It had to be adjusted manually, added comment and an entry in readme for schema generator

pkg/schemas/task_gen.go
sfc-gh-jcieslak: added comment and an entry in the readme for schema generator

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: I added this point to `next prs`. I would like to remove the need to put `$$` in the config field. It's repetitive and the user should be able to put any kind of JSON straight into this field without this strange escaping (especially when it's colliding with HCL syntax).

pkg/resources/task.go
sfc-gh-jcieslak: Yeah, I was going over user resource to look up some of the helper function you added there. Some of them were using this XProperty structs, so I just went the old way for now. I'll try to see what can be still converted in the next pr. Added to the list.

pkg/internal/collections/collection_helpers_test.go
sfc-gh-jcieslak: Yup, I merged with the main just now and ran the pre push

pkg/acceptance/bettertestspoc/assert/objectparametersassert/task_parameters_snowflake_gen.go
sfc-gh-jcieslak: The assert was re-generated, but probably the type has changed in this pr and after re-generation it was still "Medium", so I had to wrap it in `sdk.WarehouseSize` manually.

pkg/sdk/tasks_impl_gen.go
```diff
@@ -145,6 +153,11 @@ func GetRootTasks(v Tasks, ctx context.Context, id SchemaObjectIdentifier) ([]Ta
 			return nil, err
 		}
 
+		if task.TaskRelations.FinalizedRootTask != nil {
+			tasksToExamine.Push(*task.TaskRelations.FinalizedRootTask)
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm afaik task can only have one finalizer and it cannot refer to itself, so it's not needed in this case, but I can add it just in case.

pkg/sdk/tasks_impl_gen.go
sfc-gh-jcieslak: I think this could be a bigger discussion on handling certain parameters in convert and other mapping functions. For now, I would leave it as is, but also can revert because predecessors are not used in the resource.

pkg/resources/task_parameters.go
sfc-gh-jcieslak: Most of the fields were moved from the user, so I mostly focused on task-specific parameters and adjusted them. I can also adjust shared session parameters, but I was hesitant to over-validate some of the fields because, currently, I don't know exactly what values are possible to set. Maybe it should be done in SNOW-1645342? When moving the parameters to SDK.

pkg/resources/task.go
sfc-gh-jcieslak: Hmm you're probably right, I'll add a test for this case in this pr and adjust this if needed

sfc-gh-jcieslak: It seems like test passed, so I guess this is just a safeguard, maybe let's leave this comment for now and I'll take a closer look at it in the next pr. You can take a look at the test I added, imo it looks good, but maybe I did something wrong.

pkg/resources/task.go
sfc-gh-jcieslak: Yeah, I left it for the next pr to handle all the cleanup cases correctly. For now, I tried multiple **multiple** times to make the resource run the actions in the correct order.

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: But it's only referenced by id and not internally, meaning:
```terraform
finalize = "<identifier>"
```
is in config, not:
```terraform
finalize = snowflake_task.root_task.fully_qualified_name
```
That's why we need depends_on, or I misunderstood the comment?

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: Hmm missed them somehow. Uncommented \+ added asserts to the rest of the steps.

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: Yup, added.

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: I left TODO for the next pr to analyze this more in-depth. For now, I don't see why it's not there. I'll take a look if that's a mapping or Snowflake issue.

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: Created and referenced in TODO

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: True, in upcoming tests I also wanted to add tests for different kinds of things in `as` body. I'll link this comment in the desc and apply when testing other `as` options.

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: Added test cases for Finalize and After fields, added TODO for next pr to see what also could be tested

pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_gen.go
sfc-gh-jcieslak: Re-generated the asserts and removed what was not compiling and the needed asserts modified by hand were moved to _ect.go

pkg/acceptance/bettertestspoc/assert/objectparametersassert/task_parameters_snowflake_gen.go
sfc-gh-jcieslak: Just checked, the `return t.HasUserTaskManagedInitialWarehouseSize(sdk.WarehouseSize("Medium"))` is from generator

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/task_show_output_gen.go
sfc-gh-jcieslak: Moved to ext

pkg/acceptance/bettertestspoc/assert/resourceassert/task_resource_ext.go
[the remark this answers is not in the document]
sfc-gh-asawicki: Let's at least add InOrder suffix to the name of the function then (may be added in the next PR)

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/task_resource_parameters_ext.go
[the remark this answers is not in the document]
sfc-gh-asawicki: 🤦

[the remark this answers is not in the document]
sfc-gh-jmichalak: I guess we can leave it for now, or use sdk.WarehouseSizeMedium and others in the SDK converter. But I wonder how it behaves for other enum values...

pkg/resources/task_parameters.go
sfc-gh-asawicki: yes, let's not overvalidate the parameters because the problems we had (that params can have value outside the validation range if the account is modified by support)

pkg/resources/user_parameters.go
sfc-gh-jmichalak: Okay, I got confused because TODOs refer only to user parameter variables.

pkg/resources/task.go
sfc-gh-asawicki: You can add new ones! :)

sfc-gh-asawicki: (yes, it can be in the next PR)

pkg/resources/task_acceptance_test.go
sfc-gh-asawicki: oky

pkg/resources/task_acceptance_test.go
sfc-gh-jmichalak: Yup, it's fine.

sfc-gh-asawicki (dismissed):

pkg/resources/task.go
sfc-gh-jmichalak: Yeah, so this else block is executed when there's a change non-nil->nil. But in this if we expect the current state to be nil, and both of these can't happen at the same time, right?

sfc-gh-jmichalak (approved):

sfc-gh-jmichalak (approved):

pkg/resources/task.go
sfc-gh-jcieslak: [Adjusted a bit in](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3170): Not everything could be converted, but I think I changed what I could to attribute helpers, also added 2 of them (one for create and one for read).

sfc-gh-asawicki (approved):

pkg/resources/task.go
sfc-gh-jcieslak: Removed it, it seems from the tests it wasn't needed.

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: After a short analysis, It seems like a bug in Snowflake. For now, I removed this assert and added an entry in our doc about Snowflake improvements. I'll fill it later.

Issue #2951, opened 2024-07-19
[Feature]: Support snowflake_user TYPE property


mogren:
### Use Cases or Problem Statement

The existing [snowflake_user](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/user) resource does not have the TYPE property. This is a big limitation for us, since it prevents us from [creating SERVICE users](https://docs.snowflake.com/en/sql-reference/sql/create-user#label-user-type-property-person), which is a great security feature.

We are currently getting a number of Critical security warnings in Trust Center because of this issue.

### Category

category:resource

### Object type(s)

resource:user

### Proposal

Add a `type` property that accepts `PERSON | SERVICE | LEGACY_SERVICE | NULL`

### How much impact is this issue causing?

Medium

### Additional Information

Commented on a related PR here:

https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/2947#issuecomment-2239850099

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-asawicki:
Hey, we will handle this as a part of user redesign (that is currently happening).

Most probably, we will split the resource into three different ones (snowflake_user for people; snowflake_service_user and snowflake_legacy_service_user for the two others).

Redesigned snowflake_user may be ready for the next release but snowflake_service_user and snowflake_legacy_service_user will most probably have to wait a version.

mogren:
@sfc-gh-asawicki Hi Artur, any update on this?

sfc-gh-jcieslak:
Hey @mogren  👋 
What exactly would you like to know? If you're asking about the user rework, it was released as part of v0.95.0. The rest of the user resources (service and legacy_service) have to still wait a little (after today's release, a version or two). Artur is on vacation right now, but I'm guessing that will be one of his priorities after he comes back next week.

MrStemo:
Hi,

Following the Snowflake security update email we received yesterday we now have a deadline to update users to 'service users' by 30th September.

Is there any chance this is available in the TF by then?

Thanks

sfc-gh-jcieslak:
Hey 👋 
As far as I know, starting in October, the only thing that will be enforced is multi-factor authentication (MFA), not service users. Nevertheless, the timeline of missing user types will be discussed on Monday. I'll let you know the outcome of those discussions.

mogren:
[...]

PR #3119, opened 2024-10-08
feat: Add service user and legacy service user resources

sfc-gh-asawicki:
Introduce `snowflake_service_user` and `snowflake_legacy_service_user` resources:
- Add new resources (almost the same schema and the same limitations as `snowflake_user` resource - read below)
- Adjust datasource and add `type` to the `describe_output`
- Adjust SDK: handling types
- Add SDK integration tests to validate and document the Snowflake behavior of handling user types (create, alter set, alter unset, and parameters)
- Fix `TestAcc_User_issue2970` test
- Add migration guide
- Generate models and assertions for the new resources
- Fix sweepers

References: #2951 

##### Implementation considerations
Users of different types differ in Snowflake just slightly - a few fields are restricted from SERVICE and LEGACY_SERVICE type users. We decided to split Snowflake resources into multiple resources on our route to V1, to simplify the logic and usage (in this case, not to handle the type changes which are followed by attributes being hidden/revealed). That led to three choices for the implementation, to reuse the existing resource:
- copy-paste `snowflake_user` resource
- parameterized handling methods of `snowflake_user` resource with the user type
- enrich the helper methods with the knowledge of the schema (i.e. to change the logic of setting a field to always check first if the attribute exists in the given schema)

Ultimately, the second option was picked for the current implementation. More details of pros and cons of each solution are presented in points below.

###### copy-paste
Pros:
- Fastest implementation
- We plan to ultimately generate the resources, so at the end the implementation would be close to duplicated resources

Cons:
- Need to alter the logic in three different places
- Separate implementation of resources would require separate complete set of tests

###### parameterized handling methods
Pros:
- user object handled in "one" place
- implementation reused, so only smoke acceptance tests are justified

Cons:
- more complex logic in user resource, spread over multiple handling methods

###### enrich the helper methods
Pros:
- user resource implementation could stay the same in helper methods (because the checks would be delegated to helper methods)
- solution that can be reused in more resources
- only smoke tests would be justified

Cons:
- SDKv2 does not allow to fetch the schema in context of helper methods without exposing unexported fields
- "unsafe" solution potentially used in the majority of resources (because helper methods should be reused more and more)


MIGRATION_GUIDE.md
```diff
@@ -7,6 +7,36 @@ across different versions.
...
+- `snowflake_legacy_service_user` does not contain the following fields (because they are not supported for the user of type `LEGACY_SERVICE` in Snowflake):
+  - `first_name`
+  - `middle_name`
+  - `last_name`
+  - `mins_to_bypass_mfa`
+  - `disable_mfa`
+
+`snowflake_users` datasource was adjusted to handle different user types and `type` field was added to the `describe_output`.
+
+If you used to manage service or legacy service users through `snowflake_user` resource (e.g. using `lifecycle.ignore_changes`) or `snowflake_unsafe_execute`, please migrate to the new resources following [our guidelines on resource migration](docs/technical-documentation/resource_migration.md).
```
sfc-gh-jmichalak, line 36: It would be nice to include a short example here.

pkg/resources/custom_diffs.go
```diff
@@ -177,3 +180,22 @@ func ForceNewIfAllKeysAreNotSet(key string, keys ...string) schema.CustomizeDiff
 		return allUnset
 	})
 }
+
+func RecreateWhenUserTypeChangedExternally(userType sdk.UserType) schema.CustomizeDiffFunc {
+	return func(_ context.Context, diff *schema.ResourceDiff, _ interface{}) error {
+		if n := diff.Get("user_type"); n != nil {
+			logging.DebugLogger.Printf("[DEBUG] new external value for user type %s\n", n.(string))
+			acceptableUserTypes := []string{string(userType)}
+			if userType == sdk.UserTypePerson {
```
sfc-gh-jmichalak, line 189: Nit: I'd move this logic to SDK.

templates/resources/service_user.md.tmpl
```diff
@@ -0,0 +1,43 @@
...
+
+!> **V1 release candidate** This resource was reworked and is a release candidate for the V1. We do not expect significant changes in it before the V1. We will welcome any feedback and adjust the resource if needed. Any errors reported will be resolved with a higher priority. We encourage checking this resource out before the V1 release. Please follow the [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v0960--v0970) to use it.
+
+-> **Note** `snowflake_user_password_policy_attachment` will be reworked in the following versions of the provider which may still affect this resource.
+
+-> **Note** Attaching user policies will be handled in the following versions of the provider which may still affect this resource.
+
+-> **Note** Other two user types are handled in separate resources: `snowflake_legacy_service_user` for user type `legacy_service` and `snowflake_user` for user type `person`.
+
+-> **Note** External changes to `days_to_expiry` and `mins_to_unlock` are not currently handled by the provider (because the value changes continuously on Snowflake side after setting it).
```
sfc-gh-jmichalak, line 20: It looks good, but I think we can discuss how we want to note that external changes for some fields are not handled by the provider (for different reasons). For some resources, we just note this in the fields' descriptions.

pkg/acceptance/helpers/user_client.go
```diff
@@ -100,12 +118,19 @@ func (c *UserClient) SetDaysToExpiry(t *testing.T, id sdk.AccountObjectIdentifie
 	require.NoError(t, err)
 }
 
-func (c *UserClient) SetType(t *testing.T, id sdk.AccountObjectIdentifier, value string) {
+func (c *UserClient) SetType(t *testing.T, id sdk.AccountObjectIdentifier, userType sdk.UserType) {
+	t.Helper()
+	ctx := context.Background()
+
+	_, err := c.context.client.ExecForTests(ctx, fmt.Sprintf("ALTER USER %s SET TYPE = %s", id.FullyQualifiedName(), userType))
```
sfc-gh-jmichalak, line 125: Why is this not a part of the SDK?

pkg/resources/user.go
```diff
@@ -226,126 +216,207 @@ func User() *schema.Resource {
...
-	if err != nil {
-		return nil, err
-	}
+func ServiceUser() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: GetCreateUserFunc(sdk.UserTypeService),
+		UpdateContext: GetUpdateUserFunc(sdk.UserTypeService),
+		ReadContext:   GetReadUserFunc(sdk.UserTypeService, true),
+		DeleteContext: DeleteUser,
+		Description:   "Resource used to manage service user objects. For more information, check [user documentation](https://docs.snowflake.com/en/sql-reference/commands-user-role).",
```
sfc-gh-jmichalak, line 225: It should be https://docs.snowflake.com/en/sql-reference/sql/create-user. The same in other user schemas.

pkg/resources/user.go
```diff
@@ -389,25 +460,31 @@ func GetReadUserFunc(withExternalChangesMarking bool) schema.ReadContextFunc {
...
-			); err != nil {
+			showMappings := []showMapping{
+				{"login_name", "login_name", u.LoginName, u.LoginName, nil},
+				{"display_name", "display_name", u.DisplayName, u.DisplayName, nil},
+				{"disabled", "disabled", u.Disabled, fmt.Sprintf("%t", u.Disabled), nil},
+				{"default_namespace", "default_namespace", u.DefaultNamespace, u.DefaultNamespace, nil},
+				{"default_secondary_roles", "default_secondary_roles_option", u.DefaultSecondaryRoles, u.GetSecondaryRolesOption(), nil},
+			}
+			if userType == sdk.UserTypePerson {
+				showMappings = append(showMappings, showMapping{"must_change_password", "must_change_password", u.MustChangePassword, fmt.Sprintf("%t", u.MustChangePassword), nil})
```
sfc-gh-jmichalak, line 471: What about legacy service? It also has must_change_password.

sfc-gh-jmichalak (changes requested):

MIGRATION_GUIDE.md
sfc-gh-asawicki: Done

templates/resources/service_user.md.tmpl
sfc-gh-asawicki: This is a copy-paste from the user resource, so I added this here, too, for consistency. Added this topic to our documentation improvement epic.

pkg/acceptance/helpers/user_client.go
sfc-gh-asawicki: It is, it was not the part before, that's why it was handled through exec. Replaced.

But the unset was really missing in the SDK, so added it with tests and replaced too.

pkg/resources/user.go
sfc-gh-asawicki: It was set to this for user resource because of this link: https://docs.snowflake.com/en/sql-reference/commands-user-role#user-management. Changed to https://docs.snowflake.com/en/sql-reference/commands-user-role#user-management because it contains the list of all the commands, not only create.

examples/resources/snowflake_legacy_service_user/import.sh
```diff
@@ -0,0 +1 @@
+terraform import snowflake_legacy_service_user.example userName
```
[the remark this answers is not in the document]
sfc-gh-asawicki: changed for all user types

examples/resources/snowflake_legacy_service_user/resource.tf
```diff
@@ -0,0 +1,18 @@
...
+
+  default_warehouse       = "warehouse"
+  default_secondary_roles = "ALL"
+  default_role            = "role1"
+
+  rsa_public_key   = "..."
+  rsa_public_key_2 = "..."
+
+  must_change_password = true
+}
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Added for all user types

pkg/acceptance/helpers/user_client.go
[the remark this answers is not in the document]
sfc-gh-asawicki: done

pkg/sdk/testint/users_integration_test.go
```diff
@@ -726,6 +976,344 @@ func TestInt_Users(t *testing.T) {
...
+			},
+		}}
+
+		err = client.Users.Alter(ctx, user.ID(), alterOpts)
+		require.NoError(t, err)
+
+		assertions.AssertThatObject(t, objectassert.User(t, user.ID()).
+			HasDefaults(user.Name).
+			HasDisplayName("").
+			HasOwner(currentRole.Name()),
```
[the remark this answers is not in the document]
sfc-gh-asawicki: we have the HasDefaults check and this is precisely what UNSET does - it reverts to the defaults

pkg/resources/custom_diffs.go
sfc-gh-asawicki: extracted as preinitialized `AcceptableUserTypes`

sfc-gh-jmichalak (approved):

sfc-gh-jcieslak (approved):

PR #3122, opened 2024-10-09
feat: Stream on external table resource

sfc-gh-jmichalak:
- add snowflake_stream_on_external_table resource
- adjust copy grants documentation
- adjust external table helper client
- move copy grants handling to `resource_helpers_create.go`
- move baseModel to `ext`
- move common stream code to `stream_common.go`
- fix using empty columns in views recreation

## Test Plan

* [x] acceptance tests

## References

https://docs.snowflake.com/en/sql-reference/sql/create-stream

#3073
## TODO
- add remaining resources (stage, view)
- rework data source


pkg/resources/stream_common.go
```diff
@@ -134,3 +141,74 @@ var beforeSchema = &schema.Schema{
...
+	d.SetId("")
+	return nil
+}
+
+func handleStreamRead(d *schema.ResourceData,
+	id sdk.SchemaObjectIdentifier,
+	stream *sdk.Stream,
+	streamDescription *sdk.Stream,
+) error {
+	if err := d.Set("comment", stream.Comment); err != nil {
```
sfc-gh-asawicki, line 199: nit: we could use errors.Join(...) and return just one error here. Maybe it does not help that much with only 4 params but at least we return all the errors instead of just the first encountered one

pkg/resources/stream_on_external_table.go
```diff
@@ -0,0 +1,223 @@
...
+	id, err := sdk.ParseSchemaObjectIdentifier(d.Id())
+	if err != nil {
+		return nil, err
+	}
+
+	v, err := client.Streams.ShowByID(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	if err := d.Set("name", id.Name()); err != nil {
```
sfc-gh-asawicki, line 76: nit: these could be extracted because they set a few common stream fields

pkg/resources/view_acceptance_test.go
```diff
@@ -923,6 +923,63 @@ func TestAcc_View_Issue3073(t *testing.T) {
 	})
 }
 
+func TestAcc_View_IncorrectColumnsWithOrReplace(t *testing.T) {
```
sfc-gh-asawicki, line 926: nit: can you please reference the GH issue here, it helps to give context when someone will be studying these tests later not from the PR context

sfc-gh-asawicki (dismissed):

pkg/resources/stream_common.go
sfc-gh-jmichalak: I'll do this in the next PR.

pkg/resources/stream_on_external_table.go
sfc-gh-jmichalak: I'll do this in the next PR.

pkg/resources/view_acceptance_test.go
sfc-gh-jmichalak: I'll do this in the next PR.

pkg/resources/stream_on_table_acceptance_test.go
```diff
@@ -595,7 +593,6 @@ func TestAcc_StreamOnTable_InvalidConfiguration(t *testing.T) {
 		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
 			tfversion.RequireAbove(tfversion.Version1_5_0),
 		},
```
sfc-gh-jcieslak, line 595: Shouldn't it be changed to StreamOnTable checkDesctroy instead of removed completely?

pkg/resources/stream_on_external_table_acceptance_test.go
```diff
@@ -0,0 +1,586 @@
...
+					HasNameString(id.Name()),
+					assert.Check(resource.TestCheckResourceAttrWith(resourceName, "show_output.0.created_on", func(value string) error {
+						if value != createdOn {
+							return fmt.Errorf("view was recreated")
+						}
+						return nil
+					})),
+				),
+			},
+			{
```
sfc-gh-jcieslak, line 333: Shouldn't this test have at least one test case where attribute casing CREATE OR REPLACE to be called?

pkg/resources/stream_on_table_acceptance_test.go
sfc-gh-jmichalak: This test case doesn't create any resources because of invalid configuration, so there's no need to use checkDestroy.

pkg/resources/stream_on_external_table_acceptance_test.go
sfc-gh-jmichalak: Added.

pkg/resources/stream_common.go
```diff
@@ -5,58 +5,65 @@ import (
...
+		ForceNew:         true,
+		Description:      blocklistedCharactersFieldDescription("The database in which to create the stream."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"copy_grants": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     false,
+		Description: "Retains the access permissions from the original stream when a stream is recreated using the OR REPLACE clause. That is sometimes used when the provider detects changes for fields that can not be changed by ALTER. This value will not have any effect when creating a new stream.",
+		DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: I'll do this in the next PR.

sfc-gh-jcieslak (approved):

sfc-gh-asawicki (approved):

Issue #1150, opened 2022-07-21
State file does not capture whether a stream is stale


elikschultz:
**Provider version:**

0.37.0

**Terraform version:**

1.1.7

**Describe the bug:**

State files do not contain any indication of whether or not stream is stale.

**Expected behavior:**

State file should contain an indicator of whether or not a stream is stale, and if it is stale cause the stream to be recreated when a `terraform plan` / `terraform apply` is run.

sfc-gh-jmichalak:
Hi @elventear 👋

Sorry for the late response. We've released a new v0.98.0 version ([release](https://github.com/Snowflake-Labs/terraform-provider-snowflake/releases/tag/v0.98.0), [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v0970--v0980)) with new stream resources, based on stream type in [CREATE STREAM](https://docs.snowflake.com/en/sql-reference/sql/create-stream) All of the new resources contain `stale` field to detect stale streams and recreate them; and `show_output`, `describe_output` with `stale` field. Please upgrade to this version and use one of the relevant resources. Note that `stream` is now deprecated.

sfc-gh-jmichalak:
Closing due to inactivity.

PR #3129, opened 2024-10-11
feat: Add stream on directory table

sfc-gh-jmichalak:
- add new `snowflake_stream_on_directory_table` resource
- adjust SDK to handle non-qualified names for streams' tables
- address open comments from the previous [pr](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3122)
- recreate streams when they are stale (ref https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1150)
- change diff suppress on copy_grants
- improve procedure grants documentation (ref https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/3139)

## Test Plan

* [x] acceptance tests

* [ ] …

## References

https://docs.snowflake.com/en/sql-reference/sql/create-stream

## TODO
- add more acceptance tests for copy_grants for all of the relevant resources
- add streams on views
- add streams data source


pkg/resources/diff_suppressions.go
```diff
@@ -222,6 +222,22 @@ func suppressIdentifierQuoting(_, oldValue, newValue string, _ *schema.ResourceD
 	return slices.Equal(oldId, newId)
 }
 
+func suppressIdentifierQuotingPartiallyQualifiedName(_, oldValue, newValue string, _ *schema.ResourceData) bool {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Ok, now I can see why it's used, but still, I feel like this function should be treated just as a workaround and the stage name should be reported and eventually returned as fully qualified name.

pkg/resources/stream_common.go
```diff
@@ -196,19 +196,24 @@ func handleStreamRead(d *schema.ResourceData,
...
 
-	if err := d.Set(ShowOutputAttributeName, []map[string]any{schemas.StreamToSchema(stream)}); err != nil {
-		return err
-	}
-	if err := d.Set(DescribeOutputAttributeName, []map[string]any{schemas.StreamDescriptionToSchema(*streamDescription)}); err != nil {
-		return err
-	}
-	if err := d.Set(FullyQualifiedNameAttributeName, id.FullyQualifiedName()); err != nil {
-		return err
+func isStale(d *schema.ResourceData) bool {
```
sfc-gh-jcieslak, line 207: Question: What does "stale" mean in the context of a stream, and why do we want to recreate it when it's true?

pkg/sdk/streams_impl_gen.go
```diff
@@ -251,12 +251,7 @@ func (r showStreamsDbRow) convert() *Stream {
 		}
 	}
 	if r.BaseTables.Valid {
-		baseTables, err := ParseCommaSeparatedSchemaObjectIdentifierArray(r.BaseTables.String)
-		if err != nil {
-			log.Printf("[DEBUG] error converting show stream: %v", err)
-		} else {
-			s.BaseTables = baseTables
-		}
+		s.BaseTables = ParseCommaSeparatedStringArray(r.BaseTables.String, false)
```
sfc-gh-jcieslak, line 254: Why was it changed, it can return different types of identifiers ?

pkg/resources/stream_on_directory_table_acceptance_test.go
```diff
@@ -0,0 +1,342 @@
...
+	stage, cleanupStage := acc.TestClient().Stage.CreateStageWithDirectory(t)
+	t.Cleanup(cleanupStage)
+
+	model := model.StreamOnDirectoryTable("test", id.DatabaseName(), id.Name(), id.SchemaName(), stage.ID().FullyQualifiedName())
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: acc.CheckDestroy(t, resources.StreamOnTable),
```
sfc-gh-jcieslak, line 283: Hmm the test only checks if the resource was recreated and with the AlwaysIgnore suppress diff I think `copy_grants` doesn't work properly (because it's not changed). You can just assert the state and see the value is the same because of the diff suppress (I think).

pkg/resources/stream_on_directory_table_acceptance_test.go
```diff
@@ -0,0 +1,342 @@
...
+						HasFullyQualifiedNameString(id.FullyQualifiedName()).
+						HasStageString(stage.ID().Name()).
+						HasCommentString("bar"),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_StreamOnDirectoryTable_CopyGrants(t *testing.T) {
```
sfc-gh-asawicki, line 266: I'd say that this test only checks the proper suppression (it could be simplified with emptyplan checks, though), but I would rather see a longer test, that takes copy grants into consideration and we check it does (e.g. that with copy grants set to true and replacement, we still have the grants, and with set to false we don't).

sfc-gh-asawicki (changes requested):

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/resource_schema_def.go
```diff
@@ -65,4 +65,12 @@ var allResourceSchemaDefs = []ResourceSchemaDef{
 		name:   "StreamOnExternalTable",
 		schema: resources.StreamOnExternalTable().Schema,
 	},
+	{
+		name:   "StreamOnDirectoryTable",
+		schema: resources.StreamOnDirectoryTable().Schema,
+	},
+	// {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Removed.

pkg/resources/diff_suppressions.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: This is on purpose. Snowflake returns only partially qualified names (meaning no database and schema) for directory table streams. But we have them in the config, so we need to suppress them in this way here. And yes, this is just a workaround. Added a todo.

pkg/resources/stream_common.go
```diff
@@ -34,13 +36,11 @@ var streamCommonSchema = map[string]*schema.Schema{
...
-		Default:     false,
-		Description: "Retains the access permissions from the original stream when a stream is recreated using the OR REPLACE clause. That is sometimes used when the provider detects changes for fields that can not be changed by ALTER. This value will not have any effect when creating a new stream.",
-		DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
-			return oldValue != "" && oldValue != newValue
-		},
+		Type:             schema.TypeBool,
+		Optional:         true,
+		Default:          false,
+		Description:      "Retains the access permissions from the original stream when a stream is recreated using the OR REPLACE clause. That is sometimes used when the provider detects changes for fields that can not be changed by ALTER. This value will not have any effect when creating a new stream.",
+		DiffSuppressFunc: IgnoreAlways,
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Exactly. Added a comment.

pkg/resources/stream_common.go
sfc-gh-jmichalak: You can read more [here](https://docs.snowflake.com/en/user-guide/streams-intro#data-retention-period-and-staleness).

When a stream is stale, and you try reading from it, you get an error message telling you to recreate it ([link](https://community.snowflake.com/s/article/using-tasks-to-avoid-stale-streams-when-incoming-data-is-empty)). This logic is also requested by the community - see https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1150. I added a comment to make it more obvious.

pkg/sdk/streams_impl_gen.go
sfc-gh-jmichalak: Answered above.

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-jmichalak: The point of this test is to make sure that changing copy_grants doesn't recreate the whole resource.

MIGRATION_GUIDE.md
```diff
@@ -11,11 +11,12 @@ across different versions.
 
 ### *(new feature)* snowflake_stream_on_table, snowflake_stream_on_external_table resource
 
-To enhance clarity and functionality, the new resources `snowflake_stream_on_table` and `snowflake_stream_on_external_table` have been introduced to replace the previous `snowflake_stream`. Recognizing that the old resource carried multiple responsibilities within a single entity, we opted to divide it into more specialized resources.
+To enhance clarity and functionality, the new resources `snowflake_stream_on_table`, `snowflake_stream_on_external_table` and `snowflake_stream_on_directory_table` have been introduced to replace the previous `snowflake_stream`. Recognizing that the old resource carried multiple responsibilities within a single entity, we opted to divide it into more specialized resources.
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Corrected.

pkg/provider/resources/resources.go
```diff
@@ -48,8 +48,10 @@ const (
 	Stage                                                  resource = "snowflake_stage"
 	StorageIntegration                                     resource = "snowflake_storage_integration"
 	Stream                                                 resource = "snowflake_stream"
+	StreamOnDirectoryTable                                 resource = "snowflake_stream_on_directory_table"
 	StreamOnExternalTable                                  resource = "snowflake_stream_on_external_table"
 	StreamOnTable                                          resource = "snowflake_stream_on_table"
+	StreamOnView                                           resource = "snowflake_stream_on_view"
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Removed.

pkg/resources/stream_on_external_table.go
```diff
@@ -199,7 +193,7 @@ func UpdateStreamOnExternalTable(ctx context.Context, d *schema.ResourceData, me
 	}
 
 	// change on these fields can not be ForceNew because then the object is dropped explicitly and copying grants does not have effect
-	if keys := changedKeys(d, "external_table", "insert_only", "at", "before"); len(keys) > 0 {
+	if keys := changedKeys(d, "external_table", "insert_only", "at", "before"); len(keys) > 0 || isStale(d) {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Done.

MIGRATION_GUIDE.md
```diff
@@ -37,27 +33,42 @@ resource "snowflake_stream" "stream" {
...
 
   comment = "A stream."
 }
 ```
 
+Then, follow our [Resource migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/docs/technical-documentation/resource_migration.md).
 
-To use the new `stream_on_directory_table`, change the old `stream` from
+
+## v0.96.0 ➞ v0.97.0
```
sfc-gh-jcieslak, line 50: This is already present at line 10

docs/resources/stream_on_directory_table.md
```diff
@@ -72,6 +72,7 @@ resource "snowflake_stream_on_directory_table" "stream" {
 - `fully_qualified_name` (String) Fully qualified name of the resource. For more information, see [object name resolution](https://docs.snowflake.com/en/sql-reference/name-resolution).
 - `id` (String) The ID of this resource.
 - `show_output` (List of Object) Outputs the result of `SHOW STREAMS` for the given stream. (see [below for nested schema](#nestedatt--show_output))
+- `stale` (Boolean) Indicated if the stream is stale. When Terraform detects that the stream is stale, the stream is recreated with `CREATE OR REPLACE`.
```
sfc-gh-jcieslak, line 75: From the documentation page you sent me, it seems you can just call https://docs.snowflake.com/en/sql-reference/functions/system_stream_has_data to prevent the stream from becoming stale instead of recreating the whole stream, but maybe eventually we would have to re-create it anyway (?). Not sure, but we could check if that would work instead of re-creating the stream.

pkg/resources/stream_common.go
sfc-gh-jcieslak: Got it, can we put the first link to the description of the stale field then?

docs/resources/stream_on_directory_table.md
sfc-gh-jmichalak: It can be used to extend stale_after date on already non-stale streams. When you call this function on a stale stream, it doesn't make it non-stale.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: Fixed.

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-asawicki: Okay, I'll keep it open then

MIGRATION_GUIDE.md
```diff
@@ -37,27 +33,42 @@ resource "snowflake_stream" "stream" {
...
+
+### *(new feature)* snowflake_stream_on_table, snowflake_stream_on_external_table resource
+
+To enhance clarity and functionality, the new resources `snowflake_stream_on_table`, `snowflake_stream_on_external_table` and `snowflake_stream_on_directory_table` have been introduced to replace the previous `snowflake_stream`. Recognizing that the old resource carried multiple responsibilities within a single entity, we opted to divide it into more specialized resources.
+The newly introduced resources are aligned with the latest Snowflake documentation at the time of implementation, and adhere to our [new conventions](#general-changes).
+This segregation was based on the object on which the stream is created. The mapping between SQL statements and the resources is the following:
+- `ON TABLE <table_name>` -> `snowflake_stream_on_table`
+- `ON EXTERNAL TABLE <external_table_name>` -> `snowflake_stream_on_external_table` (this was previously not supported)
+- `ON STAGE <stage_name>` -> `snowflake_stream_on_directory_table`
```
sfc-gh-asawicki, line 59: we should not list it in v097 changes, its enough to have it in v098. Also, in v098 version migration you don't have to repeat this whole paragraph and other resources, it's enough to add a reference and just describe the newly introduced resource

docs/resources/stream_on_external_table.md
```diff
@@ -89,6 +89,7 @@ resource "snowflake_stream_on_external_table" "stream" {
 - `fully_qualified_name` (String) Fully qualified name of the resource. For more information, see [object name resolution](https://docs.snowflake.com/en/sql-reference/name-resolution).
 - `id` (String) The ID of this resource.
 - `show_output` (List of Object) Outputs the result of `SHOW STREAMS` for the given stream. (see [below for nested schema](#nestedatt--show_output))
+- `stale` (Boolean) Indicated if the stream is stale. When Terraform detects that the stream is stale, the stream is recreated with `CREATE OR REPLACE`. Read more on stream staleness in Snowflake [docs](https://docs.snowflake.com/en/user-guide/streams-intro#data-retention-period-and-staleness).
```
sfc-gh-asawicki, line 92: this new field should be added listed in the migration guide with explantation

pkg/resources/diff_suppressions.go
```diff
@@ -254,8 +254,3 @@ func IgnoreNewEmptyListOrSubfields(ignoredSubfields ...string) schema.SchemaDiff
 		return len(parts) == 3 && slices.Contains(ignoredSubfields, parts[2]) && new == ""
 	}
 }
-
-// IgnoreAlways should be used to ignore changes in every case.
-func IgnoreAlways(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
sfc-gh-asawicki: this should no have been deleted

pkg/resources/stream_common.go
```diff
@@ -196,24 +201,15 @@ func handleStreamRead(d *schema.ResourceData,
 	stream *sdk.Stream,
 	streamDescription *sdk.Stream,
 ) error {
+	stale, err := booleanStringToBool(*stream.Stale)
```
sfc-gh-asawicki, line 204: please extract this check with d.Set as a new helper function in resource_helpers_read.go and just use this function in the join below

sfc-gh-asawicki: Also, why we are dereferencing the pointer like this here and just not pass a proper value in Stream from SDK? This deref is not safe here.

pkg/resources/stream_on_directory_table.go
```diff
@@ -137,7 +139,8 @@ func UpdateStreamOnDirectoryTable(ctx context.Context, d *schema.ResourceData, m
 	}
 
 	// change on these fields can not be ForceNew because then the object is dropped explicitly and copying grants does not have effect
-	if keys := changedKeys(d, "stage"); len(keys) > 0 || isStale(d) {
+	// recreate when the stream is stale - see https://community.snowflake.com/s/article/using-tasks-to-avoid-stale-streams-when-incoming-data-is-empty
+	if keys := changedKeys(d, "stage", "stale"); len(keys) > 0 {
```
sfc-gh-asawicki, line 143: why we use changed keys and not the HasChange/GetChange? Also, what if this is changed here from true to false - which would happen after the recreation, right?

pkg/acceptance/helpers/stage_client.go
```diff
@@ -44,9 +44,9 @@ func (c *StageClient) CreateStageWithURL(t *testing.T, id sdk.SchemaObjectIdenti
 	return stage, c.DropStageFunc(t, id)
 }
 
-func (c *StageClient) CreateStageWithDirectory(t *testing.T) (*sdk.Stage, func()) {
+func (c *StageClient) CreateStageWithDirectory(t *testing.T, schemaId sdk.DatabaseObjectIdentifier) (*sdk.Stage, func()) {
```
sfc-gh-asawicki, line 47: If you want to pass a schema, use the `CreateStageInSchema` (or introduce new CreateStageWithDirectoryInSchema). The current usage is inconsistent with other existing helpers. It's also visible by the usage, we should not need to pass the default test schema.

pkg/sdk/testint/streams_gen_integration_test.go
```diff
@@ -137,7 +137,7 @@ func TestInt_Streams(t *testing.T) {
 	})
 
 	t.Run("CreateOnDirectoryTable", func(t *testing.T) {
-		stage, cleanupStage := testClientHelper().Stage.CreateStageWithDirectory(t)
+		stage, cleanupStage := testClientHelper().Stage.CreateStageWithDirectory(t, testClientHelper().Ids.SchemaId())
```
sfc-gh-asawicki, line 140: here is the example where we should just use the method without passing the schema

pkg/resources/stream_on_external_table_acceptance_test.go
```diff
@@ -414,6 +416,108 @@ func TestAcc_StreamOnExternalTable_CopyGrants(t *testing.T) {
...
+	var createdOn string
+
+	model := model.StreamOnExternalTable("test", id.DatabaseName(), externalTable.ID().FullyQualifiedName(), id.Name(), id.SchemaName()).WithInsertOnly(r.BooleanTrue)
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: acc.CheckDestroy(t, resources.StreamOnExternalTable),
+		Steps: []resource.TestStep{
```
sfc-gh-asawicki, line 451: I would like to see two different tests here:
1. The one you put here but modified slightly and informing that permadiff will always occur with such a setup.
2. The test with a bit more sophisticated setup:
- initial creation does not lead to stale stream (e.g. setting the schema params initially to non-zero value)
- changing the value externally on schema and checking manually (through client) that stream turned stale
- changing schema parameters back to non-zero values (+ optionally ensuring that the stream is still stale and did not magically unstale on its own)
- running the terraform to check that we recreate the stale stream and expecting no changes after

pkg/resources/stream_on_directory_table_acceptance_test.go
```diff
@@ -320,6 +321,46 @@ func TestAcc_StreamOnDirectoryTable_CopyGrants(t *testing.T) {
...
+
+	model := model.StreamOnDirectoryTable("test", id.DatabaseName(), id.Name(), id.SchemaName(), stage.ID().FullyQualifiedName())
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: acc.CheckDestroy(t, resources.StreamOnDirectoryTable),
+		Steps: []resource.TestStep{
+			// TODO (SNOW-1737932): Setting schema parameters related to retention time seems to have no affect on streams on directory tables.
```
sfc-gh-asawicki, line 350: let's find other way to make the stream stale then in this case

sfc-gh-asawicki (changes requested):

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-jcieslak: Got it, I was wondering if the state value of copy_grants is even assertable since there's this IgnoreAfterCreate diff suppress that is keeping it always the same in state.

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-jmichalak: I haven't found any. Do you have anything on your mind?

pkg/resources/stream_on_external_table_acceptance_test.go
sfc-gh-jmichalak: 1. Not sure what you mean - how modified exactly?
2. Added.

pkg/sdk/testint/streams_gen_integration_test.go
sfc-gh-jmichalak: Done.

pkg/acceptance/helpers/stage_client.go
sfc-gh-jmichalak: Done.

pkg/resources/stream_on_directory_table.go
sfc-gh-jmichalak: Well, we could use `HasChanges`, but now we can log which fields were exactly changed, which is useful information for potential debugging, IMO. 

I added a comm to RecreateWhenStreamIsStale. Basically, our desired state is `false` for this field. That's why when we detect `true`, we set a new value to `false` to detect this change and recreate the stream. But after the recreation the old value would be `false`.

pkg/resources/stream_common.go
sfc-gh-jmichalak: I decided to change it to bool and convert it in the sdk.

pkg/resources/diff_suppressions.go
sfc-gh-jmichalak: This is not used now.

docs/resources/stream_on_external_table.md
sfc-gh-jmichalak: Done.

MIGRATION_GUIDE.md

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-asawicki: I would experiment with changing the stage in some breaking ways. and additionally, I would just ask the stream team how we can make this particular type of stream stale.

pkg/resources/stream_on_external_table_acceptance_test.go
sfc-gh-asawicki: 1. FIrst of all name. And secondly, if there is a permadiff, we don't need that many steps (create and next step should be enough)

pkg/resources/stream_on_directory_table.go
sfc-gh-asawicki: But we get this info from the plan, right? I asked because this is different than the logic in other resources and I think we should have a reason for inconsistencies. Yes, logging is nice, but would we benefit also from it in other resources or is it just applicable here? If we could have leverage this for better logging in every resource then I would say that this is a discussion topic for the unified logging later on.

sfc-gh-asawicki: And thanks for the true -> false explanation.

sfc-gh-jmichalak: Yes, we have this info in the plan, but the plan doesn't say that the resource will be recreated, and why that happens - there may be changes in other fields that don't cause recreation. That's why I added this logging only in resources with copy grants.

pkg/resources/stream_on_external_table_acceptance_test.go
sfc-gh-jmichalak: I'm still not sure what you mean by that, but I changed this (only for this resource for now).

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-jmichalak: Added here.

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-jmichalak: For now, I've added this behavior to our list and reported it in a streams channel. We test this logic in other resources, so I think there's little chance for a bug here. Also, we have a follow-up Jira on our side.

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-jmichalak: I changed this a bit and added more tests.

pkg/resources/stream_on_external_table_acceptance_test.go
```diff
@@ -470,8 +539,92 @@ func TestAcc_StreamOnExternalTable_RecreateWhenStale(t *testing.T) {
...
+				PreConfig: func() {
+					acc.TestClient().Schema.Alter(t, schema.ID(), &sdk.AlterSchemaOptions{
+						Set: &sdk.SchemaSet{
+							DataRetentionTimeInDays:    sdk.Int(0),
+							MaxDataExtensionTimeInDays: sdk.Int(0),
+						},
+					})
+					assert.AssertThatObject(t, objectassert.Stream(t, id).
+						HasName(id.Name()).
+						HasStale(true),
```
sfc-gh-asawicki, line 625: like we discussed, here it would be better to revert the retention values to non-zero values in the preconfig

pkg/resources/stream_on_external_table_acceptance_test.go
sfc-gh-asawicki: One more comment on the new test (closing this thread).

sfc-gh-asawicki (dismissed):

sfc-gh-jcieslak (approved):

pkg/resources/stream_on_external_table_acceptance_test.go
sfc-gh-jmichalak: Done - also for stream on table.

pkg/resources/stream_on_table_acceptance_test.go
```diff
@@ -476,33 +476,11 @@ func TestAcc_StreamOnTable_RecreateWhenStale(t *testing.T) {
 					})),
 				),
 			},
-			// set schema parameters to bigger values ensuring that the stream is not stale
-			{
-				PreConfig: func() {
```
sfc-gh-jcieslak: Why was it removed? (same in other file)

sfc-gh-jmichalak: Because before there was one test for handling staleness. As a result of https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3129#discussion_r1801940817, it was split up, and here I removed this step not to test the same thing twice - see also https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3129#discussion_r1806639261

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

PR #3130, opened 2024-10-11
chore: Test support for object renaming

sfc-gh-jcieslak:
## Changes
- Added integration tests to show the inconsistency between ShowByID errors
- Added acceptance tests to show different behaviors of objects after renaming higher-hierarchy objects
- Added generated models for database and schema
- Added new test command to Makefile that runs the tests (they're separated from the CI tests by env variable that has to be enabled to run the tests)

## TODO
- Pick the important tests that should be running with other acceptance tests
- Transform shallow hierarchy tests into parametrized ones (similar to what we have right now in deep hierarchy test).


pkg/sdk/testint/errors_integration_test.go
```diff
@@ -0,0 +1,146 @@
...
+	doesNotExistOrNotAuthorized := sdk.ErrObjectNotExistOrAuthorized.Error() // Database '\"non-existing-database\"' does not exist or not authorized
+	doesNotExistOrOperationCannotBePerformed := "Object does not exist, or operation cannot be performed"
+
+	testCases := []struct {
+		ObjectType  sdk.ObjectType
+		ExpectedErr string
+		ShowFn      func(context.Context, sdk.SchemaObjectIdentifier) error
+	}{
+		// Only object types that use IN SCHEMA in their ShowByID implementation
+		{ObjectType: sdk.ObjectTypeTable, ExpectedErr: doesNotExistOrOperationCannotBePerformed, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Tables.ShowByID)},
```
sfc-gh-asawicki, line 35: omg. are the errors for the given objects at least consistent for unexisting db/schema? (it's hard for me to compare on GH)

sfc-gh-asawicki: The second thought: maybe after such a test, we could make the error returned from the show query consistent throughout the SDK? (so that on the resource level we can always check for the same error). And of course I don't mean we should do it now, but this is a nice improvement to document.

pkg/resources/object_renaming_acceptance_test.go
```diff
@@ -0,0 +1,1366 @@
...
+				},
+				Config:      config.FromModels(t, databaseConfigModelWithNewId, schemaModelConfigWithNewDatabaseId),
+				ExpectError: regexp.MustCompile("does not exist or not authorized"),
+			},
+		},
+	})
+}
+
+func TestAcc_ShallowHierarchy_IsInConfig_RenamedInternally_WithDependsOn_AfterRenameSchemaReferencingOldDatabaseName(t *testing.T) {
+	// Error happens during schema's Read operation and then Delete operation (schema cannot be removed).
```
sfc-gh-asawicki, line 256: how about having the destroy step only? would it help at all? (probably not, but worth asking)

pkg/resources/object_renaming_acceptance_test.go
```diff
@@ -0,0 +1,1366 @@
...
+
+type DependencyType string
+
+const (
+	ImplicitDependency  DependencyType = "implicit"
+	DependsOnDependency DependencyType = "depends_on"
+	NoDependency        DependencyType = "no_dependency"
+)
+
+func TestAcc_ShallowHierarchy_IsInConfig_RenamedInternally_WithImplicitDependency(t *testing.T) {
```
sfc-gh-asawicki, line 139: question: do you think that setting up parametrized tests for the shallow hierarchy too would also help in readability?

sfc-gh-asawicki (approved):

pkg/sdk/testint/errors_integration_test.go
sfc-gh-jcieslak: I think they are consistent (as far as I remember, I copy-pasted the whole test for the database and didn't have to change much). Yeah, we can make them as one error. The one thing that I didn't like during this whole testing was that we already do this for the error `does not exist or not authorized`; I think a better solution would be to enclose the error as it is right now but somehow unwrap it when returning errors in resources. That way, the user will get more meaningful errors instead of generic ones. We could also benefit from this when getting logs from users. That's just my thought because, as far as I remember, I had many many errors during the testing, and unwrapping could maybe speed debugging up a bit (it's still not that bad because the original errors are visible in the SQL logs).

pkg/resources/object_renaming_acceptance_test.go
sfc-gh-jcieslak: Yeah, I tried, but this step will also fail because it's the same as removing the resource after the test. Or maybe there's some other way, but I had too many tests to write to spend more time on this.

pkg/resources/object_renaming_acceptance_test.go
```diff
@@ -0,0 +1,1366 @@
...
+	_ = testenvs.GetOrSkipTest(t, testenvs.EnableObjectRenamingTest)
+	_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)
+	acc.TestAccPreCheck(t)
+
+	testCases := []struct {
+		DatabaseDependency     DependencyType
+		SchemaDependency       DependencyType
+		ExpectedFirstStepError *regexp.Regexp
+	}{
+		{DatabaseDependency: ImplicitDependency, SchemaDependency: NoDependency, ExpectedFirstStepError: regexp.MustCompile("error creating table")}, // tries to create table before schema
```
sfc-gh-jmichalak, line 978: Aren't these kind of cases also non-deterministic? A schema could be created before a table.

pkg/resources/object_renaming_acceptance_test.go
sfc-gh-jcieslak: Yes, after I wrote the deeper hierarchy I saw how the tests are much more compact (and readable). I wanted to rewrite shallow ones, but on the other hand, I thought to leave them as they are just because they're already there and maybe it's better to leave them in separation and avoid if-ology in test setups (if possible). Don't have strong feelings for this one 😅 wdyt?

pkg/resources/object_renaming_acceptance_test.go
```diff
@@ -0,0 +1,1366 @@
...
+					},
+				},
+				Config: config.FromModel(t, databaseConfigModelWithNewId) + configSchemaWithDatabaseReference(databaseConfigModelWithNewId.ResourceReference(), schemaName),
+			},
+		},
+	})
+}
+
+func TestAcc_ShallowHierarchy_IsInConfig_RenamedInternally_WithoutDependency_AfterRenameSchemaReferencingOldDatabaseName(t *testing.T) {
+	// Error happens during schema's Read operation and then Delete operation (schema cannot be removed).
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: We can, but the overall test will fail anyway. It's currently failing because the terraform is trying to remove schema on a non-existing database, so dropping it by hand wouldn't do anything.

pkg/resources/object_renaming_acceptance_test.go
```diff
@@ -0,0 +1,1366 @@
...
+				},
+				Config: config.FromModels(t, databaseConfigModel, schemaModelConfig),
+				// ExpectError: regexp.MustCompile("does not exist or not authorized"),
+			},
+		},
+	})
+}
+
+// This test checks if the order of the configuration resources has any impact on the order of resource execution (it seems to have no impact).
+func TestAcc_ShallowHierarchy_IsInConfig_RenamedExternally_WithoutDependency_AfterRenameDatabaseReferencingOldNameAndSchemaReferencingOldDatabaseName_ConfigOrderSwap(t *testing.T) {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: xd yeah, I wanted to trick the terraform to see if order in config matters, but I recently found a forum where it's explained that the order doesn't matter. What matters is how the dependencies are setup in the config (and probably other internal rules that probably depend on the combination of resource_name + resource_block_name).

pkg/resources/object_renaming_acceptance_test.go
sfc-gh-jcieslak: I also thought about it, but in this configuration, the order seems to be consistent xd. Only this one test is non-deterministic. What's even crazier is that the same test, but with a different order in the config (`_ConfigOrderSwap` test), had some impact on the execution order because it also seems to be more consistent. Don't ask me, but non-deterministic results only seem to happen in `TestAcc_ShallowHierarchy_IsInConfig_RenamedExternally_WithoutDependency_AfterRenameDatabaseReferencingOldNameAndSchemaReferencingOldDatabaseName` (I think it was this one).

pkg/sdk/testint/errors_integration_test.go
sfc-gh-asawicki: let's add this to the Sdk improvements and we'll handle it after v1

pkg/resources/object_renaming_acceptance_test.go
sfc-gh-asawicki: I have the same feeling that the deeper ones are much more readable now. Usually, I am against complex setup in parametrized tests but this use case is a bit different, we care about the permutations so setting the tests this way would improve readability and make them more understandable. I also don't feel a need for a strong push to rewrite them now, maybe in one of the next PRs or something

pkg/resources/object_renaming_acceptance_test.go
```diff
@@ -0,0 +1,1366 @@
...
+					t.Cleanup(acc.TestClient().Database.DropDatabaseFunc(t, newDatabaseId))
+				},
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("snowflake_database.test", plancheck.ResourceActionCreate),
+						plancheck.ExpectResourceAction("snowflake_schema.test", plancheck.ResourceActionCreate),
+					},
+				},
+				Config:      config.FromModels(t, databaseConfigModel, schemaModelConfigWithNewDatabaseId),
+				ExpectError: regexp.MustCompile("Failed to create schema"), // already exists
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Not really, from the test title `_AfterRenameDatabaseReferencingOldNameAndSchemaReferencingNewDatabaseName` schema is referencing a new database name. As we talked about it, changing database of schema can make it remove itself from state in Read which is good, then the terraform tries to create this schema, but fails because schema already exists in new database. Don't know if that answers the question, I'll write additional comment to help with explanation.

pkg/resources/object_renaming_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: No, the name is perfect :), and I don't have any idea for better naming overall.

pkg/resources/object_renaming_acceptance_test.go
```diff
@@ -0,0 +1,1366 @@
...
+			{
+				PreConfig: func() {
+					acc.TestClient().Database.Alter(t, databaseId, &sdk.AlterDatabaseOptions{
+						NewName: &newDatabaseId,
+					})
+					t.Cleanup(acc.TestClient().Database.DropDatabaseFunc(t, newDatabaseId))
+				},
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("snowflake_database.test", plancheck.ResourceActionCreate),
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Added comments

pkg/resources/object_renaming_acceptance_test.go
```diff
@@ -0,0 +1,1366 @@
...
+	databaseConfigModelWithNewId := model.Database("test", newDatabaseId.Name())
+
+	schemaModelConfig := model.Schema("test", databaseId.Name(), schemaName)
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: resource.ComposeAggregateTestCheckFunc(),
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Fixed

pkg/acceptance/bettertestspoc/config/model/schema_model_ext.go
```diff
@@ -0,0 +1 @@
+package model
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, it was probably used at some point, but currently it's useless, removed.

pkg/resources/object_renaming_acceptance_test.go
sfc-gh-jmichalak: Yeah, that's really weird, but we don't need to spend much time here, because almost always we'll have dependencies in such configurations.

sfc-gh-jmichalak (approved):

pkg/sdk/testint/errors_integration_test.go
sfc-gh-asawicki: Both seem valid, I would be going by the SDK improvements probably this week, so maybe I will split the whole epic into smaller ones, will see. It's fine to keep it in the current one for now.

pkg/resources/object_renaming_acceptance_test.go
sfc-gh-asawicki: keeping it open then

sfc-gh-asawicki (approved):

PR #3131, opened 2024-10-11
feat: Secrets datasource

sfc-gh-fbudzynski:
## Changes
- added datasource for secrets
- generated documentation for secrets
- ComposeCheckDestroy function
## Test Plan

* [X] acceptance tests

## References
https://docs.snowflake.com/en/sql-reference/sql/create-secret


pkg/datasources/secrets_acceptance_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+func TestAcc_Secrets_MultipleTypes(t *testing.T) {
+	prefix := random.AlphaN(4)
+	idOne := acc.TestClient().Ids.RandomSchemaObjectIdentifierWithPrefix(prefix + "A")
+	idTwo := acc.TestClient().Ids.RandomSchemaObjectIdentifierWithPrefix(prefix + "B")
+	idThree := acc.TestClient().Ids.RandomSchemaObjectIdentifierWithPrefix(prefix + "C")
+	idFour := acc.TestClient().Ids.RandomSchemaObjectIdentifierWithPrefix(prefix + "D")
+
+	integrationId := acc.TestClient().Ids.RandomAccountObjectIdentifier()
+	_, apiIntegrationCleanup := acc.TestClient().SecurityIntegration.CreateApiAuthenticationClientCredentialsWithRequest(t,
```
sfc-gh-asawicki, line 35: if you want to run such statements before PreConfig, then you should use EnableAcceptance and acc.Precheck at the start of the test functions (check e.g. TestAcc_SecurityIntegrations_OauthForCustomClients).

The reasons: currently the setup for acceptance tests is run in acc.Precheck, so usage of `acc.TestClient()` may result in strange behavior. The EnableAcceptance check prevents running it when running integration tests (for now we do go test with filters, and terraform testing framework checks for TF_ACC late)

pkg/datasources/secrets_acceptance_test.go
```diff
@@ -0,0 +1,305 @@
...
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: acc.CheckDestroy(t, resources.SecretWithClientCredentials),
+		Steps: []resource.TestStep{
+			{
+				Config: dataSecretsClientCredentials,
+				Check: assert.AssertThat(t,
+					assert.Check(resource.TestCheckResourceAttr(dsName, "secrets.#", "1")),
+
+					assert.Check(resource.TestCheckResourceAttr(dsName, "secrets.0.show_output.#", "1")),
```
sfc-gh-asawicki, line 151: there's a trick to use show_output asserts here, check `users_datasource_show_output_ext.go`

pkg/datasources/secrets_acceptance_test.go
```diff
@@ -0,0 +1,305 @@
...
+		accConfig.FromModel(t, secretModelAuthorizationCodeGrant) +
+		secretsDataAndVars()
+
+	configVariables := config.Variables{
+		"like": config.StringVariable(prefix + "%"),
+	}
+
+	dsName := "data.snowflake_secrets.test"
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
```
sfc-gh-asawicki, line 60: Check destroy is missing

pkg/datasources/secrets_acceptance_test.go
```diff
@@ -0,0 +1,305 @@
...
+)
+
+const (
+	secretWithClientCredentials      = "snowflake_secret_with_client_credentials"
+	secretWithAuthorizationCodeGrant = "snowflake_secret_with_authorization_code_grant"
+	secretWithBasicAuthentication    = "snowflake_secret_with_basic_authentication"
+	secretWithGenericString          = "snowflake_secret_with_generic_string"
+)
+
+func TestAcc_Secrets_MultipleTypes(t *testing.T) {
```
sfc-gh-asawicki, line 27: Do we benefit from this test? There are other tests that check single type, we can validate the output thoroughly there, here it would be nice to focus on filtering instead checking the output

sfc-gh-asawicki (changes requested):

docs/data-sources/secrets.md
```diff
@@ -0,0 +1,80 @@
+---
+page_title: "snowflake_secrets Data Source - terraform-provider-snowflake"
+subcategory: ""
+description: |-
+  Datasource used to get details of filtered secrets. Filtering is aligned with the current possibilities for SHOW SECRETS https://docs.snowflake.com/en/sql-reference/sql/show-secrets query. The results of SHOW and DESCRIBE are encapsulated in one output collection secrets.
+---
+
+# snowflake_secrets (Data Source)
+
+Datasource used to get details of filtered secrets. Filtering is aligned with the current possibilities for [SHOW SECRETS](https://docs.snowflake.com/en/sql-reference/sql/show-secrets) query. The results of SHOW and DESCRIBE are encapsulated in one output collection `secrets`.
```
sfc-gh-jcieslak, line 10: I think we also put here the v1 candidate note

pkg/datasources/secrets_acceptance_test.go
sfc-gh-fbudzynski: Is `acc.TestAccPreCheck(t)` sufficient or is there a need for `testenvs.GetOrSkipTest()` just like in the `TestAcc_SecurityIntegrations_OauthForCustomClients` as well?

sfc-gh-asawicki: GetOrSkipTest is needed too (that's what I meant by saying use EnableAcceptance, sorry)

pkg/datasources/secrets_acceptance_test.go
sfc-gh-fbudzynski: Will use that

pkg/datasources/secrets_acceptance_test.go
sfc-gh-fbudzynski: ok i agree, we would benefit much more by testing filtering on multiple types, and tests each type output independently in tests below

pkg/datasources/secrets_acceptance_test.go
sfc-gh-fbudzynski: Is there a way to handle checkDestroy for multiple types? Looking at `TestAcc_SecurityIntegrations_MultipleTypes` there is no check destroy either.

sfc-gh-jmichalak: I don't think it is. You can write a new function that composes checkDestroys into one.

pkg/acceptance/check_destroy.go
```diff
@@ -15,6 +15,20 @@ import (
...
 )
 
+func ComposeCheckDestroy(t *testing.T, resources ...resources.Resource) func(*terraform.State) error {
+	t.Helper()
+
+	return func(s *terraform.State) error {
+		for _, resource := range resources {
+			checkFunc := CheckDestroy(t, resource)
+			if err := checkFunc(s); err != nil {
+				return fmt.Errorf("error checking destruction for resource %s: %w", resource, err)
```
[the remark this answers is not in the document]
sfc-gh-asawicki: +1

pkg/datasources/secrets.go
```diff
@@ -104,15 +104,15 @@ func Secrets() *schema.Resource {
 
 func ReadSecrets(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	client := meta.(*provider.Context).Client
-	req := sdk.NewShowSecretRequest()
+	req := *sdk.NewShowSecretRequest()
```
sfc-gh-asawicki, line 107: why this change?

pkg/datasources/secrets_acceptance_test.go
```diff
@@ -303,3 +215,169 @@ func TestAcc_Secrets_WithGenericString(t *testing.T) {
...
+	idFour := acc.TestClient().Ids.RandomSchemaObjectIdentifier()
+
+	integrationId := acc.TestClient().Ids.RandomAccountObjectIdentifier()
+	_, apiIntegrationCleanup := acc.TestClient().SecurityIntegration.CreateApiAuthenticationClientCredentialsWithRequest(t,
+		sdk.NewCreateApiAuthenticationWithClientCredentialsFlowSecurityIntegrationRequest(integrationId, true, "test_oauth_client_id", "test_oauth_client_secret").
+			WithOauthAllowedScopes([]sdk.AllowedScope{{Scope: "first_scope"}, {Scope: "second_scope"}}),
+	)
+	t.Cleanup(apiIntegrationCleanup)
+
+	// ERROR Insufficient privileges to operate on account ... for role BASIC_PRIVILEGES
```
sfc-gh-asawicki, line 243: let's uncomment it and run on CI (you can also verify it on higher role now)

sfc-gh-asawicki (changes requested):

examples/data-sources/snowflake_secrets/data-source.tf
```diff
@@ -0,0 +1,46 @@
...
+}
+
+output "like_prefix_output" {
+  value = data.snowflake_secrets.like_prefix.secrets
+}
+
+# Filtering (in)
+data "snowflake_secrets" "in" {
+  in {
+    schema = "schema_fully_qualified_name"
```
sfc-gh-jcieslak, line 30: I think we try to use `fully_qualified_name` computed values instead of strings, so this should be `schema = snowflake_schema.test.fully_qualified_name`, check other examples on that

pkg/datasources/secrets.go
sfc-gh-fbudzynski: ig it is not necessary so i will leave it as it was before so this is oversight from my side

pkg/datasources/secrets_acceptance_test.go
sfc-gh-fbudzynski: i have not seen any test for `in application` and `in application package` so those have been removed, and altered the `in account` test to use prefix so it doesnt show any other secrets defined outside of this test.

examples/data-sources/snowflake_secrets/data-source.tf
sfc-gh-fbudzynski: changed it

examples/data-sources/snowflake_secrets/data-source.tf
```diff
@@ -27,7 +27,7 @@ output "like_prefix_output" {
 # Filtering (in)
 data "snowflake_secrets" "in" {
   in {
-    schema = "schema_fully_qualified_name"
+    schema = "snowflake_schema.test.fully_qualified_name"
```
sfc-gh-jcieslak, line 30: It shouldn't be quoted (it's a reference to another object, [here's](https://developer.hashicorp.com/terraform/tutorials/configuration-language/dependencies#manage-implicit-dependencies) a link to docs on implicit dependencies). We could add the actual resource (`resource "snowflake_schema" "test" { ...<fields>... }`), but we do this rarely, so I think there's no need.

sfc-gh-fbudzynski: Thanks for the docs, i have changed it to be used as reference to another obj, so removed the quotes

sfc-gh-jmichalak (approved):

sfc-gh-jcieslak (approved):

sfc-gh-asawicki (approved):

PR #3110, opened 2024-10-02
feat: Secret resource

sfc-gh-fbudzynski:
## Changes
- add `snowflake_secret_with_client_credentials` resource
- add `snowflake_secret_with_authorization_code_grant` resource
- add `snowflake_secret_with_basic_authentication` resource
- add `snowflake_secret_with_generic_string` resource
- fix parsing oauth_scopes list with `ParseCommaSeparatedStringArray()`

## Test Plan

* [x] acceptance tests

## References

https://docs.snowflake.com/en/sql-reference/sql/create-secret

## TODO
- datasource
- tests for externally changed secret type

PR #3141, opened 2024-10-17
fix: handle external change of secret type

sfc-gh-fbudzynski:
## Changes
* Added secret_type computed filed to secret resources
* Added CustomDiff for handling external change of secret_type
* Refactored show_and_describe_handlers for flat describe_output and show_output
* Tested external change of secret type for each resource

## Test Plan

* [x] acceptance tests

## References

* https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3110#discussion_r1793543864
* https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3110#discussion_r1802006700


pkg/resources/secret_with_oauth_authorization_code_grant_acceptance_test.go
```diff
@@ -314,12 +314,73 @@ func TestAcc_SecretWithAuthorizationCodeGrant_ExternalChange(t *testing.T) {
...
 					assert.AssertThat(t,
-						resourceassert.SecretWithAuthorizationCodeResource(t, secretModel.ResourceReference()).
+						resourceassert.SecretWithAuthorizationCodeGrantResource(t, secretModel.ResourceReference()).
 							HasOauthRefreshTokenExpiryTimeString(refreshTokenExpiryDateTime),
 						assert.Check(resource.TestCheckResourceAttrSet(secretModel.ResourceReference(), "describe_output.0.oauth_refresh_token_expiry_time")),
 					),
 				),
 			},
+			{
+				RefreshState: true,
```
sfc-gh-jcieslak, line 324: I see this kind of test case for the first time, could you briefly describe what it does/tests ?

pkg/resources/secret_with_oauth_client_credentials_acceptance_test.go
```diff
@@ -297,3 +297,62 @@ func TestAcc_SecretWithClientCredentials_EmptyScopesList(t *testing.T) {
 		},
 	})
 }
+
+func TestAcc_SecretWithClientCredentials_ExternalSecretTypeChange(t *testing.T) {
```
sfc-gh-jcieslak, line 301: What would happen if we would change this object into `oauth_authorization_code_grant`, since they have the same secret type?

pkg/sdk/testint/secrets_gen_integration_test.go
```diff
@@ -75,7 +75,7 @@ func TestInt_Secrets(t *testing.T) {
 			objectassert.Secret(t, id).
 				HasName(id.Name()).
 				HasComment("a").
-				HasSecretType(sdk.SecretTypeOAuth2).
+				HasSecretType(string(sdk.SecretTypeOAuth2)).
```
sfc-gh-jcieslak, line 78: Question to the changes in this file: why enums were changed to strings? With asserts I guess it's a limitation of the generator, but what about secretDetails?

pkg/resources/secret_with_oauth_authorization_code_grant_acceptance_test.go
sfc-gh-fbudzynski: When discussing the `refresh_token_expiry_time` field with @sfc-gh-jmichalak we wanted to check if there will be a perma diff or not when invoking Read on that resource in order to be sure if we needed a diff suppress.

There was no perma diff because this filed is not being overwritten by value in describe, it is handled by using `setStateToValuesFromConfig`

`withExternalChangesMarking` and `handleExternalValueChangesToObjectInDescribe` is being used only to detect the external change on that field.

pkg/sdk/testint/secrets_gen_integration_test.go
sfc-gh-fbudzynski: It have changed in order to use `var AcceptableSecretTypes = map[SecretType]string` and pass secret types as sdk.SecretType type as it is done for the user_type CustomDiff `RecreateWhenUserTypeChangedExternally` 

But in case of secrets it is not really needed so it will be changed back

pkg/resources/secret_with_basic_authentication.go
```diff
@@ -43,8 +43,9 @@ func SecretWithBasicAuthentication() *schema.Resource {
 
 		CustomizeDiff: customdiff.All(
 			ComputedIfAnyAttributeChanged(secretBasicAuthenticationSchema, ShowOutputAttributeName, "name", "comment"),
```
sfc-gh-jmichalak, line 45: `name`can also be removed from here (it has ForceNew).

pkg/resources/custom_diffs.go
```diff
@@ -195,3 +195,32 @@ func RecreateWhenUserTypeChangedExternally(userType sdk.UserType) schema.Customi
 		return nil
 	}
 }
+
+func RecreateWhenSecretTypeChangedExternally(secretType string, oauthType *sdk.OauthSecretType) schema.CustomizeDiffFunc {
+	return func(_ context.Context, diff *schema.ResourceDiff, _ interface{}) error {
+		if n := diff.Get("secret_type"); n != nil {
+			logging.DebugLogger.Printf("[DEBUG] new external value for secret type %s\n", n.(string))
+			if secretType != strings.ToUpper(n.(string)) {
```
sfc-gh-jmichalak, line 203: Instead of comparing strings here, we can introduce a function converting from a string to a secret type - just like in other resources.

pkg/resources/custom_diffs.go
```diff
@@ -195,3 +195,32 @@ func RecreateWhenUserTypeChangedExternally(userType sdk.UserType) schema.Customi
 		return nil
 	}
 }
+
+func RecreateWhenSecretTypeChangedExternally(secretType string, oauthType *sdk.OauthSecretType) schema.CustomizeDiffFunc {
```
sfc-gh-jmichalak, line 199: It would be great to have this unit tested.

pkg/resources/custom_diffs.go
```diff
@@ -195,3 +195,32 @@ func RecreateWhenUserTypeChangedExternally(userType sdk.UserType) schema.Customi
...
+			if secretType != strings.ToUpper(n.(string)) {
+				return errors.Join(diff.SetNew("secret_type", "<changed externally>"), diff.ForceNew("secret_type"))
+			}
+			// both client_credentials and authorization_code_grant secrets have the same type: "OAUTH2"
+			// to detect the external type change we need to check fields that are required in one, but should be absent in the other
+			// we will check if the 'oauth_refresh_token_expiry_time' is present in the describe_output
+			// since it is required in authorization_code_grant flow and should be empty in client_credentials flow
+			if strings.ToUpper(n.(string)) == string(sdk.SecretTypeOAuth2) && oauthType != nil {
+				var forceNew bool
+				rt := diff.Get("describe_output.0.oauth_refresh_token_expiry_time").(string)
```
sfc-gh-jmichalak, line 212: I think it would be more readable to have a variable `isRefreshTokenExpiryTimeEmpty` and use this bool in the switch. wdyt?

pkg/resources/secret_with_oauth_authorization_code_grant.go
```diff
@@ -146,9 +147,11 @@ func ReadContextSecretWithAuthorizationCodeGrant(withExternalChangesMarking bool
 			return diag.FromErr(err)
 		}
 
-		if withExternalChangesMarking {
-			if err = handleExternalValueChangesToObjectInDescribe(d,
-				describeMapping{"oauth_refresh_token_expiry_time", "oauth_refresh_token_expiry_time", secretDescription.OauthRefreshTokenExpiryTime.String(), secretDescription.OauthRefreshTokenExpiryTime.String(), nil},
+		// if secret type is changed externally, we wont be able to read oauth_refresh_token_expiry_time value (since it will not be provided)
```
sfc-gh-jmichalak, line 150: I think in such cases, we should check if we can return early (immediately after show/desc and setting the type). We're setting some values here, but the resource is going to be recreated anyway in this scenario.

pkg/resources/secret_common.go
```diff
@@ -33,6 +33,11 @@ var secretCommonSchema = map[string]*schema.Schema{
 		ForceNew:         true,
 		DiffSuppressFunc: suppressIdentifierQuoting,
 	},
+	"secret_type": {
+		Type:        schema.TypeString,
+		Computed:    true,
+		Description: "Specifies a type for the secret.",
```
sfc-gh-jmichalak, line 39: You can add a comment saying that we use this field for checking external values and recreating the resources, if needed. The reason is that in general we have our computed fields in `show_output` and `describe_output` and I think users expect configurable fields here.

pkg/resources/show_and_describe_handlers.go
```diff
@@ -35,36 +34,26 @@ func handleExternalChangesToObjectInShow(d *schema.ResourceData, mappings ...sho
...
-type showMapping struct {
-	nameInShow     string
+// handleExternalChangesToObjectInShow assumes that show output is kept in ShowOutputAttributeName attribute
+func handleExternalChangesToObjectInShow(d *schema.ResourceData, mappings ...outputMapping) error {
+	return handleExternalChangesToObject(d, ShowOutputAttributeName, mappings...)
+}
+
+// handleExternalChangesToObjectInFlatDescribe assumes that describe output is kept in DescribeOutputAttributeName attribute
+// It is to be used with flat - show like describe_output schemas
+// To handle external changes to describe with properties like collections use `handleExternalChangesToObjectInDescribe()`
```
sfc-gh-jmichalak, line 44: 👍

sfc-gh-jmichalak (changes requested):

pkg/sdk/secrets_def.go
```diff
@@ -7,10 +7,19 @@ import (
...
+
 const (
-	SecretTypePassword      = "PASSWORD"
-	SecretTypeOAuth2        = "OAUTH2"
-	SecretTypeGenericString = "GENERIC_STRING"
+	SecretTypePassword                               = "PASSWORD"
+	SecretTypeOAuth2                                 = "OAUTH2"
+	SecretTypeGenericString                          = "GENERIC_STRING"
+	OAuth2ClientCredentialsFlow      OauthSecretType = "CLIENT_CREDENTIALS"       // #nosec G101
+	OAuth2AuthorizationCodeGrantFlow OauthSecretType = "AUTHORIZATION_CODE_GRANT" // #nosec G101
```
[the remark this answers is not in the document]
sfc-gh-fbudzynski: I wanted to distinguished those two types because those are not used the same way as 
- SecretTypePassword
- SecretTypeOAuth2 
- SecretTypeGenericString
are used. 
Those two types are used to recognize the change between `SecretWithClientCredentials` and `SecretWithAuthorizationCodeGrant`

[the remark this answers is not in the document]
sfc-gh-jmichalak: True, I read that later. But still, I think we can have typed the previous 3 types, e.g., `SecretType.`

pkg/resources/custom_diffs.go
sfc-gh-fbudzynski: i agree thats a better name for this variable

pkg/resources/secret_common.go
sfc-gh-fbudzynski: will be added

pkg/sdk/secrets_def.go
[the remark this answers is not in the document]
sfc-gh-fbudzynski: Agreed

pkg/resources/custom_diffs.go
sfc-gh-asawicki, line 199: The setup of this func is a bit convoluted and delegated to the user's invocation. We know which resources result in which type and what additional logic is needed. Let's move this logic inside the function. To do that, let's have all 4 values of secret type as input and adjust the logic properly.

sfc-gh-asawicki (changes requested):

pkg/resources/secret_with_oauth_authorization_code_grant.go
sfc-gh-fbudzynski: im not really sure how to check it and return earlier in a way that checking the type change for same `OAUTH2` type will work... we relay on describe_output in that case, so it needs to be set either way

pkg/resources/custom_diffs.go
sfc-gh-fbudzynski: ok i did something similar to what is done for the user_type, that way the invocation is done only by providing one argument, one of the four types of creating the secret

pkg/sdk/secrets_def.go
```diff
@@ -2,26 +2,50 @@ package sdk
...
+	}
 }
 
 const (
-	SecretTypePassword                               = "PASSWORD"
-	SecretTypeOAuth2                                 = "OAUTH2"
-	SecretTypeGenericString                          = "GENERIC_STRING"
-	OAuth2ClientCredentialsFlow      OauthSecretType = "CLIENT_CREDENTIALS"       // #nosec G101
-	OAuth2AuthorizationCodeGrantFlow OauthSecretType = "AUTHORIZATION_CODE_GRANT" // #nosec G101
+	SecretTypePassword                     SecretType = "PASSWORD"
```
sfc-gh-asawicki, line 34: nit: I don't like that this is one enum here. Two different use cases are mixed here: SQL TYPE value and creation variants. I'm okay with merging this but let's talk later how we can improve this.

sfc-gh-asawicki (approved):

pkg/resources/secret_with_oauth_authorization_code_grant_acceptance_test.go
sfc-gh-jcieslak: Diffs after read are usually implicitly caught by the testing framework, but If we want to check that explicitly let's make a comment above this test case to know what are we actually asserting and what do we expect to happen.

sfc-gh-jcieslak (approved):

pkg/resources/secret_with_oauth_authorization_code_grant.go
sfc-gh-jmichalak: That's okay, I thought it would be simpler :/

sfc-gh-jmichalak (dismissed):

sfc-gh-jmichalak (approved):

Issue #753, opened 2021-11-09
Terraform inappropriratly renames columns instead of adding/removing what is being specified


asosnovsky:
**Provider Version**: 0.25.19

**Terraform Version**: 1.0.10

**Describe the bug**

Say I have a deployment like this

```terraform

resource "snowflake_database" "db" {
  name = "test_db"
}

resource "snowflake_schema" "schema" {
  database = snowflake_database.db.name
  name     = "example"
}

resource "snowflake_table" "tb" {
  database = snowflake_database.db.name
  schema   = snowflake_schema.schema.name
  name     = "tb1"
  column {
    name    = "name"
    type    = "VARCHAR(255)"
    comment = "User name"
  }
  column {
    name    = "age"
    type    = "NUMERIC"
    comment = "User's age"
  }
}
```

I applied the changes, and then one day a requirement is added, we need a new column added to `tb1`!
So when I include the column:

```
resource "snowflake_table" "tb" {
  database = snowflake_database.db.name
  schema   = snowflake_schema.schema.name
  name     = "tb1"
  column {
    name    = "name"
    type    = "VARCHAR(255)"
    comment = "User name"
  }
  column {
    name = "new_col"
    type = "VARCHAR(255)"
  }
  column {
    name    = "age"
    type    = "NUMERIC"
    comment = "User's age"
  }
}
```
I would expect there to be a change that specifies a new column is to be added, instead I get this:
![image](https://user-images.githubusercontent.com/7451445/140973546-159d43bd-dbd3-4879-8a6a-f5797a7b6c28.png)

Notice how the column is shifting everything down, renames an existing column and then adds a column that was already there.

Additionally, I noticed that if after this apply, I were to say remove the 'user name' i.e.

```terraform
resource "snowflake_table" "tb" {
  database = snowflake_database.db.name
  schema   = snowflake_schema.schema.name
  name     = "tb1"
  column {
    name = "new_col"
    type = "VARCHAR(255)"
  }
  column {
    name    = "age"
    type    = "NUMERIC"
    comment = "User's age"
  }
}
```

It shifts all the columns upward...
![image](https://user-images.githubusercontent.com/7451445/140973884-37c20730-77c3-48aa-8357-ed320102cca7.png)

I suspect this is due to the implementation detail of storing the NewColumn as a list and not checking if we are actually looking at the same column. Perhaps generating a unique hash to represent each column during the creation phase would resolve this issue (though it may be a big breaking change).

[...]

Issue #420, opened 2021-01-21
Support the ability to alter table...rename a column


sfc-gh-dsandler:
As a data modeler, I need to rename a column.  I would like to use the snowflake_table resource to support this action

From pure-DDL perspective, it would like this:

`alter table mydb.myschema.mytbl rename column col1  to column1;`

to be supported with a ne property called OLD_NAME in the nested schema for column (as such):

```
resource "snowflake_table" "my_tbl" {
  name     = "MYTBL"
  database = "MYDB"
  schema   = "MYSCHEMA"
  column   {
            name = "COLUMN1"
            old_name = "COL1"
            type = "STRING" 
  } 
  comment  = "test table"

  depends_on = [snowflake_schema.schema]
}
```

kumarjoshi:
This would be a great feature. Plus, the ability to change data type would be great too.

juan-yunis:
I would like to see this feature released any time soon!

mddataminr:
I would love this functionality too. I imagine the issue is maintaining the column id across each iteration. ie, if you change the name, right now TF is going to treat it like a drop and add. However, if the column had on id field, you could change attributes of it without losing the relationship. 

The problem with this is that you'd now have a TF field to create with each column that is only valuable occassionally. 
`resource "snowflake_table" "my_tbl" {
  name     = "MYTBL"
  database = "MYDB"
  schema   = "MYSCHEMA"
  column   {
            id = "column_1_id"
            name = "COLUMN1"
            type = "STRING" 
  } 
  comment  = "test table"

  depends_on = [snowflake_schema.schema]
}`
Or maybe we pass in a column resource instead of the current structure? Something like:
`resource "snowflake_table" "my_tbl" {
  name     = "MYTBL"
  database = "MYDB"
  schema   = "MYSCHEMA"
  resource "snowflake_column" "column1"   {
            name = "COLUMN1"
            type = "STRING" 
  } 
  comment  = "test table"

  depends_on = [snowflake_schema.schema]
}`

I'm not sure if the option above is syntactically possible in TF, but something along either line would be very helpful.

sagar-raythatha:
I would love to have this feature. I am new to Go language but I would be happy to collaborate with anyone to implement this feature.

elikschultz:
I am also experiencing issues with this. To add a little more color, the problem extends beyond not being able to alter a column in place, as trying to rename a column can also cause unintended changes to other columns that are not being modified.

[...]

PR #3147, opened 2024-10-22
chore: Detect changes in lists and sets

sfc-gh-jcieslak:
Test cases for changes in lists and sets. From the given time, I only went through essential cases that consisted of:
- Ignoring the order of list items after creation.
- Having the ability to update an item while ignoring the order.
For the testing, I created a test resource because currently, we don't have anything to test more complex examples of certain resource behaviors (temporary providers we create for custom diff testing are not sufficient in this case). The resource is only added to the resource list whenever a special env is set (we could remove it entirely and leave some documentation in the resource file (and acc test file) on how to prepare for the tests). The imitation of external storage was done by creating a struct and its global instance (some of the things needed to be exported since external changes tested in acceptance tests needed to access those). Certain resource fields were researched to test different approaches, each is described in the implementation file.

Also added an assert on lists/sets that is able to assert items in order independent manner (it was needed for the tests of the proposals).

> Note: Only lists were tested, because there was no major issue with them in current resources. For the lists the following issues were addressed: #420, #753

## Next pr
- Apply (parameterized tests in object renaming test cases) https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3130#discussion_r1802651147


pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+
+	d.SetId("identifier")
+
+	return ReadObjectRenamingListsAndSets(ctx, d, meta)
+}
+
+func UpdateObjectRenamingListsAndSets(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if d.HasChange("list") {
+		// It wasn't working with d.getChange(). It was returning null elements in one of the test case's steps.
+		oldList := d.GetRawState().AsValueMap()["list"].AsValueSlice()
```
sfc-gh-asawicki, line 212: question: can we get nil in any of these two (at the `["list"]`)? If so, we should handle it.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: objectRenamingConfigList([]map[string]any{
+					{"name": "", "string": "111", "int": 111},
+					{"name": "", "string": "222", "int": 222},
+					{"name": "", "string": "333", "int": 333},
```
sfc-gh-jmichalak, line 31: We should include plans to indicate what really happens inside the resource.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					assert.HasListItemsOrderIndependent("snowflake_object_renaming.test", "list", []map[string]string{
+						{"name": "", "string": "444", "int": "444"},
+						{"name": "", "string": "555", "int": "555"},
+						{"name": "", "string": "333", "int": "333"},
+						{"name": "", "string": "222", "int": "222"},
+					}),
+				),
+			},
+			// Add an item that is identical to another one (currently, failing)
```
sfc-gh-jmichalak, line 120: Could you elaborate on that?

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+		finalState := ObjectRenamingDatabaseInstance.ManuallyOrderedList
+
+		for index, item := range ObjectRenamingDatabaseInstance.ManuallyOrderedList {
+			order := index + 1
+			newValue, err := collections.FindFirst(newManuallyOrderedList.([]any), func(item any) bool {
+				return item.(map[string]any)["order"].(int) == order
+			})
+
+			// Items were removed
+			if errors.Is(err, collections.ErrObjectNotFound) {
```
sfc-gh-asawicki, line 264: I don't quite get the logic here: if we did not find the given order it can mean that only a single object was removed but all the subsequent ones exist, right?

Okay, I think I understand, so this is this assumption that there will be no gaps. But with this assumption, why we can't just initially check the new length and just trim initially?

Also, it would be nice to have this assumption validated on the resource level (i.e. that there are no gaps and that it starts with 1).

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+			{
+				Config: objectRenamingConfigList([]map[string]any{
+					{"name": "", "string": "444", "int": 444},
+					{"name": "", "string": "555", "int": 555},
+					{"name": "", "string": "333", "int": 333},
+					{"name": "", "string": "222", "int": 222},
+				}),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					assert.HasListItemsOrderIndependent("snowflake_object_renaming.test", "list", []map[string]string{
+						{"name": "", "string": "444", "int": "444"},
```
sfc-gh-jmichalak, line 96: `name` is not used and can be omitted.

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+			addedItem.Order = ""
+			ObjectRenamingDatabaseInstance.OrderedList = append(ObjectRenamingDatabaseInstance.OrderedList, addedItem)
+		}
+
+		// The implementation is not complete due to mentioned issues with computed order in DiffSuppressFunc
+	}
+
+	if d.HasChange("manually_ordered_list") {
+		_, newManuallyOrderedList := d.GetChange("manually_ordered_list")
+		hasNewItems := len(newManuallyOrderedList.([]any)) > len(ObjectRenamingDatabaseInstance.ManuallyOrderedList)
```
sfc-gh-asawicki, line 252: and what if one item was removed and one was added? it has a new item then, right? it could be named HasMoreItems or something like this.

sfc-gh-asawicki: also nit: we don't need it in the first loop, so we can move instantiation till later, right?

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+					assert.HasListItemsOrderIndependent("snowflake_object_renaming.test", "list", []map[string]string{
+						{"name": "", "string": "222", "int": "222"},
+						{"name": "", "string": "444", "int": "444"},
+						{"name": "", "string": "333", "int": "333"},
+					}),
+				),
+			},
+			// Remove, shift, and add one item (at the end)
+			{
+				Config: objectRenamingConfigList([]map[string]any{
```
sfc-gh-jmichalak, line 58: There are missing cases for external changes.

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+
+		// Copy the value from external source and compare with the new state to create the final state that will be saved
```
sfc-gh-asawicki, line 254: question: this is nice for testing but we may not have the order from Snowflake, right? How this would behave if the external source does not provide the order?

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+	}
+}
+
+func ignoreOrderAfterFirstApplyWithManuallyOrderedList(parentKey string) schema.SchemaDiffSuppressFunc {
+	return func(key string, oldValue string, newValue string, d *schema.ResourceData) bool {
+		if strings.HasSuffix(key, ".#") {
+			return false
+		}
+
+		// Waw state is not null after first apply
```
sfc-gh-asawicki, line 447: nit: Waw -> Raw

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+					// Matching hashes indicate the order changed, but the item stayed in the config, so suppress the change
+					if oldItem.Hash() == newItemHash {
+						newItemWasAlreadyPresent = true
+					}
+				}
+
+				oldItemIsStillPresent := false
+
+				// Sizes of config and state may not be the same
+				if len(d.GetRawState().AsValueMap()[parentKey].AsValueSlice()) > index {
```
sfc-gh-jmichalak, line 380: These functions are deeply nested - pls try using `collections.Map`, extracting some code and returning earlier.

sfc-gh-jmichalak (changes requested):

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+				Check: resource.ComposeAggregateTestCheckFunc(
+					assert.HasListItemsOrderIndependent("snowflake_object_renaming.test", "list", []map[string]string{
+						{"name": "", "string": "444", "int": "444"},
+						{"name": "", "string": "555", "int": "555"},
+						{"name": "", "string": "333", "int": "333"},
+						{"name": "", "string": "222", "int": "222"},
+					}),
+				),
+			},
+			// Replace all items
```
sfc-gh-asawicki, line 103: These are not replaced, config is exactly the same.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+				}),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					assert.HasListItemsOrderIndependent("snowflake_object_renaming.test", "list", []map[string]string{
+						{"name": "", "string": "444", "int": "444"},
+						{"name": "", "string": "555", "int": "555"},
+						{"name": "", "string": "333", "int": "333"},
+						{"name": "", "string": "222", "int": "222"},
+					}),
+				),
+			},
```
sfc-gh-asawicki, line 119: I would like a test for:
- less items than before
- removing all items and later adding some

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+			//			{"name": "", "string": "222", "int": "222"},
+			//			{"name": "", "string": "222", "int": "222"},
+			//		}),
+			//	),
+			// },
+		},
+	})
+}
+
+// This test researches the possibility of performing update instead of remove + add item
```
sfc-gh-asawicki, line 143: how this test proves/disproves this possibility if we only check the final state?

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+	acc "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvs"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_BasicListFlow(t *testing.T) {
```
sfc-gh-asawicki, line 16: General thought: this test is pretty straightforward; it just checks if the state is equal to config but does not validate suppressing diff/actions that happen underneath, so I am not sure if it really proves something. Let's discuss this today.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+					{"name": "name2", "order": 2},
+					{"name": "name4", "order": 3},
+				}),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_object_renaming.test", "manually_ordered_list.0.name", "name2"),
+					resource.TestCheckResourceAttr("snowflake_object_renaming.test", "manually_ordered_list.1.name", "name2"),
+					resource.TestCheckResourceAttr("snowflake_object_renaming.test", "manually_ordered_list.2.name", "name4"),
+				),
+			},
+			// Change the order
```
sfc-gh-asawicki, line 318: Change the order in config but not change the desired order of items

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	return fmt.Sprintf(`
+resource "snowflake_object_renaming" "test" {
+	%s
+}
+`, generatedListItems)
+}
+
+func TestAcc_BasicManuallyOrderedListFlow(t *testing.T) {
```
sfc-gh-asawicki, line 282: it would be nice to see this in action also for item with name + type (like e.g. argument/column) + order so that we check how much work is needed to handle any number of attributes in the given item.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
sfc-gh-asawicki: Maybe if we keep access to "external state" we could also have access to some resource actions log, and we could verify what actions are happening (so that we can better validate if the logic conforms with the expectations and so that we can later use it in practice)

sfc-gh-asawicki (changes requested):
Let's discuss today on the meeting

pkg/acceptance/bettertestspoc/assert/commons.go
```diff
@@ -101,3 +105,85 @@ func AssertThatObject(t *testing.T, objectAssert InPlaceAssertionVerifier) {
...
+
+						if strings.HasSuffix(attr, "%") || strings.HasSuffix(attr, "#") {
+							continue
+						}
+
+						attrParts := strings.SplitN(attr, ".", 2)
+						index, indexErr := strconv.Atoi(attrParts[0])
+						isIndex := indexErr == nil
+
+						if len(attrParts) > 1 && isIndex {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: No, it should work for both nested and simple types

pkg/acceptance/bettertestspoc/assert/commons.go
```diff
@@ -101,3 +105,85 @@ func AssertThatObject(t *testing.T, objectAssert InPlaceAssertionVerifier) {
...
+		}
+
+		errs := make([]error, 0)
+		for _, actualItem := range actualItems {
+			found := false
+			for _, expectedItem := range expectedItems {
+				if maps.Equal(actualItem, expectedItem) {
+					found = true
+				}
+			}
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

pkg/acceptance/bettertestspoc/assert/commons.go
```diff
@@ -101,3 +105,85 @@ func AssertThatObject(t *testing.T, objectAssert InPlaceAssertionVerifier) {
...
+	return func(state *terraform.State) error {
+		var actualItems []map[string]string
+
+		// Allocate space for actualItems and assert length
+		for key, value := range state.RootModule().Resources {
+			if resourceKey == key {
+				for attrKey, attrValue := range value.Primary.Attributes {
+					if strings.HasPrefix(attrKey, attributePath) {
+						attr := strings.TrimPrefix(attrKey, attributePath+".")
+
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm I think it's hard to extract from this code anything useful for other functions. The nesting was lowered a bit, maybe it's ok now?

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+		UpdateContext: UpdateObjectRenamingListsAndSets,
+		ReadContext:   ReadObjectRenamingListsAndSets,
+		DeleteContext: DeleteObjectRenamingListsAndSets,
+
+		Schema: objectRenamingListsAndSetsSchema,
+	}
+}
+
+func CreateObjectRenamingListsAndSets(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	ObjectRenamingDatabaseInstance.List = collections.Map(d.Get("list").([]any), func(item any) objectRenamingDatabaseListItem {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+		}
+
+		ObjectRenamingDatabaseInstance.ManuallyOrderedList = finalState
+	}
+
+	return ReadObjectRenamingListsAndSets(ctx, d, meta)
+}
+
+func ReadObjectRenamingListsAndSets(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	list := collections.Map(ObjectRenamingDatabaseInstance.List, func(t objectRenamingDatabaseListItem) map[string]any {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Currently, It would be only used in the orderedList read. I don't dislike the idea, I was thinking about adding it, but maybe next pr or wait for the other use case for it (it's a 2-3 lines of improvement right now).

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,518 @@
...
+}
+
+func ignoreListOrderAfterFirstApply(parentKey string) schema.SchemaDiffSuppressFunc {
+	return func(key string, oldValue string, newValue string, d *schema.ResourceData) bool {
+		if strings.HasSuffix(key, ".#") {
+			return false
+		}
+
+		// Raw state is not null after first apply
+		if !d.GetRawState().IsNull() && oldValue != newValue {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yes, removed

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
sfc-gh-jcieslak: Added a note, it's not supported because same items result in same hashes and multiple hashes of the same values are not yet supported in this approach.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
sfc-gh-jcieslak: As discussed, added type and tested all the supported and unsupported cases we talked about.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
sfc-gh-jcieslak: 👍 changed in the new test

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,463 @@
...
+					{"name": "name6", "order": 4},
+				}),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_object_renaming.test", "manually_ordered_list.0.name", "name2"),
+					resource.TestCheckResourceAttr("snowflake_object_renaming.test", "manually_ordered_list.1.name", "name4"),
+					resource.TestCheckResourceAttr("snowflake_object_renaming.test", "manually_ordered_list.2.name", "name5"),
+					resource.TestCheckResourceAttr("snowflake_object_renaming.test", "manually_ordered_list.3.name", "name6"),
+				),
+			},
+			// Removed externally
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Added separated tests for those

pkg/acceptance/bettertestspoc/assert/commons.go
```diff
@@ -101,3 +105,85 @@ func AssertThatObject(t *testing.T, objectAssert InPlaceAssertionVerifier) {
...
+								actualItems[i] = make(map[string]string)
+							}
+						}
+					}
+				}
+			}
+		}
+
+		// Gather all actual items
+		for key, value := range state.RootModule().Resources {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: okay, let's leave it for the future

pkg/acceptance/bettertestspoc/assert/commons.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: It would be great to include this in tests.

pkg/acceptance/bettertestspoc/assert/commons.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: Nit: still we can merge these two ifs, the same below.

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -248,91 +314,196 @@ func UpdateObjectRenamingListsAndSets(ctx context.Context, d *schema.ResourceDat
...
-	}
 
-	manuallyOrderedList := make([]map[string]any, len(ObjectRenamingDatabaseInstance.ManuallyOrderedList))
-	for index, item := range ObjectRenamingDatabaseInstance.ManuallyOrderedList {
-		manuallyOrderedList[index] = map[string]any{
-			"name":  item.Name,
-			"order": index + 1, // We start from 1, because there may be some implications with 0 and the fact it's treated as unset value in Terraform
+		itemAdded := len(ObjectRenamingDatabaseInstance.ManuallyOrderedList) > len(d.Get("manually_ordered_list").([]any))
+		itemRemoved := len(ObjectRenamingDatabaseInstance.ManuallyOrderedList) < len(d.Get("manually_ordered_list").([]any))
+		if withExternalChangesMarking && d.Get("manually_ordered_list") != nil && itemAdded || itemRemoved {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, I think it would resolve the expression the same in this case, but yeah. I'll make all the adjustments after @sfc-gh-asawicki review.

sfc-gh-asawicki (dismissed):

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -143,14 +175,45 @@ var objectRenamingListsAndSetsSchema = map[string]*schema.Schema{
...
+						}
+						return "", fmt.Errorf("invalid type: %s", value)
+					}),
+				},
 			},
 		},
 	},
+	"invalid_operation": {
+		Type:     schema.TypeString,
+		Computed: true,
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Added

pkg/resources/object_renaming_lists_and_sets.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

pkg/acceptance/bettertestspoc/assert/commons.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Not sure if that should be done here, I'll add a note in the assert generator read me to utilize this function and test it more thoroughly.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,1218 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_BasicListFlow(t *testing.T) {
+	_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)
+	_ = testenvs.GetOrSkipTest(t, testenvs.EnableObjectRenamingTest)
+	acc.TestAccPreCheck(t)
+
+	// TODO: Add planchecks
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: No, I already applied more planchecks and missed that somehow when removing todos 🤦. Removed.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,1218 @@
...
+					assert.ContainsExactlyInAnyOrder("snowflake_object_renaming.test", "list", []map[string]string{
+						{"string": "111", "int": "111"},
+						{"string": "222", "int": "222"},
+						{"string": "333", "int": "333"},
+					}),
+				),
+			},
+			// Introduce duplicates
+			{
+				Config: objectRenamingConfigList([]map[string]any{
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yes, added a comment in the test case desc

pkg/resources/object_renaming_lists_and_sets.go
```diff
@@ -0,0 +1,694 @@
...
+		oldListMapped := mapObjectRenamingDatabaseListItemFromValue(oldList)
+		newListMapped := mapObjectRenamingDatabaseListItemFromValue(newList)
+
+		addedItems, removedItems := ListDiff(oldListMapped, newListMapped)
+
+		for _, removedItem := range removedItems {
+			ObjectRenamingDatabaseInstance.List = slices.DeleteFunc(ObjectRenamingDatabaseInstance.List, func(item ObjectRenamingDatabaseListItem) bool {
+				shouldRemove := item == removedItem
+				if shouldRemove {
+					ObjectRenamingDatabaseInstance.ChangeLog.Removed = append(ObjectRenamingDatabaseInstance.ChangeLog.Removed, map[string]any{
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: 👍 I'll add this right after merging this one.

pkg/resources/object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -0,0 +1,1218 @@
...
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: objectRenamingConfigManuallyOrderedList([]map[string]any{
+					{"name": "nameTwo", "type": "STRING", "order": 20},
+					{"name": "nameOne", "type": "TEXT", "order": 10},
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I could use the structs used in the implementation and convert them underneath the config function. Next pr.

sfc-gh-jmichalak (approved):

PR #3152, opened 2024-10-23
feat: Rework provider configuration fields

sfc-gh-jmichalak:
- extract env names
- add an entry to migration guide
- add tests for invalid configurations
- improve validations

## Test Plan

* [x] acceptance tests

* [ ] …

## Open topics
- Do we want to move doc helpers to a common package? We could reuse some functions with the `resource` package.

## TODO
- add new fields from the doc
- more advanced testing will be done in tickets for auth flows and for config hierarchy
- use helpers to get values from the schema


pkg/provider/provider.go
```diff
@@ -600,8 +599,16 @@ func ConfigureProvider(s *schema.ResourceData) (interface{}, error) {
 		config.Region = v.(string)
 	}
 
-	if v, ok := s.GetOk("validate_default_parameters"); ok && v.(bool) {
-		config.ValidateDefaultParameters = gosnowflake.ConfigBoolTrue
+	if v := s.Get("validate_default_parameters").(string); v != resources.BooleanDefault {
+		parsed, err := resources.BooleanStringToBool(v)
```
sfc-gh-asawicki, line 603: IMO we shouldn't depend on the resources package here. Please extract the special values and than we can reuse them in both packages (and datasources too if needed).

pkg/internal/snowflakeenvs/snowflake_environment_variables.go
```diff
@@ -1,12 +1,48 @@
 package snowflakeenvs
 
 const (
-	Account    = "SNOWFLAKE_ACCOUNT"
-	User       = "SNOWFLAKE_USER"
-	Password   = "SNOWFLAKE_PASSWORD"
-	Role       = "SNOWFLAKE_ROLE"
+	Account                   = "SNOWFLAKE_ACCOUNT"
```
sfc-gh-asawicki, line 4: +1

pkg/provider/provider_helpers.go
```diff
@@ -13,21 +13,83 @@ import (
...
+	authenticationTypeSnowflake,
+	authenticationTypeOauth,
+	authenticationTypeExternalBrowser,
+	authenticationTypeOkta,
+	authenticationTypeJwt,
+	authenticationTypeTokenAccessor,
+	authenticationTypeUsernamePasswordMfa,
+}
+
+func toAuthenticatorType(s string) (gosnowflake.AuthType, error) {
```
sfc-gh-asawicki, line 49: let's unit test this (like the ones in the SDK)

pkg/provider/provider_helpers.go
```diff
@@ -187,3 +201,29 @@ func GetAccessTokenWithRefreshToken(
...
+		_, err := normalize(val.(string))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		return nil
+	}
+}
+
+// validateBooleanString is similar to the one we use in resources, but in the provider config we also need to explicitly allow default value
+var validateBooleanString = resources.StringInSlice([]string{resources.BooleanTrue, resources.BooleanFalse, resources.BooleanDefault}, false)
```
sfc-gh-asawicki, line 229: Let's distinguish these two validation not only in comment but also by name (either by validateBooleanStringInProvider/InResource or validateBooleanSTringWithDefault - other options will be welcome too )

pkg/provider/provider_helpers.go
```diff
@@ -187,3 +201,29 @@ func GetAccessTokenWithRefreshToken(
 	}
 	return result.AccessToken, nil
 }
+
+func envNameFieldDescription(description, envName string) string {
+	return fmt.Sprintf("%s Can also be sourced from the `%s` environment variable.", description, envName)
+}
+
+// TODO(SNOW-1752043 - next pr): move to common with resources
```
sfc-gh-asawicki, line 209: next Pr will touch again the same files, because it will have to be exported after moved to another package, maybe let's move it already. To not alter all the usages in the resources package, you can temporarily create unexported method there, and removal of it can be handled in the subsequent PR

MIGRATION_GUIDE.md
```diff
@@ -9,6 +9,20 @@ across different versions.
 
 ## v0.97.0 ➞ v0.98.0
 
+### *(behavior change)* Provider configuration rework
+On our road to v1, we have decided to rework configuration to address the most common issues. We have created a list of topics we wanted to address before v1. We will prepare an announcement soon. The following subsections describe the things addressed in the v0.98.0.
```
sfc-gh-asawicki, line 13: add here the link to the roadmap entry, this rework was communicated a few months back

MIGRATION_GUIDE.md
```diff
@@ -9,6 +9,20 @@ across different versions.
...
+- `user`, `warehouse`, `role` - added a validation for an account object identifier
+- `validate_default_parameters`, `client_request_mfa_token`, `client_store_temporary_credential`, `ocsp_fail_open`,  - to easily handle three-value logic (true, false, unknown) in provider's config, type of these fields was changed from boolean to string. For more details about default values, please refer to the [changes before v1](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/v1-preparations/CHANGES_BEFORE_V1.md#default-values) document.
+- `client_ip` - added a validation for an IP address
+- `port` - added a validation for a port number
+- `okta_url`, `token_accessor.token_endpoint`, `client_store_temporary_credential` - added a validation for a URL address
+- `login_timeout`, `request_timeout`, `jwt_expire_timeout`, `client_timeout`, `jwt_client_timeout`, `external_browser_timeout` - added a validation for setting this value to at least `0`
+- `authenticator` - added a possibility to configure JWT flow with `SNOWFLAKE_JWT`; the previous value `JWT` was left for compatibility
```
sfc-gh-asawicki, line 24: please add that it will be removed with the v1

pkg/provider/provider.go
```diff
@@ -223,105 +219,106 @@ func Provider() *schema.Provider {
...
+				Type:             schema.TypeString,
+				Description:      envNameFieldDescription("When true the MFA token is cached in the credential manager. True by default in Windows/OSX. False for Linux.", snowflakeenvs.ClientRequestMfaToken),
+				Optional:         true,
+				DefaultFunc:      schema.EnvDefaultFunc(snowflakeenvs.ClientRequestMfaToken, resources.BooleanDefault),
+				ValidateDiagFunc: validateBooleanString,
 			},
 			"client_store_temporary_credential": {
-				Type:        schema.TypeBool,
-				Description: "When true the ID token is cached in the credential manager. True by default in Windows/OSX. False for Linux. Can also be sourced from the `SNOWFLAKE_CLIENT_STORE_TEMPORARY_CREDENTIAL` environment variable.",
+				Type:        schema.TypeString,
```
sfc-gh-asawicki, line 296: if the validations works differently for provider and resource configs, is this type change safe for the provider config? (it was safe for the resource)

pkg/provider/provider.go
```diff
@@ -558,14 +555,14 @@ var (
 	configureClientError error //nolint:errname
 )
 
-func ConfigureProvider(s *schema.ResourceData) (interface{}, error) {
+func ConfigureProvider(ctx context.Context, s *schema.ResourceData) (any, diag.Diagnostics) {
 	// hacky way to speed up our acceptance tests
-	if os.Getenv("TF_ACC") != "" && os.Getenv("SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE") == "true" {
+	if os.Getenv(resource.EnvTfAcc) != "" && os.Getenv(string(testenvs.ConfigureClientOnce)) == "true" {
```
sfc-gh-asawicki, line 560: we should not have the dependecy to testenvs from this production code (that's the reason for which it was kept like this).

Also the dependency to github.com/hashicorp/terraform-plugin-testing/helper/resource" should be removed.

pkg/provider/provider.go
```diff
@@ -734,25 +753,34 @@ func ConfigureProvider(s *schema.ResourceData) (interface{}, error) {
 		config.DisableTelemetry = v.(bool)
 	}
 
-	if v, ok := s.GetOk("client_request_mfa_token"); ok && v.(bool) {
-		config.ClientRequestMfaToken = gosnowflake.ConfigBoolTrue
+	if v := s.Get("client_request_mfa_token").(string); v != resources.BooleanDefault {
```
sfc-gh-asawicki, line 756: nit: for the next PR (or even the next one); we already have some helpers in the resources that allow as making such blocks of code much shorter and more readable, it does not differ on the conceptual level from what happens in the resources' create methods. It would be great to extract such methods and reuse them. This logic will be much more compact and much less repetitive then

pkg/provider/provider_acceptance_test.go
```diff
@@ -154,6 +154,79 @@ func TestAcc_Provider_configureClientOnceSwitching(t *testing.T) {
...
+			{
+				Config:      providerConfigWithRole(testprofiles.Default, nonExisting),
+				ExpectError: regexp.MustCompile("Role 'NON-EXISTENT' specified in the connect string does not exist or not authorized."),
+			},
+			{
+				Config:      providerConfigWithWarehouse(testprofiles.Default, nonExisting),
+				ExpectError: regexp.MustCompile("The requested warehouse does not exist or not authorized."),
+			},
+			{
+				Config: providerConfigWithWarehouseAndDisabledValidation(testprofiles.Default, nonExisting),
```
sfc-gh-asawicki, line 182: what's the expectation in this last step?

pkg/provider/provider_acceptance_test.go
```diff
@@ -168,6 +241,103 @@ provider "snowflake" {
 `, profile) + datasourceConfig()
 }
 
+func providerConfigWithRole(profile, role string) string {
```
sfc-gh-asawicki, line 244: please add to our model builder generators readme that we should support model creation for the provider (if not added already)

sfc-gh-asawicki (changes requested):

pkg/provider/provider.go
sfc-gh-jcieslak: \+ you forgot to add boolean validator here

pkg/provider/provider.go
```diff
@@ -38,183 +44,173 @@ func init() {
...
 			"okta_url": {
-				Type:        schema.TypeString,
-				Description: "The URL of the Okta server. e.g. https://example.okta.com. Can also be sourced from the `SNOWFLAKE_OKTA_URL` environment variable.",
-				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("SNOWFLAKE_OKTA_URL", nil),
+				Type:             schema.TypeString,
+				Description:      envNameFieldDescription("The URL of the Okta server. e.g. https://example.okta.com.", snowflakeenvs.OktaUrl),
+				Optional:         true,
+				DefaultFunc:      schema.EnvDefaultFunc(snowflakeenvs.OktaUrl, nil),
+				ValidateDiagFunc: validation.ToDiagFunc(validation.IsURLWithHTTPorHTTPS),
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Yeah, we use Capital names instead of UPPER case, but this function is from the plugin SDK ([docs](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.33.0/helper/validation#IsURLWithHTTPorHTTPS)).

pkg/provider/provider_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/provider/provider_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/provider/provider.go
sfc-gh-jmichalak: Done.

MIGRATION_GUIDE.md

MIGRATION_GUIDE.md

pkg/provider/provider_helpers.go
sfc-gh-jmichalak: Done.

pkg/provider/provider_helpers.go
sfc-gh-jmichalak: Done.

pkg/provider/provider.go
sfc-gh-jmichalak: Done.

pkg/helpers/special_values.go
```diff
@@ -0,0 +1,60 @@
+package helpers
```
sfc-gh-asawicki, line 1: We should move away from this package (use internal/*); we will do with a bigger cleanup but as the general rule, but if we are moving them now, let's place them more correctly, e.g. we have:
- internal/provider
- internal/provider/docs
maybe we could add internal/provider/validators for validations (then usage is nicer - validators.ValidateXyz and we won't have to move it again), special values may reside it  internal/provider and helpers/doc can go to internal/provider/docs.

pkg/provider/provider_acceptance_test.go
```diff
@@ -178,13 +178,49 @@ func TestAcc_Provider_useNonExistentDefaultParams(t *testing.T) {
...
+			acc.TestAccPreCheck(t)
+			testenvs.AssertEnvNotSet(t, snowflakeenvs.User)
+			testenvs.AssertEnvNotSet(t, snowflakeenvs.Password)
+		},
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders: map[string]resource.ExternalProvider{
```
sfc-gh-asawicki, line 204: nit: you had in the doc a proposal to extract short func to setup such tests, so maybe let's add this method and use it in the new tests (even without using it in all the places)

sfc-gh-asawicki (changes requested):

pkg/provider/provider_acceptance_test.go
sfc-gh-jmichalak: I can do this in the next pr.

pkg/helpers/special_values.go
sfc-gh-jmichalak: Done.

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

PR #3155, opened 2024-10-24
feat: SDK Connection

sfc-gh-fbudzynski:
## Info
- `ignore edition check` has been removed since I can no longer execute an sql operation with it on snowflake
- `Syntax error: unexpected 'IGNORE'`

## Changes
- connection sdk
- connection snowflake object assertions

## Test Plan

* [x] unit tests
* [x]  integration tests

## TODO
* [uncomment integration tests for failover](https://snowflakecomputing.atlassian.net/browse/SNOW-1763442)
* [Unskip integration tests](https://snowflakecomputing.atlassian.net/browse/SNOW-1002023)
* Resource
* Datasource

## References

 * https://docs.snowflake.com/en/sql-reference/sql/create-connection

PR #3162, opened 2024-10-29
feat: Connection resource

sfc-gh-fbudzynski:
## Changes
* connection resource
* connection documentation
* added connection to migration guide  
## Test Plan

* [x] acceptance tests

## TODO
* [Uncomment acceptance tests for failover](https://snowflakecomputing.atlassian.net/browse/SNOW-1763442)
* [Unskip integration tests](https://snowflakecomputing.atlassian.net/browse/SNOW-1002023)
* Datasource

## References

* https://docs.snowflake.com/en/sql-reference/sql/create-connection
* https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3155


pkg/resources/connection.go
```diff
@@ -0,0 +1,321 @@
...
+		Description:      blocklistedCharactersFieldDescription("String that specifies the identifier (i.e. name) for the connection. Must start with an alphabetic character and may only contain letters, decimal digits (0-9), and underscores (_). For a primary connection, the name must be unique across connection names and account names in the organization. For a secondary connection, the name must match the name of its primary connection."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"as_replica_of": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ForceNew:         true,
+		Description:      "Specifies the identifier for a primary connection from which to create a replica (i.e. a secondary connection).",
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
```
sfc-gh-fbudzynski, line 31: Do we want to have ForceNew here?

pkg/resources/connection.go
```diff
@@ -0,0 +1,321 @@
...
+					Severity: diag.Error,
+					Summary:  "Failed to retrieve connection.",
+					Detail:   fmt.Sprintf("Connection name: %s, Err: %s", id.FullyQualifiedName(), err),
+				},
+			}
+		}
+
+		if withExternalChangesMarking {
+			if err := handleExternalChangesToObjectInShow(d,
+				outputMapping{"is_primary", "is_primary", connection.IsPrimary, connection.IsPrimary, nil},
```
sfc-gh-fbudzynski, line 166: Not sure if this is needed

examples/resources/snowflake_connection/resource.tf
```diff
@@ -0,0 +1,37 @@
+## Minimal
+resource "snowflake_connection" "basic" {
+  name = "connection_name"
+}
+
+## Enable failover to accounts
+resource "snowflake_connection" "with_enable_failover_list" {
```
sfc-gh-jmichalak, line 7: Nit: I don't think we need this one. This additional field is already covered below.

pkg/acceptance/bettertestspoc/assert/resourceassert/connection_resource_ext.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func (c *ConnectionResourceAssert) HasAsReplicaOfIdentifier(expected sdk.ExternalObjectIdentifier) *ConnectionResourceAssert {
+	expectedString := strings.ReplaceAll(expected.FullyQualifiedName(), `"`, "")
+	c.AddAssertion(assert.ValueSet("as_replica_of", expectedString))
+	return c
+}
+
+func (c *ConnectionResourceAssert) HasEnableFailoverToAccounts(expected ...sdk.AccountIdentifier) *ConnectionResourceAssert {
```
sfc-gh-jmichalak, line 17: In case of comparing lists, we should assert their length first.

pkg/sdk/connections_dto_gen.go
```diff
@@ -38,11 +38,11 @@ type ToAccountsRequest struct {
 	Accounts []AccountIdentifier
 }
 
-type SetRequest struct {
+type ConnectionSetRequest struct {
```
sfc-gh-jmichalak, line 41: This change looks good, but it's missing from `def` file. Was this file not generated properly, or `def` file should be adjusted?

pkg/schemas/connection_gen.go
```diff
@@ -0,0 +1,86 @@
...
+	connectionSchema["snowflake_region"] = connection.SnowflakeRegion
+	connectionSchema["created_on"] = connection.CreatedOn.String()
+	connectionSchema["account_name"] = connection.AccountName
+	connectionSchema["name"] = connection.Name
+	if connection.Comment != nil {
+		connectionSchema["comment"] = connection.Comment
+	}
+	connectionSchema["is_primary"] = connection.IsPrimary
+	connectionSchema["primary"] = connection.Primary
+	connectionSchema["failover_allowed_to_accounts"] = connection.FailoverAllowedToAccounts
```
sfc-gh-jmichalak, line 79: Hmm, shouldn't we map to the accounts' fully qualified names here?

pkg/resources/connection.go
```diff
@@ -0,0 +1,313 @@
...
+		Type:         schema.TypeBool,
+		Optional:     true,
+		RequiredWith: []string{"as_replica_of"},
+	},
+	"enable_failover_to_accounts": {
+		Type:        schema.TypeList,
+		Optional:    true,
+		Description: "Enables failover for given connection to provided accounts. Specifies a list of accounts in your organization where a secondary connection for this primary connection can be promoted to serve as the primary connection. Include your organization name for each account in the list.",
+		MinItems:    1,
+		Elem:        &schema.Schema{Type: schema.TypeString},
```
sfc-gh-jmichalak, line 41: I think we should have `DiffSuppressFunc: suppressIdentifierQuoting` in elem.

pkg/resources/connection.go
```diff
@@ -0,0 +1,313 @@
...
+		enableFailoverToAccountsList := make([]sdk.AccountIdentifier, 0)
+		for _, enableToAccount := range enableFailoverConfig {
+			accountInConfig := enableToAccount.(string)
+			accountIdentifier := sdk.NewAccountIdentifierFromFullyQualifiedName(accountInConfig)
+
+			enableFailoverToAccountsList = append(enableFailoverToAccountsList, accountIdentifier)
+		}
+
+		err := client.Connections.Alter(ctx, sdk.NewAlterConnectionRequest(id).
+			WithEnableConnectionFailover(*sdk.NewEnableConnectionFailoverRequest().
```
sfc-gh-jmichalak, line 123: It regards the sdk, but shouldn't the account list be required in `NewEnableConnectionFailoverRequest`? Having empty `ENABLE FAILOVER TO ACCOUNTS` doesn't make sense.

pkg/resources/connection.go
```diff
@@ -0,0 +1,313 @@
...
+		}
+		currentAccountIdentifier := sdk.NewAccountIdentifier(sessionDetails.OrganizationName, sessionDetails.AccountName)
+
+		enableFailoverToAccounts := make([]string, 0)
+		for _, allowedAccount := range connection.FailoverAllowedToAccounts {
+			allowedAccountIdentifier := sdk.NewAccountIdentifierFromFullyQualifiedName(allowedAccount)
+			if currentAccountIdentifier.FullyQualifiedName() == allowedAccountIdentifier.FullyQualifiedName() {
+				continue
+			}
+			enableFailoverToAccounts = append(enableFailoverToAccounts, allowedAccountIdentifier.Name())
```
sfc-gh-jmichalak, line 193: We're not saving the current account to state, which is ok, but what if it's present in the config?

pkg/resources/connection.go
```diff
@@ -0,0 +1,313 @@
...
+			)
+			if err != nil {
+				return diag.FromErr(err)
+			}
+		}
+	}
+
+	if d.HasChange("is_primary") {
+		if is_primary := d.Get("is_primary").(bool); is_primary {
+			err := client.Connections.Alter(ctx, sdk.NewAlterConnectionRequest(id).WithPrimary(is_primary))
```
sfc-gh-jmichalak, line 267: When `is_primary` is removed or set to false, this branch won't execute. We should handle this other case (maybe with forcenew, since we don't have an option to revert `PRIMARY` in `ALTER`).

You can use our stringbools, take a look at handling `is_secret` in https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/pkg/resources/view.go

pkg/resources/connection_acceptance_test.go
```diff
@@ -0,0 +1,421 @@
...
+				),
+			},
+		},
+	})
+}
+
+// TODO: [SNOW-1763442]
+/*
+func TestAcc_Connection_ExternalChangeToIsPrimary(t *testing.T) {
+	// TODO: [SNOW-1002023]: Unskip; Business Critical Snowflake Edition needed
```
sfc-gh-jmichalak, line 326: We should uncomment this test - it will be skipped anyway.

pkg/resources/connection.go
```diff
@@ -0,0 +1,313 @@
...
+	},
+	FullyQualifiedNameAttributeName: schemas.FullyQualifiedNameSchema,
+}
+
+func Connection() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: CreateContextConnection,
+		ReadContext:   ReadContextConnection(true),
+		UpdateContext: UpdateContextConnection,
+		DeleteContext: DeleteContextConnection,
```
sfc-gh-jmichalak, line 64: CustomDiff is missing.

sfc-gh-jmichalak (changes requested):

pkg/resources/connection_acceptance_test.go
```diff
@@ -0,0 +1,421 @@
...
+
+	getConnectionUrl := func(organizationName, objectName string) string {
+		return strings.ToLower(fmt.Sprintf("%s-%s.snowflakecomputing.com", organizationName, objectName))
+	}
+
+	connectionModel := model.Connection("t", id.Name())
+	connectionModelWithComment := model.Connection("t", id.Name()).WithComment(comment)
+
+	// TODO: [SNOW-1763442]
+	/*
```
sfc-gh-jcieslak, line 45: Hmm maybe this could be extracted to another test, it would improve the readability util we'll take care of it in the linked tickets.

pkg/resources/connection.go
sfc-gh-fbudzynski: As far as i understand the documentation correctly, replicated connection can be promoted to serve as primary connection, which essentially (to me) means that it can create a chain of replications

I agree that the best solution would be to somehow transfer secondary to primary connection, but how can this be achieved? Thats why i've decided to leave it as one resource, since secondary connection can be promoted to and serve as primary

There is no good solution for now, since connection as one resource has more responsibility and its state (primary or secondary) is dependent on `is_primary`. On the other hand how can we transfer secondary_connection resource to primary one in a way that won't be confusing to users?

examples/resources/snowflake_connection/resource.tf
sfc-gh-fbudzynski: true

pkg/sdk/connections_dto_gen.go
sfc-gh-fbudzynski: def should be adjusted

pkg/schemas/connection_gen.go
sfc-gh-fbudzynski: i don't think this field in showoutput is used anywhere, but it wouldn't be a bad idea to have them listed as accounts' fully qualified names

pkg/resources/connection.go
sfc-gh-fbudzynski: +1

pkg/resources/connection.go
sfc-gh-fbudzynski: it could but there is a disadvantage to this approach, meaning `is_primary` would have to be required.
For now it is optional and provided only with `as_replica_of`. So when we try to set it with `d.Set("is_primary", connection.IsPrimary)` TF will perform action changing `is_primary = null` to `is_primary = true`

@sfc-gh-jmichalak mentioned [here](https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3162#discussion_r1822189178) that maybe it should be provided to both replicated and primary connection? wdyt?

pkg/acceptance/bettertestspoc/config/model/connection_model_ext.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"strings"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/config"
+	tfconfig "github.com/hashicorp/terraform-plugin-testing/config"
+)
+
+func (c *ConnectionModel) WithAsReplicaOfIdentifier(asReplicaOf sdk.ExternalObjectIdentifier) *ConnectionModel {
+	asReplicaOfString := strings.ReplaceAll(asReplicaOf.FullyQualifiedName(), `"`, "")
```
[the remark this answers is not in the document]
sfc-gh-fbudzynski: it is for tests with replication, i will uncomment them since those test are skipped either way as @sfc-gh-jmichalak mentioned

pkg/resources/connection_acceptance_test.go
sfc-gh-fbudzynski: will do

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/connection_show_output_ext.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func (c *ConnectionShowOutputAssert) HasPrimaryIdentifier(expected sdk.ExternalObjectIdentifier) *ConnectionShowOutputAssert {
+	expectedString := strings.ReplaceAll(expected.FullyQualifiedName(), `"`, "")
```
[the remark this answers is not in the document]
sfc-gh-fbudzynski: problem here is that snowflake holds the values as `org_name.account_name.connection_name` and using FullyQualifiedName() on ExternalObjectIdentifier results in: `"org_name"."account_name"."connection_name"` (mind the quotes)

pkg/resources/connection.go
sfc-gh-asawicki: I have answered to the thread internally; IMO until we clarify the behavior on SF side, let's keep to separate resources and the promotion would mean dropping and recreating (not by forceNew but removal of old resource in config and adding the new one).

Of course, we have to also then react to external type changes (after split).

pkg/resources/connection.go
sfc-gh-asawicki: Unless this is the way to clean it?

docs/resources/connection.md
```diff
@@ -0,0 +1,102 @@
...
+  name          = "connection_name"
+  as_replica_of = "<organization_name>.<account_name>.<connection_name>"
+  is_primary    = false
+}
+
+# As replica with promotion to primary
+resource "snowflake_connection" "replica_with_promotion" {
+  name          = "connection_name"
+  as_replica_of = "<organization_name>.<account_name>.<connection_name>"
+  is_primary    = true
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I vote for `snowflake_primary_connection` and `snowflake_secondary_connection` resources

pkg/acceptance/bettertestspoc/assert/resourceassert/connection_resource_ext.go
```diff
@@ -0,0 +1,27 @@
...
+	for i, v := range expected {
+		c.AddAssertion(assert.ValueSet(fmt.Sprintf("enable_failover_to_accounts.%d", i), v.Name()))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: note: we have this handled inconsistently. IMO we should have a clear separation in method naming for collection assertions. My proposal:
- HasExactlyXInOrder: we should check for the length and each elem in the given order
- HasExactlyX: we should check for the length and each elem but in any order
- HasX: we shouldn't care about the length and order

WDYT? I will add it to the assertions cleanup but we can discuss already. My comment is not blocking.

pkg/resources/connection.go
```diff
@@ -0,0 +1,313 @@
...
+	"is_primary": {
+		Type:         schema.TypeBool,
+		Optional:     true,
+		RequiredWith: []string{"as_replica_of"},
+	},
+	"enable_failover_to_accounts": {
+		Type:        schema.TypeList,
+		Optional:    true,
+		Description: "Enables failover for given connection to provided accounts. Specifies a list of accounts in your organization where a secondary connection for this primary connection can be promoted to serve as the primary connection. Include your organization name for each account in the list.",
+		MinItems:    1,
```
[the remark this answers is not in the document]
sfc-gh-asawicki: It worked =, we had test for it but I am not sure if we still have it. AFAIR it was for secondary roles in user which implementation was changed from list to enum. We can still just add validation test here and check if the validation works or not.

pkg/resources/connection.go
sfc-gh-fbudzynski: to clean it there is a `alter connection <conn_name> disable failover`

pkg/resources/connection.go
sfc-gh-fbudzynski: What if the promotion of secondary connection to primary has been made externally? How should we react? as i understand there is no way to demote the primary back to secondary unless its "parent" removes the account that created secondary connection for `enable failover to accounts` list 

One option i can think of is that when we detect external change of `show_output.is_primary` from false to true, then we alter the parent of secondary connection and remove the account which created it from the list

pkg/sdk/connections_impl_gen.go
```diff
@@ -119,18 +115,41 @@ func (r *ShowConnectionRequest) toOpts() *ShowConnectionOptions {
...
+	} else {
+		c.Primary = primaryExternalId
+	}
+
+	allowedToAccounts := make([]AccountIdentifier, 0)
+	allowedAccountsString := ParseCommaSeparatedStringArray(r.FailoverAllowedToAccounts, false)
+	for _, accountString := range allowedAccountsString {
+		accountId, err := ParseAccountIdentifier(accountString)
+		if err != nil {
+			log.Printf("unable to print account identifier for 'enable failover to accounts' account %s, err = %v", accountString, err)
```
sfc-gh-jcieslak, line 146: unable to **parse**

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/connection_show_output_ext.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Even special characters are not affecting the output ? e.g. having a connection with name that contains `@`? if it's still unquoted, then it's strange.

MIGRATION_GUIDE.md
```diff
@@ -9,20 +9,60 @@ across different versions.
 
 ## v0.97.0 ➞ v0.98.0
 
-### *(new feature)* connection resource
-Added a new resource for managing connections.
+### _(new feature)_ connection resources
```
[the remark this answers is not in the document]
sfc-gh-fbudzynski: no clue either, i used pre-push and additionally linter on its own to check whether all is as it should. I will change it back to `*` but i have no clue what it has been changed. Maybe because of different formatting tool (?)

pkg/acceptance/bettertestspoc/assert/resourceassert/connection_resource_ext.go
```diff
@@ -2,19 +2,21 @@ package resourceassert
...
 import (
 	"fmt"
-	"strings"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 )
 
+/*
 func (c *ConnectionResourceAssert) HasAsReplicaOfIdentifier(expected sdk.ExternalObjectIdentifier) *ConnectionResourceAssert {
```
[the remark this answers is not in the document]
sfc-gh-fbudzynski: if it is not used it will be removed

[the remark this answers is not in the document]
sfc-gh-fbudzynski: It will be removed since the connection has been split into two resources it has been moved to `resourceassert/secondary_connection_ext.go`

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/connection_show_output_ext.go
```diff
@@ -2,19 +2,19 @@ package resourceshowoutputassert
...
-	"strings"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 )
 
 func (c *ConnectionShowOutputAssert) HasPrimaryIdentifier(expected sdk.ExternalObjectIdentifier) *ConnectionShowOutputAssert {
-	expectedString := strings.ReplaceAll(expected.FullyQualifiedName(), `"`, "")
-	c.AddAssertion(assert.ResourceShowOutputValueSet("primary", expectedString))
+	// expectedString := strings.ReplaceAll(expected.FullyQualifiedName(), `"`, "")
```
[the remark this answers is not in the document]
sfc-gh-fbudzynski: will do

pkg/sdk/connections_gen_test.go
```diff
@@ -47,7 +50,7 @@ func TestConnections_Create(t *testing.T) {
 		opts := defaultOpts()
 		opts.name = id
 		opts.AsReplicaOf = &externalId
-		assertOptsValidAndSQLEquals(t, opts, "CREATE CONNECTION %s AS REPLICA OF %s", id.FullyQualifiedName(), externalId.FullyQualifiedName())
+		assertOptsValidAndSQLEquals(t, opts, "CREATE CONNECTION %s AS REPLICA OF %s", id.FullyQualifiedName(), strings.ReplaceAll(externalId.FullyQualifiedName(), `"`, ""))
```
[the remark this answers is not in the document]
sfc-gh-fbudzynski: yeah there is no need to use it here, i use it to check whether there would be difference between externalId with FullyQualifiedName and with removed quotes

pkg/resources/connection.go
sfc-gh-fbudzynski: after splitting i've added a custom diff on secondary_connection resource, which detects if the new value has been changed externally to `true` if so, we should re-create the secondary_connection to still serve as secondary, not primary one.

pkg/resources/connection.go
sfc-gh-asawicki: We should force recreation if there is no suitable alter (maybe let's handle external promotion in a separate PR?)

pkg/acceptance/bettertestspoc/assert/resourceassert/connection_resource_ext.go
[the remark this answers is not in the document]
sfc-gh-asawicki: I meant that we should have three families of functions and use them depending on the use-case

MIGRATION_GUIDE.md
```diff
@@ -9,20 +9,60 @@ across different versions.
...
+
+Added a new resources for managing connections. We decided to split connection into two separate resources based on whether the connection is primary or a replica (secondary). i.e.:
+
+- `snowflake_connection` is used as primary connection, with ability to enable failover to other accounts.
+- `snowflake_secondary_connection` is used as replica (secondary) connection.
+
+In order to promote a secondary connection to primary, `snowflake_secondary_connection` should be removed from the config, and new `snowflake_connection` resource should be created with the same name.
```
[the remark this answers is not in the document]
sfc-gh-asawicki: +1, let's suggest the resource migration (they can alter manually and import the promoted one - less destructive) as an alternative - they both have pros and cons - resource migration is manual but less destructive while removal and addition to config is straightforward but destructive

pkg/resources/connection.go
```diff
@@ -0,0 +1,266 @@
...
+
+func Connection() *schema.Resource {
+	return &schema.Resource{
+		CreateContext: CreateContextConnection,
+		ReadContext:   ReadContextConnection,
+		UpdateContext: UpdateContextConnection,
+		DeleteContext: DeleteContextConnection,
+
+		CustomizeDiff: customdiff.All(
+			ComputedIfAnyAttributeChanged(connectionSchema, ShowOutputAttributeName, "comment", "primary", "failover_allowed_to_accounts"),
```
sfc-gh-jmichalak, line 59: There is no `primary` field.

MIGRATION_GUIDE.md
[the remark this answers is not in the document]
sfc-gh-asawicki: still, please validate such changes before the re-review

pkg/resources/connection.go
sfc-gh-fbudzynski: talking only about the `failover_to_accounts` list, the current account (which has created that connection) is always present in the list at snowflake. We don't want to always specify that the connection itself is in the list. Although if it is provided in the config, it will fail with snowflake error `The connection cannot be failed over to an account in the same region` We could prevent that from happening by adding the same if to Create method or we could enforce that this list should be required and always filled with at least the account of the creator of connection

pkg/resources/secondary_connection.go
```diff
@@ -0,0 +1,190 @@
...
+	},
+	"comment": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies a comment for the secondary connection.",
+	},
+	ShowOutputAttributeName: {
+		Type:        schema.TypeList,
+		Computed:    true,
+		Description: "Outputs the result of `SHOW CONNECTIONS` for the given secret.",
```
sfc-gh-jmichalak, line 45: For the given **connection**. The same in `connection.go`.

pkg/sdk/connections_impl_gen.go
sfc-gh-jmichalak, line 146: We already have `ParseCommaSeparatedSchemaObjectIdentifierArray` for schema object identifiers. We can add a function to parse account identifiers, IMO.

pkg/resources/secondary_connection.go
```diff
@@ -0,0 +1,190 @@
...
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		Description:      blocklistedCharactersFieldDescription("String that specifies the identifier (i.e. name) for the connection. Must start with an alphabetic character and may only contain letters, decimal digits (0-9), and underscores (_). For a secondary connection, the name must match the name of its primary connection."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"is_primary": {
+		Type:        schema.TypeBool,
+		Computed:    true,
+		Description: "Indicates if the connection has been changed to primary. If change is detected, the secondary connection will be recreated.",
```
sfc-gh-jmichalak, line 28: Why is there no `enable_failover_to_accounts`?

pkg/provider/resources/resources.go
```diff
@@ -11,6 +11,8 @@ const (
 	ApiAuthenticationIntegrationWithJwtBearer              resource = "snowflake_api_authentication_integration_with_jwt_bearer"
 	ApiIntegration                                         resource = "snowflake_api_integration"
 	AuthenticationPolicy                                   resource = "snowflake_authentication_policy"
+	Connection                                             resource = "snowflake_connection"
```
sfc-gh-jmichalak, line 14: I think it `snowflake_primary_connection` would be more clear. Same for file names, struct names, etc.

pkg/sdk/connections_impl_gen.go
sfc-gh-fbudzynski: will add one

pkg/resources/secondary_connection.go
sfc-gh-fbudzynski: im not sure if there should be, secondary connection does not serve as primary one, to me it shouldn't have the list of accounts that could replicate it, since it cannot be replicated (unless promoted to primary)

sfc-gh-jmichalak: Okay - I was just asking.

MIGRATION_GUIDE.md
```diff
@@ -9,10 +9,54 @@ across different versions.
...
-### *(new feature)* connection resource
-Added a new resource for managing connections.
+### *(new feature)* connection resources
+
+Added a new resources for managing connections. We decided to split connection into two separate resources based on whether the connection is primary or a replica (secondary). i.e.:
+
+- `snowflake_connection` is used as primary connection, with ability to enable failover to other accounts.
+- `snowflake_secondary_connection` is used as replica (secondary) connection.
+
+In order to promote secondary_connection to primary, resources need to be migrated (check [resource migration](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/docs/technical-documentation/resource_migration.md)) or re-created and imported using the following SQL statements on Snowflake Worksheet:
```
sfc-gh-jcieslak, line 19: This is good, but slightly rephrase that `In order to promote secondary_connection to primary, resources need to be removed from the state, re-created manually using the following SQL statements ... then imported again, but now as snowflake_connection (check resource migration).`

MIGRATION_GUIDE.md
```diff
@@ -9,6 +9,22 @@ across different versions.
 
 ## v0.97.0 ➞ v0.98.0
 
+### *(new feature)* connection resources
+
+Added a new resources for managing connections. We decided to split connection into two separate resources based on whether the connection is primary or a replica (secondary). i.e.:
+
+- `snowflake_connection` is used as primary connection, with ability to enable failover to other accounts.
```
sfc-gh-asawicki, line 16: nit: snowflake_primary_connection

pkg/resources/secondary_connection.go
```diff
@@ -52,14 +52,15 @@ var secondaryConnectionSchema = map[string]*schema.Schema{
...
-		CreateContext: CreateSecondaryContextConnection,
-		ReadContext:   ReadSecondaryContextConnection,
-		UpdateContext: UpdateSecondaryContextConnection,
-		DeleteContext: DeleteSecondaryContextConnection,
-		Description:   "Resource used to manage secondary connections. For more information, check [connection documentation](https://docs.snowflake.com/en/sql-reference/sql/create-connection.html).",
+		CreateContext: CreateContextSecondaryConnection,
+		ReadContext:   ReadContextSecondaryConnection,
+		UpdateContext: UpdateContextSecondaryConnection,
+		DeleteContext: DeleteContextSecondaryConnection,
+		Description:   "Resource used to manage secondary connections. To promote secondary connection to primary check [migraton guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#connection-resources). For more information, check [connection documentation](https://docs.snowflake.com/en/sql-reference/sql/create-connection.html).",
```
sfc-gh-jmichalak, line 59: Hmm I'm not sure if linking to the migration guide is the best option here. Maybe we should just point to the primary resource here, and add a note in template/resources. The same for the other direction. WDYT?

MIGRATION_GUIDE.md
```diff
@@ -9,6 +9,22 @@ across different versions.
...
+- `snowflake_secondary_connection` is used as replica (secondary) connection.
+
+In order to promote secondary_connection to primary, resources need to be migrated (check [resource migration](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/docs/technical-documentation/resource_migration.md)) or re-created and imported using the following SQL statements on Snowflake Worksheet:
```
sfc-gh-asawicki, line 19: nit: this is fine to have this info in the migration guide, but eventually, it would be great to also include this in the resource docs (it's important information for later)

pkg/acceptance/bettertestspoc/assert/objectassert/connection_snowflake_gen.go
```diff
@@ -111,6 +125,19 @@ func (c *ConnectionAssert) HasPrimary(expected string) *ConnectionAssert {
 	return c
 }
 
+/*
```
sfc-gh-asawicki, line 128: nit: why this is commented out?

pkg/resources/custom_diffs.go
```diff
@@ -270,3 +270,15 @@ func RecreateWhenStreamIsStale() schema.CustomizeDiffFunc {
 		return nil
 	}
 }
+
+// TODO: [SNOW-1763442] unable to test now, as there is no test accounts with different regions
+// RecreateWhenSecondaryConnectionChangedExternally detects if the secondary connection was promoted externally to serve as primary.
+// If so, it sets the `is_primary` field to `false` which is our desired value for secondary_connection
+func RecreateWhenSecondaryConnectionPromotedExternally() schema.CustomizeDiffFunc {
```
sfc-gh-asawicki, line 277: nit: we have to handle this the other way around too for primary connection demotion

pkg/resources/custom_diffs_test.go
```diff
@@ -2,11 +2,10 @@ package resources_test
 
 import (
 	"context"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
```
sfc-gh-asawicki, line 5: nit: imports ordering

pkg/resources/custom_diffs_test.go
```diff
@@ -1082,3 +1081,40 @@ func Test_RecreateWhenSecretTypeChangedExternallyForOAuth2(t *testing.T) {
...
+		name              string
+		expectedIsPrimary string
+		stateValue        map[string]string
+	}{
+		{
+			name:              "changed from is_primary from false to true",
+			expectedIsPrimary: "false",
+			stateValue: map[string]string{
+				"is_primary": "true",
+			},
```
sfc-gh-jcieslak, line 1096: Add more tests cases:
- going from true to false
- no changes (it would be probably caught in acc tests, but just in case) true -> true and false -> false

pkg/resources/custom_diffs.go
```diff
@@ -270,3 +270,15 @@ func RecreateWhenStreamIsStale() schema.CustomizeDiffFunc {
...
+	return func(_ context.Context, diff *schema.ResourceDiff, _ any) error {
+		if _, newValue := diff.GetChange("is_primary"); newValue.(bool) {
+			return diff.SetNew("is_primary", false)
```
sfc-gh-asawicki, line 280: hmmm, this logic is not aligned with others (check RecreateWhenUserTypeChangedExternally, RecreateWhenSecretTypeChangedExternally, or RecreateWhenResourceTypeChangedExternally) - there is ForceNew missing - it's a problem but it can be handled in the next PR too

pkg/resources/primary_connection_acceptance_test.go
```diff
@@ -0,0 +1,277 @@
...
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_PrimaryConnection_Basic(t *testing.T) {
+	// TODO: [SNOW-1002023]: Unskip; Business Critical Snowflake Edition needed
+	_ = testenvs.GetOrSkipTest(t, testenvs.TestFailoverGroups)
+
```
sfc-gh-asawicki, line 29: nit: add
```
_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)
acc.TestAccPreCheck(t)
```
to all the tests that use TestClient in their set-up

pkg/sdk/parsers.go
```diff
@@ -52,3 +52,18 @@ func ParseCommaSeparatedSchemaObjectIdentifierArray(value string) ([]SchemaObjec
 	}
 	return ids, nil
 }
+
+// ParseCommaSeparatedAccountIdentifierArray can be used to parse Snowflake output containing a list of account identifiers
+// in the format of ["organization1.account1", "organization2.account2", ...],
+func ParseCommaSeparatedAccountIdentifierArray(value string) ([]AccountIdentifier, error) {
```
sfc-gh-asawicki, line 58: nit: do we have unit test for this?

sfc-gh-asawicki (approved):
For me, all the nits and comments (from my review) can be handled in the next separate PR

pkg/schemas/connection_gen.go
```diff
@@ -36,6 +36,7 @@ var ShowConnectionSchema = map[string]*schema.Schema{
 	"is_primary": {
 		Type:     schema.TypeBool,
 		Computed: true,
+		ForceNew: true,
```
sfc-gh-jcieslak, line 39: Is it necessary? ForceNew behavior is already handled in the resources (primary and secondary connection), right?

pkg/resources/connection.go
sfc-gh-jcieslak: I think you are already handling it correctly, so this can be resolved.

sfc-gh-jmichalak (approved):

pkg/resources/connection.go
sfc-gh-jcieslak: I think this can be resolved.

sfc-gh-jcieslak (approved):
Approved. There are still open threads and new comments that are not crucial and can be applied in the next pr.

pkg/resources/secondary_connection.go
sfc-gh-fbudzynski: agree, description will be more clear, and migration of the resource will be in documentation

pkg/schemas/connection_gen.go
sfc-gh-fbudzynski: yup, will be removed

pkg/resources/custom_diffs_test.go
sfc-gh-fbudzynski: this will be further tested in next PR, as well as `ParseCommaSeparatedAccountIdentifierArray()`

pkg/sdk/parsers.go
sfc-gh-fbudzynski: not yet, i will add them in next PR, as well as tests for the customDiff

PR #3164, opened 2024-10-30
fix: Handle external type changes in stream resources

sfc-gh-jmichalak:
- handle external type changes in stream resources

## Test Plan

* [x] acceptance tests

* [ ] …

### DIFF
--- a/pkg/resources/custom_diffs.go
+++ b/pkg/resources/custom_diffs.go
@@ -228,6 +228,35 @@ func RecreateWhenSecretTypeChangedExternally(secretType sdk.SecretType) schema.C
 	}
 }
 
+// RecreateWhenStreamTypeChangedExternally recreates a stream when argument streamType is different than in the state.
+func RecreateWhenStreamTypeChangedExternally(streamType sdk.StreamSourceType) schema.CustomizeDiffFunc {
+	return RecreateWhenResourceTypeChangedExternally("stream_type", streamType, sdk.ToStreamSourceType)
+}
+
+// RecreateWhenResourceTypeChangedExternally recreates a resource when argument wantType is different than the value in typeField.
+func RecreateWhenResourceTypeChangedExternally[T ~string](typeField string, wantType T, toType func(string) (T, error)) schema.CustomizeDiffFunc {
+	return func(_ context.Context, diff *schema.ResourceDiff, _ interface{}) error {
+		if n := diff.Get(typeField); n != nil {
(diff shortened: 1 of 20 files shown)


pkg/resources/custom_diffs.go
```diff
@@ -228,6 +228,35 @@ func RecreateWhenSecretTypeChangedExternally(secretType sdk.SecretType) schema.C
...
+			// if the type is empty, the state is empty - do not recreate
+			if gotTypeRaw == "" {
+				return nil
+			}
+
+			gotType, err := toType(gotTypeRaw)
+			if err != nil {
+				return fmt.Errorf("unknown type: %w", err)
+			}
+			if gotType != wantType {
```
sfc-gh-asawicki, line 252: please add the same comment as in the user specific func here.

nit: can we reuse this func in user and in secret (secret has some additional logic; user empty type is a valid case AFAIK so differs from this impl - but still, for me it makes sense if we can reuse the function and currently was not reused in other places)?

pkg/resources/stream_on_directory_table_acceptance_test.go
```diff
@@ -459,3 +459,56 @@ func TestAcc_StreamOnDirectoryTable_InvalidConfiguration(t *testing.T) {
...
+				),
+			},
+			// external change with a different type
+			{
+				PreConfig: func() {
+					table, cleanupTable := acc.TestClient().Table.CreateWithChangeTracking(t)
+					t.Cleanup(cleanupTable)
+					acc.TestClient().Stream.DropFunc(t, id)()
+					_, cleanup := acc.TestClient().Stream.CreateOnTableWithRequest(t, sdk.NewCreateOnTableStreamRequest(id, table.ID()))
+					t.Cleanup(cleanup)
```
sfc-gh-asawicki, line 495: nit: let's assert in this preconfig that after it we have other type

sfc-gh-asawicki (dismissed):

pkg/resources/custom_diffs.go
sfc-gh-jmichalak: Added a comment.

Yeah, initially, I thought about reusing this even more, but I decided not to spend more time on it - I can do it in the next PR.

pkg/resources/stream_on_directory_table_acceptance_test.go
sfc-gh-jmichalak: Done.

sfc-gh-jcieslak (approved):

PR #3180, opened 2024-11-07
chore: Unskip auth config tests

sfc-gh-jmichalak:
- add a new v097-compatible config file
- unskip auth related tests
- for older providers, set the config file to be v097-compatible

## Test Plan

* [x] acceptance tests

* [ ] …

## TODO (next pr)
- read from the proper secrets and remove temporary secrets


.github/workflows/tests.yml
```diff
@@ -41,7 +41,11 @@ jobs:
 
       - name: Create and populate .snowflake/config file
         id: create_config
-        run: mkdir $HOME/.snowflake && echo "${{ secrets.SNOWFLAKE_CONFIG_FILE }}" > $HOME/.snowflake/config
+        run: mkdir -p $HOME/.snowflake && echo "${{ secrets.TMP_JMICHALAK_SNOWFLAKE_CONFIG_FILE }}" > $HOME/.snowflake/config
+
+      - name: Create and populate .snowflake/config_v097_compatible file
```
sfc-gh-asawicki, line 46: just marking as to-be-changed before the merge (proper secrets)

pkg/acceptance/testing.go
```diff
@@ -208,3 +210,22 @@ func ExternalProviderWithExactVersion(version string) map[string]resource.Extern
 		},
 	}
 }
+
+// SetV097CompatibleConfigPath sets a new config path in a relevant env variable for a file that is compatible with v0.97.
+func SetV097CompatibleConfigPathEnv(t *testing.T) func() {
```
sfc-gh-asawicki, line 215: It is a bit misleading that function name says `set` but the truth is that this just provides a function to set it. I would either change the name of the function or - more preferably -  do not return the function but inline the body here.

sfc-gh-asawicki (changes requested):
There are 2 tests with external providers that were added in the meantime and failed in your pipeline, please update them too.

pkg/acceptance/testing.go
```diff
@@ -208,3 +210,22 @@ func ExternalProviderWithExactVersion(version string) map[string]resource.Extern
...
+	t.Helper()
+	return func() {
+		home, err := os.UserHomeDir()
+		require.NoError(t, err)
+		configPath := filepath.Join(home, ".snowflake", "config_v097_compatible")
+		t.Setenv(snowflakeenvs.ConfigPath, configPath)
```
sfc-gh-jcieslak, line 221: this will set env only for the test it is invoked in, right?

sfc-gh-jcieslak (dismissed):

pkg/acceptance/testing.go
sfc-gh-jmichalak: Not exactly. According to the [docs](https://pkg.go.dev/testing#T.Setenv):
> Setenv calls os.Setenv(key, value) and uses Cleanup to restore the environment variable to its original value after the test.
>
> Because Setenv affects the whole process, it cannot be used in parallel tests or tests with parallel ancestors.

pkg/acceptance/testing.go
sfc-gh-jmichalak: I changed the names to `...Func`. We can't inline them because they need to take `*testing.T` and return `func()` to satisfy `PreConfig`.

sfc-gh-asawicki: Well... You can wrap it in func in the PreConfig. This is how we do this in the majority use cases throughout our tests. It's easier then to set multiple things in the PreConfig.

sfc-gh-jmichalak: Done.

pkg/resources/network_policy_acceptance_test.go
```diff
@@ -479,7 +479,7 @@ func TestAcc_NetworkPolicy_Issue2236(t *testing.T) {
 				// Identifier quoting mismatch (no diff suppression)
 				ExpectNonEmptyPlan: true,
 				PreConfig: func() {
-					acc.SetV097CompatibleConfigPathEnv(t)()
+					func() { acc.SetV097CompatibleConfigPathEnv(t) }()
```
sfc-gh-asawicki, line 482: nit (next PR): you don't have to wrap here

sfc-gh-asawicki (approved):

sfc-gh-asawicki (changes requested):
integration tests pipeline is failing (please check it)

sfc-gh-jcieslak (dismissed):

PR #3181, opened 2024-11-07
fix: Fix user resource import

sfc-gh-asawicki:
Handle user import correctly.

Password is empty after the `snowflake_user` import; we can't read it from the config or from Snowflake.
During the next terraform plan+apply it's updated to the "same" value.
It results in an error on Snowflake side: `New password rejected by current password policy. Reason: 'PRIOR_USE'.`
The error will be ignored on the provider side (after all, it means that the password in state is the same as on Snowflake side). Still, plan+apply is needed after importing user.

### DIFF
--- a/pkg/resources/user.go
+++ b/pkg/resources/user.go
@@ -512,17 +512,17 @@ func GetReadUserFunc(userType sdk.UserType, withExternalChangesMarking bool) sch
 			// can't read disable_mfa
 			d.Set("user_type", u.Type),
 
-			func() error {
+			func(rd *schema.ResourceData, ud *sdk.UserDetails) error {
 				var errs error
 				if userType == sdk.UserTypePerson {
 					errs = errors.Join(
-						setFromStringPropertyIfNotEmpty(d, "first_name", userDetails.FirstName),
-						setFromStringPropertyIfNotEmpty(d, "middle_name", userDetails.MiddleName),
-						setFromStringPropertyIfNotEmpty(d, "last_name", userDetails.LastName),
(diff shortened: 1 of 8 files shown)


sfc-gh-jcieslak (dismissed):

pkg/resources/user.go
```diff
@@ -677,6 +679,40 @@ func GetUpdateUserFunc(userType sdk.UserType) func(ctx context.Context, d *schem
...
+	if userType == sdk.UserTypePerson || userType == sdk.UserTypeLegacyService {
+		setPassword := sdk.UserAlterObjectProperties{}
+		unsetPassword := sdk.UserObjectPropertiesUnset{}
+		if err := stringAttributeUpdate(d, "password", &setPassword.Password, &unsetPassword.Password); err != nil {
+			return err
+		}
+		if (setPassword != sdk.UserAlterObjectProperties{}) {
+			err := client.Users.Alter(ctx, id, &sdk.AlterUserOptions{Set: &sdk.UserSet{ObjectProperties: &setPassword}})
+			if err != nil {
+				if strings.Contains(err.Error(), "Error: 003002 (28P01)") || strings.Contains(err.Error(), "Reason: 'PRIOR_USE'") {
```
sfc-gh-jmichalak, line 697: I think this should be handled on SDK level, and we could return a custom error from there.

sfc-gh-asawicki: Okay, I can migrate it with the next PR

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (approved):

PR #3202, opened 2024-11-14
feat: Tasks v1 readiness part3

sfc-gh-jcieslak:
- [ ] Apply comments https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3170
- [ ] Apply comments from https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3113
  - [x] Cron https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3113#discussion_r1789842978
  - [ ] Resource logic https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3113#discussion_r1789869045
  - [ ] Refactor SDK suspend root logic for tasks (and overall suspend logic in SDK/resource) https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3113#discussion_r1789872814
- [ ] Move some of the logic to SDK (if possible)
  - [ ] https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3170#discussion_r1838382878
- [ ] Refactor task resuming in task resource (most likely with the use of defer) because currently, there may be cases that error can cause tasks to be not resumed.
- [ ] Tests
  - [ ] External changes - https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3113#discussion_r1789903936
  - [ ] Add more complicated DAG structures to show the resource can handle more complex structures
  - [ ] Calling (`as` field) - https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3113#discussion_r1789900590
  - [ ] For showing how the DAG of tasks could be owned by a role other than the one created with Terraform (also with less privileges, only to run the task).
  - [x] Check in one acceptance test why finalizer task in show_output is not set (is that Snowflake or mapping error).
- [x] Data source
- [ ] Examples, documentation, and migration guide
- [ ] Keep manually changed files after regeneration https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3113#discussion_r1802561725
- [x] Make config without $$ escapes needed
- [ ] Support session paramters
- [ ] Analyze non-deterministic test cases
- [ ] Check test tasks_gen_integration_test.go:937 (and see why it's non deterministic).
- [ ] Re-generate and list all the issues with asserts and models


pkg/resources/task.go
```diff
@@ -197,6 +198,15 @@ func Task() *schema.Resource {
 			ComputedIfAnyAttributeChanged(taskSchema, FullyQualifiedNameAttributeName, "name"),
 			taskParametersCustomDiff,
 		),
+
```
sfc-gh-asawicki, line 201: reviedog

pkg/resources/task.go
```diff
@@ -587,35 +595,40 @@ func ReadTask(withExternalChangesMarking bool) schema.ReadContextFunc {
...
-					cron := task.Schedule[len("USING CRON "):]
-					if err := d.Set("schedule", []any{map[string]any{
-						"using_cron": cron,
-					}}); err != nil {
-						return err
-					}
-				case strings.Contains(upperSchedule, "MINUTE"):
-					minuteParts := strings.Split(upperSchedule, " ")
-					minutes, err := strconv.Atoi(minuteParts[0])
+				managedInitialWarehouseSizeIdx := slices.IndexFunc(taskParameters, func(p *sdk.Parameter) bool {
```
sfc-gh-asawicki, line 598: I still don't get why we need to handle this parameter in other way than the others (e.g. why we do not handle it like the parameters in warehouse resource)

pkg/resources/task_state_upgraders.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func v098TaskStateUpgrader(ctx context.Context, rawState map[string]any, meta any) (map[string]any, error) {
+	if rawState == nil {
+		return rawState, nil
+	}
+
+	rawState["condition"] = rawState["when"]
+	rawState["started"] = booleanStringFromBool(rawState["enabled"].(bool))
```
sfc-gh-asawicki, line 15: what if the previous one was false by default? in other boolean strings too

pkg/resources/task_state_upgraders.go
```diff
@@ -0,0 +1,49 @@
...
+	rawState["allow_overlapping_execution"] = booleanStringFromBool(rawState["allow_overlapping_execution"].(bool))
+	if rawState["after"] != nil {
```
sfc-gh-asawicki, line 17: what if after was not an array for some reason, should we check it before getting len?

pkg/resources/task_state_upgraders.go
```diff
@@ -0,0 +1,49 @@
...
+		newAfter := make([]string, len(rawState["after"].([]any)))
+		for i, name := range rawState["after"].([]any) {
```
sfc-gh-asawicki, line 19: where these before required to be fully qualified in the config or not? If so, we need some migration guide informing about the config changes

pkg/resources/task_state_upgraders.go
```diff
@@ -0,0 +1,49 @@
...
+			newAfter[i] = sdk.NewSchemaObjectIdentifier(rawState["database"].(string), rawState["schema"].(string), name.(string)).FullyQualifiedName()
+		}
+		rawState["after"] = newAfter
+	}
+	if rawState["session_parameters"] != nil {
```
sfc-gh-asawicki, line 24: type check first?

pkg/sdk/tasks_gen_test.go
```diff
@@ -552,3 +554,66 @@ func TestTasks_Execute(t *testing.T) {
 		assertOptsValidAndSQLEquals(t, opts, "EXECUTE TASK %s RETRY LAST", id.FullyQualifiedName())
 	})
 }
+
+func TestParseTaskSchedule(t *testing.T) {
```
sfc-gh-asawicki, line 558: In the comment for this function `America/Los_Angeles` is mentioned and it is not here in any test

pkg/resources/task_acceptance_test.go
```diff
@@ -306,6 +304,303 @@ func TestAcc_Task_Updates(t *testing.T) {
...
+			},
+		},
+	})
+}
+
+/*
+DAG structure (the test proves child3 won't have any issues with updates in the following scenario):
+
+		 child1
+		/		\
```
sfc-gh-asawicki, line 413: nit: it does not look well on GH, does it look good in the IDE?

pkg/resources/task_acceptance_test.go
```diff
@@ -306,6 +304,303 @@ func TestAcc_Task_Updates(t *testing.T) {
...
+			configvariable.StringVariable(child2.ID().FullyQualifiedName()),
+		))
+
+	comment := random.Comment()
+	basicConfigModelAfterUpdate := model.TaskWithId("test", child3Id, true, "SELECT 1").
+		WithAfterValue(configvariable.SetVariable(
+			configvariable.StringVariable(child1.ID().FullyQualifiedName()),
+			configvariable.StringVariable(child2.ID().FullyQualifiedName()),
+		)).
+		WithSqlStatement("SELECT 123"). // Overrides sql_statement
```
sfc-gh-asawicki, line 449: why not as `TaskWithId` parameter

pkg/resources/task_acceptance_test.go
```diff
@@ -1929,3 +2392,182 @@ resource "snowflake_task" "test" {
 }
 `, id.DatabaseName(), id.SchemaName(), id.Name(), errorIntegrationId.Name())
 }
+
+func TestAcc_Task_StateUpgrade(t *testing.T) {
```
sfc-gh-asawicki, line 2396: Can we split it into:
1. No optional fields
2. All optional fields
?

pkg/resources/task_acceptance_test.go
```diff
@@ -27,10 +28,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
-// TODO(SNOW-1348116 - next pr): More tests for complicated DAGs
```
sfc-gh-asawicki, line 30: not sure if the added test is enough for complicated DAG, let's discuss tomorrow

sfc-gh-asawicki (changes requested):

pkg/sdk/tasks_gen.go
```diff
@@ -213,6 +216,33 @@ func (v *Task) IsStarted() bool {
...
+		strings.HasSuffix(upperSchedule, "MINUTES"):
+		minuteParts := strings.Split(upperSchedule, " ")
+		minutes, err := strconv.Atoi(minuteParts[0])
+		if err != nil {
+			return nil, err
+		}
+
+		return &TaskSchedule{Minutes: minutes}, nil
+	}
+	return nil, fmt.Errorf("invalid schedule format: %s", schedule)
```
sfc-gh-jmichalak, line 243: Nit: move this to `default`

pkg/sdk/tasks_gen_test.go
```diff
@@ -552,3 +554,66 @@ func TestTasks_Execute(t *testing.T) {
...
+		ExpectedTaskSchedule *TaskSchedule
+		Error                string
+	}{
+		"valid schedule: m minutes": {
+			Schedule:             "5 m",
+			ExpectedTaskSchedule: &TaskSchedule{Minutes: 5},
+			Error:                "",
+		},
+		"valid schedule: M minutes": {
+			Schedule:             "5 m",
```
sfc-gh-jmichalak, line 570: Did you mean `"5 M"`?

pkg/sdk/tasks_gen_test.go
```diff
@@ -552,3 +554,66 @@ func TestTasks_Execute(t *testing.T) {
...
+			ExpectedTaskSchedule: &TaskSchedule{Minutes: 5},
+			Error:                "",
```
sfc-gh-jmichalak, line 572: `Error` should be omitted in such cases.

pkg/resources/task.go
```diff
@@ -93,16 +101,10 @@ var taskSchema = map[string]*schema.Schema{
...
-			// TODO(SNOW-1348116 - next pr): Currently config has to be passed with $$ prefix and suffix. The best solution would be to put there only json, so it could be retrieved from file, etc. Move $$ adding to the SDK.
-			func(k, oldValue, newValue string, d *schema.ResourceData) bool {
-				return strings.Trim(oldValue, "$") == strings.Trim(newValue, "$")
-			},
-		),
-		Description: "Specifies a string representation of key value pairs that can be accessed by all tasks in the task graph. Must be in JSON format.",
+		Type:             schema.TypeString,
+		Optional:         true,
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("config"),
+		Description:      "Specifies a string representation of key value pairs that can be accessed by all tasks in the task graph. Must be in JSON format.",
```
sfc-gh-jmichalak, line 107: Can we add it to the config that `$$` is added automatically?

pkg/resources/task_acceptance_test.go
```diff
@@ -306,6 +304,303 @@ func TestAcc_Task_Updates(t *testing.T) {
...
+}
+
+/*
+DAG structure (the test proves child3 won't have any issues with updates in the following scenario):
+
+		 child1
+		/		\
+	 root 		 child3
+		\		/
+		 child2
```
sfc-gh-jmichalak, line 416: Nit: root should be on top

pkg/resources/task_acceptance_test.go
```diff
@@ -306,6 +304,303 @@ func TestAcc_Task_Updates(t *testing.T) {
...
+						HasCommentString(comment).
+						HasSqlStatementString("SELECT 123"),
+					resourceshowoutputassert.TaskShowOutput(t, basicConfigModelAfterUpdate.ResourceReference()).
+						HasCreatedOnNotEmpty().
+						HasName(child3Id.Name()).
+						HasDatabaseName(child3Id.DatabaseName()).
+						HasSchemaName(child3Id.SchemaName()).
+						HasState(sdk.TaskStateStarted).
+						HasComment(comment).
+						HasDefinition("SELECT 123"),
```
sfc-gh-jmichalak, line 498: Can we check somewhere that `after` is correct?

sfc-gh-jmichalak (changes requested):

pkg/resources/task.go
sfc-gh-jcieslak: review dogged

pkg/resources/task.go
sfc-gh-jcieslak: removed this one, it's handled in the common parameter function

pkg/resources/task_state_upgraders.go
sfc-gh-jcieslak: mb, started is boolean, in this case we'll preserve the default, but after upgrade it's required so it has to be explicitly defined as not started (I'll mention it in the migration guide). In the other case we can only map this to false and if the user won't define it after we'll make a plan to change it to "default".

pkg/resources/task_state_upgraders.go
sfc-gh-jcieslak: Hmm I checked and first version of the tasks (4 years old) had string `after` I guess it depends on how far do we want to look. I'll add explicit check.

pkg/resources/task_state_upgraders.go
sfc-gh-jcieslak: Good point, added.

pkg/resources/task_state_upgraders.go
sfc-gh-jcieslak: Sure

pkg/sdk/tasks_gen_test.go
sfc-gh-jcieslak: True, added

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: Tru, done

pkg/sdk/tasks_gen_test.go
sfc-gh-jcieslak: True

pkg/resources/task.go
sfc-gh-jcieslak: I mean, does it matter what happens underneath on such low level? I'm not sure, I would say the most important info is that the accepted format should be in JSON.

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: I would leave it this way, formatter doesn't like these kinds of comments and when changing the structure of current graph it formats strangely.

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: Added TODO at the top of the test file

sfc-gh-asawicki (approved):

pkg/resources/task.go
sfc-gh-jmichalak: But someone may think that it needs to be wrapped in `$$`, and then the config would be double-wrapped, right?

sfc-gh-jmichalak (approved):

PR #3210, opened 2024-11-20
feat: Tag association v1 readiness

sfc-gh-jmichalak:
- rework `tag_association` resource
- return `nil` from GetTag instead of failing
- add more tests regarding tag/masking policy: assert that `ALTER MASKING POLICY SET TAG` differs from `ALTER TAG SET MASKING POLICY`
- support tagging account for identifiers with org name
- support `IF EXISTS` for unsetting tags
- add notes about manually unassigning policies from objects, add a todo with an issue number
- fix a wrong issue number in essential objects list

## Test Plan

* [x] acceptance tests

* [ ] …

## References

https://docs.snowflake.com/en/user-guide/object-tagging
https://docs.snowflake.com/en/sql-reference/functions/system_get_tag
#3145 
#1910 
#2943
#3235

## TODO
- use generated config and asserts, remove old test `tf` files

## Ideas
- extract a separate resource for tagging accounts?

PR #3215, opened 2024-11-21
chore: Remove deprecated resources, data sources, and fields in config

sfc-gh-jmichalak:
- remove deprecated data sources and resources
  - some of the tag related fields are not removed - will be done after merging https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3210
  - some resources (procedures, functions, unsafe_execute) are not removed - can be done after reworking 
- remove deprecated fields in the provider config
  - deprecated fields in the resources/data sources that we haven't reworked are still there
- remove some unused code from `snowflake` package
- remove deprecated `JWT` value from `authenticator`
- improve doc generator, so that the headers are added only in case of non-empty lists
  - I have left some comments in the templates because the generated files must not be blank

## Test Plan

* [ ] acceptance tests

* [ ] …

## TODO
- remove objects related with tags, procedures, functions, unsafe_execute


pkg/provider/provider_helpers.go
```diff
@@ -3,20 +3,17 @@ package provider
...
 	"net/url"
-	"os"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/mitchellh/go-homedir"
```
sfc-gh-asawicki, line 19: nit: I see this only lives in framework PoC, maybe let's remove it from there too and get rid of this dependency at all?

pkg/sdk/config.go
```diff
@@ -214,7 +214,6 @@ func GetConfigFileName() (string, error) {
 
 // TODO(SNOW-1787920): improve TOML parsing
 type ConfigDTO struct {
-	Account                        *string             `toml:"account"`
```
sfc-gh-asawicki, line 217: nit (question): what is our current policy in toml parsing: ignore unknown or fail on unknown? If it's the latter I would add a sentence in migration guide which reminds that these fields should be removed both from config but also from the toml config file.

pkg/sdk/helper_test.go
```diff
@@ -46,3 +48,10 @@ func testClient(t *testing.T, profile string) *Client {
 
 	return client
 }
+
+func configAccountId(t *testing.T, cfg *gosnowflake.Config) AccountIdentifier {
```
sfc-gh-asawicki, line 52: this is repeated but in my service user on CI change there will be other alternative instead of this func, so nothing to do yet

v1-preparations/LIST_OF_REMOVED_RESOURCES_FOR_V1.md
```diff
@@ -1,14 +1,10 @@
-Deprecated resources that will be removed with the V1:
+Deprecated resources that are removed with the V1:
 
-* [snowflake_database_old](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/0.97.0/docs/resources/database_old)
-* [snowflake_tag_masking_policy_association](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/0.97.0/docs/resources/tag_masking_policy_association)
-* [snowflake_procedure](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/0.97.0/docs/resources/procedure)
```
sfc-gh-asawicki, line 5: nit: let's still leave the ones that will be removed as a separate paragraph to remember about them (these files don;t care for the binary but if we remove them and not leave a TODO we might forget about it)

sfc-gh-asawicki (changes requested):

pkg/datasources/parameters_acceptance_test.go
```diff
@@ -99,38 +97,6 @@ func TestAcc_Parameters_TransactionAbortOnErrorCanBeSet(t *testing.T) {
 	})
 }
 
-// proves https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2353 is fixed
-// done on user, to not interfere with other parallel tests on the same account
-func TestAcc_Parameters_QuotedIdentifiersIgnoreCaseCanBeSet(t *testing.T) {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Yeah, this is just a check on a parameter, and this is already checked in user acceptance tests - I left a comment in https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3215/files#diff-c56411787ea970a3740f6f0252b3ec7ec26c6a173e98efad897190fdac606128.

MIGRATION_GUIDE.md
```diff
@@ -7,6 +7,36 @@ across different versions.
...
 > We highly recommend upgrading the versions one by one instead of bulk upgrades.
 
+## v0.99.0 ➞ v1.0.0
+
+### Removed deprecated objects
+All of the deprecated objects are removed from v1 release. This includes:
+<!-- TODO(next pr): link to entries in the migration guide regarding details for each of the resource/data source-->
+- Resources
+  - `snowflake_database_old`
+  - `snowflake_role`
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Good catch. Let's discuss it separately, and do this in another PR, if needed.

v1-preparations/LIST_OF_REMOVED_RESOURCES_FOR_V1.md
sfc-gh-jmichalak: Done.

pkg/sdk/config.go
sfc-gh-jmichalak: We ignore fields that are not in ConfigDTO. We fail when there is a data type mismatch on a known field. Added more unit tests for that. But still, I added more examples in the migration guide.

pkg/provider/provider_helpers.go
sfc-gh-jmichalak: Done.

docs/technical-documentation/resource_migration.md
```diff
@@ -1,7 +1,7 @@
 
 # Resource migration
 
-Here's a guide on how to migrate deprecated resources to their new counter-parts.
+Here's a guide on how to migrate deprecated resources to their new counter-parts before v1 release.
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Reverted.

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

PR #3223, opened 2024-11-25
chore: Add usage tracking for the rest of the resources and fix views

sfc-gh-jcieslak:
## Changes
- Add usage tracking to the remaining resources (had to change their method signature to the new one containing context)
- Fix views (trimmed the tracking comment from the text column that contains the command run to create the view)
  - Same for materialized view


pkg/internal/tracking/query_test.go
```diff
@@ -3,12 +3,48 @@ package tracking
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/stretchr/testify/assert"
 	"testing"
```
sfc-gh-jmichalak, line 7: Reviewdog. You should be able to set up your IDE to run `goimports` on each save, it's very helpful.

pkg/internal/tracking/query_test.go
```diff
@@ -3,12 +3,48 @@ package tracking
...
+			Input:    "select 1",
+			Expected: "select 1",
+		},
+		{
+			Input:    "select 1; --some comment",
+			Expected: "select 1; --some comment",
+		},
+		{
+			Input:    fmt.Sprintf("select 1; --%s", MetadataPrefix),
+			Expected: fmt.Sprintf("select 1; --%s", MetadataPrefix),
```
sfc-gh-jmichalak, line 28: I don't get why we expect the query with metadata prefix here. Shouldn't we also trim?

sfc-gh-jmichalak (dismissed):

pkg/resources/oauth_integration_test.go
```diff
@@ -37,10 +38,10 @@ func TestOAuthIntegrationCreate(t *testing.T) {
 		).WillReturnResult(sqlmock.NewResult(1, 1))
 		expectReadOAuthIntegration(mock)
 
-		err := resources.CreateOAuthIntegration(d, &internalprovider.Context{
+		err := resources.CreateOAuthIntegration(context.Background(), d, &internalprovider.Context{
 			Client: sdk.NewClientFromDB(db),
 		})
-		r.NoError(err)
+		r.NotEmpty(err)
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: because err change from `error` to `diag.Diagnostics` which is a type of `[]diag.Diagnostic`, that's why NotEmpty. I can change e.g. the name from `err` to `diag`, but I kind of skimmed through those files as they will be removed or greatly changed in the future.

pkg/sdk/dynamic_table.go
```diff
@@ -3,6 +3,7 @@ package sdk
 import (
 	"context"
 	"database/sql"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/tracking"
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: 🤦

pkg/internal/tracking/query_test.go
sfc-gh-jcieslak: I think I do have it, but I have to be in this file to run it. I'll see if I can setup some pre push or pre commit job that would run pre-push

pkg/internal/tracking/query_test.go
sfc-gh-jcieslak: I didn't have the space afterwards, but I can make it so that this one is also removed. On the other hand it's not valid because the JSON is not there. I don't think this case would ever be a problem, but I can change it.

pkg/sdk/views_impl_gen.go
```diff
@@ -277,7 +278,7 @@ func (r viewDBRow) convert() *View {
 		view.Comment = r.Comment.String
 	}
 	if r.Text.Valid {
-		view.Text = r.Text.String
+		view.Text = tracking.TrimMetadata(r.Text.String)
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: When validating it I checked the acceptance tests for an object and they seem to be working correctly after this change. I'll add the intergration tests.

pkg/internal/tracking/query_test.go
sfc-gh-jcieslak: Done

pkg/resources/oauth_integration_test.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Changed names to diags and assertion to `r.Empty`

sfc-gh-jmichalak (approved):

PR #3226, opened 2024-11-26
fix: Small fixes and adjustments

sfc-gh-jmichalak:
Changes included (they are really small or fix some wrong changes done during tag rework that were not released yet):
- smaller fixes extracted from https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3210/files
- support tagging account for identifiers with org name
- add notes about manually unassigning policies from objects, add a todo with an issue number
- fix a wrong issue number in essential objects list

Changes NOT included
- rework tag_association resource
- return nil from GetTag instead of failing
- add more tests regarding tag/masking policy: assert that ALTER MASKING POLICY SET TAG differs from ALTER TAG SET MASKING POLICY
- support IF EXISTS for unsetting tags

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3210


pkg/sdk/tags_impl.go
```diff
@@ -149,6 +149,14 @@ func (s *SetTagRequest) toOpts() *setTagOptions {
 			o.column = String(id.Name())
 		}
 	}
+	// TODO(SNOW-1818976): Remove this workaround. Currently ALTER "ORGNAME"."ACCOUNTNAME" SET TAG does not work, but ALTER "ACCOUNTNAME" does.
```
sfc-gh-jcieslak, line 152: Hmm, is this even allowed to modify an account in a different organization? If not, then having only account name seems reasonable.

sfc-gh-jcieslak (approved):

pkg/sdk/tags_impl.go
sfc-gh-jmichalak: Yes, you can do sth like
```
alter account "ORGNAME"."ACCOUNTNAME" set OAUTH_ADD_PRIVILEGED_ROLES_TO_BLOCKED_LIST=true;
```

PR #3227, opened 2024-11-26
chore: Add ShowByID filtering generation

sfc-gh-fbudzynski:
## Changes
- Added ShowByID filtering options to `implement_functions` template
- New filed in `Operation` struct with Filtering options

To generate filtering options, add filtering options in `ShowByIDOperation`:
```go
).ShowByIdOperation(
	g.ShowByIDLikeFiltering,
	g.ShowByIDExtendedInFiltering,
)
```
## Test Plan

Generation for `Like` and `ExtendedIn` for SchemaObjectidentifier (Secrets):
* [x] acceptance tests 
* [x] integration tests 
* [x] unit tests 
---
Generation for `Like` and `In` for SchemaObjectIdentifier (Streamlits):
* [x] acceptance tests 
* [x] integration tests 
* [x] unit tests 
---
Generation for `Like` for AccountObjectIdentifier (Connections):
* [x] acceptance tests 
* [x] integration tests 
* [x] unit tests 
---
Generation for `ApplicationName` for DatabaseObjectIdentifier (Application Roles):
* [x] acceptance tests 
* [x] integration tests 
* [x] unit tests 

## Notes 
- `WithIn` and `WithLike` are mostly used in `ShowByID()` 
- no objects use `Limit` and `StartsWith` in `ShowByID()`
- `application_roles` is the only `DatabaseObejctIdentifier` with a `_def` file
- `application_roles` does not implement any of the common filtering options e.g. `OptionalLike()` or `OptionalIn()` 
- `application_roles` implement `WithApplicationName` filtering option that has been added


pkg/datasources/secrets.go
```diff
@@ -107,7 +107,7 @@ func ReadSecrets(ctx context.Context, d *schema.ResourceData, meta any) diag.Dia
 	req := sdk.NewShowSecretRequest()
 
 	handleLike(d, &req.Like)
-	err := handleExtendedIn(d, &req.In)
+	err := handleExtendedIn(d, &req.ExtendedIn)
```
sfc-gh-jmichalak, line 110: I'm open for discussions, but I'm not sure if it's necessary. On the other hand, in Snowflake, is just `IN`, and initially `ExtendedIn` was used only to distinguish from `In` with limited filtering capabilities.

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
```diff
@@ -13,9 +14,21 @@
         return resultList, nil
         }
     {{ else if eq .Name "ShowByID" }}
+        {{ $idkind := .ObjectInterface.ObjectIdentifierKind }}
         func (v *{{ $impl }}) ShowByID(ctx context.Context, id {{ .ObjectInterface.IdentifierKind }}) (*{{ .ObjectInterface.NameSingular }}, error) {
-        // TODO: adjust request if e.g. LIKE is supported for the resource
-        {{ $impl }}, err := v.Show(ctx, NewShow{{ .ObjectInterface.NameSingular }}Request())
+            request := NewShow{{ .ObjectInterface.NameSingular }}Request()
+            {{- if not .ShowByIDFiltering }}
+                // TODO: adjust request if e.g. LIKE is supported for the resource
```
sfc-gh-jmichalak, line 21: We can leave the default behavior as it is, but we can point the SDK users to use `	g.ShowByIDLikeFiltering,` and others more explicitly.

pkg/sdk/poc/generator/show_by_id_filtering.go
```diff
@@ -0,0 +1,46 @@
...
+func (s *Operation) withFiltering(filtering ...ShowByIDFilteringKind) *Operation {
+	for _, f := range filtering {
+		switch f {
+		case ShowByIDLikeFiltering:
+			s.ShowByIDFiltering = append(s.ShowByIDFiltering, ShowByIDFiltering{
+				Kind:            "Like",
+				Args:            "Pattern: String(id.Name())",
+				IdentifierBased: false,
+			})
+		case ShowByIDInFiltering:
```
sfc-gh-jmichalak, line 31: We can use it (non-extended In) in one object to show if it works.

pkg/sdk/poc/generator/show_by_id_filtering.go
```diff
@@ -0,0 +1,46 @@
+package generator
+
+type ShowByIDFiltering struct {
+	Kind            string
+	Args            string
+	IdentifierBased bool
```
sfc-gh-jmichalak, line 6: I wonder if we could use an enum instead? What if we have more showByIDs which have a different behavior?  Such flag could be confusing.

sfc-gh-jmichalak (changes requested):

pkg/sdk/poc/generator/show_by_id_filtering.go
```diff
@@ -0,0 +1,46 @@
...
+
+const (
+	// Enables filtering with: Like
+	ShowByIDLikeFiltering ShowByIDFilteringKind = iota
+	// Enables filtering with: In
+	// Based on the identifier Kind
+	ShowByIDInFiltering
+	// Enables filtering with: ExtendedIn
+	// Based on the identifier Kind
+	ShowByIDExtendedInFiltering
```
sfc-gh-jmichalak, line 19: In your doc you proposed sth like
```
	ShowByIDSchemaFiltering     ShowByIDFilteringKind = "In{Schema: id.SchemaId()}"
	ShowByIDInDatabaseFiltering ShowByIDFilteringKind = "In{Database: id.DatabaseId()}"
```
I'm not saying it's bad, but I'm curious about the other approach here.

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
sfc-gh-fbudzynski: How about: // TODO: adjust request  if e.g. LIKE is supported for the resource with `ShowByIDLikeFiltering` and other filters for "ShowByIdOperation" in the definition file."

pkg/sdk/poc/generator/show_by_id_filtering.go
sfc-gh-fbudzynski: good idea, i will generate some objects to show all the filtering options

pkg/sdk/poc/generator/show_by_id_filtering.go
sfc-gh-fbudzynski: You're right, that would probably be a better approach than just a flag

pkg/sdk/poc/generator/show_by_id_filtering.go
sfc-gh-fbudzynski: I don't think we can skip the In vs. ExtendedIn because those filterings use different structs. For ExtendedIn, we need to provide the ExtendedIn struct, and for Simple In, just the In struct.

I aimed to make it similar in use to what we have for the `ShowOperation`:
```go
g.NewQueryStruct("ShowSecret").
	...
	OptionalLike().
	OptionalExtendedIn(),
```
and then for the filtering we use:
```go
).ShowByIdOperation(
	g.ShowByIDLikeFiltering,
	g.ShowByIDExtendedInFiltering,
)
```

pkg/datasources/secrets.go
sfc-gh-jcieslak: For me, it's more readable that way because I know it's the extended version, but both options are ok.

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
sfc-gh-jmichalak: Sounds good.

pkg/datasources/secrets.go
sfc-gh-fbudzynski: I'll leave it as is, meaning `In` will not be changed to `ExtendedIn`

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
sfc-gh-fbudzynski: I rephrased it a bit:
// TODO: adjust request with resource supported filtering e.g. for LIKE use 'ShowByIDLikeFiltering' in 'ShowByIdOperation()' inside the definition file.

pkg/sdk/poc/generator/show_by_id_filtering.go
sfc-gh-fbudzynski: generated `streamlits` as example, which has the simple `In` filtering

pkg/sdk/poc/generator/operation.go
```diff
@@ -160,8 +162,11 @@ func (i *Interface) ShowOperation(doc string, dbRepresentation *dbStruct, resour
 	return i
 }
 
-func (i *Interface) ShowByIdOperation() *Interface {
-	return i.newNoSqlOperation(string(OperationKindShowByID))
+func (i *Interface) ShowByIdOperation(filtering ...ShowByIDFilteringKind) *Interface {
+	op := i.newNoSqlOperation(string(OperationKindShowByID))
+	op.ObjectInterface = i
```
sfc-gh-fbudzynski, line 167: The `operation.ObjectInterface` is being set here, because without it `operation.ObjectInterface.ObjectIdentifierKind()` can't be used before executing the template. It results in nil pointer dereference, same as for`Interface.NameLowerCased()`.

I wanted to use it before the template execution to invoke the filtering just by calling `{{ .WithFiltering }}` in the template.

I'm open for discussion on this one.

sfc-gh-jcieslak: I'm fine with assigning it in the constructor (`newNoSqlOperation` in this case), it's not bad and imo could be added to other constructors that are `Interface` methods

pkg/sdk/poc/generator/show_by_id_filtering.go
```diff
@@ -14,32 +13,68 @@ const (
...
+
+func newShowByIDLikeFiltering(string) ShowByIDFiltering {
+	return newShowByIDFiltering("Like", "Like", "Pattern: String(id.Name())", nil)
+}
+
+func newShowByIDInFiltering(identifierKind string) ShowByIDFiltering {
+	return newShowByIDFiltering("In", "In", "%[1]v: id.%[1]vId()", &identifierKind)
+}
+
+func newShowByIDExtendedInFiltering(identifierKind string) ShowByIDFiltering {
```
sfc-gh-jmichalak, line 64: Can we have an enum for `identifierKind`?

pkg/sdk/poc/generator/show_by_id_filtering.go
```diff
@@ -14,32 +13,68 @@ const (
...
+}
+
+var filteringMapping = map[ShowByIDFilteringKind]func(string) ShowByIDFiltering{
+	ShowByIDLikeFiltering:       newShowByIDLikeFiltering,
+	ShowByIDInFiltering:         newShowByIDInFiltering,
+	ShowByIDExtendedInFiltering: newShowByIDExtendedInFiltering,
+	ShowByIDLimitFiltering:      newShowByIDLimitFiltering,
+}
+
+func newShowByIDFiltering(name, kind, args string, identifierKind *string) ShowByIDFiltering {
```
sfc-gh-jmichalak, line 44: How about using a variadic function?

pkg/sdk/secrets_def.go
```diff
@@ -241,15 +241,17 @@ var SecretsDef = g.NewInterface(
...
+	g.DescriptionMappingKindSingleValue,
+	"https://docs.snowflake.com/en/sql-reference/sql/desc-secret",
+	secretDetailsDbRow,
+	secretDetails,
+	g.NewQueryStruct("DescribeSecret").
+		Describe().
+		SQL("SECRET").
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).ShowByIdOperation(
```
sfc-gh-asawicki, line 254: nit: why the changed ordering? Now the Show methods are not next to each other in the generated interface.

pkg/sdk/streamlits_def.go
```diff
@@ -103,7 +103,10 @@ var StreamlitsDef = g.NewInterface(
 		OptionalLike().
 		OptionalIn().
 		OptionalLimit(),
-).ShowByIdOperation().DescribeOperation(
+).ShowByIdOperation(
```
sfc-gh-asawicki, line 106: both were added to def using SchemaOBjectIdentifier, let's check how it behaves for:
- accountObjectIdentifier
- databaseObjectIdentifier

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
```diff
@@ -13,9 +14,17 @@
         return resultList, nil
         }
     {{ else if eq .Name "ShowByID" }}
+        {{ $idkind := .ObjectInterface.ObjectIdentifierKind }}
         func (v *{{ $impl }}) ShowByID(ctx context.Context, id {{ .ObjectInterface.IdentifierKind }}) (*{{ .ObjectInterface.NameSingular }}, error) {
-        // TODO: adjust request if e.g. LIKE is supported for the resource
-        {{ $impl }}, err := v.Show(ctx, NewShow{{ .ObjectInterface.NameSingular }}Request())
+            request := NewShow{{ .ObjectInterface.NameSingular }}Request()
+            {{- if not .ShowByIDFiltering }}
```
sfc-gh-asawicki, line 20: is it a desired situation that no filtering is provided? I think we should enforce putting option here, e.g. some special option for no filtering or smth like that. We can discuss it.

pkg/sdk/poc/generator/operation.go
```diff
@@ -85,11 +87,11 @@ func addDescriptionMapping(op *Operation, from, to *Field) {
 	op.DescribeMapping = newMapping("convert", from, to)
 }
 
-func (i *Interface) newNoSqlOperation(kind string) *Interface {
+func (i *Interface) newNoSqlOperation(kind string) *Operation {
```
sfc-gh-asawicki, line 90: nit: why we return Operation here?

sfc-gh-asawicki (changes requested):

pkg/sdk/poc/generator/operation.go
sfc-gh-fbudzynski: Operation is returned here because we want to change the operation just like for the `newOperationWithDBMappin`. Here we want to add filtering to the `Operation.ShowByIDFiltering`.

Alternatively, we could return Interface as before and call the `withFiltering` inside the `newNoSqlOperation`, but then we would have to add `filtering ...ShowByIDFilteringKind` as arguments to it.

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
sfc-gh-fbudzynski: agreed, we could add `ShowByIDNoFiltering` option

pkg/sdk/secrets_def.go
sfc-gh-fbudzynski: the order does not have to be changed, I will regenerate them with correct ordering

pkg/sdk/poc/generator/operation.go
sfc-gh-asawicki: let's do this a free nhanging func and not an Interface's method (like we discussed verbally)

pkg/sdk/poc/generator/show_by_id_filtering.go
sfc-gh-fbudzynski: will be done by formatting arguments in the create function (as discussed verbally). This way the identifier kind is removed from `newShowByIDFiltering()` function and there is no need for variadic anymore.

pkg/sdk/poc/generator/show_by_id_filtering.go
```diff
@@ -0,0 +1,113 @@
...
+			Name: "ApplicationName",
+			Kind: "",
+			Args: "id.DatabaseId()",
+		},
+	}
+}
+
+func (s *Operation) withFiltering(filtering ...ShowByIDFilteringKind) *Operation {
+	for _, filteringKind := range filtering {
+		switch filteringKind {
```
sfc-gh-fbudzynski, line 97: Discussed with @sfc-gh-jmichalak to reject the usage of 
```go
map[ShowByIDFilteringKind]func(idPrefix) ShowByIDFiltering
``` 
and use a switch instead, so that there is no need to append more arguments to each function if new filtering is added

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
```diff
@@ -14,8 +15,13 @@
         }
     {{ else if eq .Name "ShowByID" }}
         func (v *{{ $impl }}) ShowByID(ctx context.Context, id {{ .ObjectInterface.IdentifierKind }}) (*{{ .ObjectInterface.NameSingular }}, error) {
-        // TODO: adjust request if e.g. LIKE is supported for the resource
-        {{ $impl }}, err := v.Show(ctx, NewShow{{ .ObjectInterface.NameSingular }}Request())
+            request := NewShow{{ .ObjectInterface.NameSingular }}Request()
+            {{- range .ShowByIDFiltering }}
+                {{- if not (eq .Name "NoFiltering") -}}.
```
sfc-gh-fbudzynski, line 20: mind the dot at the end here
checking for "NoFiltering" name allows for proper adding filtering options, one after another, and if `NoFiltering` option is applied, the dot will not be present

pkg/sdk/poc/generator/operation.go
```diff
@@ -161,7 +162,17 @@ func (i *Interface) ShowOperation(doc string, dbRepresentation *dbStruct, resour
 }
 
 func (i *Interface) ShowByIdOperation() *Interface {
-	return i.newNoSqlOperation(string(OperationKindShowByID))
+	op := newNoSqlOperation(string(OperationKindShowByID))
+	i.Operations = append(i.Operations, op)
+	return i
+}
+
+func (i *Interface) ShowByIdOperationWithFiltering(filter ShowByIDFilteringKind, filtering ...ShowByIDFilteringKind) *Interface {
```
sfc-gh-fbudzynski, line 170: New Operation forcing users to use filtering options (as discussed with @sfc-gh-asawicki)

pkg/sdk/streamlits_def.go
sfc-gh-fbudzynski: generated the examples:
* accountObjectIdentifier - connections with `Like` filtering
* databaseObjectIdentifier - application_roles with `ApplicationName` filtering as this object does not implement common `Like` or `In` filtering

sfc-gh-asawicki (approved):

pkg/sdk/poc/generator/interface.go
```diff
@@ -28,7 +26,8 @@ func (i *Interface) NameLowerCased() string {
 	return startingWithLowerCase(i.Name)
 }
 
-// ObjectIdentifierKind returns the level of the object identifier (e.g. for DatabaseObjectIdentifier, it returns "Database")
-func (i *Interface) ObjectIdentifierKind() string {
-	return strings.Replace(i.IdentifierKind, "ObjectIdentifier", "", 1)
+// ObjectIdentifierKind returns the level of the object identifier (e.g. for DatabaseObjectIdentifier, it returns the prefix "Database")
+func (i *Interface) ObjectIdentifierPrefix() idPrefix {
+	// return strings.Replace(i.IdentifierKind, "ObjectIdentifier", "", 1)
```
sfc-gh-jcieslak, line 31: remove if it doesn't serve any purpose

pkg/sdk/poc/generator/operation.go
```diff
@@ -162,10 +161,17 @@ func (i *Interface) ShowOperation(doc string, dbRepresentation *dbStruct, resour
...
+	op := newNoSqlOperation(string(OperationKindShowByID))
 	op.ObjectInterface = i
-	op.withFiltering(filtering...)
+	op.withFiltering(append([]ShowByIDFilteringKind{filter}, filtering...)...)
```
sfc-gh-jcieslak, line 173: nit: you can do it a little bit shorter with `(append(filtering, filter)...)`

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
sfc-gh-jcieslak: Wouldn't it be easier if the `.ShowByIDFiltering` would have no items? Then the check wouldn't be necessary, plus you can have two items in the collection, one being `NoFiltering` and the second `SomeFilter`. What I mean is, there's nothing you can do now to prevent unwanted behavior. Maybe instead of `NoFiltering` filtering option, we should rename the default function to emphasize the fact that no filtering is generated there (enum value is also good, but as mentioned, it could be used with other filters making it just a tag value).

sfc-gh-jmichalak: +1, I think having empty slice and removing this `if` would be better.

sfc-gh-jmichalak (approved):

pkg/sdk/poc/generator/interface.go
sfc-gh-fbudzynski: It is used to fill the "In" filtering with ObjectIdentifier prefixes based on the object identifier type e.g. DatabaseObjectIdentifier etc. So that the filtering is defined once, and is applicable for any ObejctIdentifier type.

Issue #3171, opened 2024-11-05
[Bug]: snowflake_oauth_integration_for_custom_clients - blocked_roles_list should be optional


imre-kerr-sb1:
### Terraform CLI Version

1.9.8

### Terraform Provider Version

0.97.0

### Terraform Configuration

```terraform
resource "snowflake_oauth_integration_for_custom_clients" "example" {
  name               = "example"
  enabled            = true
  oauth_client_type  = "CONFIDENTIAL"
  oauth_redirect_uri = "https://example.com/complete/snowflake"
  blocked_roles_list = ["ACCOUNTADMIN", "SECURITYADMIN"] # Fails if ORGADMIN is present in the account
  # blocked_roles_list = ["ACCOUNTADMIN", "ORGADMIN", "SECURITYADMIN"] # Fails if ORGADMIN is *not* present in the account
}
```

### Category

category:resource

### Object type(s)

resource:oauth_integration

### Expected Behavior

`blocked_roles_list` should be optional, matching the syntax of `CREATE SECURITY_INTEGRATION` (https://docs.snowflake.com/en/sql-reference/sql/create-security-integration-oauth-snowflake)

### Actual Behavior

`blocked_roles_list` is mandatory. This is annoying, since I have to check if I'm operating on an orgadmin account or not to give the correct list. Bit of a niche case, but could be avoided entirely by just making the field optional.

Error message when missing orgadmin in an orgadmin account:
```
╷
│ Error: 003629 (42501): Roles [ACCOUNTADMIN, ORGADMIN, SECURITYADMIN] are blocked since parameter OAUTH_ADD_PRIVILEGED_ROLES_TO_BLOCKED_LIST is enabled.
│ 
│   with snowflake_oauth_integration_for_custom_clients.example,
│   on main.tf line 1, in resource "snowflake_oauth_integration_for_custom_clients" "example":
│    1: resource "snowflake_oauth_integration_for_custom_clients" "example" {
│ 
╵
```
Error message when including orgadmin in a non-orgadmin account:
```
╷
│ Error: 001008 (22023): SQL compilation error:
│ invalid value [ORGADMIN] for parameter 'BLOCKED_ROLES_LIST'
│ 
│   with snowflake_oauth_integration_for_custom_clients.example,
│   on main.tf line 1, in resource "snowflake_oauth_integration_for_custom_clients" "example":
│    1: resource "snowflake_oauth_integration_for_custom_clients" "example" {
[...]

PR #3267, opened 2024-12-11
fix: Make blocked_roles_field optional in OAuth security integrations

sfc-gh-jmichalak:
- make the `blocked_roles_field` optional in OAuth security integrations
- when tried to use `NormalizeAndCompareIdentifiersInSet`, the diff was suppressed in step 2 of the new tests, resulting in the role not being removed - this will be fixed next

## Test Plan

* [x] acceptance tests

* [ ] …

## References

#3171

## TODO
- use `NormalizeAndCompareIdentifiersInSet` in this field; also in external oauth resource


pkg/resources/external_oauth_integration.go
```diff
@@ -93,10 +93,14 @@ var externalOauthIntegrationSchema = map[string]*schema.Schema{
...
-		Type:          schema.TypeSet,
-		Elem:          &schema.Schema{Type: schema.TypeString},
-		Optional:      true,
-		Description:   relatedResourceDescription("Specifies the list of roles that the client can set as the primary role.", resources.AccountRole),
+		Type:             schema.TypeSet,
+		Elem:             &schema.Schema{Type: schema.TypeString},
+		Optional:         true,
+		Description:      relatedResourceDescription("Specifies the list of roles that the client can set as the primary role.", resources.AccountRole),
+		DiffSuppressFunc: SuppressIfAny(
+		// TODO(SNOW-1517937): uncomment
```
sfc-gh-asawicki, line 101: when? I see the linked issue is in progress?

sfc-gh-asawicki (dismissed):

sfc-gh-jcieslak (approved):

pkg/resources/external_oauth_integration.go
sfc-gh-jmichalak: Ah, this ticket should be slightly different. For now, we have these lists optional, but all of them across external, custom and partner oauths are missing `NormalizeAndCompareIdentifiersInSet`. Let's do this as a follow-up.

PR #3269, opened 2024-12-11
feat: Basic functions implementation

sfc-gh-asawicki:
Prepare most of the java resource implementation:
- add common drop method (also to procedures)
- fix parameters schema and add mapping (also for procedures)
- fix parameter handling
- handle parameters and test
- handle arguments, return type, runtime version, imports, target path, language
- add default values to arguments
- improve function details with mapping logic (for easier use in resource)
- add a bunch of common functions to handle this family of resources
- handle basic rename
- add TABLE data type (needs logic and tests)
- handle external language change
- test no arguments
- regenerate model builders and the docs
- rename a few attributes
- change requirements for some fields

Next PRs:
- TABLE function improvements and tests
- handle secrets, external access integrations, packages, return not null, and comments
- Add a similar PR for java procedure (reuse what we can)
- Add PR with all other function types
- datasources

### DIFF
--- a/pkg/acceptance/helpers/ids_generator.go
+++ b/pkg/acceptance/helpers/ids_generator.go
@@ -4,7 +4,9 @@ import (
 	"strings"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers/random"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/datatypes"
 )
 
 type IdsGenerator struct {
@@ -97,12 +99,22 @@ func (c *IdsGenerator) NewSchemaObjectIdentifierWithArguments(name string, argum
 	return sdk.NewSchemaObjectIdentifierWithArguments(c.SchemaId().DatabaseName(), c.SchemaId().Name(), name, arguments...)
 }
 
+func (c *IdsGenerator) NewSchemaObjectIdentifierWithArgumentsNewDataTypes(name string, arguments ...datatypes.DataType) sdk.SchemaObjectIdentifierWithArguments {
+	legacyDataTypes := collections.Map(arguments, func(dt datatypes.DataType) sdk.DataType { return sdk.LegacyDataTypeFrom(dt) })
+	return sdk.NewSchemaObjectIdentifierWithArguments(c.SchemaId().DatabaseName(), c.SchemaId().Name(), name, legacyDataTypes...)
+}
+
 func (c *IdsGenerator) NewSchemaObjectIdentifierWithArgumentsInSchema(name string, schemaId sdk.DatabaseObjectIdentifier, argumentDataTypes ...sdk.DataType) sdk.SchemaObjectIdentifierWithArguments {
 	return sdk.NewSchemaObjectIdentifierWithArgumentsInSchema(schemaId, name, argumentDataTypes...)
 }
 
 func (c *IdsGenerator) RandomSchemaObjectIdentifierWithArguments(arguments ...sdk.DataType) sdk.SchemaObjectIdentifierWithArguments {
 	return sdk.NewSchemaObjectIdentifierWithArguments(c.SchemaId().DatabaseName(), c.SchemaId().Name(), c.Alpha(), arguments...)
+}
+
+func (c *IdsGenerator) RandomSchemaObjectIdentifierWithArgumentsNewDataTypes(arguments ...datatypes.DataType) sdk.SchemaObjectIdentifierWithArguments {
+	legacyDataTypes := collections.Map(arguments, func(dt datatypes.DataType) sdk.DataType { return sdk.LegacyDataTypeFrom(dt) })
+	return sdk.NewSchemaObjectIdentifierWithArguments(c.SchemaId().DatabaseName(), c.SchemaId().Name(), c.Alpha(), legacyDataTypes...)
 }
 
 func (c *IdsGenerator) Alpha() string {
--- a/pkg/acceptance/helpers/ids_generator.go
+++ b/pkg/acceptance/helpers/ids_generator.go
@@ -100,7 +100,7 @@ func (c *IdsGenerator) NewSchemaObjectIdentifierWithArguments(name string, argum
 }
 
 func (c *IdsGenerator) NewSchemaObjectIdentifierWithArgumentsNewDataTypes(name string, arguments ...datatypes.DataType) sdk.SchemaObjectIdentifierWithArguments {
-	legacyDataTypes := collections.Map(arguments, func(dt datatypes.DataType) sdk.DataType { return sdk.LegacyDataTypeFrom(dt) })
+	legacyDataTypes := collections.Map(arguments, sdk.LegacyDataTypeFrom)
 	return sdk.NewSchemaObjectIdentifierWithArguments(c.SchemaId().DatabaseName(), c.SchemaId().Name(), name, legacyDataTypes...)
 }
 
@@ -113,7 +113,7 @@ func (c *IdsGenerator) RandomSchemaObjectIdentifierWithArguments(arguments ...sd
 }
 
 func (c *IdsGenerator) RandomSchemaObjectIdentifierWithArgumentsNewDataTypes(arguments ...datatypes.DataType) sdk.SchemaObjectIdentifierWithArguments {
-	legacyDataTypes := collections.Map(arguments, func(dt datatypes.DataType) sdk.DataType { return sdk.LegacyDataTypeFrom(dt) })
+	legacyDataTypes := collections.Map(arguments, sdk.LegacyDataTypeFrom)
 	return sdk.NewSchemaObjectIdentifierWithArguments(c.SchemaId().DatabaseName(), c.SchemaId().Name(), c.Alpha(), legacyDataTypes...)
 }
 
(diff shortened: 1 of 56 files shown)


pkg/sdk/functions_ext.go
```diff
@@ -69,9 +92,135 @@ func functionDetailsFromRows(rows []FunctionDetail) (*FunctionDetails, error) {
...
+	if args, err := parseFunctionOrProcedureSignature(v.Signature); err != nil {
+		errs = append(errs, err)
+	} else {
+		v.NormalizedArguments = args
+	}
+
 	return v, errors.Join(errs...)
 }
 
+func parseFunctionDetailsImport(details FunctionDetails) ([]NormalizedPath, error) {
```
sfc-gh-jcieslak, line 129: nit: I believe this could be simplified with `sdk.ParseCommaSeparatedStringArray` + `collections.MapErr` combo

sfc-gh-jcieslak: also `parseFunctionOrProcedureSignature` (but with different cutset), but let's leave those as they are

pkg/sdk/functions_ext.go
```diff
@@ -69,9 +92,135 @@ func functionDetailsFromRows(rows []FunctionDetail) (*FunctionDetails, error) {
...
+// TODO [SNOW-1850370]: test with strange arg names (first integration test)
+func parseFunctionOrProcedureArgument(arg string) (*NormalizedArgument, error) {
+	log.Printf("[DEBUG] parsing argument: %s", arg)
+	trimmed := strings.TrimSpace(arg)
+	idx := strings.Index(trimmed, " ")
+	if idx < 0 {
+		return nil, fmt.Errorf("arg %s cannot be split into arg name, data type, and default", arg)
+	}
+	argName := trimmed[:idx]
+	rest := strings.TrimSpace(trimmed[idx:])
```
sfc-gh-jcieslak, line 216: 🤔  isn't `DEFAULT` as part of signature (after data type)?

pkg/resources/function_java_acceptance_test.go
```diff
@@ -0,0 +1,435 @@
...
+						HasTargetPathEmpty().
+						HasNoRuntimeVersion().
+						HasFunctionDefinitionString(definition).
+						HasFunctionLanguageString("JAVA").
+						HasFullyQualifiedNameString(id.FullyQualifiedName()),
+					resourceshowoutputassert.FunctionShowOutput(t, functionModel.ResourceReference()).
+						HasIsSecure(false),
+					assert.Check(resource.TestCheckResourceAttr(functionModel.ResourceReference(), "arguments.0.arg_name", argName)),
+					assert.Check(resource.TestCheckResourceAttr(functionModel.ResourceReference(), "arguments.0.arg_data_type", dataType.ToSql())),
+					assert.Check(resource.TestCheckResourceAttr(functionModel.ResourceReference(), "arguments.0.arg_default_value", "")),
```
sfc-gh-jcieslak, line 77: maybe assert all show_output fields once since it is (or not, don't remember) used to detect changes in some of the fields.

pkg/resources/function_commons.go
```diff
@@ -308,15 +340,28 @@ func functionBaseSchema() map[string]schema.Schema {
 			},
 			Description: "Assigns the names of [secrets](https://docs.snowflake.com/en/sql-reference/sql/create-secret) to variables so that you can use the variables to reference the secrets when retrieving information from secrets in handler code. Secrets you specify here must be allowed by the [external access integration](https://docs.snowflake.com/en/sql-reference/sql/create-external-access-integration) specified as a value of this CREATE FUNCTION command’s EXTERNAL_ACCESS_INTEGRATIONS parameter.",
 		},
-		// TODO [SNOW-1348103]: because of https://docs.snowflake.com/en/sql-reference/sql/create-function#id6, maybe it will be better to split into stage + path
 		"target_path": {
-			Type:     schema.TypeString,
+			Type:     schema.TypeSet,
```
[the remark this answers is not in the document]
sfc-gh-asawicki: For now, the tests passed with setting it as set, it won't be a breaking change if we'll decide to change the type later (schema will stay the same for the user), so I will leave it as is and pray :D

pkg/resources/function_commons.go
```diff
@@ -338,9 +383,203 @@ func functionBaseSchema() map[string]schema.Schema {
...
+	functionDetails    *sdk.FunctionDetails
+	functionParameters []*sdk.Parameter
+}
+
+func readFunctionArgumentsCommon(d *schema.ResourceData, args []sdk.NormalizedArgument) error {
+	if len(args) == 0 {
+		// TODO [SNOW-1348103]: handle empty list
+		return nil
+	}
+	if currentArgs, ok := d.Get("arguments").([]map[string]any); !ok {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Yes: the default values which are not returned by SF. Had no other idea than just to update what we have - leaving the defaults unchanged. I will add a comment about it.

pkg/resources/function_commons.go
```diff
@@ -338,9 +383,203 @@ func functionBaseSchema() map[string]schema.Schema {
...
+		return d.Set("arguments", currentArgs)
+	}
+}
+
+func readFunctionImportsCommon(d *schema.ResourceData, imports []sdk.NormalizedPath) error {
+	if len(imports) == 0 {
+		// don't do anything if imports not present
+		return nil
+	}
+	imps := make([]map[string]any, len(imports))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Done

pkg/resources/function_java.go
```diff
@@ -36,17 +47,155 @@ func FunctionJava() *schema.Resource {
...
+		stringAttributeCreateBuilder(d, "function_definition", request.WithFunctionDefinitionWrapped),
+	)
+	if errs != nil {
+		return diag.FromErr(errs)
+	}
+
+	if err := client.Functions.CreateForJava(ctx, request); err != nil {
+		return diag.FromErr(err)
+	}
+	d.SetId(helpers.EncodeResourceIdentifier(id))
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Yes, I don't think we'll change what's saved here because this is the form used in all the CREATE/ALTER/DESCRIBE/SHOW methods.

pkg/sdk/functions_ext.go
sfc-gh-asawicki: They both could, I will leave a TODO about trying to merge with the existing helpers - I was anticipating here a bit different logic, that's why initially I have used a new one.

pkg/sdk/functions_ext.go
sfc-gh-asawicki: It's not, check `create function for Java - default argument value` - it's part of the SHOW (arguments) but not a part of DESCRIBE's signature. Still: it does not provide a value, so the only thing we can do in the future is to use this DEFAULT from SHOW to discover external changes in adding/removing the DEFAULT overall.

pkg/resources/function_java_acceptance_test.go
sfc-gh-asawicki: I will; here I wanted to check the arguments behavior quickly (for the sake of this logic for defaults).

pkg/sdk/functions_ext.go
sfc-gh-jcieslak: Kk I was just wondering what we are parsing here, if the input is coming from describe's signature then it's fine. I was just wondering if `strings.TrimSpace(trimmed[idx:])` could possibly encounter something more than just data type.

sfc-gh-jcieslak (approved):

pkg/sdk/functions_ext.go
```diff
@@ -69,9 +92,136 @@ func functionDetailsFromRows(rows []FunctionDetail) (*FunctionDetails, error) {
...
+	return functionDetailsImports, nil
+}
+
+func parseStageLocationPath(location string) (*NormalizedPath, error) {
+	log.Printf("[DEBUG] parsing stage location path part: %s", location)
+	idx := strings.Index(location, "/")
+	if idx < 0 {
+		return nil, fmt.Errorf("part %s cannot be split into stage and path", location)
+	}
+	stageRaw := strings.TrimPrefix(strings.TrimSpace(location[:idx]), "@")
```
sfc-gh-jmichalak, line 156: What about stages with `/` in their name? If they're not supported, let's mention it in the docs.

pkg/resources/function_commons.go
```diff
@@ -214,6 +230,11 @@ func functionBaseSchema() map[string]schema.Schema {
 						DiffSuppressFunc: DiffSuppressDataTypes,
 						Description:      "The argument type.",
 					},
+					"arg_default_value": {
```
sfc-gh-jmichalak, line 233: I think we could drop `arg_` prefix, in other fields as well, wdyt?

sfc-gh-jmichalak (approved):

pkg/resources/function_commons.go
sfc-gh-asawicki: Yes, I will do it.

pkg/acceptance/helpers/ids_generator.go
```diff
@@ -105,6 +112,11 @@ func (c *IdsGenerator) RandomSchemaObjectIdentifierWithArguments(arguments ...sd
 	return sdk.NewSchemaObjectIdentifierWithArguments(c.SchemaId().DatabaseName(), c.SchemaId().Name(), c.Alpha(), arguments...)
 }
 
+func (c *IdsGenerator) RandomSchemaObjectIdentifierWithArgumentsNewDataTypes(arguments ...datatypes.DataType) sdk.SchemaObjectIdentifierWithArguments {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Noted

PR #3307, opened 2024-12-19
chore: Generate marshal json for each model

sfc-gh-asawicki:
- generate marshal and `depends_on` for each resource and data source model
- replace deprecated `FromModel` with the new one utilizing the aforementioned marshaling
- fix test with explicit empty list (and add it to special variables)
- fix tests with explicit null value (and add it to special variables)
- generalize multiline handling and test it
- use multiline strings for definitions in functions and procedures config builders
- simplify function and procedure definitions
- format whitespace in function and procedure definitions

For the next PRs (added to the issue):
- current config building for functions/procedures (especially definitions) is complex:
  - definition is build using multiline golang strong and sprintf
  - whitespace is formatted
  - this is added in the config builder wrapped in special multiline marker
  - hcl-compatible json is generated
  - json converted to hcl
  - hcl formatted using custom formatters
This should be refined and made simpler (e.g. newline replacement, encoding problems, etc.)

- marking in config builders generators which fields should be handled as multiline by default (because now e.g. SQL function requires the definition, so it's generated in basic builder as a string input, and not the multiline one; because of that it has to be added again with `WithFunctionDefinitionValue(...)` which is confusing)


pkg/acceptance/bettertestspoc/config/model/procedure_sql_model_ext.go
```diff
@@ -1,32 +1,21 @@
...
-}
-
 func ProcedureSqlBasicInline(
 	resourceName string,
 	id sdk.SchemaObjectIdentifierWithArguments,
 	returnType datatypes.DataType,
 	procedureDefinition string,
 ) *ProcedureSqlModel {
-	return ProcedureSql(resourceName, id.DatabaseName(), id.Name(), procedureDefinition, returnType.ToSql(), id.SchemaName())
+	return ProcedureSql(resourceName, id.DatabaseName(), id.Name(), procedureDefinition, returnType.ToSql(), id.SchemaName()).
```
sfc-gh-jmichalak, line 17: Same here.

pkg/acceptance/bettertestspoc/config/config_test.go
```diff
@@ -35,6 +35,9 @@ resource "snowflake_share" "test" {
 				Item{IntField: 2, StringField: "second item"},
 			).
 			WithSingleObject("one", 2).
+			WithTextFieldExplicitNull().
+			WithListFieldEmpty().
+			WithMultilineField("some\nmultiline\ncontent").
```
sfc-gh-jmichalak, line 40: Nit: test the behavior of `\t`.

sfc-gh-jmichalak (approved):

pkg/acceptance/bettertestspoc/config/model/procedure_javascript_model_ext.go
```diff
@@ -1,32 +1,21 @@
...
-}
-
 func ProcedureJavascriptBasicInline(
 	resourceName string,
 	id sdk.SchemaObjectIdentifierWithArguments,
 	returnType datatypes.DataType,
 	procedureDefinition string,
 ) *ProcedureJavascriptModel {
-	return ProcedureJavascript(resourceName, id.DatabaseName(), id.Name(), procedureDefinition, returnType.ToSql(), id.SchemaName())
+	return ProcedureJavascript(resourceName, id.DatabaseName(), id.Name(), procedureDefinition, returnType.ToSql(), id.SchemaName()).
```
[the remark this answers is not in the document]
sfc-gh-asawicki: This is on purpose now.

I added to the PR description the next step which explains it.

pkg/acceptance/bettertestspoc/config/config_test.go
sfc-gh-asawicki: I'll do it in one of the next PRs

sfc-gh-jmichalak (approved):

Issue #3245, opened 2024-12-04
[Bug]: snowflake_account_parameter WEEK_START


danielefrigo:
### Terraform CLI Version

1.9.8

### Terraform Provider Version

0.99

### Terraform Configuration

```terraform
While creating a snowflake_account_parameter resource (parameter WEEK_START), Terraform accepts only values 0 or 1, while this parameter accepts any value from 0 to 7.
https://docs.snowflake.com/en/sql-reference/parameters#week-start
```

### Category

category:resource

### Object type(s)

resource:account_parameter

### Expected Behavior

accept any value from 0 to 7

### Actual Behavior

accept only 0 or 1

### Steps to Reproduce

`resource "snowflake_account_parameter" "week_start" {
  key   = "WEEK_START"
  value = 7
}
`

### How much impact is this issue causing?

High

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-asawicki:
Hey @danielefrigo. Thanks for reaching out to us.

Accounts are being currently reworked and we should soon have a separate resource to manage the given account, however, we will still keep snowflake_account_parameter (at least for v1.x), so we will fix this before the next release.

sfc-gh-jmichalak:
Hi @danielefrigo 👋 

This has been fixed in v1.0.1 ([migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v100--v101), [release](https://github.com/Snowflake-Labs/terraform-provider-snowflake/releases/tag/v1.0.1)). Please upgrade to this version.

sfc-gh-jmichalak:
Closing due to inactivity.

Issue #3116, opened 2024-10-07
[Feature]: Add ENFORCE_NETWORK_RULES_FOR_INTERNAL_STAGES parameter


yuukiyamanaka:
### Use Cases or Problem Statement

When I tried to set `ENFORCE_NETWORK_RULES_FOR_INTERNAL_STAGES` parameter using `snowflake_account_parameter` resource,

```
resource "snowflake_account_parameter" "enforce_network_rules_for_internal_stages" {
  key   = "ENFORCE_NETWORK_RULES_FOR_INTERNAL_STAGES"
  value = "true"
}
```

I encounter the below error:
```
Error: Invalid object parameter: ENFORCE_NETWORK_RULES_FOR_INTERNAL_STAGES

with snowflake_account_parameter.enforce_network_rules_for_internal_stages
```

### Category

category:resource

### Object type(s)

resource:account_parameter

### Proposal

I think this parameter is still in public preview, so the provider does not support it yet.
Do you have plans to add this parameter? If not, I would like to submit a PR.

### How much impact is this issue causing?

Low

### Would you like to implement a fix?

- [X] Yeah, I'll take it 😎

sfc-gh-asawicki:
Hey @yuukiyamanaka. Thanks for reaching out to us.

We will shortly visit the rework of `snowflake_account` and `snowflake_account_parameter` resources (check https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/v1-preparations/ESSENTIAL_GA_OBJECTS.MD and https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/v1-preparations/REMAINING_GA_OBJECTS.MD). We will add this parameter then.

One more important thing to mention: as I can see in the docs https://docs.snowflake.com/en/sql-reference/parameters#enforce-network-rules-for-internal-stages, this parameter is currently in public preview. We focus on the generally available features in this provider because the preview features tend to change.

Before adding this parameter, you can also use the [snowflake_unsafe_execute resource](https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/unsafe_execute) to set this parameter through terraform.

yuukiyamanaka:
Thank you for response!
I understand the current situation.

sfc-gh-jmichalak:
Hi @yuukiyamanaka 👋 

[...]

Issue #3044, opened 2024-09-04
[Bug]: Missing account parameters in provider


dlouseiro:
### Terraform CLI Version

1.5.7

### Terraform Provider Version

0.92.0

### Terraform Configuration

```terraform
resource "snowflake_account_parameter" "cortex_enabled_cross_region" {
  key   = "CORTEX_ENABLED_CROSS_REGION"
  value = "AWS_EU"
}

resource "snowflake_account_parameter" "saml_identity_provider" {
  key   = "SAML_IDENTITY_PROVIDER"
  value = "<some-super-long-string>"
}

resource "snowflake_account_parameter" "sso_login_page" {
  key   = "SSO_LOGIN_PAGE"
  value = "true"
}
```

### Category

category:resource

### Expected Behavior

The `terraform apply` would not raise any errors.

### Actual Behavior

Provider fails with:

```
│ Error: Invalid object parameter: CORTEX_ENABLED_CROSS_REGION
│
│   with snowflake_account_parameter.cortex_enabled_cross_region,
│   on account.tf line 25, in resource "snowflake_account_parameter" "cortex_enabled_cross_region":
│   25: resource "snowflake_account_parameter" "cortex_enabled_cross_region" {
│
╵
╷
│ Error: Invalid object parameter: SAML_IDENTITY_PROVIDER
│
│   with snowflake_account_parameter.saml_identity_provider,
│   on account.tf line 50, in resource "snowflake_account_parameter" "saml_identity_provider":
│   50: resource "snowflake_account_parameter" "saml_identity_provider" {
│
╵
╷
│ Error: Invalid object parameter: UI_NOTIFICATIONS
│
│   with snowflake_account_parameter.ui_notifications,
│   on account.tf line 70, in resource "snowflake_account_parameter" "ui_notifications":
│   70: resource "snowflake_account_parameter" "ui_notifications" {
│
```

### Steps to Reproduce

Running `terraform apply`

### How much impact is this issue causing?

Medium

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jcieslak:
Hey 👋 
[...]

PR #3310, opened 2024-12-20
fix: Fixes in account parameters

sfc-gh-jmichalak:
- Add missing parameters based on the docs and output of SHOW PARAMETERS IN ACCOUNT
- Add missing unit tests and implementations
- Add missing validations
- Relax validations for ints - just validate if they are non-negative
- Update docs

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

#3044 
#3116
#3245

## TODO
- Do not add missing user parameters - this will be done in SNOW-1844996. Here, they are added only to account and session level
- More acceptance and integration tests should be done on a separate account. This should be done during account resource follow-up (SNOW-1866453)


pkg/resources/account_parameter_acceptance_test.go
```diff
@@ -131,4 +131,46 @@ func TestAcc_AccountParameter_Issue3025(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: acc.CheckAccountParameterUnset(t, sdk.AccountParameterRequireStorageIntegrationForStageCreation),
+		Steps: []resource.TestStep{
+			{
+				Config: config.FromModel(t, model),
```
sfc-gh-asawicki, line 145: We'll have problem here because in #3307 this method is completely removed (because it was deprecated). Instead, `FromModels` should be used. But we can do a quick follow-up after both merges.

sfc-gh-asawicki (approved):

sfc-gh-asawicki (approved):

PR #3343, opened 2025-01-18
fix: change EnforceNetworkRulesForInternalStages ddl to parameter

yuukiyamanaka:
`ENFORCE_NETWORK_RULES_FOR_INTERNAL_STAGES` requires boolean parameter so I change the setting from ddl:keyword to ddl:parameter.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/3344


sfc-gh-jmichalak (approved):

sfc-gh-jmichalak:
There is a missing entry about the fix from the migration guide, but we can add it on our side.

sfc-gh-jmichalak (changes requested):
Please set the target branch to `dev` we will update the contribution guide about this change.

Also, there is a missing entry about the fix from the migration guide, but we can add it on our side.

sfc-gh-jmichalak (approved):

yuukiyamanaka:
Thank you!

Issue #3300, opened 2024-12-17
[Bug]: Documentation of snowflake_grant_ownership misleading: hybrid tables are not treated separately for granting purposes.


nicolamarangoni:
### Terraform CLI Version

1.9.2

### Terraform Provider Version

1.0.0

### Company Name

ProSiebenSat.1

### Terraform Configuration

```terraform
Hello,



Best regards,

Nicola
```

### Category

category:resource

### Object type(s)

resource:grant_ownership

### Expected Behavior

HYBRID TABLES not included in the list object_type_plural

### Actual Behavior

HYBRID TABLES present in the list object_type_plural

### Steps to Reproduce

Grant ownership resource with:

```
  on {
    future {
      object_type_plural = "HYBRID TABLES"
      in_database        = "MY_DATABASE"
    }
  }
```

### How much impact is this issue causing?

Low

### Additional Information

Actually, transferring ownership of tables also transfer ownership of hybrid tables, it seems that hybrid tables are not considered separately for grants.

I think that the documentation of the resource snowflake_grant_ownership in the Terraform provider is misleading: https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/grant_ownership

In the list of object_type_plural, HYBRID TABLES are mentioned. That's why I tried to transfer ownership of hybrid tables, first with terraform, then manually with SQL.

However, the GRANT OWNERSHIP statement created by Terraform doesn't work.

Try:
```
GRANT OWNERSHIP ON ALL HYBRID TABLES IN DATABASE my_database  TO ROLE my_role;
GRANT OWNERSHIP ON ALL HYBRID TABLES IN SCHEMA my_database.my_schema  TO ROLE my_role;

GRANT OWNERSHIP ON FUTURE HYBRID TABLES IN DATABASE my_database  TO ROLE my_role;
GRANT OWNERSHIP ON FUTURE HYBRID TABLES IN SCHEMA my_database.my_schema  TO ROLE my_role;
```

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
Hi @nicolamarangoni 👋 

[...]

PR #3368, opened 2025-01-30
chore: Add notes about managing grants on hybrid table

sfc-gh-jmichalak:
- Add a note about managing grants on hybrid tables. Currently, this fails in Snowflake and grants should be managed through TABLE.
   - It has not been confirmed whether this is a bug in Snowflake or just in the docs (probably the former). See DOC-4953 for more info.
  - Add a test proving this behavior.
- Add hybrid table client helper.
- This change does not fix the issue, but it will be done in SNOW-1022645.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

#3300

### DIFF
--- a/docs/resources/grant_application_role.md
+++ b/docs/resources/grant_application_role.md
@@ -13,7 +13,7 @@ description: |-
 
 ```terraform
 locals {
-  application_role_identifier = "\"my_appplication\".\"app_role_1\""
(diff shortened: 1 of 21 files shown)


sfc-gh-jmichalak:
I checked other "special" tables, and for the moment HYBRID is the only one failing.

sfc-gh-asawicki (approved):

PR #3370, opened 2025-01-31
chore: Add preview and stable categories to resources and data sources in docs

sfc-gh-jmichalak:
- Group stable and preview resources and data sources in docs.
- Adjust a few descriptions regarding preview status.

## Test Plan

* [ ] acceptance tests

* [ ] …


docs/resources/grant_privileges_to_database_role.md
```diff
@@ -1,7 +1,6 @@
 ---
-# generated by https://github.com/hashicorp/terraform-plugin-docs
```
sfc-gh-asawicki, line 2: why this was removed?

docs/resources/grant_privileges_to_account_role.md
```diff
@@ -1,7 +1,6 @@
 ---
-# generated by https://github.com/hashicorp/terraform-plugin-docs
```
sfc-gh-asawicki, line 2: why this was removed?

docs/resources/execute.md
```diff
@@ -1,7 +1,6 @@
 ---
-# generated by https://github.com/hashicorp/terraform-plugin-docs
```
sfc-gh-asawicki, line 2: why was it removed?

sfc-gh-asawicki (changes requested):
Additional two general comments:
1. Let's add in the migration guide that we change the structure (add links there).
2. I am a bit worried about the existing relative links:
  - between stable and preview (and the other way around) - because adding hierarchy probably changes them (we can check it in other providers using subcategories)
  - between resource/data source and guide (and the other way around) - because again, they will be one level deeper

sfc-gh-jcieslak (approved):
Approved, but as Artur mentioned there may be issues with relative links.

docs/resources/grant_privileges_to_database_role.md
sfc-gh-jmichalak: This seems to be not needed, as other resources do not have such comments.

docs/resources/grant_privileges_to_account_role.md
sfc-gh-jmichalak: This seems to be not needed, as other resources do not have such comments.

docs/resources/execute.md
sfc-gh-jmichalak: This seems to be not needed, as other resources do not have such comments.

docs/resources/grant_ownership.md
```diff
@@ -1,7 +1,6 @@
 ---
-# generated by https://github.com/hashicorp/terraform-plugin-docs
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: This seems to be not needed, as other resources do not have such comments.

sfc-gh-jmichalak:
I added a migration guide entry. I checked several providers, and the categories do not change link addresses.

Issue #3322, opened 2024-12-30
SNOW-1869750: [Feature]: don't crash the Provider if the authentication is set to key-pair, but `private_key` is not specified


ahsanshafiq:
### Terraform CLI Version

1.10.2

### Terraform Provider Version

0.94.1

### Terraform Configuration

```terraform
terraform {
  required_providers {
    snowflake = {
      source  = "Snowflake-Labs/snowflake"
      version = "~> 0.94.1"
    }
  }
}
```

### Category

category:resource

### Expected Behavior

terraform apply tfplan executes without terraform provider crashing.

### Actual Behavior

```
	github.com/jmoiron/sqlx@v1.3.5/sqlx.go:642 +0x48
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk.NewClient(0x0?)
	github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/client.go:153 +0x12f
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider.ConfigureProvider(0x20e014f?)
	github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/provider.go:764 +0x19d2
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Provider).Configure(0xc0009c5a40, {0x24b1278, 0xc000c7da10}, 0xc0009de910)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.33.0/helper/schema/provider.go:296 +0x1b8
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ConfigureProvider(0xc0007d3998, {0x24b1278?, 0xc000c7cc90?}, 0xc000c39800)
	github.com/hashicorp/terraform-plugin-sdk/v2@v2.33.0/helper/schema/grpc_provider.go:611 +0x3a9
github.com/hashicorp/terraform-plugin-mux/tf5to6server.v5tov6Server.ConfigureProvider({{0x24c7fc0?, 0xc0007d3998?}}, {0x24b1278?, 0xc000c7cc90?}, 0xc000c397a0?)
	github.com/hashicorp/terraform-plugin-mux@v0.15.0/tf5to6server/tf5to6server.go:86 +0x182
[...]

Issue #3316, opened 2024-12-20
[Feature] SNOW-1866797: Update snowflake go driver to v1.12.0 or later


jmeekinsint:
### Use Cases or Problem Statement

Update snowflake go driver to v1.12.0 or later

### Category

category:other

### Proposal

Update snowflake go driver to v1.12.0 or later - to allow for Custom Okta auth

### How much impact is this issue causing?

Low

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-dszmolka:
hi - sure, will bump it. will keep this thread posted.

sfc-gh-jmichalak:
Hi @jmeekinsint,

We've released v1.0.4 ([release](https://github.com/Snowflake-Labs/terraform-provider-snowflake/releases/tag/v1.0.4), [migration guide](https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#v103--v104)) with the updated driver to v1.13.0.

sfc-gh-jmichalak:
Closing due to inactivity.

PR #3389, opened 2025-02-07
chore: Upgrade deps

sfc-gh-jmichalak:
- Upgrade libraries, tools, and Go version to 1.22.
- Add a note about the new field in the driver.
- Ignore linter warnings about using the new field (this will be continued in SNOW-1917271).
- Replace `ConcatSlices` with `slices.Concat`.
- Exclude a listing rule for the generated structs.
- Remove `avast/retry-go` library. Use our own solution instead (it was already used in other places before).
- Remove `golangci-lint` from tools. It is not needed, because of a custom installation in the Makefile.
- Increase a warehouse test timeout.
- Fix tests of granting all streamlits and add a note to the docs about the grants.
- Remove the deprecated linter.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

#3322 
#3316

## TODO
- Upgrade Go to 1.23 and 1.24 when it's available
- Close relevant dependabot PRs.


pkg/sdk/functions_impl_gen.go
```diff
@@ -110,12 +110,7 @@ func (r *CreateForJavaFunctionRequest) toOpts() *CreateForJavaFunctionOptions {
 	if r.Arguments != nil {
 		s := make([]FunctionArgument, len(r.Arguments))
 		for i, v := range r.Arguments {
-			s[i] = FunctionArgument{
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Fixed.

pkg/sdk/testint/failover_groups_integration_test.go
```diff
@@ -267,27 +266,16 @@ func TestInt_CreateSecondaryReplicationGroup(t *testing.T) {
...
-			retry.OnRetry(func(n uint, err error) {
-				log.Printf("[DEBUG] Retrying client.FailoverGroups.Drop(): #%d", n+1)
-			}),
-			retry.Delay(1*time.Second),
-			retry.Attempts(3),
-		)
+		dropFailoverGroupOnClient := func() (error, bool) {
+			return client.FailoverGroups.Drop(ctx, failoverGroup.ID(), nil), true
+		}
+		err := util.Retry(3, 1*time.Second, dropFailoverGroupOnClient)
```
sfc-gh-jcieslak, line 272: shouldn't we use the `assert.EventuallyXX`? or at least use `assert.NoError` to continue the test cleanup. If the first cleanup fails the other one won't be called which may result in object leftovers after the failed test.

pkg/sdk/testint/failover_groups_integration_test.go
```diff
@@ -711,27 +699,16 @@ func TestInt_FailoverGroupsAlterTarget(t *testing.T) {
...
+		require.NoError(t, err)
```
sfc-gh-jcieslak, line 706: same here

sfc-gh-asawicki (approved):

pkg/sdk/testint/failover_groups_integration_test.go
sfc-gh-jmichalak: Done.

sfc-gh-jcieslak (approved):

PR #3407, opened 2025-02-21
fix: Fix dispatch event checkout

sfc-gh-jmichalak:
- Fix checking out code in the repository dispatch event by specifying the repository and ref.
- Print more debug info during hash verification.

## Test Plan
Dispatch events utilize pipelines from the default branch, and custom dispatch events cannot be run before merge—see https://github.com/orgs/community/discussions/24657?sort=top. This is why this change was not tested in this repository, as the changes already need to be present on the default branch. 

Instead, I tested it on my personal repository with a simplified setup, but preserving the logic. I confirmed that for dispatch events, the checkout was done to the default origin branch, and after these changes, the checkout is done on the correct sha in the forked branch.

## References
https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3400#issuecomment-2666395813
## TODO
Check the dispatch after it's merged to the default branch.


.github/workflows/tests.yml
```diff
@@ -28,8 +28,20 @@ jobs:
...
+          echo ========
+          echo \"${{ github.event.client_payload.pull_request.head.repo.full_name }}\"
+          echo \"${{ github.event.client_payload.pull_request.head.ref }}\"
 
-      - uses: actions/checkout@v3
+      - name: Checkout Code (Repository Dispatch Event)
+        if: (github.event_name == 'repository_dispatch')
+        uses: actions/checkout@v3
+        with:
+          repository: ${{ github.event.client_payload.pull_request.head.repo.full_name }}
```
sfc-gh-asawicki, line 39: do we need to specify the repository name? Isn't it the default in this action?

sfc-gh-asawicki (changes requested):

.github/workflows/tests.yml
sfc-gh-jcieslak: Should be fine (?) `Default: ${{ github.repository }}` not sure what is the difference between the two. Maybe there's some kind of additional data that can be passed to the repository field that would make it able to checkout on the fork (?)

sfc-gh-asawicki (approved):

.github/workflows/tests.yml
sfc-gh-jmichalak: Removed.

sfc-gh-jcieslak (approved):

PR #3411, opened 2025-02-24
chore: Add test time measurements

sfc-gh-jcieslak:
## Changes
- Add a set of tools to measure test times and print the summary at the end

## Tests
- The solution was already tested on unit and integration tests and seems to be working correctly. The only limitation is that sub-tests are separated by `/` sign which shouldn't be used in any test name (I checked that none of our test is using it, so we should be fine).

## Next steps
- Adjust all the tests to include the `measureTest` function at the start of one of them (we can agree in which cases we would like to skip measuring sub-tests, e.g., in the case of unit tests, which are really fast).

## References
Alternatives considered:
- [Using testify suite package](https://github.com/stretchr/testify?tab=readme-ov-file#suite-package)
- [Using custom code, similar to the approach in the grpc-go codebase (similar to testify suite package, but without any dependency)](https://github.com/grpc/grpc-go/blob/master/internal/grpctest/grpctest.go#L104)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/Snowflake-Labs/terraform-provider-snowflake
 
-go 1.22.10
+go 1.24.0
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
--- a/pkg/internal/measurement/test_measurements.go
+++ b/pkg/internal/measurement/test_measurements.go
@@ -1,0 +1,74 @@
+package measurement
+
+import (
+	"fmt"
+	"golang.org/x/exp/maps"
+	"slices"
+	"strings"
+	"testing"
+	"time"
+)
+
+type TestMeasurement struct {
+	TestName        string
+	Duration        time.Duration
+	SubMeasurements map[string]*TestMeasurement
+}
+
+func NewTestMeasurement(t *testing.T) *TestMeasurement {
+	m := new(TestMeasurement)
+	m.TestName = t.Name()
+	m.SubMeasurements = make(map[string]*TestMeasurement)
+	return m
+}
+
+type TestMeasurements struct {
+	Measurements map[string]*TestMeasurement
+}
+
+func NewTestMeasurements() *TestMeasurements {
+	return &TestMeasurements{
+		Measurements: make(map[string]*TestMeasurement),
+	}
+}
+
+func PrintMeasurementSummary(measurements *TestMeasurements, truncateDuration time.Duration) {
+	fmt.Println("Summary of test measurements (sorted by execution time)")
+	values := maps.Values(measurements.Measurements)
+	slices.SortFunc(values, func(a, b *TestMeasurement) int {
+		return int(b.Duration.Nanoseconds() - a.Duration.Nanoseconds())
+	})
+	printMeasurements(0, values, truncateDuration)
+}
+
+func printMeasurements(level int, measurements []*TestMeasurement, truncateDuration time.Duration) {
+	for _, measurement := range measurements {
+		fmt.Printf("%-10s%s%s\n", measurement.Duration.Truncate(truncateDuration), strings.Repeat(" → ", level), measurement.TestName)
+		printMeasurements(level+1, maps.Values(measurement.SubMeasurements), truncateDuration)
+	}
+}
+
+func MeasureTestTime(testMeasurements *TestMeasurements, t *testing.T) {
+	t.Helper()
+	start := time.Now()
+	m := NewTestMeasurement(t)
+	testNameParts := strings.Split(m.TestName, "/")
+	if len(testNameParts) == 1 {
+		if _, ok := testMeasurements.Measurements[m.TestName]; ok {
+			panic("this shouldn't happen")
+		}
+
+		testMeasurements.Measurements[m.TestName] = m
+	} else {
+		currentMeasurement := testMeasurements.Measurements[testNameParts[0]]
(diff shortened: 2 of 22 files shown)


pkg/sdk/accounts_test.go
```diff
@@ -10,8 +10,64 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestATest(t *testing.T) {
```
sfc-gh-jmichalak, line 13: How about moving to `pkg/internal/measurement/`

pkg/sdk/setup_test.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/measurement"
+	"os"
+	"testing"
+	"time"
+)
+
+var testMeasurements = measurement.NewTestMeasurements()
+
+func measureTest(t *testing.T) {
```
sfc-gh-jmichalak, line 12: Shouldn't it belong to `measurement` package?

pkg/internal/measurement/test_measurements.go
```diff
@@ -0,0 +1,74 @@
...
+		for index, part := range testNameParts[1:] {
+			if value, ok := currentMeasurement.SubMeasurements[part]; ok {
+				currentMeasurement = value
+			} else {
+				m.TestName = strings.Join(testNameParts[index+1:], "/")
+				currentMeasurement.SubMeasurements[part] = m
+			}
+		}
+	}
+	t.Cleanup(func() { m.Duration = time.Now().Sub(start) })
```
sfc-gh-jmichalak, line 73: Pls add a method in `TestMeasurement`  for this.

pkg/internal/measurement/test_measurements.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		testMeasurements.Measurements[m.TestName] = m
+	} else {
+		currentMeasurement := testMeasurements.Measurements[testNameParts[0]]
+		for index, part := range testNameParts[1:] {
+			if value, ok := currentMeasurement.SubMeasurements[part]; ok {
+				currentMeasurement = value
+			} else {
+				m.TestName = strings.Join(testNameParts[index+1:], "/")
```
sfc-gh-jmichalak, line 68: This is not clear, why we're omitting the previous parts and joining the next ones?

pkg/internal/measurement/test_measurements.go
```diff
@@ -0,0 +1,74 @@
...
+		printMeasurements(level+1, maps.Values(measurement.SubMeasurements), truncateDuration)
+	}
+}
+
+func MeasureTestTime(testMeasurements *TestMeasurements, t *testing.T) {
+	t.Helper()
+	start := time.Now()
+	m := NewTestMeasurement(t)
+	testNameParts := strings.Split(m.TestName, "/")
+	if len(testNameParts) == 1 {
```
sfc-gh-jmichalak, line 56: Nit: IMO this whole block could be extracted and simplified as a recursion. Also, why we're not doing a map check with panic in `else`?

pkg/internal/measurement/test_measurements.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func NewTestMeasurement(t *testing.T) *TestMeasurement {
+	m := new(TestMeasurement)
+	m.TestName = t.Name()
+	m.SubMeasurements = make(map[string]*TestMeasurement)
+	return m
+}
+
+type TestMeasurements struct {
```
sfc-gh-jmichalak, line 25: I'm wondering if we really need this. Can we just pass `[]TestMeasurement` to the functions?

sfc-gh-jmichalak (changes requested):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Snowflake-Labs/terraform-provider-snowflake
 
-go 1.22.10
+go 1.24.0
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yeah, I know; this should be reversed to the one used in `dev.` I had to bump to see if `t.Context()` would be helpful.

pkg/sdk/accounts_test.go
sfc-gh-jcieslak: I'll either move or remove it completely. It was used only for testing purposes.

pkg/sdk/setup_test.go
sfc-gh-jcieslak: No, because you want to have separate lists for unit, integration, and acceptance tests. Separate method ensures the right default container is used (the body is in measurement package anyway, so most of the part is extracted).

pkg/internal/measurement/test_measurements.go
sfc-gh-jcieslak: It's only for visualization purposes, instead of seeing
```
Test
-> Test/Sub-Test
```
you'll see
```
Test
-> Sub-Test
```

pkg/internal/measurement/test_measurements.go
sfc-gh-jcieslak: Maybe it could be like that, I'll take a look next time.

pkg/internal/measurement/test_measurements.go
sfc-gh-jcieslak: It could be done with recursion, but this also works. I'll see next why I didn't check for panic (or value presence in the map).

Issue #3402, opened 2025-02-20
[Bug]: terraform import crashing for one of our Snowflake accounts


polarn:
### Terraform CLI Version

1.10.5

### Terraform Provider Version

1.0.3

### Company Name

Validio AB

### Terraform Configuration

```terraform
resource "snowflake_account" "demo" {
  name                 = "DEMO"
  admin_name           = "terraform"
  admin_rsa_public_key = "removed"
  admin_user_type      = "SERVICE"
  email                = "hello@example.com"
  edition              = "ENTERPRISE"
  region_group         = "PUBLIC"
  region               = "GCP_EUROPE_WEST4"
  # comment              = "SNOWFLAKE" # Not possible to change
  is_org_admin         = "true"
  grace_period_in_days = 7

  lifecycle {
    ignore_changes = [
      region_group,
    ]
  }
}
```

### Category

category:import

### Object type(s)

resource:account

### Expected Behavior

Expected account to be imported in state when running:

```
terraform import snowflake_account.demo "<masked>.DEMO"
```

### Actual Behavior

The provider crashes without importing, here is a log output:

```
2025-02-20T09:43:00.217+0100 [DEBUG] provider.terraform-provider-snowflake_v1.0.3: 2025/02/20 09:43:00 [DEBUG] sql-conn-query: [query SELECT IS_ROLE_IN_SESSION('"ORGADMIN"') AS IS_ROLE_IN_SESSION --terraform_provider_usage_tracking {"json_schema_version":"1","version":"v1.0.3","resource":"snowflake_account","operation":"import"} err <nil> duration 184.42044ms args {}] (EW61508)
2025-02-20T09:43:00.503+0100 [DEBUG] provider.terraform-provider-snowflake_v1.0.3: 2025/02/20 09:43:00 [DEBUG] sql-conn-query: [query SHOW ACCOUNTS LIKE 'DEMO' --terraform_provider_usage_tracking {"json_schema_version":"1","version":"v1.0.3","resource":"snowflake_account","operation":"import"} err <nil> duration 284.412593ms args {}] (EW61508)
[...]

PR #3417, opened 2025-02-25
fix: account for null comment

sfc-gh-jcieslak:
## Changes
Fixes null comment case in the account resource described in https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/3402. Because of the limitation in the terraform testing framework, I couldn't create an account externally and then import the created account as a first step of the test (because the testing framework seems to be only working on already existing objects that can be found in-state). That's why I had to test this change manually by using `make install-tf` and using a locally built provider. I confirm that before accounts created with explicitly null comment `COMMENT = null` were failing and after applying this change they are not failing anymore.


sfc-gh-jmichalak (approved):
How about adding a test in `resources/manual_tests`?

pkg/resources/manual_tests/handling_account_null_comment/main.tf
```diff
@@ -0,0 +1,23 @@
...
+}
+
+provider "snowflake" {
+}
+
+resource "snowflake_account" "test_account" {
+  grace_period_in_days = 3
+  name                 = "<name>"
+  admin_name           = "<admin_name>"
+  admin_password       = "<admin_password>"
```
sfc-gh-jmichalak, line 17: Nit: in other tests we mark such values with `# TODO: Replace`.

pkg/resources/manual_tests/handling_account_null_comment/handling_account_null_comment.md
```diff
@@ -0,0 +1,46 @@
+# Handling account null comment
+
+This test shows that the problem from [this issue](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/3402)
```
sfc-gh-jmichalak, line 3: Nit: please note what exactly the issue is, so there's no need to click the link.

pkg/resources/manual_tests/handling_account_null_comment/handling_account_null_comment.md
```diff
@@ -0,0 +1,46 @@
...
+    ADMIN_PASSWORD = '<password>'
+    ADMIN_USER_TYPE = SERVICE
+    EMAIL = '<email>'
+    EDITION = STANDARD
+    COMMENT = NULL;
+```
+
+## Test steps 
+
+1. Build the provider by running `make install-tf`.
```
sfc-gh-jmichalak, line 23: I'm not sure this setup should be a part of this guide. However, it's valuable - can we extract this one level up?

pkg/resources/account.go
```diff
@@ -336,12 +341,17 @@ func ReadAccount(withExternalChangesMarking bool) schema.ReadContextFunc {
...
+				comment = *account.Comment
+			}
+
 			if err = handleExternalChangesToObjectInShow(d,
 				outputMapping{"edition", "edition", *account.Edition, *account.Edition, nil},
 				outputMapping{"is_org_admin", "is_org_admin", *account.IsOrgAdmin, booleanStringFromBool(*account.IsOrgAdmin), nil},
 				outputMapping{"region_group", "region_group", regionGroup, regionGroup, nil},
 				outputMapping{"snowflake_region", "region", account.SnowflakeRegion, account.SnowflakeRegion, nil},
-				outputMapping{"comment", "comment", *account.Comment, *account.Comment, nil},
+				outputMapping{"comment", "comment", comment, comment, nil},
```
sfc-gh-asawicki, line 354: this line is funny TBH

pkg/resources/manual_tests/handling_account_null_comment/handling_account_null_comment.md
sfc-gh-asawicki: Yup, we have the instructions for building and attaching the provider in https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/CONTRIBUTING.md#advanced-debugging. There's no need to repeat these, let's just add: build and setup provider as described. in ... (or smth similar)

sfc-gh-asawicki (changes requested):

pkg/resources/manual_tests/handling_account_null_comment/handling_account_null_comment.md
sfc-gh-jcieslak: Done

pkg/resources/manual_tests/handling_account_null_comment/handling_account_null_comment.md
sfc-gh-jcieslak: Done

pkg/resources/account.go
sfc-gh-jcieslak: Yeah, I think that we should make some adjustments to handle null pointers inside `handleExternalChangesToObjectInShow` instead of defining non-pointer values before the mappings.

sfc-gh-asawicki (approved):

PR #3426, opened 2025-02-28
chore: Protect pentesting user and role from sweepers

sfc-gh-jcieslak:
## Changes
- Adjust sweepers to ignore users and roles used for testing.

### DIFF
--- a/pkg/sdk/sweepers.go
+++ b/pkg/sdk/sweepers.go
@@ -6,8 +6,6 @@ import (
 	"log"
 	"slices"
 	"strings"
-
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeroles"
 )
 
 func SweepAfterIntegrationTests(client *Client, suffix string) error {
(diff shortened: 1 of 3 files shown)


pkg/sdk/sweepers.go
```diff
@@ -201,7 +203,7 @@ func getRoleSweeper(client *Client, suffix string) func() error {
 			return fmt.Errorf("sweeping roles ended with error, err = %w", err)
 		}
 		for _, role := range roles {
-			if strings.HasSuffix(role.Name, suffix) && !slices.Contains([]string{"ACCOUNTADMIN", "SECURITYADMIN", "SYSADMIN", "ORGADMIN", "USERADMIN", "PUBLIC"}, role.Name) {
+			if strings.HasSuffix(role.Name, suffix) && !slices.Contains([]string{"ACCOUNTADMIN", "SECURITYADMIN", "SYSADMIN", "ORGADMIN", "USERADMIN", "PUBLIC", snowflakeroles.PentestingRole.Name()}, role.Name) {
```
sfc-gh-jmichalak, line 206: Nit: can we use []sdk.AccountObjectIdentifier here with roles from `snowflakeroles` package?

sfc-gh-jmichalak (commented):
1. Looks like docs check fails
2. Nit: this would not matter in this case, but in general we should not cancel tests on PRs with changes in the sweepers.

sfc-gh-asawicki (approved):

pkg/sdk/sweepers.go
sfc-gh-asawicki: I dont think so, currently the sweepers are in SDK pkg so it would be circular dep, there is a TODO for it

sfc-gh-jcieslak: Yup, I tried and got circular dep error

sfc-gh-jmichalak (approved):

PR #3432, opened 2025-03-10
fix: Limit TOML file size

sfc-gh-jmichalak:
- Limit TOML file size to 10MB.
- Move common OS-related functions to a wrapper package.
- Log interactions with OS.

## Test Plan

* [x] acceptance tests

* [ ] …

## TODO
Check the behavior on Windows and adjust the tests.


sfc-gh-jcieslak (approved):

pkg/internal/os/os_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/os"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/testhelpers"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoadConfigFileThatIsTooBig(t *testing.T) {
+	if os.IsRunningOnWindows {
```
sfc-gh-asawicki, line 13: this `if` seems to be not needed (if it works on windows - so let's check the manual tests)

pkg/internal/os/os_test.go
```diff
@@ -0,0 +1,27 @@
+package os_test
+
+import (
```
sfc-gh-asawicki, line 3: nit: all methods should have tests (at least smoke ones to check if the output in simple case is the same). Can be done later.

pkg/internal/os/os.go
```diff
@@ -0,0 +1,57 @@
...
+	"os"
+	"runtime"
+)
+
+const (
+	IsRunningOnWindows = runtime.GOOS == "windows"
+	maxFileSizeInMb    = 10
+)
+
+func Stat(path string) (os.FileInfo, error) {
```
sfc-gh-asawicki, line 15: nit: Let's add a comment to each public method

pkg/provider/provider_acceptance_test.go
```diff
@@ -296,6 +296,31 @@ func TestAcc_Provider_tomlConfig(t *testing.T) {
 	})
 }
 
+func TestAcc_Provider_tomlConfigIsTooBig(t *testing.T) {
+	_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)
+	acc.TestAccPreCheck(t)
+	t.Setenv(string(testenvs.ConfigureClientOnce), "")
+
+	tmpServiceUser := acc.TestClient().SetUpTemporaryServiceUser(t)
```
sfc-gh-asawicki, line 304: nit: we don't need that setup because the size is validated first

pkg/sdk/config.go
```diff
@@ -8,13 +8,13 @@ import (
...
 	"net"
 	"net/url"
-	"os"
 	"path/filepath"
 	"slices"
 	"strings"
 	"time"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/os"
```
sfc-gh-asawicki, line 17: question: should we keep the "os" name of the package? It may be misleading for ppl, maybe it would help to be more explicit here like oslogging or oswrapper or anything like this?
@sfc-gh-jmichalak @sfc-gh-jcieslak

sfc-gh-asawicki (changes requested):

pkg/internal/os/os_test.go
sfc-gh-jmichalak: Probably, my plan is to merge this and permissions and check all at once on a Windows machine. If it's possible, I'll remove that.

pkg/sdk/config.go
sfc-gh-jcieslak: Hmm that's true, especially for someone new to the project, maybe "oswrap" would be enough?

pkg/internal/os/os.go
```diff
@@ -0,0 +1,57 @@
+package os
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"runtime"
+)
+
+const (
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: I think it's correct. I'll check this as an exercise later. I moved this to a function.

pkg/internal/os/os_test.go
sfc-gh-jmichalak: I'll do this later with permissions.

pkg/internal/os/os.go
sfc-gh-jmichalak: Done.

pkg/provider/provider_acceptance_test.go
sfc-gh-jmichalak: Changed.

pkg/sdk/config.go
sfc-gh-jmichalak: Done.

pkg/internal/os/os_test.go
sfc-gh-asawicki: Ok, keeping it open then.

pkg/internal/os/os_test.go
sfc-gh-asawicki: Ok, keeping it open then.

sfc-gh-asawicki (approved):

PR #3444, opened 2025-03-13
fix: Handle TOML file permissions

sfc-gh-jmichalak:
- Check file permissions on non-Windows platforms.
- Add a flag for skipping the verifications.
- Adjust the testing pipeline.
- Add a note about the lack of permissions checking on Windows.
- Add a log entry for loading a specific configuration profile.
- Bump crypto pkg. It was flagged by Snyk.
- Use new config readers in tests.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/manual_tests/windows/files_test.go
```diff
@@ -1,22 +1,21 @@
...
 	"testing"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/oswrapper"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/testhelpers"
 	"github.com/stretchr/testify/require"
 )
 
-func TestLoadConfigFileThatIsTooBig(t *testing.T) {
+func TestReadFileSafeWorksOnWindows(t *testing.T) {
```
sfc-gh-asawicki, line 11: what about the too big test on windows (it was previously done here)?

pkg/sdk/client.go
```diff
@@ -110,7 +111,7 @@ func NewClient(cfg *gosnowflake.Config) (*Client, error) {
 	var err error
 	if cfg == nil {
 		log.Printf("[DEBUG] Searching for default config in credentials chain...\n")
-		cfg = DefaultConfig()
+		cfg = DefaultConfig(oswrapper.ReadFileSafe)
```
sfc-gh-asawicki, line 114: We should be able to parametrize it.

pkg/internal/oswrapper/os.go
```diff
@@ -44,24 +45,59 @@ func ReadFileSafe(path string) ([]byte, error) {
 	return readFile(path)
 }
 
+// ReadFileSkipPermissionsCheck reads a file without checking its permissions.
+func ReadFileSkipPermissionsCheck(path string) ([]byte, error) {
```
sfc-gh-asawicki, line 49: The current logic is duplicated and split (fileIsSafeToRead vs fileIsTooBig; ReadFileSafe vs ReadFileSkipPermissionsCheck). This will get complex with every new validation added. Let's rearrange it, even with the single boolean now (like shouldVerifyPermissions); later, with more validations we can do something more sophisticated than a single boolean but we should not divide this logic like currently done.

pkg/provider/provider_acceptance_test.go
```diff
@@ -322,6 +326,70 @@ func TestAcc_Provider_tomlConfigIsTooBig(t *testing.T) {
...
+	_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)
+	if oswrapper.IsRunningOnWindows() {
+		t.Skip("checking file permissions on Windows is currently done in manual tests package")
+	}
+	acc.TestAccPreCheck(t)
+	t.Setenv(string(testenvs.ConfigureClientOnce), "")
+
+	tmpServiceUser := acc.TestClient().SetUpTemporaryServiceUser(t)
+	profile := random.AlphaN(6)
+	tomlConfig := helpers.FullTomlConfigForServiceUser(t, profile, tmpServiceUser.UserId, tmpServiceUser.RoleId, tmpServiceUser.WarehouseId, tmpServiceUser.AccountId, tmpServiceUser.PrivateKey)
```
sfc-gh-asawicki, line 368: 368-372 should be extracted to helper + use TempTomlConfigForServiceUser instead of FullTomlConfigForServiceUser

nit: I wonder if we need to create the service user here, we could simply copy the existing config with the wider permissions? But maybe it's better to stay with this one.

pkg/provider/provider_acceptance_test.go
```diff
@@ -322,6 +326,70 @@ func TestAcc_Provider_tomlConfigIsTooBig(t *testing.T) {
...
+		ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				PreConfig: func() {
+					t.Setenv(snowflakeenvs.ConfigPath, configPath)
+				},
+				Config: config.FromModels(t, providermodel.SnowflakeProvider().WithProfile(profile).WithSkipTomlFilePermissionVerification(true), datasourceModel()),
```
sfc-gh-asawicki, line 384: nit: extract model

sfc-gh-asawicki (changes requested):

sfc-gh-jcieslak (approved):

pkg/sdk/client.go
sfc-gh-jmichalak: Done.

pkg/provider/provider_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/manual_tests/windows/files_test.go
sfc-gh-jmichalak: It's now being tested in `oswrapper` package. Notice that the test is not os-specific.

pkg/internal/oswrapper/os.go
```diff
@@ -50,16 +51,32 @@ func readFile(path string) ([]byte, error) {
...
+		log.Println("[DEBUG] Skipped checking file permissions on a Windows system...")
+	}
 	return nil
 }
 
+func unixFilePermissionsAreStrict(perm fs.FileMode) bool {
+	log.Println("[DEBUG] Checking file permissions on a Unix system...")
+	unsafeBits := os.FileMode(
+		0o070 | // group has any access
+			0o007, // others have any access
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Changed to `0o077`.

PR #3445, opened 2025-03-14
fix: Fix datasource acceptance tests setup part2

sfc-gh-asawicki:
Continuation of https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3443.

Use the pre-created database or create a database with the parameters set (part 2):
- the pre-created database not changed yet (it still does not have the parameters)
- going alphabetically, last is the grants datasource
- generate model builders for grants datasource
- remove file setups when unnecessary (do not migrate the ones with complex multiresource setups)
- add missing testenvs.EnableAcceptance and acc.TestAccPreCheck(t)
- do not add the builders for non-stable datasources
- added a workaround for single attribute datasources
- remove unused functions

Next PRs:
- datasources continuation
- resources
- pre-created database changes

PR #3443, opened 2025-03-13
fix: Fix datasource acceptance tests setup part1

sfc-gh-asawicki:
Use the pre-created database or create a database with the parameters set (part 1):
- the pre-created database not changed yet (it still does not have the parameters)
- going alphabetically, last is the databases datasource
- generate model builders for database role and database roles
- remove file setups when unnecessary
- extract acceptance tests nonexisting ids (similarly to integration tests)
- add missing `testenvs.EnableAcceptance` and `acc.TestAccPreCheck(t)`
- do not add the builders for non-stable datasources

Next PRs:
- datasources continuation
- resources
- pre-created database changes

PR #3450, opened 2025-03-18
fix: Fix resource acceptance tests setup part1

sfc-gh-asawicki:
Continuation of https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3443, https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3445, https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3448, and https://github.com/Snowflake-Labs/terraform-provider-snowflake/pull/3449.

Use the pre-created database or create a database with the parameters set (resource - part1):
- the pre-created database not changed yet (it still does not have the parameters)
- going alphabetically through resources, last one was grant ownership
- generate model builders for AccountRole resource
- remove file setups when unnecessary (do not migrate the ones with complex multiresource setups)
- remove database from setups where not needed
- add missing testenvs.EnableAcceptance and acc.TestAccPreCheck(t)
- do not add the builders for non-stable datasources
- skip tests in the database resource and in external_oauth_integration_acceptance_test.go (follow-up needed for these to discuss and address all the nondeterministic tests)
- remove tests for parameters that can be set on other level than account (in pkg/resources/account_parameter_acceptance_test.go)

Next PRs:
- resources continuation (1 or 2 more PRs)
- pre-created database changes
- follow-up for the problematic ones


pkg/resources/execute_acceptance_test.go
```diff
@@ -675,7 +698,6 @@ func TestAcc_Execute_grantsComplex(t *testing.T) {
 // proves https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2491
 func TestAcc_Execute_queryResultsBug(t *testing.T) {
 	resourceName := "snowflake_execute.test"
```
sfc-gh-jmichalak, line 700: Why no `	_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)` here?

pkg/resources/external_oauth_integration_acceptance_test.go
```diff
@@ -418,16 +427,25 @@ func TestAcc_ExternalOauthIntegration_completeWithJwsKeysUrlAndAllowedRolesList(
 	})
 }
 
+// TODO [SNOW-1991414]: unskip
 func TestAcc_ExternalOauthIntegration_completeWithRsaPublicKeysAndBlockedRolesList_paramSet(t *testing.T) {
-	id := acc.TestClient().Ids.RandomAccountObjectIdentifier()
+	t.Skip("Skipping this test temporarily as it messes with the account settings which potentially affect all other tests in this file.")
```
sfc-gh-jmichalak, line 432: Q: how about keeping a list of skipped tests? Could be in the ticket description with a reason.

pkg/resources/failover_group_acceptance_test.go
```diff
@@ -97,10 +103,13 @@ func TestAcc_FailoverGroupRemoveObjectTypes(t *testing.T) {
 }
 
 func TestAcc_FailoverGroupInterval(t *testing.T) {
+	_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)
+	acc.TestAccPreCheck(t)
+
 	// TODO [SNOW-1002023]: Unskip; Business Critical Snowflake Edition needed
 	_ = testenvs.GetOrSkipTest(t, testenvs.TestFailoverGroups)
```
sfc-gh-jmichalak, line 110: This check should be the first one to avoid creating the test objects. Same in other tests.

pkg/resources/grant_account_role_acceptance_test.go
```diff
@@ -197,3 +215,23 @@ func TestAcc_GrantAccountRole_IdentifierQuotingDiffSuppression(t *testing.T) {
 		},
 	})
 }
+
+func grantAccountRoleConfigWithQuotedIdentifiers(roleId sdk.AccountObjectIdentifier, parentRoleId sdk.AccountObjectIdentifier) string {
```
sfc-gh-jmichalak, line 219: We are testing only grants here, so why don't we create the account roles before the test, like in others?

pkg/resources/database_acceptance_test.go
```diff
@@ -26,7 +29,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TODO [SNOW-1991414]: discuss and address all the nondeterministic tests in this file
+// For now, this test can sometimes fail (if account parameters are changed in the meantime).
+// We could set the known parameters here, however, we need to test behavior for the database when they are not set.
+// We could try ignoring the changes to parameters too.
```
sfc-gh-jmichalak, line 35: I'm not sure if ignoring the changes alone will work. We still make assertions against parameter level, so this would need to be adjusted as well, right?

sfc-gh-jmichalak (changes requested):

pkg/resources/grant_account_role_acceptance_test.go
sfc-gh-jcieslak: The same could be done for database roles, but we are using top-level functions, which complicate adding common objects (instead of using one function with sub-tests as `t.Run()`s).

sfc-gh-jcieslak (approved):

pkg/resources/execute_acceptance_test.go
sfc-gh-asawicki: There is no setup outside of `t.Run()`, so I did not add this check (it was the same in the previous PRs)

pkg/resources/failover_group_acceptance_test.go
sfc-gh-asawicki: hmmm, I was thinking a bit differently here: when we introduce the common body/framework or just common TestMain for our tests, the enableacceptance and precheck would be first things to always run. That's why I kept this in this ordering. But you are right, objects in this test will be created even without the `TestFailoverGroups`.

I can reverse the ordering but I would rather do it in all tests as this ordering was also preserved in the previous PRs. Is it okay?

pkg/resources/grant_account_role_acceptance_test.go
sfc-gh-asawicki: I have just preserved the previously existing tests where the quoted and unquoted objects were used as dependencies. And this is the reason why they were left in the config and not precreated.

pkg/resources/database_acceptance_test.go
sfc-gh-asawicki: yes, all the options require more development but the discussions first

pkg/resources/execute_acceptance_test.go
```diff
@@ -867,7 +895,6 @@ func testAccCheckDatabaseExistence(t *testing.T, id sdk.AccountObjectIdentifier,
 
 // Result of https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/3334.
 func TestAcc_Execute_gh3334_allTimeouts(t *testing.T) {
-	resourceName := "snowflake_execute.test"
```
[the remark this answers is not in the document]
sfc-gh-asawicki: There is no setup outside of `t.Run()`, so I did not add this check (it was the same in the previous PRs)

pkg/resources/external_oauth_integration_acceptance_test.go
sfc-gh-jmichalak: I meant the tests we skip because they aren't stable, not all skipped ones. I'm just curious if we're tracking them properly.

sfc-gh-jmichalak (dismissed):

pkg/resources/external_oauth_integration_acceptance_test.go
sfc-gh-asawicki: I will prepare such a list after this series of PRs if, after discussions, there will still be some skipped tests. Is that ok?

Issue #3390, opened 2025-02-10
[Bug]: Grace Period should NOT be mandatory - Unable to import existing accounts


sivadotblog:
### Terraform CLI Version

1.6.0

### Terraform Provider Version

1.0.3

### Company Name

Humana

### Terraform Configuration

```terraform
Terraform Configuration*
 
# Complete (with SERVICE user type)
resource "snowflake_account" "account" {
 
  name                 = var.account_name
  admin_name           = var.admin_name
  admin_rsa_public_key = tls_private_key.rsa.public_key_pem
  admin_user_type      = "SERVICE"
  email                = var.account_email
  edition              = var.edition
  region_group         = "PUBLIC"
  region               = var.snowflake_region
  comment              = "built by Terraform"
  is_org_admin         = var.is_org_admin
  grace_period_in_days = 90
 
  depends_on = [tls_private_key.rsa]
}
 
import {
  to = resource.snowflake_account.account
  id = "XXX.AZ_SNOWFLAKE_CENTERWELL_EASTUS2_SBX"
}
```

### Category

category:resource

### Expected Behavior

 
Expected Behavior*
I expect the account to be successfully imported into terraform state
 
Actual Behavior*
get the following error when trying to import
 
Error: 001817 (42601): SQL compilation error: 
│ parameter value out of range: GRACE_PERIOD_IN_DAYS. Must be between 3 and 90.
 

run the code
 

### Actual Behavior

account imported successfully

### Steps to Reproduce

Steps to Reproduce*
create a snowflake account manually
copy the hcl configuration
update the account name in import
 

### How much impact is this issue causing?

Low

### Additional Information

From: Support [support@snowflake.com]
Sent: 1/30/2025 1:37 PM

Hello Pradeep,
 
Thank you for writing back.

Indeed there is no parameter ‘grace_period_in_days’, from Snowflake, it can be set only with DROP statement.

However, a possible reason why you receive the below error;

Error: 001817 (42601): SQL compilation error:
? parameter value out of range: GRACE_PERIOD_IN_DAYS. Must be between 3 and 90.

Is because when you created a Snowflake account using Terraform you did not specify  ‘grace_period_in_days’ (0/null) / or specified a different number.

When you then attempted to update the account name in import script with ‘grace_period_in_days’ as 90, the ‘grace_period_in_days’ value mismatched the previous value set (either a null value/some value) when creating the account.

[...]

PR #3451, opened 2025-03-19
fix: Adjust docs

sfc-gh-jmichalak:
- Add the note about aggressive upgrades and adjust the resource migration guide.
- Use title case in the guides.
- Add a note about recreating accounts and grace period.
- Add a note about pasting logs.
- Fix typos.

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

#3390

### DIFF
--- a/templates/resources/account.md.tmpl
+++ b/templates/resources/account.md.tmpl
@@ -17,10 +17,9 @@ description: |-
 
 ~> **Note** Changes for the following fields won't be detected: `admin_name`, `admin_password`, `admin_rsa_public_key`, `admin_user_type`, `first_name`, `last_name`, `email`, `must_change_password`. This is because these fields only supply initial values for creating the admin user. Once the account is created, the admin user becomes an independent entity. Modifying users from the account resource is challenging since it requires logging into that account. This would require the account resource logging into the account it created to read or alter admin user properties, which is impractical, because any external change to the admin user would disrupt the change detection anyway.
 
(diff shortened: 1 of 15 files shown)


templates/resources/account.md.tmpl
```diff
@@ -17,6 +17,11 @@ description: |-
 
 ~> **Note** Changes for the following fields won't be detected: `admin_name`, `admin_password`, `admin_rsa_public_key`, `admin_user_type`, `first_name`, `last_name`, `email`, `must_change_password`. This is because these fields only supply initial values for creating the admin user. Once the account is created, the admin user becomes an independent entity. Modifying users from the account resource is challenging since it requires logging into that account. This would require the account resource logging into the account it created to read or alter admin user properties, which is impractical, because any external change to the admin user would disrupt the change detection anyway.
 
+~> **Note** In order to drop the account successfully, you need to specify `grace_period_days` field (see [Snowflake docs](https://docs.snowflake.com/en/sql-reference/sql/drop-account)). This affects Terraform plans with changes on a `ForceNew` fields, causing the resource to be recreated. If you do not want to recreate the resource:
```
sfc-gh-jcieslak, line 20: Do we want to mention the import thing? `grace_period_days` is a required field, so it has to be set anyway and the recommendations are pretty general (could be defined somewhere else because it applies to all resources). If that's regarding the import usage error, I would change this to a warning, mention recreation may be hard right after import, and maybe link some of the already existing/resolved issues.

sfc-gh-jmichalak: You're right, I rephrased this one and added an entry in the resource migration guide (importing section).

sfc-gh-jcieslak (approved):

PR #3452, opened 2025-03-19
chore: Update documentation to include default values

sfc-gh-jcieslak:
## Changes
- Added a function that modifies the schema description by adding default values right after the type description (it couldn't get it into the same parentheses as type).
- Unfortunately, provider documentation couldn't be modified. (I'll still look a bit, if there's the same function for provider schema).

## Next steps
- Pipelines may fail now, because I only included a few generated files for easy review. I'll add the rest of them after first review.
- Adjust custom templates for every resource
- Check data sources
- Apply the same rules to the provider configuration


templates/resources.md.tmpl
```diff
@@ -18,13 +18,15 @@ description: |-
...
-{{ tffile .ExampleFile }}
-
 -> **Note** Instead of using fully_qualified_name, you can reference objects managed outside Terraform by constructing a correct ID, consult [identifiers guide](../guides/identifiers_rework_design_decisions#new-computed-fully-qualified-name-field-in-resources).
 <!-- TODO(SNOW-1634854): include an example showing both methods-->
 
+{{ tffile .ExampleFile }}
+
 {{- end }}
 
+-> **Note** If a field has default value, it is displayed next to the type in the schema. If the default is computed from external sources (e.g., environment variables), it displays `Default is computed`.
```
sfc-gh-jmichalak, line 28: Nit: I'd add this note at the beginning, so the users can read about it before seeing these notes. Also, remember to adjust the custom templates.

docs/resources/function_sql.md
```diff
@@ -80,9 +80,9 @@ EOT
...
+- `is_secure` (String) (Default: `default`) Specifies that the function is secure. By design, the Snowflake's `SHOW FUNCTIONS` command does not provide information about secure functions (consult [function docs](https://docs.snowflake.com/en/sql-reference/sql/create-function#id1) and [Protecting Sensitive Information with Secure UDFs and Stored Procedures](https://docs.snowflake.com/en/developer-guide/secure-udf-procedure)) which is essential to manage/import function with Terraform. Use the role owning the function while managing secure functions. Available options are: "true" or "false". When the value is not set in the configuration the provider will put "default" there which means to use the Snowflake default for this value.
```
sfc-gh-jmichalak, line 85: Hmm, I'm wondering if we could be more explicit here. We can mention this is not set by us but only in Snowflake or remove it. The same goes for `-1` values. Also, we can link https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/v1-preparations/CHANGES_BEFORE_V1.md#default-values in the template notes.

sfc-gh-jcieslak: added a warning above (it's in the documentation for alert)

templates/resources.md.tmpl
sfc-gh-jcieslak: It's above the schema. Added a point in "next steps".

docs/resources/function_sql.md
sfc-gh-jmichalak: Sounds good.

sfc-gh-jcieslak: Changed

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

PR #3453, opened 2025-03-19
fix: Add boolean env validations and unit tests for TOML config validation

sfc-gh-jmichalak:
- Add TOML validation unit tests.
- Validate boolean env variables.
- Add a link to TOML reference.
- Use testhelper in other tests.

## Test Plan

* [ ] acceptance tests

* [ ] …


pkg/provider/provider.go
```diff
@@ -474,7 +474,16 @@ func getResources() map[string]*schema.Resource {
 		"snowflake_warehouse":                                                    resources.Warehouse(),
 	}
 
-	if oswrapper.Getenv(string(testenvs.EnableObjectRenamingTest)) != "" {
+	accTestEnabled, err := oswrapper.GetenvBool("TF_ACC")
+	if err != nil {
+		log.Panicf("TF_ACC environmental variable has incorrect format: %v", err)
```
sfc-gh-jcieslak, line 479: Hmm, those two env variables are used only for tests. I don't think we should panic even if their format is incorrect as this only affect tests. Same in changes below.

sfc-gh-jmichalak: To me we can fail here because simply the provider is not configured properly. If you misconfigured TOML or TF file, the provider would fail, so I don't think we should have a different behavior here, like setting a default.

sfc-gh-asawicki: These two variables should not end up in the provider production code either way. Both options are fine with me, because we will remove them from here really soon. This potentially changes the current behavior of the provider so I would lean a bit on not panicking here).

sfc-gh-asawicki (changes requested):

pkg/provider/provider.go
sfc-gh-jmichalak: Changed - now we're using `log.Printf` with an info about a default value.

sfc-gh-jcieslak (approved):

PR #3455, opened 2025-03-20
chore: Update GitHub actions

sfc-gh-jcieslak:
## Changes
- Add workflow that validates other workflows. For now, only one rule is applied.
- Fix codeql push event

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,6 +2,9 @@ name: CodeQL
 
 on:
   push:
+    branches:
+      - main
+      - dev
   pull_request:
   schedule:
     - cron: "0 2 * * *"
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,10 +1,6 @@
 name: CodeQL
 
 on:
-  push:
-    branches:
-      - main
-      - dev
   pull_request:
   schedule:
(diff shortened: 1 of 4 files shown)


.github/workflows/workflow-rules.yml
```diff
@@ -0,0 +1,21 @@
...
+    name: Check workflows
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Check for pull_request_target
+        run: |
+          if grep -r --exclude="workflow-rules.yml" "pull_request_target" .github/workflows/; then
+            echo "pull_request_target found in workflow files"
```
sfc-gh-jmichalak, line 17: Nit: it would be great to print what file exactly.

.github/workflows/codeql-analysis.yml
```diff
@@ -2,6 +2,9 @@ name: CodeQL
 
 on:
   push:
+    branches:
+      - main
+      - dev
```
sfc-gh-jmichalak, line 7: Nit: hmm, isn't `pull_request` enough? We don't push to these branches anyway. We could also discuss if we need `schedule`.

.github/workflows/workflow-rules.yml
sfc-gh-terraform-provider-sa: Grep does it automatically, this is just an additional information

PR #3483, opened 2025-03-26
chore: Add a note about default values before every schema markdown 

sfc-gh-jcieslak:
## Changes
- Follow-up for #3452 (adding a note about default values in all custom templates)


docs/resources/account.md
```diff
@@ -91,6 +91,10 @@ variable "last_name" {
 -> **Note** Instead of using fully_qualified_name, you can reference objects managed outside Terraform by constructing a correct ID, consult [identifiers guide](../guides/identifiers_rework_design_decisions#new-computed-fully-qualified-name-field-in-resources).
 <!-- TODO(SNOW-1634854): include an example showing both methods-->
 
+-> **Note** If a field has default value, it is displayed next to the type in the schema. If the default is computed from external sources (e.g., environment variables), it displays `Default is computed`.
+
+-> **Note** If a field has default value, it is displayed next to the type in the schema. If the default is computed from external sources (e.g., environment variables), it displays `Default is computed`.
```
sfc-gh-jmichalak, line 96: The text is duplicated.

templates/resources/warehouse.md.tmpl
```diff
@@ -28,7 +28,8 @@ description: |-
 
 {{- end }}
 
-{{ .SchemaMarkdown | trimspace }}
```
sfc-gh-jmichalak, line 31: Why is this removed?

docs/index.md
```diff
@@ -73,63 +72,65 @@ provider "snowflake" {
...
-- `okta_url` (String) The URL of the Okta server. e.g. https://example.okta.com. Okta URL host needs to to have a suffix `okta.com`. Read more in Snowflake [docs](https://docs.snowflake.com/en/user-guide/oauth-okta). Can also be sourced from the `SNOWFLAKE_OKTA_URL` environment variable.
-- `organization_name` (String) Specifies your Snowflake organization name assigned by Snowflake. For information about account identifiers, see the [Snowflake documentation](https://docs.snowflake.com/en/user-guide/admin-account-identifier#organization-name). Required unless using `profile`. Can also be sourced from the `SNOWFLAKE_ORGANIZATION_NAME` environment variable.
+- `account_name` (String) (Default is computed) Specifies your Snowflake account name assigned by Snowflake. For information about account identifiers, see the [Snowflake documentation](https://docs.snowflake.com/en/user-guide/admin-account-identifier#account-name). Required unless using `profile`. Can also be sourced from the `SNOWFLAKE_ACCOUNT_NAME` environment variable.
```
sfc-gh-jmichalak, line 82: I don't think this is true or informative. For example, if a user doesn't pass this field, we don't compute it. How about something like `fallback to Snowflake Go driver default...`?

Note that we already have information about env - should we add TOML file as well?

sfc-gh-jmichalak (changes requested):

templates/resources/warehouse.md.tmpl
sfc-gh-jcieslak: Fixed

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #3491, opened 2025-03-28
chore: Handle read operation safely

sfc-gh-jcieslak:
## Changes
- Added a common read function that is safely calling show by id, and in case of any issues with querying the object, it checks higher hierarchy objects to see if the object is connected to them being missing or not.
- New function is intended to be used inside newly introduced ShowByIDSafely functions on the SDK level.
- I have added a new common Snowflake error because we can use it with a combination of other errors to check whether the underlying database or schema objects are missing.
- I left the client changes (I was using them in the previous approach) because they're following the conventions of other testing clients.

## Next changes
- Apply new show by id function to all resources / data sources.
- Adjust SDK generator templates to include new function (or add readme to do it later)
- Check object renaming tests and adjust with current Read handling


pkg/resources/common.go
```diff
@@ -206,3 +208,120 @@ func PreviewFeatureDeleteContextWrapper(featureRaw string, deleteFunc schema.Del
 		return deleteFunc(ctx, d, meta)
 	}
 }
+
+func SafeShowById[T any, ID sdk.AccountObjectIdentifier | sdk.DatabaseObjectIdentifier | sdk.SchemaObjectIdentifier | sdk.SchemaObjectIdentifierWithArguments](
```
sfc-gh-asawicki, line 212: Why is this function added on the resource level and not the SDK level? I mean, it makes more sense for it to be inside the SDK, so that any caller (like in our case resources and datasources) does not need to worry about the underlying SF behavior. And this affects some points from the PR description (like turning into acceptance tests/exporting the client, etc.)

sfc-gh-asawicki (changes requested):
Let's discuss this, I don't leave any more comments for now.

pkg/sdk/testint/show_by_id_integration_test.go
```diff
@@ -0,0 +1,131 @@
...
+	t.Cleanup(cleanupNetworkPolicy)
+
+	value, err := sdk.SafeShowById(testClient(t), networkPolicyShowById, testContext(t), networkPolicy.ID())
+	assert.NotNil(t, value)
+	assert.NoError(t, err)
+
+	cleanupNetworkPolicy()
+
+	value, err = sdk.SafeShowById(testClient(t), networkPolicyShowById, testContext(t), networkPolicy.ID())
+	assert.Nil(t, value)
```
sfc-gh-jmichalak, line 26: Nit: there is no need to assert values when we expect errors anyway. Same in other functions.

pkg/sdk/testint/show_by_id_integration_test.go
```diff
@@ -0,0 +1,131 @@
...
+	assert.ErrorIs(t, err, sdk.ErrObjectNotFound)
+}
+
+func TestInt_SafeShowByIdOnDatabaseObjectIdentifier(t *testing.T) {
+	databaseRoleShowById := func(ctx context.Context, id sdk.DatabaseObjectIdentifier) (*sdk.DatabaseRole, error) {
+		return testClientHelper().DatabaseRole.Show(t, id)
+	}
+
+	databaseRole, cleanupDatabaseRole := testClientHelper().DatabaseRole.CreateDatabaseRole(t)
```
sfc-gh-jmichalak, line 35: Missing t.Cleanup. In other places as well.

pkg/sdk/show_by_id.go
```diff
@@ -0,0 +1,78 @@
...
+		return result, err
+	}
+
+	if err != nil {
+		var zeroValue T
+		errs := []error{err}
+
+		// ErrObjectNotExistOrAuthorized or ErrDoesNotExistOrOperationCannotBePerformed can only happen
+		// when the higher hierarchy object is not accessible for some reason during the "main" showById.
+		shouldCheckHigherHierarchies := errors.Is(err, ErrObjectNotExistOrAuthorized) || errors.Is(err, ErrDoesNotExistOrOperationCannotBePerformed)
```
sfc-gh-jmichalak, line 29: This logic can be simplified - if it's false we can just handle AccountObjectIdentifier  separately (it does not have any hierarchy anyway) and `return zeroValue, errors.Join(errs...)`

pkg/sdk/show_by_id.go
```diff
@@ -0,0 +1,78 @@
...
+	if errors.Is(err, ErrObjectNotFound) {
+		return result, err
+	}
+
+	if err != nil {
+		var zeroValue T
+		errs := []error{err}
+
+		// ErrObjectNotExistOrAuthorized or ErrDoesNotExistOrOperationCannotBePerformed can only happen
+		// when the higher hierarchy object is not accessible for some reason during the "main" showById.
```
sfc-gh-jmichalak, line 28: Nit: can we have a link to docs, if possible?

pkg/sdk/show_by_id.go
```diff
@@ -0,0 +1,78 @@
...
+			return result, err
+		case DatabaseObjectIdentifier:
+			if shouldCheckHigherHierarchies {
+				if _, err := client.Databases.ShowByID(ctx, id.DatabaseId()); err != nil {
+					errs = append(errs, err)
+				}
+			}
+
+			return zeroValue, errors.Join(errs...)
+		case SchemaObjectIdentifier:
```
sfc-gh-jmichalak, line 42: That's the same as SchemaObjectIdentifierWithArguments, right?

sfc-gh-jmichalak (changes requested):

pkg/sdk/database_role_impl.go
```diff
@@ -44,13 +44,15 @@ func (v *databaseRoles) Show(ctx context.Context, request *ShowDatabaseRoleReque
...
 
 func (v *databaseRoles) ShowByID(ctx context.Context, id DatabaseObjectIdentifier) (*DatabaseRole, error) {
-	request := NewShowDatabaseRoleRequest(id.DatabaseId()).WithLike(Like{Pointer(id.Name())})
-	databaseRoles, err := v.Show(ctx, request)
-	if err != nil {
-		return nil, err
-	}
-
-	return collections.FindFirst(databaseRoles, func(r DatabaseRole) bool { return r.Name == id.Name() })
+	return SafeShowById(v.client, func(ctx context.Context, id DatabaseObjectIdentifier) (*DatabaseRole, error) {
```
sfc-gh-asawicki, line 47: I struggle here. One option is to leave it like this, which makes this behavior a default but is unnecessarily complex. The alternative is to leave the old ShowById intact and use this wrapper externally (on the caller side only) or extract a wrapping method in each interface (which would use the new wrapper + existing ShowById which makes it easy). It also allows the caller to pick their intended behavior - aligned with SF or improved. We also need to think about the solution's generativeness, as our SDK is generated to some extent.

Keeping it in mind I would prefer changing it (I don't have a strong preference yet over wrapper on the called side or the new method in the interface). WDYT? (we can sync internally first)

pkg/resources/task_acceptance_test.go
```diff
@@ -824,7 +824,8 @@ func TestAcc_Task_CallingProcedure(t *testing.T) {
 	_ = testenvs.GetOrSkipTest(t, testenvs.EnableAcceptance)
 	acc.TestAccPreCheck(t)
 
-	procedure := acc.TestClient().Procedure.Create(t, sdk.DataTypeNumber)
+	procedure, cleanupProcedure := acc.TestClient().Procedure.Create(t, sdk.DataTypeNumber)
```
sfc-gh-asawicki, line 827: let's add a test showing that the currently introduced showbyid behavior works better than previously (with the removal of one of the hierarchy levels)

sfc-gh-asawicki (changes requested):

pkg/sdk/testint/show_by_id_integration_test.go
sfc-gh-jcieslak: Adjusted

pkg/sdk/testint/show_by_id_integration_test.go
sfc-gh-jcieslak: Added

pkg/sdk/show_by_id.go
sfc-gh-jcieslak: Made the logic a bit shorter.

pkg/sdk/show_by_id.go
sfc-gh-jcieslak: Initially wanted to leave it as is, but in the end merged both cases

pkg/resources/common.go
sfc-gh-jcieslak: Yup, after discussion the implementation was moved to the SDK.

pkg/sdk/database_role_impl.go
sfc-gh-jcieslak: After discussion, adjusted the implementation to be handled by a separate function in the sdk interfaces

pkg/resources/task_acceptance_test.go
sfc-gh-jcieslak: Added in table acc tests, I skipped database and schema because they were already handling it well. Added to the table acc tests, but also I would like to check the object renaming tests which are usually skipped, because they should currently fail (added to the next steps list)

pkg/resources/table_acceptance_test.go
```diff
@@ -2260,3 +2262,49 @@ resource "snowflake_table" "test_table" {
...
+				},
+				PreConfig:   func() { schemaCleanup() },
+				Config:      tableConfig(tableId.Name(), tableId.DatabaseName(), tableId.SchemaName()),
+				ExpectError: regexp.MustCompile("object does not exist or not authorized"),
+			},
+			// The New version removes table from the state which ends up with create operation
+			// and the "error creating table" error (because of the missing underlying schema).
+			{
+				ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+				Config:                   tableConfig(tableId.Name(), tableId.DatabaseName(), tableId.SchemaName()),
```
sfc-gh-asawicki, line 2305: there will be a conflict on this method as it has different signature now (only id)

pkg/resources/table_acceptance_test.go
```diff
@@ -2260,3 +2262,49 @@ resource "snowflake_table" "test_table" {
...
+		CheckDestroy: acc.CheckDestroy(t, resources.Table),
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"snowflake": {
+						VersionConstraint: "=1.0.5",
+						Source:            "Snowflake-Labs/snowflake",
+					},
+				},
+				Config: tableConfig(tableId.Name(), tableId.DatabaseName(), tableId.SchemaName()),
```
sfc-gh-asawicki, line 2288: Let's assert something here

pkg/sdk/testint/procedures_integration_test.go
```diff
@@ -2170,23 +2170,44 @@ def filter_by_role(session, table_name, role):
...
+		procedureId := testClientHelper().Ids.RandomSchemaObjectIdentifierWithArgumentsInSchema(schemaId)
+		_, err := client.Procedures.ShowByIDSafely(ctx, procedureId)
 		assert.Error(t, err)
 		assert.ErrorIs(t, err, sdk.ErrObjectNotFound)
 		assert.ErrorIs(t, err, sdk.ErrObjectNotExistOrAuthorized)
 		assert.ErrorIs(t, err, sdk.ErrDoesNotExistOrOperationCannotBePerformed)
 	})
 
-	t.Run("show procedure by id - missing schema", func(t *testing.T) {
+	t.Run("show by id safely - missing schema", func(t *testing.T) {
```
sfc-gh-asawicki, line 2207: nit: one more test with missing procedure?

pkg/sdk/testint/tables_integration_test.go
```diff
@@ -976,24 +977,49 @@ func TestInt_TablesShowByID(t *testing.T) {
...
+			databaseId := testClientHelper().Ids.RandomAccountObjectIdentifier()
+			schemaId := testClientHelper().Ids.RandomDatabaseObjectIdentifierInDatabase(databaseId)
+			tableId := testClientHelper().Ids.RandomSchemaObjectIdentifierInSchema(schemaId)
+			_, err := client.Tables.ShowByIDSafely(ctx, tableId)
+			assert.Error(t, err)
+			assert.ErrorIs(t, err, sdk.ErrObjectNotFound)
+			assert.ErrorIs(t, err, sdk.ErrDoesNotExistOrOperationCannotBePerformed)
+		})
+
+		t.Run("show by id safely: missing schema", func(t *testing.T) {
```
sfc-gh-asawicki, line 1016: nit: one more test with missing table for safely?

sfc-gh-asawicki (changes requested):
Please also add an action point (and in SDK generator readme) to add this new interface method (it will be easier to generate it everywhere)

pkg/sdk/testint/tables_integration_test.go
sfc-gh-jcieslak: Added

pkg/resources/table_acceptance_test.go
sfc-gh-jcieslak: Added plan and `id` assert

pkg/resources/table_acceptance_test.go
```diff
@@ -3,10 +3,13 @@ package resources_test
 import (
 	"context"
 	"fmt"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
 	"regexp"
 	"strconv"
 	"testing"
 
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvs"
+
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Fixed

pkg/resources/table_acceptance_test.go
sfc-gh-jcieslak: Adjusted

pkg/resources/table_acceptance_test.go
```diff
@@ -2260,3 +2263,60 @@ resource "snowflake_table" "test_table" {
...
+		},
+		CheckDestroy: acc.CheckDestroy(t, resources.Table),
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders: map[string]resource.ExternalProvider{
+					"snowflake": {
+						VersionConstraint: "=1.0.5",
+						Source:            "Snowflake-Labs/snowflake",
+					},
+				},
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

pkg/sdk/procedures_gen.go
```diff
@@ -18,6 +18,7 @@ type Procedures interface {
 	Drop(ctx context.Context, request *DropProcedureRequest) error
 	Show(ctx context.Context, request *ShowProcedureRequest) ([]Procedure, error)
 	ShowByID(ctx context.Context, id SchemaObjectIdentifierWithArguments) (*Procedure, error)
+	ShowByIDSafely(ctx context.Context, id SchemaObjectIdentifierWithArguments) (*Procedure, error)
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I'm also not that satisfied with the name, but:
a) for now, we are the only user of the SDK (we know what this function does)
b) even once SDK will be it's own thing, I think a simple go doc describing the function's purpose should make the developer aware of the difference and why it may be considered a safer alternative (I can describe it in the following pr)
c) for now we don't have any other better name that would replace the current one.

Idk, I would go with this name, at least for now.

pkg/sdk/show_by_id.go
```diff
@@ -16,58 +16,44 @@ func SafeShowById[T any, ID AccountObjectIdentifier | DatabaseObjectIdentifier |
...
-					if errors.Is(err, ErrObjectNotFound) {
-						return zeroValue, errors.Join(errs...)
-					}
-				}
+			if _, err := client.Schemas.ShowByID(ctx, schemaObjectId.SchemaId()); err != nil {
+				errs = append(errs, err)
 
-				if _, err := client.Databases.ShowByID(ctx, id.DatabaseId()); err != nil {
-					errs = append(errs, err)
+				if errors.Is(err, ErrObjectNotFound) {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: At this point, it's already checking higher hierarchies, so there's no point in checking `shouldCheckHigherHierarchies` (it's checked higher).

pkg/sdk/procedures_gen.go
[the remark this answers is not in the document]
sfc-gh-asawicki: I'm fine with both.

sfc-gh-asawicki (approved):

pkg/sdk/show_by_id.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: Yeah, but Schemas.ShowByID can return a separate error on the missing database. It would be handled below, but it's a bit inconsistent.

sfc-gh-jmichalak (approved):

pkg/sdk/errors.go
```diff
@@ -22,6 +22,7 @@ var (
 	ErrObjectNotExistOrAuthorized               = NewError("object does not exist or not authorized")
 	ErrDoesNotExistOrOperationCannotBePerformed = NewError("object does not exist, or operation cannot be performed")
 	ErrAccountIsEmpty                           = NewError("account is empty")
+	ErrGrantPartiallyExecuted     = NewError("grant partially executed")
```
sfc-gh-jmichalak, line 25: gofmt?

sfc-gh-jmichalak (approved):

pkg/sdk/errors.go
sfc-gh-jcieslak: ;_____;

sfc-gh-jmichalak (approved):

PR #3545, opened 2025-04-10
chore: Masking random values

sfc-gh-jcieslak:
## Changes
- Updated other random helpers to include masking string
- Used masking string in login names and passwords
- Replaced the use of AdminName in places where it wasn't admin name, but rather an identifier or other value that should be masked.

### DIFF
--- a/pkg/acceptance/helpers/ids_generator.go
+++ b/pkg/acceptance/helpers/ids_generator.go
@@ -43,6 +43,10 @@ func (c *IdsGenerator) RandomAccountObjectIdentifier() sdk.AccountObjectIdentifi
 	return sdk.NewAccountObjectIdentifier(c.Alpha())
 }
 
+func (c *IdsGenerator) SensitiveRandomAccountObjectIdentifier() sdk.AccountObjectIdentifier {
+	return sdk.NewAccountObjectIdentifier(random.SensitiveAlphanumeric())
+}
+
 func (c *IdsGenerator) RandomAccountObjectIdentifierWithPrefix(prefix string) sdk.AccountObjectIdentifier {
 	return sdk.NewAccountObjectIdentifier(c.AlphaWithPrefix(prefix))
 }
--- a/pkg/acceptance/helpers/ids_generator.go
+++ b/pkg/acceptance/helpers/ids_generator.go
@@ -44,7 +44,7 @@ func (c *IdsGenerator) RandomAccountObjectIdentifier() sdk.AccountObjectIdentifi
 }
 
 func (c *IdsGenerator) RandomSensitiveAccountObjectIdentifier() sdk.AccountObjectIdentifier {
-	return sdk.NewAccountObjectIdentifier(random.SensitiveAlphanumeric())
+	return c.RandomAccountObjectIdentifierWithPrefix(random.SensitiveAlphanumeric())
 }
 
 func (c *IdsGenerator) RandomAccountObjectIdentifierWithPrefix(prefix string) sdk.AccountObjectIdentifier {
--- a/pkg/acceptance/helpers/ids_generator.go
+++ b/pkg/acceptance/helpers/ids_generator.go
@@ -43,7 +43,7 @@ func (c *IdsGenerator) RandomAccountObjectIdentifier() sdk.AccountObjectIdentifi
 	return sdk.NewAccountObjectIdentifier(c.Alpha())
 }
 
-func (c *IdsGenerator) SensitiveRandomAccountObjectIdentifier() sdk.AccountObjectIdentifier {
+func (c *IdsGenerator) RandomSensitiveAccountObjectIdentifier() sdk.AccountObjectIdentifier {
 	return sdk.NewAccountObjectIdentifier(random.SensitiveAlphanumeric())
 }
 
--- a/pkg/acceptance/helpers/random/random_helpers.go
+++ b/pkg/acceptance/helpers/random/random_helpers.go
@@ -11,7 +11,7 @@ import (
 )
 
 // generatedRandomValue is used to mask random values in GitHub Action logs.
-// It always starts with a letter and contains only letters and numbers.
+// It always starts with a letter and contains only letters.
 var generatedRandomValue string
 
 func init() {
@@ -36,7 +36,7 @@ func Comment() string {
 // 090088 (22000): ADMIN_NAME can only contain letters, numbers and underscores.
 // 090089 (22000): ADMIN_NAME must start with a letter.
 func AdminName() string {
-	return SensitiveAlphanumeric()
+	return SensitiveAlpha()
 }
 
 func Email() string {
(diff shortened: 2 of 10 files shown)


pkg/acceptance/helpers/ids_generator.go
```diff
@@ -43,6 +43,10 @@ func (c *IdsGenerator) RandomAccountObjectIdentifier() sdk.AccountObjectIdentifi
 	return sdk.NewAccountObjectIdentifier(c.Alpha())
 }
 
+func (c *IdsGenerator) RandomSensitiveAccountObjectIdentifier() sdk.AccountObjectIdentifier {
+	return sdk.NewAccountObjectIdentifier(random.SensitiveAlphanumeric())
```
sfc-gh-asawicki, line 47: we should still have test object suffix for this id

pkg/acceptance/helpers/random/random_helpers.go
```diff
@@ -31,15 +32,31 @@ func Comment() string {
...
+// SensitiveString returns a random string prefixed with a generated random value that is masked in GitHub Action logs.
+// The string returned by SensitiveString always starts with a letter and contains only letters, numbers, and symbols.
+func SensitiveString() string {
+	return generatedRandomValue + StringN(30)
+}
+
+// SensitiveAlphanumeric returns a random string prefixed with a generated random value that is masked in GitHub Action logs.
+// The string returned by SensitiveAlphanumeric always starts with a letter and contains only letters and numbers.
+func SensitiveAlphanumeric() string {
+	return generatedRandomValue + AlphanumericN(30)
```
sfc-gh-asawicki, line 59: nit: do we need these to be that long? (30)

pkg/resources/account_acceptance_test.go
```diff
@@ -32,7 +32,7 @@ func TestAcc_Account_Minimal(t *testing.T) {
 	_ = testenvs.GetOrSkipTest(t, testenvs.TestAccountCreate)
 
 	organizationName := acc.TestClient().Context.CurrentAccountId(t).OrganizationName()
-	id := random.AdminName()
+	id := random.SensitiveString()
```
sfc-gh-asawicki, line 35: if this is the account id then we should use the sensitive account id here and change it type accordingly (so that we have this test object suffix present) - in all tests in this file

pkg/resources/account_acceptance_test.go
```diff
@@ -423,17 +423,17 @@ func TestAcc_Account_IgnoreUpdateAfterCreationOnCertainFields(t *testing.T) {
...
 	accountId := sdk.NewAccountIdentifier(organizationName, id)
 
-	firstName := random.AdminName()
-	lastName := random.AdminName()
+	firstName := random.SensitiveAlphanumeric()
```
sfc-gh-asawicki, line 429: name with numbers?

pkg/resources/legacy_service_user_acceptance_test.go
```diff
@@ -49,7 +50,7 @@ func TestAcc_LegacyServiceUser_BasicFlows(t *testing.T) {
 
 	userModelAllAttributes := model.LegacyServiceUser("w", id.Name()).
 		WithPassword(pass).
-		WithLoginName(id.Name() + "_login").
+		WithLoginName(loginName + "_login").
```
sfc-gh-asawicki, line 53: if we are no longer using the name + _login, but a dedicated var, then we can remove the + _login logic, right?

pkg/resources/service_user_acceptance_test.go
```diff
@@ -45,7 +47,7 @@ func TestAcc_ServiceUser_BasicFlows(t *testing.T) {
 		WithComment(newComment)
 
 	userModelAllAttributes := model.ServiceUser("w", id.Name()).
-		WithLoginName(id.Name() + "_login").
+		WithLoginName(loginName + "_login").
```
sfc-gh-asawicki, line 50: same here

sfc-gh-asawicki (changes requested):

pkg/resources/legacy_service_user_acceptance_test.go
sfc-gh-jmichalak: This affects other suffixes as well.

pkg/acceptance/helpers/ids_generator.go
sfc-gh-jcieslak: Right, added

pkg/acceptance/helpers/random/random_helpers.go
sfc-gh-jcieslak: No, it's fine to make it shorter, changed to 10

pkg/resources/account_acceptance_test.go
sfc-gh-jcieslak: True, done

pkg/resources/account_acceptance_test.go
sfc-gh-jcieslak: Old AdminName also used Alphanumeric, so I went with the same setup

pkg/resources/legacy_service_user_acceptance_test.go
sfc-gh-jcieslak: I'm not sure if this is what we talk about, but I assume this would be the same or similar as here we are testing that the login name can be changed, so in this case we could either create another login name like this:
```go
newLoginName := random.Sensitive()
```
but I just went with what was there and re-used one login name with postfix. That being said I'm not sure if dedicated var would resolve this as we would anyway need two values to be able to test the change.

pkg/resources/service_user_acceptance_test.go
sfc-gh-jcieslak: same as above?

pkg/resources/account_acceptance_test.go
```diff
@@ -576,16 +576,16 @@ func TestAcc_Account_UpgradeFrom_v0_99_0(t *testing.T) {
...
 
+	id := random.SensitiveString()
 	email := random.Email()
-	name := random.AdminName()
 	adminName := random.AdminName()
 	adminPassword := random.Password()
-	firstName := random.AdminName()
-	lastName := random.AdminName()
+	firstName := random.SensitiveAlphanumeric()
+	lastName := random.SensitiveAlphanumeric()
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Well, that's true, changed

sfc-gh-jmichalak (approved):

pkg/resources/legacy_service_user_acceptance_test.go
sfc-gh-asawicki: yes, for the next steps you require `newLoginName`, let's do this :)

pkg/resources/account_acceptance_test.go
sfc-gh-asawicki: Hmm, not truly done, as this id still does not have the test object suffix, right?

pkg/resources/account_acceptance_test.go
sfc-gh-asawicki: yeah, but it was probably an unfortunate use of the func, right?

sfc-gh-asawicki (changes requested):

pkg/resources/account_acceptance_test.go
sfc-gh-jcieslak: Those were changed with the @sfc-gh-jmichalak comment to just AlphaN

pkg/resources/account_acceptance_test.go
sfc-gh-jcieslak: Changed to RandomSensitiveAccountObjectIdentifier which has test object suffic

pkg/resources/legacy_service_user_acceptance_test.go
sfc-gh-jcieslak: Done

sfc-gh-asawicki (approved):

PR #3562, opened 2025-04-11
fix: Split acceptance tests

sfc-gh-asawicki:
TODO


pkg/resources/account_role_acceptance_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build !account_level_tests
```
sfc-gh-jcieslak, line 1: Shouldn't this be in the account level tests? (also provider, user, account _acceptance_test.go)

pkg/resources/legacy_service_user_acceptance_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build !account_level_tests
```
sfc-gh-jcieslak, line 1: Shouldn't the user be on this list, too? It's account level and contains parameters

pkg/sdk/testint/tables_integration_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build account_level_tests
```
sfc-gh-jcieslak, line 1: I wouldn't expect tables to be included, but I guess that's confusing because of the tag name (but I don't have a better alternative, so I would leave it as is).

pkg/sdk/testint/users_integration_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build !account_level_tests
```
sfc-gh-jcieslak, line 1: Shouldn't it be included?

sfc-gh-jcieslak (approved):

pkg/resources/account_acceptance_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build !account_level_tests
```
sfc-gh-jmichalak, line 1: It should be account_level_tests, right? The same with account attachments.

pkg/resources/secondary_database_acceptance_test.go
```diff
@@ -20,7 +22,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TODO [SNOW-1991414]: discuss and address all the nondeterministic tests in this file
```
sfc-gh-jmichalak, line 23: What about this?

pkg/sdk/testint/tables_integration_test.go
sfc-gh-jmichalak: Hmm, we can comment here explaining why this happens. We can also check if we could set AccountParameterDataRetentionTimeInDays on a database/schema instead of the account.

CONTRIBUTING.md
```diff
@@ -54,7 +54,7 @@ role = "<your role>"
 host="<host of your account, e.g. organisation-account_name.snowflakecomputing.com>"
 ```
 
-To be able to run all the tests you additionally need two more profiles `[secondary_test_account]` and `[incorrect_test_profile]`:
+To be able to run all the tests you additionally the second profile `[secondary_test_account]`:
 
 ```sh
 [secondary_test_account]
```
sfc-gh-jmichalak, line 60: Can we also update account -> account_name + organization_name?

.github/workflows/tests-account-level.yml
```diff
@@ -0,0 +1,183 @@
+# A big chunk is a copy of tests.yml.
+# Reason: you can only set GH action concurrency on workflow or job level.
+# Our current test setup requires a lot of steps.defaults:
+# We want to have a separate workflow for account-level tests, so that we are sure, that there is only one instance across all active PRs.
+# This will be deduplicated while working on better job definitions.
```
sfc-gh-jmichalak, line 5: Do we have a ticket for this?

sfc-gh-jmichalak (changes requested):

pkg/resources/account_role_acceptance_test.go
sfc-gh-asawicki: No, we don't need to add it to all the account level tests, only to the ones that interact with account settings. We can rename this tag if you have any valid proposal/

pkg/resources/legacy_service_user_acceptance_test.go
sfc-gh-asawicki: You are right about the account parameters, I will check if we should move all the test or is it enough to just move the few

pkg/resources/account_acceptance_test.go
sfc-gh-asawicki: no, no need, because the new account is created here.

pkg/resources/secondary_database_acceptance_test.go
sfc-gh-asawicki: wdym? This whole action is addressing these nondeterministic ones.

.github/workflows/tests-account-level.yml
sfc-gh-asawicki: we have a TODO in the backlog cleanup

pkg/resources/secondary_database_acceptance_test.go
sfc-gh-jmichalak: Cool, I got confused because this ticket 
- lacks descriptions 
- is in TODO (which is fine if we don't cover everything here)
and I wanted to be sure we're not missing anything here.

pkg/resources/legacy_service_user_acceptance_test.go
sfc-gh-asawicki: I am moving all the user tests to the account-level ones. If they will take too much time (and remember, this job is run only once across all our PRs), then we will extract some of them.

pkg/sdk/testint/tables_integration_test.go
sfc-gh-asawicki: I moved it back as @sfc-gh-jmichalak proposal (setting on schema level) worked.

pkg/sdk/testint/users_integration_test.go
sfc-gh-asawicki: Added, as there are account level parameters checked.

CONTRIBUTING.md
sfc-gh-asawicki: Replaced. I also added links to our guides.

pkg/resources/legacy_service_user_acceptance_test.go
sfc-gh-asawicki: Account level tests on this PR (after the recent changes) took 22 minutes, I think it's a good start.

pkg/resources/account_acceptance_test.go
sfc-gh-asawicki: And they are skipped mostly because they should not be run on the given account. I added a note to the issue, I am not modifying them now.

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (approved):

PR #3570, opened 2025-04-15
fix: Handle drop operation safely

sfc-gh-jcieslak:
## Changes
- Add a common drop function (with tests) that accounts for missing high-hierarchy objects
  - I added the implicit `.WithIfExists(true)` option
- Add DropSafely method that uses a common drop function
- Test errors that may occur for drop operations with missing high-hierarchy objects
- Add new error to distinct SDK error that is skippable from underlying Snowflake ones (check ideas section)
- Add test implementation for tables and streamlits with tests
- Add common delete function for resoruces (wanted to keep it as short and simple as possible, but still ended up with two functions that need to be passed; the only solution that I thought of to shorten the function further is replacing parameters with resource name and map to correct arguments (parsing and Drop function) inside the common delete function basing on resource name. It could be less safe approach, because of the missing mappings, but maybe a cleaner and better one).
  - For more complex drop operations (like for accounts), I would keep the dedicated drop function

## Next steps
- Adjust templates to include DropSafely method
- Move common "safe" drop and show functions to one file
- Use in all resources
- Go through object renaming tests and see how the changes with drop and read affect those

## Ideas
- Use ErrSkippable in the safe show function (like in the current safe drop implementation) to distinct ErrObjectNotFound (error form the collections.Find) from the actual skippable operation.


pkg/resources/manual_tests/safe_delete_operation/safe_delete_operation.md
```diff
@@ -0,0 +1,101 @@
...
+Because of the limitations of terraform plugin testing framework, we cannot test the safe delete operation
+automatically.
+This test provides a guide how to test the safe delete operation manually.
+
+## Snowflake setup
+
+Before running Terraform tests, you have to create a simple database:
+
+```snowflake
+CREATE DATABASE TEST_DATABASE;
```
sfc-gh-asawicki, line 12: nit: maybe with some random part so that we don't overlap with anyone?

pkg/resources/manual_tests/safe_delete_operation/safe_delete_operation.md
```diff
@@ -0,0 +1,101 @@
...
+we will test the delete operation without the safe delete operation enabled and in the second part we will enable it.
+
+### Part 1. delete operation before the changes
+
+1. Run `terraform apply -auto-approve` to create a new table.
+2. Run `terraform apply -destroy` to delete the table (do not confirm the deletion by inputting `yes` into the console).
+3. Run the following SQL statement to remove the public schema:
+
+```snowflake
+DROP SCHEMA TEST_DATABASE.PUBLIC;
```
sfc-gh-asawicki, line 66: nit: let's maybe create a dedicated schema (dropping public schema and recreating it is smelly)

pkg/resources/manual_tests/safe_delete_operation/safe_delete_operation.md
```diff
@@ -0,0 +1,101 @@
+# Safe delete operation
+
+Because of the limitations of terraform plugin testing framework, we cannot test the safe delete operation
```
sfc-gh-asawicki, line 3: nit: I have one other idea for this test, going around the limitation:
1. we create a dedicated test resource
2. we implement the read function so that it does not fail with removed schema (even no-op would work, right?)
3. We remove schema in a preconfig step
and it should work, right?

I don't want this change now, but let's add a subtask in SNOW-933289 and a TODO here, WDYT?

pkg/helpers/helpers.go
```diff
@@ -95,6 +95,10 @@ func DecodeSnowflakeID(id string) sdk.ObjectIdentifier {
 	}
 }
 
+func DecodeSnowflakeIDErr[ID sdk.ObjectIdentifierConstraint](id string) (ID, error) {
```
sfc-gh-jmichalak, line 98: It's not clear why this function always returns a nil error. Do we need it?

pkg/sdk/errors.go
```diff
@@ -19,15 +19,16 @@ var (
...
 	ErrDoesNotExistOrOperationCannotBePerformed = NewError("object does not exist, or operation cannot be performed")
+	ErrObjectNotExistOrAuthorized               = NewError("object does not exist or not authorized")
 	ErrAccountIsEmpty                           = NewError("account is empty")
 	ErrGrantPartiallyExecuted                   = NewError("grant partially executed")
 
 	// snowflake-sdk errors.
 	ErrInvalidObjectIdentifier = NewError("invalid object identifier")
 	ErrDifferentDatabase       = NewError("database must be the same")
 	ErrDifferentSchema         = NewError("schema must be the same")
+	ErrSkippable               = NewError("Terraform Provider error indicating that the operation failed, but is skippable")
```
sfc-gh-jmichalak, line 31: We should change the error message because SDK shouldn't rely on its clients.

pkg/sdk/errors_test.go
```diff
@@ -132,3 +134,23 @@ func TestWriteTree(t *testing.T) {
...
+	testCases := []struct {
+		err         error
+		shouldMatch bool
+	}{
+		{err: nil},
+		{err: ErrObjectNotFound},
+		{err: errors.New(ErrSkippable.Error())},
+		{err: NewError(ErrSkippable.Error())},
+		{err: ErrSkippable, shouldMatch: true},
+		{err: errors.Join(ErrObjectNotFound, ErrSkippable), shouldMatch: true},
```
sfc-gh-jmichalak, line 148: Nit: It's not very readable, I'd split this slice into matching and non-matching errors.

pkg/sdk/drop.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// ErrObjectNotExistOrAuthorized can only happen
+	// when the higher hierarchy object is not accessible for some reason during the "main" drop operation.
+	shouldCheckHigherHierarchies := errors.Is(err, ErrObjectNotExistOrAuthorized)
+
+	if !shouldCheckHigherHierarchies {
+		return err
+	}
+
+	if err != nil {
```
sfc-gh-asawicki, line 28: I wonder if this logic is not to complicated unnecessarily:
1. each branch (apart from Account one) checks for the same error: `errors.Is(err, ErrObjectNotFound)` and appends the same `ErrSkippable`
2. other errors do not matter also on the caller side (only ErrSkippable results in a successful run)

Because of that, this logic could be simplified to:
- (on the caller side) if err != nil then diad from err
- (on the impl side) if errors.Is(err, ErrObjectNotFound) then return nil

we would get rid of the dedicated err which is only used as a conditional, we wouldn't need the errs joining, WDYT?

pkg/sdk/identifier_helpers.go
```diff
@@ -18,6 +18,10 @@ type ObjectIdentifier interface {
 	FullyQualifiedName() string
 }
 
+type ObjectIdentifierConstraint interface {
```
sfc-gh-asawicki, line 21: I think we could utilize it in other places too, can we add a subtask to SNOW-933289 and a TODO here?

pkg/sdk/testint/drop_integration_test.go
```diff
@@ -0,0 +1,118 @@
...
+	invalidSchemaIdOnValidDatabase := testClientHelper().Ids.RandomDatabaseObjectIdentifier()
+	invalidTableIdOnValidDatabase := testClientHelper().Ids.RandomSchemaObjectIdentifierInSchema(invalidSchemaIdOnValidDatabase)
+
+	err = sdk.SafeDrop(testClient(t), tableDrop(invalidTableIdOnValidDatabase), ctx, invalidTableIdOnValidDatabase)
+	assert.Error(t, err)
+	assert.ErrorIs(t, err, sdk.ErrSkippable)
+	assert.ErrorIs(t, err, sdk.ErrObjectNotFound)
+	assert.ErrorIs(t, err, sdk.ErrObjectNotExistOrAuthorized)
+
+	invalidDatabaseId := testClientHelper().Ids.RandomAccountObjectIdentifier()
```
sfc-gh-asawicki, line 73: nit: we already have :
- NonExistingAccountObjectIdentifier
- NonExistingDatabaseObjectIdentifier
- NonExistingSchemaObjectIdentifier
that could be used in such tests (the non existing part is always the last part, so e.g. in schema object id, the db and sc exist but the object not)

sfc-gh-asawicki (changes requested):

pkg/sdk/testint/tables_integration_test.go
```diff
@@ -853,6 +853,54 @@ func TestInt_Table(t *testing.T) {
...
+		err := client.Tables.Drop(ctx, sdk.NewDropTableRequest(nonExistingTableId).WithIfExists(sdk.Bool(true)))
+		require.Error(t, err)
+		require.ErrorIs(t, err, sdk.ErrObjectNotExistOrAuthorized)
+	})
+
+	t.Run("drop safely table", func(t *testing.T) {
+		table, tableCleanup := testClientHelper().Table.Create(t)
+		err := client.Tables.DropSafely(ctx, table.ID())
+		if err != nil {
+			t.Cleanup(tableCleanup)
```
sfc-gh-jmichalak, line 877: Nit: the helpers use IF EXISTS, so we can just call this immediately after creating the object.

pkg/sdk/testint/tables_integration_test.go
```diff
@@ -853,6 +853,54 @@ func TestInt_Table(t *testing.T) {
...
+		_, err = client.Tables.ShowByID(ctx, table.ID())
+		require.ErrorIs(t, err, collections.ErrObjectNotFound)
+	})
+
+	t.Run("drop safely table in non-existing schema", func(t *testing.T) {
+		nonExistingSchemaId := testClientHelper().Ids.RandomDatabaseObjectIdentifier()
+		nonExistingTableId := testClientHelper().Ids.RandomSchemaObjectIdentifierInSchema(nonExistingSchemaId)
+		err := client.Tables.DropSafely(ctx, nonExistingTableId)
+		require.Error(t, err)
+		require.ErrorIs(t, err, sdk.ErrSkippable)
```
sfc-gh-jmichalak, line 890: What about checking `ErrObjectNotFound` here, and in other cases as well?

sfc-gh-jmichalak (changes requested):
This approach has one challenge: [errors.Is](https://pkg.go.dev/errors#Is) matches any vertex, so we need to be careful when joining errors that can be skipped. We can accidentally skip an error.

pkg/sdk/drop.go
sfc-gh-jmichalak: +1, wrapping errors like this and calling errors.Is may have unexpected results when we aren't careful.

pkg/acceptance/helpers/sequence_client.go
```diff
@@ -0,0 +1,35 @@
...
+func (c *SequenceClient) client() sdk.Sequences {
+	return c.context.client.Sequences
+}
+
+func (c *SequenceClient) DropFunc(t *testing.T, id sdk.SchemaObjectIdentifier) func() {
+	t.Helper()
+	ctx := context.Background()
+
+	return func() {
+		err := c.client().Drop(ctx, sdk.NewDropSequenceRequest(id).WithIfExists(sdk.Bool(true)))
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, good point, I'm not sure, the cases SafeDrop coverts, should end up as errors in tests anyway. So I think that maybe just fixing the test is better than adding additional overhead to make it disappear without an error. But maybe there are some cases where this could be useful? My thinking is that:
- on the account level Drop == SafeDrop
- on the lower levels it doesn't matter because we are dropping the database at the end of the tests
tldr; with our current setup, I don't think we need to do it

pkg/resources/manual_tests/safe_delete_operation/safe_delete_operation.md
sfc-gh-jcieslak: Sure, added

pkg/resources/manual_tests/safe_delete_operation/safe_delete_operation.md
sfc-gh-jcieslak: Hmm, yeah, this could be a good case for testing resource. I'll create a ticket for it (SNOW-2043814).

pkg/sdk/drop.go
sfc-gh-jcieslak: Adjusted the implementation

pkg/sdk/identifier_helpers.go
sfc-gh-jcieslak: Yeah, created SNOW-2043829 and left a TODO

pkg/helpers/helpers.go
sfc-gh-jcieslak: It's temporary wrapper function for old resource identifier decoder `DecodeSnowflakeID`. Ultimately, the `DecodeSnowflakeID` function and this one will be removed once we'll refactor all resources to use identifier parsing functions.

pkg/sdk/errors.go
sfc-gh-jcieslak: Because the implementation changed, I removed the error

pkg/sdk/errors_test.go
sfc-gh-jcieslak: Because the implementation of SafeDrop changed, I removed the error and tests

pkg/sdk/testint/tables_integration_test.go
sfc-gh-jcieslak: True, changed

pkg/sdk/testint/drop_integration_test.go
sfc-gh-jcieslak: I added missing NonExisting- identifiers

pkg/sdk/drop.go
```diff
@@ -7,8 +7,8 @@ import (
 
 // SafeDrop is a helper function that wraps a drop function and handles common error cases that
 // relate to missing high hierarchy objects when dropping lower ones like schemas, tables, views, etc.
-// Whenever an object is missing or the higher hierarchy object is not accessible, it will return ErrObjectNotFound error,
-// which can be leveraged with [errors.Is] to handle the logic in case of missing objects.
+// Whenever an object is missing or the higher hierarchy object is not accessible, it will return nil error,
```
sfc-gh-jmichalak, line 10: Nit: `it will not return an error`.

And `which means the caller can safely ignore the error and proceed` sounds a bit confusing because now, the error handling is "default" by the caller, (certainly not **ignored** by the caller), right? The point of this function is to not return an error when objects higher in hierarchy don't exist as well, and IMO this comment does not reflect this enough.

pkg/resources/manual_tests/safe_delete_operation/safe_delete_operation.md
```diff
@@ -9,11 +10,10 @@ This test provides a guide how to test the safe delete operation manually.
 Before running Terraform tests, you have to create a simple database:
 
 ```snowflake
-CREATE DATABASE TEST_DATABASE;
+CREATE DATABASE TEST_SAFE_DELETE_OPERATION_DATABASE;
+CREATE DATABASE TEST_SCHEMA;
```
sfc-gh-jmichalak, line 14: CREATE SCHEMA?

sfc-gh-jmichalak (dismissed):

pkg/resources/manual_tests/safe_delete_operation/safe_delete_operation.md
sfc-gh-jcieslak: lul, done

sfc-gh-jmichalak (approved):

pkg/acceptance/helpers/sequence_client.go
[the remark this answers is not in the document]
sfc-gh-asawicki: ok, let's leave it for now

sfc-gh-asawicki (approved):

PR #3573, opened 2025-04-16
chore: Rename ConfigDTO to LegacyConfigDTO

sfc-gh-jmichalak:
- Rename ConfigDTO to LegacyConfigDTO.
- Move LegacyConfigDTO to separate files.
- Make LoadConfigFile generic. This way it will able to read a new ConfigDTO format (next PR).
- Fix failing external volume creation.

## TODO
- Implement new ConfigDTO with underscores and renamed `tracing` field.
- Make it able to read both ConfigDTO and LegacyConfigDTO (latter by default for now).


pkg/sdk/legacy_config_test.go
```diff
@@ -0,0 +1,332 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testvars"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeenvs"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/testhelpers"
+	"github.com/snowflakedb/gosnowflake"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TODO [SNOW-1827309]: use toml config builder instead of hardcoding
+func TestLoadConfigFile(t *testing.T) {
```
sfc-gh-jmichalak, line 20: I didn't add `legacy` suffix, will do in the next PR.

sfc-gh-jmichalak: Done here.

pkg/sdk/config.go
```diff
@@ -361,12 +249,12 @@ func pointerUrlAttributeSet(src *string, dst **url.URL) error {
 	return nil
 }
 
-func LoadConfigFile(path string, verifyPermissions bool) (map[string]ConfigDTO, error) {
+func LoadConfigFile[T LegacyConfigDTO](path string, verifyPermissions bool) (map[string]T, error) {
```
sfc-gh-jcieslak, line 252: I'm not sure how this helps with type migration, but it's going to be changed, then it's fine.

sfc-gh-jcieslak (approved):

pkg/sdk/config.go
sfc-gh-jmichalak: As it's stated in the description, this will be addressed later. Basically, we can do something like `T ConfigDTO | LegacyConfigDTO` here and just call this conditionally, based on a bool flag.

PR #3574, opened 2025-04-17
fix: Network policy test fixes

sfc-gh-jcieslak:
## Changes
- Adjust tests to account for new network policy
  - Change default value for user network policy parameter to RESTRICTED_ACCESS and change it's default level to ACCOUNT
  - Adjust CheckUserPasswordPolicyAttachmentDestroy logic to only check for password policies (not all policies as it was previously doing)
  - Remove TestAcc_UserPasswordPolicyAttachment_gh3005 as it was referring to the old provider code (v0.87.0), which had a bug and was mapping POLICY_STATUS as a non-null field. It was fixed later on and it will be always failing because of the network policy on the account that has always null status (I'm guessing status for this policy doesn't matter).

## Next steps
  - Adjust parameter asserts, particularly HasAllDefaults function to check for default values on the level by checking with locally defined default value (in the object_parameters_def.go) rather than from Snowflake. Because of the current behavior, the network policy would always fail, because the Snowflake default is set to an empty string, not RESTRICTED_ACCESS, as it is in our parameter definitions.
    - Create a follow-up ticket and put it into commented out `HasDefaultParameterValueOnLevel(sdk.UserParameterNetworkPolicy, sdk.ParameterTypeAccount)`

### DIFF
--- a/pkg/acceptance/bettertestspoc/assert/objectparametersassert/user_parameters_snowflake_gen.go
+++ b/pkg/acceptance/bettertestspoc/assert/objectparametersassert/user_parameters_snowflake_gen.go
@@ -70,7 +70,7 @@ func (u *UserParametersAssert) HasDefaultParameterValueOnLevel(parameterName sdk
 func (u *UserParametersAssert) HasAllDefaults() *UserParametersAssert {
 	return u.
 		HasDefaultParameterValueOnLevel(sdk.UserParameterEnableUnredactedQuerySyntaxError, sdk.ParameterTypeSnowflakeDefault).
-		// TOOD: (temp comment) HasDefaultParameterValueOnLevel(sdk.UserParameterNetworkPolicy, sdk.ParameterTypeAccount).
+		// TOOD(SNOW-2048330): HasDefaultParameterValueOnLevel(sdk.UserParameterNetworkPolicy, sdk.ParameterTypeAccount).
 		HasDefaultParameterValueOnLevel(sdk.UserParameterPreventUnloadToInternalStages, sdk.ParameterTypeSnowflakeDefault).
 		HasDefaultParameterValueOnLevel(sdk.UserParameterAbortDetachedQuery, sdk.ParameterTypeSnowflakeDefault).
 		HasDefaultParameterValueOnLevel(sdk.UserParameterAutocommit, sdk.ParameterTypeSnowflakeDefault).
--- a/pkg/acceptance/bettertestspoc/assert/objectparametersassert/user_parameters_snowflake_gen.go
+++ b/pkg/acceptance/bettertestspoc/assert/objectparametersassert/user_parameters_snowflake_gen.go
@@ -70,7 +70,7 @@ func (u *UserParametersAssert) HasDefaultParameterValueOnLevel(parameterName sdk
 func (u *UserParametersAssert) HasAllDefaults() *UserParametersAssert {
 	return u.
 		HasDefaultParameterValueOnLevel(sdk.UserParameterEnableUnredactedQuerySyntaxError, sdk.ParameterTypeSnowflakeDefault).
-		HasDefaultParameterValueOnLevel(sdk.UserParameterNetworkPolicy, sdk.ParameterTypeSnowflakeDefault).
+		// TOOD: (temp comment) HasDefaultParameterValueOnLevel(sdk.UserParameterNetworkPolicy, sdk.ParameterTypeAccount).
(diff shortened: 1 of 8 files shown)


sfc-gh-jmichalak (dismissed):

pkg/acceptance/bettertestspoc/assert/objectparametersassert/user_parameters_snowflake_gen.go
```diff
@@ -1024,7 +1024,7 @@ func (u *UserParametersAssert) HasDefaultEnableUnredactedQuerySyntaxErrorValueEx
 }
 
 func (u *UserParametersAssert) HasDefaultNetworkPolicyValueExplicit() *UserParametersAssert {
-	return u.HasNetworkPolicy("")
+	return u.HasNetworkPolicy("RESTRICTED_ACCESS")
```
sfc-gh-jcieslak, line 1027: I was wondering if we want to have it somewhere as constant, but could find a good place for it (can add it as follow-up or in this pr). It would be a good to gather values that are test specific (testing envs as well as known predefined identifiers as this one).

sfc-gh-jcieslak: Added a proposal for `testenvidentifiers` package

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/user_resource_parameters_ext.go
```diff
@@ -1,6 +1,7 @@
 package resourceparametersassert
 
 import (
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvidentifiers"
```
sfc-gh-asawicki, line 4: reviewdawg

pkg/acceptance/bettertestspoc/assert/objectparametersassert/user_parameters_snowflake_gen.go
```diff
@@ -70,7 +70,7 @@ func (u *UserParametersAssert) HasDefaultParameterValueOnLevel(parameterName sdk
 func (u *UserParametersAssert) HasAllDefaults() *UserParametersAssert {
 	return u.
 		HasDefaultParameterValueOnLevel(sdk.UserParameterEnableUnredactedQuerySyntaxError, sdk.ParameterTypeSnowflakeDefault).
-		HasDefaultParameterValueOnLevel(sdk.UserParameterNetworkPolicy, sdk.ParameterTypeSnowflakeDefault).
+		// TOOD: (temp comment) HasDefaultParameterValueOnLevel(sdk.UserParameterNetworkPolicy, sdk.ParameterTypeAccount).
```
sfc-gh-asawicki, line 73: I read the PR description but I don't understand why this is commented out

pkg/acceptance/testenvidentifiers/test_identifiers.go
```diff
@@ -0,0 +1,7 @@
+package testenvidentifiers
+
+import "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+
+var (
```
sfc-gh-asawicki, line 5: reviewdog

sfc-gh-asawicki (changes requested):

pkg/acceptance/bettertestspoc/assert/objectparametersassert/user_parameters_snowflake_gen.go
sfc-gh-asawicki: Discussed

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/user_resource_parameters_ext.go
sfc-gh-jcieslak: Don

pkg/acceptance/testenvidentifiers/test_identifiers.go
sfc-gh-jcieslak: Don

pkg/acceptance/bettertestspoc/assert/objectparametersassert/user_parameters_snowflake_gen.go
sfc-gh-jcieslak: Added a ticket number

sfc-gh-asawicki (approved):

PR #3576, opened 2025-04-17
chore: Apply safe functions

sfc-gh-jcieslak:
## Changes
- Move "safe" operations into one file
- Add Safe read and drop on the SDK level
- Adjust the sdk templates, so that safe functions are also generated with their regular counterparts
- Added IfExists option to the managed accounts (it's not documented, but works; reported on doc discuss)

## Next PRs
- Adjust resources and data sources to use new functions
- Check object renaming tests and adjust with current Read and Drop handling
- Add TODO with SNOW-2048276 for External functions in the SDK


pkg/sdk/external_functions_gen.go
```diff
@@ -10,6 +10,7 @@ type ExternalFunctions interface {
 	Alter(ctx context.Context, request *AlterExternalFunctionRequest) error
 	Show(ctx context.Context, request *ShowExternalFunctionRequest) ([]ExternalFunction, error)
 	ShowByID(ctx context.Context, id SchemaObjectIdentifierWithArguments) (*ExternalFunction, error)
+	ShowByIDSafely(ctx context.Context, id SchemaObjectIdentifierWithArguments) (*ExternalFunction, error)
 	Describe(ctx context.Context, id SchemaObjectIdentifierWithArguments) ([]ExternalFunctionProperty, error)
```
sfc-gh-asawicki, line 14: question: no drop here for external functions? they currenlty use the drop for normal functions, right?

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
```diff
@@ -5,56 +5,69 @@
...
             {{ end }}
         {{ end }}
+    {{ else if eq .Name "Drop" }}
+        func (v *{{ $impl }}) Drop(ctx context.Context, request *{{ .OptsField.DtoDecl }}) error {
+            opts := request.toOpts()
+            return validateAndExec(v.client, ctx, opts)
+        }
+
+        func (v *{{ $impl }}) DropSafely(ctx context.Context, id {{ .ObjectInterface.IdentifierKind }}) error {
+            return SafeDrop(v.client, func() error { return v.Drop(ctx, NewDrop{{ .ObjectInterface.NameSingular }}Request(id).WithIfExists(true)) }, ctx, id)
```
sfc-gh-asawicki, line 65: nit: what if the given drop does not support the WithIfExists clause?

sfc-gh-asawicki (approved):

pkg/sdk/external_functions_gen.go
sfc-gh-jcieslak: Yes

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
sfc-gh-jcieslak: It would fail, but no object yet wasn't supporting it, so based on that, I don't think should have an issue with this in the future.

sfc-gh-asawicki: Pkay, let's leave it for now

pkg/sdk/external_functions_gen.go
sfc-gh-asawicki: Let's add a task in our SDK improvements epic for handling such situations on the SDK level (if not present yet) - even if it would here mean just passing the execution to the functions drop (the same probably with tables)

sfc-gh-jmichalak (approved):

pkg/sdk/external_functions_gen.go
sfc-gh-jcieslak: Created SNOW-2048276 (I'll add todo here in the next pr.

PR #3578, opened 2025-04-18
feat: Handle operations in resources using safe functions

sfc-gh-jcieslak:
## Changes
- Update Read functions to use Safe read
- Update Delete functions for objects and remove old ones (left the old delete functions for more complicated cases like Account delete)
- Adjust SDK implementation for ShowByID in a few cases (they were not returning the error expected by SafeShowByID)
- Added a new entry migration guide

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -18,7 +18,7 @@ across different versions.
 Some objects in Snowflake are created in hierarchy, for example, tables (database → schema → table).
 When the user wants to remove the higher-hierarchy object (like a database), the lower-hierarchy objects should be removed beforehand. 
 Otherwise, Terraform would fail to remove the lower-hierarchy objects from the state, 
-and without manual state management it wouldn't be possible to remove this object, ending up in broken state.
+and without manual state management it wouldn't be possible to remove this object, ending up in broken state (reference issue: [#1243](https://github.com/snowflakedb/terraform-provider-snowflake/issues/1243)).
 This behavior was described more in detail in [our documentation](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/guides/object_renaming_research_summary#renaming-higher-hierarchy-objects).
 
 For improved usability, we adjusted the Read and Delete operation implementations for all resources, 
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -14,6 +14,36 @@ across different versions.
 
 ## v1.1.0 ➞ v1.2.0
 
+### New behavior for Read and Delete operations when removing high-hierarchy objects
+Some objects in Snowflake are created in hierarchy, for example, tables (database → schema → table).
+When the user wants to remove the higher-hierarchy object (like a database), the lower-hierarchy objects should be removed beforehand. 
+Otherwise, Terraform would fail to remove the lower-hierarchy objects from the state, 
+and without manual state management it wouldn't be possible to remove this object, ending up in broken state.
+This behavior was described more in detail in [our documentation](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/guides/object_renaming_research_summary#renaming-higher-hierarchy-objects).
+
+For improved usability, we adjusted the Read and Delete operation implementations for all resources, 
+so that now, they're able to know the higher-hierarchy object is missing, and they can safely remove themselves from the state.
+
+To demonstrate this behavior, let's take the following configuration:
+```terraform
+resource "snowflake_table" "test" {
+  database = "TEST_DATABASE"
+  schema   = "PUBLIC"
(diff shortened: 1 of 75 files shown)


pkg/resources/storage_integration.go
```diff
@@ -204,11 +210,19 @@ func ReadStorageIntegration(ctx context.Context, d *schema.ResourceData, meta an
 		return diag.FromErr(fmt.Errorf("storage integration read, error decoding id: %s as sdk.AccountObjectIdentifier, got: %T", d.Id(), id))
 	}
 
-	s, err := client.StorageIntegrations.ShowByID(ctx, id)
+	s, err := client.StorageIntegrations.ShowByIDSafely(ctx, id)
 	if err != nil {
-		log.Printf("[DEBUG] storage integration (%s) not found", d.Id())
-		d.SetId("")
-		return nil
+		if errors.Is(err, sdk.ErrObjectNotFound) {
```
sfc-gh-jmichalak, line 215: This change is missing from the migration guide.

pkg/resources/sequence.go
```diff
@@ -114,8 +123,18 @@ func ReadSequence(ctx context.Context, d *schema.ResourceData, meta any) diag.Di
 	client := meta.(*provider.Context).Client
 
 	id := helpers.DecodeSnowflakeID(d.Id()).(sdk.SchemaObjectIdentifier)
-	seq, err := client.Sequences.ShowByID(ctx, id)
+	seq, err := client.Sequences.ShowByIDSafely(ctx, id)
 	if err != nil {
+		if errors.Is(err, sdk.ErrObjectNotFound) {
+			d.SetId("")
```
sfc-gh-jmichalak, line 129: Migration guide?

pkg/resources/api_integration.go
```diff
@@ -220,10 +227,13 @@ func ReadAPIIntegration(ctx context.Context, d *schema.ResourceData, meta interf
...
 
-	integration, err := client.ApiIntegrations.ShowByID(ctx, id)
+	integration, err := client.ApiIntegrations.ShowByIDSafely(ctx, id)
 	if err != nil {
-		log.Printf("[DEBUG] api integration (%s) not found", d.Id())
-		d.SetId("")
+		if errors.Is(err, sdk.ErrObjectNotFound) {
+			log.Printf("[DEBUG] api integration (%s) not found. Marking the resource as removed.", d.Id())
+			d.SetId("")
+			return nil
```
sfc-gh-jmichalak, line 235: Why return nil? In other resources we return a warning here.

MIGRATION_GUIDE.md
```diff
@@ -14,6 +14,36 @@ across different versions.
 
 ## v1.1.0 ➞ v1.2.0
 
+### New behavior for Read and Delete operations when removing high-hierarchy objects
+Some objects in Snowflake are created in hierarchy, for example, tables (database → schema → table).
+When the user wants to remove the higher-hierarchy object (like a database), the lower-hierarchy objects should be removed beforehand.
```
sfc-gh-jmichalak, line 19: Can we specify why this happens? Is it also affecting situations where a parent object is managed by Terraform as well?

sfc-gh-jmichalak (changes requested):

pkg/resources/storage_integration.go
sfc-gh-jcieslak: The migration guide is specified for all resources (If that's what we are talking about here). I can emphasize this more in the migration guide.

pkg/resources/sequence.go
sfc-gh-jcieslak: Same as above.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: I wanted to reference an issue regarding this, but could find one. I'll try to find it and link here. It should give more context, and also I can mention some of the stuff from the guide which is linked (that this leads from either poorly managed dependencies in Terraform, or partly managed infra).

pkg/resources/grant_privileges_to_share_acceptance_test.go
```diff
@@ -638,6 +638,8 @@ func TestAcc_GrantPrivilegesToShare_NoOnOption(t *testing.T) {
 
 // proves https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2621 doesn't apply to this resource
 func TestAcc_GrantPrivilegesToShare_RemoveShareOutsideTerraform(t *testing.T) {
+	t.Skip("Should be addressed in SNOW-2048471 - without it the test cannot be properly asserted in the terraform testing framework")
```
sfc-gh-asawicki, line 641: without what?

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #3579, opened 2025-04-18
chore: Test proper diff suppression for data types

sfc-gh-asawicki:
Test thoroughly the proposed logic for data types diff suppression improved logic for top-level datatype and for nested datatype. Mainly:
- Add two dedicated test resources
- Test the internally discussed variant
- Introduce simple unknows in the data types
- Introduce new comparison method
- Prepare helpers for easier handling in the production resources

Follow-up (next PRs):
- solve all TODOs left with next PR (implement all datatypes and add tests; handle changes with size; improve list methods - while working on production resources; extract the helpers to better-named file)
- add state upgrader (top level + list) + tests
- use in masking policy resource
- use in row access policy resource
- migration guide + documentation (before GA) + guidelines (after GA)


pkg/resources/test_resource_data_type_diff_handling_impl.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("[DEBUG] correctly parsed data type %v", dataType)
+	return dataType, nil
+}
+
+func readChangedDatatypeCommon(d *schema.ResourceData, key string) (datatypes.DataType, error) {
+	log.Printf("[DEBUG] reading updated value for datatype field %s", key)
+	_, n := d.GetChange(key)
```
sfc-gh-jcieslak, line 71: Is it the same as the function above, but with a different function name and logs? In updates, I think that `_, n := d.GetChange(key)` and `dataTypeRawConfig := d.Get(key).(string)` would be the same thing. Just confirming that was intended.

pkg/sdk/datatypes/number.go
```diff
@@ -92,10 +108,24 @@ func parseNumberDataTypeWithoutPrecisionAndScale(raw sanitizedDataTypeRaw) (*Num
...
-		return &NumberDataType{DefaultNumberPrecision, DefaultNumberScale, raw.matchedByType}, nil
+		return &NumberDataType{DefaultNumberPrecision, DefaultNumberScale, raw.matchedByType, false, false}, nil
 	}
 }
 
 func areNumberDataTypesTheSame(a, b *NumberDataType) bool {
 	return a.precision == b.precision && a.scale == b.scale
 }
+
+func areNumberDataTypesDefinitelyDifferent(a, b *NumberDataType) bool {
```
sfc-gh-jcieslak, line 119: pretty confusing, maybe change to `areNumberDataTypesDeeplyEqual`?

pkg/sdk/datatypes/data_types.go
```diff
@@ -156,3 +156,67 @@ func castSuccessfully[T any](a T, b DataType, invoke func(a T, b T) bool) bool {
...
 }
+
+// AreDefinitelyDifferent compares any two data types.
+// If both data types are nil it returns false.
+// If only one data type is nil it returns true.
+// It returns true for different underlying types.
+// For the same type it performs type-specific check.
+// TODO: test this function if approved
+// TODO: implement all missing checks if approved
+func AreDefinitelyDifferent(a DataType, b DataType) bool {
```
sfc-gh-jcieslak, line 167: As mentioned in Text or Number, I think it's easier to think about equality than something being not equal or different.

pkg/resources/test_resource_data_type_diff_handling_list_acceptance_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	externalChange := func(configValue string, externalValue string, expectChanges bool) DataTypeDiffHandlingTestCase {
+		return DataTypeDiffHandlingTestCase{
+			ConfigValue:   configValue,
+			ExternalValue: externalValue,
+			ExpectChanges: expectChanges,
+		}
+	}
+
+	testCases := []DataTypeDiffHandlingTestCase{
```
sfc-gh-jcieslak, line 70: Not sure if this is the same as in the other test, but if it is, maybe it would be better to somehow merge it (would be easier in dedicated package as it wouldn't "pollute" the scope of `resources` package

pkg/resources/test_resource_data_type_diff_handling.go
```diff
@@ -0,0 +1,109 @@
+package resources
```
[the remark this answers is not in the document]
sfc-gh-asawicki: for now, I kept them here on purpose, like the others existing ones, ultimately we will move them but then we need to also make the separate package for helpers, so that they can be used both by resources and test resources. There will be a ticket created for that (on my TODO list).

pkg/sdk/datatypes/number.go
sfc-gh-asawicki: No. These is not about the deep equality. If you compare NUMBER(38, 0) with Snowflake NUMBER, they may be different but may be also the same. So you can think about the comparison made as three value result (are they equal): YES, NO, MAYBE. So definitely is only the NO part. I am open to changing the name to something different (if you come up with a better name). More important part is the logic here. Is there a problem with it?

pkg/sdk/datatypes/data_types.go
sfc-gh-asawicki: but then in the usage places you need to `!` the invocation, sometimes in multiple places. This way it's more straightforward from the usage place.

pkg/resources/test_resource_data_type_diff_handling_impl.go
sfc-gh-asawicki: True, I think we can merge them (the logic was a bit different initially but now they do the same as you pointed out)

pkg/resources/test_resource_data_type_diff_handling_list_acceptance_test.go
sfc-gh-asawicki: I will add this merge and extraction to the aforementioned packages cleanup task.

pkg/sdk/datatypes/number.go
sfc-gh-asawicki: I will add the clarification to the `AreDefinitelyDifferent` method

sfc-gh-jcieslak: Yeah, I think documenting this potential 3-value result and that this function only focuses on `NO` results would be sufficient.

pkg/sdk/datatypes/data_types.go
sfc-gh-jcieslak: Discussed above, closing this one.

pkg/resources/test_resource_data_type_diff_handling_impl.go
sfc-gh-asawicki: Done

pkg/sdk/datatypes/number.go
sfc-gh-asawicki: Added the clarification

pkg/resources/test_resource_data_type_diff_handling_list_acceptance_test.go
sfc-gh-asawicki: Added TODO [SNOW-2054208]

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (approved):

pkg/resources/test_resource_data_type_diff_handling_acceptance_test.go
```diff
@@ -0,0 +1,233 @@
...
+		externalChange("NUMBER(38, 0)", "NUMBER", false),
+		externalChange("NUMBER(38, 0)", "NUMBER(38)", false),
+		externalChange("NUMBER(38, 0)", "NUMBER(20)", true),
+		externalChange("NUMBER(38, 0)", "NUMBER(20, 3)", true),
+		externalChange("NUMBER(38, 0)", "NUMBER(38, 2)", true),
+	}
+
+	for _, testCase := range testCases {
+		tc := testCase
+		t.Run(fmt.Sprintf("TestAcc_TestResource_DataTypeDiffHandling config value: %s, new config value: %s, external value: %s, expecitng changes: %t", tc.ConfigValue, tc.NewConfigValue, tc.ExternalValue, tc.ExpectChanges), func(t *testing.T) {
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Fixed in follow-up

PR #3583, opened 2025-04-22
chore: Masking random values

sfc-gh-jcieslak:
## Changes
- Masking values like certs from printing
- Adjust random value length


pkg/acceptance/bettertestspoc/config/custom_variables.go
```diff
@@ -31,16 +31,27 @@ func ReplacementPlaceholderVariable(placeholder ReplacementPlaceholder) replacem
...
+	return wrapperVariable{SnowflakeProviderConfigMultilineMarker, content}
+}
+
+// QuotedWrapperVariable returns Variable containing quoted content wrapped with SnowflakeProviderConfigQuoteMarker later replaced by HclFormatter.
+func QuotedWrapperVariable(content string) wrapperVariable {
+	return wrapperVariable{SnowflakeProviderConfigQuoteMarker, content}
+}
+
+// UnquotedWrapperVariable returns Variable containing unquoted content wrapped with SnowflakeProviderConfigUnquoteMarker later replaced by HclFormatter.
+func UnquotedWrapperVariable(content string) wrapperVariable {
```
sfc-gh-asawicki, line 55: let's rename it to something more high-level, like: VariableReference, and we could only put a name of the variable (instead of content), and internally here add the `var.` part. What do you think? Or we can go even more generic: ReferenceVariable, so that we could use it for all references potentially. (I like the dedicated VariableReference better but I'm open)

sfc-gh-asawicki (changes requested):

sfc-gh-jmichalak (dismissed):

pkg/acceptance/bettertestspoc/config/custom_variables.go
sfc-gh-jcieslak: Added VariableReference that uses UnquotedWrapperVariable.

pkg/datasources/security_integrations_acceptance_test.go
```diff
@@ -516,8 +528,19 @@ func TestAcc_SecurityIntegrations_Saml2(t *testing.T) {
 	issuerURL := acc.TestClient().Context.IssuerURL(t)
 	comment := random.Comment()
 
+	temporaryVariableName := "saml2_x509_cert"
+	temporaryVariableDefinition := fmt.Sprintf(`
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done, since it's kind of similar to what we are doing with the dynamic block, it put it under the same package

sfc-gh-asawicki (dismissed):

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

PR #3613, opened 2025-04-25
chore: Handle forbidden attribute names in resource model builder generation

sfc-gh-asawicki:
The model builder for the managed account was not generating correctly because its schema has `type` which is a keyword in Golang. To work around that:
- the logic for forbidden attribute names was added to generator helpers
- templates for resource model builders were adjusted
- the resulting model builder has:
  - name with added underscore
  - methods `With` are generated without this underscore


sfc-gh-jmichalak (approved):

pkg/internal/genhelpers/util.go
```diff
@@ -60,3 +65,10 @@ func WriteCodeToFile(buffer *bytes.Buffer, fileName string) error {
 	}
 	return nil
 }
+
+func SanitizeAttributeName(name string) string {
+	if slices.Contains(forbiddenAttributeNames, name) {
```
sfc-gh-jcieslak, line 70: Contains or rather Equal?

sfc-gh-jcieslak (approved):

pkg/internal/genhelpers/util.go
sfc-gh-asawicki: contains (ii checks if attribute is in slice of forbidden names)

PR #3615, opened 2025-04-25
chore: Fix ordering in resource model builder constructors

sfc-gh-asawicki:
DRAFT

### DIFF
--- a/pkg/internal/genhelpers/resource_schema_details_extractor.go
+++ b/pkg/internal/genhelpers/resource_schema_details_extractor.go
@@ -22,6 +22,8 @@ type SchemaAttribute struct {
 	Required         bool
 }
 
+var preferredFrontOrdering = []string{"database", "schema", "name"}
+
 // TODO: test
 func ExtractResourceSchemaDetails(name string, resourceSchema map[string]*schema.Schema) ResourceSchemaDetails {
 	orderedAttributeNames := make([]string, 0)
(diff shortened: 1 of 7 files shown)


pkg/internal/genhelpers/resource_schema_details_extractor.go
```diff
@@ -30,6 +32,13 @@ func ExtractResourceSchemaDetails(name string, resourceSchema map[string]*schema
 	}
 	slices.Sort(orderedAttributeNames)
 
+	for _, v := range slices.Backward(preferredFrontOrdering) {
```
sfc-gh-jmichalak, line 35: IMO, it would be more readable if this logic was used in `slices.SortFunc`. Or, at least we can add a comment/extract the code.

sfc-gh-jmichalak (approved):

pkg/internal/genhelpers/resource_schema_details_extractor.go
```diff
@@ -16,26 +16,38 @@ func (s ResourceSchemaDetails) ObjectName() string {
...
+	AttributeSubType schema.ValueType
+	Required         bool
 }
 
+var preferredFrontOrdering = []string{"database", "schema", "name"}
+
 // TODO: test
-func ExtractResourceSchemaDetails(name string, schema map[string]*schema.Schema) ResourceSchemaDetails {
+func ExtractResourceSchemaDetails(name string, resourceSchema map[string]*schema.Schema) ResourceSchemaDetails {
 	orderedAttributeNames := make([]string, 0)
```
sfc-gh-jcieslak, line 29: nit: Shorter way would be to call `orderedAttributeNames := maps.Keys(resourceSchema)`

pkg/internal/genhelpers/resource_schema_details_extractor.go
sfc-gh-jcieslak: It think SortFunc way sounds better, but would be more verbose and harder to understand, because in the sorting function you would have to differentiate between `front` and `other` values and handle comparisons differently between those categories.

sfc-gh-jcieslak (approved):

pkg/internal/genhelpers/resource_schema_details_extractor.go
sfc-gh-asawicki: I will extract and parameterize it in one of the next PRs, hopefully, making it more readable.

pkg/internal/genhelpers/resource_schema_details_extractor.go
sfc-gh-asawicki: I will add it

PR #3619, opened 2025-04-25
chore: Add TOML config generator

sfc-gh-jmichalak:
## Changes
- Add TOML config file generator with builders.
- Show the usage in one of the provider's acceptance tests.

## TODO
- Use the TOML builders in other tests (address SNOW-1827309).


sfc-gh-asawicki (changes requested):
I left a few comments but don't mind them for now. Let's talk on Monday as in this case I was expecting a bit different approach.

Namely, the approach that we have in the SDK already. The model builders were done this way, as we don't have a struct that we can base the generation on. In the SDK, having the structs, allowed us to generate the Builders (request builders) out of the structs.

In the TOML case, we already have ConfigDTO and LegacyConfigDTO. We could just apply the existing SDK generator logic and just add a part where we generate toml out of the already present tags; it would be much more straightforward.

sfc-gh-jcieslak (approved):

pkg/acceptance/helpers/config_toml_creator.go
```diff
@@ -148,14 +149,16 @@ authenticator = 'SNOWFLAKE_JWT'
...
-[%[1]s]
-user = '%[2]s'
-password = '%[7]s'
-role = '%[3]s'
-organization_name = '%[5]s'
-account_name = '%[6]s'
-warehouse = '%[4]s'
-authenticator = 'SNOWFLAKE'
-`, profile, userId.Name(), roleId.Name(), warehouseId.Name(), accountIdentifier.OrganizationName(), accountIdentifier.AccountName(), pass)
+	config := sdk.NewConfigDTO().
```
sfc-gh-asawicki, line 152: For the next PR (with usage in more places): let's extract a helper method that takes the ConfigDTO and profile name and returns the string.

Issue #3622, opened 2025-04-25
[Bug]: Updating snowflake_tag_association in-place fails


angela-mao:
### Terraform CLI Version

1.9.0

### Terraform Provider Version

1.0.5

### Terraform Configuration

```terraform
resource "snowflake_tag_association" "test_association" {
  object_identifiers = [snowflake_pipe.test.fully_qualified_name]
  object_type        = "PIPE"
  tag_id             = snowflake_tag.test.fully_qualified_name
  tag_value          = "Example"
}
```

### Category

category:resource

### Object type(s)

resource:tag_association

### Expected Behavior

I had changed the tag value and expected `terraform apply` to run successfully.

### Actual Behavior

I got an error saying `panic: runtime error: comparing uncomparable type sdk.SchemaObjectIdentifier` and based on the logs (don't have them all anymore) but it said it was due to [line 269 of the `ListDiffWithCommonItems` function](https://github.com/snowflakedb/terraform-provider-snowflake/blob/c4f1e8fd55150e40d8a556580016ff83fe65bdaf/pkg/resources/helpers.go#L269) that is used [here](https://github.com/snowflakedb/terraform-provider-snowflake/blob/37e70acd1b855927fceb13a1da7b3ebb05203b82/pkg/resources/tag_association.go#L227).

I believe it's because `oldIds` and `newIds` are returned by `ExpandObjectIdentifierSet` which returns objects of type `sdk.SchemaObjectIdentifer` which is a custom data type and can't be used with `slices.Contains`.

### Steps to Reproduce

1. Copy the following configuration
2. Run `terraform apply`
3. Now change the tag value
4. Try running `terraform apply`

### How much impact is this issue causing?

Low

### Additional Information

At the moment a workaround is to remove the tags first and then add the new tags but it would be nice to get this fixed to make the process smoother and easier.

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jcieslak:
Hey @angela-mao 👋 
Yesterday, we released a new provider version (v2.1.0), which contains the mentioned fix. Please, try to use the latest version and let me know if the issue has been resolved. Thanks.

sfc-gh-jcieslak:
Closing due to long inactivity, the fix was provided in the version provided above. Please, open another issue if the problem still exists.

PR #3626, opened 2025-04-29
fix: 3622 issue

sfc-gh-jcieslak:
## Changes
- Fixes the #3622 issue where tag association would fail for objects that use `sdk.SchemaObjectidentifier`, which is not a `comparable` type.
- The same error is expected to happen for `sdk.SchemaObjectIdentifierWithArguments`, but only functions and procedures use it, which don't support tagging (even if they would, we would wait for them to be implemented in the resource directly).

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -18,6 +18,16 @@ across different versions.
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.0.0 ➞ v2.0.1
+
+### *(bugfix)* Fixed `snowflake_tag_association` resource
+
+Reference: [#3622](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3622).
+
+The `snowflake_tag_association` resource was crashing for objects that are created on schema level.
+This was fixed, and now you can create tag associations for objects that are created on schema level, but
+we are advising to use the fields in target resources instead, e.g., `snowflake_table.tag`.
+
 ## v1.2.1 ➞ v2.0.0
 
 ### Supported architectures
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -22,11 +22,10 @@ across different versions.
 
 ### *(bugfix)* Fixed `snowflake_tag_association` resource
 
-Reference: [#3622](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3622).
+The `snowflake_tag_association` resource was crashing when performing the update operation (e.g., because the `tag_value` was changed)
+for objects that are created on schema level. This was fixed, and now you can create tag associations for objects that are created on schema level.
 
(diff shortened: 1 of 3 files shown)


MIGRATION_GUIDE.md
```diff
@@ -18,6 +18,16 @@ across different versions.
...
 
+## v2.0.0 ➞ v2.0.1
+
+### *(bugfix)* Fixed `snowflake_tag_association` resource
+
+Reference: [#3622](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3622).
+
+The `snowflake_tag_association` resource was crashing for objects that are created on schema level.
+This was fixed, and now you can create tag associations for objects that are created on schema level, but
+we are advising to use the fields in target resources instead, e.g., `snowflake_table.tag`.
```
sfc-gh-jmichalak, line 29: Why we are advising this? The advised method is to use association resources because our end goal is to remove `tag` fields, right?

pkg/resources/tag_association_acceptance_test.go
```diff
@@ -687,3 +687,47 @@ resource "snowflake_tag_association" "test" {
 }
 `, tagId.DatabaseName(), tagId.SchemaName(), tagId.Name(), tagValue, sdk.ObjectTypeSchema, schemaId.DatabaseName(), schemaId.Name())
 }
+
+// proves https://github.com/snowflakedb/terraform-provider-snowflake/issues/3622 is fixed
+func TestAcc_TagAssociation_issue_3622(t *testing.T) {
```
sfc-gh-jmichalak, line 692: It seems that this issue was not detected because we tested only account- and database-level objects above. We can also add a test with changed `object_identifiers` field with schema-level objects and with columns.

sfc-gh-jcieslak: As mentioned in the description of the pr, It will currently fail, because our general identifier parsing function won't handle identifier with arguments.

MIGRATION_GUIDE.md
```diff
@@ -18,6 +18,16 @@ across different versions.
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.0.0 ➞ v2.0.1
+
+### *(bugfix)* Fixed `snowflake_tag_association` resource
+
+Reference: [#3622](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3622).
```
sfc-gh-asawicki, line 25: nit: we usually add a reference as a last part of the section (check others entries)

MIGRATION_GUIDE.md
sfc-gh-asawicki: Yes, we should not advise using tags inside objects.

sfc-gh-asawicki (changes requested):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Done

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Done

sfc-gh-asawicki (approved):

PR #3628, opened 2025-04-29
feat: Add image repositories to SDK

sfc-gh-jmichalak:
- Generate SDK code.
- Generate object assertions.
- There are two limitations. They are marked, tested and have linked jiras:
  - UNSET COMMENT does not work - instead we use SET COMMENT=``;
  - creating with names containing lowercase characters does not work - instead we provide a separate schema, similarly to streamlits/warehouses.
- Fix a few typos.


pkg/sdk/testint/image_repositories_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInt_ImageRepositories(t *testing.T) {
+	client := testClient(t)
+	ctx := testContext(t)
+
+	// TODO(SNOW-2070746): The name must not use special characters, including lowercase ones. Remove this after fix on snowflake side.
+	db, dbCleanup := testClientHelper().Database.CreateDatabase(t)
```
sfc-gh-asawicki, line 22: Let's use CreateDatabaseWithParametersSet preemptively (CreateDatabase is still used in a few places, but we will be reducing its usage because of the potential account-level params interaction on lower-level objects).

Also, I don't quite get the explanation here, as there are special characters used in this Database name (test object suffix).

pkg/sdk/testint/image_repositories_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+			HasOwner(snowflakeroles.Accountadmin.Name()).
+			HasOwnerRoleType("ROLE").
+			HasComment(comment).
+			HasPrivatelinkRepositoryUrl(""),
+		)
+	})
+
+	// TODO(SNOW-2070746): The name must not use special characters, including lowercase ones. Remove this after fix on snowflake side.
+	t.Run("create with an ID with lowercase letters fails", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
```
sfc-gh-asawicki, line 79: how do we know there are lowercase characters here?

sfc-gh-asawicki (changes requested):

sfc-gh-jcieslak (approved):

pkg/sdk/testint/image_repositories_integration_test.go
sfc-gh-jmichalak: As we discussed IRL, this setup (and the one below) is correct. The issue was about misunderstanding `_` character, as it's a valid character in image repositories ids. They just need to be compatible with names without quotes - ID like `TEST_TEST` is valid without wrapping in `"`.

Next PR:
- I'll change the database creation to CreateDatabaseWithParametersSet
- I'll adjust the comments slightly to better grasp this limitation

pkg/sdk/testint/image_repositories_integration_test.go
sfc-gh-jmichalak: Similarly to above.

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

Issue #3629, opened 2025-04-29
[Bug]: Granting a database role directly to a user breaks snowflake_grant_database_role resource on subsequent plans


Andrew-Lindsay42:
### Terraform CLI Version

1.9.8

### Terraform Provider Version

2.0.0

### Terraform Configuration

```terraform
resource "snowflake_database" "database" {
    name = "EXAMPLE_DATABASE"
}

resource "snowflake_database_role" "db_role" {
    database = snowflake_database.database.fully_qualified_name
    name = "EXAMPLE_DB_ROLE"
}

resource "snowflake_account_role" "account_role" {
  name = "ACCOUNT_ROLE"
}

resource "snowflake_grant_database_role" "db_role_grant" {
  database_role_name = snowflake_database_role.db_role.fully_qualified_name
  parent_role_name = snowflake_account_role.account_role.fully_qualified_name
}

resource "snowflake_user" "user" {
  name = "JOEBLOGGS@EMAIL.COM"
}
```

### Category

category:grants

### Object type(s)

resource:grant_database_role

### Expected Behavior

On apply, Terraform should create a database, a database role, an account role, a user, and grant the database role to the account role.

```
Plan: 5 to add, 0 to change, 0 to destroy.
snowflake_account_role.account_role: Creating...
snowflake_database.database: Creating...
snowflake_user.user: Creating...
snowflake_account_role.account_role: Creation complete after 0s [id=ACCOUNT_ROLE]
snowflake_user.user: Creation complete after 1s [id=JOEBLOGGS@EMAIL.COM]
snowflake_database.database: Creation complete after 2s [id=EXAMPLE_DATABASE]
snowflake_database_role.db_role: Creating...
snowflake_database_role.db_role: Creation complete after 1s [id="EXAMPLE_DATABASE"."EXAMPLE_DB_ROLE"]
snowflake_grant_database_role.db_role_grant: Creating...
snowflake_grant_database_role.db_role_grant: Creation complete after 1s [id="EXAMPLE_DATABASE"."EXAMPLE_DB_ROLE"|ROLE|"ACCOUNT_ROLE"]

Apply complete! Resources: 5 added, 0 changed, 0 destroyed.
```

If the database role is granted directly to the user through regular SQL outside of Terraform, nothing should happen on a replan.

```
No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

[...]

PR #3633, opened 2025-04-30
fix: issue 3629

sfc-gh-jcieslak:
## Changes
- Fix SHOW GRANTS OF output for newly introduced user privileges in 2025_02 BCR that was pointed out in #3629
- Add migration guide entry
- Add GrantClient helpers for granting roles to users
- Add necessary tests to prove the issue was resolved/is not happening in other granting-role resources


pkg/resources/grant_database_role_acceptance_test.go
```diff
@@ -421,3 +422,63 @@ resource "snowflake_grant_database_role" "test" {
...
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config:            grantDatabaseRoleIssue3629Config(databaseRoleId, parentRoleId),
+				ExternalProviders: acc.ExternalProviderWithExactVersion("2.0.0"),
+			},
+			{
+				PreConfig: func() {
+					acc.TestClient().BcrBundles.EnableBcrBundle(t, "2025_02")
```
sfc-gh-asawicki, line 448: let's set up this BCR test as:
- account level test (to not interfere with other tests enabling this BCR)
- let's do the BCR tests on the secondary account (for the time-being)

pkg/resources/grant_database_role_acceptance_test.go
```diff
@@ -421,3 +422,63 @@ resource "snowflake_grant_database_role" "test" {
...
+				ExpectError:       regexp.MustCompile("Provider produced inconsistent result after apply"),
+			},
+			{
+				ProtoV6ProviderFactories: acc.TestAccProtoV6ProviderFactories,
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction("snowflake_database_role.test", plancheck.ResourceActionNoop),
+					},
+				},
+				Config: grantDatabaseRoleIssue3629Config(databaseRoleId, parentRoleId),
```
sfc-gh-asawicki, line 462: also some expectations here

pkg/sdk/grants_impl.go
```diff
@@ -263,12 +263,14 @@ func (v *grants) Show(ctx context.Context, opts *ShowGrantOptions) ([]Grant, err
...
 			resultList[i].GranteeName = NewAccountObjectIdentifier(granteeName)
-		} else if !slices.Contains([]ObjectType{ObjectTypeRole, ObjectTypeShare}, grant.GrantedTo) {
+		} else if !slices.Contains([]ObjectType{ObjectTypeRole, ObjectTypeShare, ObjectTypeUser}, grant.GrantedTo) {
 			id, err := ParseDatabaseObjectIdentifier(granteeNameRaw)
 			if err != nil {
 				return nil, err
 			}
 			resultList[i].GranteeName = id
+		} else if grant.GrantedTo == ObjectTypeUser {
+			resultList[i].GranteeName = NewAccountObjectIdentifier(strings.TrimLeft(granteeNameRaw, "USER$"))
```
sfc-gh-asawicki, line 273: is this the only place where this affects us?

pkg/sdk/grants_impl.go
```diff
@@ -263,12 +263,14 @@ func (v *grants) Show(ctx context.Context, opts *ShowGrantOptions) ([]Grant, err
...
 			}
 			resultList[i].GranteeName = NewAccountObjectIdentifier(granteeName)
-		} else if !slices.Contains([]ObjectType{ObjectTypeRole, ObjectTypeShare}, grant.GrantedTo) {
+		} else if !slices.Contains([]ObjectType{ObjectTypeRole, ObjectTypeShare, ObjectTypeUser}, grant.GrantedTo) {
 			id, err := ParseDatabaseObjectIdentifier(granteeNameRaw)
 			if err != nil {
 				return nil, err
 			}
 			resultList[i].GranteeName = id
+		} else if grant.GrantedTo == ObjectTypeUser {
```
sfc-gh-asawicki, line 272: if we add this logic, let's make an integration test where we verify that we parse the id correctly

sfc-gh-asawicki (changes requested):

pkg/resources/grant_database_role_acceptance_test.go
```diff
@@ -421,3 +422,63 @@ resource "snowflake_grant_database_role" "test" {
...
+	t.Cleanup(userCleanup)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config:            grantDatabaseRoleIssue3629Config(databaseRoleId, parentRoleId),
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I would say the expectations are implicit (no changes after apply are detected / no error is visible), but I can add explicit plan check.

pkg/resources/grant_database_role_acceptance_test.go
sfc-gh-jcieslak: Done

pkg/sdk/grants_impl.go
```diff
@@ -263,12 +263,14 @@ func (v *grants) Show(ctx context.Context, opts *ShowGrantOptions) ([]Grant, err
 				granteeName = granteeName[strings.IndexRune(granteeName, '.')+1:]
 			}
 			resultList[i].GranteeName = NewAccountObjectIdentifier(granteeName)
-		} else if !slices.Contains([]ObjectType{ObjectTypeRole, ObjectTypeShare}, grant.GrantedTo) {
+		} else if !slices.Contains([]ObjectType{ObjectTypeRole, ObjectTypeShare, ObjectTypeUser}, grant.GrantedTo) {
```
sfc-gh-jmichalak, line 266: Nit: IMO, this block is slowly getting out of hand. How about the following predicates:
- `if !(valueSet(opts.Of) && valueSet(opts.Of.DatabaseRole)) ` - leave as is
- `else if slices.Contains([]ObjectType{ObjectTypeRole, ObjectTypeShare}, grant.GrantedTo)` - handle only ObjectTypeRole and ObjectTypeShare, inside this block, we should move what's currently in `else`
- `else if grant.GrantedTo == ObjectTypeUser` - leave as is
- `else` - move what's in this block (the specific cases would be handled above; here we use a "default" behavior)

Alternatively, we can extract a separate block with `if !(valueSet(opts.Of) && valueSet(opts.Of.DatabaseRole))`, as this condition is slightly different than checking ObjectTypes, and use a separate switch for different ObjectTypes. WDYT?

pkg/sdk/grants_impl.go
sfc-gh-jcieslak: Done

pkg/resources/grant_database_role_acceptance_test.go
sfc-gh-jmichalak: Still, it's marked as `//go:build !account_level_tests`. I believe this test should be extracted and marked as `//go:build account_level_tests`.

sfc-gh-jmichalak (commented):
Please add an entry to the migration guide.

pkg/acceptance/helpers/execute_client..go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/require"
+)
+
+type ExecuteClient struct {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

MIGRATION_GUIDE.md
```diff
@@ -18,6 +18,18 @@ across different versions.
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.0.0 ➞ v2.0.1
+
+### *(bugfix)* Fixed snowflake_grant_database_role resource
+granting a database role directly to a user breaks
```
sfc-gh-jmichalak, line 24:
```suggestion
Granting a database role directly to a user breaks.
```

sfc-gh-jmichalak (approved):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Added by accident, removed.

pkg/sdk/grants_impl.go
sfc-gh-jcieslak: I agree that we should clean this up, but I would do it as part of a dedicated ticket. I'll create one, leave todo, and link this comment in the ticket.

sfc-gh-jmichalak (approved):

pkg/resources/grant_database_role_acceptance_test.go
sfc-gh-asawicki: as above

pkg/acceptance/helpers/grant_client.go
```diff
@@ -362,3 +363,22 @@ func (c *GrantClient) ShowGrantsToDatabaseRole(t *testing.T, databaseRoleId sdk.
...
+func (c *GrantClient) GrantDatabaseRoleToUser(t *testing.T, databaseRoleId sdk.DatabaseObjectIdentifier, userId sdk.AccountObjectIdentifier) {
+	t.Helper()
+	ctx := context.Background()
+
+	// TODO(SNOW-2095669): Update when the client is updated to support this
+	_, err := c.context.client.ExecForTests(ctx, fmt.Sprintf("GRANT DATABASE ROLE %s TO USER %s", databaseRoleId.FullyQualifiedName(), userId.FullyQualifiedName()))
+	require.NoError(t, err)
+}
+
+func (c *GrantClient) GrantAccountRoleToUser(t *testing.T, accountRoleId sdk.AccountObjectIdentifier, userId sdk.AccountObjectIdentifier) {
```
sfc-gh-asawicki, line 376: there is already existing method `GrantRoleToUser` in the `RoleClient`

pkg/resources/grant_account_role_acceptance_test.go
```diff
@@ -227,3 +233,64 @@ resource "snowflake_grant_account_role" "test" {
...
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config:            accconfig.FromModels(t, providerModel) + grantAccountRoleIssue3629Config(accountRoleId, parentRoleId),
+				ExternalProviders: acc.ExternalProviderWithExactVersion("2.0.0"),
```
sfc-gh-asawicki, line 258: why we are pinning a version of the provider in this test? We don't want this BCR to affect the newest provider, right?

pkg/resources/grant_database_role_account_level_acceptance_test.go
```diff
@@ -0,0 +1,95 @@
...
+					assert.Check(resource.TestCheckResourceAttr("snowflake_grant_database_role.test", "id", helpers.EncodeResourceIdentifier(databaseRoleId.FullyQualifiedName(), sdk.ObjectTypeRole.String(), parentRoleId.FullyQualifiedName()))),
+				),
+			},
+		},
+	})
+}
+
+func grantDatabaseRoleIssue3629Config(databaseRoleId sdk.DatabaseObjectIdentifier, accountRoleId sdk.AccountObjectIdentifier) string {
+	return fmt.Sprintf(`
+resource "snowflake_database_role" "test" {
```
sfc-gh-asawicki, line 81: do we need to create them here, instead of quick setup in test? Also, why not use builders for them (for the existing ones so db role and acc role)?

pkg/resources/grant_database_role_account_level_acceptance_test.go
```diff
@@ -0,0 +1,95 @@
...
+	t.Cleanup(userCleanup)
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() { acc.TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config:            config.FromModels(t, providerModel) + grantDatabaseRoleIssue3629Config(databaseRoleId, parentRoleId),
```
sfc-gh-asawicki, line 44: do we need the first step? I guess such simple behavior is already documented in some other test

sfc-gh-asawicki (changes requested):

pkg/resources/grant_database_role_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: It's already checked, maybe GitHub didn't show that in a good way (see `grant_database_role_account_level_acc_test.go`, I'm asserting `ids` there)

pkg/resources/grant_database_role_acceptance_test.go
sfc-gh-jcieslak: It is extracted now

pkg/resources/grant_database_role_acceptance_test.go
sfc-gh-jcieslak: The id checks are there

pkg/acceptance/helpers/grant_client.go
sfc-gh-jcieslak: I'll remove it here then, but I'm wondering then if we should put all grant methods here instead of in e.g. RoleClient or DatabaseRoleClient that are dedicated for role/database role management. It would be better to have them in GrantClient (I think).

pkg/resources/grant_account_role_acceptance_test.go
sfc-gh-jcieslak: You're right, we can use the current version here

pkg/resources/grant_database_role_account_level_acceptance_test.go
sfc-gh-jcieslak: Right, I look at other tests and assumed they're not implemented yet. Refactored.

pkg/resources/grant_database_role_account_level_acceptance_test.go
sfc-gh-jcieslak: The issue occurs in Read operation, given that I think the first step could be removed as Read occurs right after Create.

PR #3645, opened 2025-05-06
fix: Bring back account retry functionality

sfc-gh-jcieslak:
## Changes
- Bring back the retry functionality in the account's Create operation, as it may still take some time to create a new account when creating one in other region than the current account.


pkg/resources/usage_tracking_acceptance_test.go
```diff
@@ -25,7 +25,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
-	"github.com/stretchr/testify/require"
```
sfc-gh-asawicki, line 28: remove the upper one pls and leave this one

sfc-gh-asawicki (changes requested):

pkg/resources/account.go
```diff
@@ -282,6 +286,17 @@ func CreateAccount(ctx context.Context, d *schema.ResourceData, meta any) diag.D
 		return diag.FromErr(err)
 	}
 
+	if err := util.Retry(5, 3*time.Second, func() (error, bool) {
+		_, err = client.Accounts.ShowByID(ctx, id)
+		if err != nil {
+			log.Printf("[DEBUG] retryable operation resulted in error: %v", err)
+			return nil, false
```
sfc-gh-jmichalak, line 293: Why are we not returning an error here? Shouldn't we do this only for "not found" errors?

sfc-gh-jmichalak (changes requested):

pkg/resources/account.go
sfc-gh-jcieslak: The `util.Retry` would leave early if we would return error here. We can only log here and return the error at the end after all retries.

pkg/resources/usage_tracking_acceptance_test.go
sfc-gh-jcieslak: Done

sfc-gh-asawicki (approved):

pkg/resources/account.go
sfc-gh-jmichalak: What about non-retriable operations, like authentication errors, etc?

sfc-gh-jcieslak: Then we would fail anyway in Read operation that happens right after Create one.

sfc-gh-jmichalak (dismissed):

sfc-gh-jmichalak (approved):

PR #3646, opened 2025-05-06
fix: Set value into correct field for field transformers

yoiki:
This PR fixes an issue in the field transformers where certain fields were not being correctly assigned when set ddl options.  
The bug specifically affected cases where both `equals` and `parentheses` were set on a field.

The only affected case found was in `secrets_def.go`, so the DDL was regenerated for that file using the SDK.

## Test Plan
* [x] integration tests for `secrets_gen.go`


sfc-gh-jcieslak (approved):

sfc-gh-asawicki:
Thanks for the contribution @yoiki. Merged :)

Issue #2624, opened 2024-03-15
Allow configuring `storage_aws_external_id` for `snowflake_storage_integration`


AleksaC:
### Terraform CLI and Provider Versions

Terraform v1.7.5
on darwin_arm64

### Use Cases or Problem Statement

There is a cycle between the storage integration and role used for the storage integration since the role needs Snowflake AWS user arn as well as external id in its trust policy. Since user is the same for all storage integrations in the same snowflake account it can be provided as a variable, but that still leaves external id. 

Fortunately Snowflake [allows setting](https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration#cloud-provider-parameters-cloudproviderparams) external id when creating the storage integration so I could just generate the storage integration and pass it to both the role as well as the storage integration removing the cycle.

### Proposal

Make the `storage_aws_external_id` field `Optional` instead of `Computed` since it's supported in the underlying API.

### How much impact is this issue causing?

Medium

sfc-gh-jcieslak:
Hey @AleksaC. Thanks for reaching out to us.

We are currently working on redesigning existing resources as part of https://github.com/Snowflake-Labs/terraform-provider-snowflake/blob/main/ROADMAP.md#supporting-all-snowflake-ga-features. We will consider it when we will be going over `snowflake_storage_integration`.

AleksaC:
Thanks for the response @sfc-gh-jcieslak. 

Do you have any ETA? Would you accept a PR to expedite this, since the change seems relatively straightforward to implement and shouldn't impact the redesign?

parsoj:
I'm also seeing this issue (a cyclic dependency between the AWS IAM role, and the Snowflake storage integration) - 

However I believe the fix is a bit more involved than just allowing the user to set `external_id` 

AWS IAM roles require a valid assume role policy to be defined when the role is created, not afterwards. There is a potential work-around where you can:

1. create the IAM role, and give the role a secondary "vestigial" assume_role policy (it can't be an empty stub though - it has to point to a real&valid IAM Principal) 
2. create the storage integration, using the ARN of the now created role
3. create a separate IAM policy document that allows sts:AssumeRole to that user & external ID. 

This isn't ideal though - since it basically requires us to give more permissions than we'd otherwise want - and it obscures the actual permissions for the IAM role since the actual assume role policy isn't the main attached one. 

[...]

PR #3659, opened 2025-05-09
feature: Support setting `storage_aws_external_id` in storage integration resource

yoiki:
## Summary

This PR adds support for specifying the `storage_aws_external_id` field when creating and updating a Snowflake storage integration resource.

Previously, the field was treated as `Computed` and could not be explicitly set via Terraform, which caused issues in scenarios involving AWS IAM role trust relationships requiring a fixed external ID.

## What was changed

- Updated the Create and Alter function in sdk to pass the value to the Snowflake API if specified.
- Changed `storage_aws_external_id` attribute to be `Optional`.
- Added acceptance test to verify the behavior.
- Updated doc of storage integration.

## Test Plan
* [x] integration tests for storage integration
* [ ] acceptance tests for storage integration

## References

- Resolves: https://github.com/snowflakedb/terraform-provider-snowflake/issues/2624


pkg/resources/storage_integration.go
```diff
@@ -373,6 +376,17 @@ func UpdateStorageIntegration(ctx context.Context, d *schema.ResourceData, meta
 			}
 		}
 
+		if d.HasChange("storage_aws_external_id") {
+			if v, ok := d.GetOk("storage_aws_external_id"); ok {
+				s3SetParams.WithStorageAwsExternalId(v.(string))
+			} else {
+				if err := client.StorageIntegrations.Alter(ctx, sdk.NewAlterStorageIntegrationRequest(id).
```
yoiki, line 383: Currently, when both storage_aws_object_acl and storage_aws_external_id are updated, two separate ALTER queries are executed.
I think it would be better to combine them into a single query, but I’m not sure what the best approach would be.
I’d appreciate any suggestions on how to improve this!

templates/resources/storage_integration.md.tmpl
```diff
@@ -0,0 +1,38 @@
+---
+page_title: "{{.Name}} {{.Type}} - {{.ProviderName}}"
```
sfc-gh-jmichalak, line 2: I believe this file is not needed. Please remove it and regenerate docs with `make pre-push`.

pkg/resources/storage_integration_acceptance_test.go
```diff
@@ -38,6 +38,47 @@ func TestAcc_StorageIntegration_Empty_StorageAllowedLocations(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_storage_integration.test", "name", id.Name()),
+					resource.TestCheckResourceAttr("snowflake_storage_integration.test", "enabled", "true"),
+					resource.TestCheckResourceAttr("snowflake_storage_integration.test", "storage_aws_role_arn", awsRoleArn),
+					resource.TestCheckResourceAttr("snowflake_storage_integration.test", "storage_aws_external_id", awsExternalId),
+					resource.TestCheckResourceAttr("snowflake_storage_integration.test", "storage_allowed_locations.#", "1"),
+					resource.TestCheckResourceAttr("snowflake_storage_integration.test", "storage_allowed_locations.0", "s3://foo/"),
+				),
+			},
+		},
```
sfc-gh-jmichalak, line 78: We can also add a test step for detecting external changes - use something similar to https://github.com/snowflakedb/terraform-provider-snowflake/blob/spcs/pkg/resources/tag_acceptance_test.go#L127, but first add a helper function in https://github.com/snowflakedb/terraform-provider-snowflake/blob/spcs/pkg/acceptance/helpers/storage_integration_client.go#L27.

pkg/resources/storage_integration.go
sfc-gh-jmichalak: This resource's implementation is quite old, and it hasn't been reworked yet. IMO, for now, we can leave it as (follow the former behavior). 

If you want to improve this, you can just mimic the SET behavior - use
```
	var runUnsetStatement bool
	unsetReq := sdk.NewStorageIntegrationUnsetRequest()
```
above, and execute unset once below `if runSetStatement`. Then, in this block, the unset version should look similar to the set one. Please also remember about unset in `storage_blocked_locations`.

sfc-gh-jmichalak (changes requested):

pkg/resources/storage_integration.go
yoiki: @sfc-gh-jmichalak Thanks for the suggestion! I’ve addressed this in [daca2b1](https://github.com/snowflakedb/terraform-provider-snowflake/pull/3659/commits/daca2b1f84a06002eb1698d41580eaf40c64c2d5).
Please take another look when you have time.

templates/resources/storage_integration.md.tmpl
yoiki: @sfc-gh-jmichalak Got it! I’ve removed the file and ran make pre-push.  
Addressed in [76c18f0](https://github.com/snowflakedb/terraform-provider-snowflake/pull/3659/commits/76c18f0328068c5a43041e34901bd46b9ec593dc).

I’m still not totally clear on why the template isn’t necessary—if you don’t mind, I’d love to learn more about that.
Addressed in

pkg/resources/storage_integration_acceptance_test.go
yoiki: @sfc-gh-jmichalak Thanks! I’ve addressed this in [0129e85](https://.../). Let me know if anything’s still missing.

pkg/resources/storage_integration_acceptance_test.go
```diff
@@ -146,10 +190,17 @@ func TestAcc_StorageIntegration_AWS_Update(t *testing.T) {
 					resource.TestCheckResourceAttr("snowflake_storage_integration.test", "storage_allowed_locations.0", "s3://foo/"),
 					resource.TestCheckNoResourceAttr("snowflake_storage_integration.test", "storage_blocked_locations"),
 					resource.TestCheckNoResourceAttr("snowflake_storage_integration.test", "storage_aws_object_acl"),
+					resource.TestCheckNoResourceAttr("snowflake_storage_integration.test", "storage_aws_external_id"),
 					resource.TestCheckResourceAttr("snowflake_storage_integration.test", "comment", ""),
 				),
 			},
 			{
+				PreConfig: func() {
```
sfc-gh-jcieslak, line 198: I would separate the steps that are applying configuration changes from the ones that should apply changes after external ones. I propose to:
- Start with previous setup: `unset` -> `set` -> `unset`
- Copy-paste two steps so that we end up with: `unset` -> `set` -> `set` (ex) -> `unset` -> `unset` (ex)
- The two steps that are marked with `(ex)` will serve as steps where we'll put PreConfig that will:
  - Unset all fields (that are possible to unset) in the `set` step
  - Set all (or just the ones that are set by the `set` step) AWS fields in the `unset` step

That way, we can be sure that altering storage integration is working from both configuration and external changes, wdyt?

yoiki: @sfc-gh-jcieslak Thanks! I applied your suggestion in [5560d35](https://github.com/snowflakedb/terraform-provider-snowflake/commit/5560d35feb08a964f8d8202b1be7067eb7d979cb) — the test now follows the unset -> set -> set (ex) -> unset -> unset (ex) pattern with PreConfig handling the external modifications as proposed. Let me know if you think any further adjustment is needed!

pkg/resources/storage_integration.go
```diff
@@ -65,7 +66,7 @@ var storageIntegrationSchema = map[string]*schema.Schema{
 	},
 	"storage_aws_external_id": {
 		Type:        schema.TypeString,
-		Computed:    true,
+		Optional:    true,
```
sfc-gh-asawicki, line 69: I think it's not enough to set it as optional, as it is filled out as part of the read. Because of that, in case of not setting this value in terraform config, it will be read and filled out which will make the terraform to recognize it as diff (permanent diff).
First thought is of course to set it as optional+computed but we have found out the hard way, that it has some other limitations. You can read more in https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/v1-preparations/CHANGES_BEFORE_V1.md#config-values-in-the-state.
Here, I would suggest handling it through show_output changes. You can check the user's resource `display_name` attribute https://github.com/snowflakedb/terraform-provider-snowflake/blob/976b36c5702b5566eef7a1db04142f7920b7c188/pkg/resources/user.go#L44.
We can discuss further isf this is unclear.

sfc-gh-asawicki (changes requested):
LGTM. I have just one remark in addition to everything that was already written.

sfc-gh-jcieslak (approved):

pkg/resources/storage_integration_acceptance_test.go
sfc-gh-jcieslak: Looks good :)

templates/resources/storage_integration.md.tmpl
sfc-gh-jmichalak: It works like this:
- If a file like `templates/resources/<resource_name>.md.tmpl` exists, the generator uses such file.
- Else, it uses `templates/resources.md.tmpl`
- Ideally, we'd like to use `templates/resources.md.tmpl` to reduce the number of custom files. Sometimes we need to use such custom files to add notes about limitations or to mark a resource as stable. 
- No customization is needed in this case, so that this file can be removed.

I hope it's clearer :)

sfc-gh-jmichalak (approved):
LGTM, I also adjusted the PR description, as text like `Resolves #2624` may automatically close the related issue, and we don't want that.

pkg/resources/storage_integration.go
yoiki: @sfc-gh-asawicki Thanks for the feedback!
I addressed this in [4419d61](https://github.com/snowflakedb/terraform-provider-snowflake/commit/4419d61100ccf25372994fbe24a6e98a45d2d69d).
Instead of relying on `SHOW`, I switched to using the value from the `DESCRIBE` properties, since `storage_aws_external_id` appears to be included there.

I’m not entirely confident if this is the correct approach, so I’d really appreciate it if you could take a look and let me know what you think.

sfc-gh-jmichalak (approved):

pkg/resources/storage_integration.go
sfc-gh-jcieslak: Changes in [4419d61](https://github.com/snowflakedb/terraform-provider-snowflake/commit/4419d61100ccf25372994fbe24a6e98a45d2d69d) should be enough to prevent unwanted plan diffs 👍

sfc-gh-jcieslak (dismissed):

pkg/resources/storage_integration.go
```diff
@@ -132,6 +135,9 @@ func StorageIntegration() *schema.Resource {
 			StateContext: schema.ImportStatePassthroughContext,
 		},
 		Timeouts: defaultTimeouts,
+		CustomizeDiff: customdiff.All(
+			ComputedIfAnyAttributeChanged(storageIntegrationSchema, DescribeOutputAttributeName, "storage_aws_external_id", "enabled", "comment"),
```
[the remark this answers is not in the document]
yoiki: @sfc-gh-jcieslak 
I’m starting to work on this, but I noticed that the DescribeOperation definition for storage_integration in pkg/sdk/storage_integration_def.go is quite different from how other resources define their properties.

Specifically, as seen [here (lines 160–178)](https://github.com/snowflakedb/terraform-provider-snowflake/blob/4419d61100ccf25372994fbe24a6e98a45d2d69d/pkg/sdk/storage_integration_def.go#L160-L178), the output of DESCRIBE is represented as generic fields like property, property_type, and property_value, which map directly to the columns returned by the Snowflake DESCRIBE command.

This differs from other resources where the schema properties are explicitly defined — for example, see [streamlit_def.go (lines 109–141)](https://github.com/snowflakedb/terraform-provider-snowflake/blob/4419d61100ccf25372994fbe24a6e98a45d2d69d/pkg/sdk/streamlits_def.go#L109-L141).

Because of this difference, I think we should agree on how to implement the DescribeStorageIntegrationToSchema function. Personally, I suggest we follow the same structure used in other `XXToSchema` implementations (e.g., [security_integration_gen.go](https://github.com/snowflakedb/terraform-provider-snowflake/blob/4419d61100ccf25372994fbe24a6e98a45d2d69d/pkg/schemas/security_integration_gen.go)):
* Use the value of the property field as the key (e.g., "STORAGE_AWS_EXTERNAL_ID"),
* Use the corresponding property_value as the value.

If this approach makes sense, I’ll proceed with implementing it accordingly.

Also, if we feel that the current DESCRIBE schema for storage_integration in the SDK should be refactored to match the structure used in other resources, I’d suggest doing that in a separate PR to keep the scope manageable — since it might involve larger changes.

Let me know what you think!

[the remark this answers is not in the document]
sfc-gh-jcieslak: Some describe outputs are more structured with a defined schema, and some are just "key-value" outputs, like in this case. It's not that uncommon, you can find something similar in the [external_oauth_security_integration](https://github.com/snowflakedb/terraform-provider-snowflake/blob/4419d61100ccf25372994fbe24a6e98a45d2d69d/pkg/schemas/external_oauth_security_integration.go), where we define the schema and what keys we are expecting. Then, during the mapping, we receive the list of properties and map them based on their key name (which, I assume, is what you are proposing here; if yes, I think we should go with this approach).

I don't think we need to refactor the SDK in this case, as we have similar mappings in other "key-value" describe outputs. Maybe those struct mappings should be moved to the SDK at some point, but we would have to also touch other existing ones, so let's leave the current approach as it is. I'm guessing this is something we would like to have in the future, but as a nice-to-have rather than a must-have (but that's a good proposal, so I'll still discuss this with the team, to see what they think, and create a corresponding ticket for it).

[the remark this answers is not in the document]
yoiki: @sfc-gh-jcieslak Thanks for the detailed feedback!

Yes—external_oauth_security_integration is exactly the pattern I had in mind. I’ll move forward using that implementation as a reference.

I’m also happy to stick with the current approach for now. Looking ahead, I still feel the most maintainable solution would be to declare the property schema explicitly in the SDK and generate the mappings automatically. Even a smaller refactor that just unifies how we declare these properties in the SDK seems like a realistic next step.

In any case, this is all future-looking, so I’d be glad if the team could discuss it when the time is right:)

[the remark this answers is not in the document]
sfc-gh-jcieslak: Great! So we discussed it today, and we would prefer to put this aside for now (I'll create a future improvement ticket for this). We agree that the proposed approach (schema moved to SDK) is something we would like to aim for, but we don't think it's worth taking right now, as there are more important topics to address. In this case, specifically, you would have to change the way properties are read in the storage integration, which we plan to split anyway; that's why we would like to limit the work in this resource to a minimum.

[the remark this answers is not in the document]
sfc-gh-jcieslak: Hey @yoiki 👋, just wanted to check in and ask if you need any help regarding this ⬆️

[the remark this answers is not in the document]
yoiki: @sfc-gh-jcieslak 
I apologize for the delayed response. I've implemented the suggested changes in commit 04029d6f0d66a4b9a414da5ef5de51301fb60d3a. The changes include:

- Removing direct setting of storage_aws_external_id from describe output
- Adding external changes detection for storage_aws_external_id
- Always setting state from config values for storage_aws_external_id
- Updating describe output handling to use DescribeStorageIntegrationToSchema

Please review the changes and let me know if there are any adjustments needed. Thank you for your patience and guidance.

pkg/resources/storage_integration.go
```diff
@@ -201,108 +313,7 @@ func CreateStorageIntegration(ctx context.Context, d *schema.ResourceData, meta
...
-			}
-		case "AZURE_MULTI_TENANT_APP_NAME":
-			if err := d.Set("azure_multi_tenant_app_name", prop.Value); err != nil {
-				return diag.FromErr(err)
-			}
-		}
-	}
-
-	return diag.FromErr(err)
+	return GetReadStorageIntegrationFunc(true)(ctx, d, meta)
```
sfc-gh-jcieslak: here `GetReadStorageIntegrationFunc` should take in `false` instead of `true`.

pkg/resources/storage_integration.go
```diff
@@ -110,9 +114,111 @@ var storageIntegrationSchema = map[string]*schema.Schema{
...
+			Schema: schemas.DescribeStorageIntegrationSchema,
+		},
+	},
 	FullyQualifiedNameAttributeName: schemas.FullyQualifiedNameSchema,
 }
 
+func GetReadStorageIntegrationFunc(withExternalChangesMarking bool) schema.ReadContextFunc {
+	return func(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+		client := meta.(*provider.Context).Client
+		id, ok := helpers.DecodeSnowflakeID(d.Id()).(sdk.AccountObjectIdentifier)
```
sfc-gh-jmichalak: Nit: we should use 
```
	id, err := sdk.ParseAccountObjectIdentifier(d.Id())
	if err != nil {
		return nil, err
	}
```

pkg/resources/storage_integration.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: This part LGTM.

sfc-gh-jmichalak (commented):
Thanks for addressing the comments. It looks good in general. However, recently, we moved acceptance tests from `pkg/resources` to `pkg/testacc` and changed the filename convention. Please merge the newest `dev` branch and resolve conflicts (the acceptance test and `testdata` files should be placed in `pkg/testacc`; the change should be pretty straightforward).

timols:
Hi @yoiki, any chance you're able to resolve the conflicts so this PR can be merged? Happy to help if needed!

templates/resources/storage_integration.md.tmpl
sfc-gh-jmichalak: I brought it back because of the unusual handling of `describe_output`, which must be documented. I also regenerated the docs.

pkg/resources/storage_integration.go
sfc-gh-jmichalak: Done.

pkg/resources/storage_integration.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: Reworked a bit.

pkg/resources/storage_integration.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/storage_integration_gen_integration_test.go
```diff
@@ -308,6 +316,7 @@ func TestInt_StorageIntegrations(t *testing.T) {
 					WithStorageAwsObjectAcl(true).
 					WithEnabled(true).
 					WithStorageBlockedLocations(true).
+					WithStorageAwsExternalId(true).
```
sfc-gh-jcieslak, line 319: I think assert wan't adjusted to see if this unsets external id on the Snowflake side

sfc-gh-asawicki (approved):

PR #3660, opened 2025-05-09
feat: Add Image Repository resource

sfc-gh-jmichalak:
Add a new snowflake_image_repository resource.

## TODO
- Add a new snowflake_image_repositories data source.


sfc-gh-asawicki (changes requested):

MIGRATION_GUIDE.md
```diff
@@ -18,6 +18,11 @@ across different versions.
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.0.1 ➞ v2.1.0
+
+### *(new feature)* snowflake_image_repository resource
+Added a new resource for managing image repositories. See reference [docs](https://docs.snowflake.com/en/sql-reference/sql/create-image-repository). The limitation of this resource is that quoted names for special characters or case-sensitive names are not supported. The same constraint also applies to database and schema names where you create an image repository. This limitation in the provider follows the limitation in Snowflake (see the linked docs).
```
sfc-gh-jcieslak, line 24: We could add short sentence what to use given the limitations: `...or case-sensitive names are not supported...` (next sentence ->) Please, use only upper-cased identifiers with `_` as separator (e.g. SOME_IDENTIFIER).
> I'm guessing that `_` is fine?

pkg/resources/image_repository.go
```diff
@@ -0,0 +1,149 @@
...
+	)
+	return &schema.Resource{
+		CreateContext: TrackingCreateWrapper(resources.ImageRepository, CreateImageRepository),
+		ReadContext:   TrackingReadWrapper(resources.ImageRepository, ReadImageRepository),
+		UpdateContext: TrackingUpdateWrapper(resources.ImageRepository, UpdateImageRepository),
+		DeleteContext: TrackingDeleteWrapper(resources.ImageRepository, deleteFunc),
+		Description:   "Resource used to manage image repositories. For more information, check [image repositories documentation](https://docs.snowflake.com/en/sql-reference/sql/create-image-repository).",
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.ImageRepository, customdiff.All(
+			ComputedIfAnyAttributeChanged(imageRepositorySchema, ShowOutputAttributeName, "comment"),
```
sfc-gh-jcieslak, line 73: I don't think this is needed

pkg/resources/image_repository.go
```diff
@@ -0,0 +1,149 @@
...
+					Summary:  "Failed to query image repository. Marking the resource as removed.",
+					Detail:   fmt.Sprintf("Image repository id: %s, Err: %s", id.FullyQualifiedName(), err),
+				},
+			}
+		}
+		return diag.FromErr(err)
+	}
+	errs := errors.Join(
+		d.Set(ShowOutputAttributeName, []map[string]any{schemas.ImageRepositoryToSchema(imageRepository)}),
+		d.Set(FullyQualifiedNameAttributeName, imageRepository.ID().FullyQualifiedName()),
```
sfc-gh-jcieslak, line 126: I'm wondering if we should use `imageRepository` object with id parts returned from Snowflake or `id` that consists of the parts that are exactly the same as in the configuration. Theoretically this shouldn't matter, but I guess in some situations can.

sfc-gh-jcieslak (dismissed):

pkg/resources/image_repository.go
sfc-gh-jmichalak: Why? The `comment` is present in the show output schema and it can be altered externally.

sfc-gh-jcieslak: Because `comment` is not handled in Read by `handleExternalChangesToObjectInShow` (and that's good). When the field is directly put into the field without any need to check `show_output` it doesn't need to be mentioned here, that's why I think whole `CustomizeDiff` could be removed in this case.

sfc-gh-jmichalak: Still, `comment` is present in `ComputedIfAnyAttributeChanged` calls in warehouses, tasks and other resources, and in these cases it's also not handled by `handleExternalChangesToObjectInShow` and `setStateToValuesFromConfig`, so I think we should keep this.

docs/resources/image_repository.md
```diff
@@ -0,0 +1,90 @@
+---
+page_title: "snowflake_image_repository Resource - terraform-provider-snowflake"
+subcategory: "Preview"
+description: |-
+  Resource used to manage image repositories. For more information, check image repositories documentation https://docs.snowflake.com/en/sql-reference/sql/create-image-repository.
+---
+
+-> **Note** Quoted names for special characters or case-sensitive names are not supported. The same constraint also applies to database and schema names where you create an image repository. That is, database and schema names without quotes are valid when creating an image repository.
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Well, it's linked in the resource description. Added here as well.

pkg/resources/image_repository.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func UpdateImageRepository(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+	id, err := sdk.ParseSchemaObjectIdentifier(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	set := sdk.NewImageRepositorySetRequest()
+	if d.HasChange("comment") {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Done.

pkg/resources/image_repository.go
```diff
@@ -0,0 +1,149 @@
...
+
+func CreateImageRepository(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+	name := d.Get("name").(string)
+	schemaName := d.Get("schema").(string)
+	database := d.Get("database").(string)
+	id := sdk.NewSchemaObjectIdentifier(database, schemaName, name)
+
+	request := sdk.NewCreateImageRepositoryRequest(id)
+	if v, ok := d.GetOk("comment"); ok {
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Done.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: It's not exactly this set. The ID needs to be compatible with unquoted identifiers, I linked the docs here.

pkg/resources/image_repository.go
sfc-gh-jmichalak: Changed to id.FullyQualifiedName() for consistency, but I don't think it matters ultimately, because we call SHOW with ID from the state, so the IDs should match.

pkg/resources/image_repository.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/schemas"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+var imageRepositorySchema = map[string]*schema.Schema{
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: It's not documented because it does not work properly. For now, we are sticking to ForceNew.

sfc-gh-asawicki (approved):

pkg/resources/image_repository.go
sfc-gh-jcieslak: Then I would say it should be removed from there also as it doesn't serve any purpose, but fortunately it won't cause any issues so If you left it, it won't do anything

PR #3679, opened 2025-05-15
chore: Extract a separate directory for the acceptance tests

sfc-gh-asawicki:
Extract a separate directory for the acceptance tests

This change is the essential next step to simplify incoming tests of the terraform plugin framework capabilities.

### Motivation
For quite some time, we have considered setting up the acceptance tests similarly to the SDK integration tests. The main reasons to do it:
- Running acceptance tests locally does not clean up the database, schema, and warehouse created for acceptance tests (on CI, they are removed as part of sweepers; integration tests are cleaned up locally) - it will reduce the time after which certain objects are removed from our test accounts.
- We have different setups for acceptance and integration tests, when these setups should be almost the same (we recreate the same set of objects, we run the same tests); by doing it, we will be able to unify the setups.
- Test resources are located in the same package as the production resources. By extracting the acceptance tests, we mark the border; there will be a separate directory extracted only for test resources and their tests
- Acceptance tests currently reuse the same setup, helpers, etc., and they are located in three different packages, which leads to three separate setups taking time.

### Changes
- new package `testacc` is meant to contain all resource, datasource, and provider tests; it won't contain the test resources tests
- two tests (1 resource, 1 datasource; 1 account level, 1 not) moved to the new directory
  - `EnableAcceptance` env checks were removed as now it's handled inside the `TestMain`, the same with the initial call to `acc.TestAccPreCheck(t)`;
  - `acc.TestAccPreCheck(t)` method was left but empty and not removed from the `PreCheck` part of the acceptance tests (if we want to check for something else);
  - dependencies to old acceptance package were dropped (`acc.` replaced with calls to the copied methods)
  - `TestClient()` -> `testClient()`
- common setup of db, sc, and wh extracted as new methods in the `TestClient` and used in integration tests and the new package
- common integration/acceptance checks extracted as new methods in the `TestClient`
- setting up the provider extracted as `setUpProvider`
- check destroy and provider helper methods were copied to the new directory (to allow future possibility of moving all the tests)
- ci config adjusted
  - to run tests in the new directory (for account and non-account level features)
  - to separately run unit and integration tests
- unit tests in Makefile are now run by integration/acceptance test packages exclusion
- Makefile recipes adjusted to use concrete packages instead of `./...`
- One of the ChecKDestroy methods was renamed
- Unused methods removed
- Added numbers to file names in the crowded packages (so that the important setup, commons are higher in IDE) - this is the idea, which we can discuss

### Next PRs
- move all acceptance tests to this new package if aapproved
- adjust Makefile and ci jobs definitions after all tests are moved
- extract separate directory for the test resources (where we test terraform libraries behavior)
- solve TODOs with `[next PRs]`
  - Makefile cleanup
  - Use `DropSafely` for common db, sc, wh setup
  - some setup common functions were extracted; extract more and reuse between acceptance and integration tests (like `timer` and `setUpSdkClient`)
  - potentially extract test context logic into separate directory
  - decide what do we do with `SimplifiedIntegrationTestsSetup` for acceptance tests
  - rework provider setup in tests when working on terraform plugin framework PoC


pkg/testacc/2_check_destroy_test.go
```diff
@@ -0,0 +1,641 @@
...
+				return fmt.Errorf("%s %v still exists", resource, id.FullyQualifiedName())
+			} else {
+				t.Logf("resource %s (%v) was dropped successfully in Snowflake", resource, id.FullyQualifiedName())
+			}
+		}
+		return nil
+	}
+}
+
+func decodeSnowflakeId(rs *terraform.ResourceState, resource resources.Resource) (sdk.ObjectIdentifier, error) {
```
sfc-gh-jcieslak, line 71: I know this is a copy, but this function and the above CheckDestroy seem incorrect with the identifiers that are parsed here and then passed to the showByIdFunctions. Currently, the resource identifiers are inconsistent as some of the resources still use `DecodeSnowflakeID`. It's also strange, because it should raise an error (or a warning) in this case, but either I'm wrong or this was missed somehow because it's only logged and not errors out.

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (approved):
LGTM, it would be great to have a bit smaller PRs, esp. we could split moving and refactoring parts.

pkg/testacc/2_check_destroy_test.go
sfc-gh-asawicki: Yeah, it's strange, I extracted it as SNOW-2110727

sfc-gh-asawicki: Addressed in #3967.

Issue #3690, opened 2025-05-20
[Bug]: Empty Privileges Array Causes State File Corruption


yoshito-ueki:
### Terraform CLI Version

1.10.0

### Terraform Provider Version

1.0.0

### Company Name

CARTA MARKETING FIRM Inc.

### Terraform Configuration

```terraform
resource "snowflake_grant_privileges_to_database_role" "table_privileges" {
  database_role_name = snowflake_database_role.this.fully_qualified_name
  privileges         = []
  on_schema_object {
    all {
      object_type_plural = "TABLES"
      in_database        = snowflake_database_role.this.database
    }
  }
}
```

### Category

category:resource, category:snowflake, category:grants

### Object type(s)

resource:grant_account_role, resource:grant_database_role

### Expected Behavior

- When an error occurs due to privileges=[] setting, the state file should not be updated to an inconsistent state
- After an error occurs, the resource identifier in the state file should either be automatically removed or the original correct state should be maintained
- This would allow terraform plan and terraform apply to run without issues after the error is fixed

### Actual Behavior

- When running apply with privileges=[] setting, an error occurs during Snowflake provider processing
- However, by the time the error occurs, the state file has already been overwritten with privileges=[]
- This creates an inconsistency between Snowflake's actual configuration and Terraform state
- In this state, terraform plan and terraform apply cannot be executed
- Operations cannot continue unless the resource identifier is manually removed using terraform state rm
- This issue occurs not only with tables but also with schemas and databases
- It is likely that the problem affects not only database roles but account roles as well

### Steps to Reproduce

1. Create a Snowflake resource with correct privilege settings in Terraform (run apply successfully)
2. Change the privileges parameter of the same resource to an empty array []
3. Run terraform apply → Error occurs in Snowflake provider
4. Despite the error, the state file is overwritten with privileges=[]
5. Attempting to run terraform plan again results in errors due to inconsistency with Snowflake
6. To resume normal operations, you need to manually remove the resource identifier using terraform state rm

### How much impact is this issue causing?

Low

### Additional Information

The issue has been confirmed with Terraform version 2.0.0 as well, where configurations with empty privileges arrays still successfully display in plan output. Therefore, it's likely that in Terraform version 2.0.0 and later versions, the same situation would occur - plans display normally, but applying results in Snowflake provider errors and state inconsistency.

error message
```console
│ Error: Failed to parse internal identifier
│ 
[...]

PR #3695, opened 2025-05-21
fix: Empty privileges

sfc-gh-jcieslak:
## Changes
- Adjust privilege-granting resources to prevent empty privilege lists from being specified.
- Added test on `snowflake_grant_privileges_to_database_role` to ensure it's fixed (I had to add another file with empty privileges explicitly defined, because otherwise with `config.ListVariables()` empty it was treating it as "Terraform null", which triggered `ExactlyOneOf { "privileges", "all_privileges" }` validation).

Reference: #3690


MIGRATION_GUIDE.md
```diff
@@ -18,6 +18,32 @@ across different versions.
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.2.0 ➞ v2.2.1
```
sfc-gh-jmichalak, line 21: 2.2.0 hasn't been released yet.

MIGRATION_GUIDE.md
```diff
@@ -18,6 +18,32 @@ across different versions.
...
+
+### *(bugfix)* snowflake_grant_privileges_to_account_role, snowflake_grant_privileges_to_database_role, and snowflake_grant_privileges_to_share resources
```
sfc-gh-jmichalak, line 23: It's not clear from the header what is fixed.

pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -34,6 +34,7 @@ var grantPrivilegesToAccountRoleSchema = map[string]*schema.Schema{
 		Type:        schema.TypeSet,
 		Optional:    true,
 		Description: "The privileges to grant on the account role. This field is case-sensitive; use only upper-case privileges.",
+		MinItems:    1,
```
sfc-gh-jmichalak, line 37: Question: This will come back during the implementation of authoritative flags. I guess it would be done with a separate field + altered `ExactlyOneOf`?

MIGRATION_GUIDE.md
```diff
@@ -18,6 +18,32 @@ across different versions.
...
+
+Previously, it was possible to create `snowflake_grant_privileges_to_X` resources with an empty privilege list, which led to the following error:
+```
+│ Error: Failed to parse internal identifier
+│ Error: [grant_privileges_to_database_role_identifier.go:79] invalid Privileges value: , should be either a comma separated list of privileges or "ALL" / "ALL PRIVILEGES" for all privileges
+```
+After that, an identifier stored in state would be corrupted and only manual state manipulation would be able to fix it.
```
sfc-gh-jmichalak, line 30: Nit: 
```suggestion
After that, an identifier stored in state would be corrupted and only manual state manipulation would fix it.
```

MIGRATION_GUIDE.md
```diff
@@ -18,6 +18,32 @@ across different versions.
...
+We added validation to prevent this from happening. Now, if you try to create a resource with an empty privilege list, you will get the following error:
```
sfc-gh-jmichalak, line 31: Nit: 
```suggestion
We added validation to prevent this from happening. Now, if you try to create or update a resource with an empty privilege list, you will get the following error:
```

sfc-gh-jmichalak (changes requested):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Done

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Done

MIGRATION_GUIDE.md
```diff
@@ -20,6 +20,30 @@ across different versions.
...
 ## v2.1.0 ➞ v2.2.0
 
+### *(bugfix)* Fix the behavior for empty privileges list in snowflake_grant_privileges_to_account_role, snowflake_grant_privileges_to_database_role, and snowflake_grant_privileges_to_share resources
+
+Previously, it was possible to create `snowflake_grant_privileges_to_X` resources with an empty privilege list, which led to the following error:
+```
+│ Error: Failed to parse internal identifier
+│ Error: [grant_privileges_to_database_role_identifier.go:79] invalid Privileges value: , should be either a comma separated list of privileges or "ALL" / "ALL PRIVILEGES" for all privileges
+```
+After that, an identifier stored in state would be corrupted and only manual state manipulation would fix it.
```
sfc-gh-asawicki, line 30: what needs to be done on the user side? Is bumping to the newer version, having the corrupted state, enough? If not, please describe what needs to be done, if yes, please also be explicit about it.

pkg/testacc/grant_privileges_to_database_role_acceptance_test.go
```diff
@@ -1472,3 +1472,69 @@ func TestAcc_GrantPrivileges_OnObject_HybridTable_ToDatabaseRole_Fails(t *testin
 		},
 	})
 }
+
+// proves that https://github.com/snowflakedb/terraform-provider-snowflake/issues/3690 is fixed
+func TestAcc_GrantPrivileges_ToDatabaseRole_WithEmptyPrivileges(t *testing.T) {
```
sfc-gh-asawicki, line 1477: In context of my comment in the migration guide and the previous discussion we had on tests on the previous version tr just tests in any version: this is the example of a bug, I would like to see tested on the previous versus current version, as it can also then document, if you can easily get out of the problem by simply bumping the version.

sfc-gh-asawicki (changes requested):

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jcieslak: Ah, ok. Yeah, maybe in that case we could have another field dedicated to empty privileges, but as you mentioned, this is an implementation detail we can discuss later.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Added a note on how to fix the corrupted state.

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

PR #3705, opened 2025-05-26
feat: Add git repository definition

sfc-gh-ppreibisch:
Generate SDK code for git repository.


pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,155 @@
...
+).AlterOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/alter-git-repository",
+	g.NewQueryStruct("AlterGitRepository").
+		Alter().
+		SQL("GIT REPOSITORY").
+		IfExists().
+		Name().
+		OptionalQueryStructField(
+			"Set",
+			g.NewQueryStruct("GitRepositorySet").
```
sfc-gh-jmichalak, line 64: Question: does `SET ORIGIN` work?

pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,155 @@
...
+	Text("Name").
+	Text("DatabaseName").
+	Text("SchemaName").
+	Text("Origin").
+	Text("ApiIntegration").
+	Text("GitCredentials").
+	Text("Owner").
+	Text("OwnerRoleType").
+	Text("Comment").
+	Time("LastFetchedAt")
```
sfc-gh-jmichalak, line 31: Let's make sure this is present in the docs (for show and desc ops).

pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,155 @@
...
+).ShowOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/show-git-repositories",
+	gitRepositoryDbRow,
+	gitRepository,
+	g.NewQueryStruct("ShowGitRepositories").
+		Show().
+		SQL("GIT REPOSITORIES").
+		OptionalLike().
+		OptionalIn().
+		OptionalLimit(),
```
sfc-gh-jmichalak, line 109: Does it work? It's not present in the docs.

pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,155 @@
...
+).ShowByIdOperationWithFiltering(
+	g.ShowByIDLikeFiltering,
+	g.ShowByIDInFiltering,
+).CustomShowOperation(
+	"ShowGitBranches",
+	"https://docs.snowflake.com/en/sql-reference/sql/show-git-branches",
+	g.DbStruct("gitBranchesRow").
```
sfc-gh-jmichalak, line 116: Missing schema in https://docs.snowflake.com/en/sql-reference/sql/show-git-branches. It would be great to add there something similar to https://docs.snowflake.com/en/sql-reference/sql/show-git-repositories#output. Pls add this to docs-related action points.

pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,155 @@
...
+		Text("Name").
+		Text("Path").
+		Text("Checkouts").
+		Text("CommitHash"),
+	g.NewQueryStruct("ShowGitBranches").
+		Show().
+		SQL("GIT BRANCHES").
+		OptionalLike().
+		SQL("IN").
+		OptionalSQL("GIT REPOSITORY").
```
sfc-gh-jmichalak, line 131: Does setting this make any difference? Similarly, in show git tags.

pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,155 @@
...
+		Text("checkouts").
+		Text("commit_hash"),
+	g.PlainStruct("GitBranch").
+		Text("Name").
+		Text("Path").
+		Text("Checkouts").
+		Text("CommitHash"),
+	g.NewQueryStruct("ShowGitBranches").
+		Show().
+		SQL("GIT BRANCHES").
```
sfc-gh-jmichalak, line 128: IMO we can just use `SQL("SHOW GIT BRANCHES")` and remove `Show()` above. This way, we'll have a single field in the generated structs. Similarly for show git tags.

pkg/sdk/git_repository_impl_gen.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+func (r gitRepositoriesRow) convert() *GitRepository {
+	return &GitRepository{
+		CreatedOn:      r.CreatedOn,
+		Name:           r.Name,
+		DatabaseName:   r.DatabaseName,
+		SchemaName:     r.SchemaName,
+		Origin:         r.Origin,
+		ApiIntegration: r.ApiIntegration,
```
sfc-gh-jmichalak, line 151: After changing these fields to *AccountObjectIdentifier, this will need to be adjusted. You can use code like
```
	if r. ApiIntegration.Valid {
		id, err := ParseAccountObjectIdentifier(r. ApiIntegration.String)
		if err != nil {
			log.Printf("[DEBUG] failed to parse API integration in git repository: %v", err)
		} else {
			g.ApiIntegration = &id
		}
	}
```

pkg/sdk/git_repository_gen_test.go
```diff
@@ -0,0 +1,302 @@
...
+		opts.Set = &GitRepositorySet{
+			ApiIntegration: &apiIntegrationId,
+			GitCredentials: &gitCredentialsId,
+			Comment:        String("comment"),
+		}
+
+		assertOptsValidAndSQLEquals(t, opts, "ALTER GIT REPOSITORY %s SET API_INTEGRATION = %s GIT_CREDENTIALS = %s COMMENT = 'comment'", id.FullyQualifiedName(), apiIntegrationId.FullyQualifiedName(), gitCredentialsId.FullyQualifiedName())
+	})
+
+	t.Run("all options", func(t *testing.T) {
```
sfc-gh-jmichalak, line 112: This looks the same as the one above.

pkg/sdk/git_repository_gen_test.go
```diff
@@ -0,0 +1,302 @@
...
+			{
+				Name:  NewAccountObjectIdentifier("tag-name"),
+				Value: "tag-value",
+			},
+		}
+		opts.SetTags = tag
+		assertOptsValidAndSQLEquals(t, opts, "ALTER GIT REPOSITORY %s SET TAG \"tag-name\" = 'tag-value'", id.FullyQualifiedName())
+
+		opts.SetTags = nil
+		opts.UnsetTags = []ObjectIdentifier{
```
sfc-gh-jmichalak, line 144: Pls split such checks to separate `t.Run`s and remove `... = nil`.

sfc-gh-jmichalak (changes requested):
Generally, it looks OK. I asked a few questions and noted minor issues. Pls also adjust the description. This PR also adds a custom show operation, so we should mention it.

pkg/sdk/git_repository_gen_test.go
sfc-gh-ppreibisch: Yeah, I will remove that

pkg/sdk/git_repository_def.go
sfc-gh-ppreibisch: As I mentioned [here](https://github.com/snowflakedb/terraform-provider-snowflake/pull/3705#discussion_r2109138221). I will let our technical writer know about that later.

pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,155 @@
...
+	Text("SchemaName").
+	Text("Origin").
+	Text("ApiIntegration").
+	Text("GitCredentials").
+	Text("Owner").
+	Text("OwnerRoleType").
+	Text("Comment").
+	Time("LastFetchedAt")
+
+var apiIntegrationIdentifierOptions = g.IdentifierOptions().SQL("API_INTEGRATION =")
```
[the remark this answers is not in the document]
sfc-gh-ppreibisch: I will change that, thanks!

pkg/sdk/git_repository_gen_test.go
```diff
@@ -0,0 +1,302 @@
...
+		opts.Origin = origin
+		opts.GitCredentials = &gitCredentialsId
+		opts.Comment = String("comment")
+		opts.Tag = []TagAssociation{
+			{
+				Name:  NewAccountObjectIdentifier("tag-name"),
+				Value: "tag-value",
+			},
+		}
+		assertOptsValidAndSQLEquals(t, opts, "CREATE GIT REPOSITORY IF NOT EXISTS %s ORIGIN = '%s' API_INTEGRATION = %s GIT_CREDENTIALS = %s COMMENT = '%s' TAG (\"tag-name\" = 'tag-value')", id.FullyQualifiedName(), origin, apiIntegrationId.FullyQualifiedName(), gitCredentialsId.FullyQualifiedName(), "comment")
```
sfc-gh-asawicki, line 60: nit: use ` instead of " as you don't need to escape the double quotes then

pkg/sdk/git_repository_gen_test.go
```diff
@@ -0,0 +1,302 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
+	})
+
+	t.Run("validation: valid identifier for [opts.name]", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.name = emptySchemaObjectIdentifier
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("validation: exactly one field from [opts.Set opts.Unset opts.SetTags opts.UnsetTags opts.Fetch] should be present", func(t *testing.T) {
```
sfc-gh-asawicki, line 85: please test also the case when no option is selected

pkg/sdk/git_repository_gen_test.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+
+		opts.Unset = &GitRepositoryUnset{
+			Comment: Bool(true),
+		}
+
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterGitRepositoryOptions", "Set", "Unset", "SetTags", "UnsetTags", "Fetch"))
+	})
+
+	t.Run("basic", func(t *testing.T) {
```
sfc-gh-asawicki, line 98: nit: for alter test we usually name them "set", "unset", etc.

pkg/sdk/git_repository_gen_test.go
sfc-gh-asawicki, line 112: Let's split this test into multiple alter branches (easier to read, now it ends at the first fail)

pkg/sdk/git_repository_validations_gen.go
```diff
@@ -0,0 +1,85 @@
...
+	_ validatable = new(ShowGitBranchesGitRepositoryOptions)
+	_ validatable = new(ShowGitTagsGitRepositoryOptions)
+)
+
+func (opts *CreateGitRepositoryOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	if !ValidObjectIdentifier(opts.name) {
```
sfc-gh-asawicki, line 18: we could also have validations for othe identifiers used in create statement

pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,155 @@
...
+	Time("last_fetched_at")
+
+var gitRepository = g.PlainStruct("GitRepository").
+	Time("CreatedOn").
+	Text("Name").
+	Text("DatabaseName").
+	Text("SchemaName").
+	Text("Origin").
+	Text("ApiIntegration").
+	Text("GitCredentials").
```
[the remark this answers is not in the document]
sfc-gh-ppreibisch: Addressed

pkg/sdk/git_repository_def.go
```diff
@@ -0,0 +1,153 @@
...
+).CreateOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/create-git-repository",
+	g.NewQueryStruct("CreateGitRepository").
+		Create().
+		OrReplace().
+		SQL("GIT REPOSITORY").
+		IfNotExists().
+		Name().
+		TextAssignment("ORIGIN", g.ParameterOptions().SingleQuotes()).
+		Identifier("ApiIntegration", g.KindOfT[AccountObjectIdentifier](), apiIntegrationIdentifierOptions.Required()).
```
sfc-gh-jcieslak: I think `apiIntegrationIdentifierOptions` should be inlined with `gitCredentialsIdentifierOptions`. The required method you call on it here has a pointer receiver, meaning, it will modify the calling object to be required (even in places you'll define it without the Required method, like in lines 65/66).

pkg/sdk/git_repository_validations_gen.go
sfc-gh-jcieslak: note: there's `g.ValidIdentifierIfSet` that can be used in `_def.go`

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #3711, opened 2025-05-29
feat: Add embedding model support to Cortex Search Service

### LINKED ISSUE #3567: [Feature]: snowflake cortex search resource is missing the EMBEDDING_MODEL configuration
### Company Name

_No response_

in snowflake you would create the cortex service like this:

```
CREATE OR REPLACE CORTEX SEARCH SERVICE transcript_search_service
  ON transcript_text
  ATTRIBUTES region
  WAREHOUSE = cortex_search_wh
  TARGET_LAG = '1 day'
  EMBEDDING_MODEL = 'snowflake-arctic-embed-l-v2.0'
  AS (
    SELECT
        transcript_text,
        region,
        agent_id
    FROM support_transcripts
);
```

the above SQL lets you specify EMBEDDING_MODEL = 'snowflake-arctic-embed-l-v2.0' which the current terraform resource does not have an attribute for EMBEDDING_MODEL but rather uses the default `snowflake-arctic-embed-m-v1.5` with is not  multilingual.

### Category

category:resource

### Object type(s)

resource:cortex_search_service

### Proposal

resource "snowflake_cortex_search_service" "test" {
  depends_on = [snowflake_table.test]

  database   = snowflake_database.test.name
  schema     = snowflake_schema.test.name
  name       = "some_name"
  on         = "SOME_TEXT"
  target_lag = "2 minutes"
  warehouse  = "some_warehouse"
  query      = "SELECT SOME_TEXT FROM \"some_database\".\"some_schema\".\"some_table\""
  comment    = "some comment"
 **embedding_model = "snowflake-arctic-embed-l-v2.0" [ADD THIS]**
}

High

### Additional Information

_No response_

- [ ] Yeah, I'll take it 😎

kang8:
- Introduced a new optional parameter `embedding_model` in the Cortex Search Service schema.
- Updated the service creation request to handle the embedding model.
- Enhanced acceptance tests to validate the embedding model functionality.
- Added integration tests to ensure correct behavior when creating services with the embedding model.

## Test Plan
* [x] acceptance tests

## References
 #3567


pkg/sdk/testint/cortex_search_services_integration_test.go
```diff
@@ -72,6 +72,28 @@ func TestInt_CortexSearchServices(t *testing.T) {
 		require.Equal(t, comment, cortexSearchServiceById.Comment)
 	})
 
+	t.Run("create: with embedding model", func(t *testing.T) {
+		table, tableCleanup := testClientHelper().Table.CreateWithPredefinedColumns(t)
+		t.Cleanup(tableCleanup)
+
+		name := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+		embeddingModel := "snowflake-arctic-embed-m-v1.5"
+		err := client.CortexSearchServices.Create(ctx, sdk.NewCreateCortexSearchServiceRequest(name, on, testClientHelper().Ids.WarehouseId(), targetLag, buildQuery(table.ID())).WithEmbeddingModel(embeddingModel))
```
sfc-gh-jmichalak, line 81: We can just 
- Add `WithEmbeddingModel(embeddingModel)` to the `create: test complete` test.
- Add an assertion. According to the docs, this column is returned from Describe. So, we also need to add this column in the `def` file in DescribeOperation's both structs, and ensure it's present in the `convert` function in the `impl_gen.go` file (I added separate comments).
- Remove this test case.

pkg/sdk/cortex_search_services_def.go
```diff
@@ -36,6 +36,7 @@ var CortexSearchServiceDef = g.NewInterface(
 		).
 		Identifier("Warehouse", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().Equals().Required().SQL("WAREHOUSE")).
 		TextAssignment("TARGET_LAG", g.ParameterOptions().SingleQuotes().Required()).
+		OptionalTextAssignment("EMBEDDING_MODEL", g.ParameterOptions().SingleQuotes()).
 		OptionalComment().
 		PredefinedQueryStructField("QueryDefinition", "string", g.ParameterOptions().NoEquals().NoQuotes().Required().SQL("AS")).
 		// Validations
```
sfc-gh-jmichalak, line 42: Pls add `Text("embedding_model")` here. If it's nullable, use `OptionalText`.

sfc-gh-jmichalak, line 42: <p class='graphite__hidden'><i>[Re: line +120]</i></p>
Pls add `Text("EmbeddingModel")` here.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3711?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/sdk/cortex_search_services_impl_gen.go
```diff
@@ -77,6 +77,7 @@ func (r *CreateCortexSearchServiceRequest) toOpts() *CreateCortexSearchServiceOp
 
 		Warehouse:       r.Warehouse,
 		TargetLag:       r.TargetLag,
+		EmbeddingModel:  r.EmbeddingModel,
 		Comment:         r.Comment,
 		QueryDefinition: r.QueryDefinition,
 	}
```
sfc-gh-jmichalak, line 83: <p class='graphite__hidden'><i>[Re: line +152]</i></p>
Pls add converting EmbeddingModel in this function.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3711?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/resources/cortex_search_service_acceptance_test.go
```diff
@@ -108,3 +108,58 @@ func TestAcc_CortexSearchService_basic(t *testing.T) {
 		},
 	})
 }
+
+func TestAcc_CortexSearchService_withEmbeddingModel(t *testing.T) {
```
sfc-gh-jmichalak, line 112: No need to add a separate case. We can set this option in `TestAcc_CortexSearchService_basic` (`variableSet2` struct).

pkg/resources/cortex_search_service.go
```diff
@@ -167,6 +173,9 @@ func CreateCortexSearchService(ctx context.Context, d *schema.ResourceData, meta
 	if v, ok := d.GetOk("comment"); ok {
```
sfc-gh-jmichalak, line 173: <p class='graphite__hidden'><i>[Re: line +153]</i></p>
Please also set the `embedding_model`'s value in this function. This is a bit tricky because this field is optional and has a default value in Snowflake. We described this in another community PR: https://github.com/snowflakedb/terraform-provider-snowflake/pull/3659#discussion_r2092695567.

Let us know if you want to implement this. We can continue on our side, as it's a bit more complex.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3711?utm_source=unchanged-line-comment">Graphite</a>.</p>

sfc-gh-jmichalak (changes requested):
Thanks for the PR 👍 I added a few smaller comments and one bigger. Please let us know if you need any help and if you want to continue.

pkg/resources/cortex_search_service.go
kang8: I used `DESCRIBE` instead of `SHOW`, so I can get the information of the embedding_model. For more information, see 2682912f.

kang8:
Hi @sfc-gh-jmichalak, thanks again for your earlier suggestions! I've made some updates based on our discussion. When you have a moment, I'd appreciate your thoughts. Thanks!

pkg/sdk/cortex_search_services_gen.go
```diff
@@ -101,6 +102,8 @@ type cortexSearchServiceDetailsRow struct {
 	SourceDataNumRows int            `db:"source_data_num_rows"`
 	IndexingState     string         `db:"indexing_state"`
 	IndexingError     sql.NullString `db:"indexing_error"`
+	ServingState      string         `db:"serving_state"`
```
sfc-gh-jmichalak: Did you add ServingState here and below manually? I don't see relevant changes in the `def` file. We can remove this for now. If you want to add this field, please also 
- Add it to the `def` file,
- Add handling in the `impl` file (converter function),
- Add an assertion to one of the integration tests, so that we know it's being filled correctly.

pkg/resources/cortex_search_service.go
sfc-gh-jmichalak: Regarding the default values in Snowflake, it's true. The Terraform SDK has its own limitations - that's why we have a custom handling of default values with the `show_output` and `describe_output` computed fields (see our [doc](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/v1-preparations/CHANGES_BEFORE_V1.md)). 

In this case, the default value of the embedding model is non-empty, and we need to handle diffs properly when this value is not set in the configuration. This causes the acceptance test to fail. Please see the link above for more details. In short:
- Add `describe_output` to the schema 
- Inside the Read operation, we have to check for withExternalChangesMarking and only apply changes there. 
- Call setStateToValuesFromConfig with `embedding_model` 
- Set the `describe_output` in the Read operation.

As I mentioned earlier, this is a bit more complex, and we appreciate your efforts! Let us know if it's clearer now. We can also handle this on our side.

sfc-gh-jmichalak (changes requested):

pkg/sdk/cortex_search_services_gen.go
kang8: I've removed the ServingState changes in [dbdd7b5b](https://github.com/snowflakedb/terraform-provider-snowflake/pull/3711/commits/dbdd7b5bb670a13903880fffe7d77ee18fedf3ad) to keep our focus on the embedding_model implementation.

And I'm happy to address ServingState support in a separate follow-up PR for better scope management.

pkg/resources/cortex_search_service.go
kang8: Thank you for the suggestion!

I've attempted an implementation based on my current understanding and pushed the changes in 65b41b56. Please review it, I'm happy to make further adjustments based on your feedback.

pkg/resources/testdata/TestAcc_CortexSearchService_basic/2/variables.tf
```diff
@@ -35,3 +35,7 @@ variable "comment" {
 variable "table_name" {
```
sfc-gh-jmichalak: <p class='graphite__hidden'><i>[Re: line +2]</i></p>
Nit: Please remove these empty lines.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3711?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/resources/cortex_search_service.go
```diff
@@ -107,46 +121,106 @@ func CortexSearchService() *schema.Resource {
...
+					outputMapping{"embedding_model", "embedding_model", configEmbeddingModel, *cortexSearchServiceDetails.EmbeddingModel, nil},
+				); err != nil {
+					return diag.FromErr(err)
+				}
+			}
+		}
+
+		if err = setStateToValuesFromConfig(d, cortexSearchServiceSchema, []string{
+			"target_lag",
+			"warehouse",
```
sfc-gh-jmichalak: There's no need to include `target_lag` and `warehouse` in here. They are required fields. Similarly, the `comment` field, as the default value is an empty comment (it's already handled with `d.Set`). The only thing that must be done in this function:
- unchanged handling of other fields
- get `cortexSearchServiceDetails`
- add the if block with `withExternalChangesMarking`, `setStateToValuesFromConfig`, and setting DESC output - only for embedding_model, like
```
			var embeddingModel string
			if cortexSearchServiceDetails.EmbeddingModel != nil {
				embeddingModel = *cortexSearchServiceDetails.EmbeddingModel
			}
					outputMapping{"embedding_model", "embedding_model", embeddingModel, embeddingModel, nil},
			); err != nil {
				return diag.FromErr(err)
			}
		}
		if err = setStateToValuesFromConfig(d, cortexSearchServiceSchema, []string{
			"embedding_model",
		}); err != nil {
			return diag.FromErr(err)
		}
		errs := errors.Join(
			d.Set(DescribeOutputAttributeName, []map[string]any{schemas.CortexSearchServiceDetailsToSchema(*cortexSearchServiceDetails)}),
			d.Set(FullyQualifiedNameAttributeName, id.FullyQualifiedName()),
			d.Set("comment", cortexSearchService.Comment),
		)
		if errs != nil {
			return diag.FromErr(errs)
		}
```
You can check https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/resources/compute_pool.go#L220 for inspiration.

pkg/resources/cortex_search_service.go
```diff
@@ -107,46 +121,106 @@ func CortexSearchService() *schema.Resource {
...
+				outputMapping{"target_lag", "target_lag", d.Get("target_lag"), cortexSearchServiceDetails.TargetLag, nil},
+				outputMapping{"warehouse", "warehouse", d.Get("warehouse"), cortexSearchServiceDetails.Warehouse, nil},
+			); err != nil {
+				return diag.FromErr(err)
+			}
+
+			if cortexSearchServiceDetails.Comment != nil {
+				configComment := d.Get("comment").(string)
+				if err = handleExternalChangesToObjectInFlatDescribe(d,
+					outputMapping{"comment", "comment", configComment, *cortexSearchServiceDetails.Comment, nil},
```
sfc-gh-jcieslak: By calling `*cortexSearchServiceDetails.Comment` you may cause panic if comment would be `nil`. It's better to do something similar to this:
```go
var comment string
if cortexSearchServiceDetails.Comment != nil {
 comment = *cortexSearchServiceDetails.Comment
}
/// ...
	outputMapping{"comment", "comment", configComment, comment, nil},
/// ...
```

same with embedding model below

pkg/resources/cortex_search_service.go
```diff
@@ -107,46 +121,106 @@ func CortexSearchService() *schema.Resource {
...
+		if err = setStateToValuesFromConfig(d, cortexSearchServiceSchema, []string{
+			"target_lag",
+			"warehouse",
+			"comment",
+			"embedding_model",
+		}); err != nil {
+			return diag.FromErr(err)
+		}
+
+		if cortexSearchServiceDetails.EmbeddingModel != nil {
```
sfc-gh-jcieslak: This is not necessary as embedding_model would be set in the 192line (by the handleExternalChangesToObjectInFlatDescribe function)

pkg/resources/cortex_search_service.go
sfc-gh-jmichalak: Thanks, appreciate it :) We're very close, I added one tiny comment and one comment with more explanation of handling the optional field. Let us know if it makes sense.

I also noticed that `target_lag` and `warehouse` may not be handled properly, but we can leave it for now and focus only on the `embedding_model`.

pkg/resources/cortex_search_service.go
```diff
@@ -107,46 +121,106 @@ func CortexSearchService() *schema.Resource {
 	}
```
sfc-gh-jmichalak, line 126: <p class='graphite__hidden'><i>[Re: line +109]</i></p>
Please add
```
		CustomizeDiff: TrackingCustomDiffWrapper(resources.CortexSearchService,
			ComputedIfAnyAttributeChanged(cortexSearchServiceSchema, DescribeOutputAttributeName, "embedding_model"),
		),
```
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3711?utm_source=unchanged-line-comment">Graphite</a>.</p>

sfc-gh-jmichalak (changes requested):

kang8:
Thank you for the detailed review, I learned a lot from this process.

Since it took a long time and there were too many commits, plus the re-targeting of the branch to dev, I squashed the commits into one to facilitate the review.

kang8:
Additionally, I set the default value for embedding_model to [snowflake-arctic-embed-m-v1.5](https://docs.snowflake.com/en/sql-reference/sql/create-cortex-search#:~:text=If%20the%20EMBEDDING_MODEL%20is%20not%20specified%2C%20the%20default%20model%20is%20used.%20The%20default%20model%20is%20snowflake%2Darctic%2Dembed%2Dm%2Dv1.5.). I found this to be the simplest method, although it may face issues with subsequent default value changes.

pkg/resources/cortex_search_service.go
```diff
@@ -94,9 +109,13 @@ func CortexSearchService() *schema.Resource {
 		UpdateContext: PreviewFeatureUpdateContextWrapper(string(previewfeatures.CortexSearchServiceResource), TrackingUpdateWrapper(resources.CortexSearchService, UpdateCortexSearchService)),
 		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.CortexSearchServiceResource), TrackingDeleteWrapper(resources.CortexSearchService, deleteFunc)),
 
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.CortexSearchService,
+			ComputedIfAnyAttributeChanged(cortexSearchServiceSchema, DescribeOutputAttributeName, "embedding_model"),
```
sfc-gh-jcieslak, line 113: nit: there should be more fields listed

pkg/resources/cortex_search_service.go
```diff
@@ -60,6 +60,13 @@ var cortexSearchServiceSchema = map[string]*schema.Schema{
 		Required:    true,
 		Description: "Specifies the maximum target lag time for the Cortex search service.",
 	},
+	"embedding_model": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ForceNew:         true,
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInDescribe("embedding_model"),
+		Description:      "Specifies the embedding model to use for the Cortex search service.",
```
sfc-gh-jcieslak, line 68: nit: We should list of available values (or link to the documentation page where they're listed).

pkg/resources/cortex_search_service.go
```diff
@@ -107,46 +126,107 @@ func CortexSearchService() *schema.Resource {
...
+						Severity: diag.Warning,
+						Summary:  "Failed to query cortex search service. Marking the resource as removed.",
+						Detail:   fmt.Sprintf("Cortex search service id: %s, Err: %s", id.FullyQualifiedName(), err),
+					},
+				}
+			}
+			return diag.FromErr(err)
+		}
+
+		if err := d.Set("name", cortexSearchService.Name); err != nil {
```
sfc-gh-jcieslak, line 176: nit: we should remove name, database, schema from read.

pkg/sdk/testint/cortex_search_services_integration_test.go
```diff
@@ -49,27 +49,39 @@ func TestInt_CortexSearchServices(t *testing.T) {
 
 		name := testClientHelper().Ids.RandomSchemaObjectIdentifier()
 		comment := random.Comment()
-		err := client.CortexSearchServices.Create(ctx, sdk.NewCreateCortexSearchServiceRequest(name, on, testClientHelper().Ids.WarehouseId(), targetLag, buildQuery(table.ID())).WithOrReplace(true).WithComment(comment))
+		embeddingModel := "snowflake-arctic-embed-m-v1.5"
+		err := client.CortexSearchServices.Create(ctx, sdk.NewCreateCortexSearchServiceRequest(name, on, testClientHelper().Ids.WarehouseId(), targetLag, buildQuery(table.ID())).
+			WithOrReplace(true).
```
sfc-gh-jcieslak, line 54: nit: orReplace shouldn't be necessary here, can be removed

sfc-gh-jcieslak:
Thank you for your contribution!
The pr looks good, and I'm merging it. The change will be available in the next provider release, which should happen this week.

PR #3715, opened 2025-05-29
chore: Adjust resource assertions

sfc-gh-jcieslak:
## Changes
- Include type into resource assertions and make assertions based on type
- Adjust `no value` checks
- Add `isEmpty` and `isPresent` checks
- Remove repetitive assertions from `_ext.go` files due to exact same duplications

## Next pr
- Generate run generator for all files (~7k lines)

### DIFF
--- a/pkg/acceptance/bettertestspoc/assert/resourceassert/gen/model.go
+++ b/pkg/acceptance/bettertestspoc/assert/resourceassert/gen/model.go
@@ -1,12 +1,12 @@
 package gen
 
 import (
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"os"
 	"slices"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/genhelpers"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 // TODO [SNOW-1501905]: extract to commons?
--- a/pkg/acceptance/bettertestspoc/assert/resourceassert/gen/templates/assertions.tmpl
+++ b/pkg/acceptance/bettertestspoc/assert/resourceassert/gen/templates/assertions.tmpl
@@ -56,5 +56,5 @@
             return {{ $assertVar }}
         }
     {{ end }}
-    {{/* TODO(SNOW-1501905): Implement custom asserttion for not-empty collections */}}
+    {{/* TODO(SNOW-1501905): Implement custom assertion for not-empty collections */}}
 {{ end -}}
--- a/pkg/acceptance/bettertestspoc/README.md
+++ b/pkg/acceptance/bettertestspoc/README.md
@@ -382,6 +382,7 @@ func (w *WarehouseDatasourceShowOutputAssert) IsEmpty() {
     w.assertions = append(w.assertions, valueSet("show_output.#", "0"))
 }
 ```
+- add an assertion method or helper method that would be better at checking for missing complex types (like Lists, Sets, Maps, etc.), currently ValueSet with `.#` is used
 - support other mappings if needed (TODO left in `assert/objectassert/gen/model.go`)
 - consider extracting preamble model to commons (TODOs left in `assert/objectassert/gen/model.go` and in `assert/objectparametersassert/gen/model.go`)
 - get a runtime name for the assertion creator (TODOs left in `assert/objectparametersassert/gen/model.go`)


pkg/acceptance/bettertestspoc/assert/resourceassert/resource_monitor_resource_gen.go
```diff
@@ -112,12 +112,12 @@ func (r *ResourceMonitorResourceAssert) HasNoFullyQualifiedName() *ResourceMonit
 }
 
 func (r *ResourceMonitorResourceAssert) HasNoNotifyTriggers() *ResourceMonitorResourceAssert {
-	r.AddAssertion(assert.ValueNotSet("notify_triggers"))
+	r.AddAssertion(assert.ValueSet("notify_triggers.#", "0"))
```
sfc-gh-jmichalak, line 115: Let's add generic descriptions to the generator's README. As I understand,
- `HasNoX` methods check that a field is not set (`assert.ValueNotSet`, or ValueSet 0 for lists/sets),
- `HasXEmpty` methods check that a field has an empty value (0, empty string, etc),
- `HasXNotEmpty` methods check that a field is set with any value.

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/templates/assertions.tmpl
```diff
@@ -15,14 +15,45 @@
...
+// Attribute no value checks //
+///////////////////////////////
+
+{{ range .Attributes -}}
+    func ({{ $assertVar }} *{{ $assertName }}) HasNo{{ SnakeCaseToCamel .Name }}() *{{ $assertName }} {
+    {{ if or (eq .AttributeType "TypeList") (eq .AttributeType "TypeSet") -}}
+        {{ $assertVar }}.AddAssertion(assert.ValueSet("{{ .Name }}.#", "0"))
+        return {{ $assertVar }}
+    {{ else -}}
+        {{ $assertVar }}.AddAssertion(assert.ValueNotSet("{{ .Name }}"))
```
sfc-gh-jmichalak, line 28: Nit: This does not concern this PR and could be done separately, but IMO the names are slightly confusing. I'd change `assert.ValueSet` to `assert.ValueEqual`, and  `assert.ValuePresent` to `assert.ValueSet`. Or, change `assert.ValueNotSet` to `assert.ValueNotPresent`. WDYT?

sfc-gh-jmichalak (approved):

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/templates/assertions.tmpl
sfc-gh-jcieslak: Well, I personally don't have the issue with naming, but maybe because I got used to them, but If I would have to change them, I would propose:
`ValueSet` -> `ValueEqual`
`ValuePresent` (the same)
`ValueNotSet` -> `ValueNotPresent`
I don't have any strong opinions on that, I'm fine with leaving them as they are or changing them to clearer ones.

sfc-gh-asawicki: I would leave them as name changing would require making it consistent between all the assertion types.

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/model.go
```diff
@@ -35,9 +36,9 @@ func ModelFromResourceSchemaDetails(resourceSchemaDetails genhelpers.ResourceSch
 			continue
 		}
 		attributes = append(attributes, ResourceAttributeAssertionModel{
-			Name: attr.Name,
-			// TODO [SNOW-1501905]: add attribute type logic; allow type safe assertions, not only strings
```
sfc-gh-asawicki, line 39: This TODO is not completed. You have only used the type in `No..` and `Empty` checks and the assertions are still not type-safe. Please bring it back and let's for the time-being add a `IsCollection` bool, so that this is not misleading.

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/templates/assertions.tmpl
```diff
@@ -15,13 +15,44 @@
...
+{{ end -}}
+
 ////////////////////////////
 // Attribute empty checks //
 ////////////////////////////
 
 {{ range .Attributes -}}
-    func ({{ $assertVar }} *{{ $assertName }}) HasNo{{ SnakeCaseToCamel .Name }}() *{{ $assertName }} {
-        {{ $assertVar }}.AddAssertion(assert.ValueNotSet("{{ .Name }}"))
+    {{ if and (eq .AttributeType "TypeString") (not .IsRequired) -}}
```
sfc-gh-asawicki, line 40: This logic is incorrect, Why we only care for strings? Lets add it for each which is not a collection (based on the suggested boolean).

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/templates/assertions.tmpl
```diff
@@ -15,13 +15,44 @@
 
 {{ end -}}
 
+///////////////////////////////
+// Attribute no value checks //
+///////////////////////////////
+
+{{ range .Attributes -}}
+    func ({{ $assertVar }} *{{ $assertName }}) HasNo{{ SnakeCaseToCamel .Name }}() *{{ $assertName }} {
+    {{ if or (eq .AttributeType "TypeList") (eq .AttributeType "TypeSet") -}}
```
sfc-gh-asawicki, line 24: We should not leak the tf types into the template (but it will be changed after one of the previous comments).

sfc-gh-asawicki (changes requested):

sfc-gh-jmichalak (dismissed):

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/model.go
```diff
@@ -1,6 +1,7 @@
 package gen
 
 import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
```
sfc-gh-asawicki: nit: this import should be lower

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/templates/assertions.tmpl
```diff
@@ -12,17 +12,49 @@
...
+///////////////////////////////
+
+{{ range .Attributes -}}
+    {{ if not .IsCollection }}
+        func ({{ $assertVar }} *{{ $assertName }}) Has{{ SnakeCaseToCamel .Name }}NotEmpty() *{{ $assertName }} {
+            {{ $assertVar }}.AddAssertion(assert.ValuePresent("{{ .Name }}"))
+            return {{ $assertVar }}
+        }
+    {{ end }}
+    {{/* TODO(SNOW-1501905): Implement custom asserttion for not-empty collections */}}
```
sfc-gh-asawicki: nit: typo in `asserttion`

pkg/acceptance/bettertestspoc/assert/resourceassert/resource_monitor_resource_gen.go
sfc-gh-asawicki, line 115: thought for later (please add it to the readme): instead of constructing this `#` check, we could have another type of assertion instead of ValueSet (maybe it's already even added some TODO in the readme) or even not a assertion but the helper method (like we have multiple methods for showoutput or parameters assertions)

sfc-gh-jcieslak: Added

sfc-gh-jmichalak (approved):

PR #3718, opened 2025-05-29
chore: Show output assertion adjustments

sfc-gh-jcieslak:
## Changes
- Move show-related assertions to a dedicated file
- Adjust assertions (introduce `NotSet` and `Present` assertions) and generate them with new functions (only some of them; the ones that required adjustments, e.g. in the _ext file; the rest that wasn't causing any issues will be pushed in the next pr)
- Remove now repetitive functions inside `_ext.go` files
- Adjust tests to the newly introduced assertions (mostly tests for streams and external volume)

## Next prs
- Add assertion to the template that would check for parameter presence
- Generate the rest of the assertions


sfc-gh-jmichalak (approved):
Approved after the errors are fixed.

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/task_show_output_gen.go
```diff
@@ -135,7 +135,127 @@ func (t *TaskShowOutputAssert) HasBudget(expected string) *TaskShowOutputAssert
 	return t
 }
 
+// Manually adjusted
+// func (t *TaskShowOutputAssert) HasTaskRelations(expected sdk.TaskRelations) *TaskShowOutputAssert {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Yes, I mentioned it in the pr desc, this is the only thing that still has to be commented out (I didn't investigate yet why)

pkg/testacc/data_source_streams_acceptance_test.go
```diff
@@ -63,7 +63,7 @@ func TestAcc_Streams(t *testing.T) {
 						HasSourceType(sdk.StreamSourceTypeTable).
 						HasBaseTables(table.ID()).
 						HasType("DELTA").
-						HasStale("false").
+						HasStale(false).
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, that's because I cut too much of the generated files :( added more generated files to the pr which should fix compilation errors (GoLand couldn't handle those files for some reason and I had to run pre-push everytime to check for errors)

sfc-gh-jmichalak (approved):

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/model.go
```diff
@@ -53,25 +53,27 @@ func MapToResourceShowOutputAssertion(field genhelpers.Field) ResourceShowOutput
 	var assertionCreator string
 	switch {
 	case concreteTypeWithoutPtr == "bool":
-		assertionCreator = "ResourceShowOutputBoolValueSet"
+		assertionCreator = "ResourceShowOutputBoolValue"
```
sfc-gh-asawicki, line 56: I'm not sure if extracting the `Set` part inside the template and having here incomplete creator names is a good idea. I would like it reverted.

pkg/internal/genhelpers/template_commons.go
```diff
@@ -34,6 +34,11 @@ func TypeWithoutPointerAndBrackets(t string) string {
 	return without
 }
 
+func IsTypeSlice(t string) bool {
```
sfc-gh-asawicki, line 37: nit: not sure if this should be a common but let's leave it here for now.

pkg/internal/genhelpers/template_commons.go
```diff
@@ -34,6 +34,11 @@ func TypeWithoutPointerAndBrackets(t string) string {
...
+	_, ok := strings.CutPrefix(t, "[]")
```
sfc-gh-asawicki: nit: Why `Cut` and not `Has`?

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/assertions.tmpl
```diff
@@ -3,14 +3,36 @@
...
 {{- $assertVar := FirstLetter $nameLowerCase }}
+
 ////////////////////////////
 // Attribute value checks //
 ////////////////////////////
 
 {{ range .Attributes -}}
-    func ({{ $assertVar }} *{{ $assertName }}) Has{{ .Name }}(expected {{ .ConcreteType }}) *{{ $assertName }} {
-        {{ $assertVar }}.AddAssertion(assert.{{ .AssertionCreator }}("{{ SnakeCase .Name }}", {{ RunMapper .Mapper "expected" }}))
+    {{ if not (IsTypeSlice .ConcreteType) -}}
```
sfc-gh-asawicki, line 12: nit: we handle this differently than we handle e.g. object assertions where `SnowflakeObjectFieldAssertion` has the `IsOriginalTypeSlice` etc. We should make it more consistent. Please add it to the README.

pkg/testacc/resource_external_volume_acceptance_test.go
```diff
@@ -1021,8 +1025,10 @@ func TestAcc_External_Volume_Azure(t *testing.T) {
 	resourceId := helpers.EncodeResourceIdentifier(id)
 	comment := random.Comment()
 	comment2 := random.Comment()
-	allowWritesTrue := "true"
-	allowWritesFalse := "false"
+	allowWritesTrue := true
```
sfc-gh-asawicki: I don't like this rename. And tbh, why we are using a dedicated holder for `"true"` and for `"false"`? We have the boolean string that should be just used here

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_external_volume_acceptance_test.go
sfc-gh-jcieslak: The whole file is a bit messy, tbh I would prefer to inline those boolean and string boolean values, but for now I just went with what was already there.

sfc-gh-asawicki: inlining is the same amount of changes, so let's just do it

pkg/internal/genhelpers/template_commons.go
sfc-gh-jcieslak: Changed to Cut (I believe this function was something else earlier and cut was left).

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/assertions.tmpl
sfc-gh-jcieslak: That's true, added.

pkg/testacc/resource_external_volume_acceptance_test.go
sfc-gh-jcieslak: Done

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/model.go
sfc-gh-jcieslak: How would you then assert for `xPresent` or `xNotSet`?

sfc-gh-jmichalak (approved):

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/model.go
sfc-gh-asawicki: ??

Now the name of the method is constructed by what you put here as `assertionCreator` and it's then added to the explicit `Set` in the template. So if I now try to find method `ResourceShowOutputBoolValue` I won't find anything. If you keep here `ResourceShowOutputBoolValueSet` like it was before, then it's findable because you have the whole method names here.

pkg/internal/genhelpers/template_commons.go
sfc-gh-asawicki: * changed to `Has` :)

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/model.go
sfc-gh-jcieslak: Yes, I did it this way, so that the assertions that are placed here: https://github.com/snowflakedb/terraform-provider-snowflake/pull/3718/files#diff-dafae03b2c679d375d6bc2b5fa4568dad3f7377cb201726a09ca018098e2832e with `xSet`/`xNotSet`/`xPresent` postfixes can be used without changing much. In the templates, there should also be a check for the presence (added to the next prs).

sfc-gh-asawicki: OKay, let's close it, we can discuss when I get back

PR #3721, opened 2025-05-29
chore: Add location identifier

sfc-gh-jmichalak:
- Add location identifier to SDK.
- Use location identifier in service's location on stages.


pkg/sdk/identifier_helpers.go
```diff
@@ -453,3 +453,30 @@ func (i TableColumnIdentifier) FullyQualifiedName() string {
...
+		path:  path,
+	}
+}
+
+func (i LocationIdentifier) Stage() SchemaObjectIdentifier {
+	return i.stage
+}
+
+func (i LocationIdentifier) Name() string {
+	return i.path
```
sfc-gh-jcieslak, line 474: Not sure if that should be path or stage FQN, is there any reason path is returned here and not the stage id?

pkg/sdk/identifier_helpers.go
```diff
@@ -453,3 +453,30 @@ func (i TableColumnIdentifier) FullyQualifiedName() string {
...
+}
+
+func (i LocationIdentifier) FullyQualifiedName() string {
+	if i.stage.FullyQualifiedName() == "" && i.path == "" {
+		return ""
+	}
+	return fmt.Sprintf(`@%v%v`, i.stage.FullyQualifiedName(), i.path)
```
sfc-gh-jcieslak, line 481: Do we want to validate or check whether there is a slash in path or not (or at least add resource-level validator)? The path should be either empty or non-empty starting with slash

pkg/sdk/identifier_helpers_test.go
```diff
@@ -82,3 +82,27 @@ func TestDatabaseObjectIdentifier(t *testing.T) {
...
+		identifier := NewLocationIdentifier(stage, "/path/to/file")
+
+		assert.Equal(t, `@"db"."schema"."stage"/path/to/file`, identifier.FullyQualifiedName())
+	})
+
+	t.Run("empty stage and path returns empty fully qualified name", func(t *testing.T) {
+		stage := NewSchemaObjectIdentifier("", "", "")
+		identifier := NewLocationIdentifier(stage, "")
+
+		assert.Equal(t, "", identifier.FullyQualifiedName())
```
sfc-gh-jcieslak, line 106: Shouldn't it print `@""."".""`?

pkg/sdk/identifier_helpers.go
```diff
@@ -453,3 +453,30 @@ func (i TableColumnIdentifier) FullyQualifiedName() string {
 func (i TableColumnIdentifier) SchemaObjectId() SchemaObjectIdentifier {
 	return NewSchemaObjectIdentifier(i.databaseName, i.schemaName, i.tableName)
 }
+
+type LocationIdentifier struct {
```
sfc-gh-jcieslak, line 457: Should we have a parser for it?

sfc-gh-jmichalak: It's not needed for now. We'll come back to this during parsing SHOW/DESC output, but services do not return stage locations. IIRC, we could do this in streamlits.

pkg/sdk/identifier_helpers_test.go
sfc-gh-jmichalak: This happens because the FQN is done by the schema's FQN. It turns out that this is a valid stage location, but it looks like we don't support such IDs in schemas, so I'd list this as a limitation for now.

pkg/sdk/identifier_helpers.go
sfc-gh-jmichalak: Not sure either, this method is for the purpose of satisfying an ObjectIdentifier interface. `Name` is semantically not even correct in this case. I'm not sure why stage's name would be more 'correct' here.

pkg/sdk/identifier_helpers.go
sfc-gh-jmichalak: Well, we can just expect a valid path. What if the user provides something like `asdf//asdf`, this is a valid filepath, but it fails in Snowflake. Potentially, we could just use `filepath.Clean`.

PR #3730, opened 2025-06-03
chore: Small improvements

sfc-gh-jcieslak:
## Changes
- Fix the test that was failing due to NPE because of the missing query tag from the Snowflake response
- Fix the dynamic table test by increasing the timeout value
- Fix the tests that are using `random.Email()` by adjusting the length of the email and its structure
- Add note on OpenTofu support to the FAQ
- Address in-code TODO
- Run pre-push (changes caused by previous prs)

### DIFF
--- a/pkg/acceptance/helpers/information_schema_client.go
+++ b/pkg/acceptance/helpers/information_schema_client.go
@@ -56,12 +56,15 @@ func (c *InformationSchemaClient) GetQueryHistoryByQueryId(t *testing.T, limit i
 
 func (c *InformationSchemaClient) mapQueryHistory(t *testing.T, queryResult map[string]*any) QueryHistory {
 	t.Helper()
-	require.NotNil(t, queryResult["QUERY_ID"])
-	require.NotNil(t, queryResult["QUERY_TEXT"])
-	require.NotNil(t, queryResult["QUERY_TAG"])
-	return QueryHistory{
-		QueryId:   (*queryResult["QUERY_ID"]).(string),
-		QueryText: (*queryResult["QUERY_TEXT"]).(string),
-		QueryTag:  (*queryResult["QUERY_TAG"]).(string),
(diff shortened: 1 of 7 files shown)


pkg/acceptance/helpers/information_schema_client.go
```diff
@@ -56,12 +56,15 @@ func (c *InformationSchemaClient) GetQueryHistoryByQueryId(t *testing.T, limit i
...
-	require.NotNil(t, queryResult["QUERY_ID"])
-	require.NotNil(t, queryResult["QUERY_TEXT"])
-	require.NotNil(t, queryResult["QUERY_TAG"])
-	return QueryHistory{
-		QueryId:   (*queryResult["QUERY_ID"]).(string),
-		QueryText: (*queryResult["QUERY_TEXT"]).(string),
-		QueryTag:  (*queryResult["QUERY_TAG"]).(string),
+	var queryHistory QueryHistory
+	if v, ok := queryResult["QUERY_ID"]; ok && v != nil {
+		queryHistory.QueryId = (*v).(string)
```
sfc-gh-jmichalak, line 61: Q: What if these are empty? Does this get asserted outside this function, so that the test fails?

sfc-gh-jmichalak (approved):

pkg/acceptance/helpers/information_schema_client.go
sfc-gh-jcieslak: Yes, the tests will fail. The improvement here is that the mapping won't panic if one of the parameters is not there, which made the whole test run end much faster than it was supposed to.

\+ The only optional parameter from query history that is mapped here is QUERY_TAG. The tests that use functions in this file may not find the desired query tag (depending on the testing environment traffic) and fail. In that case, the query limit (that is passed to the functions in this file), should be increased for the failing test.

PR #3734, opened 2025-06-04
chore: Introduce account parameters

sfc-gh-jcieslak:
## Changes
- Update account parameters to include all parameters from the specification (the ones that were excluded in the specification were left for backward compatibility; they were listed separately from the rest of the account parameters)
- Created a new structs for setting and unsetting parameters on account
- Added unit and integration tests
  - Note: Not all parameters in integration test were used due to setup they need
- Moved common parameter helpers to the separate file

## Next prs
- Restrict the ability to run account altering tests


pkg/sdk/common_types.go
```diff
@@ -388,9 +388,10 @@ var AllLogLevels = []LogLevel{
 type TraceLevel string
 
 const (
-	TraceLevelAlways  TraceLevel = "ALWAYS"
-	TraceLevelOnEvent TraceLevel = "ON_EVENT"
-	TraceLevelOff     TraceLevel = "OFF"
+	TraceLevelAlways    TraceLevel = "ALWAYS"
+	TraceLevelOnEvent   TraceLevel = "ON_EVENT"
```
sfc-gh-jcieslak: It's strange that in https://docs.snowflake.com/en/sql-reference/parameters#trace-level it says ON_EVENT is valid option (and I remember it is), but when I wanted to set it on account, it returned an error and said it's not a valid option, and listed all possible values, and it seems that ON_EVENT was replaced by PROPAGATE. Have to confirm, but ON_EVENT doesn't work on account.

pkg/sdk/assertions_test.go
```diff
@@ -53,7 +53,8 @@ func assertSQLEquals(t *testing.T, opts any, format string, args ...any) {
 	t.Helper()
 	actual, err := structToSQL(opts)
 	require.NoError(t, err)
-	assert.Equal(t, fmt.Sprintf(format, args...), actual)
+	ex := fmt.Sprintf(format, args...)
+	assert.Equal(t, ex, actual)
```
sfc-gh-jmichalak: ?

pkg/sdk/common_types.go
sfc-gh-jmichalak: I'm fine with this, but let's mention it in our migration guide.

pkg/sdk/parameters.go
```diff
@@ -200,11 +200,11 @@ func (parameters *parameters) SetAccountParameter(ctx context.Context, parameter
 }
 
 // TODO(SNOW-1866453): add integration tests
-func (parameters *parameters) UnsetAccountParameter(ctx context.Context, parameter AccountParameter) error {
+func (v *parameters) UnsetAccountParameter(ctx context.Context, parameter AccountParameter) error {
```
sfc-gh-jmichalak, line 203: Nit: 
```suggestion
func (p *parameters) UnsetAccountParameter(ctx context.Context, parameter AccountParameter) error {
```
In other places as well.

pkg/sdk/parameters_helpers.go
```diff
@@ -0,0 +1,32 @@
+package sdk
+
+import (
+	"fmt"
+	"strconv"
+)
+
+func parseBooleanParameter(parameter, value string) (_ *bool, err error) {
```
sfc-gh-jmichalak, line 8: 1. Why return a pointer?
2. IMO, this function can be inlined - simply call strconv.ParseBool in `setBooleanValue`, just like in `setIntegerValue`.

pkg/sdk/parameters_helpers.go
```diff
@@ -0,0 +1,32 @@
...
+	b, err := strconv.ParseBool(value)
+	if err != nil {
+		return nil, fmt.Errorf("boolean value (\"true\"/\"false\") expected for %v parameter, but got %v instead", parameter, value)
+	}
+	return &b, nil
+}
+
+func setBooleanValue[P AccountParameter | SessionParameter | UserParameter | TaskParameter | ObjectParameter](parameter P, value string, setField **bool) error {
```
sfc-gh-jmichalak: Can we create a type like 
```
type Parameter interface {
	AccountParameter | SessionParameter | UserParameter | TaskParameter | ObjectParameter
}
```
and reuse it?

pkg/sdk/accounts_test.go
```diff
@@ -122,11 +122,150 @@ func TestAccountAlter(t *testing.T) {
...
+					MetricLevel:                                      Pointer(MetricLevelAll),
+					MinDataRetentionTimeInDays:                       Int(1),
+					MultiStatementCount:                              Int(0),
+					NetworkPolicy:                                    Pointer(networkPolicyId),
+					NoorderSequenceAsDefault:                         Bool(false),
+					OAuthAddPrivilegedRolesToBlockedList:             Bool(false),
+					OdbcTreatDecimalAsInt:                            Bool(true),
+					PeriodicDataRekeying:                             Bool(false),
+					PipeExecutionPaused:                              Bool(true),
+					PreventUnloadToInlineURL:                         Bool(true),
```
sfc-gh-jmichalak, line 208: PreventLoadFromInlineURL is missing here, but present in testing unset.

pkg/sdk/testint/accounts_integration_test.go
```diff
@@ -520,6 +520,389 @@ func TestInt_Account_SelfAlter(t *testing.T) {
...
+
+		err := client.Accounts.Alter(ctx, &sdk.AlterAccountOptions{
+			Set: &sdk.AccountSet{
+				Parameters: &sdk.AccountParameters{
+					AbortDetachedQuery:                               sdk.Bool(true),
+					AllowClientMFACaching:                            sdk.Bool(true),
+					AllowIDToken:                                     sdk.Bool(true),
+					Autocommit:                                       sdk.Bool(false),
+					BaseLocationPrefix:                               sdk.String("STORAGE_BASE_URL/"),
+					BinaryInputFormat:                                sdk.Pointer(sdk.BinaryInputFormatBase64),
```
sfc-gh-jmichalak, line 556: Some are missing. If we can't provide any value and setting them fails, we should add a comment here, with a follow-up ticket.

pkg/sdk/common_types.go
sfc-gh-jcieslak: ON_EVENT is still supported, and PROPAGATE is just a new value. Added to the guide

pkg/sdk/assertions_test.go
sfc-gh-jcieslak: reverted

pkg/sdk/parameters_helpers.go
sfc-gh-jcieslak: I would leave this function. It was only moved in this pr and is used in old parameter logic that I don't want to touch. This function will be changed when we'll be refactoring the parameters, and for that reason I would leave it as it is for now.

pkg/sdk/parameters_helpers.go
sfc-gh-jcieslak: sure, extracted

pkg/sdk/parameters.go
sfc-gh-jcieslak: I would lave it as it is, I changed it for the consistency (because receivers for parameters were not the same always). We can change them later, but for now I wouldn't like to make this change bigger.

pkg/sdk/accounts_test.go
sfc-gh-jcieslak: Yeah, It should be removed, done

pkg/sdk/testint/accounts_integration_test.go
sfc-gh-jcieslak: Yeah, a few. Added a todo

docs/resources/database.md
```diff
@@ -113,7 +113,7 @@ resource "snowflake_database" "primary" {
 - `suspend_task_after_num_failures` (Number) How many times a task must fail in a row before it is automatically suspended. 0 disables auto-suspending. For more information, see [SUSPEND_TASK_AFTER_NUM_FAILURES](https://docs.snowflake.com/en/sql-reference/parameters#suspend-task-after-num-failures).
 - `task_auto_retry_attempts` (Number) Maximum automatic retries allowed for a user task. For more information, see [TASK_AUTO_RETRY_ATTEMPTS](https://docs.snowflake.com/en/sql-reference/parameters#task-auto-retry-attempts).
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
-- `trace_level` (String) Controls how trace events are ingested into the event table. Valid options are: [ALWAYS ON_EVENT OFF]. For information about levels, see [TRACE_LEVEL](https://docs.snowflake.com/en/sql-reference/parameters.html#label-trace-level).
+- `trace_level` (String) Controls how trace events are ingested into the event table. Valid options are: [ALWAYS ON_EVENT PROPAGATE OFF]. For information about levels, see [TRACE_LEVEL](https://docs.snowflake.com/en/sql-reference/parameters.html#label-trace-level).
```
sfc-gh-jmichalak: Nit: we can use `docs.PossibleValuesListed` in the schema to have better formatting here. In other places as well.

sfc-gh-jmichalak (approved):

pkg/sdk/parameters_test.go
```diff
@@ -180,18 +179,15 @@ func TestToAccountParameter(t *testing.T) {
 		{input: "STATEMENT_QUEUED_TIMEOUT_IN_SECONDS", want: AccountParameterStatementQueuedTimeoutInSeconds},
 		{input: "STATEMENT_TIMEOUT_IN_SECONDS", want: AccountParameterStatementTimeoutInSeconds},
 		{input: "STORAGE_SERIALIZATION_POLICY", want: AccountParameterStorageSerializationPolicy},
-		{input: "SHARE_RESTRICTIONS", want: AccountParameterShareRestrictions},
```
sfc-gh-asawicki, line 183: question: why a few of them were removed?

pkg/sdk/testint/accounts_integration_test.go
```diff
@@ -436,7 +436,7 @@ func TestInt_Account_SelfAlter(t *testing.T) {
 	assertParameterIsDefault := func(t *testing.T, parameters []*sdk.Parameter, parameterKey string) {
 		t.Helper()
 		param, err := collections.FindFirst(parameters, func(parameter *sdk.Parameter) bool { return parameter.Key == parameterKey })
-		require.NoError(t, err)
+		require.NoError(t, err, "parameter %v not found", parameterKey)
```
sfc-gh-asawicki, line 440: nit: it would be better not to add this here but improve the error return from findFirst, however, I think there is much logic linked to the currently returned err, so could you pls add it to SNOW-1325308?

pkg/sdk/testint/accounts_integration_test.go
```diff
@@ -520,6 +520,389 @@ func TestInt_Account_SelfAlter(t *testing.T) {
...
+		t.Cleanup(externalVolumeCleanup)
+
+		createNetworkPolicyRequest := sdk.NewCreateNetworkPolicyRequest(testClientHelper().Ids.RandomAccountObjectIdentifier()).WithAllowedIpList([]sdk.IPRequest{*sdk.NewIPRequest("0.0.0.0/0")})
+		networkPolicy, networkPolicyCleanup := testClientHelper().NetworkPolicy.CreateNetworkPolicyWithRequest(t, createNetworkPolicyRequest)
+		t.Cleanup(networkPolicyCleanup)
+
+		stage, stageCleanup := testClientHelper().Stage.CreateStage(t)
+		t.Cleanup(stageCleanup)
+
+		// TODO(SNOW2138715): Test all parameters
```
sfc-gh-asawicki: Could you pls list all the parameters that need special setup here (and/or in SNOW-2138715)?

pkg/sdk/testint/accounts_integration_test.go
```diff
@@ -520,6 +520,389 @@ func TestInt_Account_SelfAlter(t *testing.T) {
...
+	t.Run("set / unset parameters", func(t *testing.T) {
+		warehouseId := testClientHelper().Ids.WarehouseId()
+
+		eventTable, eventTableCleanup := testClientHelper().EventTable.Create(t)
+		t.Cleanup(eventTableCleanup)
+
+		externalVolumeId, externalVolumeCleanup := testClientHelper().ExternalVolume.Create(t)
+		t.Cleanup(externalVolumeCleanup)
+
+		createNetworkPolicyRequest := sdk.NewCreateNetworkPolicyRequest(testClientHelper().Ids.RandomAccountObjectIdentifier()).WithAllowedIpList([]sdk.IPRequest{*sdk.NewIPRequest("0.0.0.0/0")})
```
sfc-gh-asawicki, line 533: let's discuss tomorrow morning how we can prevent this running on the accounts it shouldn't be run on

sfc-gh-asawicki (changes requested):

docs/resources/database.md
sfc-gh-jcieslak: Done

pkg/sdk/parameters_test.go
sfc-gh-jcieslak: Because this tests `ToAccountParameter`. It was modified to support only the parameters we have in our document. The parameters that were not chosen are still usable in the account parameter resource because the list with those values was moved there. The `SHARE_RESTRICTIONS` doesn't work (I couldn't set it with alter and it's not in the SHOW PARAMETERS output, and docs), and the rest of the removed parameters are only returned by SHOW PARAMETERS, but are not documented in https://docs.snowflake.com/en/sql-reference/parameters.

pkg/sdk/testint/accounts_integration_test.go
sfc-gh-jcieslak: Yup, added a new point in SNOW-1325308 and added todo here.

pkg/sdk/testint/accounts_integration_test.go
sfc-gh-jcieslak: Yup, listed here.

pkg/sdk/testint/accounts_integration_test.go
sfc-gh-jcieslak: Will be added in the next pr.

sfc-gh-jmichalak (approved):

pkg/sdk/parameters_test.go
sfc-gh-asawicki: It's worth mentioning then in the migration guide that not all of the parameters from the account_parameter resources will be supported in the current account, and a note that they should be still managed there.

sfc-gh-asawicki (changes requested):
left one answer in existing thread

pkg/sdk/parameters_test.go
sfc-gh-jcieslak: Will be added in the next prs

PR #3747, opened 2025-06-06
feat: Add job service resource

sfc-gh-jmichalak:
## Changes
- Add a job service resource with limitations (missing templates for now).
- Extract the common code of service and job service resources.

## TODO
- Missing from_template in both resources.
- External type change detection.
- Validations in both resources.
- Service data source.

### DIFF
--- a/docs/resources/job_service.md
+++ b/docs/resources/job_service.md
@@ -10,7 +10,7 @@ description: |-
 <!-- TODO(SNOW-2129584): address this limitation -->
 !> **Caution** Only asynchronous job services are supported. This resource uses `ASYNC=TRUE` during creation. In this case, the command returns immediately while the job is running. Creating a job service automatically executes the job (read more in [EXECUTE JOB SERVICE docs](https://docs.snowflake.com/en/sql-reference/sql/execute-job-service)).
 
--> **Note** For asynchronous jobs, Snowflake does not perform automatic cleanup after completion. You must explicitly execute the `DROP SERVICE` command to remove the job. If you want to execute the job again, use the [replace flag](https://developer.hashicorp.com/terraform/cli/commands/apply#replace-resource).
+-> **Note** For asynchronous jobs, Snowflake does not perform automatic cleanup after completion. You must either remove the resource or execute the `DROP SERVICE` command to remove the job. If you want to execute the job again, use the [replace flag](https://developer.hashicorp.com/terraform/cli/commands/apply#replace-resource).
 
 -> **Note** Managing services via specification templates is not yet supported. This will be addressed in the next versions.
 
--- a/templates/resources/job_service.md.tmpl
+++ b/templates/resources/job_service.md.tmpl
@@ -14,7 +14,7 @@ description: |-
 <!-- TODO(SNOW-2129584): address this limitation -->
 !> **Caution** Only asynchronous job services are supported. This resource uses `ASYNC=TRUE` during creation. In this case, the command returns immediately while the job is running. Creating a job service automatically executes the job (read more in [EXECUTE JOB SERVICE docs](https://docs.snowflake.com/en/sql-reference/sql/execute-job-service)).
 
--> **Note** For asynchronous jobs, Snowflake does not perform automatic cleanup after completion. You must explicitly execute the `DROP SERVICE` command to remove the job. If you want to execute the job again, use the [replace flag](https://developer.hashicorp.com/terraform/cli/commands/apply#replace-resource).
+-> **Note** For asynchronous jobs, Snowflake does not perform automatic cleanup after completion. You must either remove the resource or execute the `DROP SERVICE` command to remove the job. If you want to execute the job again, use the [replace flag](https://developer.hashicorp.com/terraform/cli/commands/apply#replace-resource).
 
 -> **Note** Managing services via specification templates is not yet supported. This will be addressed in the next versions.
 


pkg/resources/job_service.go
```diff
@@ -0,0 +1,107 @@
...
+			ValidateDiagFunc: validateBooleanString,
+			DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("is_async_job"),
+			Description:      booleanStringFieldDescription("Specifies whether to execute the job service asynchronously."),
+			Default:          BooleanDefault,
+		},
+	}
+	return collections.MergeMaps(serviceBaseSchema(true), jobServiceSchema)
+}()
+
+func JobService() *schema.Resource {
```
sfc-gh-jcieslak, line 24: Question: Does job service removes itself right after its job is done (or after certain amount of time) or it has to be explicitly removed?

After checking (note: AI summary):
```
Synchronous jobs: Auto cleanup happens shortly after completion.
Asynchronous jobs: Persist indefinitely, require explicit DROP SERVICE for removal. If not dropped, they will remain "forever" (i.e., until manually deleted), even after the underlying compute resources are cleaned up.
```

If the above is true, we should have a special logic for auth cleaning jobs?

pkg/resources/service_commons.go
```diff
@@ -0,0 +1,273 @@
...
+			Computed:    true,
+			Description: "Outputs the result of `DESCRIBE SERVICE` for the given service.",
+			Elem: &schema.Resource{
+				Schema: schemas.DescribeServiceSchema,
+			},
+		},
+		// TODO(next PR): add service_type and acc tests for handling ext changes.
+	}
+	if allFieldsForceNew {
+		markSettableFieldsInSchemaAsForceNew(schema)
```
sfc-gh-jcieslak: Can't you just set `ForceNew: allFieldsForceNew` directly in the schema? (and use boolean logic for more complex cases).

pkg/resources/service_commons.go
```diff
@@ -0,0 +1,273 @@
...
+
+func ReadServiceCommonFunc(withExternalChangesMarking bool, extraOutputMappingsFunc func(service *sdk.Service) []outputMapping, extraSetStateToValuesFromConfigFields []string) schema.ReadContextFunc {
+	return func(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+		client := meta.(*provider.Context).Client
+		id, err := sdk.ParseSchemaObjectIdentifier(d.Id())
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		service, err := client.Services.ShowByIDSafely(ctx, id)
```
sfc-gh-jcieslak, line 175: With synchronous service jobs, this can cause a loop of removing and creating jobs potentially every terraform apply

pkg/testacc/resource_job_service_acceptance_test.go
```diff
@@ -0,0 +1,900 @@
...
+					resourceshowoutputassert.ServiceShowOutput(t, modelBasicOnStage.ResourceReference()).
+						HasName(id.Name()),
+					assert.Check(resource.TestCheckResourceAttr(modelBasicOnStage.ResourceReference(), "describe_output.0.name", id.Name())),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_JobService_complete(t *testing.T) {
```
sfc-gh-jcieslak, line 748: this is mostly covered by `basic_fromSpecification` test (due to recreate) where you set optionals and import afterwards

sfc-gh-jmichalak: Yeah, but there I set optional parameters on the object already in the state. Here, I set all options on a resource without state.

pkg/resources/service_commons.go
sfc-gh-jmichalak: There is a use case for this: someone may want to rerun a job after/before running the whole TF workflow, IMO.

pkg/resources/job_service.go
sfc-gh-jmichalak: Synchronous jobs are left in Snowflake for a few moments after finishing execution. Jobs can be removed just like any other service, even when they are running synchronously, so I don't think we need any special logic for now. I think we can mention it in the docs.

docs/resources/job_service.md
```diff
@@ -0,0 +1,198 @@
+---
+page_title: "snowflake_job_service Resource - terraform-provider-snowflake"
+subcategory: "Preview"
+description: |-
+  Resource used to manage job services. For more information, check services documentation https://docs.snowflake.com/en/sql-reference/sql/execute-job-service.
+---
+
```
sfc-gh-asawicki, line 7: let's specify why this resource is needed and when it should be used. Guide users to use the service resource for jobs they want to persist and this one for a single use. Guide what should be done with it after using it (remove from state, tainting, etc.).

also, handle the aouto-removal of the sync jobs or prevent using them.

pkg/resources/service_commons.go
```diff
@@ -0,0 +1,273 @@
...
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/schemas"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func serviceBaseSchema(allFieldsForceNew bool) map[string]*schema.Schema {
```
sfc-gh-asawicki, line 16: I believe this is almost a copy of the previous impl, I haven't verified it

pkg/resources/service_commons.go
sfc-gh-asawicki: @sfc-gh-jcieslak is right about the loop, though. It will cause a permanent diff which is never a good thing. For execution in each run, the datasource should be used instead.

sfc-gh-asawicki (changes requested):

pkg/resources/service_commons.go
sfc-gh-jmichalak: Done.

pkg/resources/service_commons.go
sfc-gh-jmichalak: That's true.

pkg/resources/service_commons.go
sfc-gh-jmichalak: In this case, we can just disable sync jobs.

sfc-gh-jmichalak: As we agreed off-line, the resource uses async=true by default for now. I also adjusted the docs.

docs/index.md
```diff
@@ -513,7 +513,7 @@ Preview features are disabled by default and should be used with caution.
 To use them, add the relevant feature name to the `preview_features_enabled` field in the [provider configuration](#preview_features_enabled-1).
 
 <!-- Section of stable resources -->
-## Currently stable resources
```
sfc-gh-asawicki: hmmm, are these spaces being removed and added back once in a while? I am pretty sure they were removed in the previous change/review

docs/resources/job_service.md
sfc-gh-asawicki: `Guide what should be done with it after using it (remove from state, tainting, etc.).` - this part was not adressed

pkg/resources/doc_helpers.go
```diff
@@ -56,3 +56,7 @@ func copyGrantsDescription(description string) string {
 func relatedResourceDescription(description string, resource providerresources.Resource) string {
 	return fmt.Sprintf(`%s For more information about this resource, see [docs](./%s).`, description, strings.TrimPrefix(resource.String(), "snowflake_"))
 }
+
+func longDescription(parts ...string) string {
```
sfc-gh-jcieslak: nit: rename to `JoinWithSpace`? \+ shouldn't we add documentation improvements to `docs` package?

docs/index.md
sfc-gh-jmichalak: I noticed it, too. I'm not sure what the cause is. Finally, nothing is changing in the PR.

pkg/testacc/resource_job_service_acceptance_test.go
sfc-gh-jcieslak: It's recreated, so I think the state is also "fresh" there, but I'm fine with leaving this one.

sfc-gh-jmichalak: Yes, but from the user's point of view these are two completely separate cases.

pkg/resources/doc_helpers.go
sfc-gh-jmichalak: Renamed. For now, I placed this function near all the other resources helpers.

docs/resources/job_service.md
sfc-gh-jmichalak: Added more guidelines. More details of what happens in Snowflake are also available in the docs.

templates/resources/job_service.md.tmpl
```diff
@@ -12,7 +12,9 @@ description: |-
...
 
 <!-- TODO(SNOW-2129584): address this limitation -->
-!> **Caution** Only asynchronous job services are supported. This resource uses `ASYNC=TRUE` during creation.
+!> **Caution** Only asynchronous job services are supported. This resource uses `ASYNC=TRUE` during creation. In this case, the command returns immediately while the job is running. Creating a job service automatically executes the job (read more in [EXECUTE JOB SERVICE docs](https://docs.snowflake.com/en/sql-reference/sql/execute-job-service)).
+
+-> **Note** For asynchronous jobs, Snowflake does not perform automatic cleanup after completion. You must explicitly execute the `DROP SERVICE` command to remove the job. If you want to execute the job again, use the [replace flag](https://developer.hashicorp.com/terraform/cli/commands/apply#replace-resource).
```
sfc-gh-jcieslak: Idk if the DROP SERVICE info needs to be mentioned here or in this way (it will be transparent for the users when they remove the resource; it kind of sounds like they have to run this command to clean things up).

sfc-gh-jmichalak: Rephrased.

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak:
### Merge activity

* **Jun 11, 11:03 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3747).
* **Jun 11, 11:03 AM UTC**: @sfc-gh-jmichalak merged this pull request with [Graphite](https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3747).

PR #3752, opened 2025-06-09
feat: Add services data source

sfc-gh-jmichalak:
## Changes
- Add a Service data source.
- Add a new struct and handling for ServiceIn.
- Adjust extracting `in` block (extract common code). Now this is slightly different, but the behavior does not change because we have the `ExactlyOneOf` setting.

## TODO
- Missing from_template in both resources.
- Validations in both resources.


docs/data-sources/services.md
```diff
@@ -0,0 +1,269 @@
...
+---
+
+!> **Caution: Preview Feature** This feature is considered a preview feature in the provider, regardless of the state of the resource in Snowflake. We do not guarantee its stability. It will be reworked and marked as a stable feature in future releases. Breaking changes are expected, even without bumping the major version. To use this feature, add the relevant feature name to `preview_features_enabled` field in the [provider configuration](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs#schema). Please always refer to the [Getting Help](https://github.com/snowflakedb/terraform-provider-snowflake?tab=readme-ov-file#getting-help) section in our Github repo to best determine how to get help for your questions.
+
+# snowflake_services (Data Source)
+
+Data source used to get details of filtered services. Filtering is aligned with the current possibilities for [SHOW SERVICES](https://docs.snowflake.com/en/sql-reference/sql/show-services) query. The results of SHOW and DESCRIBE are encapsulated in one output collection `services`.
```
sfc-gh-asawicki: nit: in this case we could add that it include both services and service jobs.

pkg/datasources/services.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+var servicesSchema = map[string]*schema.Schema{
+	"with_describe": {
+		Type:        schema.TypeBool,
+		Optional:    true,
+		Default:     true,
+		Description: "Runs DESC SERVICE for each service returned by SHOW SERVICES. The output of describe is saved to the description field. By default this value is set to true.",
+	},
+	"jobs_only": {
```
sfc-gh-asawicki: nit: I know it's now consistent with the SQL syntax but maybe we could make more consistent names here?

Additionally, should we make them exclusive (there is no point in setting both jobs_onnly and exclude_jobs, right)?

Maybe we could even make an enum for it: ALL, JOBS_ONLY, SERVICES_ONLY instead?

sfc-gh-asawicki (changes requested):

pkg/testacc/data_source_services_acceptance_test.go
```diff
@@ -0,0 +1,285 @@
...
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: accconfig.FromModels(t, serviceModel, dataSourceModel),
+				Check: assertThat(t,
+					assert.Check(resource.TestCheckResourceAttr(dataSourceModel.DatasourceReference(), "services.#", "1")),
+
+					resourceshowoutputassert.ServiceShowOutput(t, serviceModel.ResourceReference()).
```
sfc-gh-jcieslak: Why are we asserting show output of a resource here?

pkg/testacc/data_source_services_acceptance_test.go
```diff
@@ -0,0 +1,285 @@
...
+					assert.Check(resource.TestCheckResourceAttr(dataSourceModel.DatasourceReference(), "services.0.describe_output.0.managing_object_domain", "")),
+					assert.Check(resource.TestCheckResourceAttr(dataSourceModel.DatasourceReference(), "services.0.describe_output.0.managing_object_name", "")),
+				),
+			},
+			{
+				Config: accconfig.FromModels(t, serviceModel, dataSourceModelWithoutOptionals),
+				Check: assertThat(t,
+					assert.Check(resource.TestCheckResourceAttr(dataSourceModelWithoutOptionals.DatasourceReference(), "services.#", "1")),
+
+					resourceshowoutputassert.ServicesDatasourceShowOutput(t, "snowflake_services.test").
```
sfc-gh-jcieslak: nit: Use `dataSourceModelWithoutOptionals.DatasourceReference()`

docs/data-sources/services.md
sfc-gh-jmichalak: Done. Also, mentioned that the results can be filtered by a dedicated field.

pkg/testacc/data_source_services_acceptance_test.go
sfc-gh-jmichalak: Fixed.

pkg/testacc/data_source_services_acceptance_test.go
sfc-gh-jmichalak: Usually we reference by "" value, so that's a copy-paste. But changed here.

Note that `resourceshowoutputassert.XDatasourceShowOutput` must be adjusted because it prefixes the reference with `data`. If we want to use DatasourceReference, it must be removed, and it's not consistent across other data sources.

pkg/datasources/services.go
sfc-gh-jmichalak: Done.

PR #3758, opened 2025-06-12
chore: Update docs

sfc-gh-asawicki:
- Update the contribution guidelines
- Add `USE ROLE` warning to `snowflake_execute` resource

### DIFF
--- a/templates/resources/execute.md.tmpl
+++ b/templates/resources/execute.md.tmpl
@@ -13,6 +13,8 @@ description: |-
 
 !> **Warning** This is a dangerous resource that allows executing **ANY** SQL statement. It may destroy resources if used incorrectly. It may behave incorrectly combined with other resources. Use at your own risk.
 
+!> **Warning** Do not run `USE ROLE ...` statements as they will cause the provider to produce inconsistent results.
+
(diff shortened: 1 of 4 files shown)


templates/resources/execute.md.tmpl
```diff
@@ -13,6 +13,8 @@ description: |-
 
 !> **Warning** This is a dangerous resource that allows executing **ANY** SQL statement. It may destroy resources if used incorrectly. It may behave incorrectly combined with other resources. Use at your own risk.
 
+!> **Warning** Do not run `USE ROLE ...` statements as they will cause the provider to produce inconsistent results.
```
sfc-gh-jmichalak: 1. What about others like `USE SCHEMA`, `USE DATABASE`, etc? We use fully qualified names in the provider, so this should not be a problem, but we can consider discouraging these.
2. We should mention the behavior we recommend (using provider aliases).

sfc-gh-asawicki: 1. I am a bit hesitant about the database and schema for the reason you added. I added USE WAREHOUSE, though, as this can also "break" the provider.
2. Good one, added.

sfc-gh-jcieslak (approved):

PR #3759, opened 2025-06-12
chore: Account modificaiton test assertion

sfc-gh-jcieslak:
## Changes
- Add an assert that ensures a correct environment variables and account is used for org-level actions like, creating or modifying accounts.
- Tests to ensure the assert is working
- Use assert in account tests


pkg/testacc/1_prerequisites_test.go
```diff
@@ -0,0 +1,42 @@
+package testacc
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvs"
+)
+
+func TestEnsureValidAccountIsUsed(t *testing.T) {
```
sfc-gh-asawicki: These tests are a bit tricky as 1. they operate on the same `t` they want to validate, 2. they are located in one of the package utilizing testclient only (the second one being `testint`).

Could we move these to the `acceptance/helpers` package and maybe test it on the unit level?

pkg/testacc/1_prerequisites_test.go
```diff
@@ -0,0 +1,42 @@
...
+		defer func() {
+			if !t.Skipped() {
+				t.Errorf("Expected test to be skipped due to missing %s environment variable", testenvs.TestAccountCreate)
+			}
+		}()
+		testClient().EnsureValidNonProdAccountIsUsed(t)
+	})
+
+	t.Run(fmt.Sprintf("invalid account: %s not set", testenvs.TestNonProdModifiableAccountLocator), func(t *testing.T) {
+		t.Setenv(string(testenvs.TestNonProdModifiableAccountLocator), "")
```
sfc-gh-asawicki: what if `testenvs.TestAccountCreate` is set in the CI/local settings? I think that it should be set as this tests prerequisite. The same in test above with TestNonProdModifiableAccountLocator.

sfc-gh-asawicki (changes requested):

pkg/testacc/1_prerequisites_test.go
```diff
@@ -0,0 +1,42 @@
...
+		t.Setenv(string(testenvs.TestNonProdModifiableAccountLocator), accountLocator)
+		defer func() {
+			if t.Skipped() {
+				t.Errorf("Expected test to run with valid account, but it was skipped")
+			}
+		}()
+		testClient().EnsureValidNonProdAccountIsUsed(t)
+	})
+
+	t.Run(fmt.Sprintf("invalid account: %s not set", testenvs.TestAccountCreate), func(t *testing.T) {
```
sfc-gh-jmichalak: Nit: the name is slightly misleading, as we're not validating an account name here, but check if one of the envs is set.

pkg/testacc/1_prerequisites_test.go
sfc-gh-jcieslak: Had to somehow access the account locator to set it, but moved and it works. I'm wondering if we should do it the way i left it (with setter function dedicated for testing purposes) or just export account locator which on one hand exposes it and everyone can set it, but feels "cleaner", lmk.

pkg/testacc/1_prerequisites_test.go
sfc-gh-jcieslak: Adjusted, but not sure if that's what you meant

pkg/testacc/1_prerequisites_test.go
sfc-gh-jcieslak: Adjusted

pkg/sdk/client.go
```diff
@@ -86,6 +86,10 @@ type Client struct {
 	Warehouses                   Warehouses
 }
 
+func (c *Client) SetAccountLocatorForTests(accountLocator string) {
```
sfc-gh-jmichalak, line 89: I don't like exposing this just for testing. At least we can provide a constructor hiding this or use an interface.

sfc-gh-asawicki: I don't like it neither. The cleanest way would be to turn it into and interface and mock it on the unit test level.

pkg/acceptance/helpers/1_prerequisites_test.go
```diff
@@ -0,0 +1,57 @@
...
+			if t.Skipped() {
+				t.Errorf("Expected test to run with valid account, but it was skipped")
+			}
+		}()
+		testClient.EnsureValidNonProdAccountIsUsed(t)
+	})
+
+	t.Run(fmt.Sprintf("invalid account is used: should skip the tests as %s is not set", testenvs.TestAccountCreate), func(t *testing.T) {
+		accountLocator := "ABC123123"
+		client.SetAccountLocatorForTests(accountLocator)
```
sfc-gh-asawicki, line 36: nit: this setup could be done outside of each specific t.Run

pkg/sdk/client.go
sfc-gh-asawicki: We can have a follow-up for it I guess. But quite soon.

pkg/acceptance/helpers/1_prerequisites_test.go
sfc-gh-jcieslak: Will do in the follow-up

pkg/sdk/client.go
sfc-gh-jcieslak: I'll do a follow-up with different approach

Issue #3676, opened 2025-05-13
[Bug]: Changing name attribute in snowflake_view doesn't trigger recreate


landro:
### Terraform CLI Version

1.9.5

### Terraform Provider Version

1.0.5

### Company Name

Ruter

### Terraform Configuration

```terraform
resource "snowflake_view" "view" {
  database  = "db"
  schema    = "sch"
  name      = "old_view" -> "new_view"
  ...
}
```

### Category

category:identifiers

### Object type(s)

resource:view

### Expected Behavior

Expected the old view to be deleted, and the new one to be created

### Actual Behavior

The old view was **not** deleted, and the new view was created

Plan:
---  

snowflake_view.view will be updated in-place
  ~ resource "snowflake_view" "view" {
       ...
        id                   = "\"DB\".\"SCH\".\"OLD_VIEW\"". 

Apply:
---

snowflake_view.view: Modifying... [id="db"."sch"."old_view"]
snowflake_view.view: Modifications complete after 3s [id="db"."sch"."new_view"]

### Steps to Reproduce

1. Copy config
2. run apply
3. change name attribute
4. run apply

### How much impact is this issue causing?

Medium

### Additional Information

I see no docs concerning what triggers recreate in the terraform docs. Looking at the source (https://github.com/snowflakedb/terraform-provider-snowflake/blob/976b36c5702b5566eef7a1db04142f7920b7c188/pkg/resources/view.go#L28-L33) it seems changing the db or schema triggers a recreate, whereas a name change doesn't!. I find that strange.

### Would you like to implement a fix?

- [x] Yeah, I'll take it 😎

sfc-gh-asawicki:
Hey @landro. Thanks for reaching out to us.

I haven't tested it, but from looking at the code, it seems that if you alter one of the `statement`, `is_temporary`, `is_recursive`, `column` + at the same time you change the `name`, you may end up with new view created and the old one not being dropped. Can you confirm that the `name` was not the only attribute you have changed?

landro:
A ha - we actually changed the statement and one of the columns at the same time
You are referring to https://github.com/snowflakedb/terraform-provider-snowflake/blob/976b36c5702b5566eef7a1db04142f7920b7c188/pkg/resources/view.go#L872 I presume?

sfc-gh-asawicki:
[...]

PR #3766, opened 2025-06-16
fix: Update streamlit docs and fix view rename with modification

sfc-gh-asawicki:
- Update streamlit docs
- Fix view rename with modification

References: #3676

### DIFF
--- a/pkg/resources/view.go
+++ b/pkg/resources/view.go
@@ -868,12 +868,6 @@ func UpdateView(ctx context.Context, d *schema.ResourceData, meta any) diag.Diag
 		return diag.FromErr(err)
 	}
 
-	// change on these fields can not be ForceNew because then view is dropped explicitly and copying grants does not have effect
-	if keys := changedKeys(d, "statement", "is_temporary", "is_recursive", "column"); len(keys) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/resources/view.go
```diff
@@ -886,6 +880,12 @@ func UpdateView(ctx context.Context, d *schema.ResourceData, meta any) diag.Diag
 		id = newId
 	}
 
+	// change on these fields can not be ForceNew because then view is dropped explicitly and copying grants does not have effect
+	if keys := changedKeys(d, "statement", "is_temporary", "is_recursive", "column"); len(keys) > 0 {
```
sfc-gh-jcieslak, line 884: I guess this is also an issue in other resources that have a similar approach. The issue may occur whenever one of the keys responsible for recreation is changed with the name. If it's not documented in those resources, we should write a warning or an info note on their documentation page.

sfc-gh-jcieslak (approved):

pkg/resources/view.go
sfc-gh-asawicki: I have checked and I don't see the same problem in any other resources.

PR #3768, opened 2025-06-16
chore: Add current account resource policies

sfc-gh-jcieslak:
## Changes
- Add support for policies in the current_account resource.
- Add example usage.
- Add test client for feature policy.

## Next prs
- Address https://github.com/snowflakedb/terraform-provider-snowflake/pull/3745#discussion_r2144490524 (move account parameters)
- Address https://github.com/snowflakedb/terraform-provider-snowflake/pull/3745#discussion_r2144499012 (adjust setup for tests with parameters)

## Future prs
- Add organization user groups (this was postponed to be done in SNOW-2157505 because bigger changes were needed)


docs/resources/current_account.md
```diff
@@ -5,19 +5,150 @@ description: |-
   Resource used to manage the account you are currently connected to. This resource is used to set account parameters and other account-level settings. See ALTER ACCOUNT https://docs.snowflake.com/en/sql-reference/sql/alter-account documentation for more information on resource capabilities.
 ---
 
+!> **Warning** This resource requires warehouse to be in the context. To use this resource, specify a default warehouse in the provider configuration.
+
 !> **Warning** This resource shouldn't be used with `snowflake_account_parameter` resource in the same configuration, as it may lead to unexpected behavior. Unless the `snowflake_account_parameter` is used to manage the following parameters that are not supported by `snowflake_current_account`: ENABLE_CONSOLE_OUTPUT, ENABLE_PERSONAL_DATABASE, PREVENT_LOAD_FROM_INLINE_URL. They are not supported, because they are not in the [official parameters documentation](https://docs.snowflake.com/en/sql-reference/parameters). Once they are publicly documented, they will be added to the `snowflake_current_account` resource.
```
sfc-gh-asawicki: what about compatibility with `snowflake_account_authentication_policy_attachment` and `snowflake_network_policy_attachment` after adding policies? (I think there are no other account-level policy attachment resources but please verify them too)

pkg/acceptance/helpers/feature_policy_client.go
```diff
@@ -0,0 +1,46 @@
...
+		context: context,
+		ids:     idsGenerator,
+	}
+}
+
+func (c *FeaturePolicyClient) client() *sdk.Client {
+	return c.context.client
+}
+
+func (c *FeaturePolicyClient) Create(t *testing.T) (sdk.SchemaObjectIdentifier, func()) {
```
sfc-gh-asawicki, line 28: nit: please add a TODO here linking to the epic to support feature policies (I think you need to create one)

pkg/resources/current_account.go
```diff
@@ -43,13 +80,97 @@ func CurrentAccount() *schema.Resource {
...
+		if err := client.Accounts.Alter(ctx, &sdk.AlterAccountOptions{Set: setResourceMonitor}); err != nil {
+			return diag.FromErr(err)
+		}
+	} else {
+		if err := client.Accounts.Alter(ctx, &sdk.AlterAccountOptions{Unset: &sdk.AccountUnset{ResourceMonitor: sdk.Bool(true)}}); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	handlePolicyCreate := func(kind sdk.PolicyKind, policyIdFieldPointerGetter func(*sdk.AccountSet) **sdk.SchemaObjectIdentifier, hasForce bool, unsetPolicy *sdk.AccountUnset) error {
```
sfc-gh-asawicki, line 99: questions:
1. Are there any other resources that could benefit from this helper?
2. Do they all need to be handled as a separate ALTER? (more statements, more chances for error, additionally, the mentioned in other PR `d.Partial` should be considered)

pkg/resources/current_account.go
```diff
@@ -43,13 +80,97 @@ func CurrentAccount() *schema.Resource {
...
+		set := new(sdk.AccountSet)
+		if kind == sdk.PolicyKindFeaturePolicy {
+			set.FeaturePolicySet = new(sdk.AccountFeaturePolicySet)
+		}
+
+		if hasForce {
+			set.Force = sdk.Bool(true)
```
sfc-gh-asawicki, line 106: question: the logic here is that FORCE is used for each policy supporting the parameter. However, do we really need it in create? Create starts with a clean slate or is imported. I think we should care about it only in update, and the logic should be a bit different below in the update: "is there anything assigned currently? do we have update?" (as for now it's different)

pkg/sdk/accounts.go
```diff
@@ -244,16 +244,22 @@ type AccountSet struct {
 	PasswordPolicy       *SchemaObjectIdentifier  `ddl:"identifier" sql:"PASSWORD POLICY"`
 	SessionPolicy        *SchemaObjectIdentifier  `ddl:"identifier" sql:"SESSION POLICY"`
 	AuthenticationPolicy *SchemaObjectIdentifier  `ddl:"identifier" sql:"AUTHENTICATION POLICY"`
+	FeaturePolicySet     *AccountFeaturePolicySet `ddl:"keyword"`
 	Force                *bool                    `ddl:"keyword" sql:"FORCE"`
 }
 
+type AccountFeaturePolicySet struct {
+	FeaturePolicy      *SchemaObjectIdentifier `ddl:"identifier" sql:"FEATURE POLICY"`
+	forAllApplications bool                    `ddl:"static" sql:"FOR ALL APPLICATIONS"`
```
sfc-gh-asawicki, line 271: question: do we want to allow setting `FOR ALL APPLICATIONS` in the provider?

pkg/sdk/testint/accounts_integration_test.go
```diff
@@ -1071,37 +1036,44 @@ func TestInt_Account_SelfAlter(t *testing.T) {
...
+		err := client.Accounts.Alter(ctx, &sdk.AlterAccountOptions{Set: &sdk.AccountSet{FeaturePolicySet: &sdk.AccountFeaturePolicySet{FeaturePolicy: &featurePolicyId}}})
+		require.NoError(t, err)
+		assertThatPolicyIsSetOnAccount(t, featurePolicyId)
+		t.Cleanup(func() {
+			err = client.Accounts.Alter(ctx, &sdk.AlterAccountOptions{Unset: &sdk.AccountUnset{FeaturePolicyUnset: &sdk.AccountFeaturePolicyUnset{FeaturePolicy: sdk.Bool(true)}}})
+			require.NoError(t, err)
+			assertThatNoPolicyIsSetOnAccount(t)
 		})
+
+		err = client.Accounts.Alter(ctx, &sdk.AlterAccountOptions{Set: &sdk.AccountSet{FeaturePolicySet: &sdk.AccountFeaturePolicySet{FeaturePolicy: &newFeaturePolicyId}}})
```
sfc-gh-asawicki, line 1102: nit: let's leave a comment here like "expecting error as there is another policy already attached to the account"

and below: "using FORCE allows to SET the policy without having to unset the existing policy first"

pkg/testacc/resource_current_account_acceptance_test.go
```diff
@@ -630,9 +633,519 @@ func TestAcc_CurrentAccount_Parameters(t *testing.T) {
 			},
 			// unset parameters
 			{
-				Config: config.FromModels(t, unsetParametersModel),
+				Config: config.FromModels(t, provider, unsetParametersModel),
+				Check: assertThat(t,
+					resourceassert.CurrentAccountResource(t, unsetParametersModel.ResourceReference()).
+						HasAbortDetachedQueryString("false").
```
sfc-gh-asawicki: nit: we should extract this as our default (check `HasAllDefaults` for `UserResourceParametersAssert`)

pkg/testacc/resource_current_account_acceptance_test.go
```diff
@@ -630,9 +633,519 @@ func TestAcc_CurrentAccount_Parameters(t *testing.T) {
...
+						HasUnsupportedDdlActionString("ignore").
+						HasUserTaskManagedInitialWarehouseSizeString("Medium").
+						HasUserTaskMinimumTriggerIntervalInSecondsString("30").
+						HasUserTaskTimeoutMsString("3600000").
+						HasUseCachedResultString("true").
+						HasWeekOfYearPolicyString("0").
+						HasWeekStartString("0"),
+				),
+			},
+			// Test for external changes
```
sfc-gh-asawicki, line 420: nit: let's assert for the action update here

sfc-gh-asawicki (changes requested):

docs/resources/current_account.md
sfc-gh-jcieslak: Right, checked and added a few more resources this one shouldn't be used with

pkg/acceptance/helpers/feature_policy_client.go
sfc-gh-jcieslak: Done

pkg/resources/current_account.go
sfc-gh-jcieslak: 1. Looking at the objects where policies are used, I don't think we would get much out of extracting this.
2. Unfortunately, yes, but I don't think we need to worry that much about partial execution as we are not creating anything with this resource and the code is written in a way that it wouldn't suffer as much. I'm talking mostly about create that takes into consideration that some policies could be already assigned to the account before (and assigning/removing policies is problematic). 
  2.1. tldr: I can check where Partial could be used here, but I don't think it's needed as much as in other resources, because it doesn't create anything on Snowflake side.

pkg/resources/current_account.go
sfc-gh-jcieslak: Initially I was using Update in the Create method, but one test failed and I adjusted it to be also handled during create. The test that was failing was creating fully set config in the first step. I don't remember exactly why it wasn't properly filling up the state, but after the adjustment it worked. Regarding force, it was just to save the number of SQLs, before assigning any policy you have to make sure no other policy is set on the account (otherwise you'll get Snowflake error). To avoid the error and limit SQLs, I used force that is just capable of overriding the currently set policy without the need of unsetting it.

pkg/sdk/accounts.go
sfc-gh-jcieslak: In the [syntax](https://docs.snowflake.com/en/sql-reference/sql/alter-account) it seems it's required and afaik it failed without this part, so it seems to be a truly required param.

pkg/sdk/testint/accounts_integration_test.go
sfc-gh-jcieslak: Done

pkg/testacc/resource_current_account_acceptance_test.go
sfc-gh-jcieslak: Done

pkg/testacc/resource_current_account_acceptance_test.go
sfc-gh-jcieslak: Done

docs/resources/current_account.md
sfc-gh-asawicki: Two things:
1. A similar note should be added to all mentioned resources.
2. Why `object_parameter`?

pkg/resources/current_account.go
sfc-gh-asawicki: Still, you could just run it only if there is a policy assigned to the account (and you know it, as the read happened during refresh)... but yeah, there is no harm in leaving it. Maybe later it would be nice to verify if we can use FORCE on 3 other policies and if not, followup with the proper teams to verify why there re such inconsistencies

pkg/sdk/accounts.go
sfc-gh-asawicki: oke, thought it's optional

pkg/testacc/resource_current_account_acceptance_test.go
```diff
@@ -3,6 +3,7 @@
 package testacc
 
 import (
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
```
sfc-gh-asawicki: 👮 where is this import

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_current_account_acceptance_test.go
sfc-gh-jcieslak: :( re-ran the pre-push

docs/resources/current_account.md
sfc-gh-jcieslak: 1. added
2. it has an option to set a parameter on account

PR #3771, opened 2025-06-17
chore: Clarify SDK generator future

sfc-gh-asawicki:
- Add the disclaimer to the SDK generator's README
- Add the high-priority topics section based on the internal research
- Clean up the other sections slightly
- Other: change the receiver type and name for service object


pkg/sdk/poc/README.md
```diff
@@ -41,17 +43,46 @@ make clean-generator-session_policies run-generator-session_policies
...
+- Move manually added changes to the separate `_ext.go` files (e.g. [`procedures_ext.go`](https://github.com/snowflakedb/terraform-provider-snowflake/blob/5bdcd127d9288212b10ea7b138bebc0cb770c5b9/pkg/sdk/procedures_ext.go#L19))
+- Add definitions for the objects that were written by hand before the generator was created
+- In dto builder generator, use first letter lowercase for method arguments (e.g. [`OrReplace`](https://github.com/snowflakedb/terraform-provider-snowflake/blob/5bdcd127d9288212b10ea7b138bebc0cb770c5b9/pkg/sdk/procedures_dto_builders_gen.go#L26) -> `orReplace`)
+- Reuse existing `genhelpers.NewGenerator()` or other useful tools from the `genhelpers` package (to e.g. allow running it with command line arguments to regenerate only the chosen files)
+- Refresh the example
+
 ##### Essentials
 - generate each branch of alter in tests (instead of basic and all options)
 - clean up predefined operations in generator (now casting to string)
-- (?) add mapping from db to plain struct (for now only "// TODO: Mapping" comment is generated)
```
sfc-gh-jcieslak, line 47: I don't think we can remove this, at least this part: `add mapping from db to plain struct`. I think this was referring to the struct definitions for SHOW/DESC outputs, we have to define their structure twice, which can be improved to specify it only once (you can search `g.DbStruct` and `g.PlainStruct` for examples).

pkg/sdk/poc/README.md
```diff
@@ -41,17 +43,46 @@ make clean-generator-session_policies run-generator-session_policies
...
+    B := QueryStruct("B")
+        .ListQueryStructField(A) // A []A - validations will be wrong because this is array
+    ```
+  - generation for lists is not recursive, so we're only supporting one-level mapping
+    ```go
+    []Request1{ foo Request2, bar int } // won't be converted 
+    []Request1{ foo string, bar int } // will
+    ```
+  - optional structs inside slices are not being checked for nil in the toOpts() methods in the implementation file
+  - pointer fields are not being handled correctly in lists
```
sfc-gh-jcieslak, line 71: Do you remember what it was referring to?

pkg/sdk/poc/README.md
sfc-gh-asawicki: It's about the mapping which is already above (`Fully generate "convert" function body (from objectDBRow to object) as it needs to be added manually.`)

pkg/sdk/poc/README.md
sfc-gh-asawicki: Sent on priv

PR #3800, opened 2025-06-24
chore: Add basic listings to sdk

sfc-gh-jcieslak:
## Changes
- Add basic listing CRUD operations
- Add unit tests
- Add mapping helpers. I guess we could somehow gather those eventually into one package, but I didn't find any mapping helpers for the SDK convert, so I wrote them myself because of the large number of columns returned by SHOW/DESC. Those mappings could be later utilized in automatic conversion generation.

## Next prs
- Remove unnecessary `TODO: Not in Documentation` for output struct definitions
- Add the rest of the operations
- Add integration tests
- Check for lists of object types (e.g. grant ownership, and other that enable functionalities in other resources).


pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+	OptionalIdentifier("ApplicationPackage", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().SQL("APPLICATION PACKAGE")).
+	WithValidation(g.ExactlyOneValueSet, "Share", "ApplicationPackage")
+
+var listingDbRow = g.DbStruct("listingDBRow").
+	Text("global_name").
+	Text("name").
+	Text("title").
+	OptionalText("subtitle").
+	Text("profile").
+	Text("created_on").
```
sfc-gh-jmichalak, line 55: Time? Same for other fields below.

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+	Text("GlobalName").
+	Text("Name").
+	Text("Title").
+	OptionalText("Subtitle").
+	Text("Profile").
+	Text("CreatedOn").
+	Text("UpdatedOn").
+	OptionalText("PublishedOn").
+	Field("State", g.KindOfT[ListingState]()).
+	Text("ReviewState").
```
sfc-gh-jmichalak, line 87: Q: is this an enum as well?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+	OptionalText("request_approval_type").          // TODO: Not documented
+	OptionalText("monetization_display_order").     // TODO: Not documented
+	OptionalText("legacy_uniform_listing_locators") // TODO: Not documented
+
+var listingDetails = g.PlainStruct("ListingDetails").
+	Text("GlobalName").
+	Text("Name").
+	Text("Owner").
+	Text("OwnerRoleType").
+	Text("CreatedOn").
```
sfc-gh-jmichalak, line 173: Same, Time?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+	OptionalText("Description").
+	OptionalText("ListingTerms").
+	Field("State", g.KindOfT[ListingState]()).
+	Field("Share", g.KindOfTPointer[AccountObjectIdentifier]()).
+	Field("ApplicationPackage", g.KindOfTPointer[AccountObjectIdentifier]()).
+	OptionalText("BusinessNeeds").
+	OptionalText("UsageExamples").
+	OptionalText("DataAttributes"). // TODO: Not documented
+	OptionalText("Categories").
+	OptionalText("Resources").
```
sfc-gh-jmichalak, line 187: Q: Are some of these lists of strings? If yes, we can them.

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+
+var ListingsDef = g.NewInterface(
+	"Listings",
+	"Listing",
+	g.KindOfT[AccountObjectIdentifier](),
+).
+	CreateOperation(
+		"https://docs.snowflake.com/en/sql-reference/sql/create-listing",
+		g.NewQueryStruct("CreateListing").
+			Create().
```
sfc-gh-jmichalak, line 237: Q: Do `OR REPLACE` or `TAG` work?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+			Name().
+			OptionalQueryStructField("With", listingWithDef, g.KeywordOptions()).
+			TextAssignment("AS", g.ParameterOptions().NoEquals().DoubleDollarQuotes().Required()).
+			OptionalBooleanAssignment("PUBLISH", g.ParameterOptions()).
+			OptionalBooleanAssignment("REVIEW", g.ParameterOptions()).
+			OptionalComment().
+			WithValidation(g.ValidIdentifier, "name"),
+	).
+	CustomOperation(
+		"CreateFromStage",
```
sfc-gh-jmichalak: I wonder why these two ops are completely separated in the docs. In a similar case (services), they simply used `|` notation. Maybe there are/will be more differences?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+		"https://docs.snowflake.com/en/sql-reference/sql/create-listing",
+		g.NewQueryStruct("CreateListingFromStage").
+			Create().
+			SQL("EXTERNAL LISTING").
+			IfNotExists().
+			Name().
+			OptionalQueryStructField("With", listingWithDef, g.KeywordOptions()).
+			PredefinedQueryStructField("From", "Location", g.ParameterOptions().Required().NoQuotes().NoEquals().SQL("FROM")).
+			OptionalBooleanAssignment("PUBLISH", g.ParameterOptions()).
+			OptionalBooleanAssignment("REVIEW", g.ParameterOptions()).
```
sfc-gh-jmichalak: Q: COMMENT is missing from the docs. Does it work?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+			WithValidation(g.ValidIdentifier, "name"),
+	).
+	AlterOperation(
+		"https://docs.snowflake.com/en/sql-reference/sql/alter-listing",
+		g.NewQueryStruct("AlterListing").
+			Alter().
+			SQL("LISTING").
+			IfExists().
```
sfc-gh-jmichalak, line 255: Q: This suggests that `IF EXISTS` is available for all alters, but it's missing from `ADD VERSION` in the docs. So, does it work?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+				"AddVersion",
+				g.NewQueryStruct("AddListingVersion").
+					IfNotExists().
+					Text("VersionName", g.KeywordOptions().Required()).
+					PredefinedQueryStructField("From", "Location", g.ParameterOptions().Required().NoQuotes().NoEquals().SQL("FROM")).
+					OptionalComment(),
+				g.KeywordOptions().SQL("ADD VERSION"),
+			).
+			OptionalIdentifier("RenameTo", g.KindOfTPointer[AccountObjectIdentifier](), g.IdentifierOptions().SQL("RENAME TO")).
+			OptionalQueryStructField(
```
sfc-gh-jmichalak, line 279: Q: What about UNSET COMMENT and managing tags?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+			OptionalAssignment("REVISION", g.KindOfT[ListingRevision](), g.ParameterOptions().NoQuotes()).
+			WithValidation(g.ValidIdentifier, "name"),
+	)
+
+	// TODO(next prs): Organization listing may have its interface, but most of the operations would be pass through functions to the Listings interface
+	// TODO(next prs): Show available listings
+	// TODO(next prs): Show versions in listing
+	// TODO(next prs): Describe available listing
+	// TODO(next prs): Listing manifest builder - https://docs.snowflake.com/en/progaccess/listing-manifest-reference
+	// TODO(next prs): Test mapping functions (ToListingRevision and ToListingState)
```
sfc-gh-jmichalak, line 334: Are these all in the scope of this epic? E.g., the manifest builder doesn't seem to be.

pkg/sdk/listings_impl_gen.go
```diff
@@ -0,0 +1,270 @@
...
+		ReviewState:    r.ReviewState,
+		Owner:          r.Owner,
+		OwnerRoleType:  r.OwnerRoleType,
+		TargetAccounts: r.TargetAccounts,
+		IsMonetized:    r.IsMonetized,
+		IsApplication:  r.IsApplication,
+		IsTargeted:     r.IsTargeted,
+	}
+	if state, err := ToListingState(r.State); err == nil {
+		l.State = state
```
sfc-gh-jmichalak, line 162: Pls use `log.Printf` to detect cases when e.g. Snowflake adds a new state which will not be supported by the provider yet.

pkg/sdk/listings_gen_test.go
```diff
@@ -0,0 +1,394 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("validation: exactly one field from [opts.With.Share opts.With.ApplicationPackage] should be present", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.With = &ListingWith{}
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("CreateListingOptions.With", "Share", "ApplicationPackage"))
+	})
+
+	t.Run("validation: exactly one field from [opts.With.Share opts.With.ApplicationPackage] should be present", func(t *testing.T) {
```
sfc-gh-jmichalak: This test should have a different name.

pkg/sdk/listings_gen_test.go
```diff
@@ -0,0 +1,394 @@
...
+			Share:              &shareId,
+			ApplicationPackage: &applicationPackageId,
+		}
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("CreateListingOptions.With", "Share", "ApplicationPackage"))
+	})
+
+	t.Run("basic", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.As = `
+title: "MyListing"
```
sfc-gh-jmichalak: We should extract such listing definitions to the helpers package.

pkg/sdk/listings_gen_test.go
```diff
@@ -0,0 +1,394 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("validation: exactly one field from [opts.With.Share opts.With.ApplicationPackage] should be present", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.With = &ListingWith{}
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("CreateFromStageListingOptions.With", "Share", "ApplicationPackage"))
+	})
+
+	t.Run("validation: exactly one field from [opts.With.Share opts.With.ApplicationPackage] should be present", func(t *testing.T) {
```
sfc-gh-jmichalak: Same.

pkg/sdk/listings_gen_test.go
```diff
@@ -0,0 +1,394 @@
...
+		opts.AddVersion = &AddListingVersion{}
+		assertOptsInvalidJoinedErrors(t, opts, errOneOf("AlterListingOptions", "IfExists", "AddVersion"))
+	})
+
+	t.Run("validation: exactly one field from [opts.Publish opts.Unpublish opts.Review opts.AlterListingAs opts.AddVersion opts.RenameTo opts.Set] should be present", func(t *testing.T) {
+		opts := defaultOpts()
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterListingOptions", "Publish", "Unpublish", "Review", "AlterListingAs", "AddVersion", "RenameTo", "Set"))
+	})
+
+	t.Run("validation: exactly one field from [opts.Publish opts.Unpublish opts.Review opts.AlterListingAs opts.AddVersion opts.RenameTo opts.Set] should be present", func(t *testing.T) {
```
sfc-gh-jmichalak: Same.

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+					OptionalBooleanAssignment("PUBLISH", g.ParameterOptions()).
+					OptionalBooleanAssignment("REVIEW", g.ParameterOptions()).
+					OptionalComment(),
+				g.KeywordOptions().SQL("AS"),
+			).
+			OptionalQueryStructField(
+				"AddVersion",
+				g.NewQueryStruct("AddListingVersion").
+					IfNotExists().
+					Text("VersionName", g.KeywordOptions().Required()).
```
sfc-gh-jmichalak, line 273: Q: According to the docs, this is not required (wrapped in `[]`)?

pkg/sdk/mapping_helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	"log"
+)
+
+func mapStringIfNotNil(stringField **string, sqlValue sql.NullString) {
+	if sqlValue.Valid && sqlValue.String != "" {
+		*stringField = &sqlValue.String
+	}
+}
+
+func mapStringWithMappingIfNotNil[T any](stringField **T, sqlValue sql.NullString, mapper func(string) (T, error)) {
```
sfc-gh-jmichalak: Nit:
```suggestion
func mapNullStringWithMapping[T any](stringField **T, sqlValue sql.NullString, mapper func(string) (T, error)) {
```

sfc-gh-jmichalak (changes requested):

pkg/sdk/listings_def.go
sfc-gh-jcieslak: There was a discussion on the time in the result objects. I think we should discuss it. For now, I left it as a string, and I think we may prefer it this way, as the time format may differ depending on how it's set in Snowflake.

pkg/sdk/listings_def.go
sfc-gh-jcieslak: It most likely could be represented as enum, but there's no documentation on possible values, that's why I left it as string.

pkg/sdk/listings_def.go
sfc-gh-jcieslak: same as above

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Neither works.

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Well, maybe, but this is not as bad as when creating a request, you know what values are required (because they're in the request builder constructor), by merging the operations into one we would have to make 'as' and 'fromStage' as optional values and enforce setting one with validation which is less ideal imo.

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Hmm, it works... I'll adjust the definition in a bit

pkg/sdk/listings_def.go
sfc-gh-jcieslak: if exists cannot be used with add version, checked

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Hmm, unset comment seems to work (I'll add it), tags are not (answered above)

pkg/sdk/listings_def.go
sfc-gh-jcieslak: It's not, I'll create a separate ticket for it later add assign it to this todo

pkg/sdk/listings_impl_gen.go
sfc-gh-jcieslak: Done

pkg/sdk/listings_gen_test.go
sfc-gh-jcieslak: They are in the listing client, I don't think we need such a helper in the unit tests, where we only validate how the SQL is built.

pkg/sdk/listings_def.go
sfc-gh-jcieslak: It's required, will be reported to docs team

pkg/sdk/listings_def.go
sfc-gh-asawicki: We can wait with these changes for integration tests

pkg/sdk/listings_gen_test.go
sfc-gh-asawicki: However if this is reused, we have also small helpers for unit tests (e.g. sdk/random_test.go) but for me it would be enough to extract it as a top-level func in this file

pkg/sdk/object_types.go
```diff
@@ -80,6 +80,7 @@ const (
 	ObjectTypeGitRepository        ObjectType = "GIT REPOSITORY"
 	ObjectTypeModel                ObjectType = "MODEL"
 	ObjectTypeService              ObjectType = "SERVICE"
+	ObjectTypeListing              ObjectType = "LISTING"
```
sfc-gh-asawicki, line 83: question: there are some other places (like `allObjectTypes`, `PluralObjectType` or `objectTypeSingularToPluralMap` in this file, and places in other files like granting, ownership, etc). Can you leave a TODO to support all these as part of your current listings work?

pkg/sdk/mapping_helpers.go
```diff
@@ -0,0 +1,38 @@
...
+		*boolField = &sqlValue.Bool
+	}
+}
+
+func mapStringWithMapping[T any](stringField *T, sqlValue string, mapper func(string) (T, error)) {
+	if sqlValue != "" {
+		if mappedValue, err := mapper(sqlValue); err == nil {
+			*stringField = mappedValue
+		} else {
+			log.Printf("[WARN] Failed to map string value: %s, err = %s", sqlValue, err)
```
sfc-gh-asawicki: can we log this value always? (what about the potentially sensitive values?)

pkg/sdk/mapping_helpers.go
```diff
@@ -0,0 +1,38 @@
...
+		*stringField = &sqlValue.String
+	}
+}
+
+func mapStringWithMappingIfNotNil[T any](stringField **T, sqlValue sql.NullString, mapper func(string) (T, error)) {
+	if sqlValue.Valid && sqlValue.String != "" {
+		if mappedValue, err := mapper(sqlValue.String); err == nil {
+			*stringField = &mappedValue
+		} else {
+			log.Printf("[WARN] Failed to map string value: %s, err = %s", sqlValue.String, err)
```
sfc-gh-asawicki: can we log this value always? (what about the potentially sensitive values?)

pkg/sdk/mapping_helpers.go
```diff
@@ -0,0 +1,38 @@
+package sdk
+
+import (
+	"database/sql"
+	"log"
+)
+
+func mapStringIfNotNil(stringField **string, sqlValue sql.NullString) {
```
sfc-gh-asawicki: nit: I would like to see unit tests for all these helpers

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+	OptionalText("RequestApprovalType").         // TODO: Not documented
+	OptionalText("MonetizationDisplayOrder").    // TODO: Not documented
+	OptionalText("LegacyUniformListingLocators") // TODO: Not documented
+
+var ListingsDef = g.NewInterface(
+	"Listings",
+	"Listing",
+	g.KindOfT[AccountObjectIdentifier](),
+).
+	CreateOperation(
```
sfc-gh-asawicki, line 234: question: what about organization listing?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+		"https://docs.snowflake.com/en/sql-reference/sql/create-listing",
```
sfc-gh-asawicki, line 235: question: the object in SQL is called EXTERNAL LISTING; we have also ORGANIZATION LISTING:
- what keyword is used for grants? is there a difference between these two types?
- should we call it just Listing or ExternalListing (I guess on the resource level we will be explicit so that our users understand which resource they need to use)?

pkg/sdk/listings_def.go
sfc-gh-asawicki: But we could still add two more constructors: 1 requiring yaml and one requiring stage. I'm just saying, for me we can merge these functions and we can leave them separated.

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+			SQL("EXTERNAL LISTING").
+			IfNotExists().
+			Name().
+			OptionalQueryStructField("With", listingWithDef, g.KeywordOptions()).
+			PredefinedQueryStructField("From", "Location", g.ParameterOptions().Required().NoQuotes().NoEquals().SQL("FROM")).
+			OptionalBooleanAssignment("PUBLISH", g.ParameterOptions()).
+			OptionalBooleanAssignment("REVIEW", g.ParameterOptions()).
+			WithValidation(g.ValidIdentifier, "name"),
+	).
+	AlterOperation(
```
sfc-gh-asawicki, line 250: question: are all of these appliable also to the ORGANIZATION LISTING? (I don't see a separate section for ALTER ORGANIZATION LISTING)

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+			).
+			OptionalIdentifier("RenameTo", g.KindOfTPointer[AccountObjectIdentifier](), g.IdentifierOptions().SQL("RENAME TO")).
+			OptionalQueryStructField(
+				"Set",
+				g.NewQueryStruct("ListingSet").
+					OptionalComment(),
+				g.KeywordOptions().SQL("SET"),
+			).
+			WithValidation(g.ValidIdentifier, "name").
+			WithValidation(g.ConflictingFields, "IfExists", "AddVersion").
```
sfc-gh-asawicki, line 292: Have you checked that they can't be mixed really or is it just the docs problem?

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+		g.NewQueryStruct("DescribeListing").
+			Describe().
+			SQL("LISTING").
+			Name().
+			OptionalAssignment("REVISION", g.KindOfT[ListingRevision](), g.ParameterOptions().NoQuotes()).
+			WithValidation(g.ValidIdentifier, "name"),
+	)
+
+	// TODO(next prs): Organization listing may have its interface, but most of the operations would be pass through functions to the Listings interface
+	// TODO(next prs): Show available listings
```
sfc-gh-asawicki, line 330: do we need all these other shows for the resource? If not, please sskip these and create a follow-up issue for the SDK completeness.

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+	OptionalBool("IsLimitedTrial").
+	OptionalBool("IsByRequest").
+	OptionalText("Distribution").
+	OptionalBool("IsMountlessQueryable").
+	OptionalText("RejectedOn").
+	OptionalText("OrganizationProfileName").
+	OptionalText("UniformListingLocator").
+	OptionalText("DetailedTargetAccounts")
+
+var listingDetailsDbRow = g.DbStruct("listingDetailsDBRow").
```
sfc-gh-asawicki, line 107: please leave a note that currently there are more fields listed here than in the docs

pkg/sdk/listings_def.go
```diff
@@ -0,0 +1,349 @@
...
+	OptionalText("regions").
+	OptionalText("refresh_schedule").
+	OptionalText("refresh_type").
+	Text("review_state").
+	OptionalText("rejection_reason").
+	OptionalText("unpublished_by_admin_reasons").
+	Bool("is_monetized").
+	Bool("is_application").
+	Bool("is_targeted").
+	OptionalBool("is_limited_trial").    // TODO: Not documented
```
sfc-gh-asawicki, line 143: what do you plan to do with these TODOs? we can't keep them without the issue number

sfc-gh-asawicki (changes requested):

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Merged into one (there will be less work on the resource side to differentiate between the create types, and the difference between those two is only the way of providing manifest (as after other comment I checked the `comment` property can be set on both).

pkg/sdk/listings_gen_test.go
sfc-gh-jcieslak: Done

pkg/sdk/listings_gen_test.go
sfc-gh-jcieslak: Create a helper function on top of the file

pkg/sdk/listings_gen_test.go
sfc-gh-jcieslak: Done

pkg/sdk/listings_gen_test.go
sfc-gh-jcieslak: Done

pkg/sdk/mapping_helpers.go
sfc-gh-jcieslak: renamed

pkg/sdk/object_types.go
sfc-gh-jcieslak: Good point, added in the pr desc TODO section

pkg/sdk/mapping_helpers.go
sfc-gh-jcieslak: Sure, added

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Will be added later on, I wanted to first focus on the first type. I was wondering if we would like to separate interfaces (I left the TODO at the bottom), but it would only contain create, and other operations would be passed through to this interface. I think for now maybe I'll add it here.

pkg/sdk/listings_def.go
sfc-gh-jcieslak: 1. It seems you can't grant anything to listings
2. Yes, on the resource level we should be explicit, but here maybe it should be one interface that holds both (? I also left a TODO at the bottom of the def file to see what we should go with, but I think one interface would be Ok in this case)

pkg/sdk/listings_def.go
sfc-gh-jcieslak: This will mostly come out during integration tests, but it seems like the alter is made for both

pkg/sdk/listings_def.go
sfc-gh-jcieslak: [ref](https://github.com/snowflakedb/terraform-provider-snowflake/pull/3800#discussion_r2163735696) I checked, it doesn't work

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Sure, I'll check what operations are necessary for resource implementation and create a follow-up for unimplemented things

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Will remove these, I just wanted to have a way to show how much of the fields are skipped in the documentation. Most likely I'll remove them in the next pr (after I forward this to the doc team).

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Sure, added

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Discussed with @sfc-gh-asawicki, we should, for now, continue with `strings` (mostly for the reason above). Later on, we should come up with a solution that would enable us to handle time with `time.Time` and apply it across every object, but that's a future improvement (it doesn't affect the usability on the provider side that much as you can parse and modify the dates/timestamps within tf configuration).

pkg/sdk/mapping_helpers.go
sfc-gh-jcieslak: Removed the value logging and moved the responsibility to the error producer to hide the sensitive parts whenever the parsed value is sensitive.

pkg/sdk/listings_impl_gen.go
sfc-gh-jmichalak: It seems it was overridden by the generator, similarly below.

pkg/sdk/listings_def.go
sfc-gh-jmichalak: Still not done, but we can continue for now.

pkg/sdk/listings_gen_test.go
sfc-gh-jmichalak: Yeah, for now, it's not necessary, but it's helpful for integration and acceptance tests (see SampleSpecWithContainerName as an example).

sfc-gh-jmichalak (changes requested):

pkg/sdk/listings_impl_gen.go
sfc-gh-jcieslak: fixed

sfc-gh-asawicki (approved):

pkg/sdk/listings_impl_gen.go
sfc-gh-jmichalak: Still, the error is not logged. It can be done in a follow-up.

sfc-gh-jmichalak (approved):

PR #3804, opened 2025-06-24
chore: Set up muxing in tests

sfc-gh-asawicki:
- Set up a plugin framework provider with an example resource on the acceptance tests level
- Fix review remarks from the previous change (https://github.com/snowflakedb/terraform-provider-snowflake/pull/3787)
- Bump sdkv2, plugin framework, and other terraform-related libraries to the newest versions (it's a recommended step before mixing; also, this one is related: https://github.com/hashicorp/terraform-plugin-framework/issues/1148)
- Set up the mixed provider using the framework
  - It needs to have the exact same schema, so it was generated (as we may need to introduce some changes before moving to the plugin framework) and adjusted manually where needed
  - The configure method has only required settings; the configure logic has not yet been duplicated/reused
  - A simple resource was added (not all of the methods were implemented)

Next (some follow-ups will be soon, some will be transformed into future actions; for now, all are marked with the `[mux-PR]` TODO):
- Set up muxing for functional tests
- Test the resource with Snowflake connection (requires a proper configuration step, env handling, preview features etc.; it also requires full resource implementation - we have a really simple one for this test)
- Test authentication methods and configuration correctness in the plugin framework provider
- Improvements in the config generation (blocks and collections support; versioning)
- Set provider version automatically, like in tracking
- Dig into plan modifiers
- Extract resource helpers (like to get metadata)
- The current provider built on top of the plugin framework does not use the custom configure logic (e.g, with the configure once setting)
- Improve acceptance test (check destroy + version checks)
- Add acceptance test with resources from both providers (SDK and plugin framework)


pkg/testacc/poc/gen/mapper.go
```diff
@@ -0,0 +1,96 @@
+package gen
+
+import (
+	"fmt"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/genhelpers"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type pluginFrameworkType string
```
sfc-gh-jcieslak, line 10: Do we need to have our own type when the same thing is in the `schema` package?

sfc-gh-jcieslak (approved):

pkg/testacc/poc/gen/mapper.go
sfc-gh-asawicki: In the schema package, you have the int-based enum, and this one is used for the string mappings that we use.

pkg/testacc/poc/gen/sdk_v2_provider_schemas.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+type SdkV2ProviderSchema map[string]*schema.Schema
+
+func (s SdkV2ProviderSchema) ObjectName() string {
+	return "Snowflake"
+}
+
+// SdkV2ProviderSchemas contains all provider schemas
+var SdkV2ProviderSchemas = []SdkV2ProviderSchema{
```
sfc-gh-jmichalak, line 15: Is it used somewhere?

pkg/testacc/poc/gen/mapper.go
```diff
@@ -0,0 +1,96 @@
...
+	pfType, err := pluginFrameworkTypeFromSdkV2Type(fieldSchema.Type)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ProviderSchemaEntry{
+		Key:                                key,
+		PluginFrameworkSchemaAttributeType: pfType.ToSchemaAttributeType(),
+		Description:                        fieldSchema.Description,
+		Optional:                           fieldSchema.Optional || !fieldSchema.Required,
```
sfc-gh-jmichalak, line 93: I'm not sure why we need the `|| !fieldSchema.Required` part.

pkg/testacc/plugin_framework_poc_acceptance_test.go
```diff
@@ -0,0 +1,42 @@
...
+		CheckDestroy: nil,
+		Steps: []resource.TestStep{
+			{
+				Config: someResourceConfig(id),
+				Check: assertThat(t,
+					assert.Check(resource.TestCheckResourceAttr("snowflake_some.test", "id", id.FullyQualifiedName())),
+				),
+			},
+		},
+	})
```
sfc-gh-jmichalak, line 33: Q: In the next PRs we'll get a test with both `snowflake_some` and an actual resource from the sdkv2 provider, right?

pkg/testacc/poc/gen/sdk_v2_provider_schemas.go
sfc-gh-asawicki: Yes, in the generator main

pkg/testacc/plugin_framework_poc_acceptance_test.go
sfc-gh-asawicki: Yes, good point, it's missing in the PR description, added.

pkg/testacc/poc/gen/mapper.go
sfc-gh-asawicki: TBH I don't think it's needed too. The reason I added this was that the schema struct have both these fields and they can't be used together, but at least one of `Required`, `Optional`, `Optional` and `Computed`, or `Computed` must be enabled. I guess that our current provider only uses Optional: true for everything so I will simplify this logic for now (as it is not currently needed and is not fully correct either way).

pkg/testacc/poc/gen/sdk_v2_provider_schemas.go
sfc-gh-jmichalak: OK, for some reason the search results didn't show it :/

sfc-gh-jmichalak (approved):

PR #3815, opened 2025-06-27
chore: Account test assert follow up

sfc-gh-jcieslak:
## Changes
- Address comments from #3759
- Create TestClientBuilder for use cases that require mocked `sdk.Client` and use it in `1_prerequisites_test.go`

## TODO
- Use different env name for non prod organization account (currently the same env is used)


pkg/acceptance/helpers/1_prerequisites_test.go
```diff
@@ -9,18 +9,16 @@ import (
 )
 
 func TestEnsureValidAccountIsUsed(t *testing.T) {
-	client := sdk.Client{}
+	accountLocator := "ABC123123"
+	clientBuilder := sdk.NewTestClientBuilder().WithAccountLocator(accountLocator)
 
 	testClient := TestClient{
 		context: &TestClientContext{
```
sfc-gh-jmichalak: Why don't we:
- Change `client` in `TestClientContext` to an interface
- The proper client implements this new interface almost out of the box
- We can have a test mock client in a separate package and not pollute the `sdk` package with test objects.
WDYT?

pkg/sdk/client_test_builder.go
```diff
@@ -0,0 +1,159 @@
+package sdk
```
sfc-gh-asawicki: I am not convinced we need this builder at all. We create this client through a dedicated `NewClient` method, that set's it up properly. There are no other reasons to set this client up, so I don't think we should have this builder for the testing purposes only. For the sake of your test, the only thing is needed is the interface extraction to be able to mock GetAccountLocator function.

sfc-gh-asawicki (changes requested):

pkg/acceptance/helpers/1_prerequisites_test.go
sfc-gh-jcieslak: Changing client into interface cannot be done easily. I tried to go with different approach of having AccountInformation interface that test client implements and it's mocked in the test.

pkg/sdk/client_test_builder.go
sfc-gh-jcieslak: Builder removed and GetAccountLocator extracted as AccountInformation interface (name could be different, e.g. AccountContext)

PR #3818, opened 2025-06-27
chore: Add current organization account resource sdk part

sfc-gh-jcieslak:
## Changes
- Add organization account operations to the SDK
- Create definitions for the organization accounts
- Add unit and integration tests (only for ALTERing operations)
- ❗ The mapping was taken from listings PR
- Introduce an assert for tests to only run if you're on the non-prod organization account to avoid running the tests by mistake

## TODO
- Test CREATE operations (left as a task for later as current_organization_account won't need CREATE)


pkg/sdk/mapping_helpers.go
```diff
@@ -0,0 +1,38 @@
+package sdk
```
sfc-gh-jcieslak: You can skip this file (it was copied from the listing PR and it was already adjusted and tested there)

pkg/sdk/object_types.go
```diff
@@ -80,6 +80,7 @@ const (
 	ObjectTypeGitRepository        ObjectType = "GIT REPOSITORY"
 	ObjectTypeModel                ObjectType = "MODEL"
 	ObjectTypeService              ObjectType = "SERVICE"
+	ObjectTypeOrganizationAccount  ObjectType = "ORGANIZATION_ACCOUNT"
```
sfc-gh-jcieslak: This is mostly for separation purposes, the grants and maybe other entities treat this as `ACCOUNT` object type.

pkg/acceptance/helpers/1_prerequisites_test.go
```diff
@@ -55,3 +57,90 @@ func TestEnsureValidAccountIsUsed(t *testing.T) {
...
+			{
+				AccountLocator: accountLocator,
+			},
+		},
+	}
+
+	client := sdk.Client{
+		OrganizationAccounts: organizationAccounts,
+	}
+	client.SetAccountLocatorForTests(accountLocator)
```
sfc-gh-asawicki, line 102: are we still using it? :(

pkg/sdk/object_types.go
sfc-gh-asawicki: What about plural and all other collections with object types?

pkg/sdk/organization_accounts_def.go
```diff
@@ -0,0 +1,154 @@
...
+
+	g "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/poc/generator"
+)
+
+//go:generate go run ./poc/main.go
+
+type OrganizationAccountEdition string
+
+var (
+	OrganizationAccountEditionEnterprise       OrganizationAccountEdition = "ENTERPRISE"
```
sfc-gh-asawicki, line 16: question: can't be standard, right?

pkg/acceptance/helpers/1_prerequisites.go
```diff
@@ -77,3 +77,21 @@ func (c *TestClient) EnsureValidNonProdAccountIsUsed(t *testing.T) {
...
+
+func (c *TestClient) EnsureValidNonProdOrganizationAccountIsUsed(t *testing.T) {
+	t.Helper()
+	nonProdModifiableAccountLocator := testenvs.GetOrSkipTest(t, testenvs.TestNonProdModifiableAccountLocator)
+	if c.context.client.GetAccountLocator() != nonProdModifiableAccountLocator {
+		t.Skipf("Current client account locator does not match the required non-prod modifiable account's locator set in %s env variable. Skipping test.", testenvs.TestNonProdModifiableAccountLocator)
+	}
+	organizationAccounts, err := c.context.client.OrganizationAccounts.Show(context.Background(), sdk.NewShowOrganizationAccountRequest())
+	if err != nil {
+		t.Skipf("Failed to show organization accounts, err = %v.", err)
```
sfc-gh-jmichalak: Why skip on error? We won't notice if it fails.

pkg/sdk/object_types.go
sfc-gh-jmichalak: But it's not used anywhere?

pkg/sdk/organization_accounts_def.go
```diff
@@ -0,0 +1,154 @@
...
+					OptionalIdentifier("PasswordPolicy", g.KindOfTPointer[SchemaObjectIdentifier](), g.IdentifierOptions().SQL("PASSWORD POLICY")).
+					OptionalIdentifier("SessionPolicy", g.KindOfTPointer[SchemaObjectIdentifier](), g.IdentifierOptions().SQL("SESSION POLICY")).
+					WithValidation(g.AtLeastOneValueSet, "Parameters", "ResourceMonitor", "PasswordPolicy", "SessionPolicy"),
+				g.KeywordOptions().SQL("SET"),
+			).
+			OptionalQueryStructField(
+				"Unset",
+				g.NewQueryStruct("OrganizationAccountUnset").
+					PredefinedQueryStructField("Parameters", g.KindOfTPointer[AccountParametersUnset](), g.ListOptions().NoParentheses()).
+					OptionalBool("ResourceMonitor", g.KeywordOptions().SQL("RESOURCE_MONITOR")).
```
sfc-gh-asawicki: why new OptionalBool when for other unsets we use `OptionalSQL` ?

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -0,0 +1,519 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeroles"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInt_OrganizationAccount_SelfAlter(t *testing.T) {
+	testClientHelper().EnsureValidNonProdAccountIsUsed(t)
```
sfc-gh-asawicki: not `EnsureValidNonProdOrganizationAccountIsUsed`?

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -0,0 +1,519 @@
...
+		// TODO(SNOW-2138715): Test all parameters, the following parameters were not tested due to more complex setup:
+		// - ActivePythonProfiler
+		// - CatalogSync
+		// - EnableInternalStagesPrivatelink
+		// - PythonProfilerModules
+		// - S3StageVpceDnsName
+		// - SamlIdentityProvider
+		// - SimulatedDataSharingConsumer
+		err := client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithSet(*sdk.NewOrganizationAccountSetRequest().WithParameters(sdk.AccountParameters{
+			AbortDetachedQuery:                               sdk.Bool(true),
```
sfc-gh-asawicki: I guess we use the exact same setting for the "normal" account setup, let's extract

pkg/acceptance/helpers/1_prerequisites_test.go
```diff
@@ -55,3 +57,90 @@ func TestEnsureValidAccountIsUsed(t *testing.T) {
...
+}
+
+func (m *mockOrganizationAccounts) Create(ctx context.Context, request *sdk.CreateOrganizationAccountRequest) error {
+	return nil
+}
+
+func (m *mockOrganizationAccounts) Alter(ctx context.Context, request *sdk.AlterOrganizationAccountRequest) error {
+	return nil
+}
+
```
sfc-gh-jmichalak, line 78: Nit: can we verify that the mock implementation satisfies the interface?

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -0,0 +1,519 @@
...
+			WeekStart:                                        sdk.Int(1),
+		})))
+		require.NoError(t, err)
+
+		// TODO: Make for organization accounts, but passthrough the account interface
+		parameters, err := client.Accounts.ShowParameters(ctx)
+		require.NoError(t, err)
+		require.NotEmpty(t, parameters)
+
+		assertParameterValueSetOnAccount(t, parameters, string(sdk.AccountParameterAbortDetachedQuery), "true")
```
sfc-gh-asawicki: We have objectparameters asserts that could be used here. Additionally, we could extract a common validation for this and "normal" account tests.

pkg/sdk/organization_accounts_def.go
```diff
@@ -0,0 +1,154 @@
...
+			OptionalComment().
+			WithValidation(g.ValidIdentifier, "name").
+			WithValidation(g.AtLeastOneValueSet, "AdminPassword", "AdminRsaPublicKey"),
+	).
+	AlterOperation(
+		"https://docs.snowflake.com/en/sql-reference/sql/alter-organization-account",
+		g.NewQueryStruct("AlterOrganizationAccount").
+			Alter().
+			SQL("ORGANIZATION ACCOUNT").
+			OptionalIdentifier("Name", g.KindOfTPointer[AccountObjectIdentifier](), g.IdentifierOptions()).
```
sfc-gh-jmichalak, line 63: Does specifying the name work for all of the operations? According to the docs, it's only for renaming and dropping the old URL.

EDIT: I saw the validations below, but we should consider splitting the ops, IMO cc @sfc-gh-asawicki

pkg/sdk/organization_accounts_def.go
```diff
@@ -0,0 +1,154 @@
...
+			OptionalQueryStructField(
+				"Set",
+				g.NewQueryStruct("OrganizationAccountSet").
+					// Currently, Organization Accounts use the same set of parameters as regular accounts
+					PredefinedQueryStructField("Parameters", g.KindOfTPointer[AccountParameters](), g.ListOptions().NoParentheses()).
+					OptionalIdentifier("ResourceMonitor", g.KindOfTPointer[AccountObjectIdentifier](), g.IdentifierOptions().Equals().SQL("RESOURCE_MONITOR")).
+					OptionalIdentifier("PasswordPolicy", g.KindOfTPointer[SchemaObjectIdentifier](), g.IdentifierOptions().SQL("PASSWORD POLICY")).
+					OptionalIdentifier("SessionPolicy", g.KindOfTPointer[SchemaObjectIdentifier](), g.IdentifierOptions().SQL("SESSION POLICY")).
+					WithValidation(g.AtLeastOneValueSet, "Parameters", "ResourceMonitor", "PasswordPolicy", "SessionPolicy"),
```
sfc-gh-jmichalak: Not exactly one value set?

pkg/sdk/organization_accounts_def.go
```diff
@@ -0,0 +1,154 @@
...
+				g.KeywordOptions().SQL("UNSET"),
+			).
+			OptionalSetTags().
+			OptionalUnsetTags().
+			OptionalQueryStructField(
+				"RenameTo",
+				g.NewQueryStruct("OrganizationAccountRename").
+					Identifier("NewName", g.KindOfTPointer[AccountObjectIdentifier](), g.IdentifierOptions().Required().SQL("RENAME TO")).
+					OptionalBooleanAssignment("SAVE_OLD_URL", g.ParameterOptions()),
```
sfc-gh-jmichalak, line 91: Can we just reuse the struct we use in renaming accounts?

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -0,0 +1,519 @@
+//go:build !account_level_tests
```
sfc-gh-jmichalak: I think these tests should be on the account level. We are modifying the current account here.

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -0,0 +1,519 @@
...
+	ctx := testContext(t)
+
+	err := client.Sessions.UseRole(ctx, snowflakeroles.GlobalOrgAdmin)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		err := client.Sessions.UseRole(ctx, snowflakeroles.Accountadmin)
+		require.NoError(t, err)
+	})
+	t.Cleanup(testClientHelper().Role.UseRole(t, snowflakeroles.GlobalOrgAdmin))
```
sfc-gh-jmichalak, line 22: Nit: Can we have these calls extracted as a helper (similarly to temporarily setting envs)?

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -0,0 +1,519 @@
...
+
+	assertParameterValueSetOnAccount := func(t *testing.T, parameters []*sdk.Parameter, parameterKey string, parameterValue string) {
```
sfc-gh-jmichalak: We have similar helpers in the account integration tests. Let's extract them.

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -0,0 +1,519 @@
...
+	assertParameterIsDefault := func(t *testing.T, parameters []*sdk.Parameter, parameterKey string) {
+		t.Helper()
+		param, err := collections.FindFirst(parameters, func(parameter *sdk.Parameter) bool { return parameter.Key == parameterKey })
+		require.NoError(t, err, "parameter %v not found", parameterKey)
+		require.NotNil(t, param)
+		require.Equal(t, (*param).Default, (*param).Value)
+		require.Equal(t, sdk.ParameterType(""), (*param).Level)
+	}
+
+	assertThatPolicyIsSetOnAccount := func(t *testing.T, id sdk.SchemaObjectIdentifier) {
```
sfc-gh-jmichalak: What if we have the policy of correct name, but incorrect type?

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -0,0 +1,519 @@
...
+		require.NoError(t, err)
+
+		// Set another resource monitor without unsetting the previous one
+		err = client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithSet(*sdk.NewOrganizationAccountSetRequest().WithResourceMonitor(resourceMonitor2.ID())))
+		require.NoError(t, err)
+
+		err = client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithUnset(*sdk.NewOrganizationAccountUnsetRequest().WithResourceMonitor(true)))
+		require.NoError(t, err)
+
+		// TODO(ticket number): Currently, there's no way to query resource monitor to verify to was unset properly.
```
sfc-gh-jmichalak: ticket number?

sfc-gh-jmichalak (changes requested):
Please also make sure that the Snowflake docs are adjusted (mainly quotes/double quotes).

pkg/sdk/object_types.go
sfc-gh-jcieslak: Removed, I used Account one on the organization account object

pkg/acceptance/helpers/1_prerequisites_test.go
sfc-gh-jcieslak: It's changed in https://github.com/snowflakedb/terraform-provider-snowflake/pull/3815, not sure why those changes are here 🤔 I missed them somehow

pkg/sdk/object_types.go
sfc-gh-jcieslak: Removed this one, as in every case so far it's referred as `ACCOUNT` instead of `ORGANIZATION_ACCOUNT`

pkg/sdk/organization_accounts_def.go
sfc-gh-jcieslak: Yes

pkg/sdk/organization_accounts_def.go
sfc-gh-jcieslak: Yeah, don't remember why I needed it, but now it seems i can use this one. Changed and removed OptionalBool as it doesn't have a use case now.

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: Yup, changed

pkg/acceptance/helpers/1_prerequisites.go
sfc-gh-jcieslak: I don't think we want to error out in this case. For now, it doesn't matter that much because those tests are now run manually and it would be bad if this error would make our tests fail on CI (and it could because of the insufficient privileges).

pkg/acceptance/helpers/1_prerequisites_test.go
sfc-gh-jcieslak: it would otherwise throw compilation error because `mockOrganizationAccounts` couldn't be assigned as `OrganizationAccounts` implementation.

pkg/sdk/organization_accounts_def.go
sfc-gh-jcieslak: In comparison to account interface, we have self alters and other account alters under one Alter method. I went with the same approach here.

pkg/sdk/organization_accounts_def.go
sfc-gh-jcieslak: True, refactored

pkg/sdk/organization_accounts_def.go
sfc-gh-jcieslak: We could, but because it's a small struct I would leave it this way for the OrganizationAccount prefix consistency across operations

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: as with account case, it doesn't matter as they are all manually run. In the end, I think they should be extracted to their own package where those build tags wouldn't matter that much.

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: Done (Idk if you meant the above UseRoles but that I extracted)

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: extracted

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: generated and used objectparameters instead

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: extracted

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: very unlikely but added a kind

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: added

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jmichalak: I see, but IMO we still mark it accordingly, so that they are not left like that by accident.

pkg/acceptance/helpers/1_prerequisites.go
sfc-gh-jmichalak: Yes, but as a caller of this function, I expect that after calling, it's guaranteed that we use a non-prod account.

pkg/sdk/testint/account_commons_integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/require"
+)
+
+// SetAndUnsetAccountParametersTest is a common test used for different account kinds.
+func SetAndUnsetAccountParametersTest(
+	setParameters func(sdk.AccountParameters) error,
+	unsetAllParameters func(ctx context.Context) error,
+	showParameters func(ctx context.Context) ([]*sdk.Parameter, error),
```
sfc-gh-jmichalak, line 18: Nit: `unsetAllParameters` and `showParameters` take context but `setParameters` doesn't which is inconsistent.

sfc-gh-jmichalak (approved):

pkg/acceptance/helpers/role_client.go
```diff
@@ -25,17 +25,22 @@ func (c *RoleClient) client() sdk.Roles {
 }
 
 func (c *RoleClient) UseRole(t *testing.T, roleId sdk.AccountObjectIdentifier) func() {
+	t.Helper()
+	return c.UseRoleWithClient(t, c.context.client, roleId)
+}
+
+func (c *RoleClient) UseRoleWithClient(t *testing.T, client *sdk.Client, roleId sdk.AccountObjectIdentifier) func() {
```
sfc-gh-asawicki: client shouldnt be one of the parameters. The goal behind having testclients is that they keep the clients internally and we don't modify them

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -3,69 +3,25 @@
...
 	ctx := testContext(t)
 
-	err := client.Sessions.UseRole(ctx, snowflakeroles.GlobalOrgAdmin)
-	require.NoError(t, err)
-
-	t.Cleanup(func() {
-		err := client.Sessions.UseRole(ctx, snowflakeroles.Accountadmin)
-		require.NoError(t, err)
-	})
+	t.Cleanup(testClientHelper().Role.UseRoleWithClient(t, client, snowflakeroles.GlobalOrgAdmin))
```
sfc-gh-asawicki: we should not modify our client under test through the test helper. PLease modify it directly

pkg/acceptance/helpers/role_client.go
sfc-gh-jcieslak: removed

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: reverted back to the direct approach

pkg/acceptance/helpers/1_prerequisites.go
sfc-gh-jcieslak: Adjusted

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: changed, but still, I don't think it matters that much

pkg/sdk/testint/account_commons_integration_test.go
sfc-gh-jcieslak: changed

sfc-gh-jmichalak (dismissed):

PR #3805, opened 2025-06-24
feat: Add support for PROGRAMMATIC_ACCESS_TOKEN authenticator

sfc-gh-jmichalak:
Add support for PROGRAMMATIC_ACCESS_TOKEN authenticator

- Add `PROGRAMMATIC_ACCESS_TOKEN` as a valid authenticator option
- Update documentation to include the new authentication method
- Add acceptance tests for PAT authentication
- Fixed a typo in the migration guide

## References
* [Snowflake documentation on Programmatic Access Tokens](https://docs.snowflake.com/en/user-guide/programmatic-access-tokens)

PR #3819, opened 2025-06-27
feat: Add programmatic access token support to SDK

sfc-gh-jmichalak:
Add User Programmatic Access Token SDK Support

- Add, modify, rotate, remove, and show programmatic access token operations
- Support custom operations with returned values. Previously, only `Show...` operations were supported. Now, any operation having a show mapping is supported.
- Unsetting is not documented, but it works correctly.
- The SDK is coupled to object identifiers. PATs do not have identifiers (only names), and usual ID operations, like queries with `"A"."B"`, are not supported. That's why I decided not to introduce a new identifier for now. 
- This means that most of the operations are custom, and we don't have wrapper methods like `...Safely`. For now, they can be added to the `ext` file. Let me know what you think.

## TODO
- Add integration tests.
- Add resource.
- Implement token rotation.
- Add data source.

## References
* https://docs.snowflake.com/en/sql-reference/sql/alter-user-add-programmatic-access-token and related


sfc-gh-jmichalak:
* **#3819** <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3819?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3819?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3805** <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3805?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `dev`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

pkg/sdk/user_programmatic_access_tokens_gen_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	t.Run("with rename to", func(t *testing.T) {
+		opts := &ModifyUserProgrammaticAccessTokenOptions{
+			name:     name,
+			RenameTo: String("new_token"),
+		}
+		assertOptsValidAndSQLEquals(t, opts, `ALTER USER MODIFY PROGRAMMATIC ACCESS TOKEN "%s" RENAME TO "new_token"`, name)
+	})
+
+	t.Run("with set: all attributes", func(t *testing.T) {
```
sfc-gh-jcieslak, line 131: nit: could be with `User` field

pkg/sdk/user_programmatic_access_tokens_impl_gen.go
```diff
@@ -0,0 +1,169 @@
...
+	resultList := convertRows[addProgrammaticAccessTokenResultDBRow, AddProgrammaticAccessTokenResult](dbRows)
+	return resultList, nil
+}
+
+func (v *userProgrammaticAccessTokens) Modify(ctx context.Context, request *ModifyUserProgrammaticAccessTokenRequest) error {
+	opts := request.toOpts()
+	return validateAndExec(v.client, ctx, opts)
+}
+
+func (v *userProgrammaticAccessTokens) Rotate(ctx context.Context, request *RotateUserProgrammaticAccessTokenRequest) ([]RotateProgrammaticAccessTokenResult, error) {
```
sfc-gh-jcieslak: I'm guessing add and rotate are returning one row always. Because of that, maybe modify this or leave a TODO for later to make it a single return value (in the implementation we should then use `validateAndQueryOne`).

pkg/sdk/user_programmatic_access_tokens_validations_gen.go
```diff
@@ -0,0 +1,112 @@
...
+		return ErrNilOptions
+	}
+	var errs []error
+	// adjusted manually
+	if !ValidObjectName(opts.name) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	// adjusted manually
+	if opts.User != nil && !ValidObjectIdentifier(opts.User) {
+		errs = append(errs, fmt.Errorf("invalid user: %s", opts.User.FullyQualifiedName()))
```
sfc-gh-jcieslak: nit: we have dedicated error function for invalid identifiers

sfc-gh-jcieslak (approved):

pkg/sdk/user_programmatic_access_tokens_def.go
```diff
@@ -0,0 +1,161 @@
...
+var rotateProgrammaticAccessTokenResultDef = g.PlainStruct("RotateProgrammaticAccessTokenResult").
+	Text("TokenName").
+	Text("TokenSecret").
+	Text("RotatedTokenName")
+
+var UserProgrammaticAccessTokensDef = g.NewInterface(
+	"UserProgrammaticAccessTokens",
+	"UserProgrammaticAccessToken",
+	g.KindOfT[AccountObjectIdentifier](),
+).CustomShowOperation(
```
sfc-gh-asawicki, line 85: I am not convinced whether we should extract an SDK interface for PAT handling. Why? You already mentioned a few reasons in the PR description: they don't have id, all operations are custom. Additionally, they are all connected with USER on Snowflake side. For other objects, we put such methods inside the target object's interface (in this case `Users`). Remember, this is the SDK for Snowflake which should be aligned with Snowflake (even if we pick to make user and PAT objects separated on the tf level).
There are some pros of extraction, however, the main disadvantage I see here is inconsistency with most of the existing objects.
Let's discuss this @sfc-gh-jcieslak @sfc-gh-jmichalak .

pkg/sdk/user_programmatic_access_tokens_def.go
```diff
@@ -0,0 +1,161 @@
...
+	"Add",
+	"https://docs.snowflake.com/en/sql-reference/sql/alter-user-add-programmatic-access-token",
+	addProgrammaticAccessTokenResultDBRowDef,
+	addProgrammaticAccessTokenResultDef,
+	g.NewQueryStruct("AddUserProgrammaticAccessToken").
+		Alter().
+		SQL("USER").
+		IfExists().
+		OptionalIdentifier("User", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions()).
```
sfc-gh-asawicki: optional? let's make it required (not putting an identifier here interacts with the current user and we can choose to be stricter here - stricter and more consistent with other SDK objects)

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
```diff
@@ -3,7 +3,7 @@
 {{ $impl := .NameLowerCased }}
 
 {{ range .Operations }}
-    {{ if and (hasPrefix .Name "Show") .ShowMapping }}
+    {{ if .ShowMapping }}
```
sfc-gh-asawicki, line 6: question: have this change affected any other objects generation?

pkg/sdk/poc/generator/templates/interface.tmpl
```diff
@@ -4,7 +4,7 @@ import "context"
 
 type {{ .Name }} interface {
 	{{- range .Operations }}
-		{{- if and (hasPrefix .Name "Show") .ShowMapping }}
+		{{- if .ShowMapping }}
```
sfc-gh-asawicki, line 7: question: have this change affected any other objects generation?

pkg/sdk/user_programmatic_access_tokens_impl_gen.go
```diff
@@ -0,0 +1,169 @@
...
+	"log"
+)
+
+var _ UserProgrammaticAccessTokens = (*userProgrammaticAccessTokens)(nil)
+
+type userProgrammaticAccessTokens struct {
+	client *Client
+}
+
+func (v *userProgrammaticAccessTokens) Add(ctx context.Context, request *AddUserProgrammaticAccessTokenRequest) ([]AddProgrammaticAccessTokenResult, error) {
```
sfc-gh-asawicki: Is it ever a multi-row response? If not, I would like all these functions to return a single row

pkg/sdk/user_programmatic_access_tokens_validations_gen.go
```diff
@@ -0,0 +1,112 @@
...
+	_ validatable = new(ShowUserProgrammaticAccessTokenOptions)
+)
+
+func (opts *AddUserProgrammaticAccessTokenOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	// adjusted manually
+	if !ValidObjectName(opts.name) {
```
sfc-gh-asawicki: TBH I don't get why we are using a new ValidObjectName function here instead of going with the existing one for identifiers. Underneath, it's still an identifier, and even if it's optional (however, I left a comment that IMO it shouldn't), this could be handled the old way

sfc-gh-asawicki (changes requested):
Please add to the PR description that integration tests will be added separately.

pkg/sdk/user_programmatic_access_tokens_gen_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/poc/generator/templates/sub_templates/implementation_functions.tmpl
sfc-gh-jmichalak: No, because `ShowMapping` was only used 
- in `ShowOperation`, which forces `Show` prefix
- in `CustomOperation`, which was only used in git repositories with `ShowGitTags` and `ShowGitBranches` ops.
I checked the generation of git repositories, and it works fine.

pkg/sdk/poc/generator/templates/interface.tmpl
sfc-gh-jmichalak: ditto

pkg/sdk/user_programmatic_access_tokens_impl_gen.go
sfc-gh-jmichalak: Done. We can do it in SNOW-2183036 with a more complex approach.

pkg/sdk/user_programmatic_access_tokens_validations_gen.go
sfc-gh-jmichalak: Done.

pkg/sdk/user_programmatic_access_tokens_def.go
sfc-gh-jmichalak: Done.

pkg/sdk/user_programmatic_access_tokens_impl_gen.go
sfc-gh-jmichalak: Done.

pkg/sdk/user_programmatic_access_tokens_validations_gen.go
sfc-gh-jmichalak: Because the PAT is not an identifier. After the discussion, we decided to use AccountObjectIdentifier as a workaround (done).

pkg/sdk/user_programmatic_access_tokens_def.go
sfc-gh-jmichalak: As we discussed offline:
- Currently, the user SDK does not use a generator.
- This PAT required some changes in the generator to support queries returning a single row, but the change is not that complex.
-  PAT names can be handled like AccountObjectIdentifier.
- I added user methods that simply call the methods from the PAT client. 
- We can do a more systematic approach later (I created and linked tickets).

PR #3824, opened 2025-06-30
chore: Configure plugin framework in functional tests

sfc-gh-asawicki:
- Fix after review
- Regenerate plugin framewrok schema
- Create new terraform plugin framework based provider for functional tests


pkg/testfunctional/test_resource_plugin_framework_example.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+func (r *SomeResource) Create(ctx context.Context, request resource.CreateRequest, response *resource.CreateResponse) {
+	var data *someResourceModelV0
+	response.Diagnostics.Append(request.Plan.Get(ctx, &data)...)
+
+	name := data.Name.ValueString()
+	id := sdk.NewAccountObjectIdentifier(name)
+	data.Id = types.StringValue(id.FullyQualifiedName())
```
sfc-gh-jcieslak, line 54: question: does this id has anything to do with old ids that are set with `d.SetId`?

sfc-gh-jcieslak (approved):

pkg/testfunctional/test_resource_plugin_framework_example.go
sfc-gh-asawicki: It appears so, I will be testing it.

sfc-gh-jmichalak (approved):

PR #3830, opened 2025-07-02
chore: Add PAT integration tests

sfc-gh-jmichalak:
Add User Programmatic Access Token integration tests

- Add integration tests.
- Generate assertions.
- Add a new error for PAT not found.
- Add a new function `ShowByName` that behaves similarly to ShowById.
- Add and adjust user helpers.
- Implement and test safe functions.
- Adjust a validation.

## TODO
- Add resource.
- Implement token rotation.
- Add data source.

## References
* https://docs.snowflake.com/en/sql-reference/sql/alter-user-add-programmatic-access-token and related


pkg/acceptance/bettertestspoc/assert/objectassert/programmatic_access_token_snowflake_gen.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+type ProgrammaticAccessTokenAssert struct {
+	*assert.SnowflakeObjectAssert[sdk.ProgrammaticAccessToken, sdk.AccountObjectIdentifier]
+}
+
+// function ProgrammaticAccessToken is not supported because ProgrammaticAccessToken has a pseudo-identifier and requires a user ID for being listed in Snowflake.
+
+// TODO: check if it makes sense.
```
sfc-gh-asawicki: What about this TODO?

pkg/acceptance/bettertestspoc/assert/objectassert/programmatic_access_token_snowflake_gen.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+type ProgrammaticAccessTokenAssert struct {
+	*assert.SnowflakeObjectAssert[sdk.ProgrammaticAccessToken, sdk.AccountObjectIdentifier]
+}
+
+// function ProgrammaticAccessToken is not supported because ProgrammaticAccessToken has a pseudo-identifier and requires a user ID for being listed in Snowflake.
```
sfc-gh-asawicki, line 18: I would like all our assertions/models to be verified for regeneration and no changes soon, so please add a TODO with SNOW-1501905 and add it to Better tests poc README.

pkg/acceptance/helpers/user_client.go
```diff
@@ -192,17 +193,34 @@ func (c *UserClient) UnsetDefaultSecondaryRoles(t *testing.T, id sdk.AccountObje
...
-	return *token
+	return *token, c.RemoveProgrammaticAccessTokenFunc(t, userId, sdk.NewAccountObjectIdentifier(token.TokenName))
+}
+
+func (c *UserClient) RemoveProgrammaticAccessTokenFunc(t *testing.T, userId sdk.AccountObjectIdentifier, tokenName sdk.AccountObjectIdentifier) func() {
+	t.Helper()
+	ctx := context.Background()
+
+	return func() {
+		err := c.context.client.Users.RemoveProgrammaticAccessToken(ctx, sdk.NewRemoveUserProgrammaticAccessTokenRequest(userId, tokenName))
```
sfc-gh-asawicki: What about the situation when the user is missing? What error we will receive then? Could we use a safe variant here?

pkg/sdk/errors.go
```diff
@@ -23,6 +23,7 @@ var (
 	ErrObjectNotExistOrAuthorized               = NewError("object does not exist or not authorized")
 	ErrAccountIsEmpty                           = NewError("account is empty")
 	ErrGrantPartiallyExecuted                   = NewError("grant partially executed")
+	ErrNotFound                                 = NewError("not found")
```
sfc-gh-asawicki: Is it really just "not found"? Can we add a test where we check for this error in errors_integration_test?

The error handling is not the greatest one currently in the SDK, so I am skeptical about new errors being added (I just want us to be sure we need it). Because, e.g. on the resource level, we need to handle a different error than usual potentially, right?

pkg/sdk/object_types.go
```diff
@@ -16,70 +16,71 @@ type Object struct {
...
+	ObjectTypeAggregationPolicy       ObjectType = "AGGREGATION POLICY"
+	ObjectTypeAuthenticationPolicy    ObjectType = "AUTHENTICATION POLICY"
+	ObjectTypeHybridTable             ObjectType = "HYBRID TABLE"
+	ObjectTypeImageRepository         ObjectType = "IMAGE REPOSITORY"
+	ObjectTypeProjectionPolicy        ObjectType = "PROJECTION POLICY"
+	ObjectTypeDataMetricFunction      ObjectType = "DATA METRIC FUNCTION"
+	ObjectTypeGitRepository           ObjectType = "GIT REPOSITORY"
+	ObjectTypeModel                   ObjectType = "MODEL"
+	ObjectTypeService                 ObjectType = "SERVICE"
+	ObjectTypeProgrammaticAccessToken ObjectType = "PROGRAMMATIC ACCESS TOKEN"
```
sfc-gh-asawicki: I have left similar comment on some other recent changes (I guess listings): there are other methods and collections in this file (e.g. plural). When we will add it there and verify this objects behavior in places where all these functions/collections are used?

pkg/sdk/operation_helpers.go
```diff
@@ -113,3 +139,29 @@ func SafeShowById[T any, ID ObjectIdentifierConstraint](
 
 	return result, nil
 }
+
+// SafeShowProgrammaticAccessTokenByName is a helper function with specific implementation for PATs.
+func SafeShowProgrammaticAccessTokenByName(
+	client *Client,
+	showByName func(ctx context.Context, userName AccountObjectIdentifier, tokenName AccountObjectIdentifier) (*ProgrammaticAccessToken, error),
```
sfc-gh-asawicki: nit: userName -> userId

pkg/sdk/operation_helpers.go
```diff
@@ -113,3 +139,29 @@ func SafeShowById[T any, ID ObjectIdentifierConstraint](
...
+	ctx context.Context,
+	userName AccountObjectIdentifier,
+	tokenName AccountObjectIdentifier,
+) (*ProgrammaticAccessToken, error) {
+	result, err := showByName(ctx, userName, tokenName)
+
+	// ErrObjectNotExistOrAuthorized or ErrDoesNotExistOrOperationCannotBePerformed can only happen
+	// when the user object is not accessible for some reason during the "main" showById.
+	shouldCheckHigherHierarchies := errors.Is(err, ErrObjectNotExistOrAuthorized) || errors.Is(err, ErrDoesNotExistOrOperationCannotBePerformed)
+	if errors.Is(err, ErrObjectNotFound) || !shouldCheckHigherHierarchies {
```
sfc-gh-asawicki, line 159: question: so this new error (ErrNotFound) is only in drop?

pkg/sdk/testint/drop_integration_test.go
```diff
@@ -88,3 +88,25 @@ func TestInt_SafeDropOnSchemaObjectIdentifierWithArguments(t *testing.T) {
...
+	removeProgrammaticAccessToken := func(userId, tokenId sdk.AccountObjectIdentifier) func() error {
+		return func() error {
+			return testClient(t).Users.RemoveProgrammaticAccessToken(ctx, sdk.NewRemoveUserProgrammaticAccessTokenRequest(userId, tokenId))
+		}
+	}
+
+	err := sdk.SafeRemoveProgrammaticAccessToken(testClient(t), removeProgrammaticAccessToken(user.ID(), token.ID()), ctx, user.ID())
+	assert.NoError(t, err)
+
+	invalidUserId := NonExistingAccountObjectIdentifier
```
sfc-gh-asawicki, line 108: how about removal of incorrect tokenId on the correct user?

pkg/sdk/testint/show_by_id_integration_test.go
```diff
@@ -127,3 +127,33 @@ func TestInt_SafeShowByIdOnSchemaObjectIdentifierWithArguments(t *testing.T) {
...
+	assert.NotNil(t, value)
+	assert.NoError(t, err)
+
+	cleanupToken()
+
+	_, err = sdk.SafeShowProgrammaticAccessTokenByName(testClient(t), tokenShowByName, testContext(t), user.ID(), token.ID())
+	assert.Error(t, err)
+	assert.ErrorIs(t, err, sdk.ErrObjectNotFound)
+
+	invalidUserId := testClientHelper().Ids.RandomAccountObjectIdentifier()
```
sfc-gh-asawicki: let's split invalidUser and invalidToken into two SafeShowProgrammaticAccessTokenByName invocations and assertions

pkg/sdk/user_programmatic_access_tokens_impl_gen.go
```diff
@@ -49,6 +59,21 @@ func (v *userProgrammaticAccessTokens) Show(ctx context.Context, request *ShowUs
 	return resultList, nil
 }
 
+// Adjusted manually to include the user id in the request.
```
sfc-gh-asawicki, line 59: let's document these limitations in readme (if not documented already, as I recall you adding there something around PAT already)

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInt_UserProgrammaticAccessToken(t *testing.T) {
+	client := testClient(t)
+	ctx := testContext(t)
+
+	currentUser := testClientHelper().Context.CurrentUser(t)
```
sfc-gh-asawicki: nit: this current user should be run on `client` as potentially these are different users

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
```diff
@@ -0,0 +1,282 @@
...
+		// Assert the token is shown in the SHOW command.
+		tokenShowObject, err := client.Users.ShowProgrammaticAccessTokenByName(ctx, user.ID(), id)
+		require.NoError(t, err)
+		require.NotNil(t, tokenShowObject)
+
+		assertThatObject(t, objectassert.ProgrammaticAccessTokenFromObject(t, tokenShowObject).
+			HasName(id.Name()).
+			HasUserName(user.ID()).
+			HasRoleRestriction(snowflakeroles.Public).
+			// Assert that WithDaysToExpiry(1) takes effect.
```
sfc-gh-asawicki: Can we do it in this PR? We can e.g. set it to two and check that it's greater than 1 in the ext assertion?

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
```diff
@@ -0,0 +1,282 @@
...
+		t.Cleanup(cleanupToken)
+
+		rotateRequest := sdk.NewRotateUserProgrammaticAccessTokenRequest(user.ID(), token.ID()).
+			WithExpireRotatedTokenAfterHours(0)
+		rotateResult, err := client.Users.RotateProgrammaticAccessToken(ctx, rotateRequest)
+		require.NoError(t, err)
+		t.Cleanup(testClientHelper().User.RemoveProgrammaticAccessTokenFunc(t, user.ID(), sdk.NewAccountObjectIdentifier(rotateResult.RotatedTokenName)))
+
+		// Assert the values returned during token rotation.
+		assert.NotNil(t, rotateResult)
```
sfc-gh-asawicki, line 201: nit (some day): we could add assertions for rotateresult

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
```diff
@@ -0,0 +1,282 @@
...
+		require.ErrorIs(t, err, sdk.ErrObjectNotFound)
+	})
+
+	t.Run("show", func(t *testing.T) {
+		token1, cleanupToken1 := testClientHelper().User.AddProgrammaticAccessToken(t, user.ID())
+		t.Cleanup(cleanupToken1)
+		token2, cleanupToken2 := testClientHelper().User.AddProgrammaticAccessToken(t, user.ID())
+		t.Cleanup(cleanupToken2)
+
+		showRequest := sdk.NewShowUserProgrammaticAccessTokenRequest().WithUserName(user.ID())
```
sfc-gh-asawicki, line 267: question: I see that there are no filters in the docs, have you tried running them?

pkg/sdk/operation_helpers.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/show_by_id_integration_test.go
sfc-gh-jmichalak: It was already checked before - slightly adjusted.

pkg/acceptance/bettertestspoc/assert/objectassert/programmatic_access_token_snowflake_gen.go
sfc-gh-jmichalak: Removed.

pkg/acceptance/bettertestspoc/assert/objectassert/programmatic_access_token_snowflake_gen.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/drop_integration_test.go
sfc-gh-jmichalak: Added.

pkg/acceptance/helpers/user_client.go
sfc-gh-jmichalak: Yes, done. Check `TestInt_SafeRemoveProgrammaticAccessToken` to see the behavior.

pkg/sdk/object_types.go
sfc-gh-jmichalak: Added; with a small comment.

pkg/sdk/operation_helpers.go
sfc-gh-jmichalak: It's returned from all operations that require token names. SHOW does not - it returns all keys, and we filter them on our side with `errors.ErrObjectNotFound`.

pkg/sdk/user_programmatic_access_tokens_impl_gen.go
sfc-gh-jmichalak: Done (I made it more specific).

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
sfc-gh-jmichalak: Yes, they don't work.

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
sfc-gh-jmichalak: WDYM? This isn't a TODO; we already assert this. Added further explanation.

pkg/sdk/errors.go
sfc-gh-jmichalak: Snowflake returns `Programmatic access token S2T not found.` on all operations except SHOW. I agree this is not the ideal solution, but it works for now.

We already handle this behavior in `Safely` functions, which we'll use in the resource.

Added integration error tests.

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
sfc-gh-jmichalak: Added a TODO to the betterttestspoc readme.

pkg/sdk/errors.go
sfc-gh-asawicki: Okey, leaving this open so @sfc-gh-jcieslak can also read it during the review.

sfc-gh-asawicki (dismissed):

pkg/sdk/errors.go
sfc-gh-jcieslak: Can we make it more specific then? We're mapping based on "Contains" in error logic and I'm afraid that we could map more than we should to this error. Let's maybe make it more specific like ErrPatNotFound or ErrPatS2TNotFound and add more text into mapping below (it can be whole "Programmatic access token S2T not found" if that's what is returned for all operations except show).

pkg/sdk/object_types.go
```diff
@@ -80,6 +80,9 @@ const (
 	ObjectTypeGitRepository        ObjectType = "GIT REPOSITORY"
 	ObjectTypeModel                ObjectType = "MODEL"
 	ObjectTypeService              ObjectType = "SERVICE"
+	// ObjectTypeProgrammaticAccessToken is a pseudo-object, as it does not support the usual operations in Snowflake, but it is handled by user functions.
+	// Programmatic access tokens do not have grants and cannot be tagged.
+	ObjectTypeProgrammaticAccessToken ObjectType = "PROGRAMMATIC ACCESS TOKEN" //nolint:gosec
```
sfc-gh-jcieslak, line 85: Is the object type of this kind useful, or are we just adding it for consistency? I don't think we could use it in a typical use case for object type which is grants and filtering by object type, etc.

pkg/sdk/operation_helpers.go
```diff
@@ -56,6 +56,37 @@ func SafeDrop[ID ObjectIdentifierConstraint](
...
+	userId AccountObjectIdentifier,
+) error {
+	err := removeProgrammaticAccessToken()
+
+	// Tokens can't be removed with IF EXISTS, so we return nil if the token is not found.
+	if errors.Is(err, ErrNotFound) {
+		return nil
+	}
+
+	// ErrObjectNotExistOrAuthorized can only happen
```
sfc-gh-jcieslak: nit: `or ErrDoesNotExistOrOperationCannotBePerformed`

pkg/sdk/operation_helpers.go
```diff
@@ -56,6 +56,37 @@ func SafeDrop[ID ObjectIdentifierConstraint](
...
+	// when the user object is not accessible for some reason during the "main" removeProgrammaticAccessToken.
+	shouldCheckHigherHierarchies := errors.Is(err, ErrObjectNotExistOrAuthorized) || errors.Is(err, ErrDoesNotExistOrOperationCannotBePerformed)
+	if !shouldCheckHigherHierarchies {
+		return err
+	}
+	if err != nil {
+		if _, err := client.Users.ShowByID(ctx, userId); err != nil {
```
sfc-gh-jcieslak: Use ShowByIDSafely

pkg/sdk/operation_helpers.go
```diff
@@ -113,3 +144,29 @@ func SafeShowById[T any, ID ObjectIdentifierConstraint](
...
+
+	// ErrObjectNotExistOrAuthorized or ErrDoesNotExistOrOperationCannotBePerformed can only happen
+	// when the user object is not accessible for some reason during the "main" showById.
+	shouldCheckHigherHierarchies := errors.Is(err, ErrObjectNotExistOrAuthorized) || errors.Is(err, ErrDoesNotExistOrOperationCannotBePerformed)
+	if errors.Is(err, ErrObjectNotFound) || !shouldCheckHigherHierarchies {
+		return result, err
+	}
+	if err != nil {
+		errs := []error{err}
+		if _, err := client.Users.ShowByID(ctx, userId); err != nil {
```
sfc-gh-jcieslak: Use ShowByIDSafely

pkg/sdk/testint/drop_integration_test.go
```diff
@@ -88,3 +88,28 @@ func TestInt_SafeDropOnSchemaObjectIdentifierWithArguments(t *testing.T) {
...
+	t.Cleanup(cleanupToken)
+
+	ctx := context.Background()
+	removeProgrammaticAccessToken := func(userId, tokenId sdk.AccountObjectIdentifier) func() error {
+		return func() error {
+			return testClient(t).Users.RemoveProgrammaticAccessToken(ctx, sdk.NewRemoveUserProgrammaticAccessTokenRequest(userId, tokenId))
+		}
+	}
+
+	err := sdk.SafeRemoveProgrammaticAccessToken(testClient(t), removeProgrammaticAccessToken(user.ID(), token.ID()), ctx, user.ID())
```
sfc-gh-jcieslak: I kind of don't get why we are passing a function here. Can't we make something similar to the ShowByIDSafely, so that the interface is similar to the underlying function, but it does additional stuff? This way, it's harder to guess what we are expecting to be passed there (without looking at its usage).

Additionally, why are they not part of the PAT interface, but rather added as separate top-level functions?

pkg/sdk/operation_helpers.go
sfc-gh-jmichalak: Done.

pkg/sdk/operation_helpers.go
sfc-gh-jmichalak: Done. But we don't use the `Safely` function in `SafeDrop`.

pkg/sdk/operation_helpers.go
sfc-gh-jmichalak: Done. But we don't use the `Safely` function in `SafeDrop`.

pkg/sdk/object_types.go
sfc-gh-jmichalak: As mentioned in the comment, it's added for consistency. We don't need to add it to grants handling or filtering if we don't want to support it.

pkg/sdk/testint/drop_integration_test.go
sfc-gh-jmichalak: 1. Done. It was actually following the original Drop/Safe functions.
2. I don't understand. They are not separate top-level functions. As we discussed before, you can call them like `client.UserProgrammaticAccessTokens.ShowByIDSafely`. We're testing them here to be consistent with other `...Safely` functions. Do you want me to inline them? Let's continue in the next PR.

pkg/sdk/errors.go
sfc-gh-jmichalak: This matcher (`not found`) doesn't overlap with any existing matches, so it will not break. Let's leave this for the error rework. I don't want to rework the error matching for now.

sfc-gh-asawicki: TBH, I agree that we can be more specific with this mapping. If the response is `Programmatic access token S2T not found` let's include it whole. There might be some other errors later that also use the not found but are different errors. THis is not reworking the error mapping, it's just changing the mapper for this one

sfc-gh-jmichalak: To be clear: `S2T` is the token name in this case. I added a regex matcher and add error matcher unit tests.

pkg/sdk/testint/drop_integration_test.go
sfc-gh-jcieslak: Generally, I think DropSafe and these functions shouldn't be exported, but this is future improvement not important for now.

sfc-gh-jcieslak (approved):

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak:
### Merge activity

* **Jul 14, 9:42 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3830).
* **Jul 14, 9:42 AM UTC**: @sfc-gh-jmichalak merged this pull request with [Graphite](https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3830).

PR #3838, opened 2025-07-03
feat: Add a user_programmatic_access_tokens data source

sfc-gh-jmichalak:
Add a new data source for user programmatic access tokens.

## TODO
- Support rotating keys in the resource.


pkg/datasources/user_programmatic_access_tokens.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var userProgrammaticAccessTokensSchema = map[string]*schema.Schema{
+	"for_user": {
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      "Returns programmatic access tokens for the specified user.",
+		ValidateDiagFunc: resources.IsValidIdentifier[sdk.AccountObjectIdentifier](),
```
sfc-gh-jcieslak, line 23: nit: same validation could be added to the resource

pkg/testacc/data_source_user_programmatic_access_tokens_acceptance_test.go
```diff
@@ -0,0 +1,66 @@
...
+		WithDependsOn(modelComplete.ResourceReference())
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
```
sfc-gh-jcieslak, line 65: Let's have at least the second step with two tokens

MIGRATION_GUIDE.md
```diff
@@ -23,6 +23,11 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ## v2.3.0 ➞ v2.4.0
 
+### *(new feature)* snowflake_user_programmatic_access_tokens data source
```
sfc-gh-asawicki: nit: let's start thinking from the user's perspective for these entries. The new functionality we add is PAT token handling. We can mention that authentication was added in 2.3.0, and here we add: resource, data source, and the rotation capabilities with examples. Each can have it's small subsection, grouping them nicely. WDYT? It can be changed with the last PR, not necessarily now.

pkg/testacc/data_source_user_programmatic_access_tokens_acceptance_test.go
sfc-gh-asawicki: +1

pkg/datasources/user_programmatic_access_tokens.go
sfc-gh-jmichalak: Done.

MIGRATION_GUIDE.md

pkg/testacc/data_source_user_programmatic_access_tokens_acceptance_test.go
sfc-gh-jmichalak: Done. Also, added a test for a different user.

sfc-gh-asawicki (dismissed):

sfc-gh-jcieslak (dismissed):

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
```diff
@@ -97,7 +93,173 @@ func TestInt_UserProgrammaticAccessToken(t *testing.T) {
...
+		require.NoError(t, err)
+
+		tokenShowObject := testClientHelper().User.ShowProgrammaticAccessToken(t, user.ID(), id)
+
+		assertThatObject(t, objectassert.ProgrammaticAccessTokenFromObject(t, tokenShowObject).
+			HasName(id.Name()).
+			HasUserName(user.ID()).
+			HasNoRoleRestriction().
+			HasExpiresAtNotEmpty().
+			HasStatus(sdk.ProgrammaticAccessTokenStatusDisabled).
```
sfc-gh-jcieslak: nit: This test could be a bit shorter. Instead of using object asserts fromObject you could let it do the object retrieving, and I don't think we need all assertions (e.g. RoleRestrictions, ExpiresAt, etc.) when we're focusing on disabled property.

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
```diff
@@ -97,7 +93,173 @@ func TestInt_UserProgrammaticAccessToken(t *testing.T) {
...
+			HasNoComment().
+			HasCreatedOnNotEmpty().
+			HasCreatedBy(currentUser.Name()).
+			HasNoMinsToBypassNetworkPolicyRequirement().
+			HasRotatedToEmpty(),
+		)
+
+		unsetRequest := sdk.NewModifyUserProgrammaticAccessTokenRequest(user.ID(), id).
```
sfc-gh-jcieslak, line 120: nit: personal preference, inlining would make this a bit more readable, because firstly you see what method is being called (Modify) and then with what parameters (WithDisabled(true)), also a bit less space is used.

pkg/resources/user_programmatic_access_token.go
```diff
@@ -260,25 +260,50 @@ func UpdateUserProgrammaticAccessToken(ctx context.Context, d *schema.ResourceDa
...
+			if err := client.Users.ModifyProgrammaticAccessToken(ctx, sdk.NewModifyUserProgrammaticAccessTokenRequest(resourceId.userName, resourceId.tokenName).WithUnset(*sdk.NewModifyProgrammaticAccessTokenUnsetRequest().WithDisabled(true))); err != nil {
+				return diag.FromErr(err)
+			}
+		}
 	}
 
-	if (*set != sdk.ModifyProgrammaticAccessTokenSetRequest{}) {
-		if err := client.Users.ModifyProgrammaticAccessToken(ctx, sdk.NewModifyUserProgrammaticAccessTokenRequest(resourceId.userName, resourceId.tokenName).WithSet(*set)); err != nil {
-			return diag.FromErr(err)
+	if d.HasChange("comment") {
```
sfc-gh-asawicki, line 283: nit: do we need to split into 3 completely different requests (we could just split into disabled and rest of them)

pkg/resources/user_programmatic_access_token.go
```diff
@@ -260,25 +260,50 @@ func UpdateUserProgrammaticAccessToken(ctx context.Context, d *schema.ResourceDa
...
-	set, unset := sdk.NewModifyProgrammaticAccessTokenSetRequest(), sdk.NewModifyProgrammaticAccessTokenUnsetRequest()
-	errs := errors.Join(
-		// role_restriction and days_to_expiry are handled by forcenew
-		intAttributeUpdate(d, "mins_to_bypass_network_policy_requirement", &set.MinsToBypassNetworkPolicyRequirement, &unset.MinsToBypassNetworkPolicyRequirement),
-		booleanStringAttributeUpdate(d, "disabled", &set.Disabled, &unset.Disabled),
-		stringAttributeUpdate(d, "comment", &set.Comment, &unset.Comment),
-	)
-	if errs != nil {
-		return diag.FromErr(errs)
+	if d.HasChange("disabled") {
```
sfc-gh-asawicki, line 264: if we split like this, we still should use d.Partial

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
sfc-gh-asawicki: +1

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
```diff
@@ -97,7 +93,173 @@ func TestInt_UserProgrammaticAccessToken(t *testing.T) {
...
+			HasStatus(sdk.ProgrammaticAccessTokenStatusActive).
+			HasNoComment().
+			HasCreatedOnNotEmpty().
+			HasCreatedBy(currentUser.Name()).
+			HasNoMinsToBypassNetworkPolicyRequirement().
+			HasRotatedToEmpty(),
+		)
+	})
+
+	t.Run("modify - setting all fields does not change all object values", func(t *testing.T) {
```
sfc-gh-asawicki, line 215: please add a TODO here with the issue number (ours, not the other team's one), to change the behavior when it's fixed, the same in the resource

pkg/resources/user_programmatic_access_token.go
sfc-gh-jmichalak: No, but I prefer to keep all of these separated, just in case.

pkg/resources/user_programmatic_access_token.go
sfc-gh-jmichalak: Done. But I'm confused, why don't we use `d.Partial` after sets in each resource?

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
sfc-gh-jmichalak: But this is the first you see - `NewModifyUserProgrammaticAccessTokenRequest` means that `MODIFY` will be called.

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
sfc-gh-jmichalak: Removed some assertions. But as I said before, we can't use `objectassert.ProgrammaticAccessToken` for now because the logic is coupled to having one object ID, but we also use User ID here. We'd only save one line of code, literally.

pkg/sdk/testint/user_programmatic_access_token_integration_test.go
sfc-gh-jmichalak: Done.

pkg/resources/user_programmatic_access_token.go
sfc-gh-asawicki: Let's tak on Monday

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

PR #3843, opened 2025-07-07
feat: Implement rotating pats

sfc-gh-jmichalak:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


MIGRATION_GUIDE.md
```diff
@@ -25,7 +25,7 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ### *(new feature)* Handling Programmatic Access Tokens
 As we announced in our [roadmap](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/ROADMAP.md#pat-support), we implemented handling Programmatic Access Tokens (PATs) in the provider. In [v2.3.0](#v220--v230), we already added `PROGRAMMATIC_ACCESS_TOKEN` authenticator option.
-In this version, we enhanced the provider capabilities with handling PATs in a new resource and data source.
+In this version, we enhanced the provider capabilities with handling PATs in a new resource and data source. See more in our [Authentication Methods guide](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/guides/authentication_methods#pat-personal-access-token).
```
sfc-gh-asawicki: nit: let;s link to `2.4` and not `latest`

MIGRATION_GUIDE.md
```diff
@@ -35,6 +35,8 @@ This feature will be marked as a stable feature in future releases. Breaking cha
 #### New `snowflake_user_programmatic_access_token` resource
 Added a new preview resource for managing users' programmatic access tokens. See reference [docs](https://docs.snowflake.com/en/sql-reference/sql/alter-user-add-programmatic-access-token) and a [user guide](https://docs.snowflake.com/en/user-guide/programmatic-access-tokens) for more details.
 
+This resource also supports token rotation. See our [Authentication Methods guide](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/guides/authentication_methods#managing-pats) and the [resource documentation](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/user_programmatic_access_token).
```
sfc-gh-asawicki: as above

docs/guides/authentication_methods.md
```diff
@@ -91,8 +91,40 @@ You must fulfill the following prerequisites to generate and use programmatic ac
...
+Managing the whole PAT lifecycle is implemented in the provider with the [snowflake_programmatic_access_token](../resources/user_programmatic_access_token) resource.
+For example, to create a new PAT, you can use the following code:
+```terraform
+resource "snowflake_user_programmatic_access_token" "example" {
+  user = snowflake_user.user.name
+  name = "TOKEN"
+}
+```
+
+You can also rotate the PAT by using the `keepers` attribute and connecting it to a resource providing changing a value, like [time_rotating](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/rotating) resource from the `time` provider.
```
sfc-gh-asawicki: let's add a bit more, like: after 30 days passes, you will see terraform plan output similar to: ...

docs/resources/user_programmatic_access_token.md
```diff
@@ -68,13 +70,38 @@ resource "snowflake_user_programmatic_access_token" "complete_with_external_refe
   mins_to_bypass_network_policy_requirement = 10
   disabled                                  = false
   comment                                   = "COMMENT"
+
+  keepers = {
+    rotation_time = time_rotating.my_token_rotation.rotation_rfc3339
```
sfc-gh-asawicki: rotation_rfc3339 - let's add a comment why rotation rfc for clarity (the same in authentication guide)

docs/resources/user_programmatic_access_token.md
```diff
@@ -90,17 +117,20 @@ output "token" {
...
 - `disabled` (String) (Default: fallback to Snowflake default - uses special value that cannot be set in the configuration manually (`default`)) Disables or enables the programmatic access token. Available options are: "true" or "false". When the value is not set in the configuration the provider will put "default" there which means to use the Snowflake default for this value.
+- `expire_rotated_token_after_hours` (Number) (Default: fallback to Snowflake default - uses special value that cannot be set in the configuration manually (`-1`)) Sets the expiration time of the existing token secret to expire after the specified number of hours. You can set this to a value of 0 to expire the current token secret immediately. This field is only used when the token is rotated with `keepers` field.
+- `keepers` (Map of String) Arbitrary map of values that, when changed, will trigger a new key to be generated.
```
sfc-gh-asawicki: question: why do we implement it like as an arbitrary map and not the dedicated field?

docs/resources/user_programmatic_access_token.md
```diff
@@ -90,17 +117,20 @@ output "token" {
...
 - `mins_to_bypass_network_policy_requirement` (Number) The number of minutes during which a user can use this token to access Snowflake without being subject to an active network policy. External changes for this field won't be detected. In case you want to apply external changes, you can re-create the resource manually using "terraform taint".
 - `role_restriction` (String) The name of the role used for privilege evaluation and object creation. This must be one of the roles that has already been granted to the user. Due to technical limitations (read more [here](../guides/identifiers_rework_design_decisions#known-limitations-and-identifier-recommendations)), avoid using the following characters: `|`, `.`, `"`.
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
 
 ### Read-Only
 
 - `id` (String) The ID of this resource.
+- `rotated_token_name` (String) Name of the token that represents the prior secret. This field is updated only when the token is rotated.
```
sfc-gh-asawicki: question: does it come from Snowflake? Should we make it a list o last keys?

pkg/acceptance/bettertestspoc/assert/resourceassert/user_programmatic_access_token_resource_gen.go
```diff
@@ -52,6 +52,16 @@ func (u *UserProgrammaticAccessTokenResourceAssert) HasDisabledString(expected s
 	return u
 }
 
+func (u *UserProgrammaticAccessTokenResourceAssert) HasExpireRotatedTokenAfterHoursString(expected string) *UserProgrammaticAccessTokenResourceAssert {
+	u.AddAssertion(assert.ValueSet("expire_rotated_token_after_hours", expected))
+	return u
+}
+
+func (u *UserProgrammaticAccessTokenResourceAssert) HasKeepersString(expected string) *UserProgrammaticAccessTokenResourceAssert {
```
sfc-gh-asawicki: question: why is it generated here as string? It should have an outcome that maps are not supported

pkg/resources/user_programmatic_access_token.go
```diff
@@ -61,25 +61,34 @@ var userProgrammaticAccessTokenSchema = map[string]*schema.Schema{
...
-	// 	Optional:         true,
-	// 	Description:      "Sets the expiration time of the existing token secret to expire after the specified number of hours. You can set this to a value of 0 to expire the current token secret immediately.",
-	// 	DiffSuppressFunc: IgnoreAfterCreation,
-	// ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(0)),
-	// },
+	"expire_rotated_token_after_hours": {
+		Type:             schema.TypeInt,
+		Optional:         true,
+		Description:      "Sets the expiration time of the existing token secret to expire after the specified number of hours. You can set this to a value of 0 to expire the current token secret immediately. This field is only used when the token is rotated with `keepers` field.",
+		DiffSuppressFunc: IgnoreAfterCreation,
```
sfc-gh-asawicki: question: what if I want to rotate the token but also change its rotation period? Can I do that in Snowflake?

pkg/resources/user_programmatic_access_token.go
```diff
@@ -306,6 +315,29 @@ func UpdateUserProgrammaticAccessToken(ctx context.Context, d *schema.ResourceDa
 			}
 		}
 	}
+
+	if d.HasChange("keepers") {
+		request := sdk.NewRotateUserProgrammaticAccessTokenRequest(resourceId.userName, resourceId.tokenName)
+		// Handle expire_rotated_token_after_hours. It can't be done with the usual d.Get.
+		// The value is not set in the state, and we allow zero values, so we need to handle it from the raw config.
+		if ctyValue, ok := d.GetRawConfig().AsValueMap()["expire_rotated_token_after_hours"]; ok && !ctyValue.IsNull() {
```
sfc-gh-asawicki: can we use `GetConfigPropertyAsPointerAllowingZeroValue` here?

pkg/resources/user_programmatic_access_token.go
```diff
@@ -306,6 +315,29 @@ func UpdateUserProgrammaticAccessToken(ctx context.Context, d *schema.ResourceDa
 			}
 		}
 	}
+
+	if d.HasChange("keepers") {
+		request := sdk.NewRotateUserProgrammaticAccessTokenRequest(resourceId.userName, resourceId.tokenName)
+		// Handle expire_rotated_token_after_hours. It can't be done with the usual d.Get.
+		// The value is not set in the state, and we allow zero values, so we need to handle it from the raw config.
```
sfc-gh-asawicki: why do we not set it in state?

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
```diff
@@ -442,7 +445,174 @@ func TestAcc_UserProgrammaticAccessToken_complete(t *testing.T) {
...
+				ConfigVariables: tfconfig.ConfigVariablesFromModel(t, modelBasic),
+				Check: assertThat(t,
+					resourceassert.UserProgrammaticAccessTokenResource(t, modelBasic.ResourceReference()).
+						HasNameString(id.Name()).
+						HasNoRotatedTokenName().
+						HasUserString(user.ID().Name()),
+					assertTokenNotEmpty,
+				),
+			},
+			// rotate the token with added keepers and check that the token is updated
```
sfc-gh-asawicki: question: should the token be updated when the keepers is added? I think we could change it only if the keepers is changed (i.e. the existing value is changed). BTW, I asked another question about the need for such a map instead of a dedicated field for it.

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
```diff
@@ -442,7 +445,174 @@ func TestAcc_UserProgrammaticAccessToken_complete(t *testing.T) {
...
+				ConfigDirectory: ConfigurationDirectory("TestAcc_UserProgrammaticAccessTokenRotation"),
+				ConfigVariables: tfconfig.ConfigVariablesFromModel(t, modelWithExpireRotatedTokenAfterHours),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(modelWithExpireRotatedTokenAfterHours.ResourceReference(), plancheck.ResourceActionNoop),
+					},
+				},
+				Check: resource.ComposeAggregateTestCheckFunc(
+					assertThat(t,
+						resourceassert.UserProgrammaticAccessTokenResource(t, modelWithExpireRotatedTokenAfterHours.ResourceReference()).
```
sfc-gh-asawicki: do we need this assertions when noop and we assert that token was not rotated? same in some steps below

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
```diff
@@ -442,7 +445,174 @@ func TestAcc_UserProgrammaticAccessToken_complete(t *testing.T) {
...
+				},
+				Check: assertThat(t,
+					resourceassert.UserProgrammaticAccessTokenResource(t, modelWithKeepersDifferentValueAndExpireRotatedTokenAfterHours.ResourceReference()).
+						HasNameString(id.Name()).
+						HasRotatedTokenNameNotEmpty().
+						HasUserString(user.ID().Name()),
+					assertTokenNotRotated,
+				),
+			},
+			// rotate the token when the keepers are removed
```
sfc-gh-asawicki: why? Shouldn't it then just stay the same?

sfc-gh-asawicki (changes requested):

MIGRATION_GUIDE.md

MIGRATION_GUIDE.md

docs/resources/user_programmatic_access_token.md
sfc-gh-jmichalak: This is what comes from SF after rotating. (name, new secret, rotated token name). There is no list of related keys.

pkg/resources/user_programmatic_access_token.go
sfc-gh-jmichalak: It's not returned from SF. It only returns `expires_at`. During rotation, Snowflake creates a new token with a name like TOKEN_ROTATED_1752751196964 and sets expires_at according to this value. So, this field is not related to this object, but to the new one.

docs/resources/user_programmatic_access_token.md
sfc-gh-jmichalak: This is the standard format used in other providers: https://registry.terraform.io/providers/hashicorp/random/latest/docs#resource-keepers. This was also discussed during the design review.

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-jmichalak: This is the standard format used in other providers: https://registry.terraform.io/providers/hashicorp/random/latest/docs#resource-keepers. All the random value resources behave this way.

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-jmichalak: Removed.

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-jmichalak: I think we can leave it. This way, you don't need to run TF twice.

pkg/resources/user_programmatic_access_token.go
sfc-gh-jmichalak: No, as it's a different case, but I extracted this code to a helper function.

docs/resources/user_programmatic_access_token.md
sfc-gh-jmichalak: Done.

pkg/resources/user_programmatic_access_token.go
sfc-gh-jmichalak: Let's rephrase it: you **need** to rotate the token if you want to change its expiration date (or you can disable it, but that's a different story). The rotation period is not important at all from our side. It's set by the user, and we only care about triggering the rotation itself.

docs/guides/authentication_methods.md
sfc-gh-jmichalak: Done, in the resource docs as well.

pkg/acceptance/bettertestspoc/assert/resourceassert/user_programmatic_access_token_resource_gen.go
sfc-gh-jmichalak: I reran this, and this is the outcome. The same happens for lists. The `not supported message` is displayed only for model builders.

MIGRATION_GUIDE.md
```diff
@@ -35,6 +35,8 @@ This feature will be marked as a stable feature in future releases. Breaking cha
 #### New `snowflake_user_programmatic_access_token` resource
 Added a new preview resource for managing users' programmatic access tokens. See reference [docs](https://docs.snowflake.com/en/sql-reference/sql/alter-user-add-programmatic-access-token) and a [user guide](https://docs.snowflake.com/en/user-guide/programmatic-access-tokens) for more details.
 
+This resource also supports token rotation. See our [Authentication Methods guide](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/guides/authentication_methods#managing-pats) and the [resource documentation](https://registry.terraform.io/providers/snowflakedb/snowflake/2.4.0/docs/resources/user_programmatic_access_token).
```
sfc-gh-jcieslak: in one link you reference the latest provider version, in second the 2.4.0 one, is this intentional?

docs/guides/authentication_methods.md
```diff
@@ -91,8 +91,77 @@ You must fulfill the following prerequisites to generate and use programmatic ac
...
+  ~ resource "snowflake_user_programmatic_access_token" "example" {
+        id                                        = "\"PAT\"|\"TOKEN\""
+      ~ keepers                                   = {
+          - "rotation_time" = "2025-07-17T12:20:51Z"
+        } -> (known after apply)
+        name                                      = "TOKEN"
+        # (10 unchanged attributes hidden)
+    }
+
+  # time_rotating.my_token_rotation will be created
```
sfc-gh-jcieslak: it recreates itself after 30days? Or was it accidentally copied and pasted?

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-jcieslak: I think a better option would be to have "custom", but not particularly harder to do. An alternative schema:
```terraform
resource "pat" "..." {
  rotate_every = { # Optional block
    days = number
    hours = number
    minutes = number
    # Cron syntax?
  }
  last_rotation = computed time
}
```
With the following actions:
- No block defined = no rotation
- Adding block to the config = saving current timestamp into computed field (will be used to calculate next rotation)
- Removing block = no action
- last_rotation value + rotation time < current time = rotate the key and update last_rotation value with current time

I think we expected the keepers' map to bring more value than the "usual approach" (e.g., presented above) would bring. Despite this design being implemented in HC resource(s) I don't see additional value over primitive/object types.

(optional) for fine control CRON syntax would be good to set exact rotation periods

After second thought maybe the following would be better
```terraform
resource "pat" "..." {
  rotation = { # optional block
    rotate_every = { # required block
      days = number
      hours = number
      minutes = number
      # Cron syntax?
    }
    # + the field that would be responsible for extending expiration time (`EXPIRE_ROTATED_TOKEN_AFTER_HOURS ` param; couldn't think of a good name)
    last_rotation = computed time
  }
}
```

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-jcieslak: I think the question was more about not rotating on removal. If possible, I guess we shouldn't rotate the key.

pkg/acceptance/bettertestspoc/assert/resourceassert/user_programmatic_access_token_resource_gen.go
sfc-gh-asawicki: Yes, but why? It shouldn't. We need to check it, so add it to the README at least (with the initial diagnosis).

docs/resources/user_programmatic_access_token.md
sfc-gh-asawicki: Yup, I commented the design when I was making this comment.

pkg/resources/user_programmatic_access_token.go
sfc-gh-asawicki: The description states first: `You can set this to a value of 0 to expire the current token secret immediately` (it later mentions it's used only `when the token is rotated with`), which may be easily misinterpreted by the users. We could:
- at least rewrite the description so that it first emphasizes when the field can be used (I know it's a copy-paste of https://docs.snowflake.com/en/sql-reference/sql/alter-user-rotate-programmatic-access-token#parameters but I think we can make it clearer)
- if we proceed with the proposal I made with not having an arbitrary map for keepers, we could have an object with two fields: one for rotation trigger and one for the current tokens expiration

I am also thinking if we should not throw error when this field is altered by user WITHOUT triggering the rotation (because then by just looking at the config we don;t know what's the expiration, but maybe we don;t need to worry about this).

pkg/resources/user_programmatic_access_token.go
sfc-gh-asawicki: Hmmm, there are a few usages with GetConfigPropertyAsPointerAllowingZeroValue[int]. Is the only difference the intdefault check or something else prohibits us from using it?

pkg/resources/user_programmatic_access_token.go
sfc-gh-asawicki: Okay, but we could set it in state to make the implementation simpler, right? Also, is it really not visible in `d.GetOk`? It shouldn't be different than any other situation when something is initially not set but then appears in config

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
```diff
@@ -436,13 +439,158 @@ func TestAcc_UserProgrammaticAccessToken_complete(t *testing.T) {
...
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"days_to_expiry", "mins_to_bypass_network_policy_requirement", "token"},
+				ImportStateVerifyIgnore: []string{"days_to_expiry", "expire_rotated_token_after_hours", "mins_to_bypass_network_policy_requirement", "token"},
 			},
 		},
 	})
 }
 
-// TODO(next PR): add tests for rotating the token
+func TestAcc_UserProgrammaticAccessToken_rotating(t *testing.T) {
```
sfc-gh-asawicki, line 447: I was also thinking that we should have an acceptance test checking the path from the example (using time_rotation resource). It would validate and document what we are proposing.

You would need to:
- add this to the test definition:
```
		ExternalProviders: map[string]resource.ExternalProvider{
			"time": {
				VersionConstraint: "0.13.1",
				Source:            "hashicorp/time",
			},
		},
```
- trigger change on the time_rotation resource (e.g. by providing and changing the triggers https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/rotating#triggers-1)

docs/resources/user_programmatic_access_token.md
sfc-gh-asawicki: My take:
- the example you provided have the `keepers` used in the random resource, and the regenerated value is used in the target resource to trigger action (so differently than here where you put the keepers on the target resource)
- The rotation IMO boils down to having a single field with the following logic:
  - do nothing on creation (does not matter if the value is filled or not)
  - do nothing when changed from nothing to something (so setting for the first time)
  - ROTATE when changed from something to something else
  - do nothing when removed (we don't want to trigger recreation when user removes it)
- with such a simple field, everyone can still provide the desired conditions combinations (e.g. by concatenating them like `rotate_when        = "${var.some_var_for_manual_trigger}-${random_id.some_random_with_its_own_triggers.hex}-${time_rotation.after_30.rfc...}"`

sfc-gh-jmichalak: As we discussed, this is now implemented as a simple string.

pkg/acceptance/bettertestspoc/assert/resourceassert/user_programmatic_access_token_resource_gen.go
sfc-gh-jmichalak: I already had added it before to the readme. 

As we discussed, this field is now implemented as a simple string, but this behavior for maps can be addressed later.

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-jmichalak: Next PRs.

MIGRATION_GUIDE.md

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-jmichalak: As we discussed, the token is not rotated when this field is added/removed. Instead, we show a non-empty plan, but with no change to the `token` field. It's noted in the resource documentation.

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-jmichalak: Ditto.

pkg/resources/user_programmatic_access_token.go
sfc-gh-jmichalak: We can't use `GetConfigPropertyAsPointerAllowingZeroValue`, as here we use a -1 default. Decided to go with a different approach.

pkg/resources/user_programmatic_access_token.go
sfc-gh-jmichalak: We can, but this is a special field that has no direct connection to the object state. It's used as a parameter during the rotation. But I changed this behavior.

pkg/resources/user_programmatic_access_token.go
sfc-gh-jmichalak: Altered as discussed.

docs/guides/authentication_methods.md
sfc-gh-jmichalak: Yes, that's how this resource works. I copy-pasted the output from `terraform plan`. Note that after the changes, the outputs are now slightly different.

sfc-gh-jcieslak (approved):

pkg/testacc/resource_user_programmatic_access_token_acceptance_test.go
sfc-gh-asawicki: Oke

PR #3847, opened 2025-07-08
feat: Implement Current Organization Account

sfc-gh-jcieslak:
## Changes
- Implement and test the current organization account resource

## Next prs
- Add migration guide
- Merge current organization account and current account tests (or extract the common part)


pkg/resources/current_organization_account.go
```diff
@@ -12,6 +14,8 @@ import (
 )
 
 var currentOrganizationAccountSchema = map[string]*schema.Schema{
+	// TODO: name will be also in the organization account resource, but it won't support alters and we have to decide if we want to support renames here or there.
```
sfc-gh-jcieslak: We have to decide where do we want to handle organization account renaming (this resource would make more sense because it alters organization account; what I would do is to add required name field that on change would perform renames)

sfc-gh-asawicki: LGTM

pkg/sdk/organization_accounts_impl_gen.go
```diff
@@ -44,6 +47,34 @@ func (v *organizationAccounts) UnsetAllParameters(ctx context.Context) error {
 	return v.client.Accounts.UnsetAllParameters(ctx)
 }
 
+// UnsetPolicySafely added manually
```
sfc-gh-asawicki: nit: extract it to *_ext.go file (we have already did it e.g. for functions in SDK)

pkg/testacc/resource_current_organization_account_acceptance_test.go
```diff
@@ -0,0 +1,994 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config/model"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config/providermodel"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+// TODO(next pr): Extract common part from current account tests (they are reused here)
```
sfc-gh-asawicki: Please add a ticket and let's do it later.

pkg/testacc/resource_current_organization_account_acceptance_test.go
```diff
@@ -0,0 +1,994 @@
...
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+		PreCheck:                 func() { TestAccPreCheck(t) },
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: config.FromModels(t, provider, completeConfigModel),
+				Check: assertThat(t,
+					resourceassert.CurrentOrganizationAccountResource(t, completeConfigModel.ResourceReference()).
```
sfc-gh-asawicki, line 333: both CurrentOrganizationAccountResource and ImportedCurrentOrganizationAccountResource return CurrentOrganizationAccountResourceAssert, so the asserts that are true for both steps could be reused without such duplication. Please apply here and in other places.

pkg/testacc/resource_current_organization_account_acceptance_test.go
```diff
@@ -0,0 +1,994 @@
...
+	t.Cleanup(networkPolicyCleanup)
+
+	stage, stageCleanup := testClient().Stage.CreateStage(t)
+	t.Cleanup(stageCleanup)
+
+	provider := providermodel.SnowflakeProvider().WithWarehouse(testClient().Ids.WarehouseId().FullyQualifiedName())
+
+	unsetParametersModel := model.CurrentOrganizationAccount("test")
+
+	setParametersModel := model.CurrentOrganizationAccount("test").
```
sfc-gh-asawicki: can we reuse this and the model in one of the next tests? The same with assertions.

sfc-gh-asawicki (changes requested):
Let's schedule this part `Merge current organization account and current account tests (or extract the common part)` for later

pkg/sdk/organization_accounts_impl_gen.go
```diff
@@ -44,6 +47,34 @@ func (v *organizationAccounts) UnsetAllParameters(ctx context.Context) error {
...
+	case PolicyKindPasswordPolicy:
+		unset.WithPasswordPolicy(true)
+	case PolicyKindSessionPolicy:
+		unset.WithSessionPolicy(true)
+	default:
+		return fmt.Errorf("policy kind %s is not supported for account policies", kind)
+	}
+	err := v.client.OrganizationAccounts.Alter(ctx, NewAlterOrganizationAccountRequest().WithUnset(*unset))
+	// If the policy is not attached to the account, Snowflake returns an error.
+	if err != nil && strings.Contains(err.Error(), fmt.Sprintf("Any policy of kind %s is not attached to ACCOUNT", kind)) {
```
sfc-gh-jmichalak: Nit: we should handle it through a custom type in `errors.go`.

pkg/resources/current_organization_account.go
```diff
@@ -55,17 +59,181 @@ func CurrentOrganizationAccount() *schema.Resource {
...
+	attachedPolicies, err := client.PolicyReferences.GetForEntity(ctx, sdk.NewGetForEntityPolicyReferenceRequest(sdk.NewAccountObjectIdentifier(client.GetAccountLocator()), sdk.PolicyEntityDomainAccount))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	for _, policy := range attachedPolicies {
+		switch policy.PolicyKind {
+		case sdk.PolicyKindPasswordPolicy,
+			sdk.PolicyKindSessionPolicy:
+			if err := d.Set(strings.ToLower(string(policy.PolicyKind)), sdk.NewSchemaObjectIdentifier(*policy.PolicyDb, *policy.PolicySchema, policy.PolicyName).FullyQualifiedName()); err != nil {
```
sfc-gh-jmichalak: Nit: check the pointers to be extra safe.

pkg/resources/current_organization_account.go
```diff
@@ -55,17 +59,181 @@ func CurrentOrganizationAccount() *schema.Resource {
...
+			if err := client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithUnset(*sdk.NewOrganizationAccountUnsetRequest().WithResourceMonitor(true))); err != nil {
+				return diag.FromErr(err)
+			}
+		}
+	}
+
+	if d.HasChange("password_policy") {
+		if oldValue, newValue := d.GetChange("password_policy"); newValue != nil && newValue.(string) != "" {
+			if oldValue != nil && oldValue.(string) != "" {
+				if err := client.OrganizationAccounts.UnsetPolicySafely(ctx, sdk.PolicyKindPasswordPolicy); err != nil {
```
sfc-gh-jmichalak: Nit: we could move it to SDK: add a new method `SetPolicySafely` that unsets the old policy, if present, and then sets a new one. Same for session policies.

pkg/resources/current_organization_account.go
```diff
@@ -55,17 +59,181 @@ func CurrentOrganizationAccount() *schema.Resource {
...
+	if v, ok := d.GetOk("resource_monitor"); ok {
+		resourceMonitorId, err := sdk.ParseAccountObjectIdentifier(v.(string))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		if err := client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithSet(*sdk.NewOrganizationAccountSetRequest().WithResourceMonitor(resourceMonitorId))); err != nil {
+			return diag.FromErr(err)
+		}
+	} else {
+		if err := client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithUnset(*sdk.NewOrganizationAccountUnsetRequest().WithResourceMonitor(true))); err != nil {
```
sfc-gh-jmichalak, line 94: We need to make sure this behavior is obvious in the docs. Unsetting params during object creation will not be present in the plan, and it is not conventional.

pkg/testacc/resource_current_organization_account_acceptance_test.go
```diff
@@ -0,0 +1,994 @@
...
+				Config: config.FromModels(t, provider, unsetParametersModel),
+				Check: assertThat(t,
+					resourceassert.CurrentOrganizationAccountResource(t, setParametersModel.ResourceReference()).HasAllDefaultParameters(),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_CurrentOrganizationAccount_EmptyParameters(t *testing.T) {
```
sfc-gh-jmichalak: Q: what is tested here that is not tested in the last steps of the previous test?

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jmichalak: Similarly, for model builders. Lots of the params are reused.

sfc-gh-jmichalak (changes requested):

pkg/sdk/organization_accounts_impl_gen.go
sfc-gh-jcieslak: Done

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jcieslak: Done

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jcieslak: Done

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jcieslak: Done

pkg/sdk/organization_accounts_impl_gen.go
sfc-gh-jcieslak: What I don't like about this is that then the error becomes more generic. By not handling it through predefined error we get the information about policy kind which is useful to debug the error (of course we could spot the issue without it, but it's easier this way. I think we should consider some solution that would somehow be able to associate given error with predefined one, but would leave the original error.

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: done

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: Good idea, added

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: Done

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jcieslak: I think it was purely checking how parameters with empty default values will behave but I'm pretty sure it's also tested above, so I'll remove this one (same in current account)

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: Will be addressed in the next pr

pkg/resources/current_organization_account.go
```diff
@@ -55,17 +59,173 @@ func CurrentOrganizationAccount() *schema.Resource {
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	for _, policy := range attachedPolicies {
+		switch policy.PolicyKind {
+		case sdk.PolicyKindPasswordPolicy,
+			sdk.PolicyKindSessionPolicy:
+			if policy.PolicyDb != nil && policy.PolicySchema != nil {
+				if err := d.Set(strings.ToLower(string(policy.PolicyKind)), sdk.NewSchemaObjectIdentifier(*policy.PolicyDb, *policy.PolicySchema, policy.PolicyName).FullyQualifiedName()); err != nil {
```
sfc-gh-jmichalak: I'm not sure it will work correctly. What happens when a policy is unset externally? We should set an empty policy name then.

pkg/sdk/organization_accounts_impl_gen.go
sfc-gh-jmichalak: It's not that generic, as it contains info about a policy kind and an account. This seems pretty unique to me. In the errors package, we can simply create a new error with a policy name.

docs/resources/current_account.md
```diff
@@ -5,6 +5,8 @@ description: |-
   Resource used to manage the account you are currently connected to. This resource is used to set account parameters and other account-level settings. See ALTER ACCOUNT https://docs.snowflake.com/en/sql-reference/sql/alter-account documentation for more information on resource capabilities.
 ---
 
+!> **Warning** During create operation every parameter that is not set in the resource will be unset on the account.
```
sfc-gh-asawicki, line 10: question: is it visible and how in the creation plan?

pkg/resources/current_organization_account.go
sfc-gh-asawicki: +1, let's add a test for it if it's missing

pkg/testacc/resource_current_account_acceptance_test.go
```diff
@@ -435,41 +435,6 @@ func TestAcc_CurrentAccount_Parameters(t *testing.T) {
 	})
 }
 
-func TestAcc_CurrentAccount_EmptyParameters(t *testing.T) {
```
sfc-gh-asawicki, line 438: why this was removed?

pkg/testacc/resource_current_organization_account_acceptance_test.go
```diff
@@ -437,41 +106,6 @@ func TestAcc_CurrentOrganizationAccount_Parameters(t *testing.T) {
 	})
 }
 
-func TestAcc_CurrentOrganizationAccount_EmptyParameters(t *testing.T) {
```
sfc-gh-asawicki: why this was removed?

sfc-gh-jcieslak: As above

pkg/testacc/resource_current_account_acceptance_test.go
sfc-gh-jcieslak: This test case was already covered by test cases in the main test above, so It didn't make sense to have it.

pkg/sdk/organization_accounts_impl_gen.go
sfc-gh-jcieslak: I'll address this in the next pr then

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: Added test case

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: Update, this had to be addressed now due to identifier change. The rest of the fields will be added later.

docs/resources/current_account.md
sfc-gh-jcieslak: Unfortunately it's not and I don't think there's a way to show it in plan (I think we already tried a few times)

pkg/resources/current_organization_account.go
```diff
@@ -47,25 +54,213 @@ func CurrentOrganizationAccount() *schema.Resource {
...
 
 func CreateCurrentOrganizationAccount(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	return nil
+	client := meta.(*provider.Context).Client
+
+	organizationAccounts, err := client.OrganizationAccounts.Show(ctx, sdk.NewShowOrganizationAccountRequest())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	if len(organizationAccounts) != 1 {
```
sfc-gh-jmichalak, line 71: Q: Is having only 1 `organizationAccounts` guaranteed by SF? If so, we can move this check to the SDK and return an error from there. Of course, this can be done later.

pkg/resources/current_organization_account.go
sfc-gh-jmichalak: In this test, we simply unset the policy externally, but this policy type is not returned in `attachedPolicies`, so after read it is still filled in the state. The provider doesn't update the state, so this test still passes. Please check if after unsetting externally and rerunning TF, the policy gets attached back (in Snowsight).

IMO, we should check all the policy types at each read, and if we don't get them from `PolicyReferences.GetForEntity`, set an empty string.

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: Yes, I think it is, will modify in next pr

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: Adjusted

sfc-gh-jmichalak: Looks good now, but we're using unsafe `*policy.PolicyDb` and `*policy.PolicySchema` here. I found that we do not check nil in other places in the code, so we can fix this in a follow-up for all such places.

docs/resources/current_account.md
sfc-gh-asawicki: I think it was a different problem. Usually, during create you should see the plan, e.g. that parameter A will have value or will be known after apply, etc. Let's talk tomorrow (this is not blocking if you want to merge). Alternatively, we could do a simple create (essentially working like import), and handle update in the following step. Maybe it would be better for resource like this one?

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

pkg/sdk/organization_accounts_impl_gen.go
sfc-gh-jcieslak: Addressed in https://github.com/snowflakedb/terraform-provider-snowflake/pull/3866

docs/resources/current_account.md
sfc-gh-jcieslak: Addressed in https://github.com/snowflakedb/terraform-provider-snowflake/pull/3866

PR #3850, opened 2025-07-10
chore: Test zero values in plugin framework

sfc-gh-asawicki:
Test zero values in terraform plugin framework:
- test zero values handling for optional fields
  - add helpers for handling values in terraform plugin framework 
- functional test improvements
  - set up test HTTP server for functional tests - earlier functional tests were going through env modification
  - allow dynamically modifying the HTTP server response (DynamicHandler abstraction)
  - prepare actions log that can be reused inside functional tests
  - prepare embeddable struct allowing easy access to test server from test resources
  - extract provider context
  - test handling computed list (needed for actions log)
  - test server communciation

Next PRs:
- handling creation with external defaults
  - optional+computed
  - backing field (like show output)
  - parameters - resource plan modifier (like parameters custom diff in SDKv2)
  - parameters - backing field (like show output)
  - parameters - private
- diff suppresions
- validations
- warehouse resource in the new plugin framework
- Add README for setting up functional tests
- update old functional tests to use the HTTP server
- solve TODOs from dynamic handler


pkg/testfunctional/0_setup_test.go
```diff
@@ -28,10 +29,19 @@ func setup() {
...
+		cleanup()
 		os.Exit(1)
 	}
 }
 
+func cleanup() {
+	functionalTestLog.Printf("[INFO] Running functional tests cleanup")
+
+	if serverCleanup != nil {
+		defer serverCleanup()
```
sfc-gh-jcieslak: nit: unnecessary defer

pkg/testfunctional/0_setup_test.go
```diff
@@ -28,10 +29,19 @@ func setup() {
 	err := initialize()
 	if err != nil {
 		functionalTestLog.Printf("[ERROR] Functional tests context initialization failed with: `%s`", err)
+		cleanup()
 		os.Exit(1)
```
sfc-gh-jcieslak: Instead of calling cleanup twice (one in execute and the second here), shouldn't we just `return 1` here?

pkg/testfunctional/0_setup_test.go
sfc-gh-jmichalak: Why defer here?

pkg/testfunctional/10_plugin_framework_poc_helpers_zero_values.go
```diff
@@ -0,0 +1,56 @@
...
+			*unsetField = nil
+		} else {
+			*setField = sdk.Int(int(planned.ValueInt64()))
+		}
+	}
+}
+
+func stringAttributeCreate(stringAttribute types.String, createField **string) {
+	if !stringAttribute.IsNull() {
+		*createField = sdk.String(stringAttribute.ValueString())
```
sfc-gh-jmichalak: Nit: `stringAttribute.ValueStringPointer`

pkg/testfunctional/common/action_log.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+const ActionsLogPropertyName = "actions_log"
+
+type ActionsLogEmbeddable struct {
```
sfc-gh-jcieslak, line 14: nit: `ActionsLog` -> `ActionLogs`?

pkg/testfunctional/common/action_log.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+func AppendActions(ctx context.Context, actionsLog *ActionsLogEmbeddable, actionsProvider func() []ActionLogEntry) diag.Diagnostics {
+	existingEntries := actionsLog.ActionsLog.Elements()
+
+	actions := actionsProvider()
+
+	for _, a := range actions {
+		entry, diags := types.ObjectValue(GetActionLogEntryTypes(), map[string]attr.Value{
```
sfc-gh-jmichalak, line 69: Nit: This can be simplified by extracting and using `ActionEntry()`.

pkg/testfunctional/common/dynamic_handler.go
```diff
@@ -0,0 +1,58 @@
...
+func (h *DynamicHandler[T]) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case http.MethodGet:
+		logger.Printf("[DEBUG] Received get request. Current value %v", h.currentValue)
+		w.WriteHeader(http.StatusOK)
+		_ = json.NewEncoder(w).Encode(h.currentValue)
+	case http.MethodPost:
+		w.WriteHeader(http.StatusCreated)
+		var newValue T
+		_ = json.NewDecoder(r.Body).Decode(&newValue)
```
sfc-gh-jmichalak, line 28: Please handle both errors.

pkg/testfunctional/common/dynamic_handler.go
```diff
@@ -0,0 +1,58 @@
...
+
+func NewDynamicHandler[T any]() *DynamicHandler[T] {
+	return &DynamicHandler[T]{
+		replaceWithFunc: func(_ T, t2 T) T {
+			return t2
+		},
+	}
+}
+
+func NewDynamicHandlerWithInitialValue[T any](initialValue T) *DynamicHandler[T] {
```
sfc-gh-jmichalak: This function is not used. Besides, `replaceWithFunc` is not set here, but is required in the `ServeHTTP` function.

pkg/testfunctional/computednestedlist/test_resource_plugin_framework_computed_nested_list.go
```diff
@@ -0,0 +1,170 @@
...
+		setActionsOutputDedicated(ctx, response, data)
+	default:
+		response.Diagnostics.AddError("Use correct option", "Available options are: STRUCT, EXPLICIT, DEDICATED")
+		return
+	}
+
+	if response.Diagnostics.HasError() {
+		return
+	}
+	response.Diagnostics.Append(response.State.Set(ctx, &data)...)
```
sfc-gh-jmichalak, line 87: Q: what about Read?

pkg/testfunctional/httpserver/api.go
```diff
@@ -0,0 +1,5 @@
+package httpserver
+
+type Read struct {
```
sfc-gh-jcieslak: Shouldn't this be in pkg/testfunctional/httpserver/test_resource_plugin_framework_http_server.go as it is mostly tied to this resource?

pkg/testfunctional/common/logger.go
```diff
@@ -0,0 +1,8 @@
+package common
+
+import (
+	"log"
+	"os"
+)
+
+var logger = log.New(os.Stdout, "", log.LstdFlags)
```
sfc-gh-jcieslak, line 8: is it necessary? It doesn't look any different than the default one

pkg/testfunctional/10_plugin_framework_poc_helpers_zero_values.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+func int64AttributeCreate(int64Attribute types.Int64, createField **int) {
+	if !int64Attribute.IsNull() {
+		*createField = sdk.Int(int(int64Attribute.ValueInt64()))
+	}
+}
+
+func int64AttributeUpdate(planned types.Int64, inState types.Int64, setField **int, unsetField **int) {
```
sfc-gh-jcieslak, line 35: Note: We usually use booleans for unset fields, unless this is somewhat framework specific

pkg/testfunctional/test_resource_plugin_framework_http_server_acceptance_test.go
```diff
@@ -0,0 +1,60 @@
+package testfunctional_test
```
sfc-gh-jmichalak, line 1: I don't quite get why we need to have this testing resource.

pkg/testfunctional/0_setup_test.go
sfc-gh-asawicki: Well, it's a single cleanup rn but this was done on purpose when we add more cleanups. Deferring them will run all of them even if one of them panics. Leaving it.

pkg/testfunctional/0_setup_test.go
sfc-gh-asawicki: If we return 1 here, we need to handle the status above in execute, instead, we run cleanup again because os.Exit will prevent defer from happening. Both options work.

I reorganized in the 3rd way, maybe this one will be the most appealing.

pkg/testfunctional/common/action_log.go
sfc-gh-asawicki: No, it's the log, of all actions happening, and not a single action which results we log.

pkg/testfunctional/httpserver/api.go
sfc-gh-asawicki: Moved

pkg/testfunctional/common/logger.go
sfc-gh-asawicki: I wanted to have a dedicated one here. It does not serve a big purpose yet, but it's also not harming IMO.

pkg/testfunctional/10_plugin_framework_poc_helpers_zero_values.go
sfc-gh-asawicki: Yes and we have two methods for this: set and unset. I wanted to keep the duality here as we are familiar with it but I did not want to handle the logic for two different calls in the simple HTTP server. But it should be refined later, I left a clarifying comment and added a TODO for later.

pkg/testfunctional/0_setup_test.go
sfc-gh-asawicki: Well, it's a single cleanup rn but this was done on purpose when we add more cleanups. Deferring them will run all of them even if one of them panics.

pkg/testfunctional/10_plugin_framework_poc_helpers_zero_values.go
sfc-gh-asawicki: Done

pkg/testfunctional/common/action_log.go
sfc-gh-asawicki: It would be nice but the output is not the same (ActionLogEntry vs types.ObjectValue)

pkg/testfunctional/common/dynamic_handler.go
sfc-gh-asawicki: There is no need. This is a test server which does not have the knowledge about the test being run. It's also supposed to be simple helper for tests in which we don't want to add error handling logic. Also, if there is an error here, we will know on the test assertion level.

pkg/testfunctional/common/dynamic_handler.go
sfc-gh-asawicki: Removed

pkg/testfunctional/computednestedlist/test_resource_plugin_framework_computed_nested_list.go
sfc-gh-asawicki: It was not needed for the actions log use case. Left a TODO for later.

pkg/testfunctional/test_resource_plugin_framework_http_server_acceptance_test.go
sfc-gh-asawicki: It verifies that the server works as intended and can be used in other tests.

PR #3851, opened 2025-07-10
chore: Fix tests and update documentation after reverting changes from BCR 2025_03

sfc-gh-jmichalak:
[BCR-1944](https://docs.snowflake.com/release-notes/bcr-bundles/un-bundled/bcr-1944) with arguments attributes changes has been unbundled from BCR 2025_03. This caused our tests to fail. In this change:
- Fix function and procedure tests
- Adjust the migration guide and the BCR migration guide.
- Fix a config parser unit test.
- Skip acceptance tests showing incorrect behavior for older provider versions. They will be handled separately after this change is released in Snowflake once again.

Note that the migration guide for the backported versions will be updated in https://github.com/snowflakedb/terraform-provider-snowflake/pull/3841.

[BCR-1944]: https://snowflakecomputing.atlassian.net/browse/BCR-1944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfc-gh-jcieslak (approved):

pkg/sdk/testint/procedures_integration_test.go
```diff
@@ -2258,23 +2259,23 @@ def filter_by_role(session, table_name, role):
 		input             string
 		expectedShowValue string
 	}{
-		{"NUMBER(36, 5)", "NUMBER(36,5)"},
-		{"NUMBER(36)", "NUMBER(36,0)"},
+		{"NUMBER(36, 5)", "NUMBER"},
```
sfc-gh-asawicki, line 2262: I really dislike that we are reverting it as we will have to reintroduce it in a moment, maybe 2025_04 Bundle. BTW, is it already possible to enable 2025_04? If so, let's duplicate this test, have one for pre-2025_04, and one with enabled. That way we will have full coverage and much easier work after (only the general checks will have to be adjusted).

sfc-gh-asawicki (changes requested):

pkg/sdk/testint/procedures_integration_test.go
sfc-gh-jmichalak: BCR `2025_04` can be enabled. However, the behavior is the same - attributes are not returned. If the change gets into one of the next BCRs, we would simply revert two files, so I don't think it's a huge issue.

sfc-gh-asawicki: Discussed internally, resolving for now. We will add secondary test after this change is available in `2025_04` or any other Bundle.

sfc-gh-asawicki (approved):

PR #3853, opened 2025-07-11
chore: Add test results gathering

sfc-gh-jcieslak:
## Changes
- Adjusted makefile commands to return JSON format
- Added a new make command for CI use to save the given test command's results
- Created a new script that connects to the fourth connection profile and creates a file with content passed through stdin, then uploads it to the temporary stage, and at the end, uploads everything to the target table
- Adjusted workflow commands by piping their output to the newly introduced command (had to add special pipefail shell param to keep the correct step exit code; otherwise, we could miss failing tests)


Makefile
```diff
@@ -67,20 +67,23 @@ sweep: ## destroy the whole architecture; USE ONLY FOR DEVELOPMENT ACCOUNTS
...
 
-test-unit: ## run unit tests
-	go test -v -cover $$(go list ./... | grep -v -E "$(UNIT_TESTS_EXCLUDE_PATTERN)")
+create-test-output-directory:
+	mkdir -p $$HOME/test_results
 
-test-acceptance: ## run acceptance tests
-	TF_ACC=1 SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE=true TEST_SF_TF_REQUIRE_TEST_OBJECT_SUFFIX=1 TEST_SF_TF_REQUIRE_GENERATED_RANDOM_VALUE=1 SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test -run "^TestAcc_" -v -cover -timeout=120m ./pkg/testacc
+test-unit: create-test-output-directory ## run unit tests
+	go test -v -cover $$(go list ./... | grep -v -E "$(UNIT_TESTS_EXCLUDE_PATTERN)") -json | tee $$HOME/test_results/test_unit_output.json
```
sfc-gh-asawicki: okay, so I understand they are outputted to the console also as json, correct?

pkg/scripts/save_test_results/save_test_results.go
```diff
@@ -0,0 +1,137 @@
...
+	"os"
+	"slices"
+	"strings"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testprofiles"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+// TODO: Process rows to match given test outcome with its start to save test creation timestamp.
```
sfc-gh-asawicki: ticket numebr missing

Makefile
sfc-gh-jcieslak: Yes, I didn't find a way to output normal output to the console and JSON to the file

pkg/scripts/save_test_results/save_test_results.go
sfc-gh-jcieslak: Removed this TODO, but it will be included in future improvements in the document I prepared (which eventually will be later transformed into ticket)

Makefile
sfc-gh-jcieslak: Refactored to use the recently added test output processor (had to adjust it a bit, because it was a bit broken, or more like the output was in the incorrect place, which would make us wait for the logs whole test run instead of being real time)

.github/workflows/tests.yml
```diff
@@ -136,13 +142,17 @@ jobs:
...
         if: ${{ !cancelled() && steps.setup_terraform.conclusion == 'success' }}
 
+
       - name: Run acceptance tests
-        run: make test-acceptance
+        # pipefail needs to be set to ensure that the exit code of make test command is captured even when piped (otherwise it would take the exit code of make upload-test-results)
+        run: set -o pipefail && make test-acceptance | make upload-test-results
         if: ${{ !cancelled() && steps.setup_terraform.conclusion == 'success' }}
         env:
+          TEST_SF_TF_TEST_TYPE: "functional"
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I think this one is correct

Makefile
```diff
@@ -68,19 +68,19 @@ sweep: ## destroy the whole architecture; USE ONLY FOR DEVELOPMENT ACCOUNTS
 		fi;
 
 test-unit: ## run unit tests
-	go test -v -cover $$(go list ./... | grep -v -E "$(UNIT_TESTS_EXCLUDE_PATTERN)")
+	go test -v -cover $$(go list ./... | grep -v -E "$(UNIT_TESTS_EXCLUDE_PATTERN)") -json
```
sfc-gh-asawicki: hm, but I still think we didn't agree to change this output to json as it's much harder to read by us during development.

Makefile
```diff
@@ -98,6 +98,11 @@ test-main-terraform-use-cases-docker-compose: ## run main terraform use cases te
 process-test-output-docker-compose: ## run test output processor within docker environment
 	docker compose -f ./packaging/docker-compose.yml run --quiet-pull --rm process-test-output
 
+upload-test-results: ## run test output processor and save results to central location
```
sfc-gh-asawicki: this recipe uses some pipeline specific environment variables. Mayybe it could just be a step in the job definition instead something exposed without context in our makefile?

pkg/scripts/test_output_uploader/test_output_uploader.go
```diff
@@ -0,0 +1,78 @@
...
+	"path/filepath"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testprofiles"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+var testResultsTableId = sdk.NewSchemaObjectIdentifier("TEST_RESULTS_DATABASE", "TEST_RESULTS_SCHEMA", "TEST_RESULTS_TABLE")
+
+func main() {
+	clientConfig, err := sdk.ProfileConfig(testprofiles.Fourth)
```
sfc-gh-asawicki: to be discussed internally

sfc-gh-asawicki (changes requested):

Makefile
sfc-gh-asawicki: Hm, ok, I see that the output is logged in the results processing. Still, I think we should have a normal output when running locally (I do this for unit and functional tests).

sfc-gh-jcieslak: Made it optional by env variable which will be set in CI

.github/workflows/tests.yml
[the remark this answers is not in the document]
sfc-gh-jcieslak: fixed

Makefile
sfc-gh-jcieslak: well, it's repeated as we have two testing pipeliens, that's why maybe it would be better to have it in one place (\+ pipelines are already confusing and hiding a bit of logic makes them simpler to understand). We also have other commands that are more CI related. Maybe we should create a task to split them like so
```make
# Makefile

# Include the modular makefiles
include make/dev.mk
include make/ci.mk

.PHONY: help
help:
	@echo "Available commands:"
	@echo "  Dev: build, run, test"
	@echo "  CI:  ci-build, ci-test, ci-deploy"
```
in make you can do conditional includes like so
```make
ifeq ($(CI), true)
    include make/ci.mk
else
    include make/dev.mk
endif
```
which maybe would exluce unnecessary commands that are CI related

pkg/scripts/test_output_uploader/test_output_uploader.go
sfc-gh-jcieslak: As far as I remember we wanted it to move to already existing sweeper step which I just did

Makefile
sfc-gh-asawicki: Hmmm, let's try it in a separate PR (how many rules we could move etc). Maybe it's a good idea to split after all.

.github/workflows/tests.yml
```diff
@@ -146,13 +153,17 @@ jobs:
           terraform_wrapper: false
 
       - name: Run functional tests of the underlying terraform libraries
-        run: make test-functional
+        # pipefail needs to be set to ensure that the exit code of make test command is captured even when piped (otherwise it would take the exit code of make save-test-results)
+        run: set -o pipefail && make test-functional | make save-test-results
```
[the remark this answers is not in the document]
sfc-gh-asawicki: ?

.github/workflows/tests.yml
```diff
@@ -146,13 +153,17 @@ jobs:
...
         if: ${{ !cancelled() && steps.setup_terraform.conclusion == 'success' }}
```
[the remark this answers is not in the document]
sfc-gh-asawicki: ?

pkg/sdk/sweepers_upload_test_results_test.go
```diff
@@ -0,0 +1,66 @@
...
+	if testenvs.GetSnowflakeEnvironmentWithProdDefault() != testenvs.SnowflakeProdEnvironment {
+		t.Skip("Skipping for non-prod Snowflake environments")
+	}
+
+	testResultsTableId := sdk.NewSchemaObjectIdentifier("TEST_RESULTS_DATABASE", "TEST_RESULTS_SCHEMA", "TEST_RESULTS_TABLE")
+
+	client := fourthTestClient(t)
+
+	// Create a temporary stage for test results
+	testResultsStageId := sdk.NewSchemaObjectIdentifier("TEST_RESULTS_DATABASE", "TEST_RESULTS_SCHEMA", "TEST_RESULTS_STAGE")
```
sfc-gh-asawicki, line 23: nit: we are using fourth account but currently the protected db name is for all accounts (as we dont have per-account config). We should have it as an improvement.

sfc-gh-asawicki (approved):

pkg/sdk/sweepers_upload_test_results_test.go
```diff
@@ -0,0 +1,66 @@
...
+	files, err := filepath.Glob("/tmp/test_results_*.csv")
+	if err != nil {
+		t.Fatal("Failed to find test results files:", err)
+	}
+	t.Logf("Found %v test results files", files)
+
+	// Range over temporary files with test results and copy the data into target tables
+	for _, file := range files {
+		// Put file to the stage
+		if _, err := client.ExecUnsafe(context.Background(), fmt.Sprintf("put file://%s @%s auto_compress = true overwrite = true", file, testResultsStageId.FullyQualifiedName())); err != nil {
```
sfc-gh-jmichalak, line 46: Nit: we have several helpers for PUT. On the other hand, maybe we shouldn't fail everything immediately, but continue.

sfc-gh-jmichalak (approved):

pkg/sdk/sweepers_upload_test_results_test.go
sfc-gh-asawicki: The helpers are in test clients, and we have a TODO to move all sweeper logic to the test client helpers instead of the direct client use.

PR #3856, opened 2025-07-14
chore: Test custom type with metadata

sfc-gh-asawicki:
TODO


sfc-gh-jcieslak (approved):

pkg/testfunctional/customtypes/string_with_metadata_value.go
```diff
@@ -0,0 +1,61 @@
...
+func (v StringWithMetadataValue) Type(_ context.Context) attr.Type {
+	return StringWithMetadataType{}
+}
+
+// Equal returns true if the given value is equivalent.
+func (v StringWithMetadataValue) Equal(o attr.Value) bool {
+	other, ok := o.(StringWithMetadataValue)
+
+	if !ok {
+		return false
```
sfc-gh-jmichalak, line 32: Q: does this happen? Doesn't the framework call `Equal` on types before?

pkg/testfunctional/customtypes/string_with_metadata_value.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// ToTerraformValue returns the data contained in the *String as a tftypes.Value.
+func (v StringWithMetadataValue) ToTerraformValue(ctx context.Context) (tftypes.Value, error) {
+	raw := v.ValueString()
+	metaBytes, err := json.Marshal(v.Metadata)
+	if err != nil {
+		return types.StringValue(raw).ToTerraformValue(ctx)
+	}
+	combined := fmt.Sprintf(`%s|%s`, raw, metaBytes)
```
sfc-gh-jmichalak, line 45: Why do we join with `|`? We could just encode to JSON.

pkg/testfunctional/customtypes/string_with_metadata_value.go
sfc-gh-asawicki: It comes from hashicorp docs: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom#schema-type (check also one of their implementations: https://github.com/hashicorp/terraform-plugin-framework-jsontypes/blob/4a29773393cfb852b886644532a4742b58bfaae1/jsontypes/normalized_type.go#L38)

pkg/testfunctional/customtypes/string_with_metadata_value.go
sfc-gh-asawicki: It's the simplest join. It would matter if it worked but it doesn't, so we can join like this, right?

pkg/testfunctional/customtypes/string_with_metadata_value.go
sfc-gh-jmichalak: 😦

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (approved):

PR #3857, opened 2025-07-14
chore: Test optional with backing field

sfc-gh-asawicki:
Test backing field (like show output) handling in the Terraform Plugin Framework
- handle the default values in the dynamic handler
- handle config and external changes and test


sfc-gh-jcieslak (dismissed):

pkg/testfunctional/test_resource_plugin_framework_optional_with_backing_field.go
```diff
@@ -0,0 +1,216 @@
...
+
+	err := r.HttpServerEmbeddable.Post(*opts)
+	if err != nil {
+		diags.AddError("Could not update resource", err.Error())
+	}
+	return diags
+}
+
+func (r *OptionalWithBackingFieldResource) setUpdateActionsOutput(ctx context.Context, response *resource.UpdateResponse, opts *OptionalWithBackingFieldOpts, plan *optionalWithBackingFieldResourceModelV0, state *optionalWithBackingFieldResourceModelV0) {
+	plan.ActionsLogEmbeddable = state.ActionsLogEmbeddable
```
sfc-gh-jmichalak, line 203: Nit: This pattern is used a few times across different resources and can be reused.

sfc-gh-jmichalak (approved):

pkg/testfunctional/test_resource_plugin_framework_optional_with_backing_field.go
sfc-gh-asawicki: It can be refined later, these are test resources and these action won't be in the production code.

PR #3858, opened 2025-07-14
chore: Test parameters handling (multiple variants)

sfc-gh-asawicki:
Test parameter handling in 4 variants:
1. Like in SDKv2 (resource plan modifier instead of custom diff) - not working as we can SetNewComputed like in SDKv2
2. Handling logic in read (if different than OBJECT set null) - not working as then handling with unset config fails
3. Backing field (like show output) - works
4. Private state - works


pkg/testfunctional/test_resource_plugin_framework_parameter_handling_backing_field.go
```diff
@@ -0,0 +1,274 @@
...
+	response.Diagnostics.Append(common.AppendActions(ctx, &data.ActionsLogEmbeddable, func() []common.ActionLogEntry {
+		actions := make([]common.ActionLogEntry, 0)
+		if opts.StringValue != nil {
+			actions = append(actions, common.ActionEntry("CREATE", "string_value", *opts.StringValue))
+		}
+		return actions
+	})...)
+}
+
+func (r *ParameterHandlingBackingFieldResource) create(opts *ParameterHandlingBackingFieldOpts) diag.Diagnostics {
```
sfc-gh-jcieslak, line 136: nit: Some of the operations could also be extracted as embeddable with common logic or helper functions for simplicity, this would help with focusing on the main implementation parts

pkg/testfunctional/test_resource_plugin_framework_parameter_handling_read_logic_acceptance_test.go
```diff
@@ -0,0 +1,220 @@
...
+		base.Level = "OBJECT"
+	}
+	return base
+}
+
+func init() {
+	allTestHandlers["parameter_handling_read_logic"] = parameterHandlingReadLogicHandler
+}
+
+func TestAcc_TerraformPluginFrameworkFunctional_ParameterHandling_ReadLogic(t *testing.T) {
```
sfc-gh-jcieslak, line 42: In the pr description, you mentioned this is not working. Please point the non-working part in the test steps which shows that (couldn't find such description)

pkg/testfunctional/test_resource_plugin_framework_parameter_handling_backing_field.go
sfc-gh-jmichalak: +1, so much boilerplate

pkg/testfunctional/test_resource_plugin_framework_parameter_handling_backing_field_acceptance_test.go
```diff
@@ -0,0 +1,327 @@
...
+						planchecks.ExpectChange(resourceReference, "string_value", tfjson.ActionUpdate, nil, sdk.String(newValue)),
+						planchecks.ExpectComputed(resourceReference, "string_value", false),
+					},
+				},
+				Config: parameterHandlingBackingFieldAllSetConfig(id, resourceType, newValue),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceReference, "id", id.FullyQualifiedName()),
+					resource.TestCheckResourceAttr(resourceReference, "string_value", newValue),
+					resource.TestCheckResourceAttr(resourceReference, "string_value_backing_field.value", newValue),
+					resource.TestCheckResourceAttr(resourceReference, "string_value_backing_field.level", "OBJECT"),
```
sfc-gh-jmichalak, line 168: Why not `OTHER`?

pkg/testfunctional/test_resource_plugin_framework_parameter_handling_resource_plan_modifier.go
```diff
@@ -0,0 +1,271 @@
...
+
+	// For all other cases, if a parameter is set in the configuration, we can ignore parts needed for Computed fields.
+	if !config.StringValue.IsNull() { //nolint:staticcheck
+		return
+	}
+
+	// If the configuration is not set, perform SetNewComputed for cases like:
+	// 1. Check if the parameter value differs from the one saved in state (if they differ, we'll update the computed value).
+	// 2. Check if the parameter is set on the object level (if so, it means that it was set externally, and we have to unset it).
+	// if foundParameter.StringValue !=  || parameter.Level == objectParameterLevel {
```
sfc-gh-jmichalak, line 122: Why is this commented out?

sfc-gh-jmichalak (dismissed):

pkg/testfunctional/test_resource_plugin_framework_parameter_handling_backing_field.go
sfc-gh-asawicki: You will both love the plugin framework if you think this is a lot boilerplate :D

Added a TODO for later to include at least `create` and `update` embeds (reading logic differs between the resources and their usecases).

pkg/testfunctional/test_resource_plugin_framework_parameter_handling_read_logic_acceptance_test.go
sfc-gh-asawicki: Last step:
```// This documents that the read logic added to handle previous step messes with the logic when the parameter is removed from config.```
and 
```ExpectNonEmptyPlan: true,```

pkg/testfunctional/test_resource_plugin_framework_parameter_handling_backing_field_acceptance_test.go
sfc-gh-asawicki: Because OTHER is set externally but we have setting in config, so we expect it to be on OBJECT level after successful apply

pkg/testfunctional/test_resource_plugin_framework_parameter_handling_resource_plan_modifier.go
sfc-gh-asawicki: It's in this ModifyPlan methods comment description: `// ModifyPlan inlines resources.ParameterValueComputedIf which is our previous SDK implementation.` I didn't got to uncommenting it as it failed on sooner tests. I emphasised why it's left like this.

sfc-gh-jcieslak (dismissed):

PR #3861, opened 2025-07-15
chore: Test enum handling (suppresion and validation)

sfc-gh-asawicki:
Test generic enum handling
- custom type
- attribute plan modifier for enums
- plan modifier for backing field
- acceptance tests for different flows

This resource does not use the actions log as it does not yet handle conditional computations. It will be fixed later.

References:
- https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom#semantic-equality
- https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification

### DIFF
--- a/pkg/testfunctional/customplanmodifiers/enum_suppressor.go
+++ b/pkg/testfunctional/customplanmodifiers/enum_suppressor.go
@@ -11,9 +11,7 @@ func EnumSuppressor[T customtypes.EnumCreator[T]]() planmodifier.String {
 	return enumSuppressorModifier[T]{}
 }
 
-type enumSuppressorModifier[T customtypes.EnumCreator[T]] struct {
-	et T
-}
+type enumSuppressorModifier[T customtypes.EnumCreator[T]] struct{}
 
 func (m enumSuppressorModifier[T]) Description(_ context.Context) string {
 	return "enum suppressor"
--- a/pkg/testfunctional/customtypes/enum_generic_type.go
+++ b/pkg/testfunctional/customtypes/enum_generic_type.go
@@ -14,7 +14,6 @@ var _ basetypes.StringTypable = (*EnumType[dummyEnumType])(nil)
 
 type EnumType[T EnumCreator[T]] struct {
 	basetypes.StringType
-	et T
 }
 
 func (t EnumType[T]) String() string {
--- a/pkg/testfunctional/test_resource_plugin_framework_enum_handling.go
+++ b/pkg/testfunctional/test_resource_plugin_framework_enum_handling.go
@@ -56,14 +56,14 @@ type EnumHandlingResource struct {
 }
 
 type enumHandlingResourceModelV0 struct {
-	Name                    types.String                        `tfsdk:"name"`
-	StringValue             customtypes.EnumValue[SomeEnumType] `tfsdk:"string_value"`
-	StringValueBackingField types.String                        `tfsdk:"string_value_backing_field"`
-	Id                      types.String                        `tfsdk:"id"`
+	Name                  types.String                        `tfsdk:"name"`
+	EnumValue             customtypes.EnumValue[SomeEnumType] `tfsdk:"enum_value"`
+	EnumValueBackingField types.String                        `tfsdk:"enum_value_backing_field"`
+	Id                    types.String                        `tfsdk:"id"`
 }
 
 type EnumHandlingOpts struct {
-	StringValue *SomeEnumType
+	EnumValue *SomeEnumType
 }
 
 func (r *EnumHandlingResource) Metadata(_ context.Context, request resource.MetadataRequest, response *resource.MetadataResponse) {
@@ -78,7 +78,7 @@ func (r *EnumHandlingResource) Schema(_ context.Context, _ resource.SchemaReques
 				Description: "Name for this resource.",
 				Required:    true,
 			},
-			"string_value": schema.StringAttribute{
+			"enum_value": schema.StringAttribute{
 				CustomType:  customtypes.EnumType[SomeEnumType]{},
 				Description: "String value - enum.",
 				Optional:    true,
@@ -86,7 +86,7 @@ func (r *EnumHandlingResource) Schema(_ context.Context, _ resource.SchemaReques
 					customplanmodifiers.EnumSuppressor[SomeEnumType](),
 				},
 			},
-			"string_value_backing_field": schema.StringAttribute{
+			"enum_value_backing_field": schema.StringAttribute{
 				Description: "String value backing field.",
 				Computed:    true,
 				PlanModifiers: []planmodifier.String{
@@ -119,8 +119,8 @@ func (r *EnumHandlingResource) ModifyPlan(ctx context.Context, request resource.
 	}
 
 	// we run normal equal here, as customplanmodifiers.EnumSuppressor handles it
-	if !plan.StringValue.Equal(state.StringValue) {
-		plan.StringValueBackingField = types.StringUnknown()
+	if !plan.EnumValue.Equal(state.EnumValue) {
+		plan.EnumValueBackingField = types.StringUnknown()
 	}
 
 	response.Diagnostics.Append(response.Plan.Set(ctx, &plan)...)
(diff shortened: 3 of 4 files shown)


pkg/testfunctional/customtypes/enum_generic_helpers.go
```diff
@@ -0,0 +1,12 @@
+package customtypes
+
+type EnumCreator[T ~string] interface {
+	~string
+	FromString(string) (T, error)
+}
+
+type dummyEnumType string
+
+func (e dummyEnumType) FromString(s string) (dummyEnumType, error) {
```
sfc-gh-jcieslak, line 10: If you really want to keep this function as a method (which is strange because it doesn't need to be a method), I think it would be better to provide some top-level generic FromString function that would call this one internally.

pkg/testfunctional/customtypes/enum_generic_type.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+var _ basetypes.StringTypable = (*EnumType[dummyEnumType])(nil)
+
+type EnumType[T EnumCreator[T]] struct {
+	basetypes.StringType
+	et T
```
sfc-gh-jcieslak: is it needed? I don't see any usage of `et`

pkg/testfunctional/test_resource_plugin_framework_enum_handling.go
```diff
@@ -0,0 +1,252 @@
...
+
+type SomeEnumType string
+
+const (
+	SomeEnumTypeVersion1 SomeEnumType = "VERSION_1"
+	SomeEnumTypeVersion2 SomeEnumType = "VERSION_2"
+	SomeEnumTypeVersion3 SomeEnumType = "VERSION_3"
+)
+
+func ToSomeEnumType(s string) (SomeEnumType, error) {
```
sfc-gh-jcieslak, line 33: With top-level FromString, we would only define an enum with FromString method and this could be removed

pkg/testfunctional/test_resource_plugin_framework_enum_handling.go
```diff
@@ -0,0 +1,252 @@
...
+
+type enumHandlingResourceModelV0 struct {
+	Name                    types.String                        `tfsdk:"name"`
+	StringValue             customtypes.EnumValue[SomeEnumType] `tfsdk:"string_value"`
+	StringValueBackingField types.String                        `tfsdk:"string_value_backing_field"`
+	Id                      types.String                        `tfsdk:"id"`
+}
+
+type EnumHandlingOpts struct {
+	StringValue *SomeEnumType
```
sfc-gh-jcieslak: StringValue as the name, was confusing, rename to EnumValue?

pkg/testfunctional/customplanmodifiers/enum_suppressor.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/testfunctional/customtypes"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+)
+
+func EnumSuppressor[T customtypes.EnumCreator[T]]() planmodifier.String {
+	return enumSuppressorModifier[T]{}
+}
+
+type enumSuppressorModifier[T customtypes.EnumCreator[T]] struct {
+	et T
```
sfc-gh-jcieslak: even if you are using this `et` in normalize you don't have to keep it here right?

pkg/testfunctional/test_resource_plugin_framework_enum_handling_acceptance_test.go
```diff
@@ -0,0 +1,239 @@
...
+					},
+				},
+				Config: enumHandlingAllSetConfig(id, resourceType, newValue),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceReference, "id", id.FullyQualifiedName()),
+					resource.TestCheckResourceAttr(resourceReference, "string_value", newValue),
+					resource.TestCheckResourceAttr(resourceReference, "string_value_backing_field", newValue),
+				),
+			},
+			// remove type from config but update warehouse externally to default (still expecting non-empty plan because we do not know the default)
```
sfc-gh-jcieslak: update enum ?

sfc-gh-jcieslak (dismissed):

pkg/testfunctional/test_resource_plugin_framework_enum_handling.go
sfc-gh-jmichalak: On the other hand, `FromString` is used to satisfy an interface and have all string loaders named the same. For manual use, `ToSomeEnumType` is clearer, IMO.

pkg/testfunctional/test_resource_plugin_framework_enum_handling_acceptance_test.go
```diff
@@ -0,0 +1,239 @@
...
+	id := sdk.NewAccountObjectIdentifier("abc")
+	resourceType := fmt.Sprintf("%s_enum_handling", PluginFrameworkFunctionalTestsProviderName)
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: providerForPluginFrameworkFunctionalTestsFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			// create with known value
```
sfc-gh-jmichalak: Nit:
```suggestion
			// create with an unknown value
```

sfc-gh-jmichalak (dismissed):

pkg/testfunctional/customtypes/enum_generic_helpers.go
sfc-gh-asawicki: I don't think a top-level FromString would remove the need for having it as a method (interface impl) or switching by type and being explicit with all the available enums. But maybe I misunderstood the idea. For me, this one, with simple interface which implementation we can easily add (or generate) is good enough.

pkg/testfunctional/test_resource_plugin_framework_enum_handling.go
sfc-gh-asawicki: `ToSomeEnumType` follows our current SDK implementation (e.g. ToWarehouseType). The proposed impl introduces the need for only oneliner method implementing the enumcreator which I think is acceptable and pretty neat.

pkg/testfunctional/test_resource_plugin_framework_enum_handling.go
sfc-gh-asawicki: Renamed opts, schema, and structs.

pkg/testfunctional/test_resource_plugin_framework_enum_handling_acceptance_test.go
sfc-gh-asawicki: Done

pkg/testfunctional/test_resource_plugin_framework_enum_handling_acceptance_test.go
sfc-gh-asawicki: it should be with invalid value - fixed

pkg/testfunctional/customtypes/enum_generic_type.go
sfc-gh-asawicki: Removed from here as it is not needed in type

pkg/testfunctional/customplanmodifiers/enum_suppressor.go
sfc-gh-asawicki: Yes, removed also from the plan modiier

sfc-gh-jmichalak (approved):

PR #3865, opened 2025-07-17
chore: Test optional computed in plugin framework

sfc-gh-asawicki:
TODO


pkg/testfunctional/test_resource_plugin_framework_optional_computed_handling_acceptance_test.go
```diff
@@ -0,0 +1,96 @@
...
+					},
+				},
+				Config: optionalComputedAllSetConfig(id, resourceType),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceReference, "id", id.FullyQualifiedName()),
+					resource.TestCheckResourceAttr(resourceReference, "string_value", "some value"),
+				),
+			},
+			// remove all from config (to validate that unset is run correctly and default value is in state)
+			{
```
sfc-gh-jcieslak, line 66: Can we have an additional step to confirm there is a permadiff, and it looks like this step? (non-empty plan in test step do not always mean permadiff; I'm sure you checked that, but just to distinguish this case from non-empty plans that in next steps dissapear; such plans happen in case of in-place updates)

sfc-gh-asawicki: Done

PR #3866, opened 2025-07-17
feat: Refactor current organization account logic to improve error handling…

sfc-gh-jcieslak:
… and update comment field management

Add error handling for unsupported account policies

Refactor acceptance test for current organization account to validate non-empty comment on creation

Update validation logic to include 'Comment' field for organization account options

Add missing methods and update documentation for organization account

Add comment field to organization account schema and update related logic

Add missing fields to organization account schema and implement retrieval logic

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -81,15 +100,47 @@ func TestInt_OrganizationAccount_SelfAlter(t *testing.T) {
 
 		// Set another session policy without unsetting the previous one
 		err = client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithSet(*sdk.NewOrganizationAccountSetRequest().WithSessionPolicy(sessionPolicy2.ID())))
-		assert.ErrorContains(t, err, fmt.Sprintf("Only one %s is allowed at a time", sdk.PolicyKindSessionPolicy))
+		assert.ErrorContains(t, err, "already has a SESSION POLICY on ACCOUNT. The existing policy must be removed")
```
sfc-gh-jcieslak, line 103: I'm not yet sure but I think I got inconsistent error messages for the same actions (for now changed to the current one)

pkg/acceptance/helpers/organization_account_client.go
```diff
@@ -30,10 +31,19 @@ func (c *OrganizationAccountClient) Alter(t *testing.T, req *sdk.AlterOrganizati
 	require.NoError(t, err)
 }
 
-func (c *OrganizationAccountClient) Show(t *testing.T) sdk.OrganizationAccount {
+func (c *OrganizationAccountClient) ShowCurrent(t *testing.T) *sdk.OrganizationAccount {
```
sfc-gh-jmichalak, line 34: Why don't we have it on SDK level?

pkg/sdk/errors.go
```diff
@@ -29,6 +29,8 @@ var (
 	ErrInvalidObjectIdentifier = NewError("invalid object identifier")
 	ErrDifferentDatabase       = NewError("database must be the same")
 	ErrDifferentSchema         = NewError("schema must be the same")
+
+	ErrPolicyNotAttachedToAccount = NewError("policy kind is not attached account")
```
sfc-gh-jmichalak: Nit: 
```suggestion
	ErrPolicyNotAttachedToAccount = NewError("policy kind is not attached to account")
```

pkg/sdk/organization_accounts_gen_test.go
```diff
@@ -111,13 +111,13 @@ func TestOrganizationAccounts_Alter(t *testing.T) {
 	t.Run("validation: at least one of the fields [opts.Set.Parameters opts.Set.ResourceMonitor opts.Set.PasswordPolicy opts.Set.SessionPolicy] should be set", func(t *testing.T) {
 		opts := defaultOpts()
 		opts.Set = new(OrganizationAccountSet)
-		assertOptsInvalidJoinedErrors(t, opts, errAtLeastOneOf("AlterOrganizationAccountOptions.Set", "Parameters", "ResourceMonitor", "PasswordPolicy", "SessionPolicy"))
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterOrganizationAccountOptions.Set", "Parameters", "ResourceMonitor", "PasswordPolicy", "SessionPolicy", "Comment"))
 	})
 
 	t.Run("validation: at least one of the fields [opts.Unset.Parameters opts.Unset.ResourceMonitor opts.Unset.PasswordPolicy opts.Unset.SessionPolicy] should be set", func(t *testing.T) {
```
sfc-gh-jmichalak: Update also test names.

pkg/sdk/testint/organization_accounts_integration_test.go
```diff
@@ -81,15 +100,47 @@ func TestInt_OrganizationAccount_SelfAlter(t *testing.T) {
...
 
 		err = client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithUnset(*sdk.NewOrganizationAccountUnsetRequest().WithSessionPolicy(true)))
 		require.NoError(t, err)
 		assertThatNoPolicyIsSetOnAccount(t)
 
 		// Unset session policy when there's no password policy attached
 		err = client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithUnset(*sdk.NewOrganizationAccountUnsetRequest().WithSessionPolicy(true)))
-		assert.ErrorContains(t, err, fmt.Sprintf("Any policy of kind %s is not attached to ACCOUNT", sdk.PolicyKindSessionPolicy))
+		require.NoError(t, err)
```
sfc-gh-jmichalak, line 111: Q: Why is this assertion changed?

pkg/resources/current_organization_account.go
```diff
@@ -19,6 +23,12 @@ var currentOrganizationAccountSchema = map[string]*schema.Schema{
 		Required:    true,
 		Description: "Specifies the identifier (i.e. name) for the account. It must be unique within an organization, regardless of which Snowflake Region the account is in and must start with an alphabetic character and cannot contain spaces or special characters except for underscores (_). Note that if the account name includes underscores, features that do not accept account names with underscores (e.g. Okta SSO or SCIM) can reference a version of the account name that substitutes hyphens (-) for the underscores.",
 	},
+	"comment": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("comment"),
+		Description:      externalChangesNotDetectedFieldDescription("Specifies a comment for the organization account."),
```
sfc-gh-jmichalak: Why is it with externalChangesNotDetectedFieldDescription, even though it's in SHOW?

pkg/resources/current_organization_account.go
```diff
@@ -64,74 +85,23 @@ func CurrentOrganizationAccount() *schema.Resource {
...
 
-	setParameters := new(sdk.AccountParameters)
-	if diags := handleAccountParametersCreate(d, setParameters); diags != nil {
-		return diags
-	}
-	if *setParameters != (sdk.AccountParameters{}) {
-		if err := client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithSet(*sdk.NewOrganizationAccountSetRequest().WithParameters(*setParameters))); err != nil {
-			return diag.FromErr(err)
-		}
+	if id.Name() != currentOrganizationAccount.AccountName {
```
sfc-gh-jmichalak: Shouldn't we check it also in import?

pkg/resources/current_organization_account.go
```diff
@@ -188,6 +192,19 @@ func UpdateCurrentOrganizationAccount(ctx context.Context, d *schema.ResourceDat
 		d.SetId(id.Name())
 	}
 
+	if d.HasChange("comment") {
+		newComment := d.Get("comment").(string)
+		if newComment != "" {
+			if err := client.OrganizationAccounts.Alter(ctx, sdk.NewAlterOrganizationAccountRequest().WithSet(*sdk.NewOrganizationAccountSetRequest().WithComment(newComment))); err != nil {
```
sfc-gh-jmichalak, line 208: Nit: This pattern is used a lot, we can extract it as a helper.

pkg/testacc/resource_current_organization_account_acceptance_test.go
```diff
@@ -328,6 +329,25 @@ func TestAcc_CurrentOrganizationAccount_Complete(t *testing.T) {
...
+					resourceassert.CurrentOrganizationAccountResource(t, completeConfigModel.ResourceReference()).
+						HasNameString(currentOrganizationAccountName).
+						HasAllDefaultParameters().
+						HasCommentEmpty().
+						HasResourceMonitorString(resourceMonitor.ID().Name()).
+						HasPasswordPolicyEmpty().
+						HasSessionPolicyEmpty(),
+				),
+				// Create works as an import, so with filled fields we expect the first plan not to be empty
+				ExpectNonEmptyPlan: true,
```
sfc-gh-jmichalak, line 346: I just checked that TF always calls `Create`. I'm not sure if there's a better way of handling non-creatable resources, but I wish there was better support for it.

templates/resources/current_organization_account.md.tmpl
```diff
@@ -11,17 +11,19 @@ description: |-
...
 
-!> **Warning** During create operation every parameter and policy that is not set in the resource will be unset on the organization account. However, if you pass the incorrect `name` field, it will return an error, and no changes will be made to the account until a correct name is passed.
+!> **Warning** Create in this resource works differently from other resources as it doesn't create anything on the Snowflake side. It works mostly as import, any planned changes will be visible after the first apply. It's done this way, because of the Terraform limitation which prevents us from presenting changes during resource creation.
```
sfc-gh-jmichalak: 1. What about `However, if you pass the incorrect `name` field, it will return an error, and no changes will be made to the account until a correct name is passed.`
2. `prevents us from *modiyfing* changes in terraform plan`

MIGRATION_GUIDE.md
```diff
@@ -23,6 +23,9 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ## v2.3.0 ➞ v2.4.0
 
+### *(new feature)* snowflake_current_organization_account resource
+Added a new preview resource for managing the organization account that the provider is currently connected to. It's capable of managing attached parameters, resource_monitors, and more. See reference docs for [ALTERING ORGANIZATION ACCOUNT](https://docs.snowflake.com/en/sql-reference/sql/alter-organization-account). You can read about the resource limitations in the documentation in the registry.
+
```
sfc-gh-jmichalak, line 28: Add also `This feature will be marked as...`

pkg/sdk/testint/organization_accounts_integration_test.go
sfc-gh-jcieslak: The old one was up-to-date before when there was not that much fields in `set`/`unset` structs, this is also consistent with what we have in regular accounts, so I think we'll be safer this way.

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: My mistake, it was copied from account schema, `DiffSuppressFunc` also seems to be not needed in this context (it was needed before we switched the tactic with Create operation).

pkg/resources/current_organization_account.go
sfc-gh-jcieslak: I would leave it as it is, each field has a slightly different handling in this case and the `if`s are not that long. I tried to create a helper, and it would be a bit shorter but rather less readable.

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jcieslak: yeah, this way at least the users will be able to to see the changes (with update-like logic in create the changes are not visible; they shown as `known after apply`)

templates/resources/current_organization_account.md.tmpl
sfc-gh-jcieslak: adjusted the `name` description and this warning

pkg/acceptance/helpers/organization_account_client.go
sfc-gh-jcieslak: Hmm we can add it as improvement, but for later, as there will be more entries in SHOW ORGANIZATION ACCOUNTS (e.g. during replication or similar operations). I'm not sure how those entries will look like, so I would rather keep this functionality in test and not the actual resource code.

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-asawicki: In some resources, we set fields during creation. Usually, only the ones that are not later filled out by Read. Which fields cause a nonempty plan here? Also, the next step starts with a NoOp assertion. So I am lost., What is really happening here?

pkg/testacc/resource_current_organization_account_acceptance_test.go
```diff
@@ -371,7 +391,63 @@ func TestAcc_CurrentOrganizationAccount_NameValidationOnCreate(t *testing.T) {
...
+				Config: config.FromModels(t, emptyPropertiesModel),
+				Check: assertThat(t,
+					resourceassert.CurrentOrganizationAccountResource(t, emptyPropertiesModel.ResourceReference()).
+						HasNameString(currentOrganizationAccountName).
+						HasCommentString(comment),
+					resourceshowoutputassert.OrganizationAccountShowOutput(t, emptyPropertiesModel.ResourceReference()).
+						HasComment(comment),
+				),
+				// The plan indicates the comment should change but later steps prove that our show output helps suppress it
+				ExpectNonEmptyPlan: true,
```
sfc-gh-asawicki: please assert it then as PostApplyPostRefresh

sfc-gh-asawicki: Also, as I have written above, in some resources we set them on import (check e.g. user resources)

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jcieslak: This wasn't completely right, corrected

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jcieslak: This test fundamentally was incorrect imo (it was showing wrong behavior). Adjusted.

pkg/testacc/resource_current_organization_account_acceptance_test.go
```diff
@@ -328,6 +331,39 @@ func TestAcc_CurrentOrganizationAccount_Complete(t *testing.T) {
...
+						HasSessionPolicyEmpty(),
+				),
+				// If the config is set for all fields, we expect to see them as changes in the plan (because they're different from the default values).
+				// They will be applied in the next step.
+				ExpectNonEmptyPlan: true,
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PostApplyPostRefresh: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(completeConfigModel.ResourceReference(), plancheck.ResourceActionUpdate),
+						// Expect changes for all fields that are set in the config: parameters (chose to show only 3 parameters of a different type as an example), resource monitor, password policy, session policy, comment
+						planchecks.ExpectChange(completeConfigModel.ResourceReference(), "abort_detached_query", tfjson.ActionUpdate, sdk.String("false"), sdk.String("true")),
```
sfc-gh-jmichalak, line 351: Nit: add a check for a field that we do not expect any change in.

sfc-gh-jmichalak (approved):

pkg/testacc/resource_current_organization_account_acceptance_test.go
sfc-gh-jcieslak: We expect a change in every field because the model contains values that are all different from the things we have in Snowflake. Check `TestAcc_CurrentOrganizationAccount_NonParameterValues` first step to see there are no changes planned since all fields are matching the infra.

sfc-gh-jmichalak: Yeah, it looks correct. Maybe this comment was done on an older revision.

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #3873, opened 2025-07-22
chore: Listings integration tests

sfc-gh-jcieslak:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/acceptance/helpers/listing_client.go
```diff
@@ -0,0 +1,83 @@
...
+	err := c.client().Alter(ctx, req)
+	require.NoError(t, err)
+}
+
+func (c *ListingClient) DropFunc(t *testing.T, id sdk.AccountObjectIdentifier) func() {
+	t.Helper()
+	ctx := context.Background()
+
+	return func() {
+		if err := c.client().DropSafely(ctx, id); !errors.Is(err, sdk.ErrObjectNotFound) {
```
sfc-gh-jmichalak: I already wrote in the DropSafely function, but this check shouldn't be needed here, and this should be handled there.

pkg/acceptance/helpers/listing_client.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (c *ListingClient) Show(t *testing.T, id sdk.AccountObjectIdentifier) (*sdk.Listing, error) {
+	t.Helper()
+	return c.client().ShowByID(context.Background(), id)
+}
+
+func (c *ListingClient) BasicManifest(t *testing.T) string {
+	t.Helper()
+	return `title: "title"
```
sfc-gh-jmichalak: Q: Did you check if the title is unique?

pkg/acceptance/bettertestspoc/assert/objectassert/listing_snowflake_ext.go
```diff
@@ -0,0 +1,120 @@
+// Code generated by assertions generator; DO NOT EDIT.
```
sfc-gh-jmichalak: Pls remove this line.

pkg/sdk/listings_def.go
```diff
@@ -121,49 +121,49 @@ var listingDetailsDbRow = g.DbStruct("listingDetailsDBRow").
...
 	OptionalText("unpublished_by_admin_reasons").
 	Bool("is_monetized").
 	Bool("is_application").
 	Bool("is_targeted").
-	OptionalBool("is_limited_trial").    // TODO: Not documented
-	OptionalBool("is_by_request").       // TODO: Not documented
-	OptionalText("limited_trial_plan").  // TODO: Not documented
-	OptionalText("retried_on").          // TODO: Not documented
-	OptionalText("scheduled_drop_time"). // TODO: Not documented
+	OptionalBool("is_limited_trial").
```
sfc-gh-jmichalak, line 143: I still can't see these fields in https://docs.snowflake.com/en/sql-reference/sql/desc-listing. Will they be added soon?

pkg/sdk/listings_def.go
```diff
@@ -324,6 +350,19 @@ var ListingsDef = g.NewInterface(
...
+		listingVersion,
+		g.NewQueryStruct("ShowListings").
+			Show().
+			SQL("VERSIONS IN LISTING").
+			Name().
+			OptionalLimit().
+			WithValidation(g.ValidIdentifier, "name"),
 	)
 
 	// TODO(next prs): Organization listing may have its interface, but most of the operations would be pass through functions to the Listings interface
```
sfc-gh-jmichalak, line 369: <p class='graphite__hidden'><i>[Re: line +370]</i></p>
Nit: remove this line.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3873?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/sdk/listings_dto_gen.go
```diff
@@ -76,3 +77,8 @@ type DescribeListingRequest struct {
 	name     AccountObjectIdentifier // required
 	Revision *ListingRevision
 }
+
+type ShowVersionsListingRequest struct {
+	name  AccountObjectIdentifier // required
+	Limit *LimitFrom
```
sfc-gh-jmichalak, line 83: `FROM` is missing from the docs, pls report it.

pkg/sdk/listings_impl_gen.go
```diff
@@ -28,6 +28,16 @@ func (v *listings) Drop(ctx context.Context, request *DropListingRequest) error
 }
 
 func (v *listings) DropSafely(ctx context.Context, id AccountObjectIdentifier) error {
+	// Adjusted manually
+	l, err := v.ShowByIDSafely(ctx, id)
```
sfc-gh-jmichalak: I think this breaks the interface now. When you use `DropSafely` and the object does not exist, it shouldn't fail, right? We could also use ALTER IF EXISTS below.

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+
+	client := testClient(t)
+	ctx := testContext(t)
+
+	accountId := testClientHelper().Context.CurrentAccountId(t)
+	basicManifest := testClientHelper().Listing.BasicManifest(t)
+	testClientHelper().Stage.PutOnStageDirectoryWithContent(t, stage.ID(), "manifest.yml", "basic", basicManifest)
+	basicManifestStageLocation := sdk.NewStageLocation(stage.ID(), "basic/")
+
+	targetAccount := fmt.Sprintf("%s.%s", accountId.OrganizationName(), accountId.AccountName())
```
sfc-gh-jmichalak, line 38: Q: Don't accountId's `Name` or `FullyQualifiedName` work here?

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+title: title
+subtitle: subtitle
+description: description
+listing_terms:
+  type: OFFLINE
+targets:
+  accounts: [%s]
+`, targetAccount)
+	testClientHelper().Stage.PutOnStageDirectoryWithContent(t, stage.ID(), "manifest.yml", "with_target", basicManifestWithTarget)
+	basicManifestWithTargetStageLocation := sdk.NewStageLocation(stage.ID(), "with_target/")
```
sfc-gh-jmichalak: Wow, what a setup 😮  Can we move it to some helpers, as it will be reused in resource and data source acc tests?

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+
+	assertCompleteWithShare := func(t *testing.T, id sdk.AccountObjectIdentifier) {
+		t.Helper()
+
+		listingDetails, err := client.Listings.Describe(ctx, id)
+		assert.NoError(t, err)
+		assert.Equal(t, share.ID().Name(), listingDetails.Share.Name())
+
+		assertThatObject(t,
+			objectassert.Listing(t, id).
```
sfc-gh-jmichalak, line 113: As I understand, these assertions are a bit tricky because we have multiple exclusive options, and there are lots of fields we can't control. If so, we can extract an assertion of non-controllable fields, IMO.

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+				HasState(sdk.ListingStateDraft).
+				HasNoReviewState(),
+		)
+
+		err = client.Listings.Alter(ctx, sdk.NewAlterListingRequest(id).WithReview(true))
+		assert.NoError(t, err)
+
+		assertThatObject(t,
+			objectassert.Listing(t, id).
+				HasState(sdk.ListingStateDraft).
```
sfc-gh-jmichalak, line 260: Q: Is nothing changed? Shouldn't we expect at least a ReviewState?

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+subtitle: different_subtitle
+description: description
+listing_terms:
+  type: OFFLINE
+`
+
+		err := client.Listings.Alter(ctx, sdk.NewAlterListingRequest(listing.ID()).
+			WithAlterListingAs(*sdk.NewAlterListingAsRequest(basicManifestWithDifferentSubtitle).
+				WithPublish(false).
+				WithReview(false).
```
sfc-gh-jmichalak, line 300: Q: Could you explain how this behaves? Above, when we just set the state, these options are exclusive, but here they are not. What if we set them to true, for example? I'm just wondering what we can achieve with both these alters.

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+				WithComment(comment),
+			))
+		assert.NoError(t, err)
+
+		assertThatObject(t,
+			objectassert.Listing(t, listing.ID()).
+				HasSubtitle("different_subtitle").
+				HasNoComment(),
+			// Should be HasComment(comment), but it seems the comment is not set on alter or this comment is set somewhere else
```
sfc-gh-jmichalak: Q: Did you report it? Also, maybe this could work without `PUBLISH` or `REVIEW`, like we had in PATs. Also, it works below, so to clarify: this doesn't work only when the manifest changes as well?

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+		assert.Nil(t, listingDetails.LastCommittedVersionAlias)
+		assert.Nil(t, listingDetails.PublishedVersionUri)
+		assert.Nil(t, listingDetails.PublishedVersionName)
+		assert.Nil(t, listingDetails.PublishedVersionAlias)
+		assert.False(t, *listingDetails.IsShare)
+		assert.Nil(t, listingDetails.RequestApprovalType)
+		assert.Empty(t, *listingDetails.MonetizationDisplayOrder)
+		assert.Empty(t, *listingDetails.LegacyUniformListingLocators)
+	})
+}
```
sfc-gh-jmichalak, line 540: I think `SHOW VERSIONS` is missing.

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+	t.Run("describe: default", func(t *testing.T) {
+		listing, listingCleanup := testClientHelper().Listing.Create(t)
+		t.Cleanup(listingCleanup)
+
+		listingDetails, err := client.Listings.Describe(ctx, listing.ID())
+		require.NoError(t, err)
+		require.NotNil(t, listingDetails)
+
+		assert.NotEmpty(t, listingDetails.GlobalName)
+		assert.Equal(t, listing.ID().Name(), listingDetails.Name)
```
sfc-gh-jmichalak, line 452: Nit: The generator doesn't support DESC natively, but you can temporarily add the DESC struct to `sdk_object_def.go` and generate the assertions.

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+		assert.Nil(t, listingDetails.LastCommittedVersionName)
+		assert.Nil(t, listingDetails.LastCommittedVersionAlias)
+		assert.Nil(t, listingDetails.PublishedVersionUri)
+		assert.Nil(t, listingDetails.PublishedVersionName)
+		assert.Nil(t, listingDetails.PublishedVersionAlias)
+		assert.False(t, *listingDetails.IsShare)
+		assert.Nil(t, listingDetails.RequestApprovalType)
+		assert.Empty(t, *listingDetails.MonetizationDisplayOrder)
+		assert.Empty(t, *listingDetails.LegacyUniformListingLocators)
+	})
```
sfc-gh-jmichalak, line 539: Nit: can you check if passing the `REVISION` field works for `DESC`? I'm fine with a simple NoError check.

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,514 @@
...
+		assert.Nil(t, listingDetails.UsageExamples)
+		assert.Nil(t, listingDetails.DataAttributes)
+		assert.Nil(t, listingDetails.Categories)
+		assert.Nil(t, listingDetails.Resources)
+		assert.Nil(t, listingDetails.Profile)
+		assert.Nil(t, listingDetails.CustomizedContactInfo)
+		assert.Nil(t, listingDetails.DataDictionary)
+		assert.Nil(t, listingDetails.DataPreview)
+		assert.Nil(t, listingDetails.Comment)
+		assert.Equal(t, "DRAFT", listingDetails.Revisions)
```
sfc-gh-jmichalak, line 477: I think `state` is missing.

sfc-gh-jmichalak (changes requested):

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Have to report them. For now, we have what is currently supported (returned by Snowflake).

pkg/sdk/listings_dto_gen.go
sfc-gh-jcieslak: Yes, I have it on my list

pkg/sdk/listings_impl_gen.go
sfc-gh-jcieslak: yeah, adjusted a bit

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: For some reason, it's not. It's in yaml anyway.

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: Hmm, I think it's pretty big because it's a setup for all integration tests. In acceptance tests for resource and data source, the setups should be smaller. If not, I could then pack the common things into a function. For now, I'm not sure what those common operations would be, so I would leave the current setup.

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: Well, the non-controllable fields are asserted in 3 common assert functions, so I think we could leave them just as proof that the mapping from Snowflake output works. The rest of the tests are just targeting specific assertions relevant to a given test. In here (if that helps) we could separate the assertions into controllable and non-controllable ones.

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: This is strictly related to the manifest configuration, for configurations without target accounts in manifest will contain review state "UNSENT", but they cannot be published. It would fail with the following error: `        	            	090881 (22023): Listing cannot be published: Missing listing fields: Business needs, Usage examples or Notebooks, Data Attributes, Documentation resources, Data Dictionary, Categories.`. I'm not yet sure if those statuses will be relevant, but I didn't want to spend more time configuring listings to behave in certain way (because I couldn't guess by reading docs).

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: Alter comment works, I'll try more variations, but I think this comment may be setting something else than listing comment (e.g. add version comment adds a comment to a given version, not the listing itself). All inconsistencies will be reported.

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: SHOW VERSIONS is called during add version, I'll extend asserts there

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: Yeah, I guess I missed that, but due to difficult setup, for now, I created a follow-up ticket so we don't stop on this. If it will be required for resource development, I'll add it.

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: It's above application package (with comment) and share

pkg/sdk/listings_def.go
sfc-gh-jcieslak: Removed SHOW VERSIONS comment

pkg/acceptance/helpers/listing_client.go
sfc-gh-jcieslak: Right, removed checks here

pkg/acceptance/helpers/listing_client.go
sfc-gh-jcieslak: They don't have to be unique afaik (you can create two listings with the same name, but I'm not sure if this would be an issue later on, but even if yes, the customer would be able to adjust the manifest as there is alter for it)

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: Let's continue without it as I have only one test with desc (second one incoming, but not all asserts will be used)

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: I'm not sure If I understand :/ could you elaborate a bit?

pkg/acceptance/helpers/listing_client.go
```diff
@@ -0,0 +1,80 @@
...
```
sfc-gh-asawicki: nit: I see that it does not matter but we could probably have here a name that allows us to rack e.g. from which test run the listing comes. So I would use here some object suffix added.

pkg/acceptance/helpers/stage_client.go
```diff
@@ -160,12 +160,17 @@ func (c *StageClient) RemoveFromStageFunc(t *testing.T, stageLocation string, pa
...
 	t.Helper()
 	ctx := context.Background()
 
 	filePath := testhelpers.TestFile(t, filename, []byte(content))
 
-	_, err := c.context.client.ExecForTests(ctx, fmt.Sprintf(`PUT file://%s @%s AUTO_COMPRESS = FALSE OVERWRITE = TRUE`, filePath, id.FullyQualifiedName()))
+	_, err := c.context.client.ExecForTests(ctx, fmt.Sprintf(`PUT file://%s @%s/%s AUTO_COMPRESS = FALSE OVERWRITE = TRUE`, filePath, id.FullyQualifiedName(), targetDirectory))
 	require.NoError(t, err)
 	t.Cleanup(func() {
 		_, err = c.context.client.ExecForTests(ctx, fmt.Sprintf(`REMOVE @%s/%s`, id.FullyQualifiedName(), filename))
```
sfc-gh-asawicki: I think this cleanup is incorrect after the change you made (the directory path is missing)

pkg/acceptance/helpers/stage_client.go
```diff
@@ -160,12 +160,17 @@ func (c *StageClient) RemoveFromStageFunc(t *testing.T, stageLocation string, pa
...
+}
+
+func (c *StageClient) PutOnStageDirectoryWithContent(t *testing.T, id sdk.SchemaObjectIdentifier, filename string, targetDirectory string, content string) {
 	t.Helper()
 	ctx := context.Background()
 
 	filePath := testhelpers.TestFile(t, filename, []byte(content))
 
-	_, err := c.context.client.ExecForTests(ctx, fmt.Sprintf(`PUT file://%s @%s AUTO_COMPRESS = FALSE OVERWRITE = TRUE`, filePath, id.FullyQualifiedName()))
+	_, err := c.context.client.ExecForTests(ctx, fmt.Sprintf(`PUT file://%s @%s/%s AUTO_COMPRESS = FALSE OVERWRITE = TRUE`, filePath, id.FullyQualifiedName(), targetDirectory))
```
sfc-gh-asawicki: BTW, you have the PutInLocationWithContent which I guess should be used instead. (yes, we need to clean these helpers for stage files management)

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,550 @@
...
+	ctx := testContext(t)
+
+	accountId := testClientHelper().Context.CurrentAccountId(t)
+	basicManifest := testClientHelper().Listing.BasicManifest(t)
+	testClientHelper().Stage.PutOnStageDirectoryWithContent(t, stage.ID(), "manifest.yml", "basic", basicManifest)
+	basicManifestStageLocation := sdk.NewStageLocation(stage.ID(), "basic/")
+
+	targetAccount := fmt.Sprintf("%s.%s", accountId.OrganizationName(), accountId.AccountName())
+	basicManifestWithTarget := fmt.Sprintf(`
+title: title
```
sfc-gh-asawicki: nit: we have this manifest in the chest client helper, can we get it from there?

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,550 @@
...
+subtitle: subtitle
+description: description
+listing_terms:
+  type: OFFLINE
+targets:
+  accounts: [%s]
+`, targetAccount)
+	testClientHelper().Stage.PutOnStageDirectoryWithContent(t, stage.ID(), "manifest.yml", "with_target", basicManifestWithTarget)
+	basicManifestWithTargetStageLocation := sdk.NewStageLocation(stage.ID(), "with_target/")
```
sfc-gh-asawicki: nit: is the `/` needed after `with_target`?

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,550 @@
...
+		listing, listingCleanup := testClientHelper().Listing.Create(t)
+		t.Cleanup(listingCleanup)
+
+		assertThatObject(t,
+			objectassert.ListingFromObject(t, listing).
+				HasSubtitle("subtitle").
+				HasNoComment(),
+		)
+
+		basicManifestWithDifferentSubtitle := `
```
sfc-gh-asawicki: nit: let's move it to helpers too

pkg/sdk/testint/listings_integration_test.go
```diff
@@ -0,0 +1,550 @@
...
+		assertThatObject(t,
+			objectassert.Listing(t, listing.ID()).
+				HasSubtitle("different_subtitle").
+				HasNoComment(),
+			// Should be HasComment(comment), but it seems the comment is not set on alter or this comment is set somewhere else
+		)
+	})
+
+	t.Run("alter: add version", func(t *testing.T) {
+		basicWithDifferentSubtitleManifest := `
```
sfc-gh-asawicki: nit: and this

pkg/acceptance/helpers/stage_client.go
sfc-gh-jcieslak: You're right, fixed

pkg/acceptance/helpers/stage_client.go
sfc-gh-jcieslak: Reverted changes here and used existing functions

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: Not quite the same, but moved this config there and retrieved it here

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: Checked, they're not needed, removed

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: moved

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jcieslak: done

pkg/acceptance/helpers/listing_client.go
sfc-gh-jcieslak: Not sure if that will be needed, but added

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jmichalak: Okay, please also write it in the code (next PR).

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jmichalak: I'm fine with leaving this for now, but it would be helpful when we have more tests.

pkg/acceptance/helpers/listing_client.go
```diff
@@ -0,0 +1,108 @@
...
+description: "description"
+listing_terms:
+  type: "OFFLINE"
+`, title), title
+}
+
+func (c *ListingClient) BasicManifestWithDifferentSubtitle(t *testing.T) (string, string) {
+	t.Helper()
+	title := c.ids.WithTestObjectSuffix("with_diff_subtitle_")
+	return fmt.Sprintf(`title: "%s"
```
sfc-gh-jmichalak, line 88: Nit: not sure what we want here exactly, but the title will be changed as well. If we want to have only a different subtitle, we should pass the title as an argument.

pkg/sdk/testint/listings_integration_test.go
sfc-gh-jmichalak: I meant that here we can set both publish and review options (although, to false), but in `alter: change state`, we do it separately, so I was wondering what's the difference here. Let's discuss off-line.

sfc-gh-jmichalak (approved):

pkg/acceptance/helpers/listing_client.go
sfc-gh-jcieslak: Yes, I know. This refers to this: https://github.com/snowflakedb/terraform-provider-snowflake/pull/3873#discussion_r2225695725 comment. I wanted to have a title that would be meaningful so that we could match by title, which test, e.g., didn't clean up the listing properly. I don't think changing the title should impact the test result (as I run the tests, it didn't matter).

sfc-gh-jmichalak (approved):

PR #3876, opened 2025-07-23
chore: Adjust sweepers and test workflows

sfc-gh-jcieslak:
## Changes
- Add network policy sweeper
- Skip file filtering for repository_dispatch events in test workflows

### DIFF
--- a/.github/workflows/tests-account-level.yml
+++ b/.github/workflows/tests-account-level.yml
@@ -30,6 +30,7 @@ jobs:
       - name: Filter changed files
         uses: dorny/paths-filter@v3
         id: changes
+        if: github.event_name != 'repository_dispatch'
         with:
           filters: |
             run_tests:
--- a/pkg/sdk/sweepers.go
+++ b/pkg/sdk/sweepers.go
@@ -84,7 +84,7 @@ func getNetworkPolicySweeper(client *Client, suffix string) func() error {
 		log.Printf("[DEBUG] Sweeping network policies with suffix %s", suffix)
 		ctx := context.Background()
 
-		nps, err := client.NetworkPolicies.Show(ctx, nil)
+		nps, err := client.NetworkPolicies.Show(ctx, NewShowNetworkPolicyRequest())
 		if err != nil {
 			return fmt.Errorf("SHOW NETWORK POLICIES ended with error, err = %w", err)
 		}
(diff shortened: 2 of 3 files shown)


.github/workflows/tests-account-level.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       - name: Filter changed files
         uses: dorny/paths-filter@v3
         id: changes
+        if: github.event_name != 'repository_dispatch'
```
sfc-gh-jmichalak, line 33: Could you explain why we need this?

pkg/sdk/sweepers.go
```diff
@@ -88,6 +79,31 @@ func getResourceMonitorSweeper(client *Client, suffix string) func() error {
...
+
+		nps, err := client.NetworkPolicies.Show(ctx, NewShowNetworkPolicyRequest())
+		if err != nil {
+			return fmt.Errorf("SHOW NETWORK POLICIES ended with error, err = %w", err)
+		}
+
+		for _, np := range nps {
+			if strings.HasSuffix(np.Name, suffix) && strings.ToUpper(np.Name) != "RESTRICTED_ACCESS" {
+				log.Printf("[DEBUG] Dropping network policy %s", np.ID().FullyQualifiedName())
+				if err := client.NetworkPolicies.Drop(ctx, NewDropNetworkPolicyRequest(np.ID()).WithIfExists(true)); err != nil {
```
sfc-gh-jmichalak, line 98: Will it work? When I tried to drop a network policy with a set network rule, I still got an error. I think we must first unassign network rules from network policies, and then (maybe it's not even necessary) remove the network rules and policies.

.github/workflows/tests-account-level.yml
sfc-gh-jcieslak: It's needed to fix ok-to-test, if you now try to run the tests on community pr the tests will fail because git command issued internally in this action doesn't have access/permission or just doesn't meet its expectations. As we always want to run the tests which are triggered by us, no filtering is needed.

pkg/sdk/sweepers.go
sfc-gh-jcieslak: Added a note on the sweeping function

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

PR #3880, opened 2025-07-23
feat: Add resource_constraint to the warehouse resource

danielford831:
## Changes
- Remove the note about the missing field. Note that the `generation` field will be added in a follow-up, with improved documentation.
- Add the `resource_constraint` field in `warehouse` resource, without `generation`-related values.
- Add acceptance tests, including migration asssertions.
- Add new helper client methods.
- Add/generate model builders and assertions.
- Adjust the data source acceptance tests.
- Update the example configuration.

## References
https://github.com/snowflakedb/terraform-provider-snowflake/issues/3258
https://docs.snowflake.com/en/sql-reference/sql/create-warehouse#optional-properties-objectproperties

## TODO
- Add the `generation` attribute.
- Report not working UNSET.
- Add the migration guide entry and update the resource documentation (and point out to limitations/valid configurations).


sfc-gh-asawicki:
Hey @danielford831. Thanks for making the contribution.

The review may take a bit longer as the team is smaller (due to the holiday season), we are working on a roadmap for the next months, and we need to consult the warehouse team internally before approving the PR. We will get back to you as soon as we sort these out. Thanks for understanding!

papaonlegs:
Hey @danielford831, here's a PR I made with Claude before @sfc-gh-asawicki pointed this one out.

It has some tests in there, if you want to use those -> https://github.com/snowflakedb/terraform-provider-snowflake/pull/3910

pkg/testacc/resource_warehouse_acceptance_test.go
```diff
@@ -2131,3 +2136,422 @@ resource "snowflake_warehouse" "test" {
...
+					resourceshowoutputassert.WarehouseShowOutput(t, warehouseModelDefault.ResourceReference()).
+						HasType(sdk.WarehouseTypeStandard).
+						HasResourceConstraint(sdk.WarehouseResourceConstraintStandardGen1),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_Warehouse_ResourceConstraint_MigrateManuallySetResourceConstraint(t *testing.T) {
```
sfc-gh-asawicki, line 2481: Let's also test migration without setting it in the config (also to veridfy the behavior in show_output when new field was added, as I am unsure if we are handling it correctly).

pkg/sdk/warehouses_test.go
```diff
@@ -274,6 +274,48 @@ func TestWarehouseDescribe(t *testing.T) {
...
+		input string
+		want  WarehouseResourceConstraint
+	}
+
+	valid := []test{
+		// Case insensitive.
+		{input: "standard_gen_1", want: WarehouseResourceConstraintStandardGen1},
+
+		// Supported Values.
+		{input: "STANDARD_GEN_1", want: WarehouseResourceConstraintStandardGen1},
```
sfc-gh-asawicki, line 288: Let's also test the WithoutGen method

pkg/resources/warehouse.go
```diff
@@ -111,6 +112,13 @@ var warehouseSchema = map[string]*schema.Schema{
 		Description:      "Specifies the maximum scale factor for leasing compute resources for query acceleration. The scale factor is used as a multiplier based on warehouse size.",
 		Default:          IntDefault,
 	},
+	"resource_constraint": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ValidateDiagFunc: sdkValidation(sdk.ToWarehouseResourceConstraintAllowedInput),
+		DiffSuppressFunc: SuppressIfAny(NormalizeAndCompare(sdk.ToWarehouseResourceConstraint), IgnoreChangeToCurrentSnowflakeValueInShow("resource_constraint")),
```
sfc-gh-asawicki: shouldn't we use `ToWarehouseResourceConstraintAllowedInput` here?

pkg/sdk/warehouses.go
```diff
@@ -132,6 +132,76 @@ func (e ScalingPolicy) FromString(s string) (ScalingPolicy, error) {
...
+	case "MEMORY_64X":
+		return WarehouseResourceConstraintMemory64X, nil
+	case "MEMORY_64X_X86":
+		return WarehouseResourceConstraintMemory64Xx86, nil
+	default:
+		return "", fmt.Errorf("invalid resource constraint: %s", s)
+	}
+}
+
+func ToWarehouseResourceConstraintAllowedInput(s string) (WarehouseResourceConstraint, error) {
```
sfc-gh-asawicki: nit: `AllowedInput` sounds odd. Let's change it e.g. explicitly to `WithoutGeneration` or something else motivating the distinction

pkg/sdk/warehouses.go
```diff
@@ -132,6 +132,76 @@ func (e ScalingPolicy) FromString(s string) (ScalingPolicy, error) {
...
+	case "MEMORY_64X":
+		return WarehouseResourceConstraintMemory64X, nil
+	case "MEMORY_64X_X86":
+		return WarehouseResourceConstraintMemory64Xx86, nil
+	default:
+		return "", fmt.Errorf("invalid resource constraint: %s", s)
+	}
+}
+
+var AllWarehouseResourceConstraints = []string{
```
sfc-gh-asawicki: nit: these are not All, these are the ones without gen

sfc-gh-asawicki (changes requested):

pkg/resources/warehouse.go
```diff
@@ -402,6 +420,14 @@ func GetReadWarehouseFunc(withExternalChangesMarking bool) schema.ReadContextFun
 		}
 
 		if withExternalChangesMarking {
+			resourceConstraint := ""
+			if w.ResourceConstraint != nil {
+				if v, err := sdk.ToWarehouseResourceConstraintAllowedInput(string(*w.ResourceConstraint)); err != nil {
```
sfc-gh-jcieslak: Wait, is it necessary to map a resource constraint to the string just to map it again to a resource constraint? It should already be in the form of an enum and `ToWarehouseResourceConstraintAllowedInput` is probably used in convert meaning the errors would be propagated properly if it returns an error, correct?

pkg/resources/warehouse.go
```diff
@@ -585,6 +613,32 @@ func UpdateWarehouse(ctx context.Context, d *schema.ResourceData, meta any) diag
...
+						return diag.FromErr(err)
+					}
+					set.ResourceConstraint = &resourceConstraint
+				} else {
+					// TODO [SNOW-1473453]: UNSET of resource constraint does not work
+					// unset.ResourceConstraint = sdk.Bool(true)
+					set.ResourceConstraint = sdk.Pointer(sdk.WarehouseResourceConstraintMemory16X)
+				}
+			} else {
+				log.Printf("[DEBUG] resource constraint is not supported for %s warehouses, ignoring", warehouseType)
```
sfc-gh-jcieslak, line 654: nit: shouldn't we return a warning in this case?

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -399,6 +431,46 @@ func TestInt_Warehouses(t *testing.T) {
...
+		})
+		t.Cleanup(warehouseCleanup)
+
+		returnedWarehouse, err := client.Warehouses.ShowByID(ctx, warehouse.ID())
+		require.NoError(t, err)
+		assert.Equal(t, sdk.WarehouseTypeSnowparkOptimized, returnedWarehouse.Type)
+		assert.Equal(t, sdk.WarehouseResourceConstraintMemory16X, *returnedWarehouse.ResourceConstraint)
+
+		alterOptions := &sdk.AlterWarehouseOptions{
+			Set: &sdk.WarehouseSet{ResourceConstraint: sdk.Pointer(sdk.WarehouseResourceConstraintMemory1X)},
```
sfc-gh-jcieslak, line 449: question: Can it be set with other parameters? The resource implementation suggests that yes, but making sure.

pkg/resources/warehouse.go
sfc-gh-jmichalak: Yes, done (with a rename).

pkg/sdk/warehouses.go
sfc-gh-jmichalak: Renamed.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: Done.

pkg/sdk/warehouses_test.go
sfc-gh-jmichalak: Done.

pkg/resources/warehouse.go
sfc-gh-jmichalak: No, because it would break handling standard warehouses for now.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jmichalak: Yes, but for the STANDARD type. Altering the type is extracted already, so I didn't want to modify the existing tests more.

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/warehouse.go
sfc-gh-jmichalak: It's not - changed. The invalid constraint would be detected in the SDK, but here we also check it against the warehouse type, though it's probably unlikely to mismatch in Snowflake.

pkg/resources/warehouse.go
```diff
@@ -294,8 +294,14 @@ func ImportWarehouse(ctx context.Context, d *schema.ResourceData, meta any) ([]*
...
-	if err = d.Set("resource_constraint", w.ResourceConstraint); err != nil {
-		return nil, err
+	// Handle resource_constraint
+	switch w.Type {
+	case sdk.WarehouseTypeSnowparkOptimized:
+		if err = d.Set("resource_constraint", w.ResourceConstraint); err != nil {
+			return nil, err
+		}
+	default:
+		log.Printf("[DEBUG] handling resource_constraint for warehouse type %s is not supported, ignoring", w.Type)
```
sfc-gh-jcieslak, line 310: question: Should we do nothing or set it to nil in this case?

pkg/testacc/resource_warehouse_acceptance_test.go
```diff
@@ -2534,3 +2531,97 @@ func TestAcc_Warehouse_ResourceConstraint_MigrateManuallySetResourceConstraint(t
...
+						HasNoResourceConstraint(),
+				),
+			},
+			{
+				Config:                   config.FromModels(t, warehouseModelSnowparkOptimized),
+				ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(warehouseModelSnowparkOptimized.ResourceReference(), plancheck.ResourceActionUpdate),
+						planchecks.ExpectChange(warehouseModelSnowparkOptimized.ResourceReference(), "resource_constraint", tfjson.ActionUpdate, sdk.String(string(sdk.WarehouseResourceConstraintMemory16X)), nil),
```
sfc-gh-asawicki: We should aim for having a no-op here, as:
- this is a stable resource
- it does not involve any action

We can either do it through a state upgrader and/or modification in show_output logic (i.e. if key was missing at all and is present, the change should not be triggered or smth). We can discuss this one internally.

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-jmichalak: Done with a state upgrader.

pkg/resources/warehouse_state_upgraders.go
```diff
@@ -90,3 +91,42 @@ func v092WarehouseSizeStateUpgrader(ctx context.Context, rawState map[string]any
...
+// - we are adding resource_constraint to the show output and to the config
+// - this means that the old output in handleExternalChangesToObject would be empty after migration
+// - that caused the plan to show a diff for resource_constraint
+// - this upgrade fixes that by setting the default value for resource_constraint
+func v2_6_0_WarehouseResourceConstraintUpgrader(ctx context.Context, rawState map[string]any, meta any) (map[string]any, error) {
+	if rawState == nil {
+		return rawState, nil
+	}
+
+	warehouseTypeRaw, ok := rawState["warehouse_type"].(string)
```
sfc-gh-asawicki: This can be empty (you can create a warehouse without specifying it, and it won't be visible in the state with our current logic).

pkg/resources/warehouse_state_upgraders.go
```diff
@@ -90,3 +91,42 @@ func v092WarehouseSizeStateUpgrader(ctx context.Context, rawState map[string]any
...
+		return nil, fmt.Errorf("cannot read warehouse_type from the state")
+	}
+	warehouseType, err := sdk.ToWarehouseType(warehouseTypeRaw)
+	if err != nil {
+		return nil, err
+	}
+
+	oldShowOutputRaw, ok := rawState[ShowOutputAttributeName].([]any)
+	if !ok || len(oldShowOutputRaw) != 1 {
+		return nil, fmt.Errorf("expected exactly one warehouse show output; got %d", len(oldShowOutputRaw))
```
sfc-gh-asawicki: I guess that if there is non show_output, then we don't need to add resource constraint to it, and we are fine, so this should just return?

pkg/resources/warehouse_state_upgraders.go
```diff
@@ -90,3 +91,42 @@ func v092WarehouseSizeStateUpgrader(ctx context.Context, rawState map[string]any
...
+	}
+	oldShowOutput, ok := oldShowOutputRaw[0].(map[string]any)
+	if !ok {
+		return nil, fmt.Errorf("cannot read warehouse show output from the state")
+	}
+
+	switch warehouseType {
+	case sdk.WarehouseTypeSnowparkOptimized:
+		oldShowOutput["resource_constraint"] = string(sdk.WarehouseResourceConstraintMemory16X)
```
sfc-gh-asawicki: why we use a hardcoded one here? I think we need to fetch it from SF (e.g. in cases that someone set it by external means - manually or through snowflake_execute).

sfc-gh-asawicki: Please test such a case too.

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-asawicki: Left comments

pkg/resources/warehouse.go
sfc-gh-jmichalak: I think it does not matter. This func will still be modified in the follow-up.

pkg/resources/warehouse_state_upgraders.go
sfc-gh-jmichalak: Now, it's read from the SHOW.

pkg/resources/warehouse_state_upgraders.go
sfc-gh-jmichalak: Are there any cases with missing `show_output`? This is more of a sanity check. If someone upgrades before importing the warehouse, the upgrader will return early anyway.

pkg/resources/warehouse_state_upgraders.go
sfc-gh-jmichalak: Added.

pkg/resources/warehouse_state_upgraders.go
sfc-gh-asawicki: but why with an error? for this upgrader, which cares only for existing show_outpu, it should return without error.

pkg/resources/warehouse_state_upgraders.go
```diff
@@ -120,11 +111,24 @@ func v2_6_0_WarehouseResourceConstraintUpgrader(ctx context.Context, rawState ma
...
+
+	if warehouseInSnowflake.ResourceConstraint != nil {
+		switch warehouseInSnowflake.Type {
+		case sdk.WarehouseTypeSnowparkOptimized:
+			oldShowOutput["resource_constraint"] = string(*warehouseInSnowflake.ResourceConstraint)
+		default:
+			log.Printf("[DEBUG] handling resource_constraint for warehouse type %s is not supported, ignoring", warehouseInSnowflake.Type)
+		}
 	}
 	rawState[ShowOutputAttributeName] = []any{oldShowOutput}
```
sfc-gh-asawicki: nit: this line could be only invoked when we made changes to show_output

pkg/resources/warehouse_state_upgraders.go
sfc-gh-jmichalak: Changed.

pkg/resources/warehouse_state_upgraders.go
```diff
@@ -90,3 +91,46 @@ func v092WarehouseSizeStateUpgrader(ctx context.Context, rawState map[string]any
...
+	warehouseInSnowflake, err := client.Warehouses.ShowByIDSafely(ctx, sdk.NewAccountObjectIdentifier(warehouseName))
+	if err != nil {
+		return nil, err
+	}
+
+	if warehouseInSnowflake.ResourceConstraint != nil {
+		switch warehouseInSnowflake.Type {
+		case sdk.WarehouseTypeSnowparkOptimized:
+			oldShowOutput["resource_constraint"] = string(*warehouseInSnowflake.ResourceConstraint)
+		default:
```
sfc-gh-asawicki: one more question - why we skip here setting the resource constraint in show output? We don't skip setting it in the show output when the resource is read, right?

sfc-gh-jmichalak: Because the other type would be supported in the next PR. Changed.

pkg/resources/warehouse_state_upgraders.go
sfc-gh-jmichalak: Done.

PR #3881, opened 2025-07-24
chore: Sweepers and Test workflow changes

sfc-gh-jcieslak:
## Changes
- Adjust sweepers to include user sweeping (to avoid policy related issues)
- Change test workflow's test object suffix so that it's unique even if test are re-run
  - This is important in the context of failed cleanup, on test workflow re-run we may encounter "object already exists" errors. By having unique object prefix this could be avoided (and because we have stale object sweepers, leftovers will be cleaned up anyway)

### DIFF
--- a/.github/workflows/tests-account-level.yml
+++ b/.github/workflows/tests-account-level.yml
@@ -22,7 +22,7 @@ jobs:
     environment: test
     env:
       ## adding additional suffix to don't have conflict in non-account level tests and their pre-created objects
-      TEST_SF_TF_TEST_OBJECT_SUFFIX: "${{ (github.event_name == 'repository_dispatch' && github.event.client_payload.pull_request.head.sha) || (github.event_name == 'pull_request' && github.event.pull_request.head.sha) }}_${{ github.run_attempt }}AL"
+      TEST_SF_TF_TEST_OBJECT_SUFFIX: ${{ ((github.event_name == 'repository_dispatch' && github.event.client_payload.pull_request.head.sha) || (github.event_name == 'pull_request' && github.event.pull_request.head.sha)) && 'AL' }}
     name: Run all account-level tests
     runs-on: ubuntu-latest
     if: (github.event_name == 'repository_dispatch') || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository)
--- a/pkg/sdk/sweepers_test.go
+++ b/pkg/sdk/sweepers_test.go
@@ -48,7 +48,6 @@ func SweepAfterAcceptanceTests(client *Client, suffix string) error {
 	return sweep(client, suffix)
 }
 
-// TODO [SNOW-867247]: move this to test code
 // TODO [SNOW-867247]: use if exists/use method from helper for dropping
 // TODO [SNOW-867247]: sweep all missing account-level objects (like users, integrations, replication groups, network policies, ...)
 // TODO [SNOW-867247]: extract sweepers to a separate dir
@@ -285,12 +284,12 @@ func nukeUsers(client *Client, suffix string) func() error {
 		if suffix != "" {
 			log.Printf("[DEBUG] Sweeping users with suffix %s", suffix)
 			userDropCondition = func(u User) bool {
-				return strings.HasSuffix(u.Name, suffix) && !slices.Contains(protectedUsers, u.ID().Name())
+				return strings.HasSuffix(u.Name, suffix)
 			}
 		} else {
 			log.Println("[DEBUG] Sweeping stale users")
 			userDropCondition = func(u User) bool {
-				return !slices.Contains(protectedUsers, u.Name) && u.CreatedOn.Before(time.Now().Add(-15*time.Minute))
+				return u.CreatedOn.Before(time.Now().Add(-15 * time.Minute))
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


sfc-gh-jmichalak (dismissed):

pkg/sdk/sweepers.go
```diff
@@ -107,6 +108,48 @@ func getNetworkPolicySweeper(client *Client, suffix string) func() error {
 	}
 }
 
+// getUserSweeper was introduced to make sure that users created during tests are cleaned up.
+// It's required as users that have connections to policies within databases, block deletion of the whole database.
+// In Snowflake, the users can be removed without prior unsetting policies, but policies cannot be removed without firstly unlinking them from other objects.
+func getUserSweeper(client *Client, suffix string) func() error {
```
sfc-gh-asawicki: We already have a nukeUsers function which does exactly the same

sfc-gh-jcieslak: Yeah, but we have the same with databases and warehouses. The problem here was that the sweepers failed in the TestSweepAll test, which was intended to remove a database that contained a policy attached to a user. By adding a user sweeper, we'll eliminate such errors. Maybe we could find some way to know which test generated a given identifier that wasn't properly cleared 🤔, but for now, this should be sufficient to stop these kinds of errors.

sfc-gh-asawicki: I know, but what we have with databases and warehouses is not good either. Additionally, the ones that are outside these sweepers are better, as they also handle the ownership changes. I am not saying we should rework the sweepers now but maybe we can add the optional suffix to the existing method and delete either by suffiix (if present) or by the date when absent?

sfc-gh-jcieslak: Moved nuke functions to sweepers. Go and refactor the user so that there's only one nukeUsers function. Nuke users can be a good candidate for making generic nuke function.

sfc-gh-jcieslak: Can add in next pr a more generic nukeObject version to avoid copy-paste here (already have somewhere a version with a suffix).

pkg/sdk/sweepers.go
```diff
@@ -275,3 +235,172 @@ func getRoleSweeper(client *Client, suffix string) func() error {
 		return nil
 	}
 }
+
+// TODO [SNOW-867247]: generalize nuke methods (sweepers too)
+// TODO [SNOW-1658402]: handle the ownership problem while handling the better role setup for tests
+func nukeWarehouses(client *Client, prefix string) func() error {
```
sfc-gh-asawicki: please leave the nuke* methods within the test code (sweepert_test).

Please move the `SweepAfterIntegrationTests`, `SweepAfterAcceptanceTests`, and `sweep` instead (first, as this shouldn't be a production code in the first place; second, as it will be easier to review the changes you did while merging old sweepers with the nukes).

sfc-gh-asawicki (changes requested):

pkg/sdk/sweepers.go
sfc-gh-jcieslak: Done

.github/workflows/tests-account-level.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     environment: test
     env:
       ## adding additional suffix to don't have conflict in non-account level tests and their pre-created objects
-      TEST_SF_TF_TEST_OBJECT_SUFFIX: ${{ ((github.event_name == 'repository_dispatch' && github.event.client_payload.pull_request.head.sha) || (github.event_name == 'pull_request' && github.event.pull_request.head.sha)) && 'AL' }}
+      TEST_SF_TF_TEST_OBJECT_SUFFIX: "${{ (github.event_name == 'repository_dispatch' && github.event.client_payload.pull_request.head.sha) || (github.event_name == 'pull_request' && github.event.pull_request.head.sha) }}_${{ github.run_attempt }}AL"
```
sfc-gh-asawicki: let's not add the run_attempt yet (as this changes the sweepers behavior and we should discuss it first)

pkg/sdk/sweepers_test.go
```diff
@@ -39,6 +40,45 @@ func TestSweepAll(t *testing.T) {
...
 
+func SweepAfterIntegrationTests(client *Client, suffix string) error {
+	return sweep(client, suffix)
+}
+
+func SweepAfterAcceptanceTests(client *Client, suffix string) error {
+	return sweep(client, suffix)
+}
+
+// TODO [SNOW-867247]: move this to test code
```
sfc-gh-asawicki: this TODO is done by this move

pkg/sdk/sweepers_test.go
```diff
@@ -239,27 +279,42 @@ func nukeUsers(client *Client) func() error {
...
 	return func() error {
-		log.Println("[DEBUG] Nuking users")
 		ctx := context.Background()
 
-		users, err := client.Users.Show(ctx, &ShowUserOptions{})
+		var userDropCondition func(u User) bool
+		if suffix != "" {
+			log.Printf("[DEBUG] Sweeping users with suffix %s", suffix)
+			userDropCondition = func(u User) bool {
+				return strings.HasSuffix(u.Name, suffix) && !slices.Contains(protectedUsers, u.ID().Name())
```
sfc-gh-asawicki: nit: the protected users condition should be there no matter the condition, so maybe let's extract this part to the if below?

sfc-gh-asawicki (changes requested):

.github/workflows/tests-account-level.yml
sfc-gh-jcieslak: Ok, reverted

pkg/sdk/sweepers_test.go
sfc-gh-jcieslak: Sure, extracted

PR #3885, opened 2025-07-24
chore: Listing basic resource with autogenerated content

sfc-gh-jcieslak:
## Changes
- Prepare a schema only with name type, and basic validation
- Generate model, assertions, and show schema
  - Important note about model: It was generated with dry-run and copied manually into a file that doesn't end with _gen (because otherwise it would be removed when calling generate). It was done because the generator was throwing errors, because the listing schema contains a list that is required, which produced invalid model constructors.
- This can also be a good place to discuss if we would like to go with this schema, my take on the current one:
  - There are two/three main things to take into consideration (manifest and its versioning, listing state)
    -  Manifest can be either sourced from a stage or inlined as a string. The source can be freely altered (e.g. going from inlined to stage and vice-versa). To add a new version in inlined fashion, you can just change the value in the configuration, and we would call ALTER LISTING AS and pass the new manifest (the Snowflake then generated a new version automatically, but without version alias). When the manifest is sourced from the stage a new version is explicitly created with version name (in `SHOW VERSIONS IN LISTING` it's returned as `alias`). A new version will be triggered by:
       - Specifying a new version name
       - Going from inlined approach to the stage one
       - Whenever the stage identifier or stage location changes
  - The inlined and stage versions differ in SQL (I'm talking about ALTER AS vs ALTER ADD VERSION). The inlined one can directly specify if the new version should be reviewed/published, but the stage version does not have it. I already asked the listing team how this functionality corresponds to the stage version. I'm guessing that calling separate ALTERs with REVIEW and/or PUBLISH commands should be similar to `ALTER LISTING AS .. REVIEW = x PUBLISH = y' behavior. If my assumptions are correct, then the top-level review and publish "booleans" should be sufficient as they will be referring to the state of the latest version. We could also specify this as a limitation, that the listing state (review and publish state) will be only considered for the latest version. Seeing our customers' usage we could extend its functionality.

## Next prs
- Update the schema (add/update descriptions / diff suppresses / validations)
- Fill the resource implementation
- Add acceptance tests

Reference: [ALTER LISTING](https://docs.snowflake.com/en/sql-reference/sql/alter-listing)

### DIFF
--- a/pkg/acceptance/bettertestspoc/assert/resourceassert/gen/resource_schema_def.go
+++ b/pkg/acceptance/bettertestspoc/assert/resourceassert/gen/resource_schema_def.go
@@ -105,7 +105,11 @@ var allResourceSchemaDefs = []ResourceSchemaDef{
 		name:   "LegacyServiceUser",
 		schema: resources.LegacyServiceUser().Schema,
 	},
-	// TODO(SNOW-1501905): Support the rest of attribute types (Commented because it's blocking model generator)
+	// TODO(SNOW-1501905): Support required object types (commented because it's blocking model generator)
+	// Manifest field (required object type) is not supported by the model generator.
+	// Current overrides are not sufficient to generate the model with the required attribute that would use tfconfig.Variable type.
+	// The WithManifestValue method should be used in the constructors as WithManifest is not generated at all (another generator limitation, but not that important in this case).
+	// Once it's supported, uncomment this and remove the manually written model.
 	// {
 	//	name:   "Listing",
 	//	schema: resources.Listing().Schema,
--- a/pkg/acceptance/bettertestspoc/config/model/listing_model.go
+++ b/pkg/acceptance/bettertestspoc/config/model/listing_model.go
@@ -24,9 +24,11 @@ type ListingModel struct {
 	*config.ResourceModelMeta
 }
 
-// TODO(SNOW-1501905): Adjusted manually (because the model generator does not support required list attributes yet).
-// Current overrides are not sufficient to generate the model with the required list attribute that would use tfconfig.Variable type.
-// The WithManifestValue method should be used in this case as WithManifest is not generated at all.
+// TODO(SNOW-1501905): Support required object types (written manually, because it's blocking model generator)
+// Manifest field (required object type) is not supported by the model generator.
+// Current overrides are not sufficient to generate the model with the required attribute that would use tfconfig.Variable type.
+// The WithManifestValue method should be used in the constructors as WithManifest is not generated at all (another generator limitation, but not that important in this case).
+// Once it's supported, remove this file and generate the model with generator (adjust resource_schema_def.go).
 
 func ListingWithInlineManifest(
 	resourceName string,


pkg/acceptance/bettertestspoc/config/model/listing_model.go
```diff
@@ -0,0 +1,173 @@
...
+	Publish            tfconfig.Variable `json:"publish,omitempty"`
+	Review             tfconfig.Variable `json:"review,omitempty"`
+	Share              tfconfig.Variable `json:"share,omitempty"`
+
+	DynamicBlock *config.DynamicBlock `json:"dynamic,omitempty"`
+
+	*config.ResourceModelMeta
+}
+
+// TODO(SNOW-1501905): Adjusted manually (because the model generator does not support required list attributes yet).
```
sfc-gh-asawicki: I don't think it's true. Check pkg/acceptance/bettertestspoc/config/model/gen/overrides.go

pkg/resources/listing.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var listingSchema = map[string]*schema.Schema{
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		DiffSuppressFunc: suppressIdentifierQuoting,
+		Description:      "Specifies the listing identifier (name). It must be unique within the organization, regardless of which Snowflake region the account is located in. Must start with an alphabetic character and cannot contain spaces or special characters except for underscores.",
```
sfc-gh-asawicki, line 23: Should we validate  the alphabetic character at leat?

pkg/resources/listing.go
```diff
@@ -0,0 +1,181 @@
...
+		Type:        schema.TypeList,
+		Computed:    true,
+		Description: "Outputs the result of `SHOW LISTINGS` for the given listing.",
+		Elem: &schema.Resource{
+			Schema: schemas.ShowListingSchema,
+		},
+	},
+}
+
+// TODO: Should we call it free_listing or public_listing?
```
sfc-gh-asawicki, line 97: I think it's something that other teams should provide an answer for. You can also engage Swathi

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/resource_schema_def.go
```diff
@@ -105,6 +105,12 @@ var allResourceSchemaDefs = []ResourceSchemaDef{
 		name:   "LegacyServiceUser",
 		schema: resources.LegacyServiceUser().Schema,
 	},
+
+	// TODO(SNOW-1501905): Support the rest of attribute types (Commented because it's blocking model generator)
```
sfc-gh-asawicki: What attributes? LIst is supported.

sfc-gh-asawicki (changes requested):

pkg/provider/previewfeatures/preview_features.go
```diff
@@ -48,12 +49,12 @@ const (
 	ImageRepositoryResource                       feature = "snowflake_image_repository_resource"
 	ImageRepositoriesDatasource                   feature = "snowflake_image_repositories_datasource"
 	JobServiceResource                            feature = "snowflake_job_service_resource"
+	ListingResource                               feature = "snowflake_listing_resource"
```
sfc-gh-jmichalak, line 52: Please also add it to the test file. A kind reminder about updating the repo labels as well (might be done later).

pkg/resources/listing.go
```diff
@@ -0,0 +1,181 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	},
+	"share": {
+		Type:          schema.TypeString,
+		Optional:      true,
```
sfc-gh-jmichalak, line 64: Nit: I'm sure you remember, but let's have descriptions for these fields.

pkg/resources/listing.go
sfc-gh-jmichalak: +1. When we make this decision, let's also reflect it in other places (preview feature name, resource name, generators, etc.).

sfc-gh-jcieslak: Yes, I added it to the list of questions for the listing team

pkg/resources/listing.go
sfc-gh-jcieslak: I wanted to add bare minimum schema, only for `name` I did something extra by adding diff suppress and description. I fill up the descriptions / diff suppresses / validations in the next pr (updated pr desc).

pkg/provider/previewfeatures/preview_features.go
sfc-gh-jcieslak: Done

pkg/resources/listing.go
sfc-gh-jcieslak: Forgot to specify in the pr desc that I'm going to add them later

pkg/acceptance/bettertestspoc/config/model/listing_model.go
sfc-gh-jcieslak: Tried, but wasn't successful, as in this case, I would like to use a generic tfconfig.Variable instead of concrete type (also don't want it to be an array/slice). Added further explanation in the comment.

pkg/acceptance/bettertestspoc/assert/resourceassert/gen/resource_schema_def.go
sfc-gh-jcieslak: Adjusted the comment to be clearer

sfc-gh-jmichalak (dismissed):

sfc-gh-jmichalak (approved):

PR #3886, opened 2025-07-25
chore: Add a functional test for nested sensitives in the framework

sfc-gh-jmichalak:
Add a functional test proving that nested sensitive are properly protected by the framework.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -385,7 +385,7 @@ Planning failed. Terraform encountered an error while generating this plan.
 
 Some fields, like secure function definitions, can also contain sensitive values. However, because of [SDK v2](https://developer.hashicorp.com/terraform/plugin/sdkv2) limitations:
 - There is no possibility to mark sensitive values conditionally ([reference](https://github.com/hashicorp/terraform-plugin-sdk/issues/736)). This means it is not possible to mark sensitive values based on other fields, like marking `body` based on the value of `secure` field in views, functions, and procedures. As a result, this field is not marked as sensitive. For such cases, we add disclaimers in the resource documentation.
-- There is no possibility to mark sensitive values in nested fields ([reference](https://github.com/hashicorp/terraform-plugin-sdk/issues/201)). This means the nested fields, like these in `show_output` and `describe_output` cannot be sensitive. fields, like in `show_output` and `describe_output`, cannot be marked as sensitive.
+- There is no possibility to mark sensitive values in nested fields ([reference](https://github.com/hashicorp/terraform-plugin-sdk/issues/201)). This means the nested fields, like these in `show_output` and `describe_output` cannot be marked as sensitive.
 
 Instead, we added notes in the documentation of the related resources. The full list includes:
 - `snowflake_execute` resource: `execute`, `revert`, `query` and `query_results` fields,


pkg/testfunctional/test_resource_plugin_framework_nested_sensitive_acceptance_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+// This test proves that in the framework, sensitive attributes are not exposed by default in the output.
+// They can be accessed in the output block, but it must be properly marked.
+func TestAcc_TerraformPluginFrameworkFunctional_NestedSensitive(t *testing.T) {
```
sfc-gh-asawicki, line 18: question: do we have any automated test for the SDKv2 behavior or were these only manual tests some time ago? Maybe it would be nice to document it as a SDKv2 automated test (but in the future).

sfc-gh-asawicki (dismissed):

pkg/testfunctional/test_resource_plugin_framework_nested_sensitive_acceptance_test.go
```diff
@@ -0,0 +1,70 @@
...
+`, id.Name(), resourceType, PluginFrameworkFunctionalTestsProviderName)
+}
+
+func nestedSensitiveConfigMarked(id sdk.AccountObjectIdentifier, resourceType string) string {
+	return nestedSensitiveResourceConfig(id, resourceType) + fmt.Sprintf(`
+output "nested_sensitive" {
+  value = resource.%[2]s.test.output[0].string_sensitive
+  sensitive = true
+}
+`, id.Name(), resourceType, PluginFrameworkFunctionalTestsProviderName)
```
sfc-gh-jcieslak: It seems there are too many fmt.Sprintf arguments (not all are in the format?). Check three config functions. I think the one above may have similar issue.

pkg/testfunctional/test_resource_plugin_framework_nested_sensitive_acceptance_test.go
```diff
@@ -0,0 +1,70 @@
...
+output "nested_sensitive" {
+  value = resource.%[2]s.test.output[0].string_sensitive
+}
+`, id.Name(), resourceType, PluginFrameworkFunctionalTestsProviderName)
+}
+
+func nestedSensitiveConfigMarked(id sdk.AccountObjectIdentifier, resourceType string) string {
+	return nestedSensitiveResourceConfig(id, resourceType) + fmt.Sprintf(`
+output "nested_sensitive" {
+  value = resource.%[2]s.test.output[0].string_sensitive
```
sfc-gh-jcieslak: Question: Should we also test the output value of the `output` field? In the SDKv2, there was an issue that the nested sensitive values were marking implicitly the outside object (as far as I remember). By checking if we can print the `output` field without `sensitive marking`, we'll check if the plugin framework has the same limitation in that regard. Maybe this should be another test function to make clear they have different purposes, wdyt? The question I have is whether the sensitive nested field has any impact on the top-level field (afaik, there was some kind of correlation, but I'm not sure what exactly). This can be proved later (next pr?).

pkg/testfunctional/test_resource_plugin_framework_nested_sensitive_acceptance_test.go
sfc-gh-jmichalak: Fixed, it's weird that the linter didn't catch it.

pkg/testfunctional/test_resource_plugin_framework_nested_sensitive_acceptance_test.go
sfc-gh-jmichalak: I don't think we have (we tested manually). I just created SNOW-2229605 for that.

pkg/testfunctional/test_resource_plugin_framework_nested_sensitive_acceptance_test.go
sfc-gh-jmichalak: After we discussed this offline, I added a new step for referencing the whole `output` block and created SNOW-2229605 for testing in SDKv2.

sfc-gh-asawicki (approved):

PR #3888, opened 2025-07-28
chore: Implement PoC for managing warehouse using terraform plugin framework

sfc-gh-asawicki:
Implement PoC for handling warehouse in Terraform Plugin Framework based on the existing SDKv2 implementation

#### Main differences to SDKv2 implementation
- no support for `show_output` and `parameters`; private state marshalling used instead
- parameters are no longer computed (due to earlier tests for olg logic)
- no support for identifier suppression
- `IgnoreChangeToCurrentSnowflakeValueInShow` couldn't be supported (method left for reference with annotation to create a functional test for this)
- Import state has the full reading logic
- Read after create/update is separated from the normal read (due to different types and private state being in internal package)
- no current implementation for `handleExternalChangesToObjectInShow` - to be discussed
- `setStateToValuesFromConfig` was completely omitted and it seems it's not needed

#### Changes
- Configure the framework PoC provider (`pkg/testacc/10_plugin_framework_poc_provider_configure_test.go`)
  - duplicate caching logic from the existing SDKv2 tests
  - handle only a subset of attributes (leave list of unsupported ones)
  - skip handling `skip_toml_file_permission_verification` and `use_legacy_toml_file`
  - skip handling `preview_features_enabled`
  - export `getDriverConfigFromTOML` to reuse its logic
  - export `getPrivateKey` to reuse its logic
- Add `FromString` method to `WarehouseType`, `WarehouseSize`, and `ScalingPolicy` enums\
- Export method from testfunctional (`StringEnumAttributeCreate`, `BooleanAttributeCreate`, `Int64AttributeCreate`, `StringAttributeCreate`)
- Add missing methods to testfunctional, without tests (`StringEnumAttributeUpdate`, `StringEnumAttributeUpdateSetDefaultInsteadOfUnset`, `StringEnumAttributeUpdateSetOnly`, `BooleanAttributeUpdate`, `BooleanAttributeUpdateSetDefaultInsteadOfUnset`, `Int64AttributeUpdate`, `Int64AttributeUpdateSetDefaultInsteadOfUnset`, `IdAttributeCreate`, `StringAttributeUpdate`, `IdAttributeUpdate`)
- Change signature of some helpers to return error
- Add custom plan modifer for replace when removed from config (`RequiresReplaceIfRemovedFromConfig`); without dedicated functional tests
- Extract `fully_qualified_name` as embeddable struct and provide a common schema (to be decided later what logic is needed)
- Extract `SnowflakeClientEmbeddable` to easily gain access to Snowflake Client in resources

#### Testing
There are two acceptance test files:
1. `pkg/testacc/plugin_framework_poc_warehouse_acceptance_test.go` just checking the basic setup
2. `pkg/testacc/plugin_framework_poc_warehouse_sdkv2_copy_acceptance_test.go` which is a copy of the existing SDKv2 implementation acceptance tests with the following modifications:
   - Migration tests were not copied.
   - Experimental tests were not copied.
   - Models used are the same, but with the resource type replaced.
   - Assertions used are the same, but with the resource type replaced.
   - Assertions using `r.IntDefaultString` or `r.BooleanDefault` were replaced (as such defaults are not needed in the Terraform Plugin Framework).
   - Parameter assertions were adjusted due to changed behavior (parameter values are not in state when the config does not contain them). Default parameter assertions can't be used because of this.
   - `WarehouseShowOutput` assertions were removed or replaced with Snowflake object assertions (as PoC was done using private state and not `show_output`).
   - `WarehouseResourceParameters` assertions were removed or replaced with Snowflake parameters assertions (as PoC was done using private state and not `parameters`).
   - Default extensions were not used as they don't match.
   - Expectations for tests utilizing IgnoreChangeToCurrentSnowflakeValueInShow were adjusted.
   - Computed expectations for parameters were adjusted (`ExpectComputed` -> `ExpectChange`).
   - Exchanged some old assertions with new ones for easier Snowflake checks.
   - `IgnoreAfterCreation` is not implemented, so assertions for `initially_suspended` were adjusted.
   - Identifier suppression is not implemented, so adjusted steps with resource monitor.
   - `TestCheckResourceAttr` checks were replaced with `TestCheckNoResourceAttr` for values absent in config (as plugin framework handles null differently).
   - Validation tests regex assertions were adjusted.
   - Test verifying `initially_suspended` is skipped till the `IgnoreAfterCreate` is implemented.
   - Additional step to tackle (similar test will be added to the existing resource):
```
        001425 (22023): SQL compilation error:
        invalid property combination 'RESOURCE_CONSTRAINT'='MEMORY_16X' and
        'WAREHOUSE_SIZE'='X-Small'
```

#### Other changes
- in the original warehouse:
  - change `sdk.ObjectParameterMaxConcurrencyLevel` -> `sdk.WarehouseParameterMaxConcurrencyLevel`
  - remove duplication for handling parameters in update

#### For later research/changes:
- Create follow up tasks for the `mux-PR` TODOs
- Handle `CheckDestroy` in PoC plugin framework resources (the protected enum is used now)
- Decide how we can handle `IgnoreAfterCreation`
- Handle suppression for identifiers (custom types?)
- Add functional tests for added helper methods (create and update)
- Add functional tests for `RequiresReplaceIfRemovedFromConfig`
- Follow up on fully qualified name logic
- Add `show_output` and `parameters` (preferably conditionally)
- Extract modifier for id and fully_qualified_name (like we had `ComputedIfAnyAttributeChanged`)
- Add functional test documenting that `IgnoreChangeToCurrentSnowflakeValueInShow` cannot be achieved that way.
- Discuss if we want to handle imports differently
- Discuss if we want to merge read logic
- Discuss how we want to implement `handleExternalChangesToObjectInShow` equivalent


pkg/testfunctional/10_plugin_framework_poc_helpers_enums.go
```diff
@@ -33,3 +33,54 @@ func stringEnumAttributeUpdate[T customtypes.EnumCreator[T]](planValue customtyp
 	}
 	return nil
 }
+
+// TODO [mux-PR]: add functional test for this variant (with unset) to be closer to our implementation
+func StringEnumAttributeUpdate[T customtypes.EnumCreator[T]](planValue customtypes.EnumValue[T], stateValue customtypes.EnumValue[T], setField **T, unsetField **bool) error {
```
sfc-gh-jmichalak, line 38: Is it any different from the above?

pkg/testfunctional/10_plugin_framework_poc_helpers_enums.go
```diff
@@ -33,3 +33,54 @@ func stringEnumAttributeUpdate[T customtypes.EnumCreator[T]](planValue customtyp
...
+				return err
+			}
+			*setField = sdk.Pointer(v)
+		}
+	}
+	return nil
+}
+
+// TODO [mux-PR]: add functional test for this variant
+func StringEnumAttributeUpdateSetOnly[T customtypes.EnumCreator[T]](planValue customtypes.EnumValue[T], stateValue customtypes.EnumValue[T], setField **T) error {
```
sfc-gh-jmichalak, line 72: Q: this results in a similar behavior to `ForceNewIfChangeToEmptyString` we have for SDKv2, right? If so, let's make it clear in the function name, because we also have cases where unset does not work , so we call set with an empty value, which may be confusing.

pkg/testfunctional/10_plugin_framework_poc_helpers_zero_values.go
```diff
@@ -42,10 +68,43 @@ func int64AttributeUpdate(planned types.Int64, inState types.Int64, setField **i
 	}
 }
 
-func stringAttributeCreate(stringAttribute types.String, createField **string) {
+// TODO [mux-PR]: add functional test for this variant (with unset) to be closer to our implementation
+func Int64AttributeUpdate(planned types.Int64, inState types.Int64, setField **int, unsetField **bool) error {
```
sfc-gh-jmichalak, line 72: Nit: let's improve the naming here and other places as well, bc here we set UNSET, but above we set a null to UNSET.

pkg/testacc/20_plugin_framework_poc_warehouse.go
```diff
@@ -0,0 +1,791 @@
...
+		warehouseJson.WarehousePocParametersPrivateJson = *warehouseParametersJson
+	}
+	bytes, err := json.Marshal(warehouseJson)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal json: %w", err)
+	}
+	return bytes, nil
+}
+
+func marshallWarehousePocParameters(warehouseParameters []*sdk.Parameter, privateJson *WarehousePocParametersPrivateJson) error {
```
sfc-gh-jmichalak: Nit, other places as well.
```suggestion
func marshalWarehousePocParameters(warehouseParameters []*sdk.Parameter, privateJson *WarehousePocParametersPrivateJson) error {
```

pkg/resources/warehouse.go
```diff
@@ -585,20 +585,6 @@ func UpdateWarehouse(ctx context.Context, d *schema.ResourceData, meta any) diag
 			unset.QueryAccelerationMaxScaleFactor = sdk.Bool(true)
 		}
 	}
-	if d.HasChange("max_concurrency_level") {
-		if v := d.Get("max_concurrency_level").(int); v != IntDefault {
```
sfc-gh-jcieslak, line 589: I'm guessing you removed those, because they're handled in the handleWarehouseParametersChanges, but is it really the same handling? I don't think the IntDefault is taken into account there. The raw config is used instead, but I guess this could result in the same correctly detected changes, is it right?

pkg/testacc/20_plugin_framework_poc_warehouse.go
```diff
@@ -0,0 +1,791 @@
...
+// we can't use here the WarehouseParameter type values as struct tags are pure literals
+// this is really easy to generate though
+type warehouseParametersModelV0 struct {
+	MaxConcurrencyLevel             types.Int64 `tfsdk:"max_concurrency_level"`
+	StatementQueuedTimeoutInSeconds types.Int64 `tfsdk:"statement_queued_timeout_in_seconds"`
+	StatementTimeoutInSeconds       types.Int64 `tfsdk:"statement_timeout_in_seconds"`
+}
+
+type WarehousePocPrivateJson struct {
+	WarehouseType                   sdk.WarehouseType `json:"warehouse_type,omitempty"`
```
sfc-gh-jmichalak, line 71: Q: no name?

pkg/testacc/10_plugin_framework_poc_provider_configure_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	enumValue, err := toEnumFunc(value)
+	if err != nil {
+		return err
+	}
+	*setInConfig = enumValue
+	return nil
+}
+
+func setEnumAttributeString[T ~string](stringField types.String, envName string, toEnumFunc func(string) (T, error), setInConfig *string) error {
```
sfc-gh-jcieslak, line 132: Couldn't this be merged with the function above? All of our enums are rather string-based, but as far as I remember there are some enums in the driver that are int based 🤔

pkg/testacc/10_plugin_framework_poc_provider_test.go
```diff
@@ -77,41 +88,94 @@ func (p *pluginFrameworkPocProvider) Schema(_ context.Context, _ provider.Schema
...
+			accTestLog.Printf("[DEBUG] Returning cached terraform plugin framework PoC provider configuration error")
+			response.Diagnostics.Append(configureClientErrorPluginFrameworkDiag...)
+			return
+		}
+	}
+	accTestLog.Printf("[DEBUG] No cached terraform plugin framework PoC provider configuration found or caching is not enabled; configuring a new provider")
+
+	providerCtx, clientErrorDiag := p.configureWithoutCache(ctx, request, response)
+
+	if providerCtx != nil && accTestEnabled && oswrapper.Getenv("SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES") == "true" {
```
sfc-gh-jcieslak: nit: we could check clientErrorDiag and quit early to avoid providerCtx checks, SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES could be taken from testenvs

pkg/testacc/10_plugin_framework_poc_provider_test.go
```diff
@@ -77,41 +88,94 @@ func (p *pluginFrameworkPocProvider) Schema(_ context.Context, _ provider.Schema
...
+		configurePluginFrameworkProviderCtx = providerCtx
+		configureClientErrorPluginFrameworkDiag = clientErrorDiag
+	} else {
+		configurePluginFrameworkProviderCtx = nil
+		configureClientErrorPluginFrameworkDiag = make(diag.Diagnostics, 0)
+	}
+	// no last configured provider
 
+	if clientErrorDiag.HasError() {
+		response.Diagnostics.Append(clientErrorDiag...)
```
sfc-gh-jcieslak: isn't it repetitive? You already added error diags in configureWithoutCache function

pkg/testacc/plugin_framework_poc_warehouse_sdkv2_copy_acceptance_test.go
```diff
@@ -0,0 +1,1371 @@
...
+						HasScalingPolicy(sdk.ScalingPolicyEconomy).
+						HasAutoSuspend(1200).
+						HasAutoResume(false).
+						HasResourceMonitor(resourceMonitor.ID()).
+						HasComment(newComment).
+						HasEnableQueryAcceleration(true).
+						HasQueryAccelerationMaxScaleFactor(4),
+				),
+			},
+			// change resource monitor - wrap in quotes (no change expected)
```
sfc-gh-jmichalak, line 351: Nit: this coment is incorrect now, as we assert a non-empty plan

pkg/testacc/plugin_framework_poc_warehouse_sdkv2_copy_acceptance_test.go
```diff
@@ -0,0 +1,1371 @@
...
+				Check: assertThat(t,
+					assert.Check(resource.TestCheckResourceAttr(replaceResourceReference(warehouseModelWithStatementTimeoutInSeconds86400.ResourceReference()), "statement_timeout_in_seconds", "86400")),
+					objectparametersassert.WarehouseParameters(t, id).
+						HasStatementTimeoutInSeconds(86400).
+						HasStatementTimeoutInSecondsLevel(sdk.ParameterTypeWarehouse),
+				),
+			},
+			// do not make any change (to check if there is no drift)
+			{
+				Config: replaceWithWarehousePoCResourceType(t, config.FromModels(t, warehouseModelWithStatementTimeoutInSeconds86400)),
```
sfc-gh-jmichalak: I know we also have this in sdkv2, but I think we don't need it, as it's checked automatically in the previous step, unless `ExpectNonEmptyPlan` is set.

pkg/testacc/20_plugin_framework_poc_warehouse.go
```diff
@@ -0,0 +1,791 @@
...
+func handleWarehousePocParameterImportInt(parameter *sdk.Parameter, field *types.Int64, objectLevel sdk.ParameterType, diags *diag.Diagnostics) {
+	value, err := strconv.Atoi(parameter.Value)
+	if err != nil {
+		diags.AddError(fmt.Sprintf("Handling parameter %s failed", parameter.Key), err.Error())
+		return
+	}
+	if parameter.Level == objectLevel {
+		*field = types.Int64Value(int64(value))
+	}
+}
```
sfc-gh-jcieslak, line 448: For later, we have to think of a way to structure those resources into a few files, as they may become a bit big. The problem here, in contrast to SDKv2, is that there are a lot of small functions and variables instead of bigger ones. In SDKv2 despite files being bigger or similar size, feel easier to grasp as a whole, but that may come from the fact we are more used to / familar with it.

pkg/testacc/20_plugin_framework_poc_warehouse.go
```diff
@@ -0,0 +1,791 @@
...
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	name := data.Name.ValueString()
+	id := sdk.NewAccountObjectIdentifier(name)
+
+	opts := &sdk.CreateWarehouseOptions{}
+	errs := errors.Join(
+		testfunctional.StringEnumAttributeCreate(data.WarehouseType, &opts.WarehouseType),
```
sfc-gh-jcieslak, line 462: For the future, as the models may change (e.g. we would like to introduce new model version), maybe it would be better if this kind of mapping were the model's responsibility?
or maybe this could be an overkill, as removed options wouldn't be that common, and any SDK modifications could break the old mapping anyway, but we should discuss how do we want to go about schema/model updates.

pkg/testfunctional/10_plugin_framework_poc_helpers_zero_values.go
```diff
@@ -42,10 +68,43 @@ func int64AttributeUpdate(planned types.Int64, inState types.Int64, setField **i
...
+}
+
+func StringAttributeCreate(stringAttribute types.String, createField **string) error {
 	if !stringAttribute.IsNull() {
 		*createField = stringAttribute.ValueStringPointer()
 	}
+	return nil
+}
+
+// TODO [mux-PR]: test and adjust when adding identifier suppression
```
sfc-gh-jcieslak, line 102: todo: add support for other identifier types

pkg/testfunctional/10_plugin_framework_poc_helpers_enums.go
sfc-gh-asawicki: Yes (`unsetField **T` used for simplicity in existing functional tests versus `unsetField **bool` which is more closer to our SDK impl)

pkg/testfunctional/10_plugin_framework_poc_helpers_enums.go
sfc-gh-asawicki: It is not. There is an added plan modifier doing what you ask for, `RequiresReplaceIfRemovedFromConfig` (remember, there is nothing like forceNew in the plugin framework).

And I believe that these two methods (`StringEnumAttributeUpdateSetOnly` vs `StringEnumAttributeUpdateSetDefaultInsteadOfUnset`) already  offer enough context in their name.

pkg/testfunctional/10_plugin_framework_poc_helpers_zero_values.go
sfc-gh-asawicki: Yes, because the above method is a simplified one used for functional tests for simplicity. This one follows our usual SDK implementation. For that reason, the above is still private with existing TODO.

pkg/testacc/20_plugin_framework_poc_warehouse.go
sfc-gh-asawicki: I always do that... Done.

pkg/testacc/20_plugin_framework_poc_warehouse.go
sfc-gh-asawicki: No. We don't need it as we fetch by it.

pkg/testacc/plugin_framework_poc_warehouse_sdkv2_copy_acceptance_test.go
sfc-gh-asawicki: But the intent should stay the same. There is a TODO below. I moved it higher so it stays together with this comment.

pkg/testacc/plugin_framework_poc_warehouse_sdkv2_copy_acceptance_test.go
sfc-gh-asawicki: Removing from both.

pkg/testacc/10_plugin_framework_poc_provider_configure_test.go
sfc-gh-asawicki: It's exactly for the in enum (authenticator type) that is int in the driver.

pkg/resources/warehouse.go
sfc-gh-asawicki: All the tests passed but... the detailed tests for the warehouse parameters (TestAcc_Warehouse_Parameter) are done only for the third one...But, look, that there is no Default added in schema for sdk.WarehouseParameterMaxConcurrencyLevel, so I guess it's safe to be removed and it's handled by `!d.GetRawConfig().AsValueMap()[key].IsNull`, right?

pkg/testacc/10_plugin_framework_poc_provider_test.go
sfc-gh-asawicki: Used the testenvs.

We can leave early, as we are saving the previous results (including errors). This logic is copied from the existing one in the acceptance tests.

pkg/testacc/10_plugin_framework_poc_provider_test.go
sfc-gh-asawicki: Hmmm, you are right. I need to return the diagnostics, though. I changed the `configureWithoutCache` implementation to not add any diags to the response there.

pkg/testacc/10_plugin_framework_poc_provider_test.go
sfc-gh-asawicki: Moved higher but without return.

pkg/testacc/20_plugin_framework_poc_warehouse.go
sfc-gh-asawicki: Yes.

pkg/testacc/20_plugin_framework_poc_warehouse.go
sfc-gh-asawicki: Yes, with the migration research.

pkg/testfunctional/10_plugin_framework_poc_helpers_zero_values.go
sfc-gh-asawicki: Yes.

pkg/testacc/20_plugin_framework_poc_warehouse.go
sfc-gh-jmichalak: Ok. Previously, in the output fields, we had all values returned from SF, so I guess that in the private state, we won't store db, schema, and argument types as well?

pkg/testfunctional/10_plugin_framework_poc_helpers_enums.go
sfc-gh-jmichalak: `should be handled by force new` got me confused.

pkg/resources/warehouse.go
sfc-gh-jcieslak: I think so, just double-checking as the logic is different, but it seems to be doing the same thing.

PR #3891, opened 2025-07-29
chore: Add REST API PoC for warehouse resource

sfc-gh-asawicki:
Implement PoC for handling warehouse with REST API based on the existing SDKv2 implementation and Terraform Plugin Framework PoC (#3888)
- create a simple rest API client
  - it only allows pat authorization currently
  - it offers simplified error handling
- Add rest api client to provider context (only the mixed test one)
- Provide a simplified API for warehouse management (one struct reused for each method)
- Create a new resource for warehouse; adjust the logic to use rest API client
  - Reuse parts of the Terraform Plugin Framework PoC (schema, data model)
- Perform the sanity test for creation (other tests will be adjusted next PR)
  - test is set up on the tmp user on purpose

Next PRs:
- Copy and adjust basic acceptance tests to verify full behavior

Future:
- We should have tests similar to the SDK integration tests we have today
- We need to support other auth methods
- Error handling


pkg/testacc/30_rest_api_poc_client_test.go
```diff
@@ -0,0 +1,176 @@
...
+	Account string
+	Token   string
+}
+
+func RestApiPocConfigFromDriverConfig(driverConfig *gosnowflake.Config) (*RestApiPocConfig, error) {
+	res := &RestApiPocConfig{
+		Account: strings.ToLower(driverConfig.Account),
+	}
+	if driverConfig.Token == "" {
+		return nil, fmt.Errorf("token is currently required for REST API PoC client initialization")
```
sfc-gh-jmichalak, line 32: Q: What about the account, is it really optional?

pkg/testacc/30_rest_api_poc_client_test.go
```diff
@@ -0,0 +1,176 @@
+package testacc
```
sfc-gh-jmichalak, line 1: The REST API code can be extracted to a separate package.

pkg/testacc/30_rest_api_poc_client_test.go
```diff
@@ -0,0 +1,176 @@
...
+type RestApiPocClient struct {
+	httpClient *http.Client
+	url        string
+	token      string
+
+	Warehouses WarehousesPoc
+}
+
+func (c *RestApiPocClient) doRequest(ctx context.Context, method string, path string, body io.Reader, queryParams map[string]string) (*http.Response, error) {
+	req, err := http.NewRequestWithContext(ctx, method, c.url+path, body)
```
sfc-gh-jmichalak: Nit: url.JoinPath

pkg/testacc/31_rest_api_poc_warehouses_test.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("warehousesPoc.GetByID(%s): %w", id.Name(), err)
+	}
+	return warehouse, nil
+}
+
+// Based on https://docs.snowflake.com/developer-guide/snowflake-rest-api/reference/warehouse#delete--api-v2-warehouses-name
+func (w warehousesPoc) Drop(ctx context.Context, id sdk.AccountObjectIdentifier, ifExists *bool) error {
+	queryParams := map[string]string{}
+	if ifExists != nil {
```
sfc-gh-jmichalak, line 97: What about ifExists in other operations? I can see it's supported.

pkg/testacc/33_rest_api_poc_warehouse_resource_test.go
```diff
@@ -0,0 +1,527 @@
...
+	if warehouse.StatementQueuedTimeoutInSeconds != nil {
+		privateJson.StatementQueuedTimeoutInSeconds = *warehouse.StatementQueuedTimeoutInSeconds
+	}
+	if warehouse.StatementTimeoutInSeconds != nil {
+		privateJson.StatementTimeoutInSeconds = *warehouse.StatementTimeoutInSeconds
+	}
+	return privateJson, nil
+}
+
+func marshallWarehouseRestApiPocPrivateJson(warehouseApiModel *WarehouseApiModel) ([]byte, error) {
```
sfc-gh-jmichalak: Nit: again :)

pkg/testacc/33_rest_api_poc_warehouse_resource_test.go
```diff
@@ -0,0 +1,527 @@
...
+	if prevValueBytes != nil {
+		var prevValue WarehousePocPrivateJson
+		err := json.Unmarshal(prevValueBytes, &prevValue)
+		if err != nil {
+			diags.AddError("Could not unmarshal json", err.Error())
+			return diags
+		}
+
+		if warehouse.WarehouseType != nil && *warehouse.WarehouseType != prevValue.WarehouseType {
+			data.WarehouseType = customtypes.NewEnumValue(*warehouse.WarehouseType)
```
sfc-gh-jmichalak, line 340: Hm, but what if warehouse.WarehouseType is nil, shouldn't we override the data field?

pkg/testacc/33_rest_api_poc_warehouse_resource_test.go
```diff
@@ -0,0 +1,527 @@
...
+		warehouseModel.WaitForCompletion = sdk.String("true")
+	}
+
+	// Put (CoA)
+	if err := r.client.Warehouses.CreateOrAlter(ctx, warehouseModel); err != nil {
+		response.Diagnostics.AddError("Could not run create or alter in REST API PoC warehouse", err.Error())
+		return
+	}
+
+	plan.FullyQualifiedName = types.StringValue(id.FullyQualifiedName())
```
sfc-gh-jmichalak: I think it should be before CoA because if CoA fails, we're left with the old FQN, despite setting the id after the rename.

pkg/testacc/30_rest_api_poc_client_test.go
sfc-gh-asawicki: It's not, however, it will fail the SDK client setup first, so we will never get here. I will add it for completeness.

pkg/testacc/30_rest_api_poc_client_test.go
sfc-gh-asawicki: Yes, later it can. For the PoC I think it can stay here with other PoCs

pkg/testacc/31_rest_api_poc_warehouses_test.go
sfc-gh-asawicki: It's the only one we care for it for now (it's a PoC, not a full implementation)

pkg/testacc/33_rest_api_poc_warehouse_resource_test.go
sfc-gh-asawicki: yeah.. xD

pkg/testacc/30_rest_api_poc_client_test.go
sfc-gh-asawicki: Replaced string with URL.

pkg/testacc/33_rest_api_poc_warehouse_resource_test.go
sfc-gh-asawicki: We can verify and adjust if needed when adding the tests

pkg/testacc/33_rest_api_poc_warehouse_resource_test.go
sfc-gh-asawicki: True. Changed.

pkg/testacc/10_plugin_framework_poc_provider_test.go
```diff
@@ -164,14 +163,23 @@ func (p *pluginFrameworkPocProvider) configureWithoutCache(ctx context.Context,
...
 	if client, err := sdk.NewClient(config); err != nil {
 		diags.AddError("Could not initialize client", err.Error())
 		return nil, diags
 	} else {
 		providerCtx.Client = client
 	}
 
+	// using warnings on purpose here
+	if restApiPocConfig, err := RestApiPocConfigFromDriverConfig(config); err != nil {
+		response.Diagnostics.AddWarning("Could not initialize REST API PoC client", err.Error())
```
sfc-gh-jcieslak: Let's differentiate those error messages a bit, so we'll know which branch failed

pkg/testacc/31_rest_api_poc_warehouses_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}{Name: newId.Name()})
+	if err != nil {
+		return fmt.Errorf("warehousesPoc.Rename(%s -> %s): %w", id.Name(), newId.Name(), err)
+	}
+	return nil
+}
+
+// Based on https://docs.snowflake.com/developer-guide/snowflake-rest-api/reference/warehouse#get--api-v2-warehouses-name
+func (w warehousesPoc) GetByID(ctx context.Context, id sdk.AccountObjectIdentifier) (*WarehouseApiModel, error) {
+	warehouse, err := get[WarehouseApiModel](ctx, w.client, fmt.Sprintf("warehouses/%s", id.Name()))
```
sfc-gh-jcieslak, line 87: Should those URLs be sanitized in any way? e.g., is there a Go package (from the standard library) that allows for constructing URLs (or parts of it) so that, e.g., malicious resource name won't cause any trouble. I'm guessing the risk is rather low, but just wondering.

pkg/testacc/rest_api_poc_warehouse_acceptance_test.go
```diff
@@ -0,0 +1,73 @@
...
+						HasAutoSuspend(600).
+						HasAutoResume(true).
+						HasResourceMonitor(sdk.AccountObjectIdentifier{}).
+						HasComment("").
+						HasEnableQueryAcceleration(false).
+						HasQueryAccelerationMaxScaleFactor(8),
+					objectparametersassert.WarehouseParameters(t, id).
+						HasAllDefaults(),
+				),
+			},
```
sfc-gh-jcieslak, line 62: Can we just test one more step to confirm update works?

sfc-gh-asawicki: As written, we will add more tests in the following PRs (we will copy the same ones as for plugin frmkw).

pkg/testacc/31_rest_api_poc_warehouses_test.go
sfc-gh-asawicki: It can be a question for the future, currently this is not the production code, and we can't even use it on the production yet.

pkg/testacc/10_plugin_framework_poc_provider_test.go
sfc-gh-asawicki: Done

sfc-gh-jcieslak (dismissed):

pkg/testacc/30_rest_api_poc_client_test.go
sfc-gh-jmichalak: Okay, but I see `if driverConfig.Token` twice.

sfc-gh-jmichalak (approved):

pkg/testacc/30_rest_api_poc_client_test.go
sfc-gh-asawicki: Fixed

sfc-gh-jmichalak (approved):

PR #3893, opened 2025-07-30
feat: Add listing resource implementation

sfc-gh-jcieslak:
## Changes
- Add resource implementation
- Add acceptance tests
- Fix listing definition (and regenerate)
- Add migration guide and resource documentation with examples and current limitations

### DIFF
--- a/pkg/acceptance/helpers/stage_client.go
+++ b/pkg/acceptance/helpers/stage_client.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"testing"
@@ -123,7 +124,10 @@ func (c *StageClient) PutInLocationWithContent(t *testing.T, stageLocation strin
 	require.NoError(t, err)
 	t.Cleanup(func() {
 		_, err = c.context.client.ExecForTests(ctx, fmt.Sprintf(`REMOVE %s/%s`, stageLocation, filename))
-		require.NoError(t, err)
+		// Only check the error if it's not related to the stage / file existence or access
+		if !errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) {
+			require.NoError(t, err)
+		}
 	})
 
 	return filePath
--- a/pkg/testacc/resource_cortex_search_service_acceptance_test.go
+++ b/pkg/testacc/resource_cortex_search_service_acceptance_test.go
@@ -141,7 +141,7 @@ func TestAcc_CortexSearchService_basic(t *testing.T) {
 				ImportState:       true,
 				ImportStateVerify: true,
 				// currently not set in read because the early implementation on Snowflake side did not return these values on SHOW/DESCRIBE
-				ImportStateVerifyIgnore: []string{"attributes", "on", "query", "target_lag", "warehouse"},
+				ImportStateVerifyIgnore: []string{"attributes", "on", "query", "target_lag", "warehouse", "describe_output.0.data_timestamp"},
 			},
 		},
 	})


MIGRATION_GUIDE.md
```diff
@@ -23,6 +23,11 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ## v2.4.x ➞ v2.5.0
 
+### *(new feature)* snowflake_listing resource
+Added a new preview resource for managing listings. Check the official Snowflake documentation to know more [about listings](https://other-docs.snowflake.com/en/collaboration/collaboration-listings-about). You can read about the resource limitations in the documentation in the registry.
+
+This feature will be marked as a stable feature in future releases. Breaking changes are expected, even without bumping the major version. To use this feature, add `snowflake_listing_resource` to `preview_features_enabled` field in the provider configuration.
```
sfc-gh-asawicki, line 33: nit: please add here the two limitations that you added to the resource docs (public listings and inlining manifest into tf schema), so that someone reading only this is aware.

pkg/acceptance/bettertestspoc/config/model/listing_model.go
```diff
@@ -39,7 +39,7 @@ func ListingWithInlineManifest(
 	l.WithName(name)
 	l.WithManifestValue(tfconfig.ListVariable(
 		tfconfig.MapVariable(map[string]tfconfig.Variable{
-			"from_string": tfconfig.StringVariable(manifest),
+			"from_string": config.MultilineWrapperVariable(manifest),
```
sfc-gh-asawicki, line 42: nit: I know it's not generated (i mean, it is, but it was copied), but you can add this to overrides

pkg/resources/listing.go
```diff
@@ -23,35 +23,56 @@ var listingSchema = map[string]*schema.Schema{
 		Description:      "Specifies the listing identifier (name). It must be unique within the organization, regardless of which Snowflake region the account is located in. Must start with an alphabetic character and cannot contain spaces or special characters except for underscores.",
 	},
```
sfc-gh-asawicki, line 25: what about name validation (https://github.com/snowflakedb/terraform-provider-snowflake/pull/3885#discussion_r2228572615)?

examples/resources/snowflake_listing/resource.tf
```diff
@@ -0,0 +1,61 @@
...
+EOT
+  }
+}
+
+# basic resource with manifest in a stage
+resource "snowflake_listing" "basic_staged" {
+  name = "LISTING"
+  manifest {
+    from_stage = {
+      stage = snowflake_stage.test_stage.fully_qualified_name
```
sfc-gh-jmichalak, line 20: Nit: and the stage location.

pkg/sdk/listings_gen.go
```diff
@@ -299,27 +300,27 @@ type ShowVersionsListingOptions struct {
...
 	Name              string         `db:"name"`
-	Alias             string         `db:"alias"`
+	Alias             sql.NullString `db:"alias"`
 	LocationUrl       string         `db:"location_url"`
 	IsDefault         bool           `db:"is_default"`
 	IsLive            bool           `db:"is_live"`
 	IsFirst           bool           `db:"is_first"`
 	IsLast            bool           `db:"is_last"`
-	Comment           string         `db:"comment"`
+	Comment           sql.NullString `db:"comment"`
```
sfc-gh-asawicki, line 309: I havent seen any changes in integration tests connected with these nullstring changes

pkg/testacc/resource_listing_acceptance_test.go
```diff
@@ -0,0 +1,922 @@
...
+						HasCommentString(newComment),
+					resourceshowoutputassert.ListingShowOutput(t, modelCompleteWithDifferentValues.ResourceReference()).
+						HasName(id.Name()).
+						HasTitle(listingTitleWithTargetAccountsEscaped).
+						HasSubtitle(`\"subtitle\"`).
+						HasState(sdk.ListingStateUnpublished).
+						HasComment(newComment),
+				),
+			},
+			// unset
```
sfc-gh-asawicki, line 205: is it really unset or recreate? (because share is forcenew) So I guess a step with real unset (without changing the share) is needed

pkg/testacc/resource_listing_acceptance_test.go
```diff
@@ -0,0 +1,922 @@
...
+					resourceassert.ListingResource(t, modelInitialInlined.ResourceReference()).
+						HasNameString(id.Name()).
+						HasManifestFromStringNotEmpty().
+						HasShareString(share.ID().FullyQualifiedName()).
+						HasPublishString(r.BooleanTrue),
+					resourceshowoutputassert.ListingShowOutput(t, modelInitialInlined.ResourceReference()).
+						HasName(id.Name()).
+						HasTitle(title1Escaped).
+						HasState(sdk.ListingStatePublished),
+					// Cannot assert versions (because SHOW VERSIONS IN LISTING will throw an error) for inlined manifests
```
sfc-gh-asawicki: do we have integration test asserting this error?

sfc-gh-asawicki (changes requested):

pkg/resources/listing.go
```diff
@@ -23,35 +23,56 @@ var listingSchema = map[string]*schema.Schema{
...
+		Description: externalChangesNotDetectedFieldDescription("Specifies the way manifest is provided for the listing. For more information on manifest syntax, see [Listing manifest reference](https://docs.snowflake.com/en/progaccess/listing-manifest-reference)."),
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"from_string": {
-					Type:          schema.TypeString,
-					Optional:      true,
-					ConflictsWith: []string{"manifest.0.from_stage"},
+					Type:         schema.TypeString,
+					Optional:     true,
+					Description:  "Manifest provided as a string. For more information on manifest syntax, see [Listing manifest reference](https://docs.snowflake.com/en/progaccess/listing-manifest-reference). Also, the [multiline string syntax](https://developer.hashicorp.com/terraform/language/expressions/strings#heredoc-strings) is a must here. A proper YAML indentation (2 spaces) is required.",
```
sfc-gh-jmichalak: We also wrap it in `$$` automatically.

pkg/resources/listing.go
```diff
@@ -23,35 +23,56 @@ var listingSchema = map[string]*schema.Schema{
...
 				},
 				"from_stage": {
-					Type:          schema.TypeList,
-					Optional:      true,
-					MaxItems:      1,
-					ConflictsWith: []string{"manifest.0.from_string"},
+					Type:         schema.TypeList,
+					Optional:     true,
+					MaxItems:     1,
+					Description:  "Manifest provided as a string. For more information on manifest syntax, see [Listing manifest reference](https://docs.snowflake.com/en/progaccess/listing-manifest-reference). A proper YAML indentation (2 spaces) is required.",
```
sfc-gh-jmichalak: This description is incorrect.

pkg/resources/listing.go
```diff
@@ -104,15 +129,16 @@ func Listing() *schema.Resource {
 	)
 
 	return &schema.Resource{
-		Description:   "",
+		Description: "Resource used to manage listing objects. For more information, check [listing documentation](https://other-docs.snowflake.com/en/collaboration/collaboration-listings-about).",
```
sfc-gh-jmichalak, line 133: Nit: I find it odd that the docs redirect to other-docs only for this page, let's report it IMO.

pkg/resources/listing.go
```diff
@@ -129,6 +155,67 @@ func CreateListing(ctx context.Context, d *schema.ResourceData, meta any) diag.D
...
+			location = l.(string)
+		}
+
+		req.WithFrom(sdk.NewStageLocation(stage, location))
+	}
+
+	withReq := sdk.NewListingWithRequest()
+
+	if errs := errors.Join(
+		attributeMappedValueCreateBuilder(d, "share", withReq.WithShare, sdk.ParseAccountObjectIdentifier),
```
sfc-gh-jmichalak: Nit: you can also handle the comment and from_string here.

pkg/resources/listing.go
```diff
@@ -146,8 +233,96 @@ func UpdateListing(ctx context.Context, d *schema.ResourceData, meta any) diag.D
...
+		d.SetId(helpers.EncodeSnowflakeID(newId))
+		id = newId
+	}
+
+	if d.HasChange("manifest") {
+		if d.HasChange("manifest.0.from_string") {
+			if manifest := d.Get("manifest.0.from_string").(string); manifest != "" {
+				req := sdk.NewAlterListingAsRequest(manifest)
+
+				if err := booleanStringAttributeCreate(d, "publish", &req.Publish); err != nil {
```
sfc-gh-jmichalak, line 245: Pls use `d.Partial`, in other places as well.

pkg/resources/listing.go
```diff
@@ -175,7 +350,21 @@ func ReadListing(ctx context.Context, d *schema.ResourceData, meta any) diag.Dia
...
-	_ = listing
+	listingDetails, err := client.Listings.Describe(ctx, sdk.NewDescribeListingRequest(id))
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if errs := errors.Join(
+		setOptionalValueWithMapping(d, "share", listingDetails.Share, (*sdk.AccountObjectIdentifier).FullyQualifiedName),
+		setOptionalValueWithMapping(d, "application_package", listingDetails.ApplicationPackage, (*sdk.AccountObjectIdentifier).FullyQualifiedName),
+		d.Set("publish", booleanStringFromBool(listing.State == sdk.ListingStatePublished)),
```
sfc-gh-jmichalak, line 360: Nit: maybe this field should be called `published`, as it's describing the listing state.

templates/resources/listing.md.tmpl
```diff
@@ -0,0 +1,50 @@
...
+{{- else -}}
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+{{- end }}
+---
+
+!> **Caution: Preview Feature** This feature is considered a preview feature in the provider, regardless of the state of the resource in Snowflake. We do not guarantee its stability. It will be reworked and marked as a stable feature in future releases. Breaking changes are expected, even without bumping the major version. To use this feature, add the relevant feature name to `preview_features_enabled` field in the [provider configuration](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs#schema). Please always refer to the [Getting Help](https://github.com/snowflakedb/terraform-provider-snowflake?tab=readme-ov-file#getting-help) section in our Github repo to best determine how to get help for your questions.
+
+!> **Warning** Versioning only works if your listing ever sourced the manifest from stage. This is a Snowflake limitation.
+
+!> **Warning** Changes to the manifest won't be detected by the provider automatically. You need to manually trigger updates when manifest content changes.
```
sfc-gh-jmichalak: This applies only to the stage manifests, right? Or you meant the external changes?

pkg/testacc/resource_listing_acceptance_test.go
sfc-gh-jmichalak: +1, let's also assert the update operation in the alter above.

pkg/testacc/resource_listing_acceptance_test.go
```diff
@@ -0,0 +1,922 @@
...
+				),
+			},
+			// Modify the manifest and show that no version was produced
```
sfc-gh-jmichalak, line 633: How do we show it here? I don't understand.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Added

pkg/acceptance/bettertestspoc/config/model/listing_model.go
sfc-gh-jcieslak: Added, but as commented with future TODO, I'm wondering if those overrides support nested fields (all of the previous multiline values are top-level).

pkg/resources/listing.go
sfc-gh-jcieslak: Missed it (at first I thought the comment was regarding all validations in the schema), added a custom validator

pkg/sdk/listings_gen.go
sfc-gh-jcieslak: Added those (wanted to wait for initial test run on CI since it is running on different env which can make difference in this case).

pkg/testacc/resource_listing_acceptance_test.go
sfc-gh-jcieslak: You're right, adjusted

pkg/testacc/resource_listing_acceptance_test.go
sfc-gh-jcieslak: We don't, added

examples/resources/snowflake_listing/resource.tf
sfc-gh-jcieslak: If the manifest is in the root, the location is not needed

pkg/resources/listing.go
sfc-gh-jcieslak: added a note

pkg/resources/listing.go
sfc-gh-jcieslak: Right, corrected

pkg/resources/listing.go
sfc-gh-jcieslak: Added to the list of things to report. I'll start reporting at the start of next week.

pkg/resources/listing.go
sfc-gh-jcieslak: done

pkg/resources/listing.go
sfc-gh-jcieslak: afaik partial should be added whenever an error occurs during the update, right?

pkg/resources/listing.go
sfc-gh-jcieslak: Well, in this context right now it represents a state, but later on it may not be always the case. The publish operation in Snowflake is more of a flag that says "whenever the listing passes the review should it be immediately published or not?". Because only private listings are supported, the review is automatically approved by default and the same for publishing, thus currently representing the state. When we would like to support the public listings, this could act then more like a flag rather than state. Given this info, I'm not quite sure what the name should be.

templates/resources/listing.md.tmpl
sfc-gh-jcieslak: External changes, adjusted

pkg/testacc/resource_listing_acceptance_test.go
sfc-gh-jcieslak: Added specialized assert

sfc-gh-asawicki (dismissed):

examples/resources/snowflake_listing/resource.tf
sfc-gh-jmichalak: Okay, but we can still do this in the example - actually in the complete resource; next PR.

pkg/resources/listing.go
sfc-gh-jmichalak: I think so; IMO we should discuss it because we were unsure about it lately. We can leave this for now.

pkg/resources/listing.go
sfc-gh-jmichalak: Sure, we can come back to it in public listings. Let's maybe adjust the docs: so that it's clear that for now this means the state, and we don't run PUBLISH=TRUE each time.

sfc-gh-jmichalak (dismissed):

pkg/resources/listing.go
sfc-gh-jcieslak: As this is only doc change, will do in next pr.

examples/resources/snowflake_listing/resource.tf
sfc-gh-jcieslak: Addressed in https://github.com/snowflakedb/terraform-provider-snowflake/pull/3902

pkg/resources/listing.go
sfc-gh-jcieslak: This was addressed in this pr

pkg/resources/listing.go
sfc-gh-jcieslak: Didn't know how more to express this as field description already says `Determines if the listing should be published.` which is what we were discussing here. Which rather clearly states the state in Snowflake has to be Published otherwise we'll try to reach this state.

sfc-gh-asawicki (approved):

PR #3894, opened 2025-07-31
fix(file_format): handle empty strings in `NULL_IF`

denizsurmeli:
This PR fixes a bug where if a file format has `[]` as it's `NULL_IF` value, provider sees as `[null]`, causing permadiff on file formats with empty `NULL_IF`.

See the playground for producing: https://go.dev/play/p/N_zE0rTRToA

## Test Plan

* [x] Integration tests
* [x] Acceptance tests

## References
https://github.com/snowflakedb/terraform-provider-snowflake/issues/3325
https://github.com/snowflakedb/terraform-provider-snowflake/issues/1609


sfc-gh-jmichalak:
Hi @denizsurmeli 👋 

Thank you for your contribution. The file formats are still in preview, but they aren't on our roadmap for now. We'll take a closer look at this PR, though.

I changed the base branch to the `dev`. I linked two related issues - they do not address this exact issue, but a similar one, so it's worth taking a look. For now:
- The integration test is not working. Now, when the NULL_IF is not set, it is not sent as an empty list to Snowflake, as desired. This should also be fixed—we did something similar in PreAuthorizedRolesList.
- We already have a `ParseCommaSeparatedStringArray` helper.
- We should check if we should fix other file types besides the CSV.
- We should also add an acceptance test to prove this change really fixes the permadiff.
- The migration guide entry is missing.

denizsurmeli:
@sfc-gh-jmichalak if I'm not missing anything, the PR should be ready for review

sfc-gh-jmichalak:
@denizsurmeli, thanks for a quick response. I'm on leave for the next 2 weeks, @sfc-gh-jcieslak or @sfc-gh-asawicki will take a look.

denizsurmeli:
> @denizsurmeli, thanks for a quick response. I'm on leave for the next 2 weeks, @sfc-gh-jcieslak or @sfc-gh-asawicki will take a look.

have a nice leave, thanks for the response

MIGRATION_GUIDE.md
```diff
@@ -38,6 +38,10 @@ In this version, this behavior is fixed. The provider always uses the mixed `'"n
 
 References: [#3229](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3229)
 
+
+### *(bugfix)* Fix file format `null_if` empty array permadiff
+Fixed a permadiff issue where file formats with empty `null_if = []` arrays were causing persistent differences in Terraform state.
```
sfc-gh-asawicki: nit: we usually link to the issues that are being solved, please check other entries

pkg/sdk/file_format.go
```diff
@@ -840,7 +842,7 @@ func (v *fileFormats) Describe(ctx context.Context, id SchemaObjectIdentifier) (
 				details.Options.JSONTrimSpace = &b
 			case "NULL_IF":
 				newNullIf := []NullString{}
-				for _, s := range strings.Split(strings.Trim(v, "[]"), ", ") {
+				for _, s := range ParseCommaSeparatedStringArray(v, false) {
```
sfc-gh-asawicki, line 844: is the empty string check needed only for CSV? why?

pkg/sdk/file_format_test.go
```diff
@@ -15,6 +15,17 @@ func TestFileFormatsCreate(t *testing.T) {
 		assertOptsValidAndSQLEquals(t, opts, `CREATE FILE FORMAT %s TYPE = CSV`, id.FullyQualifiedName())
 	})
 
+	t.Run("CSV with empty null_if", func(t *testing.T) {
```
sfc-gh-asawicki: nit: let's add unit test for each format, not only csv

pkg/sdk/testint/file_format_integration_test.go
```diff
@@ -105,6 +105,97 @@ func TestInt_FileFormatsCreateAndRead(t *testing.T) {
 		assert.Equal(t, &sdk.CSVEncodingGB18030, describeResult.Options.CSVEncoding)
 	})
 
+	// check that empty array is correctly read
+	t.Run("CSV", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+		err := client.FileFormats.Create(ctx, id, &sdk.CreateFileFormatOptions{
+			Type: sdk.FileFormatTypeCSV,
+			FileFormatTypeOptions: sdk.FileFormatTypeOptions{
```
sfc-gh-asawicki, line 112: let's simplify this setup. We want to check only the empty CSVNullIf behavior so we don't need all other attributes (the same with the assertions below)

sfc-gh-asawicki (changes requested):
Hey, I have left small comments, otherwise LGTM. Please wait also for @sfc-gh-jcieslak review.

pkg/testacc/resource_file_format_acceptance_test.go
```diff
@@ -493,6 +493,146 @@ func TestAcc_FileFormat_Rename(t *testing.T) {
...
+					resource.TestCheckResourceAttr("snowflake_file_format.test", "name", id.Name()),
+					resource.TestCheckResourceAttr("snowflake_file_format.test", "format_type", "ORC"),
+					resource.TestCheckResourceAttr("snowflake_file_format.test", "null_if.#", "0"),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_FileFormat_PARQUET_EmptyNullIf_NoPlanDiff(t *testing.T) {
```
sfc-gh-jcieslak: We could simplify those tests a bit and use https://go.dev/wiki/TableDrivenTests as they are identical except single value

pkg/testacc/resource_file_format_acceptance_test.go
```diff
@@ -649,3 +789,15 @@ resource "snowflake_file_format" "test" {
...
 }
+
+func fileFormatConfigEmptyNullIf(id sdk.SchemaObjectIdentifier, formatType string) string {
+	return fmt.Sprintf(`
+resource "snowflake_file_format" "test" {
+  name        = "%s"
+  database    = "%s"
+  schema      = "%s"
+  format_type = "%s"
+  null_if     = []
```
sfc-gh-jcieslak: Linked issues in the pr description are referring to the configuration where null if should be set to `null_if = ["''"]` and the provider shouldn't produce any plans to override this empty string. Please provide second test that would replicate configuration from the issues.

pkg/testacc/resource_file_format_acceptance_test.go
```diff
@@ -493,6 +493,146 @@ func TestAcc_FileFormat_Rename(t *testing.T) {
...
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		PreCheck:     func() { TestAccPreCheck(t) },
+		CheckDestroy: CheckDestroy(t, resources.FileFormat),
+		Steps: []resource.TestStep{
+			{
```
sfc-gh-jcieslak: In cases where we provide a fix to the implementation, we usually provide a test where the first step is using the configuration that causes problem in the latest version by using ExternalProviders (like here: https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/testacc/resource_user_authentication_policy_attachment_acceptance_test.go#L135) and in second step you use the current version of the provider (by using https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/testacc/resource_user_authentication_policy_attachment_acceptance_test.go#L154) to show the issue has been fixed. Please, adjust those to follow this pattern.

pkg/sdk/testint/file_format_integration_test.go
```diff
@@ -196,6 +287,40 @@ func TestInt_FileFormatsCreateAndRead(t *testing.T) {
 		assert.True(t, *describeResult.Options.JSONIgnoreUTF8Errors)
 		assert.True(t, *describeResult.Options.JSONSkipByteOrderMark)
 	})
+
+	// check that empty array is correctly read for JSON
+	t.Run("JSON", func(t *testing.T) {
```
sfc-gh-jcieslak: I guess this is also a good candidate for https://go.dev/wiki/TableDrivenTests as there are a lot of repetitions in added tests

MIGRATION_GUIDE.md
sfc-gh-jcieslak: The issues where about setting null_if = ["''"], not null_if = []. Please provide proofs in tests that both were not working correctly and this description

sfc-gh-jcieslak (changes requested):
Thank you for your contribution! I added a few comments in addition to the ones left by Artur.

denizsurmeli:
Hi, I was on a leave. I will work on this PR this week.

sfc-gh-jmichalak:
Hi @denizsurmeli 👋  What's the status of this PR? Could you update?

sfc-gh-jmichalak:
Closing due to inactivity. Note that the linked issues will be worked on during resource rework (no timelines yet).

ndimakosJLP:
Hi @sfc-gh-jmichalak
is there any update on fixing this? This was closed, but it's still needed as it was never fixed

sfc-gh-jmichalak:
Hi @ndimakosJLP 👋 

Yes, fixing this and other file format issues is on our radar for this month - see our [Roadmap](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/ROADMAP.md#feature-gap). Thanks for your patience!

PR #3902, opened 2025-08-04
chore: Simplify listing manifests

sfc-gh-jcieslak:
## Changes
- Add missing optional fields from complete listing example
- Add note about referencing files from the manifest
- Remove target accounts form sample manifests as it is not required to have at least one there


pkg/acceptance/helpers/listing_client.go
```diff
@@ -160,6 +160,6 @@ description: description
 listing_terms:
   type: OFFLINE
 targets:
-  accounts: [%s.%s]
-`, title, subtitle, targetAccount.OrganizationName(), targetAccount.AccountName()), title
+  accounts: []
```
sfc-gh-asawicki: q: do we have any test with a target account for completeness?

sfc-gh-asawicki (approved):

pkg/acceptance/helpers/listing_client.go
sfc-gh-jcieslak: We don't. Modified the helper and one test a bit, so that there is at least one integration test that shows it works properly

sfc-gh-asawicki (approved):

PR #3905, opened 2025-08-05
chore: Adjust grant privileges to share test

sfc-gh-jcieslak:
## Changes 
- Adjust grant privileges to share test to make it consistent with other tests

### DIFF
--- a/pkg/testacc/resource_grant_privileges_to_share_acceptance_test.go
+++ b/pkg/testacc/resource_grant_privileges_to_share_acceptance_test.go
@@ -181,29 +181,24 @@ func TestAcc_GrantPrivilegesToShare_OnTable(t *testing.T) {
 }
 
 func TestAcc_GrantPrivilegesToShare_OnDynamicTable(t *testing.T) {
-	database, databaseCleanup := testClient().Database.CreateDatabaseWithParametersSet(t)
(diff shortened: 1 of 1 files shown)


pkg/testacc/resource_grant_privileges_to_share_acceptance_test.go
```diff
@@ -181,29 +181,24 @@ func TestAcc_GrantPrivilegesToShare_OnTable(t *testing.T) {
 }
 
 func TestAcc_GrantPrivilegesToShare_OnDynamicTable(t *testing.T) {
-	database, databaseCleanup := testClient().Database.CreateDatabaseWithParametersSet(t)
```
sfc-gh-asawicki: please revert dedicated db (it 's like these in all other share tests - as we don't want to have any additional grants on the default db)

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_grant_privileges_to_share_acceptance_test.go
sfc-gh-jcieslak: Done

sfc-gh-asawicki (approved):

PR #3907, opened 2025-08-06
fix: schema default ddl collation handling

sfc-gh-jcieslak:
## Changes
- Fix default_ddl_collation empty string setting from non-empty value. The fix is not ideal, as despite providing a way to set an empty string, it is required to perform in two steps.

## Next steps
- Check other resources for similar issues with string parameters (created SNOW-2251223 for this)

## References

* https://github.com/snowflakedb/terraform-provider-snowflake/issues/3510


pkg/sdk/testint/schemas_integration_test.go
```diff
@@ -320,7 +320,7 @@ func TestInt_Schemas(t *testing.T) {
 				ExternalVolume:                          &externalVolumeTest,
 				Catalog:                                 &catalogIntegrationTest,
 				ReplaceInvalidCharacters:                sdk.Bool(true),
-				DefaultDDLCollation:                     sdk.String("en_US"),
+				DefaultDDLCollation:                     sdk.Pointer(sdk.StringAllowEmpty{Value: "en_US"}),
```
sfc-gh-asawicki, line 323: nit (in one of the next PRs): we should have an integration test with setting to empty

sfc-gh-asawicki (approved):

pkg/sdk/testint/schemas_integration_test.go
sfc-gh-jcieslak: Noted

sfc-gh-jmichalak (approved):

PR #3915, opened 2025-08-06
chore: Dev to main merge

sfc-gh-jcieslak:
Merge dev to main to release a new provider version


sfc-gh-asawicki (approved):
Failing tests are fine:
1. task tests depend on account-level param setting; we will move the to separate account-level pipeline
2. PAT behavior was fixed, it does not affect the resource (it handles alters separately due to the earlier incorrect behavior)
3. Hybrid tests are not caused by the faulty SQL current Snowflake unavailability).

PR #3925, opened 2025-08-10
Sarur snow 2176806 sdk

sfc-gh-sarur:
adding semantic view support

## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*

### DIFF
--- a/pkg/sdk/semantic_view_def.go
+++ b/pkg/sdk/semantic_view_def.go
@@ -40,10 +40,31 @@ var SemanticViewsDef = g.NewInterface(
 		WithValidation(g.ConflictingFields, "IfNotExists", "OrReplace"),
 ).DropOperation(
 	"https://docs.snowflake.com/en/sql-reference/sql/drop-semantic-view",
-	g.NewQueryStruct("DropGitRepository").
+	g.NewQueryStruct("DropSemanticView").
 		Drop().
 		SQL("SEMANTIC VIEW").
 		IfExists().
 		Name().
 		WithValidation(g.ValidIdentifier, "name"),
+).DescribeOperation(
+	g.DescriptionMappingKindSingleValue,
+	"https://docs.snowflake.com/en/sql-reference/sql/desc-semantic-view",
+	semanticViewDbRow,
+	semanticView,
+	g.NewQueryStruct("DescribeSemanticView").
+		Describe().
+		SQL("SEMANTIC VIEW").
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).ShowOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/show-semantic-views",
+	gitRepositoryDbRow,
+	gitRepository,
+	g.NewQueryStruct("ShowSemanticViews").
+		Show().
+		Terse().
+		SQL("SEMANTIC VIEWS").
+		OptionalLike().
+		OptionalIn().
+		OptionalLimit(),
 )
--- a/pkg/sdk/semantic_view_def.go
+++ b/pkg/sdk/semantic_view_def.go
@@ -58,8 +58,8 @@ var SemanticViewsDef = g.NewInterface(
 		WithValidation(g.ValidIdentifier, "name"),
 ).ShowOperation(
 	"https://docs.snowflake.com/en/sql-reference/sql/show-semantic-views",
-	gitRepositoryDbRow,
-	gitRepository,
+	semanticViewDbRow,
+	semanticView,
 	g.NewQueryStruct("ShowSemanticViews").
 		Show().
 		Terse().
--- a/pkg/sdk/semantic_view_def.go
+++ b/pkg/sdk/semantic_view_def.go
@@ -1,0 +1,49 @@
+package sdk
+
+import g "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/poc/generator"
+
+//go:generate go run ./poc/main.go
+
+var semanticViewDbRow = g.DbStruct("semanticViewsRow").
+	Time("created_on").
+	Text("name").
+	Text("database_name").
+	Text("schema_name").
+	Text("owner").
+	Text("owner_role_type").
+	OptionalText("comment")
+
+var semanticView = g.PlainStruct("semanticView").
+	Time("CreatedOn").
+	Text("Name").
+	Text("DatabaseName").
+	Text("SchemaName").
+	Text("Owner").
+	Text("OwnerRoleType").
+	OptionalText("Comment")
+
+var SemanticViewsDef = g.NewInterface(
+	"SemanticViews",
+	"SemanticView",
+	g.KindOfT[SchemaObjectIdentifier](),
+).CreateOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/create-semantic-view",
+	g.NewQueryStruct("CreateSemanticView").
+		Create().
+		OrReplace().
+		SQL("SEMANTIC VIEW").
+		IfNotExists().
+		Name().
+		OptionalComment().
+		OptionalCopyGrants().
+		WithValidation(g.ValidIdentifier, "name").
+		WithValidation(g.ConflictingFields, "IfNotExists", "OrReplace"),
+).DropOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/drop-semantic-view",
+	g.NewQueryStruct("DropGitRepository").
+		Drop().
+		SQL("SEMANTIC VIEW").
+		IfExists().
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+)
--- a/pkg/sdk/semantic_view_dto_gen.go
+++ b/pkg/sdk/semantic_view_dto_gen.go
@@ -3,8 +3,10 @@ package sdk
 //go:generate go run ./dto-builder-generator/main.go
 
 var (
-	_ optionsProvider[CreateSemanticViewOptions] = new(CreateSemanticViewRequest)
-	_ optionsProvider[DropSemanticViewOptions]   = new(DropSemanticViewRequest)
+	_ optionsProvider[CreateSemanticViewOptions]   = new(CreateSemanticViewRequest)
+	_ optionsProvider[DropSemanticViewOptions]     = new(DropSemanticViewRequest)
+	_ optionsProvider[DescribeSemanticViewOptions] = new(DescribeSemanticViewRequest)
+	_ optionsProvider[ShowSemanticViewsOptions]    = new(ShowSemanticViewsRequest)
 )
 
 type CreateSemanticViewRequest struct {
@@ -16,9 +18,18 @@ type CreateSemanticViewRequest struct {
 }
 
 type DropSemanticViewRequest struct {
-	OrReplace   *bool
-	IfNotExists *bool
-	name        SchemaObjectIdentifier // required
-	Comment     *string
-	CopyGrants  *bool
+	IfExists *bool
+	name     SchemaObjectIdentifier // required
+}
+
+type DescribeSemanticViewRequest struct {
+	name SchemaObjectIdentifier // required
+}
+
+type ShowSemanticViewsRequest struct {
+	Like       *Like
+	Terse      *bool
+	In         *In
+	StartsWith *string
+	Limit      *LimitFrom
 }
--- a/pkg/sdk/semantic_view_dto_gen.go
+++ b/pkg/sdk/semantic_view_dto_gen.go
@@ -1,0 +1,24 @@
+package sdk
+
+//go:generate go run ./dto-builder-generator/main.go
+
+var (
+	_ optionsProvider[CreateSemanticViewOptions] = new(CreateSemanticViewRequest)
+	_ optionsProvider[DropSemanticViewOptions]   = new(DropSemanticViewRequest)
+)
+
+type CreateSemanticViewRequest struct {
+	OrReplace   *bool
+	IfNotExists *bool
+	name        SchemaObjectIdentifier // required
+	Comment     *string
+	CopyGrants  *bool
+}
+
+type DropSemanticViewRequest struct {
+	OrReplace   *bool
+	IfNotExists *bool
+	name        SchemaObjectIdentifier // required
+	Comment     *string
+	CopyGrants  *bool
+}
--- a/pkg/sdk/semantic_view_gen_test.go
+++ b/pkg/sdk/semantic_view_gen_test.go
@@ -74,3 +74,61 @@ func TestSemanticViews_Drop(t *testing.T) {
 		assertOptsValidAndSQLEquals(t, opts, `DROP SEMANTIC VIEW IF EXISTS %s`, id.FullyQualifiedName())
 	})
 }
+
+func TestSemanticViews_Describe(t *testing.T) {
+	id := randomSchemaObjectIdentifier()
+
+	defaultOpts := func() *DescribeSemanticViewOptions {
+		return &DescribeSemanticViewOptions{
+			name: id,
+		}
+	}
+
+	t.Run("validation: nil options", func(t *testing.T) {
+		var opts *DescribeSemanticViewOptions = nil
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
+	})
+
+	t.Run("validation: valid identifier for [opts.name]", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.name = emptySchemaObjectIdentifier
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("all options", func(t *testing.T) {
+		opts := defaultOpts()
+		assertOptsValidAndSQLEquals(t, opts, "DESCRIBE SEMANTIC VIEW %s", id.FullyQualifiedName())
+	})
+}
+
+func TestSemanticViews_Show(t *testing.T) {
+	defaultOpts := func() *ShowSemanticViewsOptions {
+		return &ShowSemanticViewsOptions{}
+	}
+
+	t.Run("validation: nil options", func(t *testing.T) {
+		var opts *ShowSemanticViewsOptions = nil
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
+	})
+
+	t.Run("basic", func(t *testing.T) {
+		opts := defaultOpts()
+		assertOptsValidAndSQLEquals(t, opts, "SHOW SEMANTIC VIEWS")
+	})
+
+	t.Run("all options", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.Terse = Bool(true)
+		opts.Like = &Like{
+			Pattern: String("semantic-view-name"),
+		}
+		opts.In = &In{
+			Database: NewAccountObjectIdentifier("database-name"),
+		}
+		opts.StartsWith = String("semantic-view-name-start")
+		opts.Limit = &LimitFrom{
+			Rows: Int(10),
+		}
+		assertOptsValidAndSQLEquals(t, opts, `SHOW TERSE SEMANTIC VIEWS LIKE 'semantic-view-name' IN DATABASE "database-name" STARTS WITH 'semantic-view-name-start' LIMIT 10`)
+	})
+}
--- a/pkg/sdk/semantic_view_gen_test.go
+++ b/pkg/sdk/semantic_view_gen_test.go
@@ -1,0 +1,76 @@
+package sdk
+
+import "testing"
+
+func TestSemanticViews_Create(t *testing.T) {
+	id := randomSchemaObjectIdentifier()
+
+	defaultOpts := func() *CreateSemanticViewOptions {
+		return &CreateSemanticViewOptions{
+			name: id,
+		}
+	}
+
+	t.Run("validation: nil options", func(t *testing.T) {
+		var opts *CreateSemanticViewOptions = nil
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
+	})
+
+	t.Run("validation: valid identifier for [opts.name]", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.name = emptySchemaObjectIdentifier
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("validation: conflicting fields for [opts.IfNotExists opts.OrReplace]", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.IfNotExists = Bool(true)
+		opts.OrReplace = Bool(true)
+		assertOptsInvalidJoinedErrors(t, opts, errOneOf("CreateSemanticViewOptions", "IfNotExists", "OrReplace"))
+	})
+
+	t.Run("basic", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.OrReplace = Bool(true)
+		assertOptsValidAndSQLEquals(t, opts, "CREATE OR REPLACE SEMANTIC VIEW %s", id.FullyQualifiedName())
+	})
+
+	t.Run("all options", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.IfNotExists = Bool(true)
+		opts.Comment = String("comment")
+		assertOptsValidAndSQLEquals(t, opts, `CREATE SEMANTIC VIEW IF NOT EXISTS %s COMMENT = '%s'`, id.FullyQualifiedName(), "comment")
+	})
+}
+
+func TestSemanticViews_Drop(t *testing.T) {
+	id := randomSchemaObjectIdentifier()
+
+	defaultOpts := func() *DropSemanticViewOptions {
+		return &DropSemanticViewOptions{
+			name: id,
+		}
+	}
+
+	t.Run("validation: nil options", func(t *testing.T) {
+		var opts *DropSemanticViewOptions = nil
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
+	})
+
+	t.Run("validation: valid identifier for [opts.name]", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.name = emptySchemaObjectIdentifier
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("basic", func(t *testing.T) {
+		opts := defaultOpts()
+		assertOptsValidAndSQLEquals(t, opts, `DROP SEMANTIC VIEW %s`, id.FullyQualifiedName())
+	})
+
+	t.Run("all options", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.IfExists = Bool(true)
+		assertOptsValidAndSQLEquals(t, opts, `DROP SEMANTIC VIEW IF EXISTS %s`, id.FullyQualifiedName())
+	})
+}
--- a/pkg/sdk/client.go
+++ b/pkg/sdk/client.go
@@ -67,6 +67,7 @@ type Client struct {
 	Schemas                      Schemas
 	Secrets                      Secrets
 	SecurityIntegrations         SecurityIntegrations
+	SemanticViews                SemanticViews
 	Services                     Services
 	Sequences                    Sequences
 	SessionPolicies              SessionPolicies
@@ -198,6 +199,7 @@ func (c *Client) initialize() {
 	c.Schemas = &schemas{client: c}
 	c.Secrets = &secrets{client: c}
 	c.SecurityIntegrations = &securityIntegrations{client: c}
+	c.SemanticViews = &semanticViews{client: c}
 	c.Sequences = &sequences{client: c}
 	c.Services = &services{client: c}
 	c.SessionPolicies = &sessionPolicies{client: c}
--- a/pkg/sdk/poc/main.go
+++ b/pkg/sdk/poc/main.go
@@ -56,6 +56,7 @@ var definitionMapping = map[string]*generator.Interface{
 	"user_programmatic_access_tokens_def.go": sdk.UserProgrammaticAccessTokensDef,
 	"listings_def.go":                        sdk.ListingsDef,
 	"organization_accounts_def.go":           sdk.OrganizationAccountsDef,
+	"semantic_view_def.go":                   sdk.SemanticViewsDef,
 }
 
 func main() {
--- a/pkg/sdk/semantic_view_validations_gen.go
+++ b/pkg/sdk/semantic_view_validations_gen.go
@@ -29,3 +29,22 @@ func (opts *DropSemanticViewOptions) validate() error {
 	}
 	return JoinErrors(errs...)
 }
+
+func (opts *DescribeSemanticViewOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	if !ValidObjectIdentifier(opts.name) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	return JoinErrors(errs...)
+}
+
+func (opts *ShowSemanticViewsOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	return JoinErrors(errs...)
+}
--- a/pkg/sdk/semantic_view_validations_gen.go
+++ b/pkg/sdk/semantic_view_validations_gen.go
@@ -1,0 +1,31 @@
+package sdk
+
+var (
+	_ validatable = new(CreateSemanticViewOptions)
+	_ validatable = new(DropSemanticViewOptions)
+)
+
+func (opts *CreateSemanticViewOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	if !ValidObjectIdentifier(opts.name) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	if everyValueSet(opts.IfNotExists, opts.OrReplace) {
+		errs = append(errs, errOneOf("CreateSemanticViewOptions", "IfNotExists", "OrReplace"))
+	}
+	return JoinErrors(errs...)
+}
+
+func (opts *DropSemanticViewOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	if !ValidObjectIdentifier(opts.name) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	return JoinErrors(errs...)
+}
--- a/pkg/sdk/semantic_views_gen.go
+++ b/pkg/sdk/semantic_views_gen.go
@@ -15,8 +15,9 @@ type CreateSemanticViewOptions struct {
 	semanticView bool                   `ddl:"static" sql:"SEMANTIC VIEW"`
 	IfNotExists  *bool                  `ddl:"keyword" sql:"IF NOT EXISTS"`
 	name         SchemaObjectIdentifier `ddl:"identifier"`
-	Comment      *string                `ddl:"parameter,single_quotes" sql:"COMMENT"`
-	CopyGrants   *bool                  `ddl:"keyword" sql:"COPY GRANTS"`
+
+	Comment    *string `ddl:"parameter,single_quotes" sql:"COMMENT"`
+	CopyGrants *bool   `ddl:"keyword" sql:"COPY GRANTS"`
 }
 
 type DropSemanticViewOptions struct {
@@ -24,4 +25,20 @@ type DropSemanticViewOptions struct {
 	semanticView bool                   `ddl:"static" sql:"SEMANTIC VIEW"`
 	IfExists     *bool                  `ddl:"keyword" sql:"IF EXISTS"`
 	name         SchemaObjectIdentifier `ddl:"identifier"`
+}
+
+type DescribeSemanticViewOptions struct {
+	describe     bool                   `ddl:"static" sql:"DESCRIBE"`
+	semanticView bool                   `ddl:"static" sql:"SEMANTIC VIEW"`
+	name         SchemaObjectIdentifier `ddl:"identifier"`
+}
+
+type ShowSemanticViewsOptions struct {
+	show          bool       `ddl:"static" sql:"SHOW"`
+	Terse         *bool      `ddl:"keyword" sql:"TERSE"`
+	semanticViews bool       `ddl:"static" sql:"SEMANTIC VIEWS"`
+	Like          *Like      `ddl:"keyword" sql:"LIKE"`
+	In            *In        `ddl:"keyword" sql:"IN"`
+	StartsWith    *string    `ddl:"parameter,no_equals,single_quotes" sql:"STARTS WITH"`
+	Limit         *LimitFrom `ddl:"parameter,no_equals" sql:"LIMIT"`
 }
--- a/pkg/sdk/semantic_views_gen.go
+++ b/pkg/sdk/semantic_views_gen.go
@@ -1,0 +1,27 @@
+package sdk
+
+import (
+	"context"
+)
+
+type SemanticViews interface {
+	Create(ctx context.Context, request *CreateSemanticViewRequest) error
+	Drop(ctx context.Context, request *DropSemanticViewRequest) error
+}
+
+type CreateSemanticViewOptions struct {
+	create       bool                   `ddl:"static" sql:"CREATE"`
+	OrReplace    *bool                  `ddl:"keyword" sql:"OR REPLACE"`
+	semanticView bool                   `ddl:"static" sql:"SEMANTIC VIEW"`
+	IfNotExists  *bool                  `ddl:"keyword" sql:"IF NOT EXISTS"`
+	name         SchemaObjectIdentifier `ddl:"identifier"`
+	Comment      *string                `ddl:"parameter,single_quotes" sql:"COMMENT"`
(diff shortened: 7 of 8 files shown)


pkg/sdk/identifier_helpers.go
```diff
@@ -453,3 +453,32 @@ func (i TableColumnIdentifier) FullyQualifiedName() string {
 func (i TableColumnIdentifier) SchemaObjectId() SchemaObjectIdentifier {
 	return NewSchemaObjectIdentifier(i.databaseName, i.schemaName, i.tableName)
 }
+
+type LogicalTableIdentifier struct {
```
sfc-gh-sarur: The create semantic view command spec is here https://docs.snowflake.com/en/sql-reference/sql/create-semantic-view
Is this appropriate implementation of the fields?

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,81 @@
...
+		SQL("TABLES").
+		OptionalSQL("RELATIONSHIPS").
+		OptionalSQL("FACTS").
+		OptionalSQL("DIMENSIONS").
+		OptionalSQL("METRICS").
+		OptionalComment().
+		OptionalCopyGrants().
+		WithValidation(g.ValidIdentifier, "name").
+		WithValidation(g.ValidIdentifier, "tables").
+		WithValidation(g.ValidIdentifierIfSet, "relationships").
```
sfc-gh-sarur: I am not sure what validationTypes to use for these custom fields, should I define new ones to use here?

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,81 @@
...
+).ShowOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/show-semantic-views",
+	semanticViewDbRow,
+	semanticView,
+	g.NewQueryStruct("ShowSemanticViews").
+		Show().
+		Terse().
+		SQL("SEMANTIC VIEWS").
+		OptionalLike().
+		OptionalIn().
```
sfc-gh-sarur: the IN field has to follow a specific format defined here https://docs.snowflake.com/en/sql-reference/sql/show-semantic-views
What is the right way to enforce that? Should that be handled in the validation code?

pkg/sdk/semantic_view_gen_test.go
```diff
@@ -0,0 +1,134 @@
+package sdk
+
+import "testing"
+
+func TestSemanticViews_Create(t *testing.T) {
```
sfc-gh-sarur, line 5: I will add more test cases here as I implement the validation logic for create semantic views.

pkg/sdk/identifier_helpers.go
sfc-gh-asawicki: 1. These structs shouldn't be added to the identifier_helpers. I don't think there is a need to create a new identifier type, as semantic views are schema-level objects, so SchemaObjectIdentifier can id them.
2. These structs shouldn't be written by hand but generated based on def (check the comment in the def file).

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,81 @@
...
+).CreateOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/create-semantic-view",
+	g.NewQueryStruct("CreateSemanticView").
+		Create().
+		OrReplace().
+		SQL("SEMANTIC VIEW").
+		IfNotExists().
+		Name().
+		SQL("TABLES").
+		OptionalSQL("RELATIONSHIPS").
```
sfc-gh-asawicki: SQL or OptionalSQL are used in this DSL to mark SQL syntax where there is a given "word" in SQL that must appear or is optional, e.g. `SQL("SEMANTIC VIEW")` above. It should not be used for clauses which contain much more inside, like:
```
TABLES logicalTable

logicalTable ::=
  [ <table_alias> AS ] <table_name>
  [ PRIMARY KEY ( <primary_key_column_name> [ , ... ] ) ]
  [
    UNIQUE ( <unique_column_name> [ , ... ] )
    [ ... ]
  ]
  [ WITH SYNONYMS [ = ] ( '<synonym>' [ , ... ] ) ]
  [ COMMENT = '<comment_about_table>' ]
```
To handle these, you should define `logicalTable` using `g.NewQueryStruct` and then use it here using e.g. `OptionalQueryStructField(...)`. PLease take a look at other definition files, e.g. https://github.com/snowflakedb/terraform-provider-snowflake/blob/f80c1595e0267ec922c710dc23be1ae0bad5ee64/pkg/sdk/views_def.go#L142 and https://github.com/snowflakedb/terraform-provider-snowflake/blob/f80c1595e0267ec922c710dc23be1ae0bad5ee64/pkg/sdk/views_def.go#L264.

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,81 @@
...
+		Drop().
+		SQL("SEMANTIC VIEW").
+		IfExists().
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).DescribeOperation(
+	g.DescriptionMappingKindSingleValue,
+	"https://docs.snowflake.com/en/sql-reference/sql/desc-semantic-view",
+	semanticViewDbRow,
+	semanticView,
```
sfc-gh-jcieslak: By looking at the documentation, it seems the describe has a different output than the SHOW command. Create a new DbStruct and PlainStruct for this output if they are different.

pkg/sdk/semantic_view_def.go
sfc-gh-asawicki: These validations shouldn't cover every use case; these are more like sanity checks for the correct syntax for the SQL. This one is the correct one: `WithValidation(g.AtLeastOneValueSet, "dimensions", "metrics")`. For now, you can remove all:
```
		WithValidation(g.ValidIdentifier, "tables").
		WithValidation(g.ValidIdentifierIfSet, "relationships").
		WithValidation(g.ValidIdentifierIfSet, "facts").
		WithValidation(g.ValidIdentifierIfSet, "dimensions").
		WithValidation(g.ValidIdentifierIfSet, "metrics").
```
You should add validation directly inside the query struct definitions (check the comment above). So, e.g. windowFunctionMetricDefinition can have`PARTITION BY { <exprs_using_dimensions_or_metrics> | EXCLUDING <dimensions> } ]` so they will probably use `g.ExactlyOneValueSet` for expression or excluding or `g.ConflictingFields` depending on how you will structure this.

pkg/sdk/semantic_view_def.go
sfc-gh-asawicki: OptionalIn() uses https://github.com/snowflakedb/terraform-provider-snowflake/blob/f80c1595e0267ec922c710dc23be1ae0bad5ee64/pkg/sdk/common_types.go#L56 which is exactly what you want. You don't need to change anything to cover the referenced syntax for IN clause.

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,81 @@
...
+).DropOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/drop-semantic-view",
+	g.NewQueryStruct("DropSemanticView").
+		Drop().
+		SQL("SEMANTIC VIEW").
+		IfExists().
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).DescribeOperation(
+	g.DescriptionMappingKindSingleValue,
```
sfc-gh-jcieslak: From the docs, it seems more like a g.DescriptionMappingKindSlice where the describe output contains more than one row for a given object. Please, check different similar described outputs for reference (e.g. stages_def.go).

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,81 @@
...
+	"https://docs.snowflake.com/en/sql-reference/sql/show-semantic-views",
+	semanticViewDbRow,
+	semanticView,
+	g.NewQueryStruct("ShowSemanticViews").
+		Show().
+		Terse().
+		SQL("SEMANTIC VIEWS").
+		OptionalLike().
+		OptionalIn().
+		OptionalLimit(),
```
sfc-gh-jcieslak: Minor thing, but you're missing startsWith option

pkg/sdk/semantic_view_dto_gen.go
```diff
@@ -0,0 +1,40 @@
...
+	_ optionsProvider[DropSemanticViewOptions]     = new(DropSemanticViewRequest)
+	_ optionsProvider[DescribeSemanticViewOptions] = new(DescribeSemanticViewRequest)
+	_ optionsProvider[ShowSemanticViewsOptions]    = new(ShowSemanticViewsRequest)
+)
+
+type CreateSemanticViewRequest struct {
+	OrReplace     *bool
+	IfNotExists   *bool
+	name          SchemaObjectIdentifier   // required
+	tables        []LogicalTableIdentifier // required
```
sfc-gh-jcieslak: If you are doing any manual changes within the auto-generated files `*_gen.go`, please comment those places with e.g., "adjusted manually". This part should be auth-generated anyway (see [this comment](https://github.com/snowflakedb/terraform-provider-snowflake/pull/3925/files#r2265850552)), but that's a note for future changes. It helps during review and later re-generations (e.g. another person that didn't develop this resource would know by the comment that a given line shouldn't be overridden by the generator).

pkg/sdk/semantic_view_gen_test.go
sfc-gh-jcieslak: Just a note, the validation doesn't have to be exhaustive. We will add more validation on the Terraform schema level to ensure, e.g., the correct combination of specified fields. For wrong inputs with correct structure (e.g., a correct identifier that references a non-existing object in Snowflake), we let Snowflake fail and return the error, which we propagate to the users.

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,81 @@
...
+)
```
sfc-gh-jcieslak: Please, add `ShowByIdOperationWithFiltering(g.ShowByIDInFiltering, g.ShowByIDLikeFiltering).`

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: Thank you, that makes sense.
However, since at least one logical table is required in the create and more than one can be included, should I use the `ListQueryStructField` instead?

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: sounds good.

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: added, thanks!

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: done

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,117 @@
...
+	ListAssignment("UNIQUE", "SemanticViewColumn", g.ParameterOptions().Parentheses().NoEquals())
+
+var synonym = g.NewQueryStruct("Synonyms").
+	ListAssignment("WITH SYNONYMS", "string", g.ParameterOptions().NoEquals().Parentheses())
+
+var logicalTableAlias = g.NewQueryStruct("LogicalTableAlias").
+	Text("LogicalTableAlias", g.KeywordOptions()).
+	SQL("AS")
+
+var logicalTable = g.NewQueryStruct("LogicalTable").
```
sfc-gh-sarur, line 135: @sfc-gh-asawicki @sfc-gh-jcieslak I believe I have the logical tables working properly, PTAL and let me know what you think.

pkg/sdk/semantic_view_gen_test.go
sfc-gh-sarur: ack

pkg/sdk/semantic_view_def.go
```diff
@@ -9,36 +9,34 @@ var semanticViewDbRow = g.DbStruct("semanticViewDBRow").
...
 var semanticView = g.PlainStruct("SemanticView").
 	Time("CreatedOn").
 	Text("Name").
 	Text("DatabaseName").
 	Text("SchemaName").
+	OptionalText("Comment").
 	Text("Owner").
 	Text("OwnerRoleType").
-	OptionalText("Comment")
+	OptionalText("Extension")
```
sfc-gh-jcieslak, line 25: It's not publicly documented, do we need it?

pkg/sdk/semantic_view_def.go
```diff
@@ -52,11 +50,23 @@ var SemanticViewsDef = g.NewInterface(
 		SQL("SEMANTIC VIEW").
 		IfNotExists().
 		Name().
-		ListQueryStructField("tables", logicalTable, g.ListOptions().Parentheses().Required()).
+		SQL("TABLES").
+		ListQueryStructField("logicalTables", logicalTable, g.ListOptions().Required().Parentheses()).
+		OptionalSQL("RELATIONSHIPS").
+		ListQueryStructField("semanticViewRelationships", semanticViewRelationship, g.ListOptions().Parentheses()).
+		OptionalSQL("FACTS").
+		ListQueryStructField("semanticViewFacts", semanticExpression, g.ListOptions().Parentheses()).
```
sfc-gh-jcieslak: move optional SQLs to the listQueryStructFields (ListOption() should have SQL method)

pkg/sdk/semantic_view_def.go
```diff
@@ -89,5 +99,54 @@ var SemanticViewsDef = g.NewInterface(
...
+	ListAssignment("PRIMARY KEY", "SemanticViewColumn", g.ParameterOptions().Parentheses().NoEquals())
+
+var uniqueKey = g.NewQueryStruct("UniqueKeys").
+	ListAssignment("UNIQUE", "SemanticViewColumn", g.ParameterOptions().Parentheses().NoEquals())
+
+var synonym = g.NewQueryStruct("Synonyms").
+	ListAssignment("WITH SYNONYMS", "string", g.ParameterOptions().NoEquals().Parentheses())
+
+var logicalTableAlias = g.NewQueryStruct("LogicalTableAlias").
+	Text("LogicalTableAlias", g.KeywordOptions()).
```
sfc-gh-jcieslak: nit: `KeywordOptions().Required()` for query structs with one field you can mark this field as required. It's `nit` so you can skip it for now. The only difference it would make for Request builder. Currently, It will be `NewLogicalTableAliasRequest().WithLogicalTableAlias(alias)` vs `NewLogicalTableAliasRequest(alias)`

pkg/sdk/semantic_view_def.go
```diff
@@ -89,5 +99,54 @@ var SemanticViewsDef = g.NewInterface(
...
+var relationshipAlias = g.NewQueryStruct("RelationshipAlias").
+	Text("RelationshipAlias", g.KeywordOptions()).
+	SQL("AS")
+
+var relationshipTableAlias = g.NewQueryStruct("RelationshipTableAlias").
+	Text("RelationshipTableAlias", g.KeywordOptions())
+
+var semanticViewRelationship = g.NewQueryStruct("SemanticViewRelationship").
+	OptionalQueryStructField("relationshipAlias", relationshipAlias, g.KeywordOptions()).
+	OptionalQueryStructField("tableName", relationshipTableAlias, g.KeywordOptions().Required()).
```
sfc-gh-jcieslak: is it table name or table alias? The documentation says table_alias and I'm not sure what it is. If it's table_name then it should use Identifier method (it's required so no Optional prefix). Maybe also the documentation in Snowflake should be changed?

pkg/sdk/semantic_view_def.go
```diff
@@ -89,5 +99,54 @@ var SemanticViewsDef = g.NewInterface(
...
+	ListQueryStructField("relationshipColumnNames", semanticViewColumn, g.ListOptions().NoEquals().Parentheses().Required()).
+	SQL("REFERENCES").
+	OptionalQueryStructField("refTableName", relationshipTableAlias, g.KeywordOptions().Required()).
```
sfc-gh-jcieslak: Same here, I think it should be identifier.

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,152 @@
...
+		OptionalIn().
+		OptionalStartsWith().
+		OptionalLimit(),
+)
+
+var primaryKey = g.NewQueryStruct("PrimaryKeys").
+	ListAssignment("PRIMARY KEY", "SemanticViewColumn", g.ParameterOptions().Parentheses().NoEquals())
+
+var uniqueKey = g.NewQueryStruct("UniqueKeys").
+	ListAssignment("UNIQUE", "SemanticViewColumn", g.ParameterOptions().Parentheses().NoEquals())
```
sfc-gh-sarur: does the "UNIQUE" keyword need to be repeated for each row?

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: Discussed during the meeting. It's fine, we'll check on our side.

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: I think this makes sense. The "at least one" validation could be added by manually adjusting the generated validations, but let's skip this for now. We'll add this kind of validation during the resource implementation, which will lead to a correctly filled request.

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: Discussed during the meeting, it's correct

pkg/sdk/semantic_view_def.go
```diff
@@ -89,5 +99,54 @@ var SemanticViewsDef = g.NewInterface(
...
 )
 
+var primaryKey = g.NewQueryStruct("PrimaryKeys").
+	ListAssignment("PRIMARY KEY", "SemanticViewColumn", g.ParameterOptions().Parentheses().NoEquals())
+
+var uniqueKey = g.NewQueryStruct("UniqueKeys").
+	ListAssignment("UNIQUE", "SemanticViewColumn", g.ParameterOptions().Parentheses().NoEquals())
+
+var synonym = g.NewQueryStruct("Synonyms").
+	ListAssignment("WITH SYNONYMS", "string", g.ParameterOptions().NoEquals().Parentheses())
```
sfc-gh-jcieslak: Test if single quotes are used for items inside

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: The UNIQUE key does have to be repeated, verified it and added it to the test case.

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: they seem good to me

pkg/sdk/semantic_view_gen_test.go
```diff
@@ -0,0 +1,257 @@
...
+				sqlExpression:           &SemanticSqlExpression{SqlExpression: dimensionExpression},
+				synonyms:                &Synonyms{WithSynonyms: []string{"'test3'", "'test4'"}},
+				Comment:                 String("dimension_comment"),
+			},
+		}
+		metricsObj := []SemanticExpression{
+			{
+				qualifiedExpressionName: &QualifiedExpressionName{QualifiedExpressionName: metricName},
+				sqlExpression:           &SemanticSqlExpression{SqlExpression: metricExpression},
+				synonyms:                &Synonyms{WithSynonyms: []string{"'test5'", "'test6'"}},
```
sfc-gh-jcieslak: The synonyms should be passed as a regular string, and the proper quoting in SQL should be specified in the definition file

pkg/sdk/semantic_view_gen_test.go
```diff
@@ -0,0 +1,257 @@
...
+			semanticViewFacts:         factsObj,
+			Dimensions:                Bool(true),
+			semanticViewDimensions:    dimensionsObj,
+			Metrics:                   Bool(true),
+			semanticViewMetrics:       metricsObj,
+		}
+		assertOptsValidAndSQLEquals(t, opts, `CREATE SEMANTIC VIEW IF NOT EXISTS %s TABLES (%s AS %s PRIMARY KEY (pk1.1, pk1.2) UNIQUE (uk1.3) UNIQUE (uk1.4) WITH SYNONYMS ('test1', 'test2') COMMENT = '%s', %s AS %s PRIMARY KEY (pk2.1, pk2.2) WITH SYNONYMS ('test3', 'test4') COMMENT = '%s') RELATIONSHIPS (%s AS %s (pk1.1, pk1.2) REFERENCES %s (pk2.1, pk2.2)) FACTS (%s AS %s WITH SYNONYMS ('test1', 'test2') COMMENT = '%s') DIMENSIONS (%s AS %s WITH SYNONYMS ('test3', 'test4') COMMENT = '%s') METRICS (%s AS %s WITH SYNONYMS ('test5', 'test6') COMMENT = '%s') COMMENT = '%s'`, id.FullyQualifiedName(), tableAlias1, logicalTableId1.FullyQualifiedName(), *logicalTableComment1, tableAlias2, logicalTableId2.FullyQualifiedName(), *logicalTableComment2, relationshipAlias1, tableAlias1, tableAlias2, factName, factExpression, *factsObj[0].Comment, dimensionName, dimensionExpression, *dimensionsObj[0].Comment, metricName, metricExpression, *metricsObj[0].Comment, "comment")
```
sfc-gh-jcieslak: I'm guessing primary key and unique column names should be wrapped in double quotes as they are case-sensitive (also to be defined in the definition file). Every identifier or name that can be wrapped within double quotes should be wrapped in them in the final SQL because then this identifier/name will be case-sensitive, which is important in our case.

sfc-gh-jcieslak (dismissed):

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: Saw there's a test for it, but the quoting for synonyms inside was incorrect (there was no quoting when documentation says single quotes should be used).

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,152 @@
...
+	semanticViewDbRow,
+	semanticView,
+	g.NewQueryStruct("ShowSemanticViews").
+		Show().
+		Terse().
+		SQL("SEMANTIC VIEWS").
+		OptionalLike().
+		OptionalIn().
+		OptionalStartsWith().
+		OptionalLimit(),
```
sfc-gh-jmichalak:
```suggestion
		OptionalLimitFrom(),
```

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,152 @@
...
+	semanticExpression,
+).DropOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/drop-semantic-view",
+	g.NewQueryStruct("DropSemanticView").
+		Drop().
+		SQL("SEMANTIC VIEW").
+		IfExists().
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).DescribeOperation(
```
sfc-gh-jmichalak: It seems that ALTER is missing. Even though it only has 1 field, we should still support it.

pkg/sdk/semantic_view_gen_test.go
sfc-gh-jmichalak: Precisely, we can use the `SingleQuotes` method.

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,152 @@
...
+
+var semanticSqlExpression = g.NewQueryStruct("SemanticSqlExpression").
+	Text("SqlExpression", g.KeywordOptions().NoQuotes())
+
+var semanticExpression = g.NewQueryStruct("SemanticExpression").
+	OptionalQueryStructField("qualifiedExpressionName", qualifiedExpressionName, g.KeywordOptions().Required()).
+	SQL("AS").
+	OptionalQueryStructField("sqlExpression", semanticSqlExpression, g.KeywordOptions().Required()).
+	OptionalQueryStructField("synonyms", synonym, g.ParameterOptions().NoEquals()).
+	OptionalComment()
```
sfc-gh-jmichalak, line 171: <p class='graphite__hidden'><i>[Re: line +153]</i></p>
Question: Are we planning to implement windowFunctionMetricDefinition?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3925?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/sdk/semantic_view_impl_gen.go
```diff
@@ -0,0 +1,233 @@
...
+					u[j] = SemanticViewColumn{
+						Name: w.Name,
+					}
+				}
+				s[i].relationshipRefColumnNames = u
+			}
+		}
+		opts.semanticViewRelationships = s
+		opts.Relationships = Bool(true)
+	}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: I think this is not relevant, provided we adjust the optional query structs, like we discussed above.

sfc-gh-jmichalak (changes requested):
I also agree with other comments.

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: looks like they just added the alter command to the documentation, I have added it to the def

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: for my understanding, what is difference between using Limit and LimitFrom?

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: The synonyms should be specified in this format `WITH SYNONYMS ('test1', 'test2')`, and the test is checking against that. I tried writing the ListAssignment as `ListAssignment("WITH SYNONYMS", "string", g.ParameterOptions().NoEquals().Parentheses().SingleQuotes())` but that ends up creating the SQL like `WITH SYNONYMS '(test1, test2)'` instead. How can I fix that? Is it better to enforce that somewhere else?

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: In the semantic view creation if we have specified an alias for the table, then we have to use the alias here. If we did not assign an alias to the table, then we have to use the qualified table name. Since we can use either, I did not make it an identifier (per my understanding the alias will not work as an identifier since it does not exist until the create stmt is executed, please correct me if I'm wrong). 
Should I change that field name to something like `tableNameOrAlias` instead to make it clearer to the user?

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: same as above, this can be either an alias or the qualified name, so identifier may not work

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: missed that, I will add it.

pkg/sdk/semantic_view_gen_test.go
sfc-gh-sarur: Using the SingleQuotes() method like this 
`ListAssignment("WITH SYNONYMS", "string", g.ParameterOptions().NoEquals().Parentheses().SingleQuotes())` 
did not work, as the synonyms were created like this `WITH SYNONYMS '(test1, test2)'` which is wrong. How can I make it add the quotes like `WITH SYNONYMS ('test1', 'test2')` instead?

pkg/sdk/semantic_view_gen_test.go
sfc-gh-sarur: The primary and unique keys should not be in quotes, here is an example from a working create stmt : 
```
CREATE OR REPLACE SEMANTIC VIEW tpcds_nlq_view
  TABLES (
       store as SNOWFLAKE_SAMPLE_DATA.TPCDS_SF1.store PRIMARY KEY (s_store_sk) UNIQUE (s_store_id) UNIQUE (s_store_name),
       store_sales as SNOWFLAKE_SAMPLE_DATA.TPCDS_SF1.store_sales PRIMARY KEY (ss_item_sk, ss_ticket_number),
       web_sales as SNOWFLAKE_SAMPLE_DATA.TPCDS_SF1.web_sales PRIMARY KEY (ws_item_sk, ws_order_number)
)
```

Should they be Identifiers instead of strings then?

pkg/sdk/semantic_view_impl_gen.go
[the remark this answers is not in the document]
sfc-gh-sarur: agreed.

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: Sounds good.

pkg/sdk/semantic_view_impl_gen.go
```diff
@@ -0,0 +1,227 @@
...
+		StartsWith: r.StartsWith,
+		Limit:      r.Limit,
+	}
+	return opts
+}
+
+func (r semanticViewDBRow) convert() *SemanticView {
+	// TODO: Mapping
+	return &SemanticView{}
+}
```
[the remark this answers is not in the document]
sfc-gh-sarur: this got undone when I ran make run and I did not notice, I have added the mappings

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,182 @@
...
+	SQL("ORDER BY").
+	OptionalText("OrderByClause", g.KeywordOptions()).
+	OptionalText("WindowFrameClause", g.KeywordOptions())
+
+var windowFunctionMetricDefinition = g.NewQueryStruct("WindowFunctionMetricDefinition").
+	Text("WindowFunction", g.KeywordOptions().Required()).
+	SQL("AS").
+	Text("Metric", g.KeywordOptions().Required()).
+	SQL("OVER").
+	OptionalQueryStructField("OverClause", windowFunctionOverClause, g.KeywordOptions().Parentheses())
```
sfc-gh-sarur: The Parentheses() method for KeywordOptions() is not adding the parentheses around the OverClause as expected. Is it supposed to work this way? Or should I be using a different transformer type for such a field?

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: Done, but ran into an issue with adding parentheses, please see : https://github.com/snowflakedb/terraform-provider-snowflake/pull/3925/files#r2286983341

pkg/sdk/semantic_view_def.go
sfc-gh-jmichalak: We use `OptionalLimit ` for objects that support only `LIMIT` sql clause (like https://docs.snowflake.com/en/sql-reference/sql/show-authentication-policies). In this case, `FROM` is also supported, so we should use `OptionalLimitFrom`.

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,182 @@
...
+).AlterOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/alter-semantic-view",
+	g.NewQueryStruct("AlterSemanticView").
+		Alter().
+		SQL("SEMANTIC VIEW").
+		IfExists().
+		Name().
+		OptionalTextAssignment("SET COMMENT", g.ParameterOptions().SingleQuotes()).
+		OptionalSQL("UNSET COMMENT").
+		WithValidation(g.ValidIdentifier, "name"),
```
sfc-gh-jmichalak: Pls also add `WithValidation(g.ExactlyOneValueSet, "SetComment", "UnsetComment"),`. Can be done in the follow-ups.

pkg/sdk/semantic_view_def.go
sfc-gh-jmichalak: So, in https://docs.snowflake.com/en/sql-reference/sql/create-semantic-view, instead of 
`  [ METRICS ( semanticExpression [ , ... ] ) ]` should be `  [ METRICS ( windowFunctionMetricDefinition [ , ... ] ) ]`, right?

pkg/sdk/semantic_view_gen_test.go
```diff
@@ -0,0 +1,315 @@
...
+			{
+				windowFunctionMetricDefinition: &WindowFunctionMetricDefinition{
+					WindowFunction: "metric1",
+					as:             true,
+					Metric:         "SUM(table_1.metric_1)",
+					over:           true,
+					OverClause: &WindowFunctionOverClause{
+						partitionBy:       true,
+						PartitionByClause: String("table_1.dimension_2, table_1.dimension_3"),
+						orderBy:           true,
```
sfc-gh-jmichalak: Nit: there is no need to set these keywords to `true`, as they should be in the query anyway.

pkg/sdk/semantic_view_gen_test.go
```diff
@@ -0,0 +1,315 @@
...
+			{
+				semanticExpression: &SemanticExpression{
+					qualifiedExpressionName: &QualifiedExpressionName{QualifiedExpressionName: metricName},
+					sqlExpression:           &SemanticSqlExpression{SqlExpression: metricExpression},
+					synonyms:                &Synonyms{WithSynonyms: []string{"'test5'", "'test6'"}},
+					Comment:                 String("metric_comment"),
+				},
+			},
+			{
+				windowFunctionMetricDefinition: &WindowFunctionMetricDefinition{
```
sfc-gh-jmichalak, line 207: Nit: in `def` we already have `WithValidation(g.ExactlyOneValueSet, "semanticExpression", "windowFunctionMetricDefinition")`, so we should test that:
- Setting both should fail.
- Setting one of the options is okay - for now, we only test windowFunctionMetricDefinition through `metricsObj[1]`, we can e.g. leave only windowFunctionMetricDefinition here and make a short separate unit test for semanticExpression.

pkg/sdk/semantic_view_def.go
```diff
@@ -0,0 +1,182 @@
...
+	OptionalQueryStructField("sqlExpression", semanticSqlExpression, g.KeywordOptions().Required()).
+	OptionalQueryStructField("synonyms", synonym, g.ParameterOptions().NoEquals()).
+	OptionalComment()
+
+var windowFunctionOverClause = g.NewQueryStruct("WindowFunctionOverClause").
+	SQL("PARTITION BY").
+	OptionalText("PartitionByClause", g.KeywordOptions()).
+	SQL("ORDER BY").
+	OptionalText("OrderByClause", g.KeywordOptions()).
+	OptionalText("WindowFrameClause", g.KeywordOptions())
```
sfc-gh-jmichalak, line 176: This part is missing: `[ ASC | DESC ] [ NULLS { FIRST | LAST } ] `, can be added later.

pkg/sdk/semantic_view_def.go
sfc-gh-jmichalak: I believe in `external_volumes.go` we have a similar case. We defined a helper struct:
```go
	OptionalQueryStructField(
		"Encryption",
		g.NewQueryStruct("ExternalVolumeGCSEncryption").
			Assignment("TYPE", g.KindOfT[GCSEncryptionType](), g.ParameterOptions().SingleQuotes().Required()).
			OptionalTextAssignment("KMS_KEY_ID", g.ParameterOptions().SingleQuotes()),
		g.ListOptions().Parentheses().NoComma().SQL("ENCRYPTION ="),
	)
```

and used it below in
```go
			OptionalQueryStructField(
				"AddStorageLocation",
				storageLocationDef,
				g.ParameterOptions().SQL("ADD STORAGE_LOCATION"),
			).
```

I think we should just use `g.ListOptions().Parentheses().NoComma().SQL("OVER")` instead of the current `g.KeywordOptions` and remove `	SQL("OVER").` above.

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: No, metrics can be created using either the semantic expression format or the window function format. That's why I created a new struct for metrics with one member for each type

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: I held off on adding this as I would need to make another struct for WindowFrameClause and I thought the additional nesting was not worth it, since we can validate this separately. Is it worth branching out further to add this?

pkg/sdk/semantic_view_gen_test.go
sfc-gh-sarur: metricObjs[0] is testing the semanticExpression format as well. I will add a failure case for when both are present.

pkg/sdk/semantic_view_gen_test.go
sfc-gh-sarur: I made a mistake in the def file, these should be OptionalSQL and only added if the associated clause is present. Since the .SQL() method for KeywordOptions() does not seem to be working, I had to add them separately

pkg/sdk/semantic_view_gen_test.go
sfc-gh-sarur: done.

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: Please do, the `tableNameOrAlias` should contain one field that is just an `OptionalText` and the other one `OptionalIdentifier` of SchemaObjectIdentifier.

FYI: alias could be an identifier, but of type AccountObjectIdentifier, which contains only one part (an example would be database identifier "DATABASE_NAME"). For SchemaObjectIdentifier we expect three parts "database"."schema"."table" which wouldn't make sense for an alias (I'm guessing).

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: Please proceed with `tableNameOrAlias` approach.

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: I don't think there's a better way than just using a nested object with one `Text` required value that is annotated with `SingleQuotes()`. Here's an example: https://github.com/snowflakedb/terraform-provider-snowflake/blob/e1cffd331b8f27e104c35c3403a3765c1327aa62/pkg/sdk/notification_integrations_def.go#L82

Snowflake doc reference: https://docs.snowflake.com/en/sql-reference/sql/create-notification-integration-email

pkg/sdk/semantic_view_gen_test.go
sfc-gh-jcieslak: Please, follow the steps in: https://github.com/snowflakedb/terraform-provider-snowflake/pull/3925#discussion_r2293078690

pkg/sdk/semantic_view_gen_test.go
sfc-gh-jcieslak: Are they case-sensitive? If yes, they should most likely be quoted with double quotes; if not, they can stay as they are right now.

If they have to be quoted, you can proceed with the same approach mentioned in: https://github.com/snowflakedb/terraform-provider-snowflake/pull/3925#discussion_r2293078690 where you can have SemanticViewPrimaryKey struct with one required `Text` field and use it in the ListAssignment instead of regular string.

pkg/sdk/semantic_view_gen_test.go
```diff
@@ -35,6 +35,37 @@ func TestSemanticViews_Create(t *testing.T) {
 		assertOptsInvalidJoinedErrors(t, opts, errOneOf("CreateSemanticViewOptions", "IfNotExists", "OrReplace"))
 	})
 
+	t.Run("validation: exactly one field for [metricDefinition.semanticExpression metricDefinition.windowFunctionMetricDefinition]", func(t *testing.T) {
+		metricsObj := []MetricDefinition{
+			{
+				semanticExpression: &SemanticExpression{
+					qualifiedExpressionName: &QualifiedExpressionName{QualifiedExpressionName: "metricName"},
```
sfc-gh-jcieslak: nit(doesn't have to be changed): we don't need to fill SemanticExpression and WindowFunctionMetricDefinition to observe errExactlyOneOf error

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: Done

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: Done

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: Done.

sfc-gh-sarur (commented):
addressed code review comments

pkg/sdk/semantic_view_impl_gen.go
```diff
@@ -0,0 +1,264 @@
...
+
+func (r semanticViewDetailsRow) convert() *SemanticViewDetails {
+	return &SemanticViewDetails{
+		ObjectKind:    r.ObjectKind,
+		ObjectName:    r.ObjectName,
+		ParentEntity:  r.ParentEntity,
+		Property:      r.Property,
+		PropertyValue: r.PropertyValue,
+	}
```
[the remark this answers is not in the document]
sfc-gh-sarur: @sfc-gh-jcieslak @sfc-gh-jmichalak looks like these method signatures have been updated on the dev branch. I did not rebase to dev as we discussed, so the code is working locally. But IIUC, these pipelines will continue to fail and merging will be blocked until I rebase.

pkg/sdk/semantic_view_def.go
sfc-gh-jcieslak: Done

PR #3926, opened 2025-08-11
fix: Change convert signature for a single object

sfc-gh-asawicki:
It's the first PR of a few aiming to improve SDK's stability and reliability by including an error in the return from conversion attempt. Till now, the errors there were only logged which caused unexpected behavior in some cases (like improper id parsing for functions and , because of that, not finding the function with an empty id).

The change contains:
- Rename `convertibleRow` to `convertibleRowDeprecated`
- Introduce `conversionErrorWrapped` method (it's needed as `convert` is either used from `convertRows` or directly)
- Alter the conversion for warehouse
- Add unit tests for warehouse conversion

Next PRs:
- Migrate all other objects that are not generated from defs
- Alter generation
- Migrate all other objects that are generated
- Rename `convertRowsErr` to `convertRows` when all objects are migrated
- Rename `convertErr` to `convert` when all objects are migrated
- Remove `convertibleRowDeprecated` when all objects are migrated

Future improvements:
- Add compile checks like `var _ convertibleRowDeprecated[DatabaseRole] = new(databaseRoleDBRow)` for each convertible (generate it)

Questions:
- Do we want to have extensive unit tests for each conversion?
- Do we want to generate these tests?

### DIFF
--- a/pkg/sdk/helpers_proposal.go
+++ b/pkg/sdk/helpers_proposal.go
@@ -80,7 +80,7 @@ func convertRows[T convertibleRowDeprecated[U], U any](dbRows []T) []U {
 }
 
 type convertibleRow[T any] interface {
-	// TODO [SNOW-2259477]: rename to convertErr
+	// TODO [SNOW-2259477]: rename to convert
 	convertErr() (*T, error)
 }
 
--- a/pkg/sdk/helpers_proposal.go
+++ b/pkg/sdk/helpers_proposal.go
@@ -66,11 +66,11 @@ func createIfNil[T any](t *T) *T {
 	return t
 }
 
-type convertibleRow[T any] interface {
+type convertibleRowDeprecated[T any] interface {
 	convert() *T
 }
 
-func convertRows[T convertibleRow[U], U any](dbRows []T) []U {
+func convertRows[T convertibleRowDeprecated[U], U any](dbRows []T) []U {
 	resultList := make([]U, len(dbRows))
 	for i, row := range dbRows {
 		resultList[i] = *(row.convert())
--- a/pkg/sdk/helpers_proposal.go
+++ b/pkg/sdk/helpers_proposal.go
@@ -78,6 +78,22 @@ func convertRows[T convertibleRowDeprecated[U], U any](dbRows []T) []U {
 	return resultList
 }
 
+type convertibleRow[T any] interface {
+	convertErr() (*T, error)
+}
+
+func convertRowsErr[T convertibleRow[U], U any](dbRows []T) ([]U, error) {
+	resultList := make([]U, len(dbRows))
+	for i, row := range dbRows {
+		converted, err := row.convertErr()
+		if err != nil {
+			return nil, err
+		}
+		resultList[i] = *converted
+	}
+	return resultList, nil
+}
+
 type optionsProvider[T any] interface {
 	toOpts() *T
 }
--- a/pkg/sdk/helpers_proposal.go
+++ b/pkg/sdk/helpers_proposal.go
@@ -80,9 +80,11 @@ func convertRows[T convertibleRowDeprecated[U], U any](dbRows []T) []U {
 }
 
 type convertibleRow[T any] interface {
+	// TODO [SNOW-2259477]: rename to convertErr
 	convertErr() (*T, error)
 }
 
+// TODO [SNOW-2259477]: rename to convertRows
 func convertRowsErr[T convertibleRow[U], U any](dbRows []T) ([]U, error) {
 	resultList := make([]U, len(dbRows))
 	for i, row := range dbRows {
--- a/pkg/sdk/helpers_proposal.go
+++ b/pkg/sdk/helpers_proposal.go
@@ -2,6 +2,7 @@ package sdk
 
 import (
 	"context"
+	"fmt"
 )
 
 // validatable is our sdk interface for anything that can be validated (e.g. CreateXxxOptions).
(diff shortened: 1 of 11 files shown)


pkg/sdk/helpers_proposal.go
```diff
@@ -66,18 +67,44 @@ func createIfNil[T any](t *T) *T {
...
+func convertRows[T convertibleRowDeprecated[U], U any](dbRows []T) []U {
 	resultList := make([]U, len(dbRows))
 	for i, row := range dbRows {
 		resultList[i] = *(row.convert())
 	}
 	return resultList
 }
 
+type convertibleRow[T any] interface {
+	// TODO [SNOW-2259477]: rename to convertErr
```
sfc-gh-jcieslak: nit: already done? I'm guessing it should be rename to `convert`?

pkg/sdk/helpers_proposal.go
```diff
@@ -66,18 +67,44 @@ func createIfNil[T any](t *T) *T {
...
+		converted, err := conversionErrorWrapped(row.convertErr())
+		if err != nil {
+			return nil, err
+		}
+		resultList[i] = *converted
+	}
+	return resultList, nil
+}
+
+func conversionErrorWrapped[U any](converted *U, err error) (*U, error) {
```
sfc-gh-jcieslak, line 100: nit: we could wrap one level above and remove this function

pkg/sdk/warehouses.go
```diff
@@ -529,17 +526,38 @@ func (row warehouseDBRow) convert() *Warehouse {
 		QueryAccelerationMaxScaleFactor: row.QueryAccelerationMaxScaleFactor,
 		ScalingPolicy:                   ScalingPolicy(row.ScalingPolicy),
 	}
-	if val, err := strconv.ParseFloat(row.Available, 64); err != nil {
-		wh.Available = val
+	if size, err := ToWarehouseSize(row.Size); err != nil {
```
sfc-gh-jcieslak, line 529: nit: if we are touching those converters, maybe it would be worth using (adding new) mapping_helpers?

pkg/sdk/warehouses.go
```diff
@@ -559,8 +577,8 @@ func (c *warehouses) Show(ctx context.Context, opts *ShowWarehouseOptions) ([]Wa
 	if err != nil {
 		return nil, err
 	}
-	resultList := convertRows[warehouseDBRow, Warehouse](dbRows)
-	return resultList, nil
+	resultList, err := convertRowsErr[warehouseDBRow, Warehouse](dbRows)
+	return resultList, err
```
sfc-gh-jcieslak: nit: do we want to return nil []Warehouse or whatever covertRowsErr managed to create so far?

sfc-gh-jcieslak (dismissed):

pkg/sdk/warehouses_test.go
```diff
@@ -412,3 +414,54 @@ func Test_Warehouse_ToScalingPolicy(t *testing.T) {
...
 }
+
+func Test_Warehouse_Convert(t *testing.T) {
+	correctRow := func() warehouseDBRow {
+		return warehouseDBRow{
+			Size:      string(WarehouseSizeXSmall),
+			Available: "100",
+		}
+	}
+
```
sfc-gh-jcieslak, line 425: To answer the questions in the PR description.

> Do we want to have extensive unit tests for each conversion?

I don't think so. They seem like a nice addition to separate them from other test types, but on the other hand, most errors would be spotable on the integration test level.

> Do we want to generate these tests?

As above, I wasn't really sure about adding those tests by hand. It would be a nice addition if we could generate those. Unit tests are so quick we shouldn't feel any difference in test workflow times.

pkg/sdk/helpers_proposal.go
sfc-gh-asawicki: No, check the PR description:
Introduce conversionErrorWrapped method (it's needed as convert is either used from convertRows or directly)

pkg/sdk/helpers_proposal.go
sfc-gh-asawicki: Yeah, typo, changed

pkg/sdk/warehouses.go
sfc-gh-asawicki: Yup, we can add some, to later use when convert body is generated

pkg/sdk/warehouses.go
sfc-gh-asawicki: Changed to return covertRowsErr directly

pkg/sdk/warehouses_test.go
sfc-gh-asawicki: Not every error will be visible in integration tests. The ones for basic flow and values—yes. But we don't test all the variants on the integration level (and we shouldn't). This is a good place to verify the conversion and errors.

For now, I think it's ok to add sanity checks for each obj, and we can generate them with the convert func generation, wdyt?

sfc-gh-jcieslak: Sure

pkg/sdk/warehouses.go
```diff
@@ -529,17 +526,38 @@ func (row warehouseDBRow) convert() *Warehouse {
...
 		ScalingPolicy:                   ScalingPolicy(row.ScalingPolicy),
 	}
-	if val, err := strconv.ParseFloat(row.Available, 64); err != nil {
-		wh.Available = val
+	if size, err := ToWarehouseSize(row.Size); err != nil {
+		return nil, err
+	} else {
+		wh.Size = size
+	}
+	if available := strings.TrimSpace(row.Available); available != "" {
```
sfc-gh-jmichalak, line 534: Just an idea: we could trim spaces on the deserialization level, with custom `db` tags.

PR #3932, opened 2025-08-13
chore: Find versions of provider in gh organization

sfc-gh-asawicki:
Add script to check the Snowflake Terraform Provider versions in the given GH organization

### DIFF
--- a/pkg/scripts/provider_versions_in_organization/main.go
+++ b/pkg/scripts/provider_versions_in_organization/main.go
@@ -159,9 +159,10 @@ func transformToResult(registry string, resultItem searchResultItem) []result {
 			openIdx := max(strings.LastIndex(m.Fragment[:st], "{"), 0)
 			closeIdx := strings.Index(m.Fragment[end:], "}")
 			if closeIdx == -1 {
-				closeIdx = end
+				frag = m.Fragment[openIdx:]
+			} else {
+				frag = m.Fragment[openIdx : end+closeIdx+1]
 			}
-			frag = m.Fragment[openIdx : end+closeIdx+1]
 			versionRegex := regexp.MustCompile(`version\s+=\s+"(.*)"`)
 			vMatches := versionRegex.FindStringSubmatch(frag)
 			if len(vMatches) == 0 {


pkg/scripts/provider_versions_in_organization/main.go
```diff
@@ -0,0 +1,214 @@
...
+	phraseUrl := fmt.Sprintf("%s%s?q=%s", githubAPIBase, searchEndpoint, queryEscaped)
+
+	allResults := &SearchResult{Items: []searchResultItem{}}
+	page := 1
+	for {
+		results, err := ghSearch(accessToken, phraseUrl, page)
+		if err != nil {
+			return nil, err
+		}
+		if len(results.Items) == 0 {
```
sfc-gh-jcieslak, line 109: nit: I guess this could end faster with `len(results.Items) != perPage`

pkg/scripts/provider_versions_in_organization/main.go
```diff
@@ -0,0 +1,214 @@
...
+	w := csv.NewWriter(file)
+	w.Comma = ';'
+	data := make([][]string, 0, len(results))
+	data = append(data, []string{"Registry", "Estimated Version", "RepoURL", "FileURL", "Fragment"})
+	for _, r := range results {
+		if strings.HasPrefix(r.Version, "=") {
+			r.Version = "'" + r.Version
+		}
+		if strings.HasPrefix(r.Fragment, "=") {
+			r.Fragment = "'" + r.Fragment
```
sfc-gh-jcieslak, line 202: question: what is this for?

sfc-gh-jcieslak (approved):

pkg/scripts/provider_versions_in_organization/main.go
sfc-gh-asawicki: for excel :D (it thinks that field starting with = is a formula so this is the way to make it a string instead, the quote is not visible then: https://www.thebricks.com/resources/guide-how-to-put-equal-sign-in-google-sheets-without-formula)

sfc-gh-jcieslak: 😆 oh okay

PR #3937, opened 2025-08-14
fix: Enhance connection management and logging for secondary connections

sfc-gh-jcieslak:
## Changes
- Fix connection read functionality
- Create a dedicated delete function for the secondary connection due to Snowflake change propagation (it's mostly needed for the secondary connection, and in specific cases)
- Adjusted existing tests and test client

## Notes
- There are some issues with CheckDestroy, potentially due to Snowflake changes propagation. Will create a ticket for it.


pkg/resources/manual_tests/secondary_connection_promotion/step_1.tf
```diff
@@ -0,0 +1,34 @@
...
+provider "snowflake" {}
+
+provider "snowflake" {
+  profile = "secondary_test_account"
+  alias   = second_account
+}
+
+resource "snowflake_primary_connection" "primary_connection" {
+  name = "TEST_CONNECTION"
+  enable_failover_to_accounts = ["<organization_name>.<seconary_connection_account_name>"]
```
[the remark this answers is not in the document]
sfc-gh-asawicki: +1

pkg/resources/secondary_connection.go
```diff
@@ -184,3 +180,33 @@ func UpdateContextSecondaryConnection(ctx context.Context, d *schema.ResourceDat
...
+	// 2. Setting secondary connection (con2) as primary
+	// 3. Setting previously primary connection (con1) as primary (we will have the same hierarchy as in the first step)
+	// 4. Deleting secondary connection (con2; this may fail without waiting a bit for Snowflake to propagate the changes)
+	if err := util.Retry(3, time.Second, func() (error, bool) {
+		if err := client.Connections.DropSafely(ctx, id); err != nil && strings.Contains(err.Error(), "is currently a primary connection in a replication relationship and cannot be dropped") {
+			return nil, false
+		} else {
+			log.Printf("[DEBUG] Drop secondary connection failed, err = %s", err)
+			return err, true
+		}
```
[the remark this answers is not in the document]
sfc-gh-asawicki: I dunno, pls check

docs/resources/primary_connection.md
```diff
@@ -29,6 +29,8 @@ resource "snowflake_primary_connection" "complete" {
 
 -> **Note** Instead of using fully_qualified_name, you can reference objects managed outside Terraform by constructing a correct ID, consult [identifiers guide](../guides/identifiers_rework_design_decisions#new-computed-fully-qualified-name-field-in-resources).
 
+-> **Note** This resource cannot be dropped when it has any dependent secondary connections. If you want to drop the primary connection, you must first drop all secondary connections that depend on it or promote other connection to be primary. The second option may require removing the resource from the state and removing it manually from Snowflake.
```
sfc-gh-asawicki: what if there are proper references/depends_on set? Is the destruction happening in the reverse order? Or does it need to be done in two steps?

pkg/testacc/resource_secondary_connection_acceptance_test.go
```diff
@@ -12,13 +15,16 @@ import (
...
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+	"github.com/stretchr/testify/assert"
 )
 
+// Recreation when promoting secondary to primary cannot be tested, because of the Terraform testing framework limitations.
```
sfc-gh-asawicki, line 25: Do we have these limitations described somewhere? Im curious what they are

pkg/resources/manual_tests/secondary_connection_promotion/step_1.tf
```diff
@@ -0,0 +1,34 @@
+# Prerequisites:
+# - Ensure both accounts were primary and secondary connections will be created have:
```
sfc-gh-jmichalak:
```suggestion
# - Ensure both accounts where primary and secondary connections will be created have:
```

pkg/resources/secondary_connection.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: I don't think the bot is quite correct here, but currently, when we get no error, this still does not get handled properly.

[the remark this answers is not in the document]
sfc-gh-jcieslak: Not correct, maybe it thought we are using the retry method from some other library, which behaves this way.

docs/resources/primary_connection.md
sfc-gh-jcieslak: Updated the docs. In the primary connection, we have to worry about when removing with secondary connections and in secondary connections we have to worry about creating with primary connection. Ultimately, it should be just better handled on our end (wait for them to appear). I'll create a ticket for it. We should also have a bit smarter retry function to help us in such situations, current one is pretty limited.

pkg/testacc/resource_secondary_connection_acceptance_test.go
sfc-gh-jcieslak: I don't remember; I think we don't. As far as I remember, it was connected to the deletion part. You cannot easily remove something from a state or import a resource to it during the test. Those operations are required to fix broken states. Without it, I wasn't able to show that the old version wasn't correct, because:
- In those cases, ExpectError didn't work or
- Delete didn't work, thus DeleteCheck failed
- and I think other issues, but I don't remember all of them

pkg/resources/manual_tests/secondary_connection_promotion/step_1.tf
sfc-gh-jcieslak: changed

pkg/resources/secondary_connection.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: As we discussed offline, when there is no error from `DropSafely` (err is nil), the program won't enter the first block due to `err != nil` check, but it will enter the second one and log the `nil` error. The returned values will still be correct, the problem now is just the incorrect log entry. Can be fixed later if you want.

sfc-gh-jmichalak (dismissed):

sfc-gh-jmichalak (approved):

pkg/testacc/resource_secondary_connection_acceptance_test.go
sfc-gh-asawicki: Let's note them down internally for easier reference.

PR #3940, opened 2025-08-14
chore: Change convert signature in all other resources

sfc-gh-asawicki:
Continuation of #3926.

It's the last PR aiming to improve SDK's stability and reliability by including an error in the return from conversion attempt. Till now, the errors there were only logged which caused unexpected behavior in some cases (like improper id parsing for functions and , because of that, not finding the function with an empty id).

The change contains:

- Migrate all other objects that are not generated from defs
- Alter generation
- Migrate all other objects that are generated
- Rename convertRowsErr to convertRows when all objects are migrated
- Rename convertErr to convert when all objects are migrated
- Remove convertibleRowDeprecated when all objects are migrated
- Add compile checks like var _ convertibleRowDeprecated[DatabaseRole] = new(databaseRoleDBRow) for each convertible (generated)

Future improvements:
- generated conversion tests (or at least sanity checks)
- the generated guards are not currently wrapped with other vars during the generation and there are whitespace issues

### DIFF
--- a/pkg/sdk/sql_mapping_helpers.go
+++ b/pkg/sdk/sql_mapping_helpers.go
@@ -2,18 +2,18 @@ package sdk
 
 import (
 	"database/sql"
-	"log"
+	"fmt"
 	"strconv"
 )
 
-func handleNullableBoolString(nullableBoolString sql.NullString, field *bool) {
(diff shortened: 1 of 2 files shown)


sfc-gh-jmichalak (approved):
LGTM, but the migration guide is missing.

PR #3949, opened 2025-08-20
chore: Update failover group docs

sfc-gh-jmichalak:
Update failover group docs.

### DIFF
--- a/docs/resources/failover_group.md
+++ b/docs/resources/failover_group.md
@@ -7,6 +7,8 @@ description: |-
 
 !> **Caution: Preview Feature** This feature is considered a preview feature in the provider, regardless of the state of the resource in Snowflake. We do not guarantee its stability. It will be reworked and marked as a stable feature in future releases. Breaking changes are expected, even without bumping the major version. To use this feature, add the relevant feature name to `preview_features_enabled` field in the [provider configuration](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs#schema). Please always refer to the [Getting Help](https://github.com/snowflakedb/terraform-provider-snowflake?tab=readme-ov-file#getting-help) section in our Github repo to best determine how to get help for your questions.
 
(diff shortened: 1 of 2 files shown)


docs/resources/failover_group.md
```diff
@@ -7,6 +7,8 @@ description: |-
...
 !> **Caution: Preview Feature** This feature is considered a preview feature in the provider, regardless of the state of the resource in Snowflake. We do not guarantee its stability. It will be reworked and marked as a stable feature in future releases. Breaking changes are expected, even without bumping the major version. To use this feature, add the relevant feature name to `preview_features_enabled` field in the [provider configuration](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs#schema). Please always refer to the [Getting Help](https://github.com/snowflakedb/terraform-provider-snowflake?tab=readme-ov-file#getting-help) section in our Github repo to best determine how to get help for your questions.
 
+!> **Updating `allowed_accounts`** Currently, updating the `allowed_accounts` field may fail due to an incorrect query being sent (see [#3946](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3946)). This will be fixed during the resource rework. As a workaround, use the `execute` resource to update the allowed accounts manually. After that, refresh the state with the updated `allowed_accounts` field in the resource configuration.
```
sfc-gh-jcieslak, line 10: question: wouldn't it produce infinite plan on allowed_accounts in the failover_group resource? If the users would like to update the object with execute resource, they should also use the ignore_changes lifecycle attribute, right?

sfc-gh-jcieslak (approved):

docs/resources/failover_group.md
sfc-gh-jmichalak: As we discussed offline, this will not cause an infinite plan:
- A user can add the correct account outside TF (this change will be read by the provider and stored in the state.
- The config will be updated (added the correct account to the list), so TF shouldn't detect any changes.
- This solution is okay because the only error we have is in the update function, so in this scenario the update won't be called because there will be no diff.

sfc-gh-jmichalak:
### Merge activity

* **Aug 21, 12:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3949).
* **Aug 21, 12:42 PM UTC**: @sfc-gh-jmichalak merged this pull request with [Graphite](https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/3949).

PR #3950, opened 2025-08-20
chore: Sweep security integrations

sfc-gh-jcieslak:
## Changes


pkg/sdk/sweepers_test.go
```diff
@@ -317,3 +326,62 @@ func nukeUsers(client *Client, suffix string) func() error {
...
+				return u.CreatedOn.Before(time.Now().Add(-15 * time.Minute))
+			}
+		}
+		var integrations []Integration
+		if err := client.QueryForTests(ctx, &integrations, `SHOW INTEGRATIONS`); err != nil {
+			return fmt.Errorf("SHOW INTEGRATIONS ended with error, err = %w", err)
+		}
+
+		log.Printf("[DEBUG] Found %d integrations", len(integrations))
+		_ = protectedIntegrations
```
sfc-gh-jcieslak: will be removed

pkg/sdk/sweepers_test.go
```diff
@@ -317,3 +326,62 @@ func nukeUsers(client *Client, suffix string) func() error {
...
+
+		var errs []error
+		for idx, integration := range integrations {
+			log.Printf("[DEBUG] Processing integration [%d/%d]: %s...", idx+1, len(integrations), integration.Name)
+
+			if !slices.Contains(protectedIntegrations, integration.Name) && integrationDropCondition(integration) {
+				log.Printf("[DEBUG] Dropping integration %s", integration.Name)
+				// Commented out intentionally, uncomment after review
+
+				// if _, err := client.ExecForTests(ctx, fmt.Sprintf(`DROP INTEGRATION "%s"`, integration.Name)); err != nil {
```
sfc-gh-jcieslak: will be uncommented after review

pkg/sdk/sweepers_test.go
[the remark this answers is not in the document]
sfc-gh-jmichalak: Good catch.

pkg/sdk/sweepers_test.go
sfc-gh-jmichalak: Q: Do we have any plans to have a generic integration client?

sfc-gh-jmichalak (approved):

pkg/sdk/sweepers_test.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: ehh, I submitted this comment myself, but didn't actually click "comment" 🤦 it was still in "pending" mode

pkg/sdk/sweepers_test.go
sfc-gh-jcieslak: I don't think we have such a ticket, but I can create it as it's useful in those situations

pkg/sdk/sweepers_test.go
```diff
@@ -317,3 +326,62 @@ func nukeUsers(client *Client, suffix string) func() error {
...
+				return strings.HasSuffix(u.Name, suffix)
+			}
+		} else {
+			log.Println("[DEBUG] Sweeping stale integrations")
+			integrationDropCondition = func(u Integration) bool {
+				return u.CreatedOn.Before(time.Now().Add(-15 * time.Minute))
+			}
+		}
+		var integrations []Integration
+		if err := client.QueryForTests(ctx, &integrations, `SHOW INTEGRATIONS`); err != nil {
```
sfc-gh-asawicki: I am unsure if we should blindly address all the integration types here. Don't we have any other integrations that should be protected (the listed ones are only the storage integrations)?

pkg/sdk/sweepers_test.go
sfc-gh-asawicki: when?

pkg/sdk/sweepers_test.go
sfc-gh-asawicki: For now, we are not even using the testclient in the sweepers, so such addition wouldn't be helpful now as sweepers need their own rework.

pkg/sdk/sweepers_test.go
sfc-gh-jcieslak: I looked through the existing integrations on the testing environments, and only storage integrations appeared to be pre-created. To avoid a too broad scope, we can focus specifically on a given integration type.

pkg/sdk/sweepers_test.go
sfc-gh-jcieslak: now, removed

pkg/sdk/sweepers_test.go
sfc-gh-jcieslak: What do you think about this change? We were mostly struggling with security integrations, so for now, let's repurpose this sweeper to strictly focus on them. We don't have any predefined security integrations. There's one `keycloak_saml`, but afaik it was created by Scott to test other auth types, and it can be removed.

pkg/sdk/sweepers_test.go
sfc-gh-jcieslak: repurposed the sweeper to focus only on security integrations anyway

sfc-gh-jmichalak (approved):

pkg/sdk/sweepers_test.go
sfc-gh-asawicki: LGTM

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #3953, opened 2025-08-21
fix: Fix updating privileges in grant_privileges_to_account_role

sfc-gh-jmichalak:
Fix updating privileges in grant_privileges_to_account_role.

- Skip revoking privileges when `IMPORTED PRIVILEGES` is set.
- Update resource documentation.
- Increase the timeout for `CreateDatabaseFromShare` condition.
- Add acceptance tests proving the issue in older versions and the fix in the latest versions.
- Move config files to inlined configurations.

## References

https://github.com/snowflakedb/terraform-provider-snowflake/issues/2803#issuecomment-3152992005


docs/resources/grant_privileges_to_account_role.md
```diff
@@ -12,6 +12,8 @@ description: |-
 
 ~> **Note** When granting privileges on applications (for example, the default "SNOWFLAKE" application) use `on_account_object.object_type = "DATABASE"` instead.
 
+~> **Note** When using `IMPORTED PRIVILEGES` privilege, the `with_grant_option` field is not supported. Additionally, when the `IMPORTED PRIVILEGES` privilege is not set in the config, and it is granted externally, this change is not detected because of Snowflake limitations.
```
sfc-gh-jcieslak: Additionally, let's mention the imported privileges are not available with other privileges, other privileges has to be defined in another grant_priv_to_acc_role resource

sfc-gh-jcieslak (approved):

docs/resources/grant_privileges_to_account_role.md
sfc-gh-asawicki: question: are this restrictions applied only on Snowflake or do we also apply them on schema/logic level?
Usually, we try not to be coupled with SNowflake validations logic, however, in this case, we have a crafted logic for imported pirivileges and grant option, so in addition to just adding it to the documentation we could validate it, wdyt?

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -1301,29 +1303,20 @@ func TestAcc_GrantPrivilegesToAccountRole_ImportedPrivileges(t *testing.T) {
 
 	externalShareId := createSharedDatabaseOnSecondaryAccount(t)
 
-	databaseFromShare, databaseFromShareCleanup := testClient().Database.CreateDatabaseFromShare(t, externalShareId)
+	databaseFromShare, databaseFromShareCleanup := testClient().Database.CreateDatabaseFromShareSkipWaitingForOrigin(t, externalShareId)
```
sfc-gh-asawicki: question: should we care about this await in the production resource or we just assume here that worst case scenario, terraform plan would be re-run (I'm fine with the latter).

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -1335,16 +1328,85 @@ func TestAcc_GrantPrivilegesToAccountRole_ImportedPrivileges(t *testing.T) {
 				),
 			},
 			{
-				ConfigDirectory:   ConfigurationDirectory("TestAcc_GrantPrivilegesToAccountRole/ImportedPrivileges"),
-				ConfigVariables:   configVariables,
-				ResourceName:      resourceName,
-				ImportState:       true,
-				ImportStateVerify: true,
+				ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
```
sfc-gh-asawicki: nit: why ProtoV6ProviderFactories were moved inside the steps? this is not the migration test, so they can stay at the top-level

pkg/acceptance/helpers/database_client.go
```diff
@@ -221,7 +222,20 @@ func (c *DatabaseClient) CreateDatabaseFromShare(t *testing.T, externalShareId s
 		// Origin is returned as "<revoked>" in those cases, because it's not valid sdk.ExternalObjectIdentifier parser sets it as nil.
 		// Once it turns into valid sdk.ExternalObjectIdentifier, we're ready to proceed with the actual test.
 		return database.Origin != nil
-	}, time.Minute, time.Second*6)
+	}, 3*time.Minute, time.Second*6)
+
+	return database, c.DropDatabaseFunc(t, databaseId)
+}
+
+func (c *DatabaseClient) CreateDatabaseFromShareSkipWaitingForOrigin(t *testing.T, externalShareId sdk.ExternalObjectIdentifier) (*sdk.Database, func()) {
```
sfc-gh-asawicki: should we use it also in the second place it has been used until now? (tags association integration test) And maybe get rd of the non-waiting function at all?

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/acceptance/helpers/database_client.go
sfc-gh-jmichalak: I thought about this, I'm not sure if it will work consistently, but we can give it a try, done.

docs/resources/grant_privileges_to_account_role.md
sfc-gh-jmichalak: They can't be defined in another resource, as these grants are mutually exclusive. When you have `IMPORTED PRIVILEGES`, you can't have any other grant. This is enforced only in SF.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jmichalak: I think we did not get feedback about this, so we can leave it. I updated the `shared_database` resource documentation, though.

docs/resources/grant_privileges_to_account_role.md
sfc-gh-asawicki: Okay, so what about `SNowflake validations logic, however, in this case, we have a crafted logic for imported pirivileges and grant option, so in addition to just adding it to the documentation we could validate it, wdyt `? @sfc-gh-jcieslak @sfc-gh-jmichalak

sfc-gh-asawicki (dismissed):

docs/resources/grant_privileges_to_account_role.md
sfc-gh-jcieslak: Yeah, we could add a custom validate func to check if any other privilege is set with imported privileges. For me, it can be added in the next pr, but maybe don't postpone it too much.

sfc-gh-jmichalak: Done here.

sfc-gh-asawicki (approved):

sfc-gh-jcieslak (approved):

PR #3964, opened 2025-08-26
chore: Clean up TODO comments left with "next PR" without issue number

sfc-gh-asawicki:
Add issue numbers for each TODO left with "next PR" annotation. Removed some. Solved the one with the package move (file helpers).

### DIFF
--- a/pkg/testacc/resource_view_acceptance_test.go
+++ b/pkg/testacc/resource_view_acceptance_test.go
@@ -434,7 +434,7 @@ func TestAcc_View_recursive(t *testing.T) {
 	})
 }
 
-// TODO [next PR]: currently this test is always skipped, try to fix the set up
+// TODO [SNOW-2298268]: currently, this test is always skipped, try to fix the set up
 func TestAcc_View_temporary(t *testing.T) {
 	t.Setenv(string(testenvs.ConfigureClientOnce), "")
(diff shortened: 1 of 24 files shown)


pkg/testacc/resource_view_acceptance_test.go
```diff
@@ -434,7 +434,7 @@ func TestAcc_View_recursive(t *testing.T) {
 	})
 }
 
-// TODO [next PR]: currently this test is always skipped, try to fix the set up
+// TODO [SNOW-2298268]: currently, this test is always skipped, try to fix the set up
```
sfc-gh-jcieslak, line 437: Question: how is this test always skipped, it's because of the `_ = testenvs.GetOrSkipTest(t, testenvs.ConfigureClientOnce)`?

sfc-gh-jcieslak (approved):

pkg/testacc/resource_view_acceptance_test.go
sfc-gh-asawicki: yes, because these two lines:
```
	t.Setenv(string(testenvs.ConfigureClientOnce), "")
	_ = testenvs.GetOrSkipTest(t, testenvs.ConfigureClientOnce) -- if we set it to emptyt above, it will be always skipped
```

PR #3967, opened 2025-08-26
chore: Fix check destroy logic

sfc-gh-asawicki:
CheckDestroy was silently failing, as we expected an error from Snowflake when trying to fetch the objects as part of CheckDestroy. The problem was that using the old ID parsing logic, we were always getting AccountObjectIdentifier, and later casting was returning an error, but it was treated as a successful check for an object not existing in Snowflake.
Changes:
- checking casting error before succeeding
- logic split into legacy parsing and new id parsing
- cosmetic changes
- fixed listing and storage integration

### DIFF
--- a/pkg/testacc/2_check_destroy_test.go
+++ b/pkg/testacc/2_check_destroy_test.go
@@ -32,6 +32,17 @@ func ComposeCheckDestroy(t *testing.T, resources ...resources.Resource) func(*te
 
 func CheckDestroy(t *testing.T, resource resources.Resource) func(*terraform.State) error {
 	t.Helper()
+	return checkDestroy(t, resource, decodeSnowflakeId)
+}
+
+// CheckDestroyUsingLegacyIdParsing is meant to be used in tests checking older provider versions with the legacy identifier logic.
+func CheckDestroyUsingLegacyIdParsing(t *testing.T, resource resources.Resource) func(*terraform.State) error {
+	t.Helper()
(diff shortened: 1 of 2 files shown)


sfc-gh-jmichalak (approved):

pkg/testacc/2_check_destroy_test.go
```diff
@@ -66,28 +82,90 @@ func CheckDestroy(t *testing.T, resource resources.Resource) func(*terraform.Sta
...
+		return sdk.ParseObjectIdentifierString(rs.Primary.ID)
 	}
 }
 
-type showByIdFunc func(context.Context, *sdk.Client, sdk.ObjectIdentifier) error
+func decodeSnowflakeIdLegacy(rs *terraform.ResourceState, _ resources.Resource) (sdk.ObjectIdentifier, error) {
+	return helpers.DecodeSnowflakeIDLegacy(rs.Primary.ID), nil
+}
+
+type supportedIdentifierTypes interface {
```
sfc-gh-jcieslak, line 132: nit: As far as I remember, we have an ObjectIdentifierConstraint that could be reused here

sfc-gh-jcieslak (approved):

pkg/testacc/2_check_destroy_test.go
sfc-gh-asawicki: It's a bit wider than this one and it does not include `TableColumnIdentifier`, so I created a dedicated one.

PR #3968, opened 2025-08-26
chore: Fix unsafe query integration test

sfc-gh-asawicki:
Create a dedicated database to match only one result (using the precreated database could lead to matching a database from another pipeline, e.g., account level one, where the name differs only by the "AL" suffix).
Also, create databases with parameters.

### DIFF
--- a/pkg/sdk/testint/client_unsafe_extensions_integration_test.go
+++ b/pkg/sdk/testint/client_unsafe_extensions_integration_test.go
@@ -15,7 +15,11 @@ func TestInt_Client_UnsafeQuery(t *testing.T) {
 	ctx := testContext(t)
 
 	t.Run("test show databases", func(t *testing.T) {
-		sql := fmt.Sprintf("SHOW DATABASES LIKE '%%%s%%'", testClientHelper().Ids.DatabaseId().Name())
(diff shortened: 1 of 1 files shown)


pkg/sdk/testint/client_unsafe_extensions_integration_test.go
```diff
@@ -39,11 +43,11 @@ func TestInt_Client_UnsafeQuery(t *testing.T) {
 	})
 
 	t.Run("test more results", func(t *testing.T) {
-		db1, db1Cleanup := testClientHelper().Database.CreateDatabase(t)
+		db1, db1Cleanup := testClientHelper().Database.CreateDatabaseWithParametersSet(t)
```
sfc-gh-jcieslak, line 46: question: does it matter?

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (approved):

pkg/sdk/testint/client_unsafe_extensions_integration_test.go
sfc-gh-asawicki: It should be the default method used; it should not matter in this test, but no harm in it.

PR #3971, opened 2025-08-27
chore: Move more task tests to account level

sfc-gh-asawicki:
Move the remaining tests checking the default parameter values.


pkg/testacc/data_source_tasks_account_level_acceptance_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceparametersassert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config/datasourcemodel"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+// All tests in this file are temporarily moved to account level tests due to STATEMENT_TIMEOUT_IN_SECONDS being set on warehouse level and messing with the results.
```
sfc-gh-jmichalak, line 19: What's the exact value, can't we e.g. set this parameter to sensible values? Or use a dedicated warehouse?

sfc-gh-asawicki: Our current approach for warehouse problems is to move them to the account level. Ultimately, we can experiment with a dedicated warehouse or create a new warehouse specifically for parameter alteration tests (however, I couldn't find a test messing with the default warehouse for this parameter or with USER_TASK_TIMEOUT_MS, which has connected logic - each test I found was working on a new dedicated warehouse; maybe the issue lies deeper, e.g. creating new warehouse, altering the default for the given session - it requires deeper research); however, this is not like with the pre-created database in which schema/lower objects reside. Task does not reside in the warehouse, and if you check, you will see that some of these tests are not even setting any warehouse explicitly (as this is not required to set). So, in short, we still need tests that do not set the warehouse at all. For now, this is good enough to make the tests pass more often.

sfc-gh-jmichalak (approved):

PR #3978, opened 2025-08-27
feat: add integration tests for semantic views

sfc-gh-sarur:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
+//go:build account_level_tests
```
sfc-gh-jcieslak: Please change to `//go:build !account_level_tests`

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+			HasCreatedOnNotEmpty().
+			HasOwner("ACCOUNTADMIN").
+			HasOwnerRoleType("ROLE").
+			HasComment("comment"),
+		)
+	})
+
+	t.Run("describe semantic view", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifierInSchema(schema.ID())
+		request := sdk.NewCreateSemanticViewRequest(id, logicalTables).WithSemanticViewMetrics(metrics)
```
sfc-gh-jcieslak: nit: as you are using the same / similar create I would merge this test with the above one (basically remove this one, just move the Describe and assertions related to them to the test above)

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+			{
+				ObjectKind:    "TABLE",
+				ObjectName:    "TABLE2",
+				ParentEntity:  nil,
+				Property:      "BASE_TABLE_NAME",
+				PropertyValue: table2.Name,
+			},
+		}
+		// confirm the semantic view details match
+		assert.Equal(t, expectedDescription, semanticViewDetails)
```
sfc-gh-jcieslak: Hmm I think we should use Contains rather than Equal, just in case another property will be added to the Describe output

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+	})
+
```
sfc-gh-jcieslak, line 181: let's have each sub-test for sub-branch in the create of semantic view. One sub-test for each of:
  [ RELATIONSHIPS ( relationshipDef [ , ... ] ) ]
  [ FACTS ( semanticExpression [ , ... ] ) ]
  [ DIMENSIONS ( semanticExpression [ , ... ] ) ]
  [ METRICS ( semanticExpression [ , ... ] ) ] (this one is already done)

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+		err := client.SemanticViews.Create(ctx, request)
+		require.NoError(t, err)
+		t.Cleanup(testClientHelper().SemanticView.DropFunc(t, id))
+
+		semanticViewDetails, err := client.SemanticViews.Describe(ctx, id)
+		require.NoError(t, err)
+
+		parentEntity := "TABLE1"
+		expectedDescription := []sdk.SemanticViewDetails{
+			{
```
sfc-gh-jcieslak: nit: those details take a pretty significant amount of screen, it would be better to define custom assertions for those. I can see you started them, but didn't finish them or used.

cc: @sfc-gh-jmichalak @sfc-gh-asawicki do we have any examples for custom assertion with the list of properties like here? I couldn't find any, maybe there's something I missed.

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+func TestInt_SemanticView(t *testing.T) {
+	client := testClient(t)
+	ctx := testContext(t)
+
+	// create a database and add it to the cleanup queue
+	db, dbCleanup := testClientHelper().Database.CreateDatabaseWithParametersSet(t)
+	t.Cleanup(dbCleanup)
+
+	// create a schema and add it to the cleanup queue
+	schema, schemaCleanup := testClientHelper().Schema.CreateSchemaInDatabase(t, db.ID())
```
sfc-gh-jmichalak: I don't think we need to create a custom db/schema. We can reuse the test suite resources, simply use `		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()`

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+				PropertyValue: table2.Name,
+			},
+		}
+		// confirm the semantic view details match
+		assert.Equal(t, expectedDescription, semanticViewDetails)
+	})
+
+	t.Run("alter semantic view", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifierInSchema(schema.ID())
+		request := sdk.NewCreateSemanticViewRequest(id, logicalTables).WithSemanticViewMetrics(metrics).WithComment("comment")
```
sfc-gh-jmichalak, line 184: When testing methods other than create, you can use `SemanticViewClient` from the helpers package.

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+	t.Run("drop semantic view", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifierInSchema(schema.ID())
+		request := sdk.NewCreateSemanticViewRequest(id, logicalTables).WithSemanticViewMetrics(metrics)
+
+		err := client.SemanticViews.Create(ctx, request)
+		require.NoError(t, err)
+		t.Cleanup(testClientHelper().SemanticView.DropFunc(t, id))
+
+		// confirm the semantic view exists
+		_, err = client.SemanticViews.ShowByID(ctx, id)
```
sfc-gh-jmichalak: This call is not needed, we can call DROP just after CREATE.

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+		// confirm the semantic view no longer exists
+		_, err = client.SemanticViews.ShowByID(ctx, id)
+		require.Error(t, err)
+
+		// with if exists set to true, calling Drop again should not return an error
+		err = client.SemanticViews.Drop(ctx, dropRequest)
+		require.NoError(t, err)
+
+		// with if exists set to false, Drop should fail with an error if the semantic view does not exist
+		failingDropRequest := sdk.NewDropSemanticViewRequest(id).WithIfExists(false)
```
sfc-gh-jmichalak: This is also not needed, but we can have here `WithIfExists(false)` and `require.NoError(t, err)`.

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-jmichalak: +1, we should have custom assertions. We already have an example, but not with a generic attribute list: https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/acceptance/bettertestspoc/assert/objectassert/service_describe_snowflake_ext.go, though it's not generated automatically. We can either
- use `collections.FindFirst` in each assert
- convert a list of `sdk.SemanticViewDetails` to a map with name keys.

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+			{
+				ObjectKind:    "TABLE",
+				ObjectName:    "TABLE1",
+				ParentEntity:  nil,
+				Property:      "BASE_TABLE_NAME",
+				PropertyValue: table1.Name,
+			},
+			{
+				ObjectKind:    "METRIC",
+				ObjectName:    "METRIC1",
```
sfc-gh-jmichalak: What about the case here? We defined the metric as a lowercase name `metric1`, shouldn't we expect lowercase here as well? Maybe the metric name should be wrapped in `"`? In general, the provider is case sensitive.

sfc-gh-jmichalak (changes requested):
+1 to other comments

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: done

sfc-gh-sarur: Done

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: done

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: The output of the describe is in Uppercase for most properties, regardless of whether we use lower or uppercase. That's why I used upper case in the asserts.
One exception is the metric expression, which seems to have maintained its case. I am not sure how describe decides this

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: Done

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: ack

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: ack

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: done

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-asawicki: Yes, I would prefer having it packed like e.g. `ProcedureDetails`, and generate assertions on such a struct.

sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/assert/objectassert/semantic_view_describe_snowflake_ext.go
```diff
@@ -8,12 +8,3 @@ import (
 type SemanticViewDetailsAssert struct {
 	*assert.SnowflakeObjectAssert[[]sdk.SemanticViewDetails, sdk.SchemaObjectIdentifier]
 }
-
-//func SemanticViewDetails(t *testing.T, id sdk.SchemaObjectIdentifier) *SemanticViewDetailsAssert {
```
sfc-gh-jcieslak: I don't see custom describe assertions being added and used, please add those and use instead of regular assertions on slices of describe properties

pkg/sdk/semantic_view_ext.go
```diff
@@ -0,0 +1,25 @@
+package sdk
+
+func (s *CreateSemanticViewRequest) GetName() SchemaObjectIdentifier {
+	return s.name
+}
+
+func NewSemanticViewDetails(
```
sfc-gh-jmichalak: Nit: it should be moved to the test package.

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-jmichalak: I checked this, and it looks like the casing is supported for other values (check the details section for the sql). For now we can leave this, and continue with the resource, but ultimately we'd like to support lowercase tables and columns, as we already support them in the table definition.

<details>
create or replace table "orders" ("order_date" time, "o_totalprice" int, "o_orderdate" time, "o_orderkey" string, "o_custkey" string );
create or replace table "customer"("c_custkey" string, "c_name" string);
create or replace table "lineitem" ("l_orderkey" string, "l_linenumber" int, "l_extendedprice" int, "l_discount" int);
drop table lineitem;
show view tpch_rev_analysis;
desc semantic view tpch_rev_analysis;
CREATE or replace SEMANTIC VIEW tpch_rev_analysis

  TABLES (
    "orders" AS jmichalak.public."orders"
      PRIMARY KEY ("o_orderkey")
      WITH SYNONYMS ('sales orders')
      COMMENT = 'All orders table for the sales domain',
    "customers" AS jmichalak.public."customer"
      PRIMARY KEY ("c_custkey")
      COMMENT = 'Main table for customer data',
    "line_items" AS jmichalak.public."lineitem"
      PRIMARY KEY ("l_orderkey", "l_linenumber")
      COMMENT = 'Line items in orders'
  )

  RELATIONSHIPS (
    orders_to_customers AS
      "orders" ("o_custkey") REFERENCES "customers",
    line_item_to_orders AS
      "line_items" ("l_orderkey") REFERENCES "orders"
  )

  FACTS (
    "line_items"."line_item_id" AS CONCAT("l_orderkey", '-', "l_linenumber"),
    "orders"."count_line_items" AS COUNT("line_items"."line_item_id"),
    "line_items"."discounted_price" AS "l_extendedprice" * (1 - "l_discount")
      COMMENT = 'Extended price after discount'
  )

  DIMENSIONS (
    "customers"."customer_name" AS "customers"."c_name"
      WITH SYNONYMS = ('customer name')
      COMMENT = 'Name of the customer',
    "orders"."order_date" AS "o_orderdate"
      COMMENT = 'Date when the order was placed'
    -- orders.order_year AS YEAR(o_orderdate)
    --   COMMENT = 'Year when the order was placed'
  )

  METRICS (
    "customers"."customer_count" AS COUNT("c_custkey")
      COMMENT = 'Count of number of customers',
    "orders"."order_average_value" AS AVG("orders"."o_totalprice")
      COMMENT = 'Average order value across all orders',
    "orders"."average_line_items_per_order" AS AVG("orders"."count_line_items")
      COMMENT = 'Average number of line items per order'
  )
</details

sfc-gh-jmichalak (approved):
Approved for now. For me, we can continue implementing the resource, but still, there are some adjustments needed. We can fix them on our side.

pkg/sdk/semantic_view_ext.go
sfc-gh-sarur: done

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-asawicki: I agree with @sfc-gh-jmichalak. The provider is case-sensitive so every value that is or behaves like the identifier (so can be wrapped in double quotes to preserve casing) should be handled in a case-sensitive way.
I agree though, that we can merge it like this but before rolling out the resource, it need to be adjusted.

pkg/acceptance/bettertestspoc/assert/objectassert/semantic_view_describe_snowflake_ext.go
sfc-gh-asawicki: @sfc-gh-sarur what about this comment?

sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/assert/objectassert/semantic_view_describe_snowflake_ext.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+	"slices"
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+type SemanticViewDetailsCollection struct {
```
sfc-gh-sarur, line 13: I had to add a wrapper around the details as they are in a slice but the SnowflakeObjectAssert interface expects an object

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: I need some more time to do this change, it seems to be a bit more involved than I expected. Let's merge this PR as is and I will work on this along with the data sources

sfc-gh-sarur:
@sfc-gh-asawicki @sfc-gh-jmichalak I am not sure why this test is failing in the automated run 
```FAIL: TestInt_SecurityIntegrations/AlterSAML2Integration_-_REFRESH_SAML2_SNOWFLAKE_PRIVATE_KEY```
but I don't think it is related to my changes. PTAL and let me know if it is a concern, thanks

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak:
@sfc-gh-sarur Sorry, I missed your comment. This test is not related to your changes, and is not blocking; we already reported it to one of the other teams.

This PR looks good enough, and we are merging it. Still, there are a few follow-ups to be addressed later, but before the release (see unresolved comments). This can be handled by us (SNOW-2344309).

PR #3987, opened 2025-08-29
chore: Add tests to migration script

sfc-gh-jcieslak:
## Changes
- Move most of the program logic into a struct for a better e2e testability (mostly encapsulating inputs and outputs for the program, so they're easy to override)
- Change error handling to be more concentrated in one place (Run method)
- Change the way flags and arguments are parsed by using flag.FlagSet
- Add more tests for the grant privileges to account role grants mapping

## Next prs (in order)
- Import generation
- Usage guide
- Resource id generation
- Support for more grant mappings


pkg/scripts/migration_script/program.go
```diff
@@ -0,0 +1,180 @@
...
+
+	if err := commandLine.Parse(p.Args[1:]); err != nil {
+		return nil, err
+	}
+
+	// positional arguments
+	args := commandLine.Args()
+	if len(args) != 1 {
+		return nil, fmt.Errorf("no object type specified, use -h for help")
+	}
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Changed

pkg/scripts/migration_script/mappings_grants_test.go
```diff
@@ -4,105 +4,189 @@ import (
...
   with_grant_option = false
 }
-
+`,
+		},
+		{
+			name: "grants on schema",
+			inputRows: [][]string{
+				{"privilege", "granted_on", "name", "granted_to", "grantee_name", "grant_option"},
+				{"privilege", "granted_on", "name", "granted_to", "grantee_name", "grant_option"},
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Changed

pkg/scripts/migration_script/program_test.go
```diff
@@ -0,0 +1,112 @@
...
+  account_role_name = "ROLE_NAME"
+  on_account = true
+  privileges = ["CREATE DATABASE"]
+  with_grant_option = false
+}
+`,
+		},
+		{
+			name: "basic usage - block import format",
+			args: []string{"cmd", "-import=block", "grants"},
```
sfc-gh-asawicki, line 84: question: as it's import block, where is the import block generated? Is it coming in a separate change (at least this is what I understand from the PR description)? If so, it should at least have a TODO comment here

pkg/scripts/migration_script/program.go
```diff
@@ -0,0 +1,180 @@
...
+	parsedObjectType, err := ToObjectType(args[0])
```
sfc-gh-jmichalak, line 147: As I understand from the tests, the order still matters here for the positional argument and the flag. I think we should either
- (preferred) make it available to set the flag after the positional arguments, as most CLI programs allow that
- or make it super clear and provide some better error messages, because this behavior will be confusing.
Can be done in the next PRs.

sfc-gh-jmichalak (approved):

pkg/scripts/migration_script/program_test.go
sfc-gh-jcieslak: It was added in the follow-up (https://github.com/snowflakedb/terraform-provider-snowflake/pull/3990)

pkg/scripts/migration_script/program.go
sfc-gh-jcieslak: I agree, but I'm not sure if it will be possible with standard libraries. Most of the CLI tools use something like Cobra, but I didn't want to add more dependencies. I'll try to see what can we do with this.

sfc-gh-asawicki: This is how the flag pkg works - it stops parsing at the first non-flag argument. We just need to be clear, and for me the current behavior is fine (AFAIK it's also the standard unix behavior).

PR #3990, opened 2025-08-29
chore: Add import statement generation

sfc-gh-jcieslak:
## Changes
- Add import statement and block generation to the script output
- Add tests showing the structure of identifiers is mapped correctly and both options are generated correctly (also block import statement contains right quote escaping)

### DIFF
--- a/pkg/scripts/migration_script/mappings_grants.go
+++ b/pkg/scripts/migration_script/mappings_grants.go
@@ -45,7 +45,7 @@ func HandleGrants(config *Config, csvInput [][]string) (string, error) {
 		return TransformImportModel(config, importModel)
 	})
 	if err != nil {
-		log.Printf("Errors during import transformations: %v", err)
+		return "", fmt.Errorf("errors during import transformations: %v", err)
 	}
 
 	outputBuilder := new(strings.Builder)


pkg/scripts/migration_script/mappings_grants.go
```diff
@@ -63,37 +77,79 @@ func MapGrantToModel(grantGroup []sdk.Grant) (accconfig.ResourceModel, error) {
...
-				WithWithGrantOption(grant.GrantOption),
-			nil
+		resourceModel := model.GrantPrivilegesToAccountRole("test_resource_name_on_account", grant.GranteeName.Name()).
+			WithPrivileges(privileges).
+			WithOnAccount(true).
+			WithWithGrantOption(grant.GrantOption)
+
+		return resourceModel, &ImportModel{
+			ResourceAddress: resourceModel.ResourceReference(),
+			Id: fmt.Sprintf(
```
sfc-gh-jcieslak: note: The `resources.GrantPrivilegesToAccountRoleId` could be used to construct the identifier. Initially I went for this approach, but I would like to hear wdyt, should we use this struct instead?

sfc-gh-asawicki: We should not duplicate this logic but use `GrantPrivilegesToAccountRoleId`.

pkg/scripts/migration_script/import_mappings.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+type ImportModel struct {
+	ResourceAddress string
+	Id              string
+}
+
+// IdEscaped returns the ID with escaped quotes for use in Terraform import blocks.
+func (im ImportModel) IdEscaped() string {
+	return strings.ReplaceAll(im.Id, "\"", "\\\"")
```
sfc-gh-asawicki: question: is it correct? shouldn't it be just `"` -> `\"`?

pkg/scripts/migration_script/mappings_grants_test.go
```diff
@@ -176,13 +182,17 @@ resource "snowflake_grant_privileges_to_account_role" "test_resource_name_on_sch
   privileges = ["INSERT", "SELECT"]
   with_grant_option = false
 }
+# terraform import snowflake_grant_privileges_to_account_role.test_resource_name_on_schema_object '"TEST_ROLE"|false|false|INSERT,SELECT|OnSchemaObject|OnObject|TABLE|"TEST_DATABASE"."TEST_SCHEMA"."TEST_TABLE"'
```
sfc-gh-asawicki, line 185: nit: maybe let's put import comment right before the resource and not right after? Usually comment is about something underneath it and not above it

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: Alright then, used `GrantPrivilegesToAccountRoleId`

pkg/scripts/migration_script/import_mappings.go
sfc-gh-jcieslak: Yeah, and it is, I'll use raw strings, it should clear things up

pkg/scripts/migration_script/mappings_grants_test.go
sfc-gh-jcieslak: They're in a group anyway, the example result will be:
```
resource1
resource2
resource3

import1
import2
import3
```
I think it makes sense because it's better to group those as they need to be processed further (uncomment, paste to the terminal for each, and run terraform plan/apply). Do you mean to put it also in a group above all the resources, or did you think they are usually below each resource?

pkg/scripts/migration_script/mappings_grants.go
```diff
@@ -75,81 +76,69 @@ func MapGrantToModel(grantGroup []sdk.Grant) (accconfig.ResourceModel, *ImportMo
 	slices.Sort(privileges)
 	privileges = slices.Compact(privileges)
 
+	var resourceModel accconfig.ResourceModel
+	resourceId := resources.GrantPrivilegesToAccountRoleId{
```
sfc-gh-asawicki, line 80: it's better but it's still pretty low-level and error-prone that this struct is populated manually. Could we at least extract constructors dedicated for each "situation"? The provider grants parsing logic/tests could also benefit from it. We can do it in the next PR.

pkg/scripts/migration_script/mappings_grants_test.go
sfc-gh-asawicki: Okay, so I thought that import will be next to the resource in question. So that it's clear how given resource is imported. If the ordering is preserved, it's also fine. I am just trying to be in the user's shoes here. If their only need is to mindlessly copy the output and move on, it's fine. If we want to also educate, than maybe having it paired up is a better choice. At the end, processing is easy in both options (in the paired up one, you can just search for the #-starting lines to get all imports).

Im fine with keeping it the implemented way for now.

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: Will do in one of the next prs

PR #3993, opened 2025-09-01
chore: Add resource ID generation to migration script

sfc-gh-jcieslak:
## Changes
- Add resource ID generation
  - The generation was based on the properties used for resource grouping, with a few exceptions (privileges were not taken into account, and the account name). We don't need privileges in IDs for them to be unique within one script run (we are grouping similar grant resources to have all related privileges in the same resource), and the account name doesn't need to be included, as grants can only be granted on the current account (this I'm not 100% sure of as there may be some minor cases where this may occur, but afaik only current account can participate in grants to account role, so for majority, or all cases, the account name will be the same).
  - The dot in resource ID mapping is only added for visual clarity with multi-part identifiers
  - Tested with Terraform that even 500-character resource names are handled, so we should be good regarding the resource name length

## Next prs
Document the limitations of generated resource names


pkg/scripts/migration_script/mappings.go
```diff
@@ -1,6 +1,29 @@
...
+
+var resourceIdAllowedCharacters = regexp.MustCompile("[a-zA-Z0-9\\-_]+")
+
+func MapResourceId(resourceId string) string {
+	mappedResourceId := strings.Map(func(r rune) rune {
+		switch r {
+		case '.':
+			return '_'
+		default:
+			if resourceIdAllowedCharacters.MatchString(string(r)) {
```
sfc-gh-jmichalak, line 18: Please remember to document this behavior in the readme.

pkg/scripts/migration_script/mappings.go
```diff
@@ -1,6 +1,29 @@
...
+import (
+	"regexp"
+	"strings"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
+)
+
+var resourceIdAllowedCharacters = regexp.MustCompile("[a-zA-Z0-9\\-_]+")
+
+func MapResourceId(resourceId string) string {
```
sfc-gh-jmichalak: Nit: `NormalizeResourceID` or `ToTerraformName`.

sfc-gh-jmichalak (approved):

pkg/scripts/migration_script/mappings.go
sfc-gh-jcieslak: 👍 changed

pkg/scripts/migration_script/mappings.go
sfc-gh-jcieslak: Yup I added it as next pr in the desc (and added to my todo list)

pkg/scripts/migration_script/mappings.go
```diff
@@ -1,6 +1,29 @@
...
+func NormalizeResourceId(resourceId string) string {
+	mappedResourceId := strings.Map(func(r rune) rune {
+		switch r {
+		case '.':
+			return '_'
+		default:
+			if resourceIdAllowedCharacters.MatchString(string(r)) {
+				return r
+			}
+			return -1
```
sfc-gh-asawicki, line 21: -1?

pkg/scripts/migration_script/mappings_grants.go
```diff
@@ -77,59 +77,72 @@ func MapGrantToModel(grantGroup []sdk.Grant) (accconfig.ResourceModel, *ImportMo
...
 
 	var resourceModel accconfig.ResourceModel
-	resourceId := resources.GrantPrivilegesToAccountRoleId{
+	stateResourceId := resources.GrantPrivilegesToAccountRoleId{
 		RoleName:        grant.GranteeName.(sdk.AccountObjectIdentifier),
 		WithGrantOption: grant.GrantOption,
 		Privileges:      privileges,
 	}
 
+	withGrantOption := "without"
```
sfc-gh-asawicki: this logic is odd here. The naming logic should be fully extracted to not clutter the logic here. Like a dedicated function containing all the logic.

sfc-gh-asawicki (changes requested):

pkg/scripts/migration_script/mappings.go
sfc-gh-jcieslak: Yeah, that's what the strings.Map expects (if negative number is returned, then the character is discarded rather than mapped).

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: Extracted

PR #3995, opened 2025-09-02
chore: Add more mappings for the grant migration script part1

sfc-gh-jcieslak:
## Changes
- Add support for:
  - `grant_privileges_to_database_role`
  - `grant_account_role`
  - `grant_database_role`
- Add missing Role column in the sdk.Grant (only touched the script part, but I'm guessing the Role mapping should be available in the SDK too)

## Next steps
- Create a ticket for Role mapping in the SDK (SHOW GRANTS refactor issue)
- Add other missing mappings

### DIFF
--- a/pkg/scripts/migration_script/converter_grants.go
+++ b/pkg/scripts/migration_script/converter_grants.go
@@ -14,7 +14,6 @@ type GrantCsvRow struct {
 	GrantedOn   string `csv:"granted_on"`
 	GrantOn     string `csv:"grant_on"`
 	Name        string `csv:"name"`
-	Role        string `csv:"role"`
 	GrantedTo   string `csv:"granted_to"`
 	GrantTo     string `csv:"grant_to"`
 	GranteeName string `csv:"grantee_name"`
@@ -63,15 +62,6 @@ func (row GrantCsvRow) convert() (*sdk.Grant, error) {
 		name = sdk.NewObjectIdentifierFromFullyQualifiedName(row.Name)
 	}
 
-	var role sdk.ObjectIdentifier
-	if row.Role != "" {
-		role, err = sdk.ParseDatabaseObjectIdentifier(row.Role)
-		if err != nil {
-			log.Printf("[DEBUG] Failed to parse identifier [%s], err = \"%s\"; falling back to fully qualified name conversion", row.Role, err)
-			role = sdk.NewObjectIdentifierFromFullyQualifiedName(row.Role)
-		}
-	}
-
 	return &sdk.Grant{
 		Privilege:   row.Privilege,
 		GrantedOn:   grantedOn,
@@ -79,7 +69,6 @@ func (row GrantCsvRow) convert() (*sdk.Grant, error) {
 		GrantedTo:   grantedTo,
 		GrantTo:     grantTo,
 		Name:        name,
-		Role:        role,
 		GranteeName: sdk.NewAccountObjectIdentifier(row.GranteeName),
 		GrantOption: row.GrantOption,
 		GrantedBy:   sdk.NewAccountObjectIdentifier(row.GrantedBy),
--- a/pkg/scripts/migration_script/mappings_grants.go
+++ b/pkg/scripts/migration_script/mappings_grants.go
@@ -78,26 +78,18 @@ func MapGrantToModel(grantGroup []sdk.Grant) (accconfig.ResourceModel, *ImportMo
 	slices.Sort(privileges)
 	privileges = slices.Compact(privileges)
 
-	isRoleDatabaseObjectIdentifier := func(role sdk.ObjectIdentifier) bool {
-		_, ok := role.(sdk.DatabaseObjectIdentifier)
-		return ok
-	}
-
 	switch {
-	// Granting an account role
-	// When calling SHOW GRANTS TO ROLE / USER, the USAGE privilege should be shown with the ROLE granted_on field.
-	// When calling SHOW GRANTS OF ROLE, the Role field should be populated.
-	// The granted_to field should always point to either ROLE or USER.
-	case (grant.Role != nil && !isRoleDatabaseObjectIdentifier(grant.Role) || grant.Privilege == "USAGE" && grant.GrantedOn == sdk.ObjectTypeRole && grant.Name != nil) &&
+	case grant.Privilege == "USAGE" &&
+		grant.GrantedOn == sdk.ObjectTypeRole && grant.Name != nil &&
 		(grant.GrantedTo == sdk.ObjectTypeRole || grant.GrantedTo == sdk.ObjectTypeUser):
 		return MapToGrantAccountRole(grant)
-	// Granting a database role (cases similar to the above; different handling for different SHOW GRANTS calls)
-	case (grant.Role != nil && isRoleDatabaseObjectIdentifier(grant.Role) || grant.Privilege == "USAGE" && grant.GrantedOn == sdk.ObjectTypeDatabaseRole && grant.Name != nil) &&
+	case grant.Privilege == "USAGE" &&
+		grant.GrantedOn == sdk.ObjectTypeDatabaseRole && grant.Name != nil &&
 		(grant.GrantedTo == sdk.ObjectTypeDatabaseRole || grant.GrantedTo == sdk.ObjectTypeRole):
 		return MapToGrantDatabaseRole(grant)
-	case len(privileges) > 0 && grant.GrantedTo == sdk.ObjectTypeRole:
+	case grant.GrantedTo == sdk.ObjectTypeRole:
 		return MapToGrantPrivilegesToAccountRole(grant, privileges)
-	case len(privileges) > 0 && grant.GrantedTo == sdk.ObjectTypeDatabaseRole:
+	case grant.GrantedTo == sdk.ObjectTypeDatabaseRole:
 		return MapToGrantPrivilegesToDatabaseRole(grant, privileges)
 	default:
 		return nil, nil, fmt.Errorf("unsupported grant mapping")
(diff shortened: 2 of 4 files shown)


pkg/scripts/migration_script/converter_grants.go
```diff
@@ -62,13 +63,23 @@ func (row GrantCsvRow) convert() (*sdk.Grant, error) {
 		name = sdk.NewObjectIdentifierFromFullyQualifiedName(row.Name)
 	}
 
+	var role sdk.ObjectIdentifier
+	if row.Role != "" {
+		role, err = sdk.ParseDatabaseObjectIdentifier(row.Role)
+		if err != nil {
+			log.Printf("[DEBUG] Failed to parse identifier [%s], err = \"%s\"; falling back to fully qualified name conversion", row.Name, err)
```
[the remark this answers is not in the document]
sfc-gh-asawicki: +1

pkg/scripts/migration_script/mappings.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
 )
 
-var resourceIdAllowedCharacters = regexp.MustCompile("[a-zA-Z0-9\\-_]+")
+var resourceIdAllowedCharacters = regexp.MustCompile("[a-zA-Z0-9\\-_]")
 
 func NormalizeResourceId(resourceId string) string {
 	mappedResourceId := strings.Map(func(r rune) rune {
```
sfc-gh-asawicki: let's simplify a logic here by using the ReplaceAllString with the regex negation (i.e. replace all characters NOT matching). And also this -1 will disappear. :D

pkg/scripts/migration_script/mappings_grants.go
```diff
@@ -70,12 +70,113 @@ func GroupGrants(grants []sdk.Grant) map[string][]sdk.Grant {
 func MapGrantToModel(grantGroup []sdk.Grant) (accconfig.ResourceModel, *ImportModel, error) {
```
sfc-gh-asawicki, line 70: thought: as this function grows I wonder if we should really keep all the grant types in one method, and have a complex switch logic to pick a proper case to handle. Looking at this logic, I am pretty confident we will be hit by questions of even more corner cases. I wonder if we would be better of by:
- providing a SQL fetching the right columns only for the given case
- handling the given case by setting the dedicated flag in the script.
It would lead to a simpler logic and less maintenance costs.

Let's discuss.

pkg/scripts/migration_script/mappings_grants.go
```diff
@@ -70,12 +70,113 @@ func GroupGrants(grants []sdk.Grant) map[string][]sdk.Grant {
...
+
+		// Depending on the SHOW GRANTS command options, we can either get AccountObjectIdentifier or DatabaseObjectIdentifier in granteeName column
+		if id, err := sdk.ParseDatabaseObjectIdentifier(grant.GranteeName.Name()); err == nil {
+			granteeName = id
+		}
+
+		resourceId := NormalizeResourceId(fmt.Sprintf("grant_%s_to_database_role_%s", roleIdentifier.FullyQualifiedName(), granteeName.FullyQualifiedName()))
+		resourceModel := model.GrantDatabaseRole(resourceId, roleIdentifier.FullyQualifiedName()).WithParentDatabaseRoleName(granteeName.FullyQualifiedName())
+
+		stateResourceId := fmt.Sprintf("%s|%s|%s", roleIdentifier.FullyQualifiedName(), sdk.ObjectTypeDatabaseRole.String(), granteeName.FullyQualifiedName())
```
sfc-gh-asawicki: we don;t have dedicated identifiuer types for these? let's add and reuse them

sfc-gh-asawicki (changes requested):

pkg/sdk/grants.go
```diff
@@ -213,6 +213,7 @@ type Grant struct {
 	GrantedOn   ObjectType
 	GrantOn     ObjectType
 	Name        ObjectIdentifier
+	Role        ObjectIdentifier
```
sfc-gh-jmichalak: What about GrantRow?

pkg/scripts/migration_script/mappings_grants.go
```diff
@@ -70,75 +70,267 @@ func GroupGrants(grants []sdk.Grant) map[string][]sdk.Grant {
...
+	switch {
+	// Granting an account role
+	// When calling SHOW GRANTS TO ROLE / USER, the USAGE privilege should be shown with the ROLE granted_on field.
+	// When calling SHOW GRANTS OF ROLE, the Role field should be populated.
+	// The granted_to field should always point to either ROLE or USER.
+	case (grant.Role != nil && !isRoleDatabaseObjectIdentifier(grant.Role) || grant.Privilege == "USAGE" && grant.GrantedOn == sdk.ObjectTypeRole && grant.Name != nil) &&
+		(grant.GrantedTo == sdk.ObjectTypeRole || grant.GrantedTo == sdk.ObjectTypeUser):
+		return MapToGrantAccountRole(grant)
+	// Granting a database role (cases similar to the above; different handling for different SHOW GRANTS calls)
+	case (grant.Role != nil && isRoleDatabaseObjectIdentifier(grant.Role) || grant.Privilege == "USAGE" && grant.GrantedOn == sdk.ObjectTypeDatabaseRole && grant.Name != nil) &&
```
sfc-gh-jmichalak: Nit: Move to sdk? The same above.

pkg/scripts/migration_script/mappings_grants.go
```diff
@@ -70,75 +70,267 @@ func GroupGrants(grants []sdk.Grant) map[string][]sdk.Grant {
...
 
+	isRoleDatabaseObjectIdentifier := func(role sdk.ObjectIdentifier) bool {
+		_, ok := role.(sdk.DatabaseObjectIdentifier)
+		return ok
+	}
+
+	switch {
+	// Granting an account role
+	// When calling SHOW GRANTS TO ROLE / USER, the USAGE privilege should be shown with the ROLE granted_on field.
+	// When calling SHOW GRANTS OF ROLE, the Role field should be populated.
```
sfc-gh-jmichalak: Maybe let's just limit the usage to one of them and simplify code?

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: Next pr? I'm already going through resource identifiers in other pr (creating constructors for existing ones which was requested in one of the previous migration script prs)

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: It feels there's a trade-off between ease of use and maintenance. Having specific SQLs or dedicated flags for different objects would be simpler, but still, it needs some time to adjust the code, and it would be more confusing on the user end (I think).

The current approach, despite not being the easiest to understand, is doing the job while keeping the easy approach for the user (ability to choose different show grant filtering options). There are different quirks with SHOW GRANTS, but as long as all output options are tested, we should be good (the tested ones should be documented, or if there will be a reported edge case, we would document it too, but I'm also a bit afraid how many of those we may encounter).

I'm wondering if there's a middle ground between the two approaches to simplify the logic while still maintaining the ease of use on the user's side. I didn't think of a better solution now, but I'll try to later today. We can talk tomorrow morning, what should be the direction if nothing better is proposed.

pkg/scripts/migration_script/mappings.go
sfc-gh-jcieslak: Changed, but kept the old behavior to avoid excessive underscores in identifiers

pkg/sdk/grants.go
sfc-gh-jcieslak: I didn't want to add the mapping in SDK (as it could change the production behavior, I would like to postpone the mapping and only added it for the script)

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: @sfc-gh-asawicki maybe moving (or adding) some of the logic to `sdk.Grant` (or new struct that derives from it that is defined only in the script directory) would at least partially solve the issue? We would delegate the logic for guessing what operation a given grant represents 🤔

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: I think it would be confusing for the users (or it would limit their capabilities they have with the current solution), but on the other hand, maybe we should just drop the support for one of those just for simplicity... 🤔

pkg/scripts/migration_script/mappings.go
```diff
@@ -7,22 +7,11 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
 )
 
-var resourceIdAllowedCharacters = regexp.MustCompile("[a-zA-Z0-9\\-_]+")
+var resourceIdAllowedCharacters = regexp.MustCompile("[^a-zA-Z0-9\\-_]")
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Good point, will address in the follow-up

sfc-gh-jmichalak (approved):

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-asawicki: @sfc-gh-jcieslak please write here what we have discussed today

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-asawicki: Keeping it open then, please link here the PR

pkg/scripts/migration_script/mappings.go
```diff
@@ -7,22 +7,11 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
 )
 
-var resourceIdAllowedCharacters = regexp.MustCompile("[a-zA-Z0-9\\-_]")
+var resourceIdAllowedCharacters = regexp.MustCompile("[^a-zA-Z0-9\\-_]")
```
sfc-gh-asawicki, line 10: @sfc-gh-jcieslak the name should be changed. With the negation these are not the allowed characters

pkg/scripts/migration_script/mappings.go
```diff
@@ -7,22 +7,11 @@ import (
...
-			if resourceIdAllowedCharacters.MatchString(string(r)) {
-				return r
-			}
-			return -1
-		}
-	}, resourceId)
-
-	return "snowflake_generated_" + mappedResourceId
+	resourceId = strings.ReplaceAll(resourceId, `.`, "_")
+	return "snowflake_generated_" + string(resourceIdAllowedCharacters.ReplaceAll([]byte(resourceId), []byte("")))
```
sfc-gh-asawicki, line 14: I suggested ReplaceAllString (the byte casting is not necessary)

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-asawicki: It would have to be script only struct. And then, I'm fine with it

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: It's applied here: https://github.com/snowflakedb/terraform-provider-snowflake/pull/3999

pkg/scripts/migration_script/mappings.go
sfc-gh-jcieslak: Yeah, it was adjusted in https://github.com/snowflakedb/terraform-provider-snowflake/pull/3999

pkg/scripts/migration_script/mappings.go
sfc-gh-jcieslak: Oh ok, I missed it. Will fix it later on (a small follow-up fix after I'll be back).

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: We discussed the topic offline, and decided to simplify the code by limiting the number of supported SHOW GRANTS options. The functionality stays the same, but the input has limited format.

pkg/scripts/migration_script/mappings_grants.go
sfc-gh-jcieslak: After the adjustment and that we agreed to simlify the switch I don't think it's that bad anymore, would leave as is or adjust in the follow-up

PR #3996, opened 2025-09-02
chore: Extract account level tests for oauth integrations

sfc-gh-jmichalak:
Recently, in the pipelines, we could see errors like `390950 (22023): Integration of this type already exists in the account.` in tests handling oauth security integrations for partner customers. It was happening because not all tests were properly marked as account level.
- Extract tag integration tests (object integration tests are already account-level).
- Merge resource tests and mark them all as account level.
- Extract assertion code.

TODO (Jira tickets)
- Alter the test level in the integration tests. Currently, all security integrations are marked as account-level, but only this one type needs to be.
- Change the tag association ID.


pkg/sdk/testint/helpers_test.go
```diff
@@ -152,3 +153,38 @@ func createExternalTable(t *testing.T) (*sdk.ExternalTable, func()) {
 
 	return testClientHelper().ExternalTable.CreateWithLocation(t, stage.Location())
 }
+
+func assertTagSet(t *testing.T, tagId sdk.SchemaObjectIdentifier, objectId sdk.ObjectIdentifier, objectType sdk.ObjectType, tagValue string) {
+	t.Helper()
+	client := testClient(t)
+	ctx := testContext(t)
+	returnedTagValue, err := client.SystemFunctions.GetTag(ctx, tagId, objectId, objectType)
```
sfc-gh-asawicki: nit: why we are using testClient instead of testClientHelper?
nit2: we could have a dedicated assertion to check this

Let's add two TODOs here

sfc-gh-jmichalak: 1. Done here.
2. Added a TODO.

PR #3997, opened 2025-09-02
chore: Prove session parameters behavior, task test problems, and add PoC for better provider testing cache

sfc-gh-asawicki:
- Add test proving the session parameter behavior (`TestAcc_Task_ProveSessionParameterBehavior`)
- Add test proving the problematic test behavior (`TestAcc_Task_ProveCurrentDriftBehavior`)
- Add PoC for a better cache (and use it in the added tests); it's not meant to be feature complete
- Extract pkg `invokeactionassert` temporarily for special asserts that run actions or invoke special checks (we already had one such assert, moved it there too; we can improve this type of assertions later on)
- Add a special `assert` (`UpdateAccountParameterTemporarily`) that does not check anything, but alters the parameter as it needs to happen between apply and refresh to achieve the problematic situation
- Generate model builder and resource assertion for `snowflake_execute`; add custom methods and checks
- Apply the single attribute workaround to the provider model builder

### DIFF
--- a/pkg/testacc/resource_task_account_level_acceptance_test.go
+++ b/pkg/testacc/resource_task_account_level_acceptance_test.go
@@ -72,7 +72,6 @@ func TestAcc_Task_ProveSessionParameterBehavior(t *testing.T) {
 }
 
 func TestAcc_Task_ProveCurrentDriftBehavior(t *testing.T) {
-	testClient().Parameter.UnsetAccountParameter(t, sdk.AccountParameterStatementTimeoutInSeconds)
 	id := testClient().Ids.RandomSchemaObjectIdentifier()
 	statement := "SELECT 1"
 


sfc-gh-jcieslak (approved):

pkg/testacc/1_setup_provider_test.go
```diff
@@ -121,3 +131,62 @@ func configureProviderWithConfigCache(ctx context.Context, d *schema.ResourceDat
...
+		if cached, ok := providerInitializationCache[key]; accTestEnabled && ok {
+			accTestLog.Printf("[DEBUG] Returning cached provider configuration result for key %s", key)
+			if cached.providerCtx != nil {
+				accTestLog.Printf("[DEBUG] Returning cached provider configuration context")
+				return cached.providerCtx, nil
+			} else if cached.clientErrorDiag.HasError() {
+				accTestLog.Printf("[DEBUG] Returning cached provider configuration error")
+				return nil, cached.clientErrorDiag
+			}
+		}
```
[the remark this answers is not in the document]
sfc-gh-asawicki: We will decide about this in SNOW-2312385, as the accTestEnabled guard will be most probably removed

pkg/testacc/1_setup_provider_test.go
```diff
@@ -121,3 +131,62 @@ func configureProviderWithConfigCache(ctx context.Context, d *schema.ResourceDat
...
+		// TODO [SNOW-2312385]: do we still keep this guard if this configure func override is only used in the acceptance tests already?
+		accTestEnabled, err := oswrapper.GetenvBool("TF_ACC")
+		if err != nil {
+			accTestEnabled = false
+			accTestLog.Printf("[ERROR] TF_ACC environmental variable has incorrect format: %v, using %v as a default value", err, accTestEnabled)
+		}
+
+		// check if we cached initialized provider context with the key already
+		if cached, ok := providerInitializationCache[key]; accTestEnabled && ok {
+			accTestLog.Printf("[DEBUG] Returning cached provider configuration result for key %s", key)
```
sfc-gh-jmichalak, line 163: Nit: this suggests that we are returning here, but there's a case where we don't return early, let's rephrase.

sfc-gh-jmichalak (approved):

pkg/testacc/1_setup_provider_test.go
sfc-gh-asawicki: I added it to SNOW-2312385.

PR #4002, opened 2025-09-03
feat: Add generation field to warehouses

sfc-gh-jmichalak:
TODO


pkg/acceptance/planchecks/expect_no_change_plan_check.go
```diff
@@ -0,0 +1,78 @@
...
+
+var _ plancheck.PlanCheck = expectChangePlanCheck{}
+
+type expectNoChangePlanCheck struct {
+	resourceAddress string
+	attribute       string
+}
+
+// TODO [SNOW-1473409]: test
+func (e expectNoChangePlanCheck) CheckPlan(_ context.Context, req plancheck.CheckPlanRequest, resp *plancheck.CheckPlanResponse) {
```
sfc-gh-asawicki, line 21: We already have usages like:
```
planchecks.ExpectChange("snowflake_database.test", "data_retention_time_in_days", tfjson.ActionNoop, sdk.String("25"), sdk.String("25")),
```
which serve the same purpose.

This new method uses a lot of code from the existing one. Let's utilize the existing one by any of the following:
1. Use the method directly and remove the new one.
2. Add only `ExpectNoChangeOnField` delegating to the existing one (with one more parameter - value).

If picking option 2, please add to the next PRs using this method instead of Noop above.

pkg/sdk/warehouses.go
```diff
@@ -203,6 +203,55 @@ var AllWarehouseResourceConstraintsWithoutGenerations = []string{
...
+	}
+}
+
+var AllWarehouseGenerations = []string{
+	string(WarehouseGenerationStandardGen1),
+	string(WarehouseGenerationStandardGen2),
+}
+
+func IsWarehouseResourceConstraintForStandard(s WarehouseResourceConstraint) bool {
+	return slices.Contains([]string{string(WarehouseResourceConstraintStandardGen1), string(WarehouseResourceConstraintStandardGen2)}, string(s))
```
sfc-gh-asawicki: nit: extract this similarly to AllWarehouseResourceConstraintsWithoutGenerations

pkg/sdk/warehouses.go
```diff
@@ -203,6 +203,55 @@ var AllWarehouseResourceConstraintsWithoutGenerations = []string{
...
+}
+
+func WarehouseResourceConstraintToWarehouseGeneration(s WarehouseResourceConstraint) (WarehouseGeneration, error) {
```
sfc-gh-asawicki: nit: let's make it a method on `WarehouseResourceConstraint`

pkg/sdk/warehouses.go
```diff
@@ -203,6 +203,55 @@ var AllWarehouseResourceConstraintsWithoutGenerations = []string{
...
+	case WarehouseResourceConstraintStandardGen1:
+		return WarehouseGenerationStandardGen1, nil
+	case WarehouseResourceConstraintStandardGen2:
+		return WarehouseGenerationStandardGen2, nil
+	default:
+		return "", fmt.Errorf("invalid resource constraint for generation: %s", s)
+	}
+}
+
+func WarehouseGenerationToWarehouseResourceConstraint(s WarehouseGeneration) (WarehouseResourceConstraint, error) {
```
sfc-gh-asawicki: nit: let's make it a method on `WarehouseGeneration`

pkg/resources/warehouse.go
```diff
@@ -452,9 +502,21 @@ func GetReadWarehouseFunc(withExternalChangesMarking bool) schema.ReadContextFun
...
+						generation, err := sdk.WarehouseResourceConstraintToWarehouseGeneration(*w.ResourceConstraint)
+						if err != nil {
+							return diag.FromErr(err)
+						}
+						outputMappings = append(outputMappings, outputMapping{"resource_constraint", "generation", string(*w.ResourceConstraint), string(generation), nil})
+					}
 				default:
 					log.Printf("[DEBUG] handling resource_constraint for warehouse type %s is not supported, ignoring", w.Type)
 				}
+			} else {
```
sfc-gh-asawicki: question: why do we need this else? Also, are the empty values correct?

pkg/resources/warehouse.go
```diff
@@ -452,9 +502,21 @@ func GetReadWarehouseFunc(withExternalChangesMarking bool) schema.ReadContextFun
 					} else {
 						outputMappings = append(outputMappings, outputMapping{"resource_constraint", "resource_constraint", string(*w.ResourceConstraint), string(*w.ResourceConstraint), nil})
 					}
+				case sdk.WarehouseTypeStandard:
+					if !sdk.IsWarehouseResourceConstraintForStandard(*w.ResourceConstraint) {
```
sfc-gh-asawicki: question: This logic is fully based on the Snowflake response: constraint and type. Currently, the type strictly dictates what resource constraint values are possible. Let's say, that in the future, there will be some new value, available for both types: should we fail with error like now, or should we just swallow it, so that older binaries continue to work?

pkg/resources/warehouse.go
```diff
@@ -452,9 +502,21 @@ func GetReadWarehouseFunc(withExternalChangesMarking bool) schema.ReadContextFun
...
+						return diag.FromErr(fmt.Errorf("resource constraint %s is not supported for standard warehouses", *w.ResourceConstraint))
+					} else {
+						generation, err := sdk.WarehouseResourceConstraintToWarehouseGeneration(*w.ResourceConstraint)
+						if err != nil {
+							return diag.FromErr(err)
+						}
+						outputMappings = append(outputMappings, outputMapping{"resource_constraint", "generation", string(*w.ResourceConstraint), string(generation), nil})
```
sfc-gh-asawicki: Thought: I think we should consider a different implementation here. This whole logic would simplify if we extracted two separate fields in show_output and filled them based on the resource constraint value.

pkg/resources/warehouse.go
```diff
@@ -117,7 +117,36 @@ var warehouseSchema = map[string]*schema.Schema{
...
 		DiffSuppressFunc: SuppressIfAny(NormalizeAndCompare(sdk.ToWarehouseResourceConstraintWithoutGeneration), IgnoreChangeToCurrentSnowflakeValueInShow("resource_constraint")),
-		Description:      fmt.Sprintf("Specifies the resource constraint for the warehouse. Please check [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/sql/create-warehouse#optional-properties-objectproperties) for required warehouse sizes for each resource constraint. Valid values are (case-insensitive): %s.", possibleValuesListed(sdk.AllWarehouseResourceConstraintsWithoutGenerations)),
+		Description:      fmt.Sprintf("Specifies the resource constraint for the warehouse. Only available for snowpark-optimized warehouses. For setting generation please use the `generation` field. Please check [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/sql/create-warehouse#optional-properties-objectproperties) for required warehouse sizes for each resource constraint. Valid values are (case-insensitive): %s.", possibleValuesListed(sdk.AllWarehouseResourceConstraintsWithoutGenerations)),
+		ConflictsWith:    []string{"generation"},
+	},
+	"generation": {
```
sfc-gh-asawicki, line 123: Please add state upgrader to fill this value

pkg/testacc/resource_warehouse_acceptance_test.go
```diff
@@ -2303,6 +2339,200 @@ func TestAcc_Warehouse_ResourceConstraint(t *testing.T) {
...
+					resourceassert.WarehouseResource(t, warehouseModelStandardAndGeneration.ResourceReference()).
+						HasWarehouseTypeString(string(sdk.WarehouseTypeStandard)).
+						HasNoResourceConstraint().
+						HasGenerationString(string(sdk.WarehouseGenerationStandardGen1)),
+					resourceshowoutputassert.WarehouseShowOutput(t, warehouseModelStandardAndGeneration.ResourceReference()).
+						HasType(sdk.WarehouseTypeStandard).
+						HasResourceConstraint(sdk.WarehouseResourceConstraintStandardGen1),
+				),
+			},
+			// import when resource constraint in config
```
sfc-gh-asawicki: nit: resource constraint -> generation

pkg/testacc/resource_warehouse_acceptance_test.go
```diff
@@ -2303,6 +2339,200 @@ func TestAcc_Warehouse_ResourceConstraint(t *testing.T) {
...
+						HasNameString(id.Name()).
+						HasWarehouseTypeString(string(sdk.WarehouseTypeStandard)).
+						HasNoResourceConstraint().
+						HasGenerationString(string(sdk.WarehouseGenerationStandardGen1)),
+					resourceshowoutputassert.ImportedWarehouseShowOutput(t, helpers.EncodeResourceIdentifier(id)).
+						HasType(sdk.WarehouseTypeStandard).
+						HasResourceConstraint(sdk.WarehouseResourceConstraintStandardGen1),
+				),
+			},
+			// change resource constraint in config
```
sfc-gh-asawicki: nit: resource constraint -> generation

pkg/testacc/resource_warehouse_acceptance_test.go
```diff
@@ -2303,6 +2339,200 @@ func TestAcc_Warehouse_ResourceConstraint(t *testing.T) {
...
+					resourceassert.WarehouseResource(t, warehouseModelStandardAndGeneration2.ResourceReference()).
+						HasWarehouseTypeString(string(sdk.WarehouseTypeStandard)).
+						HasGenerationString(string(sdk.WarehouseGenerationStandardGen2)).
+						HasNoResourceConstraint(),
+					resourceshowoutputassert.WarehouseShowOutput(t, warehouseModelStandardAndGeneration2.ResourceReference()).
+						HasType(sdk.WarehouseTypeStandard).
+						HasResourceConstraint(sdk.WarehouseResourceConstraintStandardGen2),
+				),
+			},
+			// remove resource constraint from config
```
sfc-gh-asawicki: nit: resource constraint -> generation

pkg/testacc/resource_warehouse_acceptance_test.go
```diff
@@ -2529,6 +2881,57 @@ func TestAcc_Warehouse_ResourceConstraint_MigrateManuallySetResourceConstraint(t
...
+		WithWarehouseTypeEnum(sdk.WarehouseTypeStandard)
+
+	resource.Test(t, resource.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: CheckDestroy(t, resources.Warehouse),
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders: ExternalProviderWithExactVersion("2.5.0"),
```
sfc-gh-asawicki: 2.6.0?

pkg/testacc/resource_warehouse_acceptance_test.go
```diff
@@ -2529,6 +2881,57 @@ func TestAcc_Warehouse_ResourceConstraint_MigrateManuallySetResourceConstraint(t
 	})
 }
 
+func TestAcc_Warehouse_Generation_MigrateManuallySetGeneration(t *testing.T) {
```
sfc-gh-asawicki, line 2916: please add test with no generation set in the config

pkg/testacc/resource_warehouse_acceptance_test.go
```diff
@@ -2462,12 +2709,115 @@ func TestAcc_Warehouse_ResourceConstraint_MixedWarehouseTypes(t *testing.T) {
 						planchecks.ExpectDrift(warehouseModelDefault.ResourceReference(), "show_output.0.resource_constraint", sdk.String(string(sdk.WarehouseResourceConstraintStandardGen1)), sdk.String(string(sdk.WarehouseResourceConstraintMemory16X))),
 						planchecks.ExpectChange(warehouseModelDefault.ResourceReference(), "resource_constraint", tfjson.ActionUpdate, sdk.String(string(sdk.WarehouseResourceConstraintMemory16X)), nil),
 						planchecks.ExpectChange(warehouseModelDefault.ResourceReference(), "warehouse_type", tfjson.ActionUpdate, sdk.String(string(sdk.WarehouseTypeSnowparkOptimized)), nil),
+						planchecks.ExpectNoChangeOnField(warehouseModelDefault.ResourceReference(), "generation"),
+						planchecks.ExpectComputed(warehouseModelDefault.ResourceReference(), r.ShowOutputAttributeName, true),
```
sfc-gh-asawicki, line 2737: maybe I missed it but do we have a test in which we externally change both the warehouse type and resource constraint? I would like to see what happens with the current logic then

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-jmichalak: Done, in other tests as well.

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/warehouse.go
sfc-gh-jmichalak: This was not needed in the approach before. Now, we follow a different approach - done.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: Done.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: Done.

pkg/resources/warehouse.go
sfc-gh-jmichalak: Warehouses created a long time ago, before this field was added, have an empty value here.

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-jmichalak: There already is - `TestAcc_Warehouse_Generation_MigrateStandardWithoutGeneration`.

pkg/testacc/resource_warehouse_acceptance_test.go
sfc-gh-jmichalak: Yes, we have such a step in `TestAcc_Warehouse_ResourceConstraint_MixedWarehouseTypes`, see `external change of the generation`.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: Done.

pkg/resources/warehouse.go
sfc-gh-jmichalak: Done.

pkg/resources/warehouse.go
sfc-gh-jmichalak: Reworked the logic entirely. Now we just handle these fields despite of the warehouse type. Additionally, we have to clear the generation and resource_constraint fields when they are changed.

pkg/acceptance/planchecks/expect_no_change_plan_check.go
sfc-gh-jmichalak: Yes, but this assertion only works if the plan is empty. When there is a diff on another field, this assertion does not check it properly. 

Note that the action is on the object-level, not the field-level (`change.Change.Actions`...), and in this plancheck we don't verify it. We only compare the before/after values.

As for the code duplication, IMO it should be done in SNOW-1473409, as it's noted in `expect_drift_plan_check.go`.

sfc-gh-asawicki: Okay, let's leave improvement for later, you are right that these planchecks require more attention overall

pkg/resources/warehouse.go
```diff
@@ -432,6 +453,13 @@ func GetReadWarehouseFunc(withExternalChangesMarking bool) schema.ReadContextFun
 		}
 
 		if withExternalChangesMarking {
+			generation, resourceConstraint := "", ""
```
sfc-gh-asawicki: nit: var instead of empty initialization

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -471,6 +482,45 @@ func TestInt_Warehouses(t *testing.T) {
...
+
+		returnedWarehouse, err = client.Warehouses.ShowByID(ctx, warehouse.ID())
+		require.NoError(t, err)
+		assertThatObject(t, objectassert.WarehouseFromObject(t, returnedWarehouse).
+			HasGeneration(sdk.WarehouseGenerationStandardGen2).
+			HasNoResourceConstraint().
+			HasType(sdk.WarehouseTypeStandard),
+		)
+
+		// TODO [SNOW-1473453]: uncomment and test when UNSET starts working correctly (expecting to unset to default STANDARD_GEN_1)
```
sfc-gh-asawicki, line 512: nit: I saw you added resource_constraint to this ticket, however, it's meant to be already fixed in 9.28, so before any other fix, let's have a dedicated one.

sfc-gh-asawicki (dismissed):

pkg/resources/warehouse.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jmichalak: Done.

PR #4004, opened 2025-09-04
chore: Fix warehouse long-running query test

sfc-gh-asawicki:
The assertion was awaiting, but the value was not refreshed from Snowflake.


sfc-gh-jmichalak (approved):

sfc-gh-asawicki:
The only failing test is connected with the session parameters and this is being worked on in a separate flow. Merging.

Issue #4001, opened 2025-09-03
[Bug]:Updating a task in snowflake by adding afters fails when pulumi up


dl-newton:
### Terraform CLI Version

1.5.7

### Terraform Provider Version

2.6.0

### Company Name

Newton

### Terraform Configuration

```terraform
This is just a translation, the issue was reported in the other issue :
provider "snowflake" {
  # creds from env vars
  role      = "ACCOUNTADMIN"
  warehouse = "COMPUTE_WH"
}

resource "snowflake_task" "parent" {
  name          = "PARENT"
  schedule      = "USING CRON 0 * * * * UTC"
  sql_statement = "select 1"
  warehouse     = "COMPUTE_WH"
}

resource "snowflake_task" "child" {
  name          = "CHILD"
  schedule      = "USING CRON 5 * * * * UTC"
  sql_statement = "select 1"
  warehouse     = "COMPUTE_WH"
  # This works on first create, but fails if you try to add it later
  after = [snowflake_task.parent.qualified_name]
}
```

### Category

category:snowflake

### Object type(s)

resource:task

### Expected Behavior

This is related to this issue : https://github.com/pulumi/pulumi-snowflake/issues/971

Here is the text recopied : 
### Describe what happened

Hi,
we are using Tasks in pulumi with [afters](https://www.pulumi.com/registry/packages/snowflake/api-docs/task/#afters_python).

The issue arises when we modify a task after its creation, it doesn't see the parameter and fails as its null.
See program below.
Its quiet urgent at the moment as we cant deploy tasks ATM. If there is a task that needs modfiication after, it breaks the whole deployment because of this.

We are interfacing with pulumi, but it's actually calling [terraform-provider-snowflake](https://github.com/snowflakedb/terraform-provider-snowflake)
The error arises at this [line](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/resources/diff_suppressions.go#L55) on the `terraform-provider-snowflake` repo:

```
    error: Detected that /Users/myUSER/.pulumi/plugins/resource-snowflake-v2.5.0/pulumi-resource-snowflake exited prematurely.
           This is *always* a bug in the provider. Please report the issue to the provider author as appropriate.
           To assist with debugging we have dumped the STDOUT and STDERR streams of the plugin:panic: can't use ElementIterator on null value [recovered]
    	panic: can't use ElementIterator on null value [recovered]
    	panic: can't use ElementIterator on null value

[...]

PR #4007, opened 2025-09-05
fix: Fix diff suppression for identifier set

sfc-gh-jmichalak:
Previously, when running `terraform plan -refresh=false`, the state of `task` was `null`, which caused panic during value extraction. Now, we check the value type and we log when it's `null` or an incorrect type.
- Fix diff suppression for identifier sets.
- Fix a few typos.

References #4001.

### DIFF
--- a/pkg/resources/diff_suppressions.go
+++ b/pkg/resources/diff_suppressions.go
@@ -50,6 +50,7 @@ func NormalizeAndCompareIdentifiersInSet(key string) schema.SchemaDiffSuppressFu
 			return false
 		}
 
+		// TODO(SNOW-2332640): Research if we can simplify this logic and keep this type safe.
 		if oldValue == "" && !d.GetRawState().IsNull() {
 			stateRaw, ok := d.GetRawState().AsValueMap()[key]
 			if ok && !stateRaw.IsNull() && stateRaw.Type().IsCollectionType() {
--- a/pkg/testacc/resource_task_acceptance_test.go
+++ b/pkg/testacc/resource_task_acceptance_test.go
@@ -1419,14 +1419,6 @@ func TestAcc_Task_WithAfter_issue4001(t *testing.T) {
 			},
 			{
 				ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
-				Config:                   config.FromModels(t, rootTaskConfigModel, childTaskConfigModelWithoutAfter),
-				Check: assertThat(t,
-					resourceassert.TaskResource(t, childTaskConfigModelWithAfter.ResourceReference()).
-						HasAfterEmpty(),
-				),
-			},
-			{
-				ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
 				Config:                   config.FromModels(t, rootTaskConfigModel, childTaskConfigModelWithAfter),
 				PlanOnly:                 true,
 				ExpectNonEmptyPlan:       true,


pkg/resources/diff_suppressions.go
```diff
@@ -52,14 +52,22 @@ func NormalizeAndCompareIdentifiersInSet(key string) schema.SchemaDiffSuppressFu
 		}
 
 		if oldValue == "" && !d.GetRawState().IsNull() {
-			if helpers.ContainsIdentifierIgnoringQuotes(ctyValToSliceString(d.GetRawState().AsValueMap()[key].AsValueSet().Values()), newValue) {
-				return true
+			stateRaw, ok := d.GetRawState().AsValueMap()[key]
```
sfc-gh-asawicki, line 55: Question: As I understand, the problem comes from an uninitialized list. Should we always initialize it to avoid having a null state? Should we at least document it for further research?

Question 2: Are there any other suppressors we should apply this logic to?

pkg/testacc/resource_task_acceptance_test.go
```diff
@@ -1373,6 +1373,54 @@ func TestAcc_Task_WithAfter(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				Config: config.FromModels(t, rootTaskConfigModel, childTaskConfigModelWithoutAfter),
+				Check: assertThat(t,
+					resourceassert.TaskResource(t, childTaskConfigModelWithAfter.ResourceReference()).
+						HasAfterEmpty(),
+				),
+			},
+			{
+				Config:             config.FromModels(t, rootTaskConfigModel, childTaskConfigModelWithAfter),
```
sfc-gh-asawicki, line 1415: We could have an additional middle step, use 2.6.0 in the first and the second steps, assert the error, and then in the third step show the fix, e.g.:
```
			{
				ExternalProviders: ExternalProviderWithExactVersion("2.6.0"),
				Config:            config.FromModels(t, rootTaskConfigModel, childTaskConfigModelWithoutAfter),
				Check: assertThat(t,
					resourceassert.TaskResource(t, childTaskConfigModelWithAfter.ResourceReference()).
						HasAfterEmpty(),
				),
			},
			{
				ExternalProviders:  ExternalProviderWithExactVersion("2.6.0"),
				Config:             config.FromModels(t, rootTaskConfigModel, childTaskConfigModelWithAfter),
				PlanOnly:           true,
				ExpectNonEmptyPlan: true,
				ExpectError:        regexp.MustCompile("can't use ElementIterator on null value"),
			},
```

sfc-gh-asawicki (changes requested):

pkg/resources/diff_suppressions.go
sfc-gh-jmichalak: 1. I think we could check if setting the default would be okay. On the other hand, the TF type system is quite brittle, so having more type assertions should help. I guess we can experiment a bit in functional tests (next PRs/TODOs). WDYT?  
2. I don't think so. I checked other calls to GetRawState and GetRawConfig, and they seem to be safe. Additionally, I deleted an unused function.

pkg/testacc/resource_task_acceptance_test.go
sfc-gh-jmichalak: 🤯 It did not work when I tested it with `		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,`, the panic was not recovered then.

pkg/resources/diff_suppressions.go
sfc-gh-asawicki: Ad 1 I am fine with leaving a TODO for future reseach as we have other priority topics atm.

Ad 2 Thanks

pkg/testacc/resource_task_acceptance_test.go
```diff
@@ -1395,23 +1395,39 @@ func TestAcc_Task_WithAfter_issue4001(t *testing.T) {
...
 			},
 			{
+				ExternalProviders:  ExternalProviderWithExactVersion("2.6.0"),
 				Config:             config.FromModels(t, rootTaskConfigModel, childTaskConfigModelWithAfter),
 				PlanOnly:           true,
 				ExpectNonEmptyPlan: true,
+				ExpectError:        regexp.MustCompile("can't use ElementIterator on null value"),
+			},
+			{
+				ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
```
sfc-gh-asawicki, line 1421: q: do we need this step? or could we just change the config while bumping? (I think I checked the latter locally during the previous review and it worked)

sfc-gh-jmichalak: Removed.

pkg/resources/diff_suppressions.go
sfc-gh-jmichalak: Created a ticket & linked it here.

PR #4016, opened 2025-09-09
fix: Fix integration and acceptance tests

sfc-gh-jmichalak:
Fix recent errors in integration and acceptance tests.
- Use REGISTER instead of ADD for application packages with enabled release channels (https://docs.snowflake.com/en/release-notes/bcr-bundles/2025_04/bcr-1977).
- Create application packages with disabled release channels for ADD and DROP integration tests.
- Update integration test for authentication policies (https://docs.snowflake.com/en/release-notes/bcr-bundles/2025_04/bcr-1971).
- Add cleanups for external tables integration tests (https://docs.snowflake.com/en/release-notes/bcr-bundles/2025_05/bcr-1989).
- Adjust a field assertion for listing integration test. This field is not used in any logic, and no change in resource is needed.
- Skip a DESC test for session policies. This object will be further reworked in SNOW-1348362, and the changes will be addressed there.
- Adjust the assertions for the user type (https://docs.snowflake.com/en/release-notes/bcr-bundles/2025_05/bcr-2067).
- Relax assertion ranges in service acceptance tests. The returned values are not deterministic.
- Adjust the setups and tests for granting privileges on shared databases - always grant USAGE before REFERENCE_USAGE.

## TODO
- Adjust the authentication policy resource (SNOW-2187814).
- Improve the BCR handling (SNOW-2330701).

### DIFF
--- a/docs/resources/account_role.md
+++ b/docs/resources/account_role.md
@@ -7,6 +7,8 @@ description: |-
 
 # snowflake_account_role (Resource)
 
+-> **Note** Users cannot execute a `DROP ROLE` command that drops the current primary role. An attempt to drop this role returns the following error: `Cannot drop role '<role_name>' because it is the current role for user '<username>'`. This means that before deleting the resource, make sure that you do not attempt to drop the current primary role. See [Usage notes](https://docs.snowflake.com/en/sql-reference/sql/drop-role#usage-notes) for more details.
+
 The resource is used for role management, where roles can be assigned privileges and, in turn, granted to users and other roles. When granted to roles they can create hierarchies of privilege structures. For more details, refer to the [official documentation](https://docs.snowflake.com/en/user-guide/security-access-control-overview).
 
 ## Example Usage
--- a/templates/resources/account_role.md.tmpl
+++ b/templates/resources/account_role.md.tmpl
@@ -11,6 +11,8 @@ description: |-
 
 # {{.Name}} ({{.Type}})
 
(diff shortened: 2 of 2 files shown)


pkg/testacc/resource_service_acceptance_test.go
```diff
@@ -546,8 +546,8 @@ func TestAcc_Service_basic_fromSpecification(t *testing.T) {
 						HasIsUpgrading(false).
 						HasManagingObjectDomainEmpty().
 						HasManagingObjectNameEmpty(),
-					assert.Check(resource.TestCheckResourceAttrWith(modelCompleteWithDifferentValues.ResourceReference(), "show_output.0.current_instances", customassert.BetweenFunc(0, 1))),
-					assert.Check(resource.TestCheckResourceAttrWith(modelCompleteWithDifferentValues.ResourceReference(), "show_output.0.target_instances", customassert.BetweenFunc(0, 1))),
+					assert.Check(resource.TestCheckResourceAttrWith(modelCompleteWithDifferentValues.ResourceReference(), "show_output.0.current_instances", customassert.BetweenFunc(0, 2))),
```
sfc-gh-asawicki, line 549: question: why we can have two here? is it from other test or what's the situation when it can happen?

sfc-gh-jmichalak: My bet is:
1. In the PreConfig, we set min=2.
2. This causes the number of instances to get bumped to 2.
3. In the resource, we have max=1, so this should be scaled down, but there may be some delay in the reconciliation.

PR #4029, opened 2025-09-15
feat: add semantic view resource - 1

sfc-gh-sarur:
Add a basic semantic view resource.
- Add a schema for the required fields.
- Do not enable the resource in the prod provider yet, as we expect a few more PRs before this resource is usable, and we don't want to block the releases.
Do not enable the resource in the test provider yet, as it can't be created in its current form. It requires either the metrics or dimensions block, which will be implemented next.
- Fix a typo in the Makefile.

## TODO
- Add the other supported fields & add this resource to the test provider.
- Handle external changes.
- Handle object renaming.
- Handle copy grants.
- Write more acceptance tests.
- Provide examples.
- Add an entry to the migration guide.
- Add a data source.
- Extract common code, especially for block handling.

### DIFF
--- a/pkg/testacc/resource_semantic_view_acceptance_test.go
+++ b/pkg/testacc/resource_semantic_view_acceptance_test.go
@@ -22,6 +22,9 @@ import (
 )
 
 func TestAcc_SemanticView_basic(t *testing.T) {
+	// TODO (next PR): unskip
+	t.Skip("Skipping semantic view acceptance test until we have metrics or dimensions block. It's required for the object creation.")
+
 	id := testClient().Ids.RandomSchemaObjectIdentifier()
 	comment, changedComment := random.Comment(), random.Comment()
 	table1, table1Cleanup := testClient().Table.Create(t)


pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var semanticViewsSchema = map[string]*schema.Schema{
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
```
sfc-gh-jmichalak, line 22: This shouldn't be ForceNew because this resource supports renaming. Pls delete this line and handle renaming in the Update function, just like e.g. https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/resources/listing.go#L224

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+		},
+	},
+	"relationships": {},
+	"facts":         {},
+	"dimensions":    {},
+	"metrics":       {},
+	"comment": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: blocklistedCharactersFieldDescription("Specifies a comment for the semantic view."),
```
sfc-gh-jmichalak, line 101: Let's remove these not implemented fields, including the comment and copy_grants. Also, let's add show_output and describe_output fields.

Regarding copy_grants: this is a tricky field and needs a special handling (see calling the `copyGrantsAttributeCreate` and `changedKeys` functions in the view resource as an example). IMO, this can be handled in the next PR.
- We should not use ForceNew, because the grants are lost, as the provider calls DROP and CREATE.
- Instead, we should use `orReplace` argument in the Create function, and call the Create function in Update.

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+func SemanticView() *schema.Resource {
+	deleteFunc := ResourceDeleteContextFunc(
+		sdk.ParseSchemaObjectIdentifier,
+		func(client *sdk.Client) DropSafelyFunc[sdk.SchemaObjectIdentifier] {
+			return client.SemanticViews.DropSafely
+		},
+	)
+	return &schema.Resource{
+		CreateContext: PreviewFeatureCreateContextWrapper(string(previewfeatures.SemanticViewResource), TrackingCreateWrapper(resources.SemanticView, CreateSemanticView)),
+		ReadContext:   PreviewFeatureReadContextWrapper(string(previewfeatures.SemanticViewResource), TrackingReadWrapper(resources.SemanticView, ReadGitRepository)),
```
sfc-gh-jmichalak:
```suggestion
		ReadContext:   PreviewFeatureReadContextWrapper(string(previewfeatures.SemanticViewResource), TrackingReadWrapper(resources.SemanticView, ReadSemanticView)),
```
Similarly, the line below.

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+		UpdateContext: PreviewFeatureUpdateContextWrapper(string(previewfeatures.SemanticViewResource), TrackingUpdateWrapper(resources.SemanticView, UpdateGitRepository)),
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.SemanticViewResource), TrackingDeleteWrapper(resources.SemanticView, deleteFunc)),
+		Description:   "Resource used to manage git repositories. For more information, check [git repositories documentation](https://docs.snowflake.com/en/sql-reference/sql/create-git-repository).",
```
sfc-gh-jmichalak: Pls update the description.

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.GitRepository, customdiff.All(
+			ComputedIfAnyAttributeChanged(gitRepositorySchema, ShowOutputAttributeName, "origin", "api_integration", "git_credentials", "comment"),
+		)),
+
+		Schema: gitRepositorySchema,
+		Importer: &schema.ResourceImporter{
+			StateContext: TrackingImportWrapper(resources.GitRepository, ImportName[sdk.SchemaObjectIdentifier]),
```
sfc-gh-jmichalak:
```suggestion
			StateContext: TrackingImportWrapper(resources.SemanticView, ImportName[sdk.SchemaObjectIdentifier]),
```

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+		ReadContext:   PreviewFeatureReadContextWrapper(string(previewfeatures.SemanticViewResource), TrackingReadWrapper(resources.SemanticView, ReadGitRepository)),
+		UpdateContext: PreviewFeatureUpdateContextWrapper(string(previewfeatures.SemanticViewResource), TrackingUpdateWrapper(resources.SemanticView, UpdateGitRepository)),
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.SemanticViewResource), TrackingDeleteWrapper(resources.SemanticView, deleteFunc)),
+		Description:   "Resource used to manage git repositories. For more information, check [git repositories documentation](https://docs.snowflake.com/en/sql-reference/sql/create-git-repository).",
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.GitRepository, customdiff.All(
+			ComputedIfAnyAttributeChanged(gitRepositorySchema, ShowOutputAttributeName, "origin", "api_integration", "git_credentials", "comment"),
+		)),
+
+		Schema: gitRepositorySchema,
```
sfc-gh-jmichalak:
```suggestion
		Schema: semanticViewSchema,
```

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+	)
+	return &schema.Resource{
+		CreateContext: PreviewFeatureCreateContextWrapper(string(previewfeatures.SemanticViewResource), TrackingCreateWrapper(resources.SemanticView, CreateSemanticView)),
+		ReadContext:   PreviewFeatureReadContextWrapper(string(previewfeatures.SemanticViewResource), TrackingReadWrapper(resources.SemanticView, ReadGitRepository)),
+		UpdateContext: PreviewFeatureUpdateContextWrapper(string(previewfeatures.SemanticViewResource), TrackingUpdateWrapper(resources.SemanticView, UpdateGitRepository)),
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.SemanticViewResource), TrackingDeleteWrapper(resources.SemanticView, deleteFunc)),
+		Description:   "Resource used to manage git repositories. For more information, check [git repositories documentation](https://docs.snowflake.com/en/sql-reference/sql/create-git-repository).",
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.GitRepository, customdiff.All(
+			ComputedIfAnyAttributeChanged(gitRepositorySchema, ShowOutputAttributeName, "origin", "api_integration", "git_credentials", "comment"),
```
sfc-gh-jmichalak: 1. resources.GitRepository -> resources.SemanticView
1. gitRepositorySchema -> semanticViewSchema
1. After adding the show_output, the string list here should contain the fields that can cause diff on show_output (for now, only the name, and tables).
1. Pls also add 
```
			ComputedIfAnyAttributeChanged(semanticViewSchema, FullyQualifiedNameAttributeName, "name"),
```

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+
+	logicalTableName := sdk.NewSchemaObjectIdentifier(databaseName, schemaName, name)
+	logicalTableRequest := sdk.NewLogicalTableRequest(logicalTableName)
+
+	return logicalTableRequest.
+		WithComment(c["comment"].(string)), nil
+}
+
+func getLogicalTableRequests(from interface{}) ([]sdk.LogicalTableRequest, error) {
+	cols := from.([]interface{})
```
sfc-gh-jmichalak: 1. Using `any` instead of `interface{}` is preferred.
1. We can change the arg type to `from []any` in the signature.
1. In general, each type cast should be made like:
```
v, ok := variable.(type)
if !ok {
// handle the error
}
```
We received a number of issues related to type assertion. This is a general remark, as in this case we can do (2).

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+				},
+				"synonym": {
+					Type:        schema.TypeList,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("List of synonyms for the logical table."),
+				},
+				"comment": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("Specifies a comment for the logical table."),
```
sfc-gh-jmichalak:
```suggestion
					Description:"Specifies a comment for the logical table.",
```

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+			Schema: map[string]*schema.Schema{
+				"table_alias": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("Specifies an alias for a logical table in the semantic view."),
+				},
+				"table_name": {
+					Type:        schema.TypeString,
+					Required:    true,
+					Description: blocklistedCharactersFieldDescription("Specifies an identifier for the logical table."),
```
sfc-gh-jmichalak: I believe this is a schema object identifier, so we need to have proper handling, pls add
```
		ValidateDiagFunc: IsValidIdentifier[sdk.SchemaObjectIdentifier](),
```

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+				},
+				"primary_key": {
+					Type:        schema.TypeList,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("Definitions of primary keys in the logical table."),
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
```
sfc-gh-jmichalak: I don't think we need to have a map here. This looks like a regular list, so it should be 
```
		Type: schema.TypeList,
		Elem: &schema.Schema{
		},
```
The same in the `unique` field.

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+		return diag.FromErr(errs)
+	}
+	return nil
+}
+
+func getLogicalTableRequest(from interface{}) (*sdk.LogicalTableRequest, error) {
+	c := from.(map[string]interface{})
+	name := c["name"].(string)
+	databaseName := c["database"].(string)
+	schemaName := c["schema"].(string)
```
sfc-gh-jmichalak: I'm confused. There are no such fields in the `tables` fields. The logical table name is passed as `table_name`, so we should parse it here with `ParseSchemaObjectIdentifier`

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,238 @@
...
+
+	logicalTableName := sdk.NewSchemaObjectIdentifier(databaseName, schemaName, name)
+	logicalTableRequest := sdk.NewLogicalTableRequest(logicalTableName)
+
+	return logicalTableRequest.
+		WithComment(c["comment"].(string)), nil
```
sfc-gh-jmichalak: This should be added conditionally, i.e. when this value is not empty.

pkg/provider/previewfeatures/preview_features.go
```diff
@@ -74,6 +74,8 @@ const (
 	ShareResource                                 feature = "snowflake_share_resource"
 	SharesDatasource                              feature = "snowflake_shares_datasource"
 	ParametersDatasource                          feature = "snowflake_parameters_datasource"
+	SemanticViewResource                          feature = "snowflake_semantic_view_resource"
+	SemanticViewDatasource                        feature = "snowflake_semantic_view_datasource"
```
sfc-gh-jmichalak:
```suggestion
```

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -0,0 +1,247 @@
+//go:build !account_level_tests
```
sfc-gh-jmichalak, line 1: 1. The acceptance tests are missing. Please create a file like `pkg/testacc/resource_semantic_view_acceptance_test.go`. You can take a look at the tests we have in `pkg/testacc/resource_git_integration_acceptance_test.go` as an inspiration.
2. This file shouldn't be added here, as it's already present in https://github.com/snowflakedb/terraform-provider-snowflake/pull/3978/files

pkg/acceptance/bettertestspoc/assert/objectassert/semantic_view_describe_snowflake_ext.go
```diff
@@ -0,0 +1,81 @@
+package objectassert
```
sfc-gh-jmichalak, line 1: Same situation as for the integration tests file (shouldn't be present here).

pkg/provider/previewfeatures/preview_features.go
sfc-gh-sarur: done

pkg/resources/semantic_views.go
sfc-gh-sarur: At the moment, we don't allow any semantic view properties to be updated other than the comment. ([doc](https://docs.snowflake.com/en/sql-reference/sql/alter-semantic-view))
So how do we handle the name in this case?

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go
sfc-gh-sarur: Agree for Primary Keys, but Unique keys is a list of lists, so it will need some nesting right? I've made an attempt at fixing it, ptal

pkg/resources/semantic_views.go
sfc-gh-sarur: I have removed the empty fields and copy_grants. 

I left the comments field since that is the only field for UpdateSemanticView to operate on. If I remove the comment field, UpdateSemanticView will essentially become a no-op. Please let me know if that is preferred for the first PR.

Ack, will handle copy_grants in the next PR

pkg/resources/semantic_views.go
sfc-gh-sarur: thank you for catching this, updated

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go
sfc-gh-sarur: You're right, fixed.

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_gen.go
```diff
@@ -0,0 +1,161 @@
...
+/////////////////////////////////////////////////
+// Basic builders (resource name and required) //
+/////////////////////////////////////////////////
+
+func SemanticView(
+	resourceName string,
+	database string,
+	schema string,
+	name string,
+	// tables ,
```
sfc-gh-sarur: The generator ran into an error when creating this file and aborted. I was able to run it using the `--dry-run --verbose` options and copy the content manually into this file. 
However, the tables attribute related stuff did not get generated, see line 126 in this file.
I commented out those rows temporarily to pass the precommits, but they are needed to be able to specify logical tables in the acceptance tests.

pkg/acceptance/bettertestspoc/assert/objectassert/semantic_view_describe_snowflake_ext.go
sfc-gh-sarur: done

sfc-gh-jmichalak:
Also, to clarify: please don't add the resource in the `provider/provider.go` file. We want to add this resource only in the testing setup of the provider. To do this:
- remove the changes from `preview_features.go`.
- add the `SemanticView()` to the testing provider setup ([location](https://github.com/snowflakedb/terraform-provider-snowflake/blob/422f2196886160b232209e8cafb1f0660ec16b69/pkg/testacc/1_setup_provider_test.go#L48)), like in the example [here](https://github.com/snowflakedb/terraform-provider-snowflake/blob/cf56dc6ffddd09f5973a7b7b2c21755d0547e0f7/pkg/acceptance/testing.go#L68).

pkg/resources/semantic_views.go
sfc-gh-jmichalak: SET supports only COMMENT, but we also have RENAME TO which changes the name. We can call RENAME TO first, and update the state, and we can continue with the SET option, just like in the example I linked.

pkg/resources/semantic_views.go
sfc-gh-jmichalak: Looks good (leaving unresolved because of the expected follow-ups).

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+	)
+	return &schema.Resource{
+		CreateContext: PreviewFeatureCreateContextWrapper(string(previewfeatures.SemanticViewResource), TrackingCreateWrapper(resources.SemanticView, CreateSemanticView)),
+		ReadContext:   PreviewFeatureReadContextWrapper(string(previewfeatures.SemanticViewResource), TrackingReadWrapper(resources.SemanticView, ReadSemanticView)),
+		UpdateContext: PreviewFeatureUpdateContextWrapper(string(previewfeatures.SemanticViewResource), TrackingUpdateWrapper(resources.SemanticView, UpdateSemanticView)),
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.SemanticViewResource), TrackingDeleteWrapper(resources.SemanticView, deleteFunc)),
+		Description:   "Resource used to manage semantic views. For more information, check [semantic views documentation](https://docs.snowflake.com/en/sql-reference/sql/create-semantic-view).",
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.SemanticView, customdiff.All(
+			ComputedIfAnyAttributeChanged(semanticViewsSchema, FullyQualifiedNameAttributeName, "name", "tables"),
```
sfc-gh-jmichalak:
```suggestion
			ComputedIfAnyAttributeChanged(semanticViewsSchema, FullyQualifiedNameAttributeName, "name"),
```

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+	return nil
+}
+
+func UpdateSemanticView(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+
+	id, err := sdk.ParseSchemaObjectIdentifier(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
```
sfc-gh-jmichalak, line 223:
```suggestion
	}
	if d.HasChange("name") {
		newId := sdk.NewSchemaObjectIdentifierInSchema(id.SchemaId(), d.Get("name").(string))

		err := client.SemanticViews.Alter(ctx, sdk.NewAlterSemanticViewRequest(id).WithRenameTo(&newId))
		if err != nil {
			return diag.FromErr(fmt.Errorf("error renaming semantic view %v err = %w", d.Id(), err))
		}

		d.SetId(helpers.EncodeResourceIdentifier(newId))
		id = newId
	}
```

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+			}
+		} else {
+			if err := client.SemanticViews.Alter(ctx, sdk.NewAlterSemanticViewRequest(id).WithUnsetComment(true)); err != nil {
+				d.Partial(true)
+				return diag.FromErr(err)
+			}
+		}
+	}
+	return ReadSemanticView(ctx, d, meta)
+}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: This will be addressed later during handling copy grants. In the meantime, the correct way is to mark tables and its subfields as ForceNew.

pkg/testacc/resource_semantic_view_acceptance_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+func TestAcc_SemanticView_basic(t *testing.T) {
+	id := testClient().Ids.RandomSchemaObjectIdentifier()
+	comment, changedComment := random.Comment(), random.Comment()
+
+	modelBasic := model.SemanticView(
+		"test",
+		id.DatabaseName(),
+		id.SchemaName(),
+		id.Name(),
```
sfc-gh-jmichalak, line 41: This one lacks the required argument for logical tables. To avoid repetition, we can define a helper logical table which can be reused, in `config/model/semantic_view_model_ext.go`.

sfc-gh-jmichalak (changes requested):

pkg/resources/semantic_views.go
sfc-gh-asawicki: We discussed that this is new, so we will leave it like this now. Later, we will add RENAME TO to the SDK and support the renaming. Please leave a TODO comment like this here in code:
```
// TODO [SNOW-2108211]: handle rename through ALTER ... RENAME TO
```

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: done

sfc-gh-jmichalak: Keeping this unresolved, as the optional fields in logical tables are missing. They can be added later.

pkg/testacc/resource_semantic_view_acceptance_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	modelCompleteWithDifferentValues := model.SemanticView(
+		"test",
+		id.DatabaseName(),
+		id.SchemaName(),
+		id.Name(),
+		[]sdk.LogicalTable{*logicalTable2},
+	).WithComment(changedComment)
+
+	resource.Test(t, resource.TestCase{
```
sfc-gh-jmichalak, line 61: This test is failing with
```
2025-09-25T02:46:30.1323528Z         Error: Unsupported argument
2025-09-25T02:46:30.1324590Z         
2025-09-25T02:46:30.1326190Z           on terraform_plugin_test.tf line 17, in resource "snowflake_semantic_view" "test":
2025-09-25T02:46:30.1327955Z           17:     database_name = "acc_test_db_AT_2B6CBE191E8B2CA05D333DF138FA352DE4618A9D"
2025-09-25T02:46:30.1329047Z         
2025-09-25T02:46:30.1330070Z         An argument named "database_name" is not expected here.
2025-09-25T02:46:30.1331069Z         
2025-09-25T02:46:30.1331963Z         Error: Unsupported argument
2025-09-25T02:46:30.1332895Z         
2025-09-25T02:46:30.1334113Z           on terraform_plugin_test.tf line 18, in resource "snowflake_semantic_view" "test":
2025-09-25T02:46:30.1335328Z           18:     logical_table_alias = "lt1"
2025-09-25T02:46:30.1336252Z         
2025-09-25T02:46:30.1337478Z         An argument named "logical_table_alias" is not expected here.
2025-09-25T02:46:30.1338499Z         
2025-09-25T02:46:30.1339409Z         Error: Unsupported argument
2025-09-25T02:46:30.1340321Z         
2025-09-25T02:46:30.1341537Z           on terraform_plugin_test.tf line 19, in resource "snowflake_semantic_view" "test":
2025-09-25T02:46:30.1343240Z           19:     schema_name = "acc_test_sc_AT_2B6CBE191E8B2CA05D333DF138FA352DE4618A9D"
2025-09-25T02:46:30.1350684Z         
2025-09-25T02:46:30.1351952Z         An argument named "schema_name" is not expected here.
2025-09-25T02:46:30.1578053Z --- FAIL: TestAcc_SemanticView_basic (0.17s)
```
Please adjust the field names in `WithTables` - they must match the fields in the resource definition:
- `database_name`, `schema_name` should be deleted and the table fully qualified name should be stored in `table_name`
- `logical_table_alias` should be renamed to `table_alias`.

pkg/resources/semantic_views.go
sfc-gh-jmichalak: I think the nested type also needs to have `Optional: true`, other than that it looks fine.

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+		Timeouts: defaultTimeouts,
+	}
+}
+
+func CreateSemanticView(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+	name := d.Get("name").(string)
+	schemaName := d.Get("schema").(string)
+	databaseName := d.Get("database").(string)
+	logicalTableRequests, err := getLogicalTableRequests(d.Get("logical_tables").([]interface{}))
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Please fix this.

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	logicalTableRequest := sdk.NewLogicalTableRequest(logicalTableName)
+
+	if c["comment"] != nil && c["comment"].(string) != "" {
+		return logicalTableRequest.
+			WithComment(c["comment"].(string)), nil
+	} else {
+		return logicalTableRequest, nil
+	}
+}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: +1

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+		return diag.FromErr(err)
+	}
+
+	semanticViewDetails, err := client.SemanticViews.Describe(ctx, id)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	errs := errors.Join(
```
sfc-gh-jmichalak, line 205: What we missed before was detecting external changes to logical tables. To do this, we need to parse the DESCRIBE result so that we use `d.Set` with a proper structure. Take a look at https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/resources/view.go#L731 for inspiration.

sfc-gh-jmichalak (changes requested):

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+		synonyms := v.GetSynonyms()
+		if synonyms != nil {
+			syns := make([]tfconfig.Variable, len(synonyms.WithSynonyms))
+			for _, synonym := range synonyms.WithSynonyms {
+				syns = append(syns, tfconfig.StringVariable(synonym.Synonym))
+			}
+			m["synonyms"] = tfconfig.ListVariable(syns...)
+		}
+		maps[i] = tfconfig.MapVariable(m)
```
sfc-gh-jcieslak: I believe this should be ObjectVariable rather than MapVariable

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+		logicalTableAlias := v.GetLogicalTableAlias()
+		if logicalTableAlias != nil {
+			m["logical_table_alias"] = tfconfig.StringVariable(logicalTableAlias.LogicalTableAlias)
+		}
+		primaryKeys := v.GetPrimaryKeys()
+		if primaryKeys != nil {
+			keys := make([]tfconfig.Variable, len(primaryKeys.PrimaryKey))
+			for _, key := range primaryKeys.PrimaryKey {
+				keys = append(keys, tfconfig.StringVariable(key.Name))
```
sfc-gh-jcieslak: nit: I think if you predefined the size above you should assign rather than append. Instead of `keys = append(keys, tfconfig.StringVariable(key.Name))` should be `keys[keyIdnex] = tfconfig.StringVariable(key.Name)`.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			m["primary_keys"] = tfconfig.ListVariable(keys...)
+		}
+		uniqueKeys := v.GetUniqueKeys()
+		if uniqueKeys != nil {
+			keys := make([]tfconfig.Variable, len(uniqueKeys))
+			for _, key := range uniqueKeys {
+				uniKeys := make([]tfconfig.Variable, len(key.Unique))
+				for _, uniKey := range key.Unique {
+					uniKeys = append(uniKeys, tfconfig.StringVariable(uniKey.Name))
```
sfc-gh-jcieslak: Same as above

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -0,0 +1,82 @@
...
+				}
+				keys = append(keys, tfconfig.ListVariable(uniKeys...))
+			}
+			m["unique_keys"] = tfconfig.ListVariable(keys...)
+		}
+		synonyms := v.GetSynonyms()
+		if synonyms != nil {
+			syns := make([]tfconfig.Variable, len(synonyms.WithSynonyms))
+			for _, synonym := range synonyms.WithSynonyms {
+				syns = append(syns, tfconfig.StringVariable(synonym.Synonym))
```
sfc-gh-jcieslak: Same here (as above)

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+				},
+				"primary_key": {
+					Type:        schema.TypeList,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("Definitions of primary keys in the logical table."),
+					Elem: &schema.Schema{
+						Type: schema.TypeString,
+					},
+				},
+				"unique": {
```
sfc-gh-jcieslak, line 65: Question: I know there may be multiple unique definitions in SQL, e.g.
```SQL
...
UNIQUE (a)
UNIQUE (b)
....
```
but does it differ from the below
```SQL
...
UNIQUE (a, b)
....
```
If not, please let's make it a single dimension **set** as in lists the order matters (if that's the case let's leave it as list).

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+				},
+				"table_name": {
+					Type:             schema.TypeString,
+					Required:         true,
+					Description:      blocklistedCharactersFieldDescription("Specifies an identifier for the logical table."),
+					ValidateDiagFunc: IsValidIdentifier[sdk.SchemaObjectIdentifier](),
+					DiffSuppressFunc: suppressIdentifierQuoting,
+				},
+				"primary_key": {
+					Type:        schema.TypeList,
```
sfc-gh-jcieslak, line 58: If the order doesn't matter let's make it TypeSet

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,268 @@
...
+	)
+	return &schema.Resource{
+		CreateContext: CreateSemanticView,
+		ReadContext:   ReadSemanticView,
+		UpdateContext: UpdateSemanticView,
+		DeleteContext: deleteFunc,
+		Description:   "Resource used to manage semantic views. For more information, check [semantic views documentation](https://docs.snowflake.com/en/sql-reference/sql/create-semantic-view).",
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.SemanticView, customdiff.All(
+			ComputedIfAnyAttributeChanged(semanticViewsSchema, FullyQualifiedNameAttributeName, "name"),
```
sfc-gh-jcieslak: As name field is marked as ForceNew this can be removed, but add `ComputedIfAnyAttributeChanged(semanticViewsSchema, ShowOutputAttributeName, "comment")` and `ComputedIfAnyAttributeChanged(semanticViewsSchema, DescribeOutputAttributeName, "comment")` (if there is a property in describe for sem views that describes comment property) instead

pkg/resources/semantic_views.go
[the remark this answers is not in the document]

pkg/resources/semantic_views.go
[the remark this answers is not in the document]

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: you're right, fixed it.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: done, but just wanted to confirm because another example (see external_volume_model_ext.go: line 29, 42 and 50) uses MapVariable to do this

pkg/resources/semantic_views.go
sfc-gh-sarur: done

pkg/resources/semantic_views.go
sfc-gh-sarur: done

pkg/resources/semantic_views.go
sfc-gh-sarur: Those 2 do behave differently, we should keep them as is

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go
```diff
@@ -0,0 +1,277 @@
...
+		aliasRequest := sdk.LogicalTableAliasRequest{LogicalTableAlias: c["table_alias"].(string)}
+		logicalTableRequest = logicalTableRequest.WithLogicalTableAlias(aliasRequest)
+	}
+
+	if c["comment"] != nil && c["comment"].(string) != "" {
+		logicalTableRequest = logicalTableRequest.WithComment(c["comment"].(string))
+	}
+
+	return logicalTableRequest, nil
+}
```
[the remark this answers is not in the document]
sfc-gh-sarur: done

pkg/resources/semantic_views.go
sfc-gh-sarur: Confirmed it with the team, primary keys should be a TypeList as the composite key order does matter

pkg/resources/semantic_views.go
sfc-gh-sarur: unique keys should be a TypeList as the composite key order does matter

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-jcieslak: Hmm, I guess there may be little to no difference in this case. The map and object are similar in tfconfig, which is why both may work here (there may be some differences, but with features we don't use here). As you changed it, let's leave the `object`, then.

pkg/resources/semantic_views.go
sfc-gh-jcieslak: Ok, Thanks for the clarification

pkg/resources/semantic_views.go
sfc-gh-jcieslak: Ok, Thanks for the clarification

sfc-gh-jcieslak (dismissed):

sfc-gh-jmichalak (dismissed):
LGTM. Let's leave other comments unresolved, as they need to be addressed later. I just reran the pipeline, let's wait for the results.

PR #4038, opened 2025-09-18
chore: Adjust warehouses after the changes in Snowflake

sfc-gh-jmichalak:
- Use UNSET instead of SET with a default value in the warehouse resource.
- Update the integration tests.

### DIFF
--- a/pkg/sdk/testint/warehouses_integration_test.go
+++ b/pkg/sdk/testint/warehouses_integration_test.go
@@ -452,7 +452,7 @@ func TestInt_Warehouses(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, sdk.WarehouseTypeSnowparkOptimized, returnedWarehouse.Type)
 		assert.Nil(t, returnedWarehouse.Generation)
-		assert.NotNil(t, returnedWarehouse.ResourceConstraint)
+		require.NotNil(t, returnedWarehouse.ResourceConstraint)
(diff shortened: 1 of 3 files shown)


pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -470,16 +470,16 @@ func TestInt_Warehouses(t *testing.T) {
...
-		// assert.Equal(t, *sdk.WarehouseResourceConstraintMemory1X, returnedWarehouse.ResourceConstraint)
+		alterOptions = &sdk.AlterWarehouseOptions{
+			Unset: &sdk.WarehouseUnset{ResourceConstraint: sdk.Bool(true)},
+		}
+		err = client.Warehouses.Alter(ctx, warehouse.ID(), alterOptions)
+		require.NoError(t, err)
+
+		returnedWarehouse, err = client.Warehouses.ShowByID(ctx, warehouse.ID())
+		require.NoError(t, err)
+		assert.NotNil(t, returnedWarehouse.ResourceConstraint)
```
sfc-gh-jcieslak: This should use `require`, not `assert`, as we will panic on the next line if this is not true

sfc-gh-jmichalak: Done, in other assertion as well.

sfc-gh-jcieslak (approved):

PR #4042, opened 2025-09-19
chore: Improve bcr handling in tests

sfc-gh-jmichalak:
BCRs are not handled consistently in our tests. In particular, bundle status is not cleaned up correctly. In this change:
- Add new system functions for getting BCR status.
- Add integration tests for handling BCRs.
- Improve testing helpers.

References:
- https://docs.snowflake.com/en/sql-reference/functions/system_show_active_behavior_change_bundles
- https://docs.snowflake.com/en/sql-reference/functions/system_behavior_change_bundle_status


pkg/sdk/system_functions.go
```diff
@@ -139,3 +141,60 @@ func (c *systemFunctions) DisableBehaviorChangeBundle(ctx context.Context, bundl
...
+		return "", fmt.Errorf("invalid behavior change bundle status: %s", s)
+	}
+	return BehaviorChangeBundleStatus(s), nil
+}
+
+func (c *systemFunctions) BehaviorChangeBundleStatus(ctx context.Context, bundle string) (BehaviorChangeBundleStatus, error) {
+	row := &struct {
+		StatusRaw string `db:"STATUS"`
+	}{}
+	sql := fmt.Sprintf(`SELECT SYSTEM$BEHAVIOR_CHANGE_BUNDLE_STATUS('%s') AS "STATUS"`, bundle)
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: This is used in tests only.

pkg/sdk/testint/system_functions_integration_account_level_integration_test.go
```diff
@@ -0,0 +1,53 @@
...
+// 2. The first bundle is always active by default.
+// 3. The second bundle is always not active by default.
+// 4. In this test we don't want to assert the bundle names as they may change. So, we simply assert that they are not empty.
+// 5. After each test, we clean up the bundle state by reverting the operation, i.e. by enabling the first bundle or disabling the second bundle.
+func TestInt_BcrBundles_AccountLevel(t *testing.T) {
+	client := testSecondaryClient(t)
+	ctx := testContext(t)
+
+	bundles := secondaryTestClientHelper().BcrBundles.ShowActiveBundles(t)
+	require.Len(t, bundles, 2)
```
sfc-gh-jcieslak: Is this assumption always true? It seems like something that could be easily broken

sfc-gh-asawicki: +1, when we were fighting with the new release 2 weeks ago, on one env we could see 3 bundles and on other just two

pkg/sdk/testint/system_functions_integration_account_level_integration_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	t.Run("show active bundles", func(t *testing.T) {
+		bundles, err := client.SystemFunctions.ShowActiveBehaviorChangeBundles(ctx)
+		require.NoError(t, err)
+		assert.Len(t, bundles, 2)
+		assert.NotEmpty(t, bundles[0].Name)
```
sfc-gh-asawicki: nit (for later): we could have object assertions generated for BehaviorChangeBundleInfo

pkg/acceptance/helpers/bcr_bundles_client_account_level.go
```diff
@@ -0,0 +1,78 @@
...
+	t.Helper()
+	ctx := context.Background()
+
+	return func() {
+		err := c.client().EnableBehaviorChangeBundle(ctx, name)
+		require.NoError(t, err)
+	}
+}
+
+func (c *BcrBundlesClient) getBcrInfo(t *testing.T, name string) sdk.BehaviorChangeBundleInfo {
```
sfc-gh-asawicki: nit: it's a read so it could be public and not in the account level test file

sfc-gh-jmichalak: Done.

pkg/sdk/testint/system_functions_integration_account_level_integration_test.go
sfc-gh-jmichalak: Yes, we can.

pkg/sdk/testint/system_functions_integration_account_level_integration_test.go
sfc-gh-jmichalak: Changed the condition to >= 2. There's no info in the docs about the exact number, so we can leave this for now.

sfc-gh-jcieslak: Yeah, I think that's better. One active bundle is still possible, but I'm guessing with much less probability (more likely we'll have more than less).

sfc-gh-jcieslak (approved):

PR #4043, opened 2025-09-19
chore: Improve imports handling in generators

sfc-gh-asawicki:
TODO


pkg/internal/genhelpers/imports_test.go
```diff
@@ -0,0 +1,346 @@
...
+}
+`
+
+		out, err := genhelpers.AddImports("", src)
+
+		require.NoError(t, err)
+		require.Equal(t, expected, string(out))
+	})
+
+	t.Run("add explicit named import", func(t *testing.T) {
```
sfc-gh-jcieslak, line 317: It seems some of the tests are repeated within this file, e.g., this one with L149 test.

pkg/internal/genhelpers/imports_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+func hello() {
+	r := re.Database
+	fmt.Println(r)
+}
+`
+
+		out, err := genhelpers.AddImports("", src)
+
+		require.NoError(t, err)
```
sfc-gh-jcieslak, line 343: Is there any case where we should expect errors being returned?

pkg/internal/genhelpers/templates/preamble.tmpl
```diff
@@ -0,0 +1,11 @@
+{{- /*gotype: github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/genhelpers.PreambleModel*/ -}}
+
+// Code generated by {{ .GeneratorName }} generator (v{{ .GeneratorVersion }}); DO NOT EDIT.
```
sfc-gh-jcieslak, line 3: Question: From now on, every modification to any generator should bump its version, right?

sfc-gh-jcieslak (approved):

pkg/internal/genhelpers/imports_test.go
sfc-gh-asawicki: Yes, they are. The first set was to experiment with the behavior, the second to verify our implementation is correct. The common setup will happen during `Extract common setup and helpers for the imports expeiment tests` listed in the PR description.

pkg/internal/genhelpers/imports_test.go
sfc-gh-asawicki: `imports.Process` can return error, so yes. I will add test for error being returned (in the follow-up PR).

pkg/internal/genhelpers/templates/preamble.tmpl
sfc-gh-asawicki: I wanted to discuss this and document what we agree. I will add it to our tomorrow agenda.

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/schema_resource_parameters_gen.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	resourceParameterAssert.AddAssertion(assert.ValueSet("parameters.#", "1"))
+	return &resourceParameterAssert
+}
+
+////////////////////////////
+// Parameter value checks //
+////////////////////////////
+
+func (s *SchemaResourceParametersAssert) HasDefaultDdlCollation(expected string) *SchemaResourceParametersAssert {
```
sfc-gh-jmichalak, line 40: Why is it only 1 param?

pkg/internal/genhelpers/imports_test.go
```diff
@@ -0,0 +1,346 @@
...
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/genhelpers"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/tools/imports"
+)
+
+// TODO [next PR - SNOW-2324252]: extract common setup, template for easier test writing
+func Test_ExperimentWithImportsProcess(t *testing.T) {
+	t.Run("add standard imports", func(t *testing.T) {
```
sfc-gh-jmichalak, line 13: Nit: use table tests.

pkg/acceptance/bettertestspoc/config/model/gen/model.go
```diff
@@ -37,8 +26,7 @@ type ResourceConfigBuilderAttributeModel struct {
 	OriginalType   schema.ValueType
 }
 
-func ModelFromResourceSchemaDetails(resourceSchemaDetails genhelpers.ResourceSchemaDetails) ResourceConfigBuilderModel {
-	additionalImports := make([]string, 0)
+func ModelFromResourceSchemaDetails(resourceSchemaDetails genhelpers.ResourceSchemaDetails, preamble *genhelpers.PreambleModel) ResourceConfigBuilderModel {
```
sfc-gh-jmichalak, line 29: Q: Why not just pass only the modelProvider to `NewGenerator` with preamble model embeeded before? I believe in `NewGenerator` we only call the modelProvider with the passed preamble, and the preamble itself is not used elsewhere.

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/schema_resource_parameters_gen.go
sfc-gh-asawicki: As it was defined like this in `objects_parameters_def` but wasn't generated. It's not the full definition and it should be extended, I added it to SNOW-1501905 and I will leave a TODO later

pkg/internal/genhelpers/imports_test.go
sfc-gh-asawicki: yes, part of `Extract common setup and helpers for the imports expeiment tests` listed in the PR description

pkg/acceptance/bettertestspoc/config/model/gen/model.go
sfc-gh-asawicki: But it is the modelProvider which constructs the model for each object, the preamble is not embedded earlier, that's why it's passed from the generator definition to the providerModel func (which does the embedding)

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/schema_resource_parameters_gen.go
sfc-gh-asawicki: Done in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4048

Issue #3959, opened 2025-08-25
[Bug]: Stage errors instead of marking resource as removed


NiklasRosenstein:
### Terraform CLI Version

1.10.0

### Terraform Provider Version

2.5.0

### Company Name

Helsing

### Terraform Configuration

```terraform
resource "snowflake_stage" "ingest" {
  name                = "S3_${local.name}_STAGE_NOLIST"
  url                 = var.url
  database            = var.database
  schema              = snowflake_schema.schema.name
  storage_integration = var.s3_int
  # HACK(@niklas.rosenstein): See https://github.com/snowflakedb/terraform-provider-snowflake/issues/3845
  file_format = "FORMAT_NAME = ${replace(snowflake_file_format.json.fully_qualified_name, "\"", "")}"
}
```

### Category

category:resource

### Object type(s)

resource:stage

### Expected Behavior

When the `snowflake_stage` object is missing from Snowflake, it should be marked as removed similar as is done in many other resource types:

```
╷
│ Warning: Failed to query stage. Marking the resource as removed.
│ 
```

### Actual Behavior

Instead, the provider errors:

```
╷
│ Error: [errors.go:22] object does not exist, or operation cannot be performed
│ 
│   with module.ingest_gitlab_raw.snowflake_stage.ingest,
│   on ../../modules/snowflake-s3-ingest-json/main.tf line 54, in resource "snowflake_stage" "ingest":
│   54: resource "snowflake_stage" "ingest" {
│ 
╵
```

### Steps to Reproduce

1. Create a `snowflake_stage` resource
2. Drop the stage directly in Snowflake
3. Try to run terraform apply again

### How much impact is this issue causing?

Low

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jcieslak:
Hey @NiklasRosenstein 👋 
I tested it on my side, and it worked properly. By the error message, I'm guessing you were removing something like a database or schema that we are not detecting in our read operation. It seems we are not using the "special" Show command that handles missing underlying database or schema. I'll update the stage implementation to use the safe show version, which should handle solve this issue.

NiklasRosenstein:
Thanks for the prompt reply @sfc-gh-jcieslak ! You're absolutely right, this is due to the schema having been removed.

[...]

Issue #3931, opened 2025-08-12
[Bug]: Dynamic_table resources crashes


liamjamesfoley:
### Terraform CLI Version

1.7.1

### Terraform Provider Version

2.5.0

### Company Name

Chime

### Terraform Configuration

```terraform
locals {
  # Flatten all table grants needed
  table_dependencies = toset(var.table_dependencies)

  # Extract unique database names from fully qualified table names
  unique_databases = toset([
    for table_path in var.table_dependencies :
    split(".", table_path)[0]
  ])

  # Extract unique schema names from fully qualified table names
  unique_schemas = toset([
    for table_path in var.table_dependencies :
    "${split(".", table_path)[0]}.${split(".", table_path)[1]}"
  ])
  schema_owner_role_name = upper("${var.database}__${var.schema}__OWNER")
}

resource "snowflake_dynamic_table" "table_name" {
  database   = var.database
  schema     = var.schema
  name       = var.name
  query      = var.query
  warehouse  = var.warehouse
  initialize = "ON_SCHEDULE" # Otherwise Terraform willl use the TF WH for the first run.

  target_lag {
    maximum_duration = var.target_lag_maximum_duration
  }
}

resource "snowflake_account_role" "service_role" {
  name = "SA__DYNAMIC_TABLE_SERVICE_ROLE_${var.database}_${var.schema}_${var.name}__ROLE"
}

resource "snowflake_grant_account_role" "subordinate_role_grant" {
  # Without this We'll get this error:
  # SQL execution error: Ownership restriction violation in a managed access schema. Grantee need to be a subordinate role of the schema owner.
  role_name        = snowflake_account_role.service_role.name
  parent_role_name = local.schema_owner_role_name
}

resource "snowflake_grant_ownership" "ownership" {
  account_role_name = snowflake_account_role.service_role.name
  on {
    object_type = "DYNAMIC TABLE"
    object_name = snowflake_dynamic_table.table_name.fully_qualified_name
  }
  depends_on = [snowflake_grant_account_role.subordinate_role_grant]
}

resource "snowflake_grant_privileges_to_account_role" "database_grants" {
  for_each          = local.unique_databases
  privileges        = ["USAGE"]
  account_role_name = snowflake_account_role.service_role.name
  on_account_object {
    object_type = "DATABASE"
    object_name = each.value
  }
}

resource "snowflake_grant_privileges_to_account_role" "schema_grants" {
[...]

PR #4045, opened 2025-09-22
fix: Bug fixes and documentation improvements

sfc-gh-jcieslak:
- Update documentation to include changes introduced as BCRs
- Update listing documentation to make it clear about current organization listing support
- Add object types and adjust valid object types used in privilege-granting resources
- Adjust stage read to use Safe SHOW method (fixes: #3959)
- Adjust dynamic table read to return an error whenever text column is empty (fixes: #3931)
- Test the adjustments with newly generated models (there were issues with generating dynamic tables, but solved it by using overrides)
- Document the changes in the migration guide

## TODO
- validate the object_types in grants and dynamic table query field are working as expected


templates/resources/listing.md.tmpl
```diff
@@ -19,6 +19,8 @@ description: |-
 
 !> **Warning** To use external resources in your manifest (e.g., company logo) you must be sourcing your manifest from a stage. Any references to external resources are relative to the manifest location in the stage.
 
+!> **Warning** Currently, this resource doesn't support [organization listings](https://docs.snowflake.com/en/user-guide/collaboration/listings/organizational/org-listing-about). We plan to add support for this in the future (through separate parameter or entirely new resource). There is a workaround for this provided in this issue: [#3982](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3982), but beware that in the next major version of the provider such workarounds won't be allowed.
```
sfc-gh-jmichalak: Nit: `may not be allowed`.

templates/resources/grant_ownership.md.tmpl
```diff
@@ -13,6 +13,10 @@ description: |-
 
 ~> **Note** Manage grants on `HYBRID TABLE` by specifying `TABLE` or `TABLES` in `object_type` field. This applies to a single object, all objects, or future objects. This reflects the current behavior in Snowflake.
 
+~> **Note** Changes introduced in [2025_01 Snowflake BCR Bundle](https://docs.snowflake.com/en/release-notes/bcr-bundles/2025_01/bcr-1870) alter how future ownership grant hierarchy works. After the change schema-level future ownership grant(s) will override those defined on the database-level. Previously, the database-level ones took the precedence.
+
+~> **Warning** Be careful when using this resource on managed schema or any object within it, as in Snowflake managed schemas follow different ownership and privilege model. For more details, please visit [Schema documentation](https://docs.snowflake.com/en/sql-reference/sql/create-schema#optional-parameters) (scroll to `WITH MANAGED ACCESS` option).
```
sfc-gh-jmichalak, line 18: Nit: can we have a separate example for handling managed objects?

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
```diff
@@ -0,0 +1,39 @@
...
+	d.WithDatabase(database)
+	d.WithSchema(schema)
+	d.WithName(name)
+	d.WithQuery(query)
+	d.WithWarehouse(warehouse)
+	return d
+}
+
+// TODO(SNOW-2357735): Remove after complex non-list type overrides are handled
+func (d *DynamicTableModel) WithTargetLag(_ []string) *DynamicTableModel {
```
sfc-gh-jmichalak: Why do we need this function? We can either remove it or here we can just pass `sdk.TargetLag`.

MIGRATION_GUIDE.md
```diff
@@ -26,6 +26,49 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ## v2.7.0 ➞ v2.7.1
 
+### *(bugfix)* Dynamic tables resource handling insufficient access and missing Text column
+
+Previously, when the `snowflake_dynamic_table` resource was created and the dynamic table's privileges were altered in a
+way that the current user lost access to view [`text` metadata field](https://docs.snowflake.com/en/user-guide/dynamic-tables-privileges#label-dynamic-tables-privileges-view-metadata),
+the resource thew internal error instead of handling the situation gracefully.
```
sfc-gh-jmichalak:
```suggestion
the resource threw an internal error instead of handling the situation gracefully.
```

MIGRATION_GUIDE.md
```diff
@@ -26,6 +26,49 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
...
+we had to adjust its reading function to handle the case when the stage is not present in Snowflake and should safely
+remove itself from the state.
+
+Now, when the stage is not present in Snowflake, it will be removed from the state without throwing an error (only an informational warning like in other resources).
+
+No changes in configuration and state are required.
+
+References: [#3959](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3959)
+
+### *(new feature)* Added missing object types in privilege-granting resources
```
sfc-gh-jmichalak, line 39: Please update the version header above.

sfc-gh-asawicki: +1

pkg/acceptance/bettertestspoc/assert/resourceassert/stage_resource_gen.go
```diff
@@ -0,0 +1,320 @@
+// Code generated by assertions generator; DO NOT EDIT.
```
sfc-gh-asawicki, line 1: this was not rebased over the newest dev (no version), please regenerate

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"log"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	tfconfig "github.com/hashicorp/terraform-plugin-testing/config"
+)
+
+// TODO(SNOW-2357735): Remove after complex non-list type overrides are handled
```
sfc-gh-asawicki: I think it's already part of SNOW-1501905, so let's not add a dedicated small task.

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
```diff
@@ -0,0 +1,39 @@
...
+	d := &DynamicTableModel{ResourceModelMeta: config.Meta(resourceName, resources.DynamicTable)}
+	d.WithDatabase(database)
+	d.WithSchema(schema)
+	d.WithName(name)
+	d.WithQuery(query)
+	d.WithWarehouse(warehouse)
+	return d
+}
+
+// TODO(SNOW-2357735): Remove after complex non-list type overrides are handled
```
sfc-gh-asawicki: as above

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
sfc-gh-asawicki: I ask in other way: we are able to generate constructor with []string, the With method is not generated but why can't we have here a proper implementation like in WithMaximumDurationTargetLag?

pkg/resources/dynamic_table.go
```diff
@@ -303,6 +303,16 @@ func ReadDynamicTable(ctx context.Context, d *schema.ResourceData, meta any) dia
 		return diag.FromErr(err)
 	}
 
+	if dynamicTable.Text == "" {
```
sfc-gh-asawicki, line 309: do we have test for this?

sfc-gh-asawicki: second question: should we error or warn here? if there is no text, we just cant handle external changes. I am not sure how did we handle this case in other secure resources but I would consider using warn here.

pkg/sdk/grants_validations.go
```diff
@@ -93,30 +95,35 @@ var validGrantToSchemaObjectTypes = []ObjectType{
...
+	ObjectTypeSnapshotPolicy,
+	ObjectTypeSnapshotSet,
 	ObjectTypeStage,
 	ObjectTypeStream,
+	ObjectTypeStreamlit,
 	ObjectTypeTable,
 	ObjectTypeTag,
 	ObjectTypeTask,
 	ObjectTypeView,
-	ObjectTypeStreamlit, // added because of https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/2656
```
sfc-gh-asawicki, line 118: question: are these two comments not valid anymore?

pkg/sdk/object_types.go
```diff
@@ -21,6 +21,7 @@ const (
 	ObjectTypeUser                 ObjectType = "USER"
 	ObjectTypeDatabaseRole         ObjectType = "DATABASE ROLE"
 	ObjectTypeDataset              ObjectType = "DATASET"
+	ObjectTypeDbtProject           ObjectType = "DBT PROJECT"
```
sfc-gh-asawicki, line 24: q: I remember talking about it some time ago: do we have in the backlog the task to have a single test verifying that all these can really be granted on?

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
sfc-gh-jcieslak: It's a matter of what can be generated and what cannot. As target_lag is a required field and it is a complex type (object), the generated constructor of the dynamic table doesn't know what to do with it, and doesn't specify a given type (it just skips it, e.g., `a string, b string, targetLag, c string`). On top of that, the target lag uses a method that is not generated, making it fail at compilation time. If the models weren't re-generated every time we call pre-push, I would adjust it manually and leave it with a well-defined constructor and compilable body, but as we do re-generate them often, I decided to make it compile and created `WithTargetLag` that shouldn't be used (making the generated constructors fail because they're invalid). Instead, I defined a constructor that properly handles what the generator cannot do atm and used it. It would be great with WithTargetLagValue could be used in generated constructor and generic tfconfig.Variable would be accepted, but it's not possible now.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Done

pkg/resources/dynamic_table.go
sfc-gh-jcieslak: I think the tests were somehow not pushed 🤔. Fixed the problem and pushed the test for the stage and dynamic table change. Well, I think the warn would work in this case, changing to warn.

pkg/sdk/grants_validations.go
sfc-gh-jcieslak: I guess they are still valid, but the difference now is that the Streamlit object type was not documented in Snowflake documentation as valid grantable type, and now it is. That's why I removed it (because it wasn't anything special anymore). If we want to keep such comments, I can bring those back.

pkg/acceptance/bettertestspoc/assert/resourceassert/stage_resource_gen.go
sfc-gh-jcieslak: Done

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
sfc-gh-jcieslak: Done, adjusted to SNOW-1501905

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
sfc-gh-jcieslak: same as above

pkg/sdk/object_types.go
sfc-gh-jcieslak: I couldn't find it, but created SNOW-2370066

templates/resources/grant_ownership.md.tmpl
sfc-gh-jcieslak: The examples would be the same, and in the case of managed schemas, I think what only changes is that the schema owner has additional privileges to act on objects within it. To be validated, but for now, I wouldn't add anything.

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
sfc-gh-asawicki: It does not address my comment.

pkg/sdk/grants_validations.go
sfc-gh-asawicki: If it's already properly documented, then the comment was invalid, good call removing it.

pkg/testacc/resource_dynamic_table_acceptance_test.go
```diff
@@ -480,3 +484,49 @@ func TestAcc_DynamicTable_issue3355_timeout(t *testing.T) {
...
+				),
+			},
+			{
+				PreConfig: func() {
+					testClient().Grant.GrantOwnershipToAccountRole(t, role.ID(), sdk.ObjectTypeDynamicTable, id)
+				},
+				Config:      accconfig.FromModels(t, dynamicTableModel),
+				ExpectError: regexp.MustCompile("the text is empty for dynamic table"),
+			},
+			// Step needed to clean up the ownership to correctly remove the dynamic table
```
sfc-gh-asawicki: let's say we fail on the previous step (e.g. expect error does not match), the dynamic table won't be cleaned up then, right? Should we leave a t.Cleanup for such a case that regrants and removes if exists?

sfc-gh-jcieslak: The test was ultimately removed as there was no way for me to properly show it behaving correctly. I'll test it manually in a second.

pkg/acceptance/bettertestspoc/config/model/dynamic_table_model_ext.go
sfc-gh-jcieslak: Changed according to the offline discussion

pkg/testacc/resource_dynamic_table_acceptance_test.go
sfc-gh-asawicki: Resolving but let's talk offline

sfc-gh-asawicki (approved):

MIGRATION_GUIDE.md
```diff
@@ -36,12 +36,55 @@ Additionally, in this change we dropped validating combinations of provider-spec
```
sfc-gh-jmichalak, line 36: <p class='graphite__hidden'><i>[Re: line +27]</i></p>
Nit: please change to `v2.7.x` before the release.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/4045?utm_source=unchanged-line-comment">Graphite</a>.</p>

sfc-gh-jmichalak (approved):

PR #4051, opened 2025-09-29
feat:  add semantic view resource - 2

sfc-gh-sarur:
Continuation of https://github.com/snowflakedb/terraform-provider-snowflake/pull/4029. 
- Add schema for the metrics block. 
- Adjust SDK to allow null values.
- Adjust DESC output setting.
-  Add this resource to the test provider & enable the acceptance test.

## TODO
- Add the other supported fields.
- Handle external changes.
- Handle object renaming.
- Handle copy grants.
- Write more acceptance tests.
- Provide examples.
- Add an entry to the migration guide.
- Add a data source.
- Extract common code, especially for block handling.


pkg/acceptance/bettertestspoc/config/model/semantic_view_model_gen.go
```diff
@@ -35,11 +36,14 @@ func SemanticView(
 	schema string,
 	name string,
 	tables []sdk.LogicalTable, // manually adjusted
+	metrics []sdk.MetricDefinition, // manually adjusted
 ) *SemanticViewModel {
 	s := &SemanticViewModel{ResourceModelMeta: config.Meta(resourceName, resources.SemanticView)}
 	s.WithDatabase(database)
 	s.WithSchema(schema)
 	s.WithName(name)
+	s.WithTables(tables)   // manually adjusted
```
sfc-gh-jmichalak, line 44: This is already done two lines below.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_gen.go
```diff
@@ -49,11 +53,14 @@ func SemanticViewWithDefaultMeta(
 	schema string,
 	name string,
 	tables []sdk.LogicalTable, // manually adjusted
+	metrics []sdk.MetricDefinition, // manually adjusted
 ) *SemanticViewModel {
 	s := &SemanticViewModel{ResourceModelMeta: config.DefaultMeta(resources.SemanticView)}
 	s.WithDatabase(database)
 	s.WithSchema(schema)
 	s.WithName(name)
+	s.WithTables(tables)   // manually adjusted
```
sfc-gh-jmichalak, line 60: Same here.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -54,6 +54,66 @@ func (s *SemanticViewModel) WithTables(tables []sdk.LogicalTable) *SemanticViewM
 	return s
 }
 
+func (s *SemanticViewModel) WithMetrics(metrics []sdk.MetricDefinition) *SemanticViewModel {
```
sfc-gh-jmichalak, line 58: This is missing from `resources/semantic_views.go`. Please add the `metrics` block & regenerate the builder files.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_gen.go
sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_gen.go
sfc-gh-sarur: got messed up when I resolved the conflicts somehow, fixed

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: done

pkg/resources/semantic_views.go
```diff
@@ -95,6 +95,85 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+		Required: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"semantic_expression": {
+					Type:        schema.TypeList,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("Specifies a semantic expression for a metric definition"),
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"qualified_expression_name": {
```
sfc-gh-jmichalak, line 117: Some fields are missing:
- a field for PRIVATE/PUBLIC values
- `<table_alias>` and `<fact_or_metric>` should be separate.

pkg/resources/semantic_views.go
```diff
@@ -95,6 +95,85 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+							"metric": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+							"over_clause": {
+								Type:     schema.TypeList,
+								Required: true,
+								Elem: &schema.Resource{
+									Schema: map[string]*schema.Schema{
+										"partition_by": {
```
sfc-gh-jmichalak, line 168: I think this should be split to `<exprs_using_dimensions_or_metrics>` and `EXCLUDING <dimensions>`.

pkg/resources/semantic_views.go
```diff
@@ -95,6 +95,85 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+											Type:        schema.TypeString,
+											Optional:    true,
+											Description: blocklistedCharactersFieldDescription("Specifies a partition by clause"),
+										},
+										"order_by": {
```
sfc-gh-jmichalak, line 173: This should be a block with more granular fields:
- <exprs_using_dimensions_or_metrics>
- ASC/DESC
- NULLS
  - FIRST
  - LAST

pkg/resources/semantic_views.go
```diff
@@ -95,6 +95,85 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
 	},
+	"metrics": {
+		Type:     schema.TypeList,
+		Required: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"semantic_expression": {
+					Type:        schema.TypeList,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("Specifies a semantic expression for a metric definition"),
```
sfc-gh-jmichalak: This and window_function should have set `ExactlyOneOf` field.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -54,6 +54,66 @@ func (s *SemanticViewModel) WithTables(tables []sdk.LogicalTable) *SemanticViewM
...
+			semExpVar := map[string]tfconfig.Variable{}
+			if semExp.Comment != nil {
+				semExpVar["comment"] = tfconfig.StringVariable(*semExp.Comment)
+			}
+			qExpName := semExp.GetQualifiedExpressionName()
+			if qExpName != nil {
+				qExpNameVar := map[string]tfconfig.Variable{
+					"QualifiedExpressionName": tfconfig.StringVariable(qExpName.QualifiedExpressionName),
+				}
+				semExpVar["qualifiedExpressionName"] = tfconfig.ObjectVariable(qExpNameVar)
```
sfc-gh-jmichalak: These fields should be lowercase - other places as well.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -54,6 +54,66 @@ func (s *SemanticViewModel) WithTables(tables []sdk.LogicalTable) *SemanticViewM
...
+			}
+			synonyms := semExp.GetSynonyms()
+			if synonyms != nil {
+				syns := make([]tfconfig.Variable, len(synonyms.WithSynonyms))
+				for j, synonym := range synonyms.WithSynonyms {
+					syns[j] = tfconfig.StringVariable(synonym.Synonym)
+				}
+				semExpVar["synonym"] = tfconfig.SetVariable(syns...)
+			}
+			m["semanticExpression"] = tfconfig.ObjectVariable(semExpVar)
```
sfc-gh-jmichalak: I believe this should be 
```suggestion
			m["semantic_expression"] = tfconfig.ListVariable(tfconfig.ObjectVariable(semExpVar))
```
Similarly below.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -54,6 +54,66 @@ func (s *SemanticViewModel) WithTables(tables []sdk.LogicalTable) *SemanticViewM
...
+			if qExpName != nil {
+				qExpNameVar := map[string]tfconfig.Variable{
+					"QualifiedExpressionName": tfconfig.StringVariable(qExpName.QualifiedExpressionName),
+				}
+				semExpVar["qualifiedExpressionName"] = tfconfig.ObjectVariable(qExpNameVar)
+			}
+			sqlExp := semExp.GetSqlExpression()
+			if sqlExp != nil {
+				sqlExpVar := map[string]tfconfig.Variable{
+					"SqlExpression": tfconfig.StringVariable(sqlExp.SqlExpression),
```
sfc-gh-jmichalak: This and qualified expressions are nested, but shouldn't be.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: Ahh I see, does it have to be wrapped in a list here because I used TypeList for the semantic_expression and window_function in the schema?

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: done

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go
sfc-gh-sarur: the public/private option is newly added and not part of the definition yet. We will have to add that in the next iteration.
Same for the table_alias and fact_or_metric. At present, the sdk only defines a field for `qualifiedExpressionName`. This will also need to be updated in the next iteration.
I have added some TODOs for this

pkg/resources/semantic_views.go
sfc-gh-sarur: The difference is because of the PRIVATE/PUBLIC field, which we don't support yet. As such, this can also be split in the next iteration.

pkg/resources/semantic_views.go
sfc-gh-sarur: Adding `ExactlyOneOf` seems to require the `MaxItems : 1` to be added not just to the nested fields, but also to the "metrics" field. But we need to be able to have multiple items in the metrics list. Is there a workaround?

pkg/resources/semantic_views.go
sfc-gh-sarur: this is also not yet supported in the SDK, it was newly added to the docs.

pkg/resources/semantic_views.go
```diff
@@ -95,6 +95,91 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+					MaxItems:    1,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"qualified_expression_name": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+							"sql_expression": {
+								Type:     schema.TypeString,
+								Required: true,
```
sfc-gh-jcieslak: Add missing description fields (could be done in a follow-up pr)

pkg/resources/semantic_views.go
```diff
@@ -95,6 +95,91 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+							"window_function": {
+								Type:        schema.TypeString,
+								Required:    true,
+								Description: blocklistedCharactersFieldDescription("The window function for the metric definition"),
+							},
+							"metric": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+							"over_clause": {
```
sfc-gh-jcieslak, line 162: Can there be more than 1 over clauses? If not let's add MaxItems: 1

pkg/resources/semantic_views.go
```diff
@@ -302,6 +391,60 @@ func getLogicalTableRequest(from any) (*sdk.LogicalTableRequest, error) {
 	return logicalTableRequest, nil
 }
 
+func getMetricDefinitionRequest(from any) (*sdk.MetricDefinitionRequest, error) {
+	c := from.(map[string]any)
+	metricDefinitionRequest := sdk.NewMetricDefinitionRequest()
+	if c["semantic_expression"] != nil {
```
sfc-gh-jcieslak: nit: A safer check would be to write `if len(c["semantic_expression"].([]any)) > 0` as it will check for null and will make sure there's at least one element in the list (but I think this would also work in this context)

pkg/resources/semantic_views.go
```diff
@@ -302,6 +391,60 @@ func getLogicalTableRequest(from any) (*sdk.LogicalTableRequest, error) {
...
+			}
+		}
+		return metricDefinitionRequest.WithSemanticExpression(*semExpRequest), nil
+	} else if c["window_function"] != nil {
+		windowFunctionDefinition := c["window_function"].([]any)[0].(map[string]any)
+		windowFunction := windowFunctionDefinition["window_function"].(string)
+		metric := windowFunctionDefinition["metric"].(string)
+		windowFuncRequest := sdk.NewWindowFunctionMetricDefinitionRequest(windowFunction, metric)
+		if windowFunctionDefinition["over_clause"] != nil {
+			overClause, ok := windowFunctionDefinition["over_clause"].(map[string]any)
```
sfc-gh-jcieslak: `over_clause` is a list of maps so you have to do similarly as above with window function `c["window_function"].([]any)[0].(map[string]any)`, depending on the number of allowed over clauses, if more than one is allowed then you should `range` over this list instead of taking the first element.

pkg/resources/semantic_views.go
```diff
@@ -302,6 +391,60 @@ func getLogicalTableRequest(from any) (*sdk.LogicalTableRequest, error) {
...
+				if overClause["partition_by"] != nil {
+					overClauseRequest = overClauseRequest.WithPartitionBy(overClause["partition_by"].(string))
+				}
+				if overClause["order_by"] != nil {
+					overClauseRequest = overClauseRequest.WithOrderBy(overClause["order_by"].(string))
+				}
+				if overClause["window_frame_clause"] != nil {
+					overClauseRequest = overClauseRequest.WithWindowFrameClause(overClause["window_frame_clause"].(string))
+				}
+				windowFuncRequest = windowFuncRequest.WithOverClause(*overClauseRequest)
```
sfc-gh-jcieslak, line 447: nit: The model functions are using pointer method receivers and are modifying the calling object, so you don't have to reassign the request, as the pointer you are calling with is the same as returned from `With` functions, which means you could turn this into: `windowFuncRequest.WithOverClause(*overClauseRequest)` (but it's not mandatory for now)

pkg/sdk/semantic_view_def.go
```diff
@@ -163,6 +163,8 @@ var qualifiedExpressionName = g.NewQueryStruct("QualifiedExpressionName").
 var semanticSqlExpression = g.NewQueryStruct("SemanticSqlExpression").
 	Text("SqlExpression", g.KeywordOptions().NoQuotes().Required())
 
+// TODO: add PUBLIC/PRIVATE optional field
```
sfc-gh-jcieslak: Todos should have ticket assigned to them (if you are planning to add them, you can use the same ticket number you are assigned to; can be done in the follow-up pr)

pkg/testacc/resource_semantic_view_acceptance_test.go
```diff
@@ -33,13 +30,19 @@ func TestAcc_SemanticView_basic(t *testing.T) {
...
+	metric1 := model.MetricDefinitionWithProps(semExp1, nil)
+	metric2 := model.MetricDefinitionWithProps(nil, windowFunc1)
 
 	modelBasic := model.SemanticView(
 		"test",
 		id.DatabaseName(),
 		id.SchemaName(),
 		id.Name(),
 		[]sdk.LogicalTable{*logicalTable1},
+		[]sdk.MetricDefinition{*metric1},
```
sfc-gh-jcieslak, line 56: No assertion updates? We need to add some to make sure all the fields are correctly mapped and used by Snowflake

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-jmichalak: Yes.

pkg/resources/semantic_views.go
sfc-gh-jmichalak: Ok, we can leave it for now, but let's keep this comment unresolved.

pkg/resources/semantic_views.go
sfc-gh-jmichalak: As above- let's keep this unresolved.

pkg/resources/semantic_views.go
sfc-gh-jmichalak: As above.

pkg/resources/semantic_views.go
sfc-gh-jmichalak: Oh, I think you're right. I believe this is just an SDK limitation, and in other places we have max items=1. So please make it clear in the `metrics` field description that these subfields are mutually exclusive.

sfc-gh-sarur: Understood, I will update the descriptions to clarify that

pkg/resources/semantic_views.go

pkg/resources/semantic_views.go
sfc-gh-sarur: there can not be multiple overclauses in one window function, I have added the max items limit

pkg/resources/semantic_views.go
sfc-gh-sarur: ack

pkg/resources/semantic_views.go
sfc-gh-sarur: ack

pkg/resources/semantic_views.go
sfc-gh-sarur: You're right, done

pkg/sdk/semantic_view_def.go
sfc-gh-sarur: updated the TODOs with ticket info

pkg/resources/semantic_views.go
sfc-gh-sarur: Done, there is only one over clause, so picking the first one is enough

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: These fields don't appear in the Show output, they appear in the Describe output, is there an example for how to add these assertions in the acceptance tests in this case?
P.S., they are currently tested in the integration tests

pkg/resources/semantic_views.go
sfc-gh-sarur: ack

pkg/resources/semantic_views.go
sfc-gh-sarur: ack

pkg/resources/semantic_views.go
```diff
@@ -95,6 +95,97 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+								Type:        schema.TypeString,
+								Optional:    true,
+								Description: blocklistedCharactersFieldDescription("Specifies a comment for the metric definition."),
+							},
+						},
+					},
+					ExactlyOneOf: []string{
+						"semantic_expression",
+						"window_function",
+					},
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Resolving, as it was already discussed above.

pkg/resources/semantic_views.go
```diff
@@ -95,6 +95,100 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				// TODO(SNOW-2396311): update the SDK with the newly added/updated fields for semantic expressions, then add them here
+				// TODO(SNOW-2396371): add PUBLIC/PRIVATE field
+				// TODO(SNOW-2398097): add table_alias
+				// TODO(SNOW-2398097): add fact_or_metric
+				"semantic_expression": {
+					Type:     schema.TypeList,
+					Optional: true,
+					Description: blocklistedCharactersFieldDescription("Specifies a semantic expression for a metric definition." +
```
sfc-gh-jmichalak, line 112: Nit: `blocklistedCharactersFieldDescription` is not needed here and in a few other places. We use it only for the string fields that are identifiers of other objects. Can be adjusted later.

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-jmichalak: We should add a couple of assertions.
- assertions for setting the fields in the state - generated `Has`
- assertions for values in Snowflake - check out e.g. compute pools - we have generated show output, but not desc output. In this case we need to iterate over the rows of DESC and match given values, like metric name, etc.

pkg/resources/semantic_views.go
```diff
@@ -156,10 +250,14 @@ func CreateSemanticView(ctx context.Context, d *schema.ResourceData, meta any) d
...
+		return diag.FromErr(err)
+	}
 	semanticViewName := sdk.NewSchemaObjectIdentifier(databaseName, schemaName, name)
 
-	request := sdk.NewCreateSemanticViewRequest(semanticViewName, logicalTableRequests)
+	request := sdk.NewCreateSemanticViewRequest(semanticViewName, logicalTableRequests).
+		WithSemanticViewMetrics(metricDefinitionRequests)
 	errs := errors.Join(
 		stringAttributeCreateBuilder(d, "comment", request.WithComment),
 	)
```
sfc-gh-jmichalak, line 263: 1. Pls use 
```
		d.Set(DescribeOutputAttributeName, schemas.SemanticViewDetailsToSchema(semanticViewDetails)),
```
2. This needs further adjustments in SDK. It looks like all columns in DESC except property and property_value can be null, so in `semantic_view_def.go`:
```
var semanticViewDetailsDbRow = g.DbStruct("semanticViewDetailsRow").
	OptionalText("object_kind").
	OptionalText("object_name").
	OptionalText("parent_entity").
	Text("property").
	Text("property_value")
```
and regenerate the sdk part

pkg/resources/semantic_views.go
```diff
@@ -302,6 +385,60 @@ func getLogicalTableRequest(from any) (*sdk.LogicalTableRequest, error) {
...
+			WithSqlExpression(semanticExpression["sql_expression"].(string))
+		semExpRequest := sdk.NewSemanticExpressionRequest(qualifiedExpNameRequest, sqlExpRequest)
+
+		if semanticExpression["comment"] != nil && semanticExpression["comment"].(string) != "" {
+			semExpRequest = semExpRequest.WithComment(semanticExpression["comment"].(string))
+		}
+
+		if semanticExpression["synonym"] != nil {
+			synonyms, ok := semanticExpression["synonym"].([]any)
+			if ok && len(synonyms) > 0 {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I believe, this is correct, the TypeSet's are mapped to *schema.Set not []any

pkg/resources/semantic_views.go
```diff
@@ -302,6 +397,60 @@ func getLogicalTableRequest(from any) (*sdk.LogicalTableRequest, error) {
...
+				var syns []sdk.Synonym
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Same here

pkg/resources/semantic_views.go
sfc-gh-sarur: ack

pkg/resources/semantic_views.go
sfc-gh-sarur: done

pkg/resources/semantic_views.go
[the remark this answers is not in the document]

pkg/resources/semantic_views.go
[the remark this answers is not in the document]
sfc-gh-sarur: done

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/assert/objectassert/semantic_view_describe_snowflake_ext.go
```diff
@@ -48,9 +48,7 @@ func (s *SemanticViewDetailsAssert) ContainsDetail(expected sdk.SemanticViewDeta
 	s.AddAssertion(func(t *testing.T, o *SemanticViewDetailsCollection) error {
 		t.Helper()
 		found := slices.ContainsFunc(o.Details, func(detail sdk.SemanticViewDetails) bool {
-			return detail.ObjectKind == expected.ObjectKind &&
-				detail.ObjectName == expected.ObjectName &&
-				detail.Property == expected.Property &&
+			return detail.Property == expected.Property &&
```
sfc-gh-jmichalak, line 51: We still should assert other fields.

sfc-gh-sarur: Ack, will fix that in the other PR to not block this one

sfc-gh-sarur: done in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4061/

PR #4061, opened 2025-10-04
Sarur snow 2176806 resource 3

sfc-gh-sarur:
## Test Plan

* [ ] acceptance tests

* [ ] …

## References

*


sfc-gh-jcieslak (commented):
You should create/update this PR's target branch to be your other PR, because now in this PR, I'm seeing changes you've made there (adding metrics). Please make sure only changes regarding relationships, dimensions, and facts (which I believe you're adding in this PR) are visible in the PR changes page.

pkg/testacc/resource_semantic_view_acceptance_test.go
```diff
@@ -72,7 +90,10 @@ func TestAcc_SemanticView_basic(t *testing.T) {
...
 		[]sdk.LogicalTable{*logicalTable2},
 		[]sdk.MetricDefinition{*metric2},
-	).WithComment(changedComment)
+	).WithComment(changedComment).
+		WithRelationships([]sdk.SemanticViewRelationship{*rel1}).
+		WithFacts([]sdk.SemanticExpression{*fact2}).
+		WithDimensions([]sdk.SemanticExpression{*dimension2})
 
 	resource.Test(t, resource.TestCase{
 		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
```
sfc-gh-jmichalak, line 128: <p class='graphite__hidden'><i>[Re: line +125]</i></p>
Please also add the assertions here.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/4061?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/resources/semantic_views.go
```diff
@@ -95,11 +95,119 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+			Schema: map[string]*schema.Schema{
+				"relationship_alias": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("Specifies an alias for the relationship."),
+				},
+				"table_name": {
+					Type:             schema.TypeString,
+					Optional:         true,
+					ValidateDiagFunc: IsValidIdentifier[sdk.SchemaObjectIdentifier](),
```
sfc-gh-jmichalak: The descriptions are missing. We are aiming to have a description in each field, if it is relevant.

pkg/resources/semantic_views.go
```diff
@@ -95,11 +95,119 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
 		},
 	},
-	"metrics": {
+	"relationships": {
 		Type:     schema.TypeList,
-		Required: true,
+		Optional: true,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"relationship_alias": {
```
sfc-gh-jmichalak: I'm struggling to understand the SQL mapping here. Maybe we should just use the names from the docs, like `<relationship_identifier`? Also, I can't see anything like `table_name`. cc @sfc-gh-jcieslak .

pkg/resources/semantic_views.go
```diff
@@ -258,6 +370,28 @@ func CreateSemanticView(ctx context.Context, d *schema.ResourceData, meta any) d
...
+		}
+		request.WithSemanticViewFacts(factsRequests)
+	}
+	if d.Get("dimensions") != nil {
+		dimensionsRequests, err := getSemanticExpressionRequests(d.Get("dimensions").([]any))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+		request.WithSemanticViewDimensions(dimensionsRequests)
+	}
```
sfc-gh-jmichalak, line 435: Nit: please use a custom wrapper and set these fields in errors.Join like below.

pkg/resources/semantic_views.go
sfc-gh-sarur: I will update the names to match the names on the docs as much as possible.

The `table_alias` field on the docs is in the sdk as `relationshipTableNameOrAlias` which has 2 fields, RelationshipTableName and RelationshipTableAlias. The customer can use either the table name (which would be a schema object identifier) or the table alias (a string). That's why I have split it into 2 separate fields here. It may be better if I contained them within a "table_name_or_alias" field here instead, so it becomes clear. I will make that change

pkg/resources/semantic_views.go

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: done

pkg/resources/semantic_views.go
sfc-gh-sarur: Can we handle this as a TODO instead?

pkg/resources/semantic_views.go
sfc-gh-jmichalak: Okay, but please don't wrap everything in `blocklistedCharactersFieldDescription`. It should be use ONLY when this field is an identifier of another object - can be fixed later.

pkg/sdk/semantic_view_gen_test.go
```diff
@@ -83,6 +84,10 @@ func TestSemanticViews_Create(t *testing.T) {
 						RelationshipTableName:  Pointer(randomSchemaObjectIdentifier()),
 						RelationshipTableAlias: String("alias"),
 					},
+					refTableNameOrAlias: &RelationshipTableAlias{
```
sfc-gh-jmichalak, line 104: This should be checked in a separate `t.Run` with a new `assertOptsInvalidJoinedErrors ` assertion only concerning `refTableNameOrAlias` because now, the assertion is still the same.

pkg/resources/semantic_views.go
```diff
@@ -95,11 +96,139 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+				"relationship_identifier": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: blocklistedCharactersFieldDescription("Specifies an optional identifier for the relationship."),
+				},
+				"table_name_or_alias": {
+					Type:     schema.TypeList,
+					Required: true,
+					MaxItems: 1,
+					Elem: &schema.Resource{
```
sfc-gh-jmichalak, line 118: Descriptions, please. And let's note that table_name and table_alias are mutually exclusive - in referenced_table_name_or_alias as well.

pkg/resources/semantic_views.go
sfc-gh-jmichalak: Yes, let's leave this open.

pkg/testacc/resource_semantic_view_acceptance_test.go
```diff
@@ -16,11 +16,12 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	_ "github.com/hashicorp/terraform-plugin-testing/plancheck"
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: @sfc-gh-sarur please resolve - remove the line with `_`.

pkg/testacc/resource_semantic_view_acceptance_test.go
```diff
@@ -49,30 +71,36 @@ func TestAcc_SemanticView_basic(t *testing.T) {
...
-		[]sdk.MetricDefinition{*metric2},
-	).WithComment(changedComment)
+		completeID.DatabaseName(),
+		completeID.SchemaName(),
+		completeID.Name(),
+		[]sdk.LogicalTable{*logicalTable1, *logicalTable2},
+		[]sdk.MetricDefinition{*metric1},
+	).WithComment(changedComment).
+		WithRelationships([]sdk.SemanticViewRelationship{*rel1}).
+		WithFacts([]sdk.SemanticExpression{*fact1}).
```
sfc-gh-jmichalak: Shouldn't these still be fact2 and dimension2? The name suggests the values should be different.

pkg/testacc/resource_semantic_view_acceptance_test.go
```diff
@@ -16,11 +16,12 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	_ "github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
 func TestAcc_SemanticView_basic(t *testing.T) {
-	id := testClient().Ids.RandomSchemaObjectIdentifier()
+	basicID, completeID := testClient().Ids.RandomSchemaObjectIdentifier(), testClient().Ids.RandomSchemaObjectIdentifier()
```
sfc-gh-jmichalak: Nit: why this change? we usually have one ID for such tests.

pkg/resources/semantic_views.go
sfc-gh-sarur: Done

pkg/resources/semantic_views.go
sfc-gh-sarur: created [SNOW-2405571](https://snowflakecomputing.atlassian.net/browse/SNOW-2405571) tracking it.

pkg/testacc/resource_semantic_view_acceptance_test.go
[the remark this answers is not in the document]
sfc-gh-sarur: done

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: When I used the same ID, the modelComplete was not getting updated with the additional relationships, facts and dimensions since we don't have the OrReplace yet. Is there a recommended way to do this?

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: Right now with alter commands we can only update the comment, so I only updated that field in the modelCompleteWithDifferentValues definition.

pkg/sdk/semantic_view_gen_test.go
sfc-gh-sarur: Done.

pkg/resources/semantic_views.go
sfc-gh-sarur: Done, I wrapped the name, schema, database and the tables.table_name field descriptions in blocklistedCharactersFieldDescription, removed everything else.

pkg/resources/semantic_views.go
sfc-gh-jmichalak: There's no need to create additional tickets like this: you can reference SNOW-2344309

pkg/resources/semantic_views.go
sfc-gh-jmichalak: Okay, makes sense. I think the syntax is not very clear in the SQL documentation.

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-jmichalak: Yes, it happens because the fields don't have `ForceNew`. Every field in the schema should be either handled in `Update` function (we can do this only for `comment`), OR have the `ForceNew` field.
So, please add the ForceNew fields, and the test issue should be fixed - it should also address https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/4061/feat-add-semantic-view-resource---3#comment-PRRC_kwDOCePWm86QOi7M

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-jmichalak: It should be handled by the ForceNew setting in the schema - check my other comment.

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: Understood, I have added the ForceNew field to all the fields except for comment and updated the tests to use the same ID. Also addressed the other comment to update the relationships, facts and dimensions in the `modelCompleteWithDifferentValues` def.

pkg/testacc/resource_semantic_view_acceptance_test.go
sfc-gh-sarur: done

pkg/resources/semantic_views.go
sfc-gh-sarur: Understood, I will do that going forward

pkg/resources/semantic_views.go
```diff
@@ -95,11 +97,157 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
+					},
+				},
+				"comment": {
+					Type:        schema.TypeString,
+					Optional:    true,
+					Description: "Specifies a comment for the dimension.",
+				},
+			},
+		},
+		AtLeastOneOf: []string{
```
sfc-gh-jcieslak, line 240: Question: Is this correct? In the SQL documentation, both are optional. If they're conditionally required (at least one of them), the SQL documentation should be slightly different imo.

pkg/testacc/resource_semantic_view_acceptance_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build !account_level_tests
+//go:build account_level_tests
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: I think this should "stay as before", but we actually changed the tags in acceptance tests, so this should be now `//go:build non_account_level_tests`

pkg/resources/semantic_views.go
sfc-gh-jmichalak: I think we discussed it earlier, as when you don't set any of these fields, the create fails, and that's why we initially disabled the tests.

sfc-gh-sarur: Yes, in the create statement we need at least a dimension or a metric defined, it is mentioned in the docs here : https://docs.snowflake.com/en/sql-reference/sql/create-semantic-view#parameters-for-facts-dimensions-and-metrics

sfc-gh-sarur:
@sfc-gh-jmichalak sorry, I missed your comment about waiting for #4096 before merging this one, I opened a revert PR here : https://github.com/snowflakedb/terraform-provider-snowflake/pull/4099

sfc-gh-jmichalak:
That's alright; I already commented on https://github.com/snowflakedb/terraform-provider-snowflake/pull/4099 and closed it.

PR #4073, opened 2025-10-08
chore: Add a new protected role

sfc-gh-jmichalak:
Add a new protected role.

TODO (next PR): Use this role in the Oauth authenticator test.


pkg/internal/snowflakeroles/snowflake_predefined_roles.go
```diff
@@ -11,6 +11,8 @@ var (
 	UserAdmin      = sdk.NewAccountObjectIdentifier("USERADMIN")
 	Public         = sdk.NewAccountObjectIdentifier("PUBLIC")
 	PentestingRole = sdk.NewAccountObjectIdentifier("PENTESTING_ROLE")
+	// RESTRICTED is a role that has no grants. It can be used instead of PUBLIC role, so that it's not granted to any users by default.
```
sfc-gh-jcieslak, line 14: nit: It's not PUBLIC role replacement, but rather it works contrary to it (granted to all vs granted to none by default). Saying this role can be used instead of PUBLIC role is a bit misleading.

sfc-gh-jmichalak: I'll address this later.

PR #4077, opened 2025-10-08
chore: Add new tests command

sfc-gh-jcieslak:
## Changes
- Prepare build verification tests for CI
  - Created a Dockerfile right with Docker Compose that installs the needed dependencies
  - Necessary secrets and environment variables will be specified in the CI
  - Added a new make command for tests (for now, only two tests are taken into consideration)

## Next changes
- Include more tests and address unresolved comments

### DIFF
--- a/packaging/docker-compose.yml
+++ b/packaging/docker-compose.yml
@@ -5,10 +5,6 @@ services:
       dockerfile: Dockerfile
     volumes:
       - ./../:/terraform-provider-snowflake
-    # TODO: Remove below?
-      - ~/.snowflake/config:/root/.snowflake/config
-    environment:
-      - TEST_SF_TF_TEST_OBJECT_SUFFIX=dummy_value_for_docker_compose
-      - TEST_SF_TF_GENERATED_RANDOM_VALUE=dummy_value_for_docker_compose
+      - ${SNOWFLAKE_CONFIG_PATH}:/root/.snowflake/config
     working_dir: /terraform-provider-snowflake
     command: [ "make", "test-build-verification" ]


packaging/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM artifactory.ci1.us-west-2.aws-dev.app.snowflake.com/docker-remote/library/ubuntu:18.04 AS base
+
+ENV GO_VERSION 1.23.6
+ENV TERRAFORM_VERSION 1.8.4
```
sfc-gh-jmichalak, line 4: Nit: can we have the latest version?

packaging/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM artifactory.ci1.us-west-2.aws-dev.app.snowflake.com/docker-remote/library/ubuntu:18.04 AS base
```
sfc-gh-jmichalak, line 1: I think we should not create more directories in the repository root, if possible. How about `tools` or `internal`?

Makefile
```diff
@@ -86,7 +86,13 @@ test-architecture: ## check architecture constraints between packages
 	go test ./pkg/architests/... -v
 
 test-acceptance-%: ## run acceptance tests (both non-account and account level ones) for the given resource only, e.g. test-acceptance-Warehouse
-	TF_ACC=1 TF_LOG=DEBUG SNOWFLAKE_DRIVER_TRACING=debug SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE=true SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags='non_account_level_tests,account_level_tests' -run ^TestAcc_$*_ -v -timeout=20m ./pkg/testacc
+	TF_ACC=1 TF_LOG=DEBUG SNOWFLAKE_DRIVER_TRACING=debug SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE=true SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags=non_account_level_tests,account_level_tests -run ^TestAcc_$*_ -v -timeout=20m ./pkg/testacc
+
+test-build-verification: ## run build verification tests
```
sfc-gh-asawicki, line 91: nit: build verification reads oddly. maybe something more like test-main-terraform-use-cases or something leke that?

Makefile
```diff
@@ -86,7 +86,13 @@ test-architecture: ## check architecture constraints between packages
...
+	TF_ACC=1 SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE=true TEST_SF_TF_REQUIRE_TEST_OBJECT_SUFFIX=1 TEST_SF_TF_REQUIRE_GENERATED_RANDOM_VALUE=1 SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags=non_account_level_tests,account_level_tests -run "^(TestAcc_Schema_basic|TestAcc_Schema_complete)$$" -v -cover -timeout=20m ./pkg/testacc
```
sfc-gh-asawicki, line 92: to be discussed: should we have SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE for these tests? - not sure, let's keep it like that fn

Makefile
```diff
@@ -86,7 +86,13 @@ test-architecture: ## check architecture constraints between packages
...
+
+test-build-verification-ci: ## run build verification tests for CI environment
```
sfc-gh-asawicki, line 94: nit: we have ci on gh actions but we are not using this command here, it is not meant to be run here. mby just -docker or -docker-compose suffix?

packaging/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+ENV GO_VERSION 1.23.6
```
sfc-gh-asawicki, line 3: nit: can we get this version from go.mod ?

packaging/Dockerfile
sfc-gh-asawicki: I'm not sure if the latest one is the right choice. We can discuss which

sfc-gh-jcieslak: Hmm, that's my bad, I actually wanted to use the same version we use in our GH pipelines. We use version 1.7.4, not 1.8.4. I'm changing it to be the same as there.

packaging/Dockerfile
sfc-gh-jcieslak: I don't know, for me, It didn't fit into any existing directories. That's why I created a new one (the naming could be different, but I don't it's that bad either).

- `tools` are for dev tool definitions (which ones are needed) and binaries (where to hold them after installation)
- `internal` is Go construct made for code visibility, it already contains tracking package, putting this next to tracking would feel weird

I don't think having one more top-level directory is that bad, especially because we don't have that many of them.

Makefile
sfc-gh-jcieslak: Hmm, sure, but I would like to adjust it in the follow-up as build-verification command is already in use

Makefile
sfc-gh-jcieslak: As above, I agree to the change, but I'll adjust it in the follow-up

packaging/Dockerfile
sfc-gh-jcieslak: Hmm, maybe it could. I tried a few approaches now, but could get it working. I'll put it as follow-up and try if time will allow.

PR #4084, opened 2025-10-13
feat: Add Notebook object to SDK

sfc-gh-mkowalski:
* Added [Notebook](https://docs.snowflake.com/en/sql-reference/commands-notebook) object to the SDK


pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+		OrReplace().
+		SQL("NOTEBOOK").
+		IfNotExists().
+		Name().
+		PredefinedQueryStructField("From", "*Location", g.ParameterOptions().SQL("FROM").NoQuotes().NoEquals()).
+		OptionalTextAssignment("MAIN_FILE", g.ParameterOptions().SingleQuotes()).
+		OptionalComment().
+		OptionalIdentifier("QueryWarehouse", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().SQL("QUERY_WAREHOUSE").Equals()).
+		OptionalNumberAssignment("IDLE_AUTO_SHUTDOWN_TIME_SECONDS", g.ParameterOptions().NoQuotes()).
+		OptionalIdentifier("Warehouse", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().SQL("WAREHOUSE").Equals()).
```
sfc-gh-jcieslak, line 111: How did you get the fields below Warehouse? We base our definitions on public documentation https://docs.snowflake.com/en/sql-reference/sql/create-notebook which is missing those fields

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+		OptionalIdentifier("ComputePool", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().SQL("COMPUTE_POOL").Equals()).
+		ListAssignment("EXTERNAL_ACCESS_INTEGRATIONS", g.KindOfT[AccountObjectIdentifier](), g.ParameterOptions().Parentheses()).
+		OptionalTextAssignment("RUNTIME_ENVIRONMENT_VERSION", g.ParameterOptions().SingleQuotes()).
+		OptionalTextAssignment("DEFAULT_VERSION", g.ParameterOptions().NoQuotes()).
+		WithValidation(g.ValidIdentifier, "name").
+		WithValidation(g.ValidIdentifierIfSet, "QueryWarehouse").
+		WithValidation(g.ValidIdentifierIfSet, "Warehouse").
+		WithValidation(g.ConflictingFields, "IfNotExists", "OrReplace").
+		WithValidation(g.ValidIdentifierIfSet, "ComputePool"),
+).AlterOperation(
```
sfc-gh-jcieslak, line 122: Remember that any inconsistencies within the documentation should be reported to the doc team (e.g. unset option is not present in the syntax section, but it's mentioned in sections below or not all possible options are listed)

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+				OptionalSQL("SECRETS").
+				OptionalSQL("WAREHOUSE").
+				OptionalSQL("RUNTIME_NAME").
+				OptionalSQL("COMPUTE_POOL").
+				OptionalSQL("EXTERNAL_ACCESS_INTEGRATIONS").
+				OptionalSQL("RUNTIME_ENVIRONMENT_VERSION"),
+			g.ListOptions().NoParentheses().SQL("UNSET"),
+		).
+		WithValidation(g.ValidIdentifier, "name").
+		WithValidation(g.ExactlyOneValueSet, "Set", "Unset"),
```
sfc-gh-jcieslak: Add RenameTo option here

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+).DropOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/drop-notebook",
+	g.NewQueryStruct("DropNotebook").
+		Drop().
+		SQL("NOTEBOOK").
+		IfExists().
```
sfc-gh-jcieslak, line 173: Another inconsistency with the documentation, does it work in the worksheet?

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).DescribeOperation(
+	g.DescriptionMappingKindSingleValue,
+	"https://docs.snowflake.com/en/sql-reference/sql/desc-notebook",
```
sfc-gh-jcieslak, line 178: It seems there another inconsistency the output section only shows one column is returned, I'm guessing by the definition that it's not true and they return the same output, right?

pkg/sdk/notebooks_gen_test.go
```diff
@@ -0,0 +1,251 @@
...
+		var opts *AlterNotebookOptions = nil
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
+	})
+	t.Run("validation: valid identifier for [opts.name]", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.name = emptySchemaObjectIdentifier
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("validation: exactly one field from [opts.Set opts.Unset] should be present", func(t *testing.T) {
```
sfc-gh-jcieslak: in cases where we have `exactly one of` validation, we test two cases: where there're more than one field set and when none is set. You're missing the second one, please add it. Ultimately, we should modify the generator to produce two test cases instead of one for this validation rule (but later on as generator improvement, not this pr).

pkg/sdk/notebooks_impl_gen.go
```diff
@@ -0,0 +1,170 @@
...
+
+func (r notebooksRow) convert() (*Notebook, error) {
+	return &Notebook{
+		CreatedOn:      r.CreatedOn,
+		Name:           r.Name,
+		DatabaseName:   r.DatabaseName,
+		SchemaName:     r.SchemaName,
+		Comment:        &r.Comment.String,
+		Owner:          r.Owner,
+		QueryWarehouse: &AccountObjectIdentifier{r.QueryWarehouse.String},
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: That's valid approach, or just assign notebook instead of returning it immediately, handle the optional fields, and return the notebook at the end.
Similar to:
```go
n := &Notebook{ 
  // handling required fields
}

// handling optional fields
if r.QueryWarehouse.Valid {
  n.QueryWarehouse = &AccountObjectIdentifier{r.QueryWarehouse.String}
}

return n
```

pkg/sdk/notebooks_validations_gen.go
```diff
@@ -0,0 +1,95 @@
...
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	if everyValueSet(opts.IfNotExists, opts.OrReplace) {
+		errs = append(errs, errOneOf("CreateNotebookOptions", "IfNotExists", "OrReplace"))
+	}
+	if opts.ComputePool != nil && !ValidObjectIdentifier(opts.ComputePool) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	// Validation added manually.
+	if opts.IdleAutoShutdownTimeSeconds != nil && !validateIntInRangeInclusive(*opts.IdleAutoShutdownTimeSeconds, 60, 259200) {
```
sfc-gh-jcieslak: We try to limit such validations and let Snowflake validate them. The reason is that some limitations could be changed per customer, which means some of the values may sometimes be set outside of the regular range and still be valid.

pkg/sdk/notebooks_validations_gen.go
```diff
@@ -0,0 +1,95 @@
...
+			errs = append(errs, ErrInvalidObjectIdentifier)
+		}
+		if opts.Set.Warehouse != nil && !ValidObjectIdentifier(opts.Set.Warehouse) {
+			errs = append(errs, ErrInvalidObjectIdentifier)
+		}
+		if opts.Set.ComputePool != nil && !ValidObjectIdentifier(opts.Set.ComputePool) {
+			errs = append(errs, ErrInvalidObjectIdentifier)
+		}
+		// Validation added manually.
+		if opts.Set.IdleAutoShutdownTimeSeconds != nil && !validateIntInRangeInclusive(*opts.Set.IdleAutoShutdownTimeSeconds, 60, 259200) {
```
sfc-gh-jcieslak: Same here, I would remove it.

pkg/sdk/notebooks_def.go
sfc-gh-mkowalski: I got it from manually testing the object; It seems that the documentation is faulty.

pkg/sdk/notebooks_def.go
sfc-gh-mkowalski: As we discussed during the daily I will report the inconsistencies in one collective message after I perform integration testing.

pkg/sdk/notebooks_def.go
sfc-gh-mkowalski: Yes it works.

pkg/sdk/notebooks_validations_gen.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	if opts.Warehouse != nil && !ValidObjectIdentifier(opts.Warehouse) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
+	if everyValueSet(opts.IfNotExists, opts.OrReplace) {
+		errs = append(errs, errOneOf("CreateNotebookOptions", "IfNotExists", "OrReplace"))
+	}
+	if opts.ComputePool != nil && !ValidObjectIdentifier(opts.ComputePool) {
+		errs = append(errs, ErrInvalidObjectIdentifier)
+	}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: We can add this manually.

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+			"Unset",
+			g.NewQueryStruct("NotebookUnset").
+				OptionalSQL("COMMENT").
+				OptionalSQL("QUERY_WAREHOUSE").
+				OptionalSQL("SECRETS").
+				OptionalSQL("WAREHOUSE").
+				OptionalSQL("RUNTIME_NAME").
+				OptionalSQL("COMPUTE_POOL").
+				OptionalSQL("EXTERNAL_ACCESS_INTEGRATIONS").
+				OptionalSQL("RUNTIME_ENVIRONMENT_VERSION"),
```
sfc-gh-jmichalak: Add `WithValidation(g.AtLeastOneValueSet...`with all Unset fields.

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+			g.ListOptions().NoParentheses().SQL("UNSET"),
+		).
```
sfc-gh-jmichalak, line 162:
```
		OptionalSetTags().
		OptionalUnsetTags().
```
is missing.

pkg/sdk/notebooks_def.go
sfc-gh-jmichalak: And "SetTags" and "UnsetTags".

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+				OptionalQueryStructField("SecretsList", functionSecretsListWrapper, g.ParameterOptions().SQL("SECRETS").Parentheses()).
+				OptionalTextAssignment("MAIN_FILE", g.ParameterOptions().SingleQuotes()).
+				OptionalIdentifier("Warehouse", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().SQL("WAREHOUSE").Equals()).
+				OptionalTextAssignment("RUNTIME_NAME", g.ParameterOptions().SingleQuotes()).
+				OptionalIdentifier("ComputePool", g.KindOfT[AccountObjectIdentifier](), g.IdentifierOptions().SQL("COMPUTE_POOL").Equals()).
+				ListAssignment("EXTERNAL_ACCESS_INTEGRATIONS", g.KindOfT[AccountObjectIdentifier](), g.ParameterOptions().Parentheses()).
+				OptionalTextAssignment("RUNTIME_ENVIRONMENT_VERSION", g.ParameterOptions().SingleQuotes()).
+				WithValidation(g.ValidIdentifierIfSet, "QueryWarehouse").
+				WithValidation(g.ValidIdentifierIfSet, "Warehouse").
+				WithValidation(g.ValidIdentifierIfSet, "ComputePool"),
```
sfc-gh-jmichalak: Add `WithValidation(g.AtLeastOneValueSet...`

pkg/sdk/notebooks_def.go
sfc-gh-jmichalak: The outputs are slightly different. So, the structs should be separate for both SHOW and DESC.

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,132 @@
...
+	g.KindOfT[SchemaObjectIdentifier](),
+).CreateOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/create-notebook",
+	g.NewQueryStruct("CreateNotebook").
+		Create().
+		OrReplace().
+		SQL("NOTEBOOK").
+		IfNotExists().
+		Name().
+		PredefinedQueryStructField("From", "*Location", g.ParameterOptions().SQL("FROM").NoQuotes().NoEquals()).
```
sfc-gh-jmichalak: Why `NoQuotes`? In docs it's with quotes.

pkg/sdk/notebooks_gen_test.go
```diff
@@ -0,0 +1,251 @@
...
+		opts := defaultOpts()
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterNotebookOptions", "Set", "Unset"))
+
+		opts.Set = &NotebookSet{
+			Comment: String("comment"),
```
sfc-gh-jmichalak: Nit: for future:
```
opts.Set = &NotebookSet{}
opts.Unset = &NotebookUnset{}
```
would be enough.

pkg/sdk/notebooks_gen_test.go
```diff
@@ -0,0 +1,251 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("basic", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.Set = &NotebookSet{
+			Comment: String("comment"),
+		}
+		assertOptsValidAndSQLEquals(t, opts, "ALTER NOTEBOOK %s SET COMMENT = 'comment'", id.FullyQualifiedName())
+	})
```
sfc-gh-jmichalak, line 158: We should also validate that the empty set/unset fails. Check `validation: at least one of the fields...` in other objects.

pkg/sdk/notebooks_impl_gen.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func (r *DescribeNotebookRequest) toOpts() *DescribeNotebookOptions {
+	opts := &DescribeNotebookOptions{
+		name: r.name,
+	}
+	return opts
+}
+
+func (r notebooksRow) convert() (*Notebook, error) {
```
sfc-gh-jmichalak: As I noted above, we should have separate structs for show/desc and a different `convert` function.

pkg/sdk/notebooks_gen_test.go
```diff
@@ -0,0 +1,251 @@
...
+			SecretsList:                 &SecretsList{[]SecretReference{{"var_name", true, SchemaObjectIdentifier{}}}},
+			MainFile:                    String("main_file"),
+			Warehouse:                   &AccountObjectIdentifier{"sample_wh"},
+			RuntimeName:                 String("runtime_name"),
+			ComputePool:                 &AccountObjectIdentifier{"sample_cp"},
+			ExternalAccessIntegrations:  []AccountObjectIdentifier{{"test"}},
+			RuntimeEnvironmentVersion:   String("WH-RUNTIME-2.0"),
+		}
+
+		assertOptsValidAndSQLEquals(t, opts, "ALTER NOTEBOOK %s SET COMMENT = 'comment' QUERY_WAREHOUSE = \"sample_qwh\" IDLE_AUTO_SHUTDOWN_TIME_SECONDS = 3600 SECRETS = ('var_name' =) MAIN_FILE = 'main_file' WAREHOUSE = \"sample_wh\" RUNTIME_NAME = 'runtime_name' COMPUTE_POOL = \"sample_cp\" EXTERNAL_ACCESS_INTEGRATIONS = (\"test\") RUNTIME_ENVIRONMENT_VERSION = 'WH-RUNTIME-2.0'", id.FullyQualifiedName())
```
sfc-gh-jmichalak: Something is wrong here: `SECRETS = ('var_name' =)`. I think we should
- set a valid SchemaObjectIdentifier above
- add a validatoin (unfortunately, manually) for this field, and add a separate test case checking this validation.

sfc-gh-jmichalak (changes requested):

pkg/sdk/notebooks_def.go
sfc-gh-mkowalski: I looked up how it was done in a definition of [another object](https://docs.snowflake.com/en/sql-reference/sql/create-listing).

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
```diff
@@ -148,6 +148,11 @@ var allStructs = []SdkObjectDef{
 		ObjectType:   sdk.ObjectTypeStorageIntegration,
 		ObjectStruct: sdk.StorageIntegration{},
 	},
+	{
```
sfc-gh-jcieslak: generate object asserts in the next step (I think you accidentally committed it)

pkg/sdk/notebooks_def.go
sfc-gh-jcieslak: That's okay, but before resource implementation, let's discuss if we want to use them there, because as I mentioned above, we mostly focus on publicly documented API. (you can resolve after reading it)

pkg/sdk/notebooks_def.go
sfc-gh-jcieslak: RenameTo seems to be still missing from this validation

pkg/sdk/notebooks_gen_test.go
```diff
@@ -91,24 +92,28 @@ func TestNotebooks_Alter(t *testing.T) {
...
-		}
-		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterNotebookOptions", "Set", "Unset"))
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterNotebookOptions", "Set", "Unset", "SetTags", "UnsetTags"))
+	})
+
+	t.Run("validation: exactly one field from [opts.Set opts.Unset opts.SetTags opts.UnsetTags] should be present (none)", func(t *testing.T) {
+		opts := defaultOpts()
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterNotebookOptions", "Set", "Unset", "SetTags", "UnsetTags"))
+
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterNotebookOptions", "Set", "Unset", "SetTags", "UnsetTags"))
```
sfc-gh-jcieslak: two same asserts

pkg/sdk/notebooks_gen_test.go
```diff
@@ -91,24 +92,28 @@ func TestNotebooks_Alter(t *testing.T) {
...
 		opts := defaultOpts()
 		opts.name = emptySchemaObjectIdentifier
 		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
 	})
 
-	t.Run("validation: exactly one field from [opts.Set opts.Unset] should be present", func(t *testing.T) {
+	t.Run("validation: exactly one field from [opts.Set opts.Unset opts.SetTags opts.UnsetTags] should be present (too many)", func(t *testing.T) {
 		opts := defaultOpts()
-		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterNotebookOptions", "Set", "Unset"))
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterNotebookOptions", "Set", "Unset", "SetTags", "UnsetTags"))
```
sfc-gh-jcieslak: remove this assert (in this test case, we care more about the second one). This case is tested / asserted in the test case below.

pkg/sdk/notebooks_impl_gen.go
```diff
@@ -144,19 +150,68 @@ func (r *DescribeNotebookRequest) toOpts() *DescribeNotebookOptions {
...
+		Name:                        r.Name,
+		DefaultVersion:              r.DefaultVersion,
+		DefaultVersionName:          r.DefaultVersionName,
+		DefaultVersionLocationUri:   r.DefaultVersionLocationUri,
+		LastVersionName:             r.LastVersionName,
+		LastVersionLocationUri:      r.LastVersionLocationUri,
+	}
+
+	// Optionals.
+	if r.Title.Valid {
```
sfc-gh-jcieslak: FYI: we also have helpers for such mappings in the `mapping_helpers.go` in the sdk package for more compact mappings. If you can, please use them or extend them, because I'm not sure if we have ones for identifiers. Also to create identifier, we usually have NewXObjectIdentifier functions or ParseXObjectIdentifier (in case there's multi-part identifier like schema object identifier, but it's returned as one field).

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
sfc-gh-asawicki: Yes, please remove it from here in this PR.

pkg/sdk/notebooks_def.go
```diff
@@ -0,0 +1,200 @@
+package sdk
+
+import g "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/poc/generator"
+
+//go:generate go run ./poc/main.go
+
+var notebookDbRow = g.DbStruct("notebooksRow").
```
sfc-gh-asawicki: I think that we usually name them using singular, not plural

pkg/sdk/notebooks_def.go
sfc-gh-asawicki: +1

pkg/sdk/notebooks_gen_test.go
```diff
@@ -0,0 +1,273 @@
...
+		opts.QueryWarehouse = &AccountObjectIdentifier{"sample_qwh"}
+		opts.IdleAutoShutdownTimeSeconds = Int(3600)
+		opts.Warehouse = &AccountObjectIdentifier{"sample_wh"}
+		opts.RuntimeName = String("sample")
+		opts.ComputePool = &AccountObjectIdentifier{"sample_cp"}
+		opts.ExternalAccessIntegrations = []AccountObjectIdentifier{}
+		opts.RuntimeEnvironmentVersion = String("WH-RUNTIME-2.0")
+		opts.DefaultVersion = String("FIRST")
+
+		assertOptsValidAndSQLEquals(t, opts, "CREATE NOTEBOOK IF NOT EXISTS %s FROM @%s/dir/subdir MAIN_FILE = 'main_file' COMMENT = 'comment' QUERY_WAREHOUSE = \"sample_qwh\" IDLE_AUTO_SHUTDOWN_TIME_SECONDS = 3600 WAREHOUSE = \"sample_wh\" RUNTIME_NAME = 'sample' COMPUTE_POOL = \"sample_cp\" RUNTIME_ENVIRONMENT_VERSION = 'WH-RUNTIME-2.0' DEFAULT_VERSION = FIRST", id.FullyQualifiedName(), stageId.FullyQualifiedName())
```
sfc-gh-asawicki: nit: if you wrap string using "`", then you don;t need to escape double quotes.

pkg/sdk/notebooks_def.go
sfc-gh-mkowalski: Git repository uses plural: `gitRepositoryDbRow = g.DbStruct("gitRepositoriesRow")`

pkg/sdk/notebooks_def.go
sfc-gh-jmichalak: In the listings' documentation, still it's wrapped in `'`.

pkg/sdk/notebooks_impl_gen.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func (r notebooksRow) convert() (*Notebook, error) {
+	return &Notebook{
+		CreatedOn:    r.CreatedOn,
+		Name:         r.Name,
+		DatabaseName: r.DatabaseName,
+		SchemaName:   r.SchemaName,
+		Comment:      &r.Comment.String,
+		Owner:        r.Owner,
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Please adjust.

pkg/sdk/notebooks_gen_test.go
sfc-gh-jmichalak: It looks better, but for the future: use `sdk.New*Identifier` functions instead of initializing a struct.

sfc-gh-jmichalak (commented):
I resolved my comments, but +1 to the other ones.

pkg/sdk/notebooks_def.go
sfc-gh-asawicki: So it should be changed also there (but not in this PR)

pkg/sdk/notebooks_def.go
sfc-gh-mkowalski: Yes, but its definition uses `NoQuotes`:
`PredefinedQueryStructField("From", g.KindOfTPointer[Location](), g.ParameterOptions().NoQuotes().NoEquals().SQL("FROM"))`

pkg/sdk/notebooks_def.go
sfc-gh-mkowalski: I discussed it with @sfc-gh-jmichalak and we concluded there's no harm in providing it in the SDK; we should come back to it when implementing the resources.

pkg/sdk/notebooks_def.go
sfc-gh-jmichalak: So, I think that should also be with single quotes. Maybe the documentation has changed. In terms of quoting we should follow the documentation. This is a non-blocking comment, tho.

sfc-gh-jmichalak (dismissed):

pkg/sdk/notebooks_gen_test.go
```diff
@@ -0,0 +1,270 @@
...
+	t.Run("validation: exactly one field from [opts.Set opts.Unset opts.SetTags opts.UnsetTags] should be present", func(t *testing.T) {
+		opts := defaultOpts()
+
+		opts.Set = &NotebookSet{}
+		opts.Unset = &NotebookUnset{}
+
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterNotebookOptions", "Set", "Unset", "SetTags", "UnsetTags"))
+	})
+
+	t.Run("validation: exactly one field from [opts.Set opts.Unset opts.SetTags opts.UnsetTags] should be present", func(t *testing.T) {
```
sfc-gh-jmichalak, line 111: It's actually better to have unique test names. When there are two, only the first one is run. This can be addressed in the next PR.

sfc-gh-jmichalak (approved):

PR #4085, opened 2025-10-13
chore: Add the note about terraform providers command to the namespace migration guides

sfc-gh-jmichalak:
Add the note about the `terraform providers` command to the namespace migration guides.

### DIFF
--- a/CZI_UPGRADE.md
+++ b/CZI_UPGRADE.md
@@ -41,6 +41,6 @@ If you are not pinning your provider versions, you may find it useful to forcefu
 terraform init -upgrade
 ```
 
-You can verify the required providers with their namespaces and versions in your deployments with the [terraform providers](https://developer.hashicorp.com/terraform/cli/commands/providers) command.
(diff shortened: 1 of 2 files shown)


CZI_UPGRADE.md
```diff
@@ -41,4 +41,6 @@ If you are not pinning your provider versions, you may find it useful to forcefu
 terraform init -upgrade
 ```
 
->**Note**:  0.34 is the first version published after the transfer. When the provider was transferred over not all of the older releases were transferred. Only versions 0.28 and newer were transferred. If you are using a version older than 0.28, it is highly recommended to upgrade to a newer version.
\ No newline at end of file
+You can verify the required providers with their namespaces and versions in your deployments with the [terraform providers](https://developer.hashicorp.com/terraform/cli/commands/providers) command.
```
sfc-gh-asawicki: Let's add why (i.e. to verify if all namespaces in the current working directory have been replaced). I wonder if we should link 3945 as it's quite long, and the conlusion is to verify it. So maybe without linking, wdyt?

sfc-gh-jmichalak: Done. I didn't want to link the issue because it's quite long and contains two separate topics.

PR #4111, opened 2025-10-20
chore: Saving test results

sfc-gh-jcieslak:
## Changes
- Add a script that internally runs the test to process its JSON output and:
  - Output to the console any entries of type `output`
  - Save as CSV entry any entries of type `pass` or `fail`
- Address some of the https://github.com/snowflakedb/terraform-provider-snowflake/pull/4077 comments
- Added a short README on usage

### DIFF
--- a/pkg/scripts/test_output_processor/README.md
+++ b/pkg/scripts/test_output_processor/README.md
@@ -23,7 +23,7 @@ go test --tags=non_account_level_tests,account_level_tests -run TestAcc_GrantPri
 1> ./pkg/scripts/test_output_processor/output.csv
 ```
 
-The script parts were grouped by new liens for clarity:
+The script parts were grouped by new lines for clarity:
 1. We set the necessary environment variables for the test run.
 2. We run the `go test` command with the required tags, test name, verbosity, timeout, and JSON output.
 3. We pipe (notice the `|`) the JSON output to the `test_output_processor.go` script.
--- a/pkg/scripts/test_output_processor/test_output_processor.go
+++ b/pkg/scripts/test_output_processor/test_output_processor.go
@@ -29,6 +29,10 @@ func main() {
 		}
 		testResults = append(testResults, entry)
 	}
+	if err := scanner.Err(); err != nil {
+		fmt.Printf("error reading input: %v", err)
+		os.Exit(1)
+	}
 
 	csvWriter := csv.NewWriter(os.Stdout)
 


pkg/scripts/test_runner/test_runner.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+		}
+	}
+
+	return resultingBuffer
+}
+
+type SyncBuffer struct {
+	buf    *bytes.Buffer
+	reader *bufio.Reader
```
sfc-gh-jmichalak: Is the reader actually needed?

pkg/scripts/test_runner/test_runner.go
```diff
@@ -0,0 +1,221 @@
...
+
+	testResults := runTest(mappedTestType)
+	if err := processAndSaveTestResults(testWorkflowId, testResults, outputPath); err != nil {
+		log.Fatalf("Failed to processed the test results, err = %s", err)
+	}
+
+	log.Println("Successfully processed test results")
+}
+
+type TestResultEntry struct {
```
sfc-gh-jmichalak: IMO this file mixes too many concepts. We could split it for:
- parsing cmd input
- handling test results

pkg/scripts/test_runner/test_runner.go
```diff
@@ -0,0 +1,221 @@
...
+				}
+			}
+			break loop
+		default:
+			text, _ := reader.ReadBytes('\n')
+
+			var entry map[string]string
+			_ = json.Unmarshal(text, &entry)
+
+			if v, ok := entry["Output"]; ok && v != "" {
```
sfc-gh-jmichalak: I think that currently it will work only for `test-main-terraform-use-cases`, as other targets don't use a JSON output.

pkg/scripts/test_runner/test_runner.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+	}
+
+	return nil
+}
+
+func runTest(testType TestType) *bytes.Buffer {
+	cmd := exec.Command("make", TestMakeCommand[testType])
+
+	buf := NewSyncBuffer()
```
sfc-gh-jmichalak: I'm curious if we actually need async parsing. The regular test-level logs weigh ~5MB, so maybe we should start by parsing the logs sequentially.

pkg/scripts/test_runner/test_runner.go
```diff
@@ -0,0 +1,221 @@
...
+	header := []string{"TEST_RUN_ID", "PACKAGE", "TEST", "ACTION", "ELAPSED"}
+	if err := csvWriter.Write(header); err != nil {
+		return fmt.Errorf("failed to write CSV header: %w", err)
+	}
+
+	// Write test results
+	for _, result := range testResults {
+		if result.Test != "" && (result.Action == "pass" || result.Action == "fail") {
+			record := []string{
+				id,
```
sfc-gh-jmichalak: The ID is consistent across all the records. Is it needed? Maybe we could just store it in the file name.

pkg/scripts/test_runner/test_runner.go
sfc-gh-jcieslak: Are you sure? I think it's much more, for e.g., acceptance tests (remember that JSON format is a bit more verbose than the regular one), but maybe it would be manageable to parse sequentially; however, if we already have a working async solution, is it worth making it synchronous? 🤔

pkg/scripts/test_runner/test_runner.go
sfc-gh-jcieslak: Yes, but that's just a matter of adding `-json` flag to other commands. I already changed them in the PR, which adds saving test results.

pkg/scripts/test_runner/test_runner.go
sfc-gh-jcieslak: As far as I remember, it is, because it holds its state about the current position, etc., which has to be different from the buffer's state.

pkg/scripts/test_runner/test_runner.go
sfc-gh-jcieslak: The format of CSV has to match the format of the table we would like to upload to, which is why it is repeated and cannot be stored in the filename.

pkg/scripts/test_runner/test_runner.go
```diff
@@ -0,0 +1,221 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"slices"
+	"sync"
+)
+
+type TestType string
+
+const (
```
sfc-gh-asawicki: why do we need all these options when we'll be running only acceptance tests?

pkg/scripts/test_runner/test_runner.go
sfc-gh-asawicki: But for now we care only about the subset of acceptance tests ~100.

pkg/scripts/test_runner/test_runner.go
sfc-gh-asawicki: But why would we want to add json to other commands?

pkg/scripts/test_runner/test_runner.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+func ToTestType(s string) (TestType, error) {
+	if slices.Contains(AllTestTypes, TestType(s)) {
+		return TestType(s), nil
+	}
+	return "", fmt.Errorf("unknown test type: %s", s)
+}
+
+func main() {
```
sfc-gh-asawicki: I am not sure if we should have this logic in a test runner in our repo and not directly in the pipeline. Let's talk.

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #4113, opened 2025-10-21
chore: Adjust database acceptance tests to set the common testing pattern

sfc-gh-jcieslak:
## Changes
Adjust database tests to set an example on a common testing pattern for main Terraform use cases for a given object. This PR only includes the testing pattern for resources that represent an object on the Snowflake side. See the next PRs section on checking other tests, as they may differ from the ones set here. By basing on a few acceptance tests already within the project, I ended up with 

~~two tests (BasicUseCase and CompleteUseCase; the naming may be confusing because it refers more to the initial configuration rather than overall complexity, but the naming can be adjusted later on). Both are pretty minimal, but they validate basic CRUD + import + external changes functionality of a resource, which should be enough for our use case, and other tests should be used to fully test the resource's functionality.~~ 

**Update:** one basic test with CRUD + import + external changes, with the possibility to create tests that cover other cases (complete), but it depends on the resource structure and what can be covered by the basic one. The pattern used is:

Basic
- Create - without optionals
- Import - without optionals
- Update - set optionals
- Import - with optionals
- Update - unset optionals
- Update - external changes
- Empty config - to ensure no resource is in the state
- Create - with optionals

Complete
- ... more if needed (depending on a resource)

In some of the test files, we ended up merging those two into one, but I wanted to keep them separate as we need two Create steps, one with and the second without optional fields set. Alternatively, here are other approaches I propose:

1) We could use two `resource.Test` functions in one test function, which wasn't practiced before that much (I think we have one or two such tests). The advantage of this would be that the objects needed in both tests would be created only once.

2) Another approach would be to use both models at once. It was also not done that often before, but it could improve test performance and maybe test readability.

**Updated:** 3) It turns out there is a Taint option in the Terraform testing framework that enables the user to have multiple create steps within one step which eliminates the need to have two test functions (basic + complete) for basic resources. In the cases of simple resources we should have only one function (basic) with above operations.

## Next PRs
- See how it applies to a more complex object (e.g. view or warehouse)
Create similar common testing patterns for the following test types:
- Data source tests
- Resources representing connections rather than objects (e.g. grants) 
- Provider tests?

### DIFF
--- a/pkg/acceptance/bettertestspoc/assert/objectassert/database_snowflake_ext.go
+++ b/pkg/acceptance/bettertestspoc/assert/objectassert/database_snowflake_ext.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 )
 
+// TODO [SNOW-1501905]: generalize this type of assertion
 type databaseNonExistenceCheck struct {
 	id sdk.AccountObjectIdentifier
 }


pkg/testacc/resource_database_acceptance_test.go
```diff
@@ -28,38 +29,39 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestAcc_Database_Basic(t *testing.T) {
+func TestAcc_Database_BasicUseCase(t *testing.T) {
```
sfc-gh-jmichalak, line 32: We should note down somewhere what the expectations are for both of these functions. Now, it's only in the PR description, but that's not enough.

Makefile
```diff
@@ -89,7 +89,7 @@ test-acceptance-%: ## run acceptance tests (both non-account and account level o
 	TF_ACC=1 TF_LOG=DEBUG SNOWFLAKE_DRIVER_TRACING=debug SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE=true SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags=non_account_level_tests,account_level_tests -run ^TestAcc_$*_ -v -timeout=20m ./pkg/testacc
 
 test-build-verification: ## run build verification tests
-	TF_ACC=1 SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE=true TEST_SF_TF_REQUIRE_TEST_OBJECT_SUFFIX=1 TEST_SF_TF_REQUIRE_GENERATED_RANDOM_VALUE=1 SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags=non_account_level_tests,account_level_tests -run "^(TestAcc_Schema_basic|TestAcc_Schema_complete)$$" -v -cover -timeout=20m ./pkg/testacc
+	TF_ACC=1 SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE=true TEST_SF_TF_REQUIRE_TEST_OBJECT_SUFFIX=1 TEST_SF_TF_REQUIRE_GENERATED_RANDOM_VALUE=1 SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags=non_account_level_tests,account_level_tests -run "^(TestAcc_.*_BasicUseCase.*|TestAcc_.*_CompleteUseCase.*)$$" -v -cover -timeout=20m ./pkg/testacc
```
sfc-gh-asawicki, line 92: Why can't we just let them be named basic and complete as before?

pkg/testacc/resource_database_acceptance_test.go
```diff
@@ -68,84 +70,209 @@ func TestAcc_Database_Basic(t *testing.T) {
...
-					resource.TestCheckResourceAttrPtr("snowflake_database.test", "trace_level", accountTraceLevel),
-					resource.TestCheckResourceAttrPtr("snowflake_database.test", "suspend_task_after_num_failures", accountSuspendTaskAfterNumFailures),
-					resource.TestCheckResourceAttrPtr("snowflake_database.test", "task_auto_retry_attempts", accountTaskAutoRetryAttempts),
-					resource.TestCheckResourceAttrPtr("snowflake_database.test", "user_task_managed_initial_warehouse_size", accountUserTaskMangedInitialWarehouseSize),
-					resource.TestCheckResourceAttrPtr("snowflake_database.test", "user_task_timeout_ms", accountUserTaskTimeoutMs),
-					resource.TestCheckResourceAttrPtr("snowflake_database.test", "user_task_minimum_trigger_interval_in_seconds", accountUserTaskMinimumTriggerIntervalInSeconds),
-					resource.TestCheckResourceAttrPtr("snowflake_database.test", "quoted_identifiers_ignore_case", accountQuotedIdentifiersIgnoreCase),
-					resource.TestCheckResourceAttrPtr("snowflake_database.test", "enable_console_output", accountEnableConsoleOutput),
+func TestAcc_Database_CompleteUseCase(t *testing.T) {
```
sfc-gh-asawicki: So, in the PR description there is the point: `We could use two resource.Test functions in one test function`, I wouldn't go that route because:
1. it would make these two tests coupled
2. there are no docs on what are the consequences of having two tests inside one func
3. It makes it harder to filter the desired tests and they live under one

I vote for keeping them separated like here

pkg/testacc/resource_database_acceptance_test.go
```diff
@@ -68,84 +70,209 @@ func TestAcc_Database_Basic(t *testing.T) {
...
+			// Update - set optionals
+			{
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(complete.ResourceReference(), plancheck.ResourceActionUpdate),
+					},
+				},
+				Config: accconfig.FromModels(t, complete),
+				Check: assertThat(t,
+					resourceassert.DatabaseResource(t, complete.ResourceReference()).
```
sfc-gh-asawicki: Idea: for some resources we also check the Snowflake state. I think we should be doing it in these tests. That way they really cover all aspects.

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-asawicki: Most of the tests are written differently:
- Basic covers all sets/unsets, etc.
- Complete is to verify one thing that basic does not cover which is an initial creation with all parameters set
I think we should stick with this to have smaller number of changes needed with tests unification

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-jcieslak: Yeah, I agree, but this pr's purpose is to set those expectations (I gave here my proposal and expect discussions regarding the chosen approach), and then we can write them down.

Makefile
sfc-gh-jcieslak: We have a lot of Basic / Complete tests, as I mentioned, I wanted to adjust them gradually. I chose a different but similar name, so that I know which ones were adjusted and which ones are still pending. Once we have adjusted all the Basic / Complete tests, we can switch back to the original naming. This was at least my motivation / reasoning behind this change.

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-jcieslak: I agree, I just had this thought, because I saw they may require the same setup. I guess this could be solved in the future by providing a common function for defining basic/complete tests by providing the necessary "parameters" and having only one setup.

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-jcieslak: Hmm, good idea. I'll add them.

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-jcieslak: Yeah, that's true, I wanted to split those a bit to make the basic test less chunky, but as you mentioned, maybe this could be more time-consuming to adjust to. I'll adjust this one to this convention.

sfc-gh-jcieslak: I found that the Taint property in the test step we could use to taint the resource, which destroys the resource and creates it with a new configuration. This way, we could keep the missing step of creating with full config within the basic test (it's already pretty big so adding one step instead of another test function is better imo)

sfc-gh-jcieslak: As mentioned above merged into one, but complete could be used in some cases as you said to verify things that are not covered by basic case

pkg/testacc/resource_database_acceptance_test.go
```diff
@@ -28,37 +29,126 @@ import (
...
 
-	configVariables := func(id sdk.AccountObjectIdentifier, comment string) config.Variables {
-		return config.Variables{
-			"name":    config.StringVariable(id.Name()),
-			"comment": config.StringVariable(comment),
-		}
+	basic := model.Database("test", id.Name())
+
+	assertBasic := []assert.TestCheckFuncProvider{
+		objectassert.Database(t, id).
```
sfc-gh-jmichalak, line 47: Idea: we could just move such bundled assertions to bettertestspoc package - like we already have with `HasAllDefaults`, etc.

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-jmichalak: Using Taint LGTM, we can do it in this or the next PR.

Makefile
sfc-gh-asawicki: Okay, I understand. One of the other options would be just listing here all the tests that are ready (e.g. by full names), but your approach is also good.

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-asawicki: It is the option, but the execution stops on the first failed step. We could argue that we already have a lot of steps, so adding one or two more steps (depending on whether the basic test will also contain full config import, and AFAIK it does, so just one step) shouldn't harm that much but still, these are separate use cases so maybe it's better to keep them separated.
Additionally, we would need a validation that the object is fully removed from Snowflake before trying to recreate it with the full config (and we don't need that with a separate test using a completely diffferent id).
Let's discuss next week?

sfc-gh-asawicki: I answered above about merging

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-asawicki: The "problem" with moving the bundled assertions is that you need to know the values used for model. Ofc you can do this with the defaults, as we know them. As for all other assertions, like complete below, I would say that we could either:
- have a logic translating from model to assertion (e.g. bundled assertion that takes the model as parameter and filling out all fields based on it)
- keep them directly in tests - the previous one is, in a way, more error-prone, as you don't directly know what will be asserted
I would keep them here for now and we can discuss this as part of SNOW-1501905

pkg/testacc/resource_database_acceptance_test.go
```diff
@@ -265,14 +245,18 @@ func TestAcc_Database_CompleteUseCase(t *testing.T) {
...
-						HasIsTransientString("false").
-						HasReplicationEmpty().
-						HasCommentEmpty().
-						HasAllDefaultParameters(),
-				),
+				Check:  assertThat(t, assertBasic...),
+			},
+			// Create - with optionals
+			{
+				Taint: []string{complete.ResourceReference()},
```
sfc-gh-asawicki: There is no validation here that the object was deleted on Snowflake side, and that it is in fact being recreated.

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-jcieslak: As discussed, I added a new step that validates the object was removed

pkg/testacc/resource_database_acceptance_test.go
```diff
@@ -247,12 +247,18 @@ func TestAcc_Database_BasicUseCase(t *testing.T) {
 				Config: accconfig.FromModels(t, basic),
 				Check:  assertThat(t, assertBasic...),
 			},
+			// Empty config - ensure database is destroyed
+			{
+				Config: " ",
+				Check: assertThat(t,
+					objectassert.DatabaseIsMissing(t, id),
```
sfc-gh-asawicki: we already have a `ViewDoesNotExist`: in a different package and with a different implementation. We should keep them together and consistent.

pkg/testacc/resource_database_acceptance_test.go
```diff
@@ -247,12 +247,18 @@ func TestAcc_Database_BasicUseCase(t *testing.T) {
 				Config: accconfig.FromModels(t, basic),
 				Check:  assertThat(t, assertBasic...),
 			},
+			// Empty config - ensure database is destroyed
+			{
+				Config: " ",
```
sfc-gh-asawicki: nit: why not `Destroy: true`? We have it in other tests.

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-jcieslak: Yeah, that's what I was missing, adjusted

pkg/testacc/resource_database_acceptance_test.go
sfc-gh-jcieslak: Hmm, I thought it worked a bit differently, but it seems to be working as well as an empty string. Adjusted to use Destroy instead.

pkg/acceptance/bettertestspoc/assert/objectassert/database_snowflake_ext.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func DatabaseIsMissing(t *testing.T, id sdk.AccountObjectIdentifier) *DatabaseAssert {
+	t.Helper()
+	return &DatabaseAssert{
+		assert.NewSnowflakeObjectAssertWithTestClientObjectProvider(sdk.ObjectTypeDatabase, id, func(testClient *helpers.TestClient) assert.ObjectProvider[sdk.Database, sdk.AccountObjectIdentifier] {
+			return func(t *testing.T, identifier sdk.AccountObjectIdentifier) (*sdk.Database, error) {
+				t.Helper()
+				db, err := testClient.Database.Show(t, id)
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Will be addressed in the follow-up

PR #4114, opened 2025-10-21
chore: Base for main connection resource use case tests

sfc-gh-jcieslak:
Continuation of https://github.com/snowflakedb/terraform-provider-snowflake/pull/4113, but for resources that represent connections rather than a Snowflake object itself. In this case, the resource and the tests turned out to be pretty straightforward, as the grant resource(s) don't have many non-force-new optional fields. The only being is always_apply, which is thoroughly tested in other tests (thus omitted in the adjusted tests). Due to the simplicity of the grant resource(s), I suggest only going with one test covering "complete" use case, which is:
- Create with required fields
- Import
- Update by updating config
- Update by detected external changes

## Next prs
- Adjust other grant tests within the `resource_grant_privileges_to_account_role_acceptance_test.go` and other grant test files


pkg/acceptance/helpers/grant_client.go
```diff
@@ -48,6 +48,28 @@ func (c *GrantClient) GrantGlobalPrivilegesOnAccount(
 	require.NoError(t, err)
 }
 
+func (c *GrantClient) RevokeGlobalPrivilegesFromAccount(
```
sfc-gh-jmichalak: Nit:
```suggestion
func (c *GrantClient) RevokeGlobalPrivilegesFromAccountRole(
```

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -310,7 +364,7 @@ func TestAcc_GrantPrivilegesToAccountRole_OnAccount_PrivilegesReversed(t *testin
 	})
 }
 
-func TestAcc_GrantPrivilegesToAccountRole_OnAccountObject(t *testing.T) {
+func TestAcc_GrantPrivilegesToAccountRole_OnAccountObject_CompleteUseCase(t *testing.T) {
```
sfc-gh-asawicki: I think it should be basic but let's discuss in #4113

sfc-gh-jcieslak: Yeah, I was a bit confused on which one to use, but I guess this covers the baseline functionality for each option, which makes more sense to call it basic, rather than complete which we rather want to keep for more targeted use cases 🤔

pkg/acceptance/helpers/grant_client.go
sfc-gh-jcieslak: Done

PR #4115, opened 2025-10-21
chore: Base for main data source use case tests

sfc-gh-jcieslak:
The continuation of adjusting the main use case tests to fit a common pattern. For data sources, I propose two tests:
 - Basic Different filtering (singular filters + only assertions to assure correct filters are applied)
- Complete (one step with all options and filters + assertion on mappings)

Also, removed unnecessary explicit assert in services as it's asserted within ServicesDatasourceShowOutput called below.


pkg/testacc/data_source_databases_acceptance_test.go
```diff
@@ -160,6 +66,102 @@ func TestAcc_Databases_DifferentFiltering(t *testing.T) {
 	})
 }
 
+func TestAcc_Databases_CompleteUseCase(t *testing.T) {
```
sfc-gh-jmichalak, line 70: Nit: In this case, it's not about representing a complete resource/datasource. It's about using additional sources, so maybe `TestAcc_Databases_WithoutAdditionalData` (a name following the examples we have). It will also make it more clear by being a separate name than in the resources.

pkg/testacc/data_source_databases_acceptance_test.go
```diff
@@ -160,6 +66,102 @@ func TestAcc_Databases_DifferentFiltering(t *testing.T) {
...
+						HasRetentionTimeNotEmpty().
+						HasResourceGroup("").
+						HasOwnerRoleTypeNotEmpty(),
+
+					assert.Check(resource.TestCheckResourceAttr(databasesModel.DatasourceReference(), "databases.0.describe_output.#", "2")),
+					assert.Check(resource.TestCheckResourceAttrSet(databasesModel.DatasourceReference(), "databases.0.describe_output.0.created_on")),
+					assert.Check(resource.TestCheckResourceAttrSet(databasesModel.DatasourceReference(), "databases.0.describe_output.0.name")),
+					assert.Check(resource.TestCheckResourceAttr(databasesModel.DatasourceReference(), "databases.0.describe_output.0.kind", "SCHEMA")),
+
+					assert.Check(resource.TestCheckResourceAttr(databasesModel.DatasourceReference(), "databases.0.parameters.#", "1")),
```
sfc-gh-asawicki: Why are these only checking the presence and not reusing a defined default? (We could have a default parameters set for database assertion if we don't).

pkg/testacc/data_source_databases_acceptance_test.go
sfc-gh-jcieslak: This case is about checking all assertions, outputs, and their mappings. Because of that, I don't think `TestAcc_Databases_WithoutAdditionalData` is correctly describing the test, and the current format is good enough. I can still change the test names at the end of the whole refactor. It's not a big change, but I would rather move those at the end rather than do them now.

pkg/testacc/data_source_databases_acceptance_test.go
sfc-gh-jcieslak: Assertions adjusted. As we discussed today, the `value` field has the most value for the end users, but should we at least check for the presence of other subfields (can be added in the follow-up)?

PR #4119, opened 2025-10-22
chore: Regenerate sequences in SDK using new generator setup

sfc-gh-asawicki:
Continuation for: #4050

The main part of this change is to regenerate a single object to validate if the new setup is correct and to plan the regeneration of all objects in the SDK. The findings/actions needed:
- Remove sequences from the old definition list
- Remove the generation directive from the Sequences definition
- Regeneration:
  - all
    - generated code disclaimer
  - _dto_gen.go
    - `generate` directive removal
  - _gen.go
    - imports reorg
  - _impl_gen.go
    - top vars reorg
    - ShowByID impl small change
    - SafeDrop bool instead of ptr to bool
    - manually edited note added to `convert` mappings
  - validations_gen.go
    - disclaimer only
  - _dto_builders_gen.go
    - casing for attributes
    - removed pointers (adjustments throughout the codebase needed)
  - _gen_test.go
    - test names changes
    - add missing cases
    - alter tests for each option
These changes may differ in other objects but serve as valuable input for the next planned changes.

Other necessary changes made to the generation process:
- Add Makefile commands with the new SDK generator setup to the main one
- Add generate.go file to the sdk package
- Adjust example rework commands (only chosen objects by default)
- Remove path to a separate dto generator
- Adjust templates:
  - remove generate directive (to the separate dto builder gen)
  - add `conversionErrorWrapped`
  - fix whitespace in unit tests gen
  - fix nil ptr generation in unit tests
- Regenerate Sequence and Streamlit rework examples
- Replace `SdkObjectDef` with the interfaces list
- Add all definitions to the rework list

Next steps:
- Regenerate for all objects (document what parts need to be adjusted manually for future improvements):
  - 1st part: dto, interface, imp, validations
  - 2nd part: dto builders
  - 3rd part: unit tests
- Remove old generation setup completely (multiple TODOs left in earlier changes), including Makefile etc.
- Update the generation documentation, limitations, next steps, etc.

### DIFF
--- a/pkg/sdk/testint/sequences_integration_test.go
+++ b/pkg/sdk/testint/sequences_integration_test.go
@@ -68,7 +68,7 @@ func TestInt_Sequences(t *testing.T) {
 
 		comment := random.Comment()
 		request := sdk.NewCreateSequenceRequest(id).
-			WithStart(sdk.Int(1)).
+			WithStart(1).
 			WithIncrement(1).
 			WithIfNotExists(true).
 			WithValuesBehavior(sdk.ValuesBehaviorOrder).


Makefile
```diff
@@ -136,6 +136,13 @@ run-generator-%: ./pkg/sdk/%_def.go ## Run go generate on given object definitio
 	go generate $<
 	go generate ./pkg/sdk/$*_dto_gen.go
 
+generate-sdk: ## Generate all SDK objects
```
sfc-gh-jmichalak, line 139: Just for the next PRs: we should have an option here to regenerate a single object.

pkg/sdk/poc/internal/rework/sdk_definitions.go
```diff
@@ -144,3 +144,7 @@ func SafeDrop[ID ObjectIdentifierConstraint](
 func String(s string) *string {
 	return sdk.String(s)
 }
+
+func conversionErrorWrapped[U any](_ *U, _ error) (*U, error) {
+	return nil, nil
+}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: What about this?

pkg/sdk/poc/new.go
```diff
@@ -10,19 +10,14 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/poc/generator"
 )
 
-type SdkObjectDef struct {
-	name string
-	// TODO [SNOW-2324252]: can be removed?
-	file       string
-	definition *generator.Interface
-}
+// TODO [SNOW-2324252]: rename this file and move it (can't be moved currently due to import cycle: sdk needs gen fro definition and , generator needs all the definitions list
```
sfc-gh-jmichalak: Nit: This sentence looks incomplete.

pkg/sdk/poc/internal/rework/sdk_definitions.go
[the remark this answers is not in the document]
sfc-gh-asawicki: this method is there only to compile the generated code in this pkg, and no to do any logic itself, rejecting

pkg/sdk/sequences_gen_test.go
```diff
@@ -6,31 +8,36 @@ import (
...
 	}
 
 	t.Run("validation: nil options", func(t *testing.T) {
-		var opts *CreateSequenceOptions = nil
+		opts := (*CreateSequenceOptions)(nil)
 		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
 	})
 
-	t.Run("validation: incorrect identifier", func(t *testing.T) {
+	t.Run("validation: valid identifier for [opts.name]", func(t *testing.T) {
```
sfc-gh-jmichalak, line 23: Nit: this naming is inconsistent with other test names. The alternatives are:
- all test names represent a validator behavior (positive), like `identifier for [opts.name] must be valid`, etc
- all test names represent an invalid case (negative), like `invalid/incorrect identifier for [opts.name]`

I'm fine with both, but IMO we should establish what we prefer, as now, we have a mix of both positives (`valid identifier...`) and negatives (`conflicting fields`), and it may be confusing.

Makefile
sfc-gh-jcieslak, line 139: Q: Shouldn't we replace or remove the `run-generator-%` command?

sfc-gh-asawicki: From PR description:
`Remove old generation setup completely (multiple TODOs left in earlier changes), including Makefile etc.`

sfc-gh-asawicki: You have it out of the box with the new generator. Check the README in the rework package, e.g. `make generate-sdk SF_TF_GENERATOR_ARGS='--filter-object-names=Sequences'` (the changes in this PR were generated like this).

pkg/sdk/poc/new.go
sfc-gh-asawicki: There is a missing closing paren, will adjust in the next PR

Makefile
sfc-gh-asawicki: And you have this out of the box in the new gen: `make generate-sdk SF_TF_GENERATOR_ARGS='--filter-object-names=Sequences'`

pkg/sdk/sequences_gen_test.go
sfc-gh-asawicki: This is the naming that the SDK generator is generating (so it's currently like this in many objects, the names should be kept as generated to make it easier to understand what happened during regeneration). We can discuss this naming separately, I'll add to the list.

pkg/sdk/poc/new.go
sfc-gh-asawicki: Updated in this one

Makefile
sfc-gh-jmichalak: I know, but I'd like to have a specific `make` target which is no less convenient than before.

PR #4123, opened 2025-10-23
feat: Add integration tests for Notebook

sfc-gh-mkowalski:
Added integration tests for [Notebook object](https://docs.snowflake.com/en/sql-reference/commands-notebook) and slightly adjusted its definition.

### DIFF
--- a/pkg/sdk/testint/errors_integration_test.go
+++ b/pkg/sdk/testint/errors_integration_test.go
@@ -61,7 +61,7 @@ func TestInt_ShowSchemaObjectInNonExistingDatabase(t *testing.T) {
 		{ObjectType: sdk.ObjectTypeImageRepository, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).ImageRepositories.ShowByID)},
 		{ObjectType: sdk.ObjectTypeService, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Services.ShowByID)},
 		{ObjectType: sdk.ObjectTypeGitRepository, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).GitRepositories.ShowByID)},
-		{ObjectType: sdk.ObjectTypeNotebook, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Notebooks.ShowByID)},
+		{ObjectType: sdk.ObjectTypeNotebook, ExpectedErr: sdk.ErrDoesNotExistOrOperationCannotBePerformed, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Notebooks.ShowByID)},
 	}
 
 	for _, tt := range testCases {
@@ -122,7 +122,7 @@ func TestInt_ShowSchemaObjectInNonExistingSchema(t *testing.T) {
 		{ObjectType: sdk.ObjectTypeImageRepository, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).ImageRepositories.ShowByID)},
 		{ObjectType: sdk.ObjectTypeService, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Services.ShowByID)},
 		{ObjectType: sdk.ObjectTypeGitRepository, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).GitRepositories.ShowByID)},
-		{ObjectType: sdk.ObjectTypeNotebook, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Notebooks.ShowByID)},
+		{ObjectType: sdk.ObjectTypeNotebook, ExpectedErr: sdk.ErrDoesNotExistOrOperationCannotBePerformed, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Notebooks.ShowByID)},
 	}
 
 	for _, tt := range testCases {


pkg/acceptance/helpers/notebook_client.go
```diff
@@ -0,0 +1,67 @@
...
+		context: context,
+		ids:     idsGenerator,
+	}
+}
+
+func (c *NotebookClient) client() sdk.Notebooks {
+	return c.context.client.Notebooks
+}
+
+func (c *NotebookClient) Create(t *testing.T, name sdk.SchemaObjectIdentifier) (*sdk.Notebook, func()) {
```
sfc-gh-jmichalak: Nit: remove `name`. The ID (name) should be generated inside this function. And in tests, we don't need to generate it upfront. Later in tests we just use `notebook.ID` from the returned value.

pkg/acceptance/helpers/notebook_client.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return notebook, c.DropFunc(t, req.GetName())
+}
+
+func (c *NotebookClient) DropFunc(t *testing.T, id sdk.SchemaObjectIdentifier) func() {
+	t.Helper()
+	ctx := context.Background()
+
+	return func() {
+		err := c.client().Drop(ctx, sdk.NewDropNotebookRequest(id).WithIfExists(true))
```
sfc-gh-jmichalak: Nit: just use `DropSafely`.

pkg/sdk/notebooks_def.go
```diff
@@ -67,7 +67,7 @@ var notebookDetails = g.PlainStruct("NotebookDetails").
 	Text("DefaultPackages").
 	OptionalText("UserPackages").
 	OptionalText("RuntimeName").
-	OptionalText("ComputePool").
+	Field("ComputePool", "AccountObjectIdentifier").
```
sfc-gh-jmichalak: Shouldn't it be optional - with `*`?

pkg/acceptance/bettertestspoc/assert/objectassert/notebook_details_snowflake_gen.go
```diff
@@ -0,0 +1,373 @@
...
+func (s *NotebookDetailsAssert) HasComputePool(expected sdk.AccountObjectIdentifier) *NotebookDetailsAssert {
+	s.AddAssertion(func(t *testing.T, o *sdk.NotebookDetails) error {
+		t.Helper()
+		if o.ComputePool.Name() != expected.Name() {
+			return fmt.Errorf("expected compute pool: %v; got: %v", expected.Name(), o.ComputePool.Name())
+		}
+		return nil
+	})
+	return s
+}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: This should be regenerated after changes in SDK.

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+		t.Cleanup(testClientHelper().Notebook.DropFunc(t, id))
+
+		notebook, err := client.Notebooks.ShowByID(ctx, id)
+		require.NoError(t, err)
+
+		assertThatObject(t, objectassert.NotebookFromObject(t, notebook).
+			HasCreatedOnNotEmpty().
+			HasName(id.Name()).
+			HasDatabaseName(id.DatabaseName()).
+			HasSchemaName(id.SchemaName()).
```
sfc-gh-jmichalak, line 34: Let's assert at least the fields we directly control - like comment, query_warehouse...

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+		computePool, computePoolCleanup := testClientHelper().ComputePool.Create(t)
+		t.Cleanup(computePoolCleanup)
+
+		// TODO(SNOW-2398051) Some of the fields were omitted due to lack of documentation.
+		request := sdk.NewCreateNotebookRequest(id).WithIfNotExists(true).
+			WithComment("comment").
+			WithTitle("title").
+			WithMainFile("main_file").
+			WithComputePool(computePool.ID()).
+			WithIdleAutoShutdownTimeSeconds(3600).
```
sfc-gh-jmichalak, line 67: This request should contain the fields (and later assertions) that are in the docs - warehouse, query_warehouse, from.

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+	})
+
+	t.Run("alter: set", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+
+		_, notebookCleanup := testClientHelper().Notebook.Create(t, id)
+		t.Cleanup(notebookCleanup)
+
+		setRequest := sdk.NewNotebookSetRequest().WithComment("comment")
+		alterRequest := sdk.NewAlterNotebookRequest(id).WithSet(*setRequest)
```
sfc-gh-jmichalak, line 148: Other fields are missing - we should test what's in the docs.

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+			HasOwner(snowflakeroles.PentestingRole.Name()).
+			HasNoQueryWarehouse().
+			HasOwnerRoleType("ROLE"),
+		)
+	})
+
+	t.Run("alter: unset", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+
+		createRequest := sdk.NewCreateNotebookRequest(id).WithComment("comment")
```
sfc-gh-jmichalak: Here, set all fields, like in `create: complete` & below unset these fields (let's say the ones that are in SET).

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+	t.Run("drop", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+		_, notebookCleanup := testClientHelper().Notebook.Create(t, id)
+		t.Cleanup(notebookCleanup)
+
+		err := client.Notebooks.Drop(ctx, sdk.NewDropNotebookRequest(id).WithIfExists(true))
+		require.NoError(t, err)
+
+		_, err = client.Notebooks.ShowByID(ctx, id)
+		require.Error(t, err)
```
sfc-gh-jmichalak, line 309:
```suggestion
				require.ErrorIs(t, err, sdk.ErrObjectNotExistOrAuthorized)
```

sfc-gh-jmichalak (changes requested):

.vscode/launch.json
```diff
@@ -3,5 +3,6 @@
     // Hover to view descriptions of existing attributes.
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
-    "configurations": []
```
sfc-gh-jcieslak, line 6: idea: let's actually define and commit useful launchers or completely remove this file (or even .vscode directory). I would keep it and improve it as some/many of our contributors may use vscode. Let's keep it as future improvement (maybe as part of FF?).

.vscode/settings.json
```diff
@@ -2,17 +2,22 @@
     "files.insertFinalNewline": true,
     "go.testEnvFile": "${workspaceFolder}/test.env",
     "go.formatTool": "gofumpt",
+    "go.toolsEnvVars": {
+        "GOFLAGS": "-tags=account_level_tests"
```
sfc-gh-jcieslak: Isn't it a repetition of the below? + I think you also adjusted those @sfc-gh-jmichalak, right?

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+		request := sdk.NewCreateNotebookRequest(id)
+
+		err := client.Notebooks.Create(ctx, request)
+		require.NoError(t, err)
+		t.Cleanup(testClientHelper().Notebook.DropFunc(t, id))
+
+		notebook, err := client.Notebooks.ShowByID(ctx, id)
+		require.NoError(t, err)
+
+		assertThatObject(t, objectassert.NotebookFromObject(t, notebook).
```
sfc-gh-jcieslak, line 30: nit: You can make it a bit shorter by using just NotebookObject assert that calls Show internally

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+			HasComputePool(computePool.ID()).
+			HasOwner(snowflakeroles.PentestingRole.Name()).
+			HasImportUrls("[]").
+			HasExternalAccessIntegrations("[]").
+			HasExternalAccessSecrets("{}").
+			HasCodeWarehouse("SYSTEM$STREAMLIT_NOTEBOOK_WH").
+			HasIdleAutoShutdownTimeSeconds(3600).
+			HasRuntimeEnvironmentVersion("WH-RUNTIME-2.0").
+			HasName(id.Name()).
+			HasComment(&comment).
```
sfc-gh-jcieslak: nit: You can use sdk.String or sdk.Pointer if you need a reference to a string

pkg/sdk/testint/notebooks_integration_test.go
sfc-gh-jcieslak: \+ assert the parameters they change (e.g. If a parameter is only accessible through describe, add describe assert with expected values)

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,248 @@
...
+		)
+	})
+
+	t.Run("show: with like", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+		createRequest := sdk.NewCreateNotebookRequest(id).WithComment("comment")
+		_, notebookCleanup := testClientHelper().Notebook.CreateWithRequest(t, createRequest)
+		t.Cleanup(notebookCleanup)
+
+		notebook, err := testClientHelper().Notebook.Show(t, id)
```
sfc-gh-jcieslak: If you want to test show, test it on sdk.Client object, not the testClient one. In case you need notebook object, you returned it in the 215 line (currently it's ignored `_`)

pkg/sdk/testint/notebooks_integration_test.go
sfc-gh-mkowalski: It gave me this error: 
```
Error: Target error should be in err chain:
           expected: "[errors.go:23] object does not exist or not authorized"
           in chain: "object does not exist"
```

So I went with `sdk.ErrObjectNotFound`

pkg/sdk/testint/notebooks_integration_test.go
sfc-gh-mkowalski: I took this approach from the `git_repository` implementation so it also has the same issue.

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -107,10 +120,34 @@ func TestInt_Notebooks(t *testing.T) {
...
+
+		// secret, secreteCleanup := testClientHelper().Secret.CreateRandomPasswordSecret(t)
+		// t.Cleanup(secreteCleanup)
+
+		// secrets := sdk.SecretsListRequest{SecretsList: []sdk.SecretReference{{
+		// 	VariableName: "sample_secret",
+		// 	Name:         secret,
+		// }}}
+
+		// TODO: Investigate the 'Secrets' field (not present in both SHOW and DESC).
```
sfc-gh-jcieslak: When is this going to be addressed?

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -43,14 +45,26 @@ func TestInt_Notebooks(t *testing.T) {
 		computePool, computePoolCleanup := testClientHelper().ComputePool.Create(t)
 		t.Cleanup(computePoolCleanup)
 
+		warehouse, warehouseCleanup := testClientHelper().Warehouse.CreateWarehouse(t)
+		t.Cleanup(warehouseCleanup)
+
+		stage, stageCleanup := testClientHelper().Stage.CreateStage(t)
+		t.Cleanup(stageCleanup)
+
+		testClientHelper().Stage.PutOnStage(t, stage.ID(), "example.ipynb")
```
sfc-gh-jcieslak, line 57: This file is in a different directory, I don't think it should work 🤔

sfc-gh-mkowalski: I put it in the same folder as `manifest.yml` which is being referenced using `testClientHelper().Stage.PutOnStage(t, stage.ID(), "manifest.yml")` in the other part of the code. I suppose it should work in the same way; but correct me if I'm mistaken.

.vscode/settings.json
sfc-gh-jmichalak: I adjusted `go.testFlags`, this is different.

pkg/sdk/testint/notebooks_integration_test.go
sfc-gh-jmichalak: This is a missing feature, so this needs to be reported and done by a notebooks team.

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,379 @@
...
+			HasNoLiveVersionLocationUri(),
+		)
+	})
+
+	t.Run("alter: set", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+
+		request := sdk.NewCreateNotebookRequest(id)
+
+		_, notebookCleanup := testClientHelper().Notebook.CreateWithRequest(t, request)
```
sfc-gh-jmichalak: Nit: here, a simple Create should be used.

pkg/sdk/testint/notebooks_integration_test.go
```diff
@@ -0,0 +1,379 @@
...
+		)
+	})
+
+	t.Run("create - complete", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+
+		computePool, computePoolCleanup := testClientHelper().ComputePool.Create(t)
+		t.Cleanup(computePoolCleanup)
+
+		warehouse, warehouseCleanup := testClientHelper().Warehouse.CreateWarehouse(t)
```
sfc-gh-jmichalak, line 48: Nit: we missed it before, but let's actually use different warehouses for query warehouse and the regular warehouse. The reason is that in the implementation these two could get swapped, and now we wouldn't know.

pkg/sdk/testint/notebooks_integration_test.go
sfc-gh-jmichalak: Yup, it prepends `testdata`, so it should work.

pkg/sdk/testint/notebooks_integration_test.go
sfc-gh-jmichalak: OK, I think not all objects return the exact same error in case they're missing, so that's fine. But look at errors coming from `errors_integration_test.go` and adjust accordingly.

pkg/sdk/testint/notebooks_integration_test.go
sfc-gh-jcieslak: Alright, so then we should create a ticket for it ourselves and mark this todo with the ticket number, e.g., `TODO(SNOW-XXXXXX): Support secrets`. I'll show you.

pkg/sdk/testint/errors_integration_test.go
```diff
@@ -61,6 +61,7 @@ func TestInt_ShowSchemaObjectInNonExistingDatabase(t *testing.T) {
 		{ObjectType: sdk.ObjectTypeImageRepository, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).ImageRepositories.ShowByID)},
 		{ObjectType: sdk.ObjectTypeService, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Services.ShowByID)},
 		{ObjectType: sdk.ObjectTypeGitRepository, ExpectedErr: sdk.ErrObjectNotExistOrAuthorized, ShowFn: schemaObjectShowByIDWrapper(testClient(t).GitRepositories.ShowByID)},
+		{ObjectType: sdk.ObjectTypeNotebook, ExpectedErr: sdk.ErrObjectNotFound, ShowFn: schemaObjectShowByIDWrapper(testClient(t).Notebooks.ShowByID)},
```
sfc-gh-jcieslak: Hmm, those added ExpectedErrs don't look right. Usually, it's either `ErrDoesNotExistOrOperationCannotBePerformed` or `ErrObjectNotExistOrAuthorized`. I don't think notebooks should be any different. I think that's because you use the Safe version of Show and Drop inside test client functions. It means other users use the non-Safe version, so please do as in other Shows and Drops, and use non-safe versions.

sfc-gh-jmichalak: I'm confused. Why should the testing helpers not use the Safe methods to begin with? These helpers provide an abstraction layer that handles errors and additional setups, so using Safe methods feels more natural.

Maybe we should test the regular methods instead of the testing ones in this file. Let's discuss and adjust this later. For now, let's use the Unsafe versions for consistency.

sfc-gh-jcieslak: the sdk.Client is used here so the comment doesn't matter, resolving

sfc-gh-jcieslak (approved):

sfc-gh-jmichalak (dismissed):

PR #4124, opened 2025-10-23
chore: Base for main resource use case tests follow up 1

sfc-gh-jcieslak:
Follow up on adjusting resource acceptance tests according to the rules set in #4113
The list of adjusted tests:
- account role
- database role
- network policy
- schema
- secondary database


pkg/acceptance/bettertestspoc/assert/objectassert/role_snowflake_ext.go
```diff
@@ -0,0 +1,30 @@
...
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func AccountRoleIsMissing(t *testing.T, id sdk.AccountObjectIdentifier) *RoleAssert {
```
sfc-gh-jmichalak, line 13: We should consider returning a custom struct. Now, you can chain this assert with other field asserts, resulting in an assert that doesn't make sense. Also, even using `SnowflakeObjectAssert` is not ideal because it's for asserting existing objects, and this is just a hack. We're asserting not during the actual assertions, but during constructing the assert.

Given these two points, IMO we should separate asserting objects vs asserting they don't exist.

Also, I suppose this could be easily generated.

pkg/acceptance/bettertestspoc/assert/objectassert/role_snowflake_ext.go
```diff
@@ -0,0 +1,30 @@
...
+	return &RoleAssert{
+		assert.NewSnowflakeObjectAssertWithTestClientObjectProvider(sdk.ObjectTypeRole, id, func(testClient *helpers.TestClient) assert.ObjectProvider[sdk.Role, sdk.AccountObjectIdentifier] {
+			return func(t *testing.T, id sdk.AccountObjectIdentifier) (*sdk.Role, error) {
+				t.Helper()
+				accountRole, err := testClient.Role.Show(t, id)
+				if err != nil {
+					if errors.Is(err, sdk.ErrObjectNotFound) {
+						return accountRole, nil
+					}
+					return accountRole, nil
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Pls fix.

pkg/testacc/resource_schema_acceptance_test.go
```diff
@@ -96,227 +165,129 @@ func TestAcc_Schema_basic(t *testing.T) {
...
+				},
+				Config: accconfig.FromModels(t, complete),
+				Check:  assertThat(t, assertComplete...),
+			},
 		},
 	})
 }
 
-func TestAcc_Schema_complete(t *testing.T) {
+func TestAcc_Schema_CompleteUseCase(t *testing.T) {
```
sfc-gh-jmichalak, line 273: Why is it separate?

pkg/testacc/resource_secondary_database_acceptance_test.go
```diff
@@ -32,118 +37,206 @@ func TestAcc_CreateSecondaryDatabase_Basic(t *testing.T) {
...
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(complete.ResourceReference(), plancheck.ResourceActionCreate),
+					},
+				},
+				Config: accconfig.FromModels(t, complete),
+				Check:  assertThat(t, assertComplete...),
 			},
 		},
 	})
```
sfc-gh-jmichalak, line 242: <p class='graphite__hidden'><i>[Re: line +245]</i></p>
Why is it separated and the name does not follow the convention?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/snowflakedb/terraform-provider-snowflake/4124?utm_source=unchanged-line-comment">Graphite</a>.</p>

sfc-gh-jmichalak (changes requested):

pkg/acceptance/bettertestspoc/assert/objectassert/database_role_snowflake_ext.go
```diff
@@ -0,0 +1,30 @@
...
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func DatabaseRoleIsMissing(t *testing.T, id sdk.DatabaseObjectIdentifier) *DatabaseRoleAssert {
```
sfc-gh-asawicki: I left a comment about this is the previous change.

pkg/acceptance/bettertestspoc/assert/objectassert/role_snowflake_ext.go
sfc-gh-asawicki: +1 I have already left a comment about this in one of the previous changes.

pkg/acceptance/bettertestspoc/assert/objectassert/role_snowflake_ext.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

pkg/acceptance/bettertestspoc/assert/objectassert/role_snowflake_ext.go
sfc-gh-jcieslak: Adjusted

pkg/testacc/resource_schema_acceptance_test.go
sfc-gh-jcieslak: Because it contains the optional forceNew (as far as I remember only: `is_transient` field), which the basic case doesn't. I guess I could be lighter on the assertions here and just make sure the value that is different is only asserted, but in Snowflake, sometimes one value influences others, and that's why I asserted all the fields as above.

PR #4125, opened 2025-10-23
chore: Rework authentication policy resource

sfc-gh-jmichalak:
- Adjust the resource to v1 guidelines.
- Fix object renaming.
- Remove deprecated `mfa_authentication_methods` field from examples.
- Generate builders & assertions. Use them in acc tests.
- Add more acc tests.
- Add SDK helpers.

## TODO
- Add missing blocks for pat/wif/mfa policies.
- Add data source.

### DIFF
--- a/pkg/testacc/resource_authentication_policy_account_level_acceptance_test.go
+++ b/pkg/testacc/resource_authentication_policy_account_level_acceptance_test.go
@@ -19,7 +19,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
-func TestInt_AuthenticationPolicies_AccountLevel(t *testing.T) {
+func TestAcc_AuthenticationPolicies_AccountLevel(t *testing.T) {
 	t.Setenv(string(testenvs.ConfigureClientOnce), "")
 
 	client := secondaryTestClient()


docs/resources/authentication_policy.md
```diff
@@ -86,14 +85,102 @@ Optional:
...
 
-- `authentication_methods` (String)
-- `client_types` (String)
-- `comment` (String)
-- `mfa_authentication_methods` (String)
-- `mfa_enrollment` (String)
-- `name` (String)
-- `owner` (String)
-- `security_integrations` (String)
+- `authentication_methods` (List of Object) (see [below for nested schema](#nestedobjatt--describe_output--authentication_methods))
```
sfc-gh-asawicki: I am not sure about this change for the following reasons:
1. It's a breaking change (if someone was using this output).
2. I don't think we use describe output verbatim in other resources in which describe outputs the list of parameters. The usual direction is that we build the details based on the list and treat it as object.
3. There are some objections against the bloated state (e.g. 3118) and this additional info is not adding anything useful.

pkg/resources/authentication_policy_state_upgraders.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/schemas"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func v2_9_0_AuthenticationPolicyStateUpgrader(ctx context.Context, rawState map[string]any, meta any) (map[string]any, error) {
```
sfc-gh-asawicki, line 12: what about new kind filed in show output? Shouldn't it be filled here too?

sfc-gh-asawicki (changes requested):

pkg/resources/authentication_policy_state_upgraders.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	client := meta.(*provider.Context).Client
+	authenticationPolicyInSnowflake, err := client.AuthenticationPolicies.Describe(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	descOutput := schemas.AuthenticationPolicyDescriptionToSchema(authenticationPolicyInSnowflake)
+	rawState[DescribeOutputAttributeName] = []any{descOutput}
+	delete(rawState, "mfa_authentication_methods")
```
sfc-gh-jcieslak, line 33: Why is this necessary?

docs/resources/authentication_policy.md
sfc-gh-jcieslak: yeah, shouldn't we in this case add new field that has correct type or just ignore it for now and adjust it in v3?

pkg/resources/authentication_policy_state_upgraders.go
sfc-gh-jmichalak: Done.

pkg/resources/authentication_policy_state_upgraders.go
sfc-gh-jmichalak: To avoid a permadiff on this field. Previously, it was always set in the state, which is incompatible with the current logic. If it's set in the state like this, the provider causes a permadiff. Note that this field can still be managed when the BCR is disabled - check the account level test.

docs/resources/authentication_policy.md
sfc-gh-jmichalak: Done as discussed: we use only non-nested values.

sfc-gh-jcieslak (approved):

PR #4130, opened 2025-10-24
chore: Base for main resource use case tests follow up 3

sfc-gh-jcieslak:
Continuation of adjusting resource acceptance tests to the new format.

## Changes
Adjusted resource tests:
- external_oauth_integration
- oauth_integration_for_custom_clients
- oauth_integration_for_partner_applications

During the testing, I saw and adjusted two issues. Both were documented in the migration guide.


pkg/acceptance/bettertestspoc/assert/resourceassert/oauth_integration_for_custom_clients_resource_ext.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+)
+
+func (o *OauthIntegrationForCustomClientsResourceAssert) HasPreAuthorizedRolesListLen(len int) *OauthIntegrationForCustomClientsResourceAssert {
+	o.AddAssertion(assert.ValueSet("pre_authorized_roles_list.#", fmt.Sprintf("%d", len)))
+	return o
+}
+
+func (o *OauthIntegrationForCustomClientsResourceAssert) HasPreAuthorizedRolesListElem(index int, value string) *OauthIntegrationForCustomClientsResourceAssert {
```
sfc-gh-jmichalak: This is quite verbose, can we just assert the whole slice?

pkg/testacc/resource_oauth_integration_for_custom_clients_acceptance_test.go
```diff
@@ -8,455 +8,377 @@ import (
...
+		WithComment(comment)
 
 	resource.Test(t, resource.TestCase{
 		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
 		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
 			tfversion.RequireAbove(tfversion.Version1_5_0),
 		},
+		CheckDestroy: CheckDestroy(t, resourcenames.OauthIntegrationForCustomClients),
 		Steps: []resource.TestStep{
+			// Create - with all optionals (including optional ForceNew parameters)
```
sfc-gh-jmichalak, line 303: I'm not sure how ForceNew matters here, since this is the 1st step.

pkg/testacc/resource_oauth_integration_for_partner_applications_acceptance_test.go
```diff
@@ -13,270 +13,315 @@ import (
 
 	accconfig "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
 	resourcehelpers "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeroles"
+	resourcenames "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
```
sfc-gh-jmichalak: Why is this renamed here? Other resources don't need it.

pkg/testacc/resource_oauth_integration_for_custom_clients_acceptance_test.go
sfc-gh-jcieslak: This matters because in the BasicUseCase we cannot make those values different between basic and complete models, so they may be not defined there, but they are here.

pkg/acceptance/bettertestspoc/assert/resourceassert/oauth_integration_for_custom_clients_resource_ext.go
sfc-gh-jcieslak: Yes, will refactor

pkg/testacc/resource_oauth_integration_for_partner_applications_acceptance_test.go
sfc-gh-jcieslak: It happened automagically, and I can revert it

pkg/acceptance/bettertestspoc/assert/objectassert/security_integration_snowflake_ext.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+)
+
+// TODO [SNOW-1501905]: generalize this type of assertion
+type SecurityIntegrationNonExistenceCheck struct {
```
sfc-gh-asawicki: why is it in this package and not with the others?

pkg/sdk/object_types.go
```diff
@@ -91,6 +91,7 @@ const (
 	ObjectTypeProgrammaticAccessToken ObjectType = "PROGRAMMATIC ACCESS TOKEN" //nolint:gosec
 	ObjectTypeListing                 ObjectType = "LISTING"
 	ObjectTypeSemanticView            ObjectType = "SEMANTIC VIEW"
+	ObjectTypeSecurityIntegration     ObjectType = "SECURITY INTEGRATION"
```
sfc-gh-asawicki: As always when adding the object type:
- aren't there more places it needs to be added in this file?
- should it be added to some collections (used in grants, etc.)?

pkg/testacc/resource_oauth_integration_for_partner_applications_acceptance_test.go
sfc-gh-asawicki: +1

pkg/acceptance/bettertestspoc/assert/resourceassert/oauth_integration_for_custom_clients_resource_ext.go
sfc-gh-jcieslak: Done

pkg/testacc/resource_oauth_integration_for_partner_applications_acceptance_test.go
sfc-gh-jcieslak: Reverted

pkg/acceptance/bettertestspoc/assert/objectassert/security_integration_snowflake_ext.go
sfc-gh-jcieslak: Hmm, I didn't notice at first that it's in a different package. Moved all asserts there and created a common assert struct

pkg/sdk/object_types.go
sfc-gh-jcieslak: Added a comment that it's a pseudo object type, similarly to the PAT (I saw that there's also storage integration that, in my opinion, should be treated as here). It's a pseudo role, as currently in the object assert generator, we are not using the passed object type in the `sdk_object_def.go`. The objectType there is auto-generated, so we for now either:
- Leave it as manual change (SecurityIntegration -> Integration), but I didn't want to add manually adjusted files
- Add pseudo object type (at least for the time being, until the object assertions are changed). For now, I only added a comment, but I believe we should also add a TODO with ticket number, but I'm not sure which one should be the most suitable for this.

pkg/testacc/resource_oauth_integration_for_custom_clients_acceptance_test.go
sfc-gh-jmichalak: Okay, I think that in some resources we just have a separate step in `Base` for changing ForceNew parameters.

PR #4140, opened 2025-10-28
chore: Regenerate all SDK objects part3

sfc-gh-asawicki:
Continuation to https://github.com/snowflakedb/terraform-provider-snowflake/pull/4139.

- General
  - Mark the manually modified code
  - Adjust the definitions
  - Add missing unit tests
  - Adjust dto builders usage if needed (pointer -> no pointer in `With` methods)
  - Regeneration problems (ext files, interface used in client) happen because of direct import of sdk package inside the generator - we need to reorganize this (planned for the future)
  - There are sometimes problems with proper mappings when the same struct is used on different levels - it will be investigated as a follow-up (marked below as bad mapping problems)
  - Unit tests improved generation must be handled separately
    - unit tests - default and all options with variants
    - unit tests - all alter branches should be added
    - unit tests - basic options for show/describe?
- `ExternalFunctions`
  - missing option in the showbyid mapping - added
  - `ExternalFunction` `ID()` added manually - no logic for this identifier type?
  - impl: find first logic using fully qualified names instead of `Name` - we should handle this identifier type in templates
  - impl: wrong type in describe (`DescribeExternalFunctionOptions` -> `DescribeFunctionOptions`)
  - impl: to opts mapping adjustment (name without arguments)
  - impl: conversion between the same structs
  - unit tests: generated id using `randomSchemaObjectIdentifierWithArguments` and `randomSchemaObjectIdentifier` before - should we keep the generated one?
  - unit tests: no arguments describe - is it correct?
- `Functions`
  - ext file preventing clean regenerate
  - usage of `SecretsList`/`SecretsListRequest`/`DescribeFunctionOptions` in other files prevents clean regeneration (move to commons? also, do not depend on the sdk package with the generator)
  - dto builders: incorrect name: `import` - renamed to `FunctionImport` (dto builders change and invocations `WithImport` -> `WithFunctionImport`)
  - dto builders: incorrect name: `package` - renamed temporarily to `FunctionPackage` (dto builders change and invocations `WithPackage` -> `WithFunctionPackage`)
  - updated def: missing columns
  - additional methods in the `Functions` interface (`DescribeDetails` and `ShowParameters`)
  - validations: a lot validations modified/added (e.g. for argument range)
  - impl: find first logic using fully qualified names instead of `Name` - we should handle this identifier type in templates
  - unit tests: generated id using `randomSchemaObjectIdentifierWithArguments` and `randomSchemaObjectIdentifier before` - should we keep the generated one?
- `Procedures`
  - import and package like with functions - renamed (no need to adjust usage as both were required, so no With builders)
  - additional methods in the `Procedures` interface (`DescribeDetails` and `ShowParameters`)
  - `SecretsList` redeclared - we could extract to commons and reuse
  - dto: `SecretsListRequest`, `NewSecretsListRequest` redeclared - we could extract to commons and reuse
  - validations: a lot validations modified/added (e.g. for argument range)
  - showbyid `In` -> `ExtendedIn`
  - impl: find first logic using fully qualified names instead of `Name` - we should handle this identifier type in templates
  - impl a few adjustments in `toOpts` logic
  - unit tests: some validation tests miss `- more present`
  - def: `ShowById` in `_def.go` could be adjusted as we support `ShowByIDExtendedInFiltering`
- `RowAccessPolicies`
  - `Type` in `CreateRowAccessPolicyArgs` renamed to `DataType` (type problematic in dto builders)
  - `GetName` added manually
  - conversion between the same structs
- `SecurityIntegrations`
  - default: `ObjectTypeSecurityIntegration` -> `ObjectTypeIntegration`
  - validations - two added manually
  - bad mappings: e.g. `r.Set.BlockedRolesList.BlockedRolesList` -> `r.ExternalOauthBlockedRolesList.BlockedRolesList`
  - dtos and builders: `*AccountObjectIdentifier` used for `NetworkPolicy` -> multiple places
  - dtos and builders: comment string allow empty -> string

### DIFF
--- a/pkg/acceptance/helpers/notebook_client.go
+++ b/pkg/acceptance/helpers/notebook_client.go
@@ -25,6 +25,7 @@ func (c *NotebookClient) client() sdk.Notebooks {
 }
 
 func (c *NotebookClient) Create(t *testing.T) (*sdk.Notebook, func()) {
+	t.Helper()
 	id := c.ids.RandomSchemaObjectIdentifier()
 	t.Helper()
 	return c.CreateWithRequest(t, sdk.NewCreateNotebookRequest(id))


pkg/sdk/procedures_gen_test.go
```diff
@@ -1694,62 +1855,254 @@ func TestProcedures_CreateAndCallForScala(t *testing.T) {
...
+
+	t.Run("validation: exactly one field from [opts.Arguments.ArgDataTypeOld opts.Arguments.ArgDataType] should be present", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.Arguments = []ProcedureArgument{
+			{ArgName: "arg"},
+		}
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("CreateAndCallForJavaScriptProcedureOptions.Arguments", "ArgDataTypeOld", "ArgDataType"))
+	})
+
+	t.Run("validation: exactly one field from [opts.Arguments.ArgDataTypeOld opts.Arguments.ArgDataType] should be present", func(t *testing.T) {
```
sfc-gh-jcieslak, line 1912: nit: `- more present`

pkg/sdk/procedures_impl_gen.go
```diff
@@ -68,12 +70,14 @@ func (v *procedures) Show(ctx context.Context, request *ShowProcedureRequest) ([
 
 func (v *procedures) ShowByID(ctx context.Context, id SchemaObjectIdentifierWithArguments) (*Procedure, error) {
 	request := NewShowProcedureRequest().
-		WithIn(ExtendedIn{In: In{Schema: id.SchemaId()}}).
-		WithLike(Like{Pattern: String(id.Name())})
+		WithLike(Like{Pattern: String(id.Name())}).
+		// adjusted manually
+		WithIn(ExtendedIn{In: In{Schema: id.SchemaId()}})
```
sfc-gh-jcieslak, line 75: nit: ShowById in `_def.go` could be adjusted as we support `ShowByIDExtendedInFiltering`

sfc-gh-jcieslak (dismissed):

pkg/sdk/procedures_gen_test.go
sfc-gh-jmichalak: Check other places in this file, and maybe in functions as well.

sfc-gh-jmichalak (approved):

pkg/sdk/procedures_gen_test.go
sfc-gh-asawicki: Added to follow-up list

pkg/sdk/procedures_impl_gen.go
sfc-gh-asawicki: Added to follow-up list

PR #4144, opened 2025-10-28
chore: Regenerate all SDK objects part4

sfc-gh-asawicki:
Continuation to https://github.com/snowflakedb/terraform-provider-snowflake/pull/4140.

- General
  - Mark the manually modified code
  - Adjust the definitions
  - Add missing unit tests
  - Adjust dto builders usage if needed (pointer -> no pointer in `With` methods)
  - Regeneration problems (ext files, interface used in client) happen because of direct import of sdk package inside the generator - we need to reorganize this (planned for the future)
  - There are sometimes problems with proper mappings when the same struct is used on different levels - it will be investigated as a follow-up (marked below as bad mapping problems)
  - Unit tests improved generation must be handled separately
    - unit tests - default and all options with variants
    - unit tests - all alter branches should be added
    - unit tests - basic options for show/describe?
 - `Stages`
   - clean regenerate not possible due to tables reusing parts of the domain (`StageFileFormatRequest`, `StageCopyOptionsRequest`, `StageFileFormat`, `StageCopyOptions`)
   - `type` is a reserved keyword, renamed to `FileFormatType` and `EncryptionType`
   - `continue` is a reserved keyword, renamed to `Continue_`
   - Fixed tags in def
   - `ID()` based on request added manually
   - adjusted `SkipFile` (and builder methods) - this can be changed in the def later
   - adjusted `FileFormatTypeOptions` -> `FileFormatTypeOptionsRequest`
- `StorageIntegrations`
  - ext preventing a clean regenerate (`S3Protocol` used from ext in def)
  - bad file name for dto builders and for unit tests
  - Changed `UsePrivateLinkEndpoint` to `UsePrivatelinkEndpoint`
- `Tasks`
  - clean regenerate not possible because of `grants_impl` (`ShowByID`, `SuspendRootTasks`, `ResumeTasks`, `Task`, `ShowTaskRequest`, `NewShowTaskRequest`, and more)
  - updated def
  - `IsStarted` added manually
  - `TaskSchedule` added manually
  - `ParseTaskSchedule` added manually
  - `Config` and `Warehouse` in `toOpts` adjusted manually
  - second same `convert` removed manually
  - validations: additional check for session params set needed - can we define it so that it's generated correctly?
  - dtos: added `GetName` for requests
  - dtos: added new struct (`CreateTaskWarehouseRequest`) and updated some of the generated struct attributes
  - unit tests: `err` for session parameters validation adjusted manually
  - unit tests: it seems like some validations could be added to def (like `opts.Warehouse.Warehouse conflicting with opts.Warehouse.UserTaskManagedInitialWarehouseSize`)
- `UserProgrammaticAccessTokens`
  - clean regenerate not possible because of usage in users
  - `ID()` added manually
  - `ObjectType` adjusted manually
  - `RemoveByIDSafely` added manually
  - Both `ShowByID` and `ShowByIDSafely` adjusted manually
  - validations adjusted manually
  - impl: lots of changes due to adjustments in the interface
  - unit tests: errors adjustments
  - unit tests: setup adjustments
- `Views`
  - set/unset tags adjusted manually (later fix in def)
  - column name (`"null?"`) adjusted manually
  - validation: invalid identifier errors changed
  - validation: validation for `opts.RowAccessPolicy.On` added manually (should be possible to set through def?)
  - validation: manually added for columns
  - bad mappings
  - dto: `GetName` added manually
  - unit tests: all options test for create uses request instead of opts
- `GitRepositories`
  - second `convert` removed manually
- `SemanticViews`
  - renamed the files
  - validations added manually
  - dto: used request structs instead of opts
  - dto: added a lot of request structs because of above (why these structs are not created by default? list?)
  - impl: adjusted mapping to changes to requests (instead of opts)
  - dto builders: adjusted manually to the changes to requests
  - dto builders: some methods generated incorrectly, e.g. `WithsemanticViewFacts` (semantic lowercase)
  - unit tests: validation tests added manually - we should try defining them (their names should be altered too)
  - unit tests: ordering changed

### DIFF
--- a/pkg/acceptance/helpers/notebook_client.go
+++ b/pkg/acceptance/helpers/notebook_client.go
@@ -27,7 +27,6 @@ func (c *NotebookClient) client() sdk.Notebooks {
 func (c *NotebookClient) Create(t *testing.T) (*sdk.Notebook, func()) {
 	t.Helper()
 	id := c.ids.RandomSchemaObjectIdentifier()
-	t.Helper()
 	return c.CreateWithRequest(t, sdk.NewCreateNotebookRequest(id))
 }
 


pkg/sdk/stages_dto_builders_gen.go
```diff
@@ -14,162 +15,172 @@ func NewCreateInternalStageRequest(
...
-func (s *StageCopyOnErrorOptionsRequest) WithContinue(Continue *bool) *StageCopyOnErrorOptionsRequest {
-	s.Continue = Continue
+func (s *StageCopyOnErrorOptionsRequest) WithContinue_(continue_ bool) *StageCopyOnErrorOptionsRequest {
+	s.Continue_ = &continue_
 	return s
 }
 
+// WithSkipFile removed manually
+
+// added manually
```
sfc-gh-jmichalak, line 164: Nit: Why is it removed manually, and here added manually?

sfc-gh-asawicki: Because of the different type (in def it's a keyword bool for now, and it can contain changes to the SQL, so it should be a string - it will be improved later)

sfc-gh-jcieslak (approved):

PR #4148, opened 2025-10-29
chore: Update docker-compose for test output processing

sfc-gh-jcieslak:
Because we planned to use the test output processing like this:
```shell
make run-tests < go run test-output-script > output_file.csv
```

I had to adjust the approach so that it could run on CI. The most important part is that, because the go CLI is a dependency only accessible through a Docker container, I had to create another service for our test output processor. To share the data between the services, I created a shared volume between them that exists within the containers.

From the CI pov it will be invoked like so (`--silent` in make doesn't print what underlying command is being run; it's important as otherwise this command would be put into resulting csv):
```shell
make --silent process-test-output-docker-compose 1> output.csv
```

### DIFF
--- a/packaging/docker-compose.yml
+++ b/packaging/docker-compose.yml
@@ -13,8 +13,6 @@ services:
     working_dir: /terraform-provider-snowflake
     command: sh -c "make --silent test-main-terraform-use-cases > /shared/test_output.json"
   process-test-output:
-    depends_on:
-      - test-main-terraform-use-cases
     build:
       context: .
       dockerfile: Dockerfile


packaging/docker-compose.yml
```diff
@@ -9,5 +9,17 @@ services:
     volumes:
       - ./../:/terraform-provider-snowflake
       - ${SNOWFLAKE_CONFIG_PATH}:/root/.snowflake/config
+      - shared_data:/shared
     working_dir: /terraform-provider-snowflake
-    command: [ "make", "test-main-terraform-use-cases" ]
+    command: sh -c "make --silent test-main-terraform-use-cases > /shared/test_output.json"
+  process-test-output:
```
sfc-gh-jmichalak, line 15: For later, we could add a dependency to force the correct order. For now, we can comment that the order is correct because of the Jenkins setup.

sfc-gh-jcieslak: I actually don't want to set the dependency. I had it initially, but it doesn't work like I expected. I thought the second service will run after the other, but it will only make sure it is running, so when you run only the test_output_processor (without the actual test) what will happen is: The test service will run (which is nice), but after it is running the test_output_processor container will run, it will read an empty file or file that is not there yet and will output CSV with only headers (which is not nice). So even if those services are dependent, setting up dependency won't make a difference; it just has to be called correctly, unfortunately.

sfc-gh-jmichalak: I think there is a flag to run the container after the other one is completed.

PR #4154, opened 2025-10-30
chore: Improve the cache logic

sfc-gh-asawicki:
Before this change, we had only a one-entry cache. Trying to initialize the provider in another way required setting the `SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE` environment variable, effectively refreshing the cache. The problem with this approach was that every such test required refreshing the cache, was slower, and another refresh was required in the test that followed (if it was using the "normally" configured provider). In this new approach, we have a cache containing multiple entries, which should improve the performance and stability of the tests. There are a few predefined factories (e.g., all view tests require the warehouse in the session), but it's also possible to define a new dedicated entry in a test or use the no cache factory. This is the start but there are a few follow-ups for later, to make it better.

- Simplify the provider initialization caching logic (acceptance tests and PoCs)
- Remove `SF_TF_ACC_TEST_CONFIGURE_CLIENT_ONCE` and `testenvs.ConfigureClientOnce` entirely
- Extract a simple `providerInitializationCache`

Needed follow-ups (mostly documented in SNOW-2661409):
- check all the places using TestAccProvider directly
- secondary account can also have a different configuration, so for now we need to be careful; let's add some hash check for the config or something else to mitigate
- check which of the `providerFactoryWithoutCache` usages want to really be without cache and which could utilize a dedicated cache entry
- fix a few tests still requiring the warehouse in session

### DIFF
--- a/pkg/testacc/resource_tag_acceptance_test.go
+++ b/pkg/testacc/resource_tag_acceptance_test.go
@@ -16,8 +16,6 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config/model"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers/random"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvs"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeroles"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"


pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -333,8 +332,6 @@ func TestAcc_Tag_Rename(t *testing.T) {
 }
 
 func TestAcc_Tag_migrateFromVersion_0_98_0(t *testing.T) {
-	t.Setenv(string(testenvs.ConfigureClientOnce), "")
-
 	id := testClient().Ids.RandomSchemaObjectIdentifier()
```
sfc-gh-jmichalak, line 260: I'm confused about cases like these, because in the PR desc it says 
> check which of the providerFactoryWithoutCache usages want to really be without cache and which could utilize a dedicated cache entry

So, it's done here?

sfc-gh-jmichalak (dismissed):

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-asawicki: Partially:
- all `ConfigureClientOnce` were initially removed and replaced with the no cache solution
- then, the main use cases like required warehouse, secondary, and authentication, were extracted as dedicated factories
- The ones that are left would be okay to leave them like this (this is the old behavior of initializing the provider for every step), but they could be improved if they reused the same config between the steps (which is not always true).

sfc-gh-jmichalak (dismissed):

sfc-gh-jcieslak (dismissed):

PR #4165, opened 2025-11-03
chore: Adjust data source tests follow up 2

sfc-gh-jcieslak:
Follow up for https://github.com/snowflakedb/terraform-provider-snowflake/pull/4115, addressing the following data sources
- `secrets`
- `security integrations`
- `streamlits`
- `schemas`


pkg/testacc/data_source_streamlits_acceptance_test.go
```diff
@@ -18,6 +19,175 @@ import (
...
+	streamlitId3 := testClient().Ids.RandomSchemaObjectIdentifier()
+
+	mainFile := random.AlphaN(4)
+
+	streamlitModel1 := model.StreamlitWithIds("test", streamlitId1, mainFile, stage.ID())
+	streamlitModel2 := model.StreamlitWithIds("test1", streamlitId2, mainFile, stage.ID())
+	streamlitModel3 := model.StreamlitWithIds("test2", streamlitId3, mainFile, stage.ID())
+
+	datasourceModelLikeExact := datasourcemodel.Streamlits("test").
+		WithLike(streamlitId1.Name()).
```
sfc-gh-jmichalak, line 39: Q: why we have disabled DESC/SHOW PARAMETERS in schemas? Maybe we could use it in other places to speed up tests, because in filtering cases we aren't asserting these outputs anyway.

pkg/testacc/data_source_security_integrations_acceptance_test.go
```diff
@@ -21,664 +23,418 @@ import (
...
+					assert.Check(resource.TestCheckResourceAttr(apiAuthClientCredentialsWithDescribe.DatasourceReference(), "security_integrations.0.describe_output.0.comment.0.value", comment)),
 				),
 			},
 		},
 	})
 }
 
-func TestAcc_SecurityIntegrations_Saml2(t *testing.T) {
-	id := testClient().Ids.RandomAccountObjectIdentifier()
+func TestAcc_SecurityIntegrations_MultipleTypes(t *testing.T) {
```
sfc-gh-jmichalak, line 388: What's the value of this test now? I believe everything is covered in the test above?

sfc-gh-asawicki (approved):

pkg/testacc/data_source_streamlits_acceptance_test.go
sfc-gh-jcieslak: I agree, adjusted

pkg/testacc/data_source_security_integrations_acceptance_test.go
sfc-gh-jcieslak: I wanted to leave it, because it tests whether this data source can query security integrations of different types. Perhaps it should be added in other similar cases, but since this test was already present and I thought it had some value, I didn't remove it.

sfc-gh-jmichalak (approved):

Issue #4010, opened 2025-09-08
[Bug]: Creating a snowflake_current_account is failing with SAML related error message


leopasta-enable:
### Terraform CLI Version

1.5.5

### Terraform Provider Version

2.6.0

### Terraform Configuration

```terraform
resource "snowflake_network_policy" "workplaces" {
  name                      = "ENABLE_OFFICES_AND_VPNS"
  allowed_network_rule_list = [
    module.platform_management_assets.network_rule_enable_offices_fqn,
    module.platform_management_assets.network_rule_enable_vpns_fqn
    ]
  comment                   = "Enable offices and vpns"
}

resource "snowflake_current_account" "account_settings" {
  network_policy = snowflake_network_policy.workplaces.fully_qualified_name
}
```

### Category

category:resource

### Object type(s)

resource:current_account

### Expected Behavior

1. Network policy to be set
2. `terraform apply` to complete successfully

### Actual Behavior

Terraform apply fails with the error:
> module.account_assets.snowflake_current_account.account_settings: Destroying... [id=current_account]
> ╷
> │ Error: 390959 (22023): Use security integration type saml instead of account parameter.

Terraform planning the resource to be replaced:

```

Terraform will perform the following actions:

  # module.account_assets.snowflake_current_account.account_settings is tainted, so must be replaced
-/+ resource "snowflake_current_account" "account_settings" {
      ~ abort_detached_query                                       = false -> (known after apply)
      + active_python_profiler                                     = (known after apply)
      ~ allow_client_mfa_caching                                   = false -> (known after apply)
      ~ allow_id_token                                             = false -> (known after apply)
      ~ autocommit                                                 = true -> (known after apply)
      + base_location_prefix                                       = (known after apply)
      ~ binary_input_format                                        = "HEX" -> (known after apply)
      ~ binary_output_format                                       = "HEX" -> (known after apply)
      + catalog                                                    = (known after apply)
      + catalog_sync                                               = (known after apply)
      ~ client_enable_log_info_statement_parameters                = false -> (known after apply)
      ~ client_encryption_key_size                                 = 128 -> (known after apply)
      ~ client_memory_limit                                        = 1536 -> (known after apply)
      ~ client_metadata_request_use_connection_ctx                 = false -> (known after apply)
      ~ client_metadata_use_session_database                       = false -> (known after apply)
      ~ client_prefetch_threads                                    = 4 -> (known after apply)
[...]

PR #4172, opened 2025-11-05
chore: Remove account parameter

sfc-gh-jcieslak:
Related to #4010, removing deprecated account parameter

The `snowflake_account_parameter` was not adjusted as removed parameter wasn't previously listed as valid parameter there anyway.

Note: tested with and without state upgrader that would explicitly remove the field with `delete` function in Go, but that didn't seem to change anything (both versions were passing the test)

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -26,6 +26,17 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ## v2.11.x ➞ v2.12.0
 
+### *(breaking change)* The removal of `SAML_IDENTITY_PROVIDER` from `snowflake_current_account` and `snowflake_current_organization_account` resources
+
+Due to changes on the Snowflake side, the `SAML_IDENTITY_PROVIDER` parameter is now deprecated and cannot be used in Snowflake (see [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/parameters#saml-identity-provider)).
+Because of this, we have removed support for this parameter in the `snowflake_current_account` and `snowflake_current_organization_account` resources.
+Both of the resources are in preview, so we decided to introduce this change now despite being a breaking change as it makes those resources unusable without workarounds (see related issue).
+
+If you were using this parameter in your configuration, please follow instructions in the linked documentation to migrate away from it.
+If you were not using this parameter, no changes are required.
+
+Related: [#4010](https://github.com/snowflakedb/terraform-provider-snowflake/issues/4010)
+
 ### *(improvement)* snowflake_scim_integration now accepts custom role names for run_as_role
 
 Previously, the `run_as_role` field in the [snowflake_scim_integration](https://registry.terraform.io/providers/snowflakedb/snowflake/2.11.0/docs/resources/scim_integration) resource only accepted predefined role names: `OKTA_PROVISIONER`, `AAD_PROVISIONER`, or `GENERIC_SCIM_PROVISIONER`.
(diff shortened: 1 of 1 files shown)


MIGRATION_GUIDE.md
```diff
@@ -24,6 +24,17 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.10.2 ➞ v2.11.0
+
+### *(breaking change)* The removal of `SAML_IDENTITY_PROVIDER` from `snowflake_current_account` and `snowflake_current_organization_account` resources
+
+Due to changes on the Snowflake side, the `SAML_IDENTITY_PROVIDER` parameter is now deprecated (see [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/parameters#saml-identity-provider)).
```
sfc-gh-jmichalak: Let's emphasize that:
- not only it is deprecated, but it can't be used in SF (it fails with an error),
- the affected resources are in preview, so we don't need to bump the major version.

MIGRATION_GUIDE.md
```diff
@@ -24,6 +24,17 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
...
+Because of this, we have removed support for this parameter in the `snowflake_current_account` and `snowflake_current_organization_account` resources.
+If you were using this parameter in your configuration, please follow instructions in the linked documentation to migrate away from it.
```
sfc-gh-jmichalak, line 35: Nit: this is a bit confusing because I understood that the config needs to be adjusted. Let's make it clear that 
- no change in TF config is necessary (we assume that no one used this field because of the errors),
- if the users want to migrate away from this feature in Snowflake, then <...>. (I'd keep this separation because it's not related to the provider).

pkg/testacc/resource_current_account_acceptance_test.go
```diff
@@ -1018,3 +1019,40 @@ func TestAcc_CurrentAccount_Complete(t *testing.T) {
 		},
 	})
 }
+
+func TestAcc_CurrentAccount_migrateFromV2_10_0(t *testing.T) {
```
sfc-gh-jmichalak, line 1023: Nit: v2_10_2, for the other test as well.

sfc-gh-jmichalak (approved):

MIGRATION_GUIDE.md
sfc-gh-asawicki: q: Is it currently (before the change) possible to have this attribute in the terraform config without error?

sfc-gh-jcieslak: According to this page: https://docs.snowflake.com/en/user-guide/admin-security-fed-auth-configure-snowflake and `Snowflake will continue to support these deprecated parameters as long as there are implementations that use them.` it should be possible. This document also mentions SSO_LOGIN_PAGE which we should look into.

sfc-gh-asawicki: but could it be run successfully? I think it was failing recently but I dont know the current status.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Done

MIGRATION_GUIDE.md
sfc-gh-jcieslak: SAML_IDENTITY_PROVIDER cannot be set, it's failing with `Use security integration type saml instead of account parameter.` The SSO_LOGIN_PAGE can be set and it's not failing (strange, because they were announced as deprecated together as linked above, but that's how it works currently).

pkg/testacc/resource_current_account_acceptance_test.go
sfc-gh-jcieslak: My mistake, but didn't change due to cumbersome test setup (I know this works, thus 2_10_2 should work, but didn't want to change and validate it locally as it takes some time to set up). I would say, let's leave it like this, as it validates what it supposed to.

PR #4181, opened 2025-11-07
chore: Add support for desc assertions

sfc-gh-jcieslak:
## Changes
- Add support for adding assertions based on desc command (in sdk and resource output assertions)
- Generate example assertions based on listing desc output

## Next prs
- Check if an enum could be introduced instead of boolean (to enable other outputs, e.g. some objects have multiple SHOW and DESC and other query commands that could be supported with new enum or similar solution)
- Remove ObjectType usage inside the assertions (use string or sdk.ObjectType("") instead), to avoid having to specify object types that don't make sense from the Snowflake pov

### DIFF
--- a/pkg/acceptance/bettertestspoc/README.md
+++ b/pkg/acceptance/bettertestspoc/README.md
@@ -62,7 +62,8 @@ func (u *UserShowOutputAssert) HasNameAndLoginName(expected string) *UserShowOut
 ### Adding new resource describe output assertions
 Resource describe output assertions can be generated automatically. For object `abc` do the following:
 - add object you want to generate to `allStructs` slice in the `assert/objectassert/gen/sdk_object_def.go`
-- to add custom (not generated assertions) create file `abc_show_output_ext.go` in the `assert/resourceshowoutputassert` package. Example would be:
+  - remember to mark the struct with `IsDescribeOutput: true` in the `sdkObjectDef` definition
+- to add custom (not generated assertions) create file `abc_desc_output_ext.go` in the `assert/resourceshowoutputassert` package. Example would be:
 ```go
 func (u *UserShowOutputAssert) HasNameAndLoginName(expected string) *UserShowOutputAssert {
 	return u.
--- a/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/assertions.tmpl
+++ b/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/assertions.tmpl
@@ -3,7 +3,7 @@
 {{- $assertVar := FirstLetterLowercase .Name | FirstLetter }}
 
 {{- $assertName := .Name | printf "%sShowOutputAssert" }}
-{{- if .IsDataSourceOutput }}
+{{- if .IsDescribeOutput }}
     {{- $assertName = TrimSuffix .Name "Details" | printf "%sDescribeOutputAssert" }}
 {{- end}}
 
--- a/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/definition.tmpl
+++ b/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/definition.tmpl
@@ -4,7 +4,7 @@
 {{- $outputType := "Show"}}
 {{- $assertName := $name | printf "%sShowOutputAssert" }}
 
-{{- if .IsDataSourceOutput }}
+{{- if .IsDescribeOutput }}
     {{- $name = TrimSuffix .Name "Details" }}
 
     {{- $outputType = "Describe" }}
--- a/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/main/main.go
+++ b/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/main/main.go
@@ -31,7 +31,7 @@ func main() {
 }
 
 func getFilename(_ genhelpers.SdkObjectDetails, model gen.ResourceShowOutputAssertionsModel) string {
-	if model.IsDataSourceOutput {
+	if model.IsDescribeOutput {
 		return strings.TrimSuffix(genhelpers.ToSnakeCase(model.Name), "_details") + "_desc_output" + "_gen.go"
 	}
 	return genhelpers.ToSnakeCase(model.Name) + "_show_output" + "_gen.go"
--- a/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/model.go
+++ b/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/model.go
@@ -7,9 +7,9 @@ import (
 )
 
 type ResourceShowOutputAssertionsModel struct {
-	Name               string
-	IsDataSourceOutput bool
-	Attributes         []ResourceShowOutputAssertionModel
+	Name             string
+	IsDescribeOutput bool
+	Attributes       []ResourceShowOutputAssertionModel
 
 	*genhelpers.PreambleModel
 }
@@ -29,10 +29,10 @@ func ModelFromSdkObjectDetails(sdkObject genhelpers.SdkObjectDetails, preamble *
 
 	name, _ := strings.CutPrefix(sdkObject.Name, "sdk.")
 	return ResourceShowOutputAssertionsModel{
-		Name:               name,
-		IsDataSourceOutput: sdkObject.IsDataSourceOutput,
-		Attributes:         attributes,
-		PreambleModel:      preamble,
+		Name:             name,
+		IsDescribeOutput: sdkObject.IsDataSourceOutput,
+		Attributes:       attributes,
+		PreambleModel:    preamble,
 	}
 }
 


pkg/acceptance/bettertestspoc/assert/objectassert/listing_desc_snowflake_gen.go
```diff
@@ -0,0 +1,819 @@
...
+)
+
+type ListingDetailsAssert struct {
+	*assert.SnowflakeObjectAssert[sdk.ListingDetails, sdk.AccountObjectIdentifier]
+}
+
+func ListingDetails(t *testing.T, id sdk.AccountObjectIdentifier) *ListingDetailsAssert {
+	t.Helper()
+	return &ListingDetailsAssert{
+		assert.NewSnowflakeObjectAssertWithTestClientObjectProvider(sdk.ObjectTypeListingDetails, id, func(testClient *helpers.TestClient) assert.ObjectProvider[sdk.ListingDetails, sdk.AccountObjectIdentifier] {
```
sfc-gh-asawicki: The template currently uses `sdk.ObjectType{{ .Name }}` instead of the `ObjectType` from `SdkObjectDef`. The TODO (`// TODO [SNOW-2324252]: remove object type?`) had the question mark on purpose. The object type from the definition was not used at all, but perhaps it should be? This way, we could overwrite the object type for cases like the desc output or e.g., integrations. Let's discuss.

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/assertions.tmpl
```diff
@@ -1,9 +1,16 @@
 {{- /*gotype: github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen.ResourceShowOutputAssertionsModel*/ -}}
 
-{{- $assertName := .Name | printf "%sShowOutputAssert" -}}
-{{- $nameLowerCase := FirstLetterLowercase .Name -}}
+{{- $assertName := .Name | printf "%sShowOutputAssert" }}
+{{- $nameLowerCase := FirstLetterLowercase .Name }}
 {{- $assertVar := FirstLetter $nameLowerCase }}
 
+{{- if .IsDataSourceOutput }}
+    {{- $trimmedName := TrimSuffix .Name "Details" }}
```
sfc-gh-asawicki: nit: nameLowerCase and assertVar could be constructed once for both cases by just disinguishing by name (i.e. conditionally define $name - .Name or TrimSuffix .Name "Details")

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/definition.tmpl
```diff
@@ -1,29 +1,40 @@
-{{- /*gotype: github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceparametersassert/gen.ResourceParametersAssertionsModel*/ -}}
+{{- /*gotype: github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen.ResourceShowOutputAssertionsModel*/ -}}
 
-{{- $assertName := .Name | printf "%sShowOutputAssert" -}}
-{{- $nameLowerCase := FirstLetterLowercase .Name -}}
-{{- $assertVar := FirstLetterLowercase .Name | printf "%sAssert" }}
+{{- $name := .Name }}
+{{- $outputType := "Show"}}
+{{- $outputTypeLowerCase := FirstLetterLowercase $outputType }}
```
sfc-gh-asawicki: this is computed twice and both times we know the answer (Show -> show, Describe -> describe below). set output type conditionally and the this logic could be reused. Additionally, maybe the _output part could be incorporated as it;s repeated in the body

sfc-gh-asawicki (changes requested):

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/assertions.tmpl
sfc-gh-jcieslak: Slimmed this a bit

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/definition.tmpl
sfc-gh-jcieslak: True, extracted

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/assertions.tmpl
```diff
@@ -1,8 +1,11 @@
 {{- /*gotype: github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen.ResourceShowOutputAssertionsModel*/ -}}
 
-{{- $assertName := .Name | printf "%sShowOutputAssert" -}}
-{{- $nameLowerCase := FirstLetterLowercase .Name -}}
-{{- $assertVar := FirstLetter $nameLowerCase }}
+{{- $assertVar := FirstLetterLowercase .Name | FirstLetter }}
+
+{{- $assertName := .Name | printf "%sShowOutputAssert" }}
```
sfc-gh-jmichalak, line 5: Nit: this line should be under an `else`​ block.

pkg/acceptance/bettertestspoc/README.md
```diff
@@ -49,7 +49,19 @@ Here's a short description of every generated assertion:
...
+func (u *UserShowOutputAssert) HasNameAndLoginName(expected string) *UserShowOutputAssert {
+	return u.
+		HasName(expected).
+		HasLoginName(expected)
+}
+```
+
+### Adding new resource describe output assertions
+Resource describe output assertions can be generated automatically. For object `abc` do the following:
+- add object you want to generate to `allStructs` slice in the `assert/objectassert/gen/sdk_object_def.go`
```
sfc-gh-jmichalak, line 64: What about filling `IsDescribeOutput`​ here?

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/definition.tmpl
```diff
@@ -1,29 +1,39 @@
...
+{{- /*gotype: github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen.ResourceShowOutputAssertionsModel*/ -}}
 
-{{- $assertName := .Name | printf "%sShowOutputAssert" -}}
-{{- $nameLowerCase := FirstLetterLowercase .Name -}}
-{{- $assertVar := FirstLetterLowercase .Name | printf "%sAssert" }}
+{{- $name := .Name }}
+{{- $outputType := "Show"}}
+{{- $assertName := $name | printf "%sShowOutputAssert" }}
+
+{{- if .IsDataSourceOutput }}
```
sfc-gh-jmichalak: Same here.

pkg/acceptance/bettertestspoc/assert/objectassert/gen/main/main.go
```diff
@@ -30,5 +32,8 @@ func main() {
 }
 
 func getFilename(_ genhelpers.SdkObjectDetails, model gen.SnowflakeObjectAssertionsModel) string {
+	if model.IsDataSourceOutput {
```
sfc-gh-jmichalak, line 35: I think we should rename `IsDataSourceOutput`​ to `IsDescribeOutput`​ everywhere. Ultimately, I'm voting for an enum, but we can do this in a follow-up.

sfc-gh-jcieslak: Hmm, idk why I named this data source. I think I meant to describe,  but didn't notice...
Yeah, I'm also voting for enum, but wanted a quick simple solution for now. After all, probably not much will change after changing the type, but let's do it in a follow-up.

pkg/acceptance/bettertestspoc/README.md
sfc-gh-jcieslak: Added

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/definition.tmpl
sfc-gh-jcieslak: Renamed

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/gen/templates/assertions.tmpl
sfc-gh-jcieslak: This way it's simpler, because ifs introduce scope, meaning we would have to create an empty value first, then reassign it in the if or else block. Will be changed anyway with the introduction of enum.

pkg/acceptance/bettertestspoc/assert/objectassert/listing_desc_snowflake_gen.go
sfc-gh-asawicki: @sfc-gh-jcieslak I know we discussed this verbally but please add a short comment on what we decided with this one pls

sfc-gh-asawicki (approved):

pkg/acceptance/bettertestspoc/assert/objectassert/listing_desc_snowflake_gen.go
sfc-gh-jcieslak: We agreed to use for now `sdk.ObjectType("{{ .Name }}")` and in later iterations use plain string values as there's no benefit in using / passing `sdk.ObjectType` around (aligns with next prs in the PR description)

PR #4182, opened 2025-11-07
chore: Upgrade golangci-lint to v2.6.1

sfc-gh-jmichalak:
- Upgrade the linter from v1 to v2.6.1.
- Adjust the installation script to https://golangci-lint.run/docs/welcome/install/.
- Use the available [migration tool](https://golangci-lint.run/docs/product/migration-guide/).
- Remove not needed comments.
- Remove obsolete depguard configurations.
- Use autofix to address the findings.
- Add filepath cleaning.
- Use lint-fix target in pre-push instead of lint.
- Enable a few presets for the scripts package - read more in https://golangci-lint.run/docs/linters/false-positives/#exclusion-presets.

## TODO
- Follow up with adjusting the pre-push-check.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,6 +32,7 @@ linters:
     testifylint:
       disable:
         - require-error # Do not force using require instead of assert checks.
+        - empty # Do not force using Empty assertion instead of Equal assertion with empty string.
   exclusions:
     presets:
       - legacy


pkg/sdk/sql_builder_test.go
```diff
@@ -510,7 +510,7 @@ func TestBuilder_parseStruct(t *testing.T) {
 func TestBuilder_sql(t *testing.T) {
 	t.Run("test sql with no clauses", func(t *testing.T) {
 		s := builder.sql([]sqlClause{}...)
-		assert.Equal(t, "", s)
+		assert.Empty(t, s)
```
sfc-gh-asawicki: I think we had this discussion already with one of the previous bumps: Empty succeeds also for nil. I would prefer having the empty string checks explicit (also because the tf nature).

sfc-gh-asawicki (changes requested):

pkg/sdk/sql_builder_test.go
sfc-gh-asawicki: Let's discuss in the whole team

sfc-gh-jmichalak: Reverted.

PR #4195, opened 2025-11-13
feat: Add notebook resource

sfc-gh-mkowalski:
Added a [Notebook](https://docs.snowflake.com/en/sql-reference/commands-notebook) resource implementation.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -94,6 +94,23 @@ Previously, the provider's file reading functions did not clean paths. In this v
 
 No changes in configuration and state are required. The supported TOML location `~/.snowflake/config` stays the same and the behavior shouldn't be affected.
 
+### Task parameter validation handling
+
+Recently, Snowflake moved validation from runtime (task execution) to CREATE/ALTER operations for two parameters ([`AUTOCOMMIT`](https://docs.snowflake.com/en/sql-reference/parameters#autocommit) and [`SEARCH_PATH`](https://docs.snowflake.com/en/sql-reference/parameters#search-path)).
+Because of this, both parameters for tasks fail during those operations for invalid values.
+The `AUTOCOMMIT` parameter can be only set to `TRUE` (default value for this parameter), and `SEARCH_PATH` cannot be set at all. Both parameters can be unset.
+Now, when either `AUTOCOMMIT` is set to `FALSE` or `SEARCH_PATH` is set in the configuration (when creating or changing), the task resource will return warnings saying:
+
+```
+Invalid value for AUTOCOMMIT parameter: cannot be set to FALSE on a task
+Invalid value for SEARCH_PATH parameter: cannot be set on a task
+```
+
+If you have any of these parameters set in your configuration,
+please remove them to avoid the Terraform warnings and potential errors from the Snowflake side.
+The parameters may be removed in the next major version of the provider.
+Other than this, no changes in the configuration are required.
+
 ### *(bugfix)* Improved validation of identifiers with arguments
 Previously, during parsing identifiers with argument types, when the identifier format was incorrect, the provider could panic with errors like:
 ```
@@ -125,6 +142,43 @@ In a few releases, we will address other similar faulty object type cases in pri
 No changes in configuration are required.
 
 Community PR: [#4185](https://github.com/snowflakedb/terraform-provider-snowflake/pull/4185)
+
+### *(new feature)* Task seconds and hours scheduling support
+
+Added support for scheduling tasks using seconds and hours intervals in addition to the existing minutes and cron support. The `snowflake_task` resource now supports:
+
+- `schedule.seconds` - Schedule tasks to run at intervals specified in seconds (e.g., every 30 seconds)
+- `schedule.hours` - Schedule tasks to run at intervals specified in hours (e.g., every 2 hours)
+
+These new scheduling options work alongside the existing `schedule.minutes` and `schedule.using_cron` options. Only one scheduling method can be specified at a time.
+
+**Example usage:**
+
+```terraform
+resource "snowflake_task" "example_seconds" {
+  database      = "my_database"
+  schema        = "my_schema"
+  name          = "my_task_seconds"
+  sql_statement = "SELECT 1"
+  started       = true
+
+  schedule {
+    seconds = 30  # Run every 30 seconds
+  }
+}
+
+resource "snowflake_task" "example_hours" {
+  database      = "my_database"
+  schema        = "my_schema"
+  name          = "my_task_hours"
+  sql_statement = "SELECT 1"
+  started       = true
+
+  schedule {
+    hours = 2  # Run every 2 hours
+  }
+}
+```
 
 ## v2.10.0 ➞ v2.10.1
 
--- a/docs/resources/task.md
+++ b/docs/resources/task.md
@@ -9,6 +9,8 @@ description: |-
 
 !> **Warning** Due to complex conditional logic in Snowflake regarding required fields for a given type (serverless or not) and use case (scheduled, run after another task, conditional, or other) of a task, please make sure you went through the [Snowflake documentation for tasks](https://docs.snowflake.com/en/user-guide/tasks-intro) to understand which fields are required for your use case.
 
+!> **Warning** Setting `AUTOCOMMIT` to `false` and `SEARCH_PATH` to any value is now not possible for tasks in Snowflake, thus it's recommended to not configure them within the resource. The only exception is leaving (or setting) `AUTOCOMMIT` with `true` value explicitly set in the configuration, is to override the parameter that may be set to `false` value on the higher parameter hierarchy level. The parameters may be removed in the next major version of the provider.
+
 # snowflake_task (Resource)
 
 Resource used to manage task objects. For more information, check [task documentation](https://docs.snowflake.com/en/user-guide/tasks-intro).
@@ -25,6 +27,32 @@ resource "snowflake_task" "task" {
   started   = true
   schedule {
     minutes = 5
+  }
+  sql_statement = "select 1"
+}
+
+# Basic task with seconds schedule
+resource "snowflake_task" "task_seconds" {
+  database  = "database"
+  schema    = "schema"
+  name      = "task_seconds"
+  warehouse = "warehouse"
+  started   = true
+  schedule {
+    seconds = 30
+  }
+  sql_statement = "select 1"
+}
+
+# Basic task with hours schedule
+resource "snowflake_task" "task_hours" {
+  database  = "database"
+  schema    = "schema"
+  name      = "task_hours"
+  warehouse = "warehouse"
+  started   = true
+  schedule {
+    hours = 2
   }
   sql_statement = "select 1"
 }
@@ -200,7 +228,7 @@ resource "snowflake_task" "test" {
 - `quoted_identifiers_ignore_case` (Boolean) Specifies whether letters in double-quoted object identifiers are stored and resolved as uppercase letters. By default, Snowflake preserves the case of alphabetic characters when storing and resolving double-quoted identifiers (see [Identifier resolution](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#label-identifier-casing)). You can use this parameter in situations in which [third-party applications always use double quotes around identifiers](https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html#label-identifier-casing-parameter). For more information, check [QUOTED_IDENTIFIERS_IGNORE_CASE docs](https://docs.snowflake.com/en/sql-reference/parameters#quoted-identifiers-ignore-case).
 - `rows_per_resultset` (Number) Specifies the maximum number of rows returned in a result set. A value of 0 specifies no maximum. For more information, check [ROWS_PER_RESULTSET docs](https://docs.snowflake.com/en/sql-reference/parameters#rows-per-resultset).
 - `s3_stage_vpce_dns_name` (String) Specifies the DNS name of an Amazon S3 interface endpoint. Requests sent to the internal stage of an account via [AWS PrivateLink for Amazon S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/privatelink-interface-endpoints.html) use this endpoint to connect. For more information, see [Accessing Internal stages with dedicated interface endpoints](https://docs.snowflake.com/en/user-guide/private-internal-stages-aws.html#label-aws-privatelink-internal-stage-network-isolation). For more information, check [S3_STAGE_VPCE_DNS_NAME docs](https://docs.snowflake.com/en/sql-reference/parameters#s3-stage-vpce-dns-name).
-- `schedule` (Block List, Max: 1) The schedule for periodically running the task. This can be a cron or interval in minutes. (Conflicts with finalize and after; when set, one of the sub-fields `minutes` or `using_cron` should be set) For [Triggered tasks](https://docs.snowflake.com/en/user-guide/tasks-triggered), a schedule is not required. For other tasks, a schedule must be defined for a standalone task or the root task in a [task graph](https://docs.snowflake.com/en/user-guide/tasks-graphs#label-task-dag); otherwise, the task only runs if manually executed using [EXECUTE TASK](https://docs.snowflake.com/en/sql-reference/sql/execute-task) in, for example, the [snowflake_execute](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/execute) resource. A schedule cannot be specified for child tasks in a task graph. For more information on schedule restrictions, consult the [official documentation for Task object](https://docs.snowflake.com/en/user-guide/tasks-intro). (see [below for nested schema](#nestedblock--schedule))
+- `schedule` (Block List, Max: 1) The schedule for periodically running the task. This can be a cron or interval in seconds, minutes, or hours. (Conflicts with finalize and after; when set, one of the sub-fields `seconds`, `minutes`, `hours`, or `using_cron` should be set) For [Triggered tasks](https://docs.snowflake.com/en/user-guide/tasks-triggered), a schedule is not required. For other tasks, a schedule must be defined for a standalone task or the root task in a [task graph](https://docs.snowflake.com/en/user-guide/tasks-graphs#label-task-dag); otherwise, the task only runs if manually executed using [EXECUTE TASK](https://docs.snowflake.com/en/sql-reference/sql/execute-task) in, for example, the [snowflake_execute](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/execute) resource. A schedule cannot be specified for child tasks in a task graph. For more information on schedule restrictions, consult the [official documentation for Task object](https://docs.snowflake.com/en/user-guide/tasks-intro). (see [below for nested schema](#nestedblock--schedule))
 - `search_path` (String) Specifies the path to search to resolve unqualified object names in queries. For more information, see [Name resolution in queries](https://docs.snowflake.com/en/sql-reference/name-resolution.html#label-object-name-resolution-search-path). Comma-separated list of identifiers. An identifier can be a fully or partially qualified schema name. For more information, check [SEARCH_PATH docs](https://docs.snowflake.com/en/sql-reference/parameters#search-path).
 - `statement_queued_timeout_in_seconds` (Number) Amount of time, in seconds, a SQL statement (query, DDL, DML, etc.) remains queued for a warehouse before it is canceled by the system. This parameter can be used in conjunction with the [MAX_CONCURRENCY_LEVEL](https://docs.snowflake.com/en/sql-reference/parameters#label-max-concurrency-level) parameter to ensure a warehouse is never backlogged. For more information, check [STATEMENT_QUEUED_TIMEOUT_IN_SECONDS docs](https://docs.snowflake.com/en/sql-reference/parameters#statement-queued-timeout-in-seconds).
 - `statement_timeout_in_seconds` (Number) Amount of time, in seconds, after which a running SQL statement (query, DDL, DML, etc.) is canceled by the system. For more information, check [STATEMENT_TIMEOUT_IN_SECONDS docs](https://docs.snowflake.com/en/sql-reference/parameters#statement-timeout-in-seconds).
@@ -244,8 +272,10 @@ resource "snowflake_task" "test" {
 
 Optional:
 
-- `minutes` (Number) Specifies an interval (in minutes) of wait time inserted between runs of the task. Accepts positive integers only. (conflicts with `using_cron`)
-- `using_cron` (String) Specifies a cron expression and time zone for periodically running the task. Supports a subset of standard cron utility syntax. (conflicts with `minutes`)
+- `hours` (Number) Specifies an interval (in hours) of wait time inserted between runs of the task. Accepts positive integers. (conflicts with `seconds`, `minutes`, and `using_cron`)
+- `minutes` (Number) Specifies an interval (in minutes) of wait time inserted between runs of the task. Accepts positive integers. (conflicts with `seconds`, `hours`, and `using_cron`)
+- `seconds` (Number) Specifies an interval (in seconds) of wait time inserted between runs of the task. Accepts positive integers. (conflicts with `minutes`, `hours`, and `using_cron`)
+- `using_cron` (String) Specifies a cron expression and time zone for periodically running the task. Supports a subset of standard cron utility syntax. (conflicts with `seconds`, `minutes`, and `hours`)
 
 
 <a id="nestedblock--timeouts"></a>
--- a/examples/resources/snowflake_task/resource.tf
+++ b/examples/resources/snowflake_task/resource.tf
@@ -11,6 +11,32 @@ resource "snowflake_task" "task" {
   sql_statement = "select 1"
 }
 
+# Basic task with seconds schedule
+resource "snowflake_task" "task_seconds" {
+  database  = "database"
+  schema    = "schema"
+  name      = "task_seconds"
+  warehouse = "warehouse"
+  started   = true
+  schedule {
+    seconds = 30
+  }
+  sql_statement = "select 1"
+}
+
+# Basic task with hours schedule
+resource "snowflake_task" "task_hours" {
+  database  = "database"
+  schema    = "schema"
+  name      = "task_hours"
+  warehouse = "warehouse"
+  started   = true
+  schedule {
+    hours = 2
+  }
+  sql_statement = "select 1"
+}
+
 # Basic serverless task
 resource "snowflake_task" "serverless_task" {
   database                                 = "database"
--- a/pkg/acceptance/bettertestspoc/assert/resourceassert/task_resource_ext.go
+++ b/pkg/acceptance/bettertestspoc/assert/resourceassert/task_resource_ext.go
@@ -16,15 +16,39 @@ func (t *TaskResourceAssert) HasAfter(ids ...sdk.SchemaObjectIdentifier) *TaskRe
 	return t
 }
 
+func (t *TaskResourceAssert) HasScheduleSeconds(seconds int) *TaskResourceAssert {
+	t.AddAssertion(assert.ValueSet("schedule.#", "1"))
+	t.AddAssertion(assert.ValueSet("schedule.0.seconds", strconv.Itoa(seconds)))
+	t.AddAssertion(assert.ValueSet("schedule.0.minutes", "0"))
+	t.AddAssertion(assert.ValueSet("schedule.0.hours", "0"))
+	t.AddAssertion(assert.ValueSet("schedule.0.using_cron", ""))
+	return t
+}
+
 func (t *TaskResourceAssert) HasScheduleMinutes(minutes int) *TaskResourceAssert {
 	t.AddAssertion(assert.ValueSet("schedule.#", "1"))
 	t.AddAssertion(assert.ValueSet("schedule.0.minutes", strconv.Itoa(minutes)))
+	t.AddAssertion(assert.ValueSet("schedule.0.seconds", "0"))
+	t.AddAssertion(assert.ValueSet("schedule.0.hours", "0"))
+	t.AddAssertion(assert.ValueSet("schedule.0.using_cron", ""))
+	return t
+}
+
+func (t *TaskResourceAssert) HasScheduleHours(hours int) *TaskResourceAssert {
+	t.AddAssertion(assert.ValueSet("schedule.#", "1"))
+	t.AddAssertion(assert.ValueSet("schedule.0.hours", strconv.Itoa(hours)))
+	t.AddAssertion(assert.ValueSet("schedule.0.seconds", "0"))
+	t.AddAssertion(assert.ValueSet("schedule.0.minutes", "0"))
+	t.AddAssertion(assert.ValueSet("schedule.0.using_cron", ""))
 	return t
 }
 
 func (t *TaskResourceAssert) HasScheduleCron(cron string) *TaskResourceAssert {
 	t.AddAssertion(assert.ValueSet("schedule.#", "1"))
 	t.AddAssertion(assert.ValueSet("schedule.0.using_cron", cron))
+	t.AddAssertion(assert.ValueSet("schedule.0.seconds", "0"))
+	t.AddAssertion(assert.ValueSet("schedule.0.minutes", "0"))
+	t.AddAssertion(assert.ValueSet("schedule.0.hours", "0"))
 	return t
 }
 
--- a/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/task_show_output_ext.go
+++ b/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/task_show_output_ext.go
@@ -78,8 +78,18 @@ func (t *TaskShowOutputAssert) HasScheduleEmpty() *TaskShowOutputAssert {
 	return t
 }
 
+func (t *TaskShowOutputAssert) HasScheduleSeconds(seconds int) *TaskShowOutputAssert {
+	t.AddAssertion(assert.ResourceShowOutputValueSet("schedule", fmt.Sprintf("%d SECOND", seconds)))
+	return t
+}
+
 func (t *TaskShowOutputAssert) HasScheduleMinutes(minutes int) *TaskShowOutputAssert {
 	t.AddAssertion(assert.ResourceShowOutputValueSet("schedule", fmt.Sprintf("%d MINUTE", minutes)))
+	return t
+}
+
+func (t *TaskShowOutputAssert) HasScheduleHours(hours int) *TaskShowOutputAssert {
+	t.AddAssertion(assert.ResourceShowOutputValueSet("schedule", fmt.Sprintf("%d HOUR", hours)))
 	return t
 }
 
--- a/pkg/acceptance/bettertestspoc/config/datasourcemodel/streamlits_model_ext.go
+++ b/pkg/acceptance/bettertestspoc/config/datasourcemodel/streamlits_model_ext.go
@@ -6,6 +6,14 @@ import (
 	tfconfig "github.com/hashicorp/terraform-plugin-testing/config"
 )
 
+func (s *StreamlitsModel) WithInAccount() *StreamlitsModel {
+	return s.WithInValue(
+		tfconfig.ObjectVariable(map[string]tfconfig.Variable{
+			"account": tfconfig.BoolVariable(true),
+		}),
+	)
+}
+
 func (s *StreamlitsModel) WithInDatabase(databaseId sdk.AccountObjectIdentifier) *StreamlitsModel {
 	return s.WithInValue(
 		tfconfig.ObjectVariable(map[string]tfconfig.Variable{
--- a/pkg/acceptance/bettertestspoc/config/model/task_model_ext.go
+++ b/pkg/acceptance/bettertestspoc/config/model/task_model_ext.go
@@ -81,7 +81,40 @@ func (t *TaskModel) WithScheduleMinutes(minutes int) *TaskModel {
 
 func (t *TaskModel) WithScheduleCron(cron string) *TaskModel {
 	t.Schedule = tfconfig.MapVariable(map[string]tfconfig.Variable{
-		"cron": tfconfig.StringVariable(cron),
+		"using_cron": tfconfig.StringVariable(cron),
 	})
 	return t
+}
+
+func (t *TaskModel) WithScheduleSeconds(seconds int) *TaskModel {
+	t.Schedule = tfconfig.MapVariable(map[string]tfconfig.Variable{
+		"seconds": tfconfig.IntegerVariable(seconds),
+	})
+	return t
+}
+
+func (t *TaskModel) WithScheduleHours(hours int) *TaskModel {
+	t.Schedule = tfconfig.MapVariable(map[string]tfconfig.Variable{
+		"hours": tfconfig.IntegerVariable(hours),
+	})
+	return t
+}
+
+func (t *TaskModel) WithEmptySchedule() *TaskModel {
+	return t.WithScheduleValue(
+		tfconfig.ObjectVariable(map[string]tfconfig.Variable{
+			"any": tfconfig.StringVariable(string(config.SnowflakeProviderConfigSingleAttributeWorkaround)),
+		}),
+	)
+}
+
+func (t *TaskModel) WithMultipleSchedules() *TaskModel {
+	return t.WithScheduleValue(
+		tfconfig.ObjectVariable(map[string]tfconfig.Variable{
+			"hours":      tfconfig.IntegerVariable(1),
+			"minutes":    tfconfig.IntegerVariable(1),
+			"seconds":    tfconfig.IntegerVariable(1),
+			"using_cron": tfconfig.StringVariable("*/5 * * * * UTC"),
+		}),
+	)
 }
--- a/pkg/resources/task.go
+++ b/pkg/resources/task.go
@@ -69,7 +69,7 @@ var taskSchema = map[string]*schema.Schema{
 		Optional: true,
 		MaxItems: 1,
 		Description: joinWithSpace(
-			"The schedule for periodically running the task. This can be a cron or interval in minutes. (Conflicts with finalize and after; when set, one of the sub-fields `minutes` or `using_cron` should be set)",
+			"The schedule for periodically running the task. This can be a cron or interval in seconds, minutes, or hours. (Conflicts with finalize and after; when set, one of the sub-fields `seconds`, `minutes`, `hours`, or `using_cron` should be set)",
 			"For [Triggered tasks](https://docs.snowflake.com/en/user-guide/tasks-triggered), a schedule is not required. For other tasks,",
 			"a schedule must be defined for a standalone task or the root task in a [task graph](https://docs.snowflake.com/en/user-guide/tasks-graphs#label-task-dag); otherwise,",
 			"the task only runs if manually executed using [EXECUTE TASK](https://docs.snowflake.com/en/sql-reference/sql/execute-task) in, for example, the [snowflake_execute](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/execute) resource.",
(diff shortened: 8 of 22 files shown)


pkg/sdk/notebooks_gen.go
```diff
@@ -187,7 +187,7 @@ type Notebook struct {
 	QueryWarehouse *AccountObjectIdentifier
 	UrlId          string
 	OwnerRoleType  string
-	CodeWarehouse  AccountObjectIdentifier
+	CodeWarehouse  *AccountObjectIdentifier
```
sfc-gh-jmichalak: Why? The CodeWarehouse is always present (by default, it's `SYSTEM`... right)?

pkg/testacc/resource_git_repository_acceptance_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build non_account_level_tests
```
sfc-gh-jmichalak, line 1: This should be reverted.

MIGRATION_GUIDE.md
```diff
@@ -24,6 +24,12 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+### *(new feature)* snowflake_notebook resource
```
sfc-gh-jmichalak, line 29: The headers are incorrect. Now, according to the guide we'd release 2.10.2 because we had no feature changes after 2.10.1.

This change introduces a new feature, so this should be under "2.10.x -> 2.11.0".

templates/resources/notebook.md.tmpl
```diff
@@ -0,0 +1,38 @@
...
+subcategory: "Preview"
+description: |-
+{{ if gt (len (split .Description "<deprecation>")) 1 -}}
+{{ index (split .Description "<deprecation>") 1 | plainmarkdown | trimspace | prefixlines "  " }}
+{{- else -}}
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+{{- end }}
+---
+
+!> **Caution: Preview Feature** This feature is considered a preview feature in the provider, regardless of the state of the resource in Snowflake. We do not guarantee its stability. It will be reworked and marked as a stable feature in future releases. Breaking changes are expected, even without bumping the major version. To use this feature, add the relevant feature name to `preview_features_enabled` field in the [provider configuration](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs#schema). Please always refer to the [Getting Help](https://github.com/snowflakedb/terraform-provider-snowflake?tab=readme-ov-file#getting-help) section in our Github repo to best determine how to get help for your questions.
```
sfc-gh-jmichalak, line 16: Whenever we are not detecting external changes (like in the `from` field in this case), we should note that in the docs. Please follow other resources.

pkg/schemas/notebook_gen.go
```diff
@@ -0,0 +1,75 @@
...
+	if notebook.Comment != nil {
+		notebookSchema["comment"] = notebook.Comment
+	}
+	notebookSchema["owner"] = notebook.Owner
+	if notebook.QueryWarehouse != nil {
+		notebookSchema["query_warehouse"] = notebook.QueryWarehouse.Name()
+	}
+	notebookSchema["url_id"] = notebook.UrlId
+	notebookSchema["owner_role_type"] = notebook.OwnerRoleType
+	notebookSchema["code_warehouse"] = notebook.CodeWarehouse.Name()
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: This is related to the SDK change.

pkg/schemas/notebook.go
```diff
@@ -0,0 +1,192 @@
+// Code generated by SDK to schema generator (v0.1.0); DO NOT EDIT.
```
sfc-gh-jmichalak: I think this file is not connected to the generator pipeline, so pls remove this line.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+var notebookSchema = map[string]*schema.Schema{
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
```
sfc-gh-jmichalak: No need for ForceNew. We can handle renaming, look at streamlits for reference (d.HasChange("name")).

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+		Description:      blocklistedCharactersFieldDescription("The database in which to create the notebook."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"schema": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		Description:      blocklistedCharactersFieldDescription("The schema in which to create the notebook."),
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("code_warehouse"),
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: This should be adjusted.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies a comment for the notebook.",
+	},
+	"from": {
+		Type:             schema.TypeList,
+		Optional:         true,
+		ForceNew:         true,
+		Description:      "Specifies that the notebook should be created from an .ipynb file in the specified stage location.",
+		DiffSuppressFunc: suppressIdentifierQuoting,
```
sfc-gh-jmichalak: This diffsuppress is not needed - in the `stage` is enough.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+				},
+			},
+		},
+	},
+	"main_file": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ForceNew:         true,
+		Description:      "Specifies a user-specified identifier for the notebook file name.",
+		DiffSuppressFunc: suppressIdentifierQuoting,
```
sfc-gh-jmichalak: This is not a part of an ID, so `DiffSuppressFunc` is unnecessary.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+		Type:             schema.TypeString,
+		Optional:         true,
+		Description:      "Specifies the warehouse where SQL queries in the notebook are run.",
+		ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"idle_auto_shutdown_time_seconds": {
+		Type:        schema.TypeInt,
+		Optional:    true,
+		Description: "Specifies the number of seconds of idle time before the notebook is shut down automatically.",
```
sfc-gh-jmichalak: We can add a validation layer here, and add acceptance tests. See compute pool's `min_nodes`.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	},
+	"warehouse": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		Description:      "Specifies the warehouse that runs the notebook kernel and python code.",
+		ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"secrets": {
```
sfc-gh-jmichalak: This should be a list of nested fields. I guess it will be similar to services' `using` block.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	)
+	return &schema.Resource{
+		CreateContext: PreviewFeatureCreateContextWrapper(string(previewfeatures.NotebookResource), TrackingCreateWrapper(resources.Notebook, CreateNotebook)),
+		ReadContext:   PreviewFeatureReadContextWrapper(string(previewfeatures.NotebookResource), TrackingReadWrapper(resources.Notebook, GetReadNotebookFunc(true))),
+		UpdateContext: PreviewFeatureUpdateContextWrapper(string(previewfeatures.NotebookResource), TrackingUpdateWrapper(resources.Notebook, UpdateNotebook)),
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.NotebookResource), TrackingDeleteWrapper(resources.Notebook, deleteFunc)),
+		Description:   "Resource used to manage notebooks. For more information, check [notebooks documentation](https://docs.snowflake.com/en/sql-reference/sql/create-notebook).",
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.Notebook, customdiff.All(
+			ComputedIfAnyAttributeChanged(notebookSchema, ShowOutputAttributeName, "comment", "query_warehouse", "code_warehouse"),
```
sfc-gh-jmichalak: The logic for this one is: for each *AttributeName (show/desc/fully_qualified_name) we list fields that should cause the attributename to be marked as computed. That is, if a user makes an external change on field X, the attributename is changed as well.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	if notebook.CodeWarehouse != nil {
+		if err := d.Set("warehouse", notebook.CodeWarehouse.FullyQualifiedName()); err != nil {
+			return nil, err
+		}
+	}
+
+	errs := errors.Join(
+		d.Set("name", notebook.Name),
```
sfc-gh-jmichalak: Nit: this can be done with `ImportName`.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	name := d.Get("name").(string)
+	schemaName := d.Get("schema").(string)
+	databaseName := d.Get("database").(string)
+
+	notebookName := sdk.NewSchemaObjectIdentifier(databaseName, schemaName, name)
+
+	request := sdk.NewCreateNotebookRequest(notebookName)
+
+	errs := errors.Join(
+		stringAttributeCreateBuilder(d, "main_file", request.WithMainFile),
```
sfc-gh-jmichalak, line 182: Q: does SF allow setting `main_file` without `from`?

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+
+		stage, err := sdk.ParseSchemaObjectIdentifier(fromStageMap["stage"].(string))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		var location string
+		if l, ok := fromStageMap["location"]; ok {
+			location = l.(string)
+		}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: +1, this should also be covered by tests.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+
+		request.WithFrom(sdk.NewStageLocation(stage, location))
+	}
+
+	if err := client.Notebooks.Create(ctx, request); err != nil {
+		return diag.FromErr(err)
+	}
```
sfc-gh-jmichalak, line 210: We should also handle the `secrets` field here.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	errs := errors.Join(
+		stringAttributeUpdate(d, "comment", &set.Comment, &unset.Comment),
+		accountObjectIdentifierAttributeUpdate(d, "query_warehouse", &set.QueryWarehouse, &unset.QueryWarehouse),
+		accountObjectIdentifierAttributeUpdate(d, "warehouse", &set.Warehouse, &unset.Warehouse),
+	)
+	if errs != nil {
+		return diag.FromErr(errs)
+	}
+
+	if err := intAttributeUnsetFallbackUpdateWithZeroDefault(d, "idle_auto_shutdown_time_seconds", &set.IdleAutoShutdownTimeSeconds, 1800); err != nil {
```
sfc-gh-jmichalak: Nit: this could also belong to the errors.Join above.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	if !reflect.DeepEqual(set, &sdk.NotebookSetRequest{}) {
+		if err := client.Notebooks.Alter(ctx, sdk.NewAlterNotebookRequest(id).WithSet(*set)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	defaultUnset := &sdk.NotebookUnsetRequest{}
+	if !reflect.DeepEqual(unset, defaultUnset) {
+		// `warehouse` and `query_warehouse` cannot be unset in the same query
+		if unset.Warehouse != defaultUnset.Warehouse && unset.QueryWarehouse != defaultUnset.QueryWarehouse {
```
sfc-gh-jmichalak: Nit: I think we can just compare to an empty sdk.AccountObjectIdentifier. Alternatively, we can 
- extract code_warehouse above in errors.Join
- call unset on the unsetObject as usual (without handling code_warehouse)
- handle code_warehouse
We can discuss it with team.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+			return diag.FromErr(err)
+		}
+
+		errs := errors.Join(
+			d.Set(ShowOutputAttributeName, []map[string]any{schemas.NotebookToSchema(notebook)}),
+			d.Set(DescribeOutputAttributeName, []map[string]any{schemas.NotebookDetailsToSchema(notebookDetails)}),
+			d.Set(FullyQualifiedNameAttributeName, id.FullyQualifiedName()),
+			d.Set("comment", notebook.Comment),
+		)
+		if notebook.QueryWarehouse != nil {
```
sfc-gh-jmichalak, line 265: This won't work. What if it's unset externally? It won't be set to "" in this case.

docs/resources/notebook.md
```diff
@@ -0,0 +1,151 @@
...
+  schema   = "SCHEMA"
+  name     = "NOTEBOOK"
+}
+
+# complete resource
+resource "snowflake_notebook" "complete" {
+  name                            = "NOTEBOOK"
+  database                        = "DATABASE"
+  schema                          = "SCHEMA"
+  from                            = "\"<db_name>\".\"<schema_name>\".\"<stage_name>\""
```
sfc-gh-asawicki: let's have examples using references to other objects, here it could be: `from = snowflake_stage.test.fully_qualified_name`. Check a few other, newer resources and adjust also other attributes here.

pkg/acceptance/helpers/notebook_client.go
```diff
@@ -27,6 +27,7 @@ func (c *NotebookClient) client() sdk.Notebooks {
 func (c *NotebookClient) Create(t *testing.T) (*sdk.Notebook, func()) {
 	t.Helper()
 	id := c.ids.RandomSchemaObjectIdentifier()
+	t.Helper()
```
sfc-gh-asawicki: ?

pkg/acceptance/helpers/stage_client.go
```diff
@@ -100,6 +100,26 @@ func (c *StageClient) PutOnStage(t *testing.T, id sdk.SchemaObjectIdentifier, fi
 	require.NoError(t, err)
 }
 
+func (c *StageClient) PutOnStageWithPath(t *testing.T, id sdk.SchemaObjectIdentifier, filename string) string {
```
sfc-gh-asawicki: please extract common logic from this and PutInLocationWithContent instead copy-pasting

examples/resources/snowflake_notebook/resource.tf
```diff
@@ -0,0 +1,19 @@
...
+  schema   = "SCHEMA"
+  name     = "NOTEBOOK"
+}
+
+# complete resource
+resource "snowflake_notebook" "complete" {
+  name                            = "NOTEBOOK"
+  database                        = "DATABASE"
+  schema                          = "SCHEMA"
+  from                            = "\"<db_name>\".\"<schema_name>\".\"<stage_name>\""
```
sfc-gh-jcieslak: examples should be using existing resources and correlated fields: e.g. `from` field should be equal to `snowflake_stage.test.fully_qualifield_name`

examples/resources/snowflake_notebook/resource.tf
```diff
@@ -0,0 +1,19 @@
...
+  main_file                       = "MAIN_FILE.ipynb"
+  query_warehouse                 = "\"QUERY_WAREHOUSE\""
+  idle_auto_shutdown_time_seconds = number_of_seconds
```
sfc-gh-jcieslak: in this case use plain number value

examples/resources/snowflake_notebook/resource.tf
sfc-gh-jcieslak: Additionally, the example is wrong, because I can see the `from` field has different structure. Please, adjust to the current schema.

pkg/acceptance/bettertestspoc/assert/resourceassert/notebook_resource_ext.go
```diff
@@ -0,0 +1,14 @@
...
+import "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+
+func (n *NotebookResourceAssert) HasFromString(expectedPath string, expectedStageId string) *NotebookResourceAssert {
+	n.AddAssertion(assert.ValueSet("from.0.path", expectedPath))
+	n.AddAssertion(assert.ValueSet("from.0.stage", expectedStageId))
+	return n
+}
+
+func (n *NotebookResourceAssert) HasNoFromString() *NotebookResourceAssert {
+	n.AddAssertion(assert.ValueNotSet("from.0"))
```
sfc-gh-jcieslak: I think in this case preferred way of checking this would be to write `ValueSet("from.#", "0")` as presence of lists / objects / sets are checked in kind of a different way. The way I wrote would translate to "the length of the from list is 0".

pkg/resources/notebook.go
sfc-gh-asawicki: question: I can't see secrets in https://docs.snowflake.com/en/sql-reference/sql/create-notebook, can we create with them or only alter?

pkg/acceptance/helpers/notebook_client.go
sfc-gh-jcieslak: t.Helper called twice, please remove it

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"schema": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		Description:      blocklistedCharactersFieldDescription("The schema in which to create the notebook."),
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("code_warehouse"),
+	},
+	"comment": {
```
sfc-gh-asawicki: nit: the ordering of the attributes seems a bit random here, we usually follow the ordering from SQL (in the docs they will be automatically listed alphabetically)

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		if withExternalChangesMarking {
+			var warehouse string
+			if notebook.CodeWarehouse != nil {
+				warehouse = notebook.CodeWarehouse.Name()
+			}
+			if err = handleExternalChangesToObjectInShow(d,
```
sfc-gh-asawicki, line 246: what about other attributes? e.g. ExternalAccessSecrets are secrets, right?

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	"comment": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies a comment for the notebook.",
+	},
+	"from": {
+		Type:             schema.TypeList,
+		Optional:         true,
+		ForceNew:         true,
+		Description:      "Specifies that the notebook should be created from an .ipynb file in the specified stage location.",
```
sfc-gh-jcieslak: nit: I know it's written in docs this way, but I would change it to: `Specifies the location in a stage of an .ipynb file from which the notebook should be created` or something similar

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	if !reflect.DeepEqual(set, &sdk.NotebookSetRequest{}) {
+		if err := client.Notebooks.Alter(ctx, sdk.NewAlterNotebookRequest(id).WithSet(*set)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	defaultUnset := &sdk.NotebookUnsetRequest{}
+	if !reflect.DeepEqual(unset, defaultUnset) {
```
sfc-gh-asawicki: we have multiple updates, we should use `d.Partial(true)` (check other places)

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+					Type:             schema.TypeString,
+					Required:         true,
+					ForceNew:         true,
+					Description:      "Identifier of the stage where the .ipynb file is located.",
+					ValidateDiagFunc: IsValidIdentifier[sdk.SchemaObjectIdentifier](),
+					DiffSuppressFunc: suppressIdentifierQuoting,
+				},
+				"path": {
+					Type:        schema.TypeString,
+					Optional:    true,
```
sfc-gh-jcieslak, line 60: Question: what would happen if path to the file is not specified? Does it make sense to make it optional?

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -0,0 +1,434 @@
...
+	t.Cleanup(stageCleanup)
+	changedStage, changedStageCleanup := testClient().Stage.CreateStage(t)
+	t.Cleanup(changedStageCleanup)
+
+	path := testClient().Stage.PutOnStageWithPath(t, stage.ID(), "example.ipynb")
+	changedPath := testClient().Stage.PutOnStageWithPath(t, changedStage.ID(), "example.ipynb")
+
+	idleAutoShutdownTimeSeconds, changedIdleAutoShutdownTimeSeconds := 3600, 2400
+
+	modelBasic := model.Notebook(
```
sfc-gh-asawicki: nit: add a constructor taking id as parameter

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+					ForceNew:    true,
+					Description: "Location of the .ipynb file in the stage.",
+				},
+			},
+		},
+	},
+	"main_file": {
```
sfc-gh-jcieslak, line 67: Question: What will happen if this is omitted? And how does it correlate to `from` parameter, because Snowflake documentation states they're connected but I still don't get in what way.

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -0,0 +1,434 @@
...
+
+					newStage, newStageCleanup := testClient().Stage.CreateStage(t)
+					t.Cleanup(newStageCleanup)
+
+					testClient().Stage.PutOnStage(t, newStage.ID(), "example.ipynb")
+					newLocation := sdk.NewStageLocation(newStage.ID(), "")
+
+					newIdleAutoShutdownTimeSeconds := 4800
+
+					testClient().Notebook.CreateWithRequest(t, sdk.NewCreateNotebookRequest(id).
```
sfc-gh-asawicki: I don't get this step. Using `newLocation` means that forcenew on this attribute takes place, so the whole notebook will be dropped and recreated, what are we aiming to validate setting all these other fields?

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -0,0 +1,434 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeroles"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_Notebook_basic(t *testing.T) {
```
sfc-gh-asawicki: let's keep the name but let's adjust to the new format of preprod tests already (will be easier later). cc: @sfc-gh-jcieslak

pkg/resources/notebook.go
sfc-gh-jcieslak: +1 especially becasue the above Set warehouse is repetitive with Read operation

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+		stringAttributeCreateBuilder(d, "comment", request.WithComment),
+		accountObjectIdentifierAttributeCreate(d, "query_warehouse", &request.QueryWarehouse),
+		intAttributeCreateBuilder(d, "idle_auto_shutdown_time_seconds", request.WithIdleAutoShutdownTimeSeconds),
+		accountObjectIdentifierAttributeCreate(d, "warehouse", &request.Warehouse),
+	)
+	if errs != nil {
+		return diag.FromErr(errs)
+	}
+
+	if fromStage, ok := d.GetOk("from"); ok {
```
sfc-gh-jcieslak: nit: not sure if this is safe and `fromStage := d.Get("from"); len(fromStage) > 0` would be better, but probably safe

pkg/resources/notebook.go
```diff
@@ -0,0 +1,338 @@
...
+		return diag.FromErr(err)
+	}
+
+	if !reflect.DeepEqual(set, &sdk.NotebookSetRequest{}) {
+		if err := client.Notebooks.Alter(ctx, sdk.NewAlterNotebookRequest(id).WithSet(*set)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	defaultUnset := &sdk.NotebookUnsetRequest{}
```
sfc-gh-jcieslak: nit: inline

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -0,0 +1,434 @@
...
+	comment, changedComment := random.Comment(), random.Comment()
+
+	warehouse, warehouseCleanup := testClient().Warehouse.CreateWarehouse(t)
+	t.Cleanup(warehouseCleanup)
+	changedWarehouse, changedWarehouseCleanup := testClient().Warehouse.CreateWarehouse(t)
+	t.Cleanup(changedWarehouseCleanup)
+
+	queryWarehouse, queryWarehouseCleanup := testClient().Warehouse.CreateWarehouse(t)
+	t.Cleanup(queryWarehouseCleanup)
+	changedQueryWarehouse, changedQueryWarehouseCleanup := testClient().Warehouse.CreateWarehouse(t)
```
sfc-gh-jcieslak, line 228: nit: I think we could get away with creating just one warehouse ("changed" one) and as default just use already existing one accessible through testClient or I think just WarehouseName

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -0,0 +1,434 @@
...
+						HasQueryWarehouse(sdk.NewAccountObjectIdentifier("")).
+						HasCodeWarehouse(sdk.NewAccountObjectIdentifier("SYSTEM$STREAMLIT_NOTEBOOK_WH")).
+						HasOwner(snowflakeroles.Accountadmin.Name()).
+						HasOwnerRoleType("ROLE").
+						HasComment(""),
+				),
+			},
+		},
+	})
+}
```
sfc-gh-jcieslak, line 215: Hmm I told you to keep with the format of BasicUseCase / CompleteUseCase tests I just added to keep the consistency with newly introduced test format :( also adjust the test names as it's important

MIGRATION_GUIDE.md
sfc-gh-jcieslak: \+ I think merging conflicts (`## v2.9.x ➞ v2.10.0` below`)

sfc-gh-jcieslak (changes requested):

pkg/resources/notebook.go
sfc-gh-mkowalski: It allows, however it's not specified in the docs.

pkg/resources/notebook.go
sfc-gh-mkowalski: Technically docs don't even specify it's possible to modify the `code_warehouse`, so perhaps handling double unset case is unnecessary.

pkg/resources/notebook.go
sfc-gh-mkowalski: It seems it is also possible to create with them but it is not mentioned in the docs.

pkg/testacc/resource_notebook_acceptance_test.go
sfc-gh-mkowalski: I did similarly to the `git_repository` and `semantic_views` objects where the ForceNew parameters are being modified in this type of test.

pkg/resources/notebook.go
sfc-gh-mkowalski: When omitted it has a default value of `"notebook_app.ipynb"`.  From my understanding `from` holds the stage where the `.ipynb` file is located whereas `main_file` holds the location of the `.ipynb` file on the stage.

pkg/resources/notebook.go
sfc-gh-mkowalski: I think it does not make sense to make it optional; changed to required.

pkg/resources/notebook.go
sfc-gh-mkowalski: Docs suggest that they should; however I was unable to find any fitting value for them when trying to set them on snowflake so I can't be sure about it.
Adding this line `outputMapping{"external_access_secrets", "secrets", secrets, secrets, nil}` causes the refresh plan to be non-empty `secrets = jsonencode({})`.

pkg/resources/notebook.go
sfc-gh-mkowalski: I don't see any such block in the `service` object.

templates/resources/notebook.md.tmpl
sfc-gh-jmichalak: I saw the note, but I don't think we can say we're planning it because it's not available in Snowflake. We should say something like `Due to Snowflake limitations, external changes to `from` are not currently detected.`

pkg/resources/notebook.go
sfc-gh-jmichalak: Hmm, I think I wasn't clear, but ImportName with SchemaObjectIdentifier also sets the schema and the database, so setting them can be removed below.

pkg/resources/notebook.go
sfc-gh-jmichalak: Almost there - the `name` is missing in both cases. Also, handling fully_qualified_name is missing.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,341 @@
...
+			d.Set(ShowOutputAttributeName, []map[string]any{schemas.NotebookToSchema(notebook)}),
+			d.Set(DescribeOutputAttributeName, []map[string]any{schemas.NotebookDetailsToSchema(notebookDetails)}),
+			d.Set(FullyQualifiedNameAttributeName, id.FullyQualifiedName()),
+			d.Set("comment", notebook.Comment),
+		)
+		if notebook.QueryWarehouse != nil {
+			errs = errors.Join(errs, d.Set("query_warehouse", notebook.QueryWarehouse.FullyQualifiedName()))
+		} else {
+			if _, ok := d.GetOk("query_warehouse"); ok {
+				errs = errors.Join(errs, d.Set("query_warehouse", ""))
```
sfc-gh-jmichalak, line 269: Nit: let's not check the GetOk here and just set the empty string in the warehouse is nil.

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -0,0 +1,417 @@
...
+			{
+				Config:                  accconfig.FromModels(t, modelComplete),
+				ResourceName:            modelComplete.ResourceReference(),
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"from", "main_file", "idle_auto_shutdown_time_seconds"},
+			},
+		},
+	})
+}
```
sfc-gh-jmichalak, line 347: The validation tests and renaming tests are missing.

pkg/resources/notebook.go
sfc-gh-jmichalak: @sfc-gh-mkowalski, please take a look.

pkg/resources/notebook.go
```diff
@@ -0,0 +1,341 @@
...
+		stage, err := sdk.ParseSchemaObjectIdentifier(fromStageMap["stage"].(string))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		// var path string
+		// if l, ok := fromStageMap["path"]; ok {
+		// 	path = l.(string)
+		// }
+		// TODO: For some reason it works only with an empty path; I will investigate further it in the next commit.
```
sfc-gh-jmichalak: Please fix this before merging. Especially that this field is required, so I'm confused why this code is commented out.

examples/resources/snowflake_notebook/resource.tf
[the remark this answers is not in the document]
sfc-gh-jmichalak: @sfc-gh-mkowalski, please take a look.

pkg/resources/notebook.go
sfc-gh-jmichalak: https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/resources/service_commons.go#L66. The point is, we need to set both secret name and its value, like in SDK. Now we set only the name.

pkg/resources/notebook.go
sfc-gh-jmichalak: But we allow to unset both these properties - they are handled in the Update function. So, we need to
- separate unsetting one of the fields from the other one
- have a test step proving this does not fail.

pkg/resources/notebook.go
sfc-gh-jmichalak: external_access_secrets is available in DESC, so use handleExternalChangesToObjectInFlatDescribe instead.

pkg/resources/notebook.go
sfc-gh-mkowalski: I added it to the Update since creating with secrets is not supported.

examples/resources/snowflake_notebook/resource.tf
sfc-gh-jcieslak: The example is still wrong, `from` is not a single string value, but a block.
Instead of having `from = ""`it should be `from {
  stage = stage.fully_qualified_name
  path = "some/path"
}`

pkg/resources/notebook.go
sfc-gh-jcieslak: Okay, so in this case, I think that the `path` field in `from` should be optional because you could store the file in the stage's root which doesn't require any path to be specified. Or we could discuss if in such case e.g. `/` should be specified (and how it is done in other resources of similar type; working with stages)

pkg/resources/notebook.go
```diff
@@ -37,19 +37,13 @@ var notebookSchema = map[string]*schema.Schema{
...
 	"from": {
-		Type:             schema.TypeList,
-		Optional:         true,
-		ForceNew:         true,
-		Description:      "Specifies that the notebook should be created from an .ipynb file in the specified stage location.",
-		DiffSuppressFunc: suppressIdentifierQuoting,
+		Type:        schema.TypeList,
+		Optional:    true,
+		ForceNew:    true,
+		Description: "Specifies the location in a stage of an .ipynb file from which the notebook should be created.",
```
sfc-gh-jcieslak: Q: Should we add `RequiredWith: []string{"main_file"}` validation and explain more in the description that this is specifically made for creating notebook from another one and MAIN_FILE should be set? I'm asking because in the documentation it says MAIN_FILE should be set, but you mentioned there is a default main_file name (the default should be also documented)

pkg/resources/notebook.go
```diff
@@ -96,12 +95,23 @@ var notebookSchema = map[string]*schema.Schema{
...
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"secret_variable_name": {
+					Type:        schema.TypeString,
+					Required:    true,
+					Description: "The name of the secret variable.",
+				},
+				"secret_id": {
+					Type:             schema.TypeString,
+					Required:         true,
```
sfc-gh-jcieslak: let's add identifier validation `IsValidIdentifier`

pkg/resources/notebook.go
```diff
@@ -162,16 +173,12 @@ func ImportNotebook(ctx context.Context, d *schema.ResourceData, meta any) ([]*s
...
-		}
+	if _, err := ImportName[sdk.SchemaObjectIdentifier](context.Background(), d, nil); err != nil {
+		return nil, err
 	}
 
 	errs := errors.Join(
-		d.Set("name", notebook.Name),
-		d.Set("database", notebook.DatabaseName),
-		d.Set("schema", notebook.SchemaName),
+		d.Set("warehouse", notebook.CodeWarehouse.FullyQualifiedName()),
```
sfc-gh-jcieslak, line 163: this is redundant with Read, you could inline the ImportName like it's done in e.g. tag resource

pkg/resources/notebook.go
sfc-gh-jcieslak: Yes, but we don't want the situation where customer is setting secrets during create and has to run the `terraform apply` twice, because something is implemented only in Update method. We have a convention that when something can be only set through update and can be set in the config it should be handled **after** the ID for the resource is set in Create.

pkg/resources/notebook.go
```diff
@@ -253,15 +260,19 @@ func GetReadNotebookFunc(withExternalChangesMarking bool) schema.ReadContextFunc
 		}
 
 		if withExternalChangesMarking {
-			var warehouse string
-			if notebook.CodeWarehouse != nil {
-				warehouse = notebook.CodeWarehouse.Name()
-			}
+			warehouse := notebook.CodeWarehouse.Name()
 			if err = handleExternalChangesToObjectInShow(d,
 				outputMapping{"code_warehouse", "warehouse", warehouse, warehouse, nil},
```
sfc-gh-jcieslak, line 247: Looking at the documentation there's more logic to this parameter: `If no warehouse is specified when you create a notebook, Snowflake uses the default warehouse defined by the schema lineage parameter DEFAULT_STREAMLIT_NOTEBOOK_WAREHOUSE. You can set this parameter at the schema, database, or account lineage level to define a preferred warehouse.` which I don't think we are handling in this resource, right?

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -432,3 +415,87 @@ func TestAcc_Notebook_complete(t *testing.T) {
...
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: CheckDestroy(t, resources.Notebook),
+		Steps: []resource.TestStep{
+			// create object
+			{
+				Config: accconfig.FromModels(t, modelComplete),
+				Check:  assertThat(t),
```
sfc-gh-jcieslak: let's at least assert the name in the resource to confirm it's in the state

pkg/resources/notebook.go
sfc-gh-mkowalski: I think I can't do it this way because it would leave the `warehouse` unset.

pkg/resources/notebook.go
sfc-gh-mkowalski: I think I don't need to do anything since the snowflake provides the default itself and it seems that the existing tests prove that it works correctly: `HasCodeWarehouse(sdk.NewAccountObjectIdentifier("\"SYSTEM$STREAMLIT_NOTEBOOK_WH\""))`.

sfc-gh-asawicki: But we don't have test checking the following scenarios (or maybe I didn't see it):
First:
1. Create notebook without warehouse (snowflake default is set).
2. Change the warehouse externally (directly on WH).
3. What do we expect here? UNSET? Will it happen?

Second:
1. Create notebook without warehouse (snowflake default is set).
2. Change parameter on the schema level.
3. What do we expect here? It's not set, so I guess we shouldnt do anything (no-op). Would it behave like this now?

The other important question here is: is it treated as a property of notebook or a parameter? If it's treated as parameter, it should have the parameters logic like in the other resources. I think it's not because the parameter name is `DEFAULT_STREAMLIT_NOTEBOOK_WAREHOUSE` but it should be verified.

The warehouse itself is in preview, so we can discuss our options:
1. Remove it for now and investigate later. Document this limitation.
2. Check at least the behavior I shared and decide then. Document the limitation.
3. Other?

@sfc-gh-jcieslak @sfc-gh-jmichalak @sfc-gh-mkowalski wdyt?

sfc-gh-jcieslak (approved):

docs/resources/notebook.md
```diff
@@ -7,6 +7,8 @@ description: |-
 
 -> **Note** Due to Snowflake limitations, external changes to `from` are not currently detected.
 
+-> **Note** External changes to `warehouse` are not currently supported. They will be handled in the following versions of the provider which may still affect this resource.
```
sfc-gh-asawicki: So are they handled or not?

examples/resources/snowflake_notebook/resource.tf
```diff
@@ -7,10 +7,13 @@ resource "snowflake_notebook" "basic" {
...
 resource "snowflake_notebook" "complete" {
-  name                            = "NOTEBOOK"
-  database                        = "DATABASE"
-  schema                          = "SCHEMA"
-  from                            = snowflake_stage.test.fully_qualifed_name
+  name     = "NOTEBOOK"
+  database = "DATABASE"
+  schema   = "SCHEMA"
+  from {
+    stage = stage.fully_qualified_name
```
sfc-gh-asawicki: it's incorrect. It should be: snowflake_stage.test.fully_qualified_name

pkg/resources/notebook.go
sfc-gh-asawicki: I still see Optional?

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -499,3 +456,187 @@ func TestAcc_Notebook_double_warehouse_change(t *testing.T) {
...
+					assert.Check(resource.TestCheckResourceAttr(modelWithChangedName.ResourceReference(), "describe_output.0.owner", snowflakeroles.Accountadmin.Name())),
+					assert.Check(resource.TestCheckResourceAttr(modelWithChangedName.ResourceReference(), "describe_output.0.comment", "")),
+				),
+			},
+		},
+	})
+}
+
+// TODO [SNOW-2858907]: Detect external changes on warehouse.
+func TestAcc_Notebook_ExternalWarehouseChange(t *testing.T) {
```
sfc-gh-asawicki, line 550: Please add `ConfigPlanChecks` verifying: the drift, planned update, and resource action (you can check other tests how to do it).

sfc-gh-asawicki (changes requested):

docs/resources/notebook.md
sfc-gh-mkowalski: they are; it's changed in the new commit

pkg/resources/notebook.go
sfc-gh-mkowalski: It's meant to be that way (from another discussion)
`Okay, so in this case, I think that the path field in from should be optional because you could store the file in the stage's root which doesn't require any path to be specified`

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -545,14 +547,14 @@ func TestAcc_Notebook_Rename(t *testing.T) {
...
 
-// TODO [SNOW-2858907]: Detect external changes on warehouse.
 func TestAcc_Notebook_ExternalWarehouseChange(t *testing.T) {
 	id := testClient().Ids.RandomSchemaObjectIdentifier()
 
 	warehouse, warehouseCleanup := testClient().Warehouse.CreateWarehouse(t)
 	t.Cleanup(warehouseCleanup)
 
 	modelBasic := model.NotebookFromId("test", id)
+	warehouseNoQuotes := warehouse.ID().FullyQualifiedName()[1 : len(warehouse.ID().FullyQualifiedName())-1]
```
sfc-gh-jcieslak, line 557: In this case use ID().Name() (can be adjusted in the follow-up)

sfc-gh-jcieslak (approved):

pkg/resources/notebook.go
sfc-gh-jmichalak: fully_qualified_name is still missing

pkg/resources/notebook.go
sfc-gh-jmichalak: What about this?

pkg/testacc/resource_notebook_acceptance_test.go
sfc-gh-jmichalak: The renames is tested, but validations are not.

pkg/testacc/resource_notebook_acceptance_test.go
```diff
@@ -0,0 +1,646 @@
...
+	t.Cleanup(stageCleanup)
+	changedStage, changedStageCleanup := testClient().Stage.CreateStage(t)
+	t.Cleanup(changedStageCleanup)
+
+	path := testClient().Stage.PutOnStageWithPath(t, stage.ID(), "testdata", "example.ipynb")
+	changedPath := testClient().Stage.PutOnStageWithPath(t, changedStage.ID(), "testdata", "example.ipynb")
+
+	idleAutoShutdownTimeSeconds, changedIdleAutoShutdownTimeSeconds := 3600, 2400
+
+	modelComplete := model.NotebookFromId("test", id).
```
sfc-gh-jmichalak, line 241: These tests should also assert secrets.

pkg/resources/notebook.go
sfc-gh-jmichalak: Ultimately, we decided not to include this field because we should clarify it first. Resolving.

PR #4202, opened 2025-11-16
chore: Add User WIF auth to SDK

Relativity74205:
- Add User WIF auth to SDK.
- Add a new `SHOW USER WORKLOAD IDENTITY AUTHENTICATION METHODS` operation for getting WIF information.
- Add a new SHOW USERS field to SDK and to resources/data source.

### DIFF
--- a/pkg/sdk/application_roles_impl_gen.go
+++ b/pkg/sdk/application_roles_impl_gen.go
@@ -54,7 +54,6 @@ func (r *GrantApplicationRoleRequest) toOpts() *GrantApplicationRoleOptions {
 		name: r.name,
 	}
 	opts.To = KindOfRole{
-		// Edited manually (From -> To)
 		RoleName:            r.To.RoleName,
 		ApplicationRoleName: r.To.ApplicationRoleName,
 		ApplicationName:     r.To.ApplicationName,
--- a/pkg/sdk/applications_impl_gen.go
+++ b/pkg/sdk/applications_impl_gen.go
@@ -79,11 +79,9 @@ func (r *CreateApplicationRequest) toOpts() *CreateApplicationOptions {
 		Tag:         r.Tag,
 	}
 	if r.Version != nil {
-		// Manually changed UpgradeVersion -> Version
 		opts.Version = &ApplicationVersion{
 			VersionDirectory: r.Version.VersionDirectory,
 		}
-		// Manually changed UpgradeVersion -> Version
 		if r.Version.VersionAndPatch != nil {
 			opts.Version.VersionAndPatch = &VersionAndPatch{
 				Version: r.Version.VersionAndPatch.Version,
--- a/pkg/sdk/authentication_policies_impl_gen.go
+++ b/pkg/sdk/authentication_policies_impl_gen.go
@@ -83,21 +83,18 @@ func (r *CreateAuthenticationPolicyRequest) toOpts() *CreateAuthenticationPolicy
 		Comment:                  r.Comment,
 	}
 	if r.MfaPolicy != nil {
-		// adjusted manually
 		opts.MfaPolicy = &AuthenticationPolicyMfaPolicy{
 			EnforceMfaOnExternalAuthentication: r.MfaPolicy.EnforceMfaOnExternalAuthentication,
 			AllowedMethods:                     r.MfaPolicy.AllowedMethods,
 		}
 	}
 	if r.SecurityIntegrations != nil {
-		// adjusted manually
 		opts.SecurityIntegrations = &SecurityIntegrationsOption{
 			All:                  r.SecurityIntegrations.All,
 			SecurityIntegrations: r.SecurityIntegrations.SecurityIntegrations,
 		}
 	}
 	if r.PatPolicy != nil {
-		// adjusted manually
 		opts.PatPolicy = &AuthenticationPolicyPatPolicy{
 			DefaultExpiryInDays:     r.PatPolicy.DefaultExpiryInDays,
 			MaxExpiryInDays:         r.PatPolicy.MaxExpiryInDays,
@@ -105,7 +102,6 @@ func (r *CreateAuthenticationPolicyRequest) toOpts() *CreateAuthenticationPolicy
 		}
 	}
 	if r.WorkloadIdentityPolicy != nil {
-		// adjusted manually
 		opts.WorkloadIdentityPolicy = &AuthenticationPolicyWorkloadIdentityPolicy{
 			AllowedProviders:    r.WorkloadIdentityPolicy.AllowedProviders,
 			AllowedAwsAccounts:  r.WorkloadIdentityPolicy.AllowedAwsAccounts,
--- a/pkg/sdk/event_tables_impl_gen.go
+++ b/pkg/sdk/event_tables_impl_gen.go
@@ -172,14 +172,12 @@ func (r *AlterEventTableRequest) toOpts() *AlterEventTableOptions {
 	}
 	if r.AddRowAccessPolicy != nil {
 		opts.AddRowAccessPolicy = &EventTableAddRowAccessPolicy{
-			// manually changed DropAndAddRowAccessPolicy.Add to AddRowAccessPolicy
 			RowAccessPolicy: r.AddRowAccessPolicy.RowAccessPolicy,
 			On:              r.AddRowAccessPolicy.On,
 		}
 	}
 	if r.DropRowAccessPolicy != nil {
 		opts.DropRowAccessPolicy = &EventTableDropRowAccessPolicy{
-			// manually changed DropAndAddRowAccessPolicy.Drop to DropRowAccessPolicy
 			RowAccessPolicy: r.DropRowAccessPolicy.RowAccessPolicy,
 		}
 	}
@@ -205,7 +203,6 @@ func (r *AlterEventTableRequest) toOpts() *AlterEventTableOptions {
 		opts.SearchOptimizationAction = &EventTableSearchOptimizationAction{}
 		if r.SearchOptimizationAction.Add != nil {
 			opts.SearchOptimizationAction.Add = &SearchOptimization{
-				// manually changed Drop to Add
 				On: r.SearchOptimizationAction.Add.On,
 			}
 		}
--- a/pkg/sdk/generator/README.md
+++ b/pkg/sdk/generator/README.md
@@ -91,6 +91,9 @@ make generate-sdk-examples SF_TF_GENERATOR_ARGS='--filter-generation-part-names=
 make generate-sdk-examples SF_TF_GENERATOR_ARGS='--help'
 ```
 
+##### Known issues
+- The implementation of nested fields causes problems when reusing nested definitions (the same `[]Fields` slice is reused causing parent redefinition and incorrect mapping; the root cause being the lack of separation between the definition and model structs). It's currently validated programmatically and the panic is raised (`Field <field> already has a parent`). When it happens, create a function wrapper instead of directly creating a `var` with a definition.
+
 [//]: # (TODO [next PRs]: update this section)
 ### Next steps
 
@@ -199,5 +202,3 @@ find a better solution to solve the issue (add more logic to the templates ?)
 
 ##### Known issues
 - generating two converts when Show and Desc use the same data structure
-- cannot re-generate when client.go is using generated interface
-- spaces in templates (especially nested validations)
--- a/pkg/sdk/generator/defs/applications_def.go
+++ b/pkg/sdk/generator/defs/applications_def.go
@@ -14,10 +14,12 @@ var versionAndPatch = g.NewQueryStruct("VersionAndPatch").
 	TextAssignment("VERSION", g.ParameterOptions().NoEquals().NoQuotes().Required()).
 	OptionalNumberAssignment("PATCH", g.ParameterOptions().NoEquals().Required())
 
-var applicationVersion = g.NewQueryStruct("ApplicationVersion").
-	OptionalText("VersionDirectory", g.KeywordOptions().SingleQuotes()).
-	OptionalQueryStructField("VersionAndPatch", versionAndPatch, g.KeywordOptions().NoQuotes()).
-	WithValidation(g.ExactlyOneValueSet, "VersionDirectory", "VersionAndPatch")
+var applicationVersion = func() *g.QueryStruct {
+	return g.NewQueryStruct("ApplicationVersion").
+		OptionalText("VersionDirectory", g.KeywordOptions().SingleQuotes()).
+		OptionalQueryStructField("VersionAndPatch", versionAndPatch, g.KeywordOptions().NoQuotes()).
+		WithValidation(g.ExactlyOneValueSet, "VersionDirectory", "VersionAndPatch")
+}
 
 var applicationSet = g.NewQueryStruct("ApplicationSet").
 	OptionalTextAssignment("COMMENT", g.ParameterOptions().SingleQuotes()).
@@ -49,7 +51,7 @@ var applicationsDef = g.NewInterface(
 		Identifier("PackageName", g.KindOfT[sdkcommons.AccountObjectIdentifier](), g.IdentifierOptions().Required()).
 		OptionalQueryStructField(
 			"Version",
-			applicationVersion,
+			applicationVersion(),
 			g.KeywordOptions().SQL("USING"),
 		).
 		OptionalBooleanAssignment("DEBUG_MODE", g.ParameterOptions()).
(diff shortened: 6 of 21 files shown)


pkg/sdk/users.go
```diff
@@ -77,6 +77,8 @@ type User struct {
 	HasRsaPublicKey       bool
 	Type                  string
 	HasMfa                bool
+	HasPAT                bool
```
Relativity74205: Was missing in the SDK so far

Relativity74205:
A few questions about testing:
- Does it make sense to add some unit tests? I would think that the integration tests are enough
- Shall I add workload identity to some other existing unit tests somewhere? I don't fully understand the meaning behind all the tests in `users_integration_test.go` to be honest

pkg/sdk/users.go
sfc-gh-jmichalak: Nit: this change adds HasPAT field only here, and not in other places (resource's show_ouput and SDK's DESC output). Let's either revert adding it for now, or add in all other places, just like with the WIF.

pkg/sdk/users.go
```diff
@@ -269,28 +279,37 @@ func (v *users) Create(ctx context.Context, id AccountObjectIdentifier, opts *Cr
...
+	RSAPublicKeyFp        *string                                 `ddl:"parameter,single_quotes" sql:"RSA_PUBLIC_KEY_FP"`
+	RSAPublicKey2         *string                                 `ddl:"parameter,single_quotes" sql:"RSA_PUBLIC_KEY_2"`
+	RSAPublicKey2Fp       *string                                 `ddl:"parameter,single_quotes" sql:"RSA_PUBLIC_KEY_2_FP"`
+	Type                  *UserType                               `ddl:"parameter,no_quotes" sql:"TYPE"`
+	WorkloadIdentity      *[]UserObjectWorkloadIdentityProperties `ddl:"parameter,list,parentheses,no_comma" sql:"WORKLOAD_IDENTITY"`
+	Comment               *string                                 `ddl:"parameter,single_quotes" sql:"COMMENT"`
+}
+
+type UserObjectWorkloadIdentityProperties struct {
+	Type             *string                 `ddl:"parameter,no_quotes" sql:"TYPE"`
```
sfc-gh-jmichalak: Let's actually use enums - check `WarehouseType` in `warehouses.go` for reference. We should allow only listed values.

pkg/sdk/users.go
```diff
@@ -269,28 +279,37 @@ func (v *users) Create(ctx context.Context, id AccountObjectIdentifier, opts *Cr
...
+	Arn              *string                 `ddl:"parameter,single_quotes" sql:"ARN"`
+	Issuer           *string                 `ddl:"parameter,single_quotes" sql:"ISSUER"`
+	Subject          *string                 `ddl:"parameter,single_quotes" sql:"Subject"`
+	OidcAudienceList []StringListItemWrapper `ddl:"parameter,parentheses" sql:"OIDC_AUDIENCE_LIST"`
```
sfc-gh-jmichalak, line 362: Let's adjust the object validation:
- we already have a validating function. Inside `if valueSet(opts.ObjectProperties)...` let's move the second condition inside, and next to it, make proper validations
- in there, we can validate e.g. 
  - `exactlyOneValueSet` with these fields - meaning at least one of these fields should be set,
  - `arn` is set for type=aws and not set for other types, etc... (following the SF docs).
- add the unit tests in `users_test.go` showing the validation works.

docs/data-sources/users.md
```diff
@@ -150,6 +150,7 @@ Read-Only:
 - `ext_authn_uid` (String)
 - `first_name` (String)
 - `has_mfa` (Boolean)
+- `has_workload_identity` (Boolean)
```
sfc-gh-jmichalak, line 153: Nit: This change only adds new fields to SDK, but please add a migration guide entry (see the MIGRATION_GUIDE.md file), so that we won't forget about it, let's add something like 
```
## v2.10.1 ➞ v2.10.2

### *(improvement)* New fields in user resources and data sources output fields
We adjusted the `show_output` by adding the missing `has_workload_identity ` field. This concerns `user`, `service_user`, and `legacy_service_user` resources and `users` data source.
```

pkg/sdk/testint/users_integration_test.go
```diff
@@ -412,6 +414,78 @@ func TestInt_Users(t *testing.T) {
 			HasExpiresAtTimeNotEmpty().
 			HasLockedUntilTimeNotEmpty().
 			HasHasPassword(false).
+			HasHasWorkloadIdentity(false).
+			HasHasRsaPublicKey(true),
+		)
+	})
+
+	t.Run("create: all object properties - type service; WIF auth", func(t *testing.T) {
```
sfc-gh-jmichalak: I think we don't need to add this new test. Instead, we should reuse the `create: all object properties - type service` and for legacy service, if eligible. We can just add the `WorkloadIdentity` field there and do the proper assertions.

Additionally, this change also allows this option for ALTER, so we should modify the ALTER test as well (integration and unit tests).

pkg/sdk/users.go
```diff
@@ -124,6 +126,8 @@ type userDBRow struct {
 	HasRsaPublicKey       sql.NullBool   `db:"has_rsa_public_key"`
 	Type                  sql.NullString `db:"type"`
 	HasMfa                sql.NullBool   `db:"has_mfa"`
+	HasPAT                sql.NullBool   `db:"has_pat"`
+	HasWorkloadIdentity   sql.NullBool   `db:"has_workload_identity"`
```
sfc-gh-jmichalak, line 130: The unit tests are missing - please see other fields for this resource.

pkg/sdk/users.go
```diff
@@ -207,6 +211,12 @@ func (row userDBRow) convert() (*User, error) {
 	if row.HasMfa.Valid {
 		user.HasMfa = row.HasMfa.Bool
 	}
+	if row.HasPAT.Valid {
```
sfc-gh-jmichalak: This is missing from the PR description, but what about https://docs.snowflake.com/en/sql-reference/sql/show-user-workload-identity-authentication-methods? I think we'll need it (can be done in the next PR).

sfc-gh-jmichalak:
Thank you for the contribution :) About the questions:
- Yes, we would like to keep the feedback loop shorter if possible. Additionally, the SDK generator is very simple, and verifying all the quotes, and wrappers is very useful. In general, we have the following test types:
  - Unit tests for verifying the generated SQLs (just the generator is running). There, we cover all possible options, filters, etc.
  - Integration tests for verifying the SDK queries against a real Snowflake instance,
  - Acceptance tests against a real Snowflake instance that are Terraform-specific and test resource behavior.
- The integration tests should be added to the `users_integration_test.go` file (like they are now). The unit tests should be in `users_test.go`. Just follow the existing tests, e.g. `with complete options - no type` or `alter: set object properties`.

pkg/sdk/users.go
Relativity74205: Then you have changed the strategy at some point? I remember than in some PR in the past, your team said, that you are not validating such values in the provider since Snowflake development moves so fast.

pkg/sdk/users.go
Relativity74205: Hmm, also here my question about the strategy. In the past I was instructed to remove some validations since Snowflake developed so fast. When the WIF-implementation allows new options, then the provider has to change.

pkg/sdk/users.go
Relativity74205: I started to implement this manually, however, I noticed that this was implemented for PATs with a generator (`pkg/sdk/user_programmatic_access_tokens_gen.go`). How?

pkg/sdk/testint/users_integration_test.go
Relativity74205: done (integration test part), I will add it also to the unit tests as requested in other comments

pkg/sdk/users.go
Relativity74205: I will revert it for now, since this PR is already more complex than anticipated (at least from my side)

pkg/sdk/users.go
Relativity74205: 👍 this was also requested in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4202#issuecomment-3542173056.

Will do it, however, it irritated me, that a lot of user attributes were not tested in `pkg/sdk/users_test.go`. However, for workload_identity it totally makes sense to do this.

Relativity74205: Added a first unit test for WIF. Will do the rest after https://github.com/snowflakedb/terraform-provider-snowflake/pull/4202#discussion_r2534246309 and https://github.com/snowflakedb/terraform-provider-snowflake/pull/4202#discussion_r2534224670 have been clarified.

pkg/sdk/users.go
sfc-gh-jmichalak: Yeah, we used a new generator for PATs because they are a bit more complex (more queries with more parameters), and the pats kinda behave like a separate object. Here, we only need to add a new SHOW function, so we can do it manually (the user object does not use the SDK generator.

pkg/sdk/users.go
```diff
@@ -825,3 +843,75 @@ var AcceptableUserTypes = map[UserType][]string{
 	UserTypeService:       {string(UserTypeService)},
 	UserTypeLegacyService: {string(UserTypeLegacyService)},
 }
+
+// TODO: unclear how to proceed here with implementing ShowUserWorkloadIdentityAuthenticationMethods (manually or with a generator)
+//type userWorkloadIdentityAuthenticationMethodsDBRow struct {
```
sfc-gh-jmichalak: The commented code looks good in general, but  
\- we don't need Methods interface - instead, let's rename the Show method to ShowUserWorkloadIdentityAuthenticationMethods and extend the User interface  
\- let's have a unit and integration test.

pkg/sdk/users.go
sfc-gh-jmichalak: Please also add a unit test for ALTER USER ...

pkg/sdk/users.go
sfc-gh-jmichalak: We validate  
\- obviously incorrect values, like negative seconds, etc  
\- fields that are related, e.g. mutually exclusive, or required together, etc (AtLeastOneOf is an example)  
\- enums

We don't validate  
\- specific number ranges  
  
Regarding the enums, we have had quite a consistent approach for some time. We already have a WIP PR for adding the enum support for the generator.   
  
I can push the changes to this PR, if it's still unclear.

pkg/sdk/users.go
sfc-gh-jmichalak: As above (I can also adjust this one).

Relativity74205: Done, however, please check. I think that I have added the most important validations (and unit tests for them).

pkg/sdk/users.go
Relativity74205: Done, however, please check. I think that I have added the most important validations (and unit tests for them).

pkg/sdk/users.go
Relativity74205: done

pkg/sdk/users.go
Relativity74205: done, also using this show function in integration tests

pkg/sdk/users.go
Relativity74205: done, also using this show function in integration tests (and added a unit test for it)

pkg/sdk/users.go
```diff
@@ -269,28 +284,95 @@ func (v *users) Create(ctx context.Context, id AccountObjectIdentifier, opts *Cr
...
+	DefaultNamespace      *ObjectIdentifier                       `ddl:"identifier,equals" sql:"DEFAULT_NAMESPACE"`
+	DefaultRole           *AccountObjectIdentifier                `ddl:"identifier,equals" sql:"DEFAULT_ROLE"`
+	DefaultSecondaryRoles *SecondaryRoles                         `ddl:"parameter,equals" sql:"DEFAULT_SECONDARY_ROLES"`
+	MinsToBypassMFA       *int                                    `ddl:"parameter,no_quotes" sql:"MINS_TO_BYPASS_MFA"`
+	RSAPublicKey          *string                                 `ddl:"parameter,single_quotes" sql:"RSA_PUBLIC_KEY"`
+	RSAPublicKeyFp        *string                                 `ddl:"parameter,single_quotes" sql:"RSA_PUBLIC_KEY_FP"`
+	RSAPublicKey2         *string                                 `ddl:"parameter,single_quotes" sql:"RSA_PUBLIC_KEY_2"`
+	RSAPublicKey2Fp       *string                                 `ddl:"parameter,single_quotes" sql:"RSA_PUBLIC_KEY_2_FP"`
+	Type                  *UserType                               `ddl:"parameter,no_quotes" sql:"TYPE"`
+	WorkloadIdentity      *[]UserObjectWorkloadIdentityProperties `ddl:"parameter,list,parentheses,no_comma" sql:"WORKLOAD_IDENTITY"`
```
sfc-gh-jmichalak: I think we missed this initially, but why is it a slice? I think it should be a single `*[]UserObjectWorkloadIdentityProperties`​, no? Let's adjust after the merge.

pkg/sdk/users.go
```diff
@@ -269,28 +284,95 @@ func (v *users) Create(ctx context.Context, id AccountObjectIdentifier, opts *Cr
...
+	case string(WIFTypeGCP):
+		return WIFTypeGCP, nil
+	case string(WIFTypeOIDC):
+		return WIFTypeOIDC, nil
+	default:
+		return "", fmt.Errorf("invalid warehouse type: %s", s)
+	}
+}
+
+func (e WIFType) FromString(s string) (WIFType, error) {
```
sfc-gh-jmichalak: I think this is not needed (ToWIFTypeType should be preferred.

pkg/sdk/testint/users_integration_test.go
```diff
@@ -498,8 +517,15 @@ func TestInt_Users(t *testing.T) {
...
 			HasHasPassword(true).
+			HasHasWorkloadIdentity(true).
 			HasHasRsaPublicKey(true),
 		)
+
+		methods, err := client.Users.ShowUserWorkloadIdentityAuthenticationMethodOptions(ctx, id)
+		require.NoError(t, err)
+		assert.Equal(t, 1, len(methods))
+		assert.Equal(t, methods[0].Type, "OIDC")
+		assert.Equal(t, methods[0].AdditionalInfo, "{\"audienceList\":[],\"issuer\":\"https://accounts.google.com\",\"subject\":\"system:serviceaccount:service_account_namespace:service_account_name\"}")
```
sfc-gh-jmichalak: Hmm, since it looks like a proper JSON, let's break this up in SDK (can be done after merge).

pkg/sdk/users.go
```diff
@@ -269,28 +284,95 @@ func (v *users) Create(ctx context.Context, id AccountObjectIdentifier, opts *Cr
...
+	case string(WIFTypeAWS):
+		return WIFTypeAWS, nil
+	case string(WIFTypeAzure):
+		return WIFTypeAzure, nil
+	case string(WIFTypeGCP):
+		return WIFTypeGCP, nil
+	case string(WIFTypeOIDC):
+		return WIFTypeOIDC, nil
+	default:
+		return "", fmt.Errorf("invalid warehouse type: %s", s)
```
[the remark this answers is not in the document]
Relativity74205: fixed

pkg/sdk/users.go
Relativity74205: fixed

pkg/sdk/testint/users_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/users.go
```diff
@@ -269,28 +284,91 @@ func (v *users) Create(ctx context.Context, id AccountObjectIdentifier, opts *Cr
...
+			errs = append(errs, NewError("Subject must be set for GCP workload identity"))
+		}
+	case WIFTypeOIDC:
+		if !valueSet(workloadIdentity.Issuer) {
+			errs = append(errs, NewError("Issuer must be set for OIDC workload identity"))
+		}
+		if !valueSet(workloadIdentity.Subject) {
+			errs = append(errs, NewError("Subject must be set for OIDC workload identity"))
+		}
+	}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Reworked.

pkg/sdk/users.go
sfc-gh-jmichalak: Converted to a single object.

sfc-gh-jmichalak (approved):

PR #4203, opened 2025-11-17
chore: Fix data source tests

sfc-gh-jcieslak:
Changes
Adjust data source tests: 
- In most cases the `IN` filtering, if not set, does not lists all objects `IN ACCOUNT`, but at the `IN DATABASE` level where database is chosen from the context which depends on the tests execution. Because of this, some of the tests were adjusted and in the database role case, I removed one test case, as it was repetitive.
- The warehouse test was failing on one parameter which appears to be different in different test Snowflake accounts (in different environments). For now, created a ticket and commented out this part to find out the reason for this difference.

### DIFF
--- a/pkg/testacc/data_source_database_roles_acceptance_test.go
+++ b/pkg/testacc/data_source_database_roles_acceptance_test.go
@@ -29,12 +29,15 @@ func TestAcc_DatabaseRoles_BasicUseCase_DifferentFiltering(t *testing.T) {
 
 	datasourceModelLikeExact := datasourcemodel.DatabaseRoles("test", databaseRoleId1.DatabaseName()).
 		WithLike(databaseRoleId1.Name()).
-		WithInDatabase(databaseRoleId3.DatabaseName()).
 		WithDependsOn(databaseRoleModel1.ResourceReference(), databaseRoleModel2.ResourceReference(), databaseRoleModel3.ResourceReference())
 
 	datasourceModelLikePrefix := datasourcemodel.DatabaseRoles("test", databaseRoleId1.DatabaseName()).
 		WithLike(prefix+"%").
+		WithDependsOn(databaseRoleModel1.ResourceReference(), databaseRoleModel2.ResourceReference(), databaseRoleModel3.ResourceReference())
+
+	datasourceModelInDatabase := datasourcemodel.DatabaseRoles("test", databaseRoleId3.DatabaseName()).
 		WithInDatabase(databaseRoleId3.DatabaseName()).
+		WithLike(databaseRoleId3.Name()).
 		WithDependsOn(databaseRoleModel1.ResourceReference(), databaseRoleModel2.ResourceReference(), databaseRoleModel3.ResourceReference())
 
 	datasourceModelLimitRows := datasourcemodel.DatabaseRoles("test", databaseRoleId1.DatabaseName()).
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -89,7 +89,8 @@ test-acceptance-%: ## run acceptance tests (both non-account and account level o
 	TF_ACC=1 TF_LOG=DEBUG SNOWFLAKE_DRIVER_TRACING=debug SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags=non_account_level_tests,account_level_tests -run ^TestAcc_$* -v -timeout=20m ./pkg/testacc
 
 test-main-terraform-use-cases: ## run test for main terraform use cases
-	TF_ACC=1 TEST_SF_TF_REQUIRE_TEST_OBJECT_SUFFIX=1 TEST_SF_TF_REQUIRE_GENERATED_RANDOM_VALUE=1 SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags=non_account_level_tests,account_level_tests -run "^(TestAcc_.*_BasicUseCase.*|TestAcc_.*_CompleteUseCase.*)$$" -v -cover -json -timeout=20m ./pkg/testacc
+	TF_ACC=1 TEST_SF_TF_REQUIRE_TEST_OBJECT_SUFFIX=1 TEST_SF_TF_REQUIRE_GENERATED_RANDOM_VALUE=1 SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true go test --tags=account_level_tests -run "^(TestAcc_.*_BasicUseCase.*|TestAcc_.*_CompleteUseCase.*)$$" -v -cover -json -timeout=20m ./pkg/testacc
```
sfc-gh-jmichalak: Nit: we could extract the envs and other `go test` parameters to variables.

sfc-gh-jcieslak: Came back to the previous setting as it should work as one command

sfc-gh-jmichalak (approved):

pkg/testacc/data_source_database_roles_acceptance_test.go
```diff
@@ -29,15 +29,12 @@ func TestAcc_DatabaseRoles_BasicUseCase_DifferentFiltering(t *testing.T) {
 
 	datasourceModelLikeExact := datasourcemodel.DatabaseRoles("test", databaseRoleId1.DatabaseName()).
 		WithLike(databaseRoleId1.Name()).
```
sfc-gh-asawicki, line 31: it would be good to keep a test for like only, we could add a random prefix or something and use the factory without cache (if the context does something to the test as described in the PR description)

pkg/testacc/data_source_warehouses_acceptance_test.go
```diff
@@ -96,7 +96,8 @@ func TestAcc_Warehouses_CompleteUseCase(t *testing.T) {
 			HasRunningNotEmpty().
 			HasQueuedNotEmpty().
 			HasIsDefault(false).
-			HasAutoSuspend(600).
+			// TODO(SNOW-2852741): Different auto suspend default on different test environments
```
sfc-gh-asawicki, line 99: hmm, this is strange, let's discuss in the morning

pkg/testacc/data_source_database_roles_acceptance_test.go
sfc-gh-jcieslak: adjusted this test (reverted and used no cache), but I left others with `ON_ACCOUNT` if that's okay. Additionally, adjusted the services with no cache.

pkg/testacc/data_source_warehouses_acceptance_test.go
sfc-gh-jcieslak: Discussed, will work on it "on the side" to clear things up with warehouse team

sfc-gh-asawicki (approved):

PR #4221, opened 2025-11-20
feat: Enable semantic views

sfc-gh-asawicki:
- Add `snowflake_semantic_view` resource and `snowflake_semantiv_views` data source
- Fix/improve the implementation
  - Handle renaming through `ALTER`
  - Handle objects in a case-sensitive way
  - Fix metrics with window functions
  - Add more integration tests and improve the acceptance ones
  - Remove describe output from resource and data source
  - Adjust the SDK
  - Mark `table_alias` as required
  - Move SDK gen definition to the new dir
- Generate the docs
- Document al the limitations
- Update the migration guide

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -56,6 +56,14 @@ The provider offers two main types of logging:
 - Terraform execution (check [Terraform Debugging Documentation](https://www.terraform.io/internals/debugging)) - you can set it through the `TF_LOG` environment variable, e.g.: `TF_LOG=DEBUG`; it will make output of the Terraform execution more verbose.
 - Snowflake communication (using the logs from the underlying [Go Snowflake driver](https://github.com/snowflakedb/gosnowflake)) - you can set it directly in the provider config ([`driver_tracing`](https://registry.terraform.io/providers/snowflakedb/snowflake/1.0.3/docs#driver_tracing-3) attribute), by `SNOWFLAKE_DRIVER_TRACING` environmental variable (e.g. `SNOWFLAKE_DRIVER_TRACING=info`), or by `drivertracing` field in the TOML file. To see the communication with Snowflake (including the SQL commands run) we recommend setting it to `info`.
 
+**Note:** Since v2.11.0 provider version, queries are not logged by default. If you need query-level debugging, ensure you configure appropriate logging settings:
+  - `log_query_text` - when set to `true`, query text will be logged
+  - `log_query_parameters` - when set to `true`, query parameters will be logged
+
+These options can be set in the provider configuration, TOML configuration file, or via environment variables. Read [the documentation](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs#schema) for more details.
+
+Note that you still need to set the `INFO` level in `driver_tracing` field to see the query logs.
+
 As driver logs may seem cluttered, to locate the SQL commands run, search for:
 - (preferred) `--terraform_provider_usage_tracking`
 - `msg="Query:`
@@ -74,6 +82,6 @@ Please refer to [this document](https://github.com/snowflakedb/terraform-provide
 
 ### Is this provider compatible with OpenTofu?
 OpenTofu is not currently supported. While it's mostly compatible with Terraform,
-OpenTofu's [reactive implementation of new Terraform features based on community demand](https://opentofu.org/faq/#opentofu-compatibility) may decrease compatibility over time. 
+OpenTofu's [reactive implementation of new Terraform features based on community demand](https://opentofu.org/faq/#opentofu-compatibility) may decrease compatibility over time.
 We plan to research OpenTofu support in the future, but there's no timeline yet (once planned, it will appear in the [roadmap](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/ROADMAP.md)).
 For now, you must research and assess the risk of provider incompatibility.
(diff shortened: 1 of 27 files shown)


docs/data-sources/semantic_views.md
```diff
@@ -0,0 +1,185 @@
...
+}
+
+output "limit_output" {
+  value = data.snowflake_semantic_views.limit.semantic_views
+}
+
+# Without additional data (to limit the number of calls made for every found semantic view)
+data "snowflake_semantic_views" "only_show" {
+  # with_describe is turned on by default and it calls DESCRIBE SEMANTIC VIEW for every semantic view found and attaches its output to semantic_views.*.describe_output field
+  with_describe = false
```
sfc-gh-jcieslak: Adjust the example as the data source doesn't have `with_describe` yet

pkg/resources/semantic_views.go
```diff
@@ -21,55 +21,55 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
 		Type:             schema.TypeString,
 		Required:         true,
-		ForceNew:         true,
 		Description:      blocklistedCharactersFieldDescription("Specifies the identifier for the semantic view; must be unique within the schema."),
 		DiffSuppressFunc: suppressIdentifierQuoting,
 	},
 	"database": {
 		Type:             schema.TypeString,
 		Required:         true,
-		ForceNew:         true,
```
sfc-gh-jcieslak, line 31: Shouldn't database and schema stay as force new?

pkg/resources/semantic_view.go
```diff
@@ -21,55 +21,55 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
 		ForceNew:    true,
-		Description: "The list of logical tables in the semantic view.",
+		Description: externalChangesNotDetectedFieldDescription("The list of logical tables in the semantic view."),
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"table_alias": {
 					Type:        schema.TypeString,
-					Optional:    true,
-					Description: "Specifies an alias for a logical table in the semantic view.",
+					Required:    true,
```
sfc-gh-jcieslak, line 48: Isn't it optional? At least documentation suggests it `[ <table_alias> AS ] <table_name>`

pkg/resources/semantic_view.go
```diff
@@ -498,7 +501,20 @@ func UpdateSemanticView(ctx context.Context, d *schema.ResourceData, meta any) d
 		return diag.FromErr(err)
 	}
 
-	// TODO [SNOW-2108211]: handle rename through ALTER ... RENAME to
+	if d.HasChange("name") || d.HasChange("schema") || d.HasChange("database") {
```
sfc-gh-jcieslak, line 504: I think we should only react to name changes / I'm not sure if it's possible to do a rename and change schema at the same time. I think in rename you can only manipulate the name

sfc-gh-jcieslak (dismissed):

pkg/resources/semantic_views.go
sfc-gh-jcieslak: I saw you tested different schema scenario, but didn't find the different database one

pkg/resources/semantic_view.go
sfc-gh-jcieslak: saw the in-code comment

pkg/resources/semantic_view.go
sfc-gh-jcieslak: related to the comment above about force new

sfc-gh-asawicki: There are tests showing you can, also the docs:
```
You can move the object to a different database and/or schema while optionally renaming the object. To do so, specify a qualified new_name value that includes the new database and/or schema name in the form <db_name>.<schema_name>.<object_name> or <schema_name>.<object_name>, respectively.
```

pkg/testacc/data_source_semantic_views_acceptance_test.go
```diff
@@ -75,158 +66,70 @@ func TestAcc_SemanticViews_Basic(t *testing.T) {
...
 		WithDependsOn(model1.ResourceReference(), model2.ResourceReference(), model3.ResourceReference())
 
 	dataSourceModelWithLimit := datasourcemodel.SemanticViews("test").
-		WithRowsAndFrom(2, "").
+		WithRowsAndFrom(2, prefix).
 		WithDependsOn(model1.ResourceReference(), model2.ResourceReference(), model3.ResourceReference())
 
 	resource.Test(t, resource.TestCase{
-		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+		ProtoV6ProviderFactories: providerFactoryWithoutCache(),
```
sfc-gh-jmichalak, line 132: Why without cache?

sfc-gh-jmichalak (commented):
Merge conflicts and +1 to Jan's comment.

pkg/testacc/data_source_semantic_views_acceptance_test.go
sfc-gh-asawicki: The test earlier failed with 0 instead of 2 hits sometimes. I was not able to reproduce it locally. I suspect it's the context problem if reusing a previously instantiated connection to Snowflake as `SHOW` without `IN` specified uses the current context. If we don;t see similar errors in the next few weeks, it would prove it was helpful.

docs/data-sources/semantic_views.md
sfc-gh-asawicki: Adjusted

pkg/resources/semantic_views.go
sfc-gh-asawicki: Added both itn and acc test for database rename

sfc-gh-jmichalak (approved):

PR #4226, opened 2025-11-21
chore: Adjust task documentation

sfc-gh-jcieslak:
### Changes
- Adjust update handling for deprecated parameters and return errors on `set` "events"
- Add migration guide entry to inform users

### DIFF
--- a/docs/resources/task.md
+++ b/docs/resources/task.md
@@ -9,7 +9,7 @@ description: |-
 
 !> **Warning** Due to complex conditional logic in Snowflake regarding required fields for a given type (serverless or not) and use case (scheduled, run after another task, conditional, or other) of a task, please make sure you went through the [Snowflake documentation for tasks](https://docs.snowflake.com/en/user-guide/tasks-intro) to understand which fields are required for your use case.
 
-!> **Warning** Setting `AUTOCOMMIT` to `false` and `SEARCH_PATH` to any value is now not possible for tasks in Snowflake, thus it's recommended to not configure them within the resource. The only exception in leaving (or setting) `AUTOCOMMIT` with `true` value explicitly set in the configuration, is to override the parameter that may be set to `false` value on the higher parameter hierarchy level. The parameters may be removed in the next major version of the provider.
+!> **Warning** Setting `AUTOCOMMIT` to `false` and `SEARCH_PATH` to any value is now not possible for tasks in Snowflake, thus it's recommended to not configure them within the resource. The only exception is leaving (or setting) `AUTOCOMMIT` with `true` value explicitly set in the configuration, is to override the parameter that may be set to `false` value on the higher parameter hierarchy level. The parameters may be removed in the next major version of the provider.
 
 # snowflake_task (Resource)
 
--- a/templates/resources/task.md.tmpl
+++ b/templates/resources/task.md.tmpl
@@ -13,7 +13,7 @@ description: |-
 
 !> **Warning** Due to complex conditional logic in Snowflake regarding required fields for a given type (serverless or not) and use case (scheduled, run after another task, conditional, or other) of a task, please make sure you went through the [Snowflake documentation for tasks](https://docs.snowflake.com/en/user-guide/tasks-intro) to understand which fields are required for your use case.
 
-!> **Warning** Setting `AUTOCOMMIT` to `false` and `SEARCH_PATH` to any value is now not possible for tasks in Snowflake, thus it's recommended to not configure them within the resource. The only exception in leaving (or setting) `AUTOCOMMIT` with `true` value explicitly set in the configuration, is to override the parameter that may be set to `false` value on the higher parameter hierarchy level. The parameters may be removed in the next major version of the provider.
+!> **Warning** Setting `AUTOCOMMIT` to `false` and `SEARCH_PATH` to any value is now not possible for tasks in Snowflake, thus it's recommended to not configure them within the resource. The only exception is leaving (or setting) `AUTOCOMMIT` with `true` value explicitly set in the configuration, is to override the parameter that may be set to `false` value on the higher parameter hierarchy level. The parameters may be removed in the next major version of the provider.
 
 # {{.Name}} ({{.Type}})
 


pkg/resources/task_parameters.go
```diff
@@ -401,6 +399,42 @@ func handleTaskParametersUpdate(d *schema.ResourceData, set *sdk.TaskSetRequest,
 		handleParameterUpdate(d, sdk.TaskParameterUseCachedResult, &set.SessionParameters.UseCachedResult, &unset.SessionParametersUnset.UseCachedResult),
 		handleParameterUpdate(d, sdk.TaskParameterWeekOfYearPolicy, &set.SessionParameters.WeekOfYearPolicy, &unset.SessionParametersUnset.WeekOfYearPolicy),
 		handleParameterUpdate(d, sdk.TaskParameterWeekStart, &set.SessionParameters.WeekStart, &unset.SessionParametersUnset.WeekStart),
+		func() diag.Diagnostics {
+			key := strings.ToLower(string(sdk.TaskParameterAutocommit))
+			if d.HasChange(key) || !d.GetRawPlan().AsValueMap()[key].IsKnown() {
+				if !d.GetRawConfig().AsValueMap()[key].IsNull() {
+					value := d.Get(key).(bool)
+					set.SessionParameters.Autocommit = sdk.Bool(value)
```
sfc-gh-jmichalak: I think we should:  
\- disallow the autocommit parameter ONLY if it's set to unknown OR it's set to false.  
\- disallow the search_path parameter ONLY if it's set to any value in the config (incl. unknown).  
\- do similar changes during task creation.  
\- IIRC, it's not possible to assert the warnings, so I guess we can't test it easily :/

MIGRATION_GUIDE.md
```diff
@@ -35,6 +35,22 @@ Previously, the provider's file reading functions did not clean paths. In this v
...
 
+### Task deprecated parameter handling
+
+Recently, Snowflake deprecated both [`AUTOCOMMIT`](https://docs.snowflake.com/en/sql-reference/parameters#autocommit) and [`SEARCH_PATH`](https://docs.snowflake.com/en/sql-reference/parameters#search-path) parameters for tasks.
+The `AUTOCOMMIT` parameter can be only set to `TRUE` (default value for this parameter), and `SEARCH_PATH` cannot be set at all. Both parameters can be unset.
+Now, when either `AUTOCOMMIT` or `SEARCH_PATH` is set in the configuration, the task resource will return warnings saying:
+
+```
+Deprecated parameter 'AUTOCOMMIT' cannot be set on a task
+Deprecated parameter 'SEARCH_PATH' cannot be set on a task
```
sfc-gh-jmichalak: 1. Is autocommit officially deprecated? Because it's okay to set the value to true still.
2. We deprecate SEARCH_PATH with a warning, but not in the schema? I think it's entirely deprecated in SF, no?

docs/resources/task.md
```diff
@@ -7,6 +7,8 @@ description: |-
 
 !> **Sensitive values** This resource's `config`, `show_output.config` and `show_output.definition` fields are not marked as sensitive in the provider. Ensure that no personal data, sensitive data, export-controlled data, or other regulated data is entered as metadata when using the provider. For more information, see [Sensitive values limitations](../#sensitive-values-limitations) and [Metadata fields in Snowflake](https://docs.snowflake.com/en/sql-reference/metadata).
 
+!> **Warning** The `AUTOCOMMIT` and `SEARCH_PATH` are deprecated session parameters for tasks in Snowflake. It is not possible to set them in Snowflake, thus it's recommended to not configure them within the resource. The parameters may be removed in the next major version of the provider.
```
sfc-gh-asawicki: AUTOCOMMIT can be set to TRUE

pkg/resources/task_parameters.go
sfc-gh-asawicki: Could we instead leave the old behavior here and just add the `ValidateDiagFunc` for each of them, that checks the values like @sfc-gh-jmichalak described?

MIGRATION_GUIDE.md
```diff
@@ -35,6 +35,22 @@ Previously, the provider's file reading functions did not clean paths. In this v
 
 No changes in configuration and state are required. The supported TOML location `~/.snowflake/config` stays the same and the behavior shouldn't be affected.
 
+### Task deprecated parameter handling
+
+Recently, Snowflake deprecated both [`AUTOCOMMIT`](https://docs.snowflake.com/en/sql-reference/parameters#autocommit) and [`SEARCH_PATH`](https://docs.snowflake.com/en/sql-reference/parameters#search-path) parameters for tasks.
```
sfc-gh-asawicki: It's not deprecated, the validation was moved from runtime to creation alrteration. Please mention that setting these values previously on tasks was resulting in tasks that fail to start.

sfc-gh-asawicki (changes requested):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: It's for sure deprecated in tasks, but I'm not sure about the deprecation as a whole parameter. In this case, we mostly wanted to address tasks.
1. Yes, but setting it to true doesn't really change anything as it's the default
2. It can still be set on different levels, it just didn't make sense in the task context

docs/resources/task.md
sfc-gh-jcieslak: Well yes, but only because it's the default value. I was wondering if it makes sense to return warnings on non-allowed values or whenever someone is setting a value. In the end, decided to go with the latter as they're deprecated for a reason, setting values for them (other than defaults) on tasks doesn't make sense, so they should be unset.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: True, adjusted

pkg/resources/task_parameters.go
sfc-gh-jcieslak: 1. This logic is copied from usual parameter update handling so it takes into account if value was set from config or not. As far as I remember we wanted to keep old behavior (thus copy paste) with warnings whenever the parameters are set to their incorrect values in addition to error from Snowflake.
2. Added case in create
3. I think validate func doesn't have enough state and plan context to be considered unfortunately
4. Will test locally

docs/resources/task.md
sfc-gh-asawicki: This warning still shows they are deprecated (and AUTOCOMMIT still can be set - I dunno about the search path but let's nto use the deprecated word)

MIGRATION_GUIDE.md
```diff
@@ -89,6 +89,23 @@ Previously, the provider's file reading functions did not clean paths. In this v
 
 No changes in configuration and state are required. The supported TOML location `~/.snowflake/config` stays the same and the behavior shouldn't be affected.
 
+### Task deprecated parameter handling
+
+Recently, Snowflake moved validation from runtime to CREATE/ALTER operations.
```
sfc-gh-asawicki: let's be clear that this validation move was only for these two parameters (it's not e.g. for dag creation), let's be precise

MIGRATION_GUIDE.md
```diff
@@ -89,6 +89,23 @@ Previously, the provider's file reading functions did not clean paths. In this v
...
+The `AUTOCOMMIT` parameter can be only set to `TRUE` (default value for this parameter), and `SEARCH_PATH` cannot be set at all. Both parameters can be unset.
+Now, when either `AUTOCOMMIT` is set to `FALSE` or `SEARCH_PATH` is set in the configuration (when creating or changing), the task resource will return warnings saying:
+
+```
+Deprecated parameter 'AUTOCOMMIT' cannot be set to FALSE on a task
+Deprecated parameter 'SEARCH_PATH' cannot be set on a task
+```
+
+If you have any of these parameters set in your configuration,
+please remove them to avoid the Terraform warnings and potential errors from the Snowflake side.
```
sfc-gh-asawicki, line 105: TRUE on AUTOCOMMIT is available so let's not advise to remove it (as it maybe needed in cases where AUTOCOMMIT is FALSE on account level param)

MIGRATION_GUIDE.md
```diff
@@ -89,6 +89,23 @@ Previously, the provider's file reading functions did not clean paths. In this v
 
 No changes in configuration and state are required. The supported TOML location `~/.snowflake/config` stays the same and the behavior shouldn't be affected.
 
+### Task deprecated parameter handling
```
sfc-gh-jmichalak: Maybe "Task parameter validation handling"? Or smth, cause they aren't officially deprecated.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: It's not default when it's set as FALSE on schema, right?

sfc-gh-jmichalak (changes requested):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Adjusted

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Adjusted

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Yes, thank you

docs/resources/task.md
sfc-gh-jcieslak: Sure, adjusted

MIGRATION_GUIDE.md
sfc-gh-jcieslak: why schema? It can be set on account or session level

pkg/resources/task_parameters.go
```diff
@@ -401,6 +425,43 @@ func handleTaskParametersUpdate(d *schema.ResourceData, set *sdk.TaskSetRequest,
...
+							},
+						}
+					}
+					set.SessionParameters.Autocommit = sdk.Bool(true)
+				} else {
+					unset.SessionParametersUnset.Autocommit = sdk.Bool(true)
+				}
+			}
+			return nil
+		}(),
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Logic should stay as it is now

MIGRATION_GUIDE.md
sfc-gh-jcieslak: As checked when creating a new task without those parameters, when:
- Parameter is set on session level, nothing happens
- Parameter is set on account level, the value is propagated to the task (even the bad ones)

docs/resources/task.md
```diff
@@ -9,6 +9,8 @@ description: |-
 
 !> **Warning** Due to complex conditional logic in Snowflake regarding required fields for a given type (serverless or not) and use case (scheduled, run after another task, conditional, or other) of a task, please make sure you went through the [Snowflake documentation for tasks](https://docs.snowflake.com/en/user-guide/tasks-intro) to understand which fields are required for your use case.
 
+!> **Warning** Setting `AUTOCOMMIT` to `false` and `SEARCH_PATH` to any value is now not possible for tasks in Snowflake, thus it's recommended to not configure them within the resource. The only exception in leaving (or setting) `AUTOCOMMIT` with `true` value explicitly set in the configuration, is to override the parameter that may be set to `false` value on the higher parameter hierarchy level. The parameters may be removed in the next major version of the provider.
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: adjusted

sfc-gh-jmichalak (approved):

PR #4236, opened 2025-11-24
feat: Support schemas in the migration tool

sfc-gh-jmichalak:
- Add schema support to the migration tool.

Open questions (next PR?)
- Should we embed SDK objects, so we have their methods working out of the box?
- Should we have acceptance tests for all objects?

### DIFF
--- a/pkg/scripts/migration_script/README.md
+++ b/pkg/scripts/migration_script/README.md
@@ -538,48 +538,74 @@ But outputs from all of these commands must be mapped to the input CSV value of 
 SQL queries and returning all necessary information.
 
 In general, what we need to do is:
-1. Get the necessary objects in a object data source.
-1. Export the data source state with output blocks.
-1. Process the output blocks data to match the expected CSV in the script.
-1. Run the migration script with processed data.
+1. Define a data source for the objects you want to import.
+1. Use HCL (Terraform's configuration language) to transform the data: merge `show_output` with flattened `parameters` for each object.
+1. Write the transformed data to a CSV file using the `local_file` resource.
+1. Run the migration script with the generated CSV file.
+
+> **Note:** It's recommended to create a fresh Terraform environment (e.g., a new local directory with a clean state) for this data extraction process to avoid collisions with any existing data sources or state in your main Terraform workspace.
 
 Now, let's look into more details.
 
-As an example, let's import all schemas in a given database. First, we need to define a data source for schemas and declare an output:
+As an example, let's import all schemas in a given database. First, we need to define a data source for schemas and use Terraform's HCL to transform the data into CSV format:
+
 ```terraform
+terraform {
+  required_providers {
+    snowflake = {
+      source = "Snowflake-Labs/snowflake"
+    }
+    local = {
+      source = "hashicorp/local"
+    }
+  }
+}
+
 data "snowflake_schemas" "test" {
   in {
     database = "DATABASE"
   }
 }
 
-output "schemas" {
-  value = data.snowflake_schemas.test.schemas
+locals {
+  # Transform each schema by merging show_output and flattened parameters
+  schemas_flattened = [
+    for schema in data.snowflake_schemas.test.schemas : merge(
+      schema.show_output[0],
+      # Flatten parameters: convert each parameter to {param_name}_value and {param_name}_level
+      {
+        for param_key, param_values in schema.parameters[0] :
+        param_key => param_values[0].value
+      },
+      {
+        for param_key, param_values in schema.parameters[0] :
+        "${param_key}_level" => param_values[0].level
+      }
+    )
+  ]
+
+  # Get all unique keys from the first schema to create CSV header
+  csv_header = join(",", [for key in keys(local.schemas_flattened[0]) : "\"${key}\""])
+
+  # Convert each schema object to CSV row
+  csv_rows = [
+    for schema in local.schemas_flattened :
+      join(",", [for key in keys(local.schemas_flattened[0]) : "\"${lookup(schema, key, "")}\""])
+  ]
+
+  # Combine header and rows
+  csv_content = join("\n", concat([local.csv_header], local.csv_rows))
+}
+
+resource "local_file" "schemas_csv" {
+  content  = local.csv_content
+  filename = "${path.module}/schemas.csv"
 }
 ```
 
-Now, get the data with `terraform apply`. The schemas should be present in your state and in the declared output. Now, we can perform some manipulations to convert
-the output format to the desired CSV format:
-1. Extract the `show_output` data to a temporary file.
-1. Extract the `parameters` data to a temporary file and convert it to a "key": "value" format instead of list of attributes.
-1. Merge the two files into one. The resulting table is a list of objects with merged fields from `show_output` and `parameters`.
-1. Convert the JSON file to a CSV file. This CSV file can be now used in the migration script.
-
-You can find the given steps in bash below:
-```bash
-tf output -json | jq '.schemas[]' > output.json
-cat output.json | jq '[.schemas.value[] | .show_output[0]]' > show_output.json
-cat output.json | jq '[.schemas.value[] | .parameters[0] | to_entries | map({("\(.key)_value"): .value[0].value, ("\(.key)_level"): .value[0].level}) | add]' > parameters.json
-jq -s 'transpose | map(add)' show_output.json parameters.json > merged.json
-jq -r '
-  (.[0] | keys_unsorted) as $keys |
-  ($keys, (.[] | [.[$keys[]]])) | @csv
-' merged.json > merged.csv
-```
-
-Now, we can run the migration script like:
+After running `terraform apply`, the CSV file will be automatically generated at `schemas.csv`. Now, we can run the migration script like:
 ```shell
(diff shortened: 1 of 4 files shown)


pkg/scripts/migration_script/README.md
```diff
@@ -526,6 +532,58 @@ Remember that, if you chose to use the import block approach, [after importing y
...
+
+As an example, let's import all schemas in a given database. First, we need to define a data source for schemas and declare an output:
+```terraform
+data "snowflake_schemas" "test" {
+  in {
+    database = "DATABASE"
+  }
+}
+
+output "schemas" {
```
sfc-gh-jcieslak: Is there a way to redirect an output to a file? If yes, maybe we could add this as console output may be cumbersome to copy. Even if it requires other provider (e.g. one provided by HC), I think it would be clearer.

pkg/scripts/migration_script/README.md
```diff
@@ -526,6 +532,58 @@ Remember that, if you chose to use the import block approach, [after importing y
...
+  value = data.snowflake_schemas.test.schemas
+}
+```
+
+Now, get the data with `terraform apply`. The schemas should be present in your state and in the declared output. Now, we can perform some manipulations to convert
+the output format to the desired CSV format:
+1. Extract the `show_output` data to a temporary file.
+1. Extract the `parameters` data to a temporary file and convert it to a "key": "value" format instead of list of attributes.
```
sfc-gh-jcieslak: State what should be a key and what value (also it's pretty confusing as you mentioned above the level and it's not listed here?)

pkg/scripts/migration_script/README.md
sfc-gh-jcieslak: Hmm, okay, I can see the script below, but it works on the whole state (?). In this case, it's worth mentioning that it's better to create a new (perhaps local; or just fresh one since it was I believe tested on such env) Terraform environment to not collide with any existing data source

pkg/scripts/migration_script/converter_schemas.go
```diff
@@ -0,0 +1,150 @@
...
+	StorageSerializationPolicy              *string
+	SuspendTaskAfterNumFailures             *int
+	TaskAutoRetryAttempts                   *int
+	TraceLevel                              *string
+	UserTaskManagedInitialWarehouseSize     *string
+	UserTaskMinimumTriggerIntervalInSeconds *int
+	UserTaskTimeoutMs                       *int
+}
+
+// TODO: embed schema?
```
sfc-gh-jcieslak: Should it be removed since the sdk type is not embedded?

pkg/scripts/migration_script/README.md
```diff
@@ -526,6 +532,58 @@ Remember that, if you chose to use the import block approach, [after importing y
...
+  }
+}
+
+output "schemas" {
+  value = data.snowflake_schemas.test.schemas
+}
+```
+
+Now, get the data with `terraform apply`. The schemas should be present in your state and in the declared output. Now, we can perform some manipulations to convert
+the output format to the desired CSV format:
```
sfc-gh-asawicki: This is the place where I had concerns when we discussed. Essentially, our goal is to for each schema, extract three outputs, merge them, and present the result as a single entry.

There are manipulation tools in HCL (for_each, flatten, etc.), so I think that instead of processing the state, we could prepare an output that already has a proper structure (as you should be able to achieve all these jq transformations with HCL), and save it directly to csv (e.g. https://registry.terraform.io/providers/hashicorp/local/latest/docs/resources/file), or pipe it directly to script (if possible).

I'd rather run the prepared HCL than running such a script.

The important part for now could be, that we accept in the script the combined csv, so @sfc-gh-mkowalski can work on adding support for other objects, but let's refine the getting the input part.

pkg/scripts/migration_script/converter_schemas.go
sfc-gh-asawicki: I think we could, we use sdk.Grant for grants after all

pkg/scripts/migration_script/mappings_schemas.go
```diff
@@ -0,0 +1,56 @@
...
+func HandleSchemas(config *Config, csvInput [][]string) (string, error) {
+	return HandleResources[SchemaCsvRow, SchemaRepresentation](config, csvInput, MapSchemaToModel)
+}
+
+func MapSchemaToModel(schema SchemaRepresentation) (accconfig.ResourceModel, *ImportModel, error) {
+	schemaId := sdk.NewDatabaseObjectIdentifier(schema.DatabaseName, schema.Name)
+	resourceId := NormalizeResourceId(fmt.Sprintf("schema_%s", schemaId.FullyQualifiedName()))
+	resourceModel := model.Schema(resourceId, schema.DatabaseName, schema.Name)
+
+	if schema.Comment != "" {
```
sfc-gh-asawicki: nit: could we extract these conditionals into helpers too (e.g. handleIfNotEmpty, handleIf, or something like that)

pkg/scripts/migration_script/mappings_schemas_test.go
```diff
@@ -0,0 +1,117 @@
+package main
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHandleSchemaMappings(t *testing.T) {
```
sfc-gh-asawicki, line 10: nit: I think that for shcemas, where we have a bit better control than we have for grants, we could have a true acceptance test for the whole flow:
1. Setting the infra with terraform
2. getting the csv from snowflake using prepared HCL script
3. running the script on the input from csv
4. verifying that input will cause no drift on Snowflake.

But I think it can be discussed internally first (e.g. if we put it with the acceptance tests or do we want to configure it the other way).

pkg/scripts/migration_script/converter_schemas.go
sfc-gh-jmichalak: Decided to embed, after all.

sfc-gh-jmichalak: Done.

pkg/scripts/migration_script/mappings_schemas.go
sfc-gh-jmichalak: Done.

pkg/scripts/migration_script/README.md
sfc-gh-jmichalak: Done.

pkg/scripts/migration_script/README.md
sfc-gh-jmichalak: The example is below. It does not need the whole state, but just one output. Adjusted with the new env note.

pkg/scripts/migration_script/README.md
sfc-gh-jmichalak: Done (with two, as this object does not need desc_output).

pkg/scripts/migration_script/mappings_schemas_test.go
sfc-gh-jmichalak: Yes, it's part of the later work. With this one, we wanted to agree on a solution, so we can add this in the next PR. I think we could do this in the testacc package for simplicity.

pkg/scripts/migration_script/common.go
```diff
@@ -13,6 +13,18 @@ func handleOptionalFieldWithBuilder[T any](parameter *T, builder func(T) *model.
 	}
 }
 
+func handleIfNotEmpty(value string, builder func(string) *model.SchemaModel) {
+	if value != "" {
+		builder(value)
+	}
+}
+
+func handleIf(condition bool, builder func(string) *model.SchemaModel) {
```
sfc-gh-jcieslak, line 22: Shouldn't it be typed liked `handleBoolIf`?

pkg/scripts/migration_script/README.md
```diff
@@ -538,48 +538,74 @@ But outputs from all of these commands must be mapped to the input CSV value of
...
 
-Now, get the data with `terraform apply`. The schemas should be present in your state and in the declared output. Now, we can perform some manipulations to convert
-the output format to the desired CSV format:
-1. Extract the `show_output` data to a temporary file.
-1. Extract the `parameters` data to a temporary file and convert it to a "key": "value" format instead of list of attributes.
-1. Merge the two files into one. The resulting table is a list of objects with merged fields from `show_output` and `parameters`.
-1. Convert the JSON file to a CSV file. This CSV file can be now used in the migration script.
-
-You can find the given steps in bash below:
-```bash
```
sfc-gh-jcieslak, line 569: nit: Maybe don't remove it completely? As we already have the right command let's leave it as an alternative option to get the CSV file?

sfc-gh-jcieslak (approved):

pkg/scripts/migration_script/common.go
sfc-gh-jmichalak: Maybe, I can change later.

pkg/scripts/migration_script/README.md
sfc-gh-jmichalak: Let's provide one option to avoid confusion. We can later add this change because it will be saved in the review history anyway.

pkg/scripts/migration_script/mappings_schemas_test.go
sfc-gh-asawicki: ok, let's discuss the tests further in the follow-up changes

PR #4241, opened 2025-11-25
chore: Upgrade tools

sfc-gh-jmichalak:
Upgrade the tfplugindocs and gofumpt to the newest versions. Additionally, propose slight adjustments in the template format.

### DIFF
--- a/templates/resources/authentication_policy.md.tmpl
+++ b/templates/resources/authentication_policy.md.tmpl
@@ -1,5 +1,5 @@
 ---
-page_title: "{{.Name}} {{.Type}} - {{.ProviderName}}"
+page_title: "{{.Name}} {{.Type}} - {{.RenderedProviderName}}"
 subcategory: "Preview"
 description: |-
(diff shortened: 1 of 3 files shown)


templates/resources/authentication_policy.md.tmpl
```diff
@@ -32,11 +32,21 @@ description: |-
...
 
 {{ .SchemaMarkdown | trimspace }}
-{{- if .HasImport }}
+{{- if or .HasImport .HasImportIDConfig }}
 
 ## Import
 
 Import is supported using the following syntax:
+{{- end }}
+{{- if .HasImportIDConfig }}
```
sfc-gh-jcieslak, line 41: Q: Is import-by-string-id file name a convention in the new doc generator version?

sfc-gh-jmichalak: Yes, see [here](https://github.com/hashicorp/terraform-plugin-docs/blob/48cb60762516ebca3f1655bc10951be272b502b6/internal/provider/generate.go#L802), and test examples, as the docs don't mention it.

PR #4242, opened 2025-11-25
chore: Make the repository compatible with git-worktree

sfc-gh-jmichalak:
- Add a new configuration for cursor worktrees.
- Add a new flag for provider name in the docs generator. By default, the directory name is used, but this is changed in git workspaces, making the generator fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ dev-cleanup: ## cleanup development dependencies
 	rm -rf tools/bin/*
 
 docs: generate-docs-additional-files ## generate docs
-	tools/bin/tfplugindocs generate
+	tools/bin/tfplugindocs generate --provider-name=terraform-provider-snowflake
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -23,7 +23,7 @@ dev-cleanup: ## cleanup development dependencies
 	rm -rf tools/bin/*
 
 docs: generate-docs-additional-files ## generate docs
-	tools/bin/tfplugindocs generate
+	tools/bin/tfplugindocs generate --provider-name=terraform-provider-snowflake
```
sfc-gh-jcieslak, line 26: Does this change belong here or really it was added in the doc upgrade PR and for some reason landed here?

sfc-gh-jmichalak: It does belong here, as I explained in the PR description.

PR #4243, opened 2025-11-26
feat: Allow for custom run_as_role in snowflake_scim_integration

sfc-gh-jmichalak:
Allow custom roles in `run_as_role` field in scim integrations.

## TODO
- Use custom roles for other acceptance tests.
- Potentially remove the roles setup in tests.

## References
https://github.com/snowflakedb/terraform-provider-snowflake/issues/3917

### DIFF
--- a/pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_ext.go
+++ b/pkg/acceptance/bettertestspoc/assert/objectassert/task_snowflake_ext.go
@@ -88,7 +88,7 @@ func (t *TaskAssert) HasNoTargetCompletionInterval() *TaskAssert {
 	t.AddAssertion(func(t *testing.T, o *sdk.Task) error {
 		t.Helper()
 		if o.TargetCompletionInterval != nil {
-			return fmt.Errorf("expected target completion interval to have no value; got: %s", o.TargetCompletionInterval)
+			return fmt.Errorf("expected target completion interval to have no value; got: %v", o.TargetCompletionInterval)
 		}
 		return nil
 	})


pkg/sdk/security_integrations_gen.go
```diff
@@ -243,10 +243,11 @@ type CreateScimSecurityIntegrationOptions struct {
 	integrationType     string                                  `ddl:"static" sql:"TYPE = SCIM"`
 	Enabled             *bool                                   `ddl:"parameter" sql:"ENABLED"`
 	ScimClient          ScimSecurityIntegrationScimClientOption `ddl:"parameter,single_quotes" sql:"SCIM_CLIENT"`
-	RunAsRole           ScimSecurityIntegrationRunAsRoleOption  `ddl:"parameter,single_quotes" sql:"RUN_AS_ROLE"`
-	NetworkPolicy       *string                                 `ddl:"parameter,no_quotes" sql:"NETWORK_POLICY"`
-	SyncPassword        *bool                                   `ddl:"parameter" sql:"SYNC_PASSWORD"`
-	Comment             *string                                 `ddl:"parameter,single_quotes" sql:"COMMENT"`
+	// adjusted manually
+	RunAsRole     string  `ddl:"parameter,no_quotes" sql:"RUN_AS_ROLE"`
```
sfc-gh-jcieslak: q: Why does it need to be adjusted? Can't def file be adjusted accordingly?

MIGRATION_GUIDE.md
```diff
@@ -24,9 +24,24 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
...
+## v2.11.x ➞ v2.12.0
+
+### *(improvement)* snowflake_scim_integration now accepts custom role names for run_as_role
+
+Previously, the `run_as_role` field in the `snowflake_scim_integration` resource only accepted predefined role names: `OKTA_PROVISIONER`, `AAD_PROVISIONER`, or `GENERIC_SCIM_PROVISIONER`.
+
+Now, the field accepts any custom role name, allowing you to use organization-specific roles for SCIM provisioning. This field is now case-sensitive. The exception is if you set any of `okta_provisioner`, `aad_provisioner`, or `generic_scim_provisioner`.
+To maintain compatibility and avoid breaking changes, for these values, the provider makes them uppercase (like it was doing before). This will be changed in the v3 version of the provider (the field will behave like all other identifier fields).
+- If you use `OKTA_PROVISIONER`, `AAD_PROVISIONER`, or `GENERIC_SCIM_PROVISIONER` roles in Snowflake, please make them uppercase in your configuration (this will result in an empty plan).
+- If you use `okta_provisioner`, `aad_provisioner`, or `generic_scim_provisioner` roles in Snowflake, please handle them in provider with snowflake_execute, or rename the roles entirely.
```
sfc-gh-jcieslak, line 36: 1) maybe link to the resource?
2) I'm not quite sure what `handle them in provider with snowflake_execute` means, as run_as_role can be only set in CREATE the whole resource would have to be either managed fully be execute resource or at least created by it or externally and imported to the regular resource, right?

pkg/sdk/security_integrations_gen.go
sfc-gh-asawicki: +1

docs/resources/scim_integration.md
```diff
@@ -46,7 +46,7 @@ resource "snowflake_scim_integration" "test" {
 
 - `enabled` (Boolean) Specify whether the security integration is enabled.
 - `name` (String) String that specifies the identifier (i.e. name) for the integration; must be unique in your account. Due to technical limitations (read more [here](../guides/identifiers_rework_design_decisions#known-limitations-and-identifier-recommendations)), avoid using the following characters: `|`, `.`, `"`.
-- `run_as_role` (String) Specify the SCIM role in Snowflake that owns any users and roles that are imported from the identity provider into Snowflake using SCIM. Provider assumes that the specified role is already provided. Valid options are: `OKTA_PROVISIONER` | `AAD_PROVISIONER` | `GENERIC_SCIM_PROVISIONER`.
+- `run_as_role` (String) Specify the SCIM role in Snowflake that owns any users and roles that are imported from the identity provider into Snowflake using SCIM. Provider assumes that the specified role is already provided.This field is case-sensitive. The exception is using `generic_scim_provisioner`, `okta_provisioner`, or `aad_provisioner`, which are automatically converted to uppercase for backwards compatibility.
```
sfc-gh-asawicki: nit: missing spase before the newly added part

pkg/acceptance/helpers/role_client.go
```diff
@@ -63,6 +63,32 @@ func (c *RoleClient) CreateRoleGrantedToCurrentUser(t *testing.T) (*sdk.Role, fu
 	return role, roleCleanup
 }
 
+func (c *RoleClient) CreateRoleGrantedToCurrentRole(t *testing.T) (*sdk.Role, func()) {
```
sfc-gh-asawicki, line 66: nit: this method could just call the next one with `c.ids.RandomAccountObjectIdentifier()`

pkg/resources/scim_integration.go
```diff
@@ -239,6 +237,14 @@ func CreateContextSCIMIntegration(ctx context.Context, d *schema.ResourceData, m
 	return ReadContextSCIMIntegration(false)(ctx, d, meta)
 }
 
+func scimIntegrationRunAsRoleToAccountObjectIdentifier(runAsRole string) (sdk.AccountObjectIdentifier, error) {
```
sfc-gh-asawicki, line 240: I feel like this should be a part of SDK and not the tf resource impl

pkg/sdk/security_integrations_gen_test.go
```diff
@@ -445,7 +445,7 @@ func TestSecurityIntegrations_CreateScim(t *testing.T) {
 			// adjusted manually
 			name:       id,
 			ScimClient: "GENERIC",
-			RunAsRole:  "GENERIC_SCIM_PROVISIONER",
+			RunAsRole:  securityIntegrationFullyQualifiedNameQuotedRequired(NewAccountObjectIdentifier("GENERIC_SCIM_PROVISIONER")),
```
sfc-gh-asawicki: nit: could we reuse "GENERIC_SCIM_PROVISIONER" string if we are touching it?

pkg/testacc/resource_scim_integration_acceptance_test.go
```diff
@@ -255,26 +259,6 @@ func TestAcc_ScimIntegration_InvalidScimClient(t *testing.T) {
 	})
 }
 
-func TestAcc_ScimIntegration_InvalidRunAsRole(t *testing.T) {
```
sfc-gh-asawicki, line 258: This tested the validation earlier but what would haoppen if we used non-existent role now (I don't see a test for it)?

MIGRATION_GUIDE.md
sfc-gh-asawicki, line 36: question: why they need to rename the roles if they used lowercase in Snowflake? Did the previous implementation work with lowercase roles on Snowflake?

docs/resources/scim_integration.md
sfc-gh-jmichalak: Fixed.

pkg/acceptance/helpers/role_client.go
sfc-gh-jmichalak: Done.

pkg/sdk/security_integrations_gen_test.go
sfc-gh-jmichalak: We can, but we don't do it usually, and this string does not have any particular semantics right now.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: Previously, it was a lowercase enum, so it worked because we called strings.ToUpper. But now someone may want to use lowercase specifically (why? I don't know, but it's possible).

sfc-gh-jmichalak: 1. Done.
2. Yes, it's only for the use case I described above.

pkg/resources/scim_integration.go
sfc-gh-jmichalak: We do this to maintain compatibility with older provider versions. This is not an SDK-specific behavior (see my explanation in other comments).

pkg/sdk/security_integrations_gen.go
sfc-gh-jmichalak: Done.

sfc-gh-jmichalak: Reverted.

pkg/testacc/resource_scim_integration_acceptance_test.go
sfc-gh-jmichalak: Added.

pkg/sdk/security_integrations_gen_test.go
sfc-gh-asawicki: WDYM? We usually try to minimize the use of magic strings and repetition. Furthermore, this is a predefined role, and we already have it extracted in the `snowflakeroles` package as `GenericScimProvisioner`.

MIGRATION_GUIDE.md
sfc-gh-asawicki: Oh okay, so we wat the previous configs with lowercase to continue working and that's the reason why we won;t work with lowercase roles like this one.

PR #4249, opened 2025-11-28
feat: Migration script warehouses support

sfc-gh-mkowalski:
Added support for warehouses to the migration script

### DIFF
--- a/pkg/scripts/migration_script/README.md
+++ b/pkg/scripts/migration_script/README.md
@@ -89,9 +89,9 @@ where script options are:
   - `warehouses` which expects a converted CSV output from the snowflake_warehouses data source.
       To support object parameters, one should use the SHOW PARAMETERS output, and combine it with the SHOW WAREHOUSES output, so the CSV header looks like `"comment","created_on",...,"max_cluster_count","min_cluster_count","name","other",...`
       When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "WAREHOUSE".
-      The script always outputs fields that have non-empty default values in Snowflake
+      The script always outputs fields that have non-empty default values in Snowflake (they can be removed from the output)
 
-      Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid)
+      Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid, initially_suspended)
 
       For more details about using multiple sources, visit [Multiple sources section](#multiple-sources).
 
--- a/pkg/scripts/migration_script/program.go
+++ b/pkg/scripts/migration_script/program.go
@@ -166,8 +166,8 @@ object_type represents the type of Snowflake object you want to generate terrafo
 		- "warehouses" which expects a converted CSV output from the snowflake_warehouses data source
 			To support object parameters, one should use the SHOW PARAMETERS output, and combine it with the SHOW WAREHOUSES output, so the CSV header looks like "comment","created_on",...,"max_cluster_count","min_cluster_count","name","other",...
 			When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "WAREHOUSE".
-			The script always outputs fields that have non-empty default values in Snowflake
-			Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid)
+			The script always outputs fields that have non-empty default values in Snowflake (they can be removed from the output)
+			Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid, initially_suspended)
 			For more details about using multiple sources, visit https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/scripts/migration_script/README.md#multiple-sources
 			Supported resources:
 				- snowflake_warehouse
--- a/pkg/scripts/migration_script/program_test.go
+++ b/pkg/scripts/migration_script/program_test.go
@@ -67,8 +67,8 @@ object_type represents the type of Snowflake object you want to generate terrafo
 		- "warehouses" which expects a converted CSV output from the snowflake_warehouses data source
 			To support object parameters, one should use the SHOW PARAMETERS output, and combine it with the SHOW WAREHOUSES output, so the CSV header looks like "comment","created_on",...,"max_cluster_count","min_cluster_count","name","other",...
 			When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "WAREHOUSE".
-			The script always outputs fields that have non-empty default values in Snowflake
-			Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid)
+			The script always outputs fields that have non-empty default values in Snowflake (they can be removed from the output)
+			Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid, initially_suspended)
 			For more details about using multiple sources, visit https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/scripts/migration_script/README.md#multiple-sources
 			Supported resources:
 				- snowflake_warehouse


pkg/scripts/migration_script/converter_warehouses.go
```diff
@@ -0,0 +1,135 @@
...
+	ResourceMonitor                      string `csv:"resource_monitor"`
+	ResumedOn                            string `csv:"resumed_on"`
+	Running                              string `csv:"running"`
+	ScalingPolicy                        string `csv:"scaling_policy"`
+	Size                                 string `csv:"size"`
+	StartedClusters                      string `csv:"started_clusters"`
+	State                                string `csv:"state"`
+	Suspended                            string `csv:"suspended"`
+	Type                                 string `csv:"type"`
+	UUID                                 string `csv:"u_u_i_d"`
```
sfc-gh-jcieslak: is the csv field format correct? Id doesn't seem so

pkg/scripts/migration_script/converter_warehouses.go
```diff
@@ -0,0 +1,135 @@
...
+			OwnerRoleType:           row.OwnerRoleType,
+			State:                   sdk.WarehouseState(row.State),
+			ScalingPolicy:           sdk.ScalingPolicy(row.ScalingPolicy),
+			EnableQueryAcceleration: row.EnableQueryAcceleration == "true",
+		},
+	}
+
+	handler := newParameterHandler(sdk.ParameterTypeWarehouse)
+	errs := errors.Join(
+		handler.handleBooleanParameter(sdk.ParameterType(row.AutoResumeLevel), row.AutoResumeValue, &warehouseRepresentation.AutoResume),
```
sfc-gh-jcieslak: nit: can those functions accept string and convert to ParameterType internally? Also isn't there a parsing function? This way it would be safer instead of mapping whatever string to ParameterType

pkg/scripts/migration_script/program.go
```diff
@@ -161,6 +163,13 @@ object_type represents the type of Snowflake object you want to generate terrafo
 			Warning: currently secondary databases and shared databases are treated as plain databases.
 			Supported resources:
 				- snowflake_database
+		- "warehouses" which expects a converted CSV output from the snowflake_warehouses data source
+			To support object parameters, one should use the SHOW PARAMETERS output, and combine it with the SHOW WAREHOUSES output, so the CSV header looks like "comment","created_on",...,"max_cluster_count","min_cluster_count","name","other",...
+			When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "ACCOUNT".
+			For more details about using multiple sources, visit https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/scripts/migration_script/README.md#multiple-sources
+			Warning: currently secondary warehouses and shared warehouses are treated as plain warehouses.
```
sfc-gh-jcieslak: there's no such thing as secondary or shared warehouses

pkg/scripts/migration_script/README.md
```diff
@@ -81,6 +81,13 @@ where script options are:
       For more details about using multiple sources, visit the [Multiple sources section](#multiple-sources).
       Supported resources:
         - snowflake_database
+  - `warehouses` which expects a converted CSV output from the snowflake_warehouses data source
+      To support object parameters, one should use the SHOW PARAMETERS output, and combine it with the SHOW WAREHOUSES output, so the CSV header looks like `"comment","created_on",...,"max_cluster_count","min_cluster_count","name","other",...`
+      When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "ACCOUNT".
+      For more details about using multiple sources, visit [Multiple sources section](#multiple-sources).
+      Warning: currently secondary warehouses and shared warehouses are treated as plain warehouses.
```
sfc-gh-jcieslak: remove here also

pkg/scripts/migration_script/README.md
```diff
@@ -81,6 +81,12 @@ where script options are:
       For more details about using multiple sources, visit the [Multiple sources section](#multiple-sources).
       Supported resources:
         - snowflake_database
+  - `warehouses` which expects a converted CSV output from the snowflake_warehouses data source
+      To support object parameters, one should use the SHOW PARAMETERS output, and combine it with the SHOW WAREHOUSES output, so the CSV header looks like `"comment","created_on",...,"max_cluster_count","min_cluster_count","name","other",...`
+      When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "ACCOUNT".
```
sfc-gh-jmichalak:
```suggestion
      When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "WAREHOUSE".
```

pkg/scripts/migration_script/README.md
```diff
@@ -81,6 +81,12 @@ where script options are:
...
+      For more details about using multiple sources, visit [Multiple sources section](#multiple-sources).
```
sfc-gh-jmichalak, line 96: Nit: please fix formatting here, as the rendered document has a few issues like missing newlines. It concerns other objects as well.

pkg/scripts/migration_script/converter_warehouses.go
sfc-gh-jmichalak: There are only 3 parameters:

| MAX_CONCURRENCY_LEVEL |
| --- |
| STATEMENT_QUEUED_TIMEOUT_IN_SECONDS |
| STATEMENT_TIMEOUT_IN_SECONDS |

Fields like WarehouseTypeValue, etc are not parameters and they shouldn't be handled like this, but like regular fields.

sfc-gh-jmichalak: Some fields are missing from data source: actives, failed, ...

pkg/scripts/migration_script/program_test.go
```diff
@@ -232,6 +238,44 @@ import {
...
+  name = "WH_PROD"
+  auto_resume = "true"
+  auto_suspend = 450
+  comment = "Production warehouse"
+  enable_query_acceleration = "true"
+  generation = "1"
+  initially_suspended = true
+  max_cluster_count = 4
+  min_cluster_count = 1
+  query_acceleration_max_scale_factor = 30
```
sfc-gh-jmichalak: What about the parameters? They're missing here.

pkg/scripts/migration_script/mappings_warehouses_test.go
```diff
@@ -0,0 +1,106 @@
...
+}
+import {
+  to = snowflake_warehouse.snowflake_generated_warehouse_WH1
+  id = "\"WH1\""
+}
+`,
+		},
+		{
+			name: "minimal warehouse",
+			inputRows: [][]string{
```
sfc-gh-jmichalak, line 42: Maybe I'm missing something, but what's the difference with the above?

pkg/scripts/migration_script/mappings_warehouses_test.go
```diff
@@ -0,0 +1,106 @@
...
...vailable", "comment", "created_on", "enable_query_acceleration", "failed", "generation", "is_current", "is_default", "max_cluster_count", "min_cluster_count", "name", "other", "owner", "owner_role_type", "pendings", "provisioning", "query_acceleration_max_scale_factor", "queued", "quiescing", "resource_constraint", "resource_monitor", "resumed_on", "running", "scaling_policy", "size", "started_clusters", "state", "suspended", "type", "uuid", "updated_on", "auto_resume_level", "auto_resume_value", "auto_suspend_level", "auto_suspend_value", "comment_level", "comment_value", "enable_query_acceleration_level", "enable_query_acceleration_value", "generation_level", "generation_value", "initially_suspended_level", "initially_suspended_value", "max_cluster_count_level", "max_cluster_count_value", "min_cluster_count_level", "min_cluster_count_value", "query_acceleration_max_scale_factor_level", "query_acceleration_max_scale_factor_value", "resource_constraint_level", "resource_constraint_value", "resource_monitor_level", "resource_monitor_value", "scaling_policy_level", "scaling_policy_value", "warehouse_size_level", "warehouse_size_value", "warehouse_type_level", "warehouse_type_value"},
```
sfc-gh-jmichalak: Missing parameters (max_concurrency_level, etc)

pkg/scripts/migration_script/mappings_warehouses.go
```diff
@@ -0,0 +1,41 @@
...
+func HandleWarehouses(config *Config, csvInput [][]string) (string, error) {
+	return HandleResources[WarehouseCsvRow, WarehouseRepresentation](config, csvInput, MapWarehouseToModel)
+}
+
+func MapWarehouseToModel(warehouse WarehouseRepresentation) (accconfig.ResourceModel, *ImportModel, error) {
+	warehouseId := sdk.NewAccountObjectIdentifier(warehouse.Name)
+	resourceId := NormalizeResourceId(fmt.Sprintf("warehouse_%s", warehouseId.FullyQualifiedName()))
+	resourceModel := model.Warehouse(resourceId, warehouse.Name)
+
+	handleOptionalFieldWithStringBuilder(warehouse.Comment, resourceModel.WithComment)
```
sfc-gh-jmichalak: Why not handleIfNotEmpty?

pkg/scripts/migration_script/program_test.go
```diff
@@ -232,6 +238,34 @@ import {
...
+"true","600","71.43","","2024-06-06 00:00:00.000 +0000 UTC","false","","false","false","3","1","WH_BASIC","0","ADMIN","ROLE","0","0","1","0","","","2024-06-06 12:00:00.000 +0000 UTC","5","ECONOMY","XSMALL","2","AVAILABLE","STANDARD","2024-06-06 00:00:00.000 +0000 UTC","","","","","",""
+"true","450","80.00","Production warehouse","2024-06-06 00:00:00.000 +0000 UTC","true","1","false","false","4","1","WH_PROD","0","ADMIN","ROLE","0","30","1","0","MEMORY_2X","MONITOR1","2024-06-06 12:00:00.000 +0000 UTC","3","CLASSIC","MEDIUM","1","SUSPENDED","SNOWPARK-OPTIMIZED","2024-06-06 00:00:00.000 +0000 UTC","WAREHOUSE","10","WAREHOUSE","600","WAREHOUSE","300"`,
+			expectedOutput: `resource "snowflake_warehouse" "snowflake_generated_warehouse_WH_BASIC" {
+  name = "WH_BASIC"
+}
+
+resource "snowflake_warehouse" "snowflake_generated_warehouse_WH_PROD" {
+  name = "WH_PROD"
+  comment = "Production warehouse"
```
sfc-gh-jmichalak: Why this test doesn't produce other fields? E.g. warehouse type or size.

pkg/scripts/migration_script/mappings_warehouses_test.go
```diff
@@ -0,0 +1,97 @@
...
+import {
+  to = snowflake_warehouse.snowflake_generated_warehouse_PROD_WH
+  id = "\"PROD_WH\""
+}
+`,
+		},
+		{
+			name: "warehouse with all parameters set on higher level",
+			inputRows: [][]string{
+				{"auto_resume", "auto_suspend", "available", "comment", "created_on", "enable_query_acceleration", "generation", "is_current", "is_default", "max_cluster_count", "min_cluster_count", "name", "other", "owner", "owner_role_type", "provisioning", "query_acceleration_max_scale_factor", "queued", "quiescing", "resource_constraint", "resource_monitor", "resumed_on", "running", "scaling_policy", "size", "started_clusters", "state", "type", "updated_on", "max_concurrency_level_level", "max_concurrency_level_value", "statement_queued_timeout_in_seconds_level", "statement_queued_timeout_in_seconds_value", "statement_timeout_in_seconds_level", "statement_timeout_in_seconds_value"},
```
sfc-gh-jmichalak: Same here. The asserted value is missing all other fields.

pkg/scripts/migration_script/mappings_warehouses.go
```diff
@@ -0,0 +1,80 @@
...
+
+	if warehouse.EnableQueryAcceleration {
+		resourceModel.WithEnableQueryAcceleration(r.BooleanTrue)
+	}
+
+	if warehouse.QueryAccelerationMaxScaleFactor != defaultQueryAccelerationMaxScaleFactor {
+		resourceModel.WithQueryAccelerationMaxScaleFactor(warehouse.QueryAccelerationMaxScaleFactor)
+	}
+
+	if warehouse.ResourceMonitor.Name() != "" {
```
sfc-gh-jmichalak: These and above should be handled with handleIfNotEmpty. Additionally, here we assume certain default values, which can change (type, size, autoSuspend, etc). Maybe instead, we should just always add these values to the model, and in the docs state that fields with default values in Snowflake are always generated? WDYT @sfc-gh-jcieslak @sfc-gh-asawicki

pkg/scripts/migration_script/program_test.go
sfc-gh-jmichalak: What about: min_cluster_count, auto_resume, resource_constraint/generation? (For resource_constraint and generation we should have separate test because they are exclusive).

pkg/scripts/migration_script/program_test.go
```diff
@@ -232,6 +238,45 @@ import {
...
+"true","600","71.43","","2024-06-06 00:00:00.000 +0000 UTC","false","","false","false","3","1","WH_BASIC","0","ADMIN","ROLE","0","0","1","0","","","2024-06-06 12:00:00.000 +0000 UTC","5","ECONOMY","XSMALL","2","AVAILABLE","STANDARD","2024-06-06 00:00:00.000 +0000 UTC","","","","","",""
+"true","450","80.00","Production warehouse","2024-06-06 00:00:00.000 +0000 UTC","true","1","false","false","4","1","WH_PROD","0","ADMIN","ROLE","0","30","1","0","MEMORY_2X","MONITOR1","2024-06-06 12:00:00.000 +0000 UTC","3","CLASSIC","MEDIUM","1","SUSPENDED","SNOWPARK-OPTIMIZED","2024-06-06 00:00:00.000 +0000 UTC","WAREHOUSE","10","WAREHOUSE","600","WAREHOUSE","300"`,
+			expectedOutput: `resource "snowflake_warehouse" "snowflake_generated_warehouse_WH_BASIC" {
```
sfc-gh-jmichalak, line 253: Nit: let's make sure that the first obj has only required fields (only name), and the second object has every possible field (if possible.

pkg/scripts/migration_script/mappings_warehouses_test.go
```diff
@@ -0,0 +1,118 @@
...
+	}{
+		{
+			name: "warehouse without parameters",
+			inputRows: [][]string{
+				{"auto_resume", "auto_suspend", "available", "comment", "created_on", "enable_query_acceleration", "generation", "is_current", "is_default", "max_cluster_count", "min_cluster_count", "name", "other", "owner", "owner_role_type", "provisioning", "query_acceleration_max_scale_factor", "queued", "quiescing", "resource_constraint", "resource_monitor", "resumed_on", "running", "scaling_policy", "size", "started_clusters", "state", "type", "updated_on"},
+				{"true", "600", "71.43", "", "2024-06-06 00:00:00.000 +0000 UTC", "false", "", "false", "false", "3", "1", "WH1", "0", "ADMIN", "ROLE", "0", "0", "1", "0", "", "", "2024-06-06 12:00:00.000 +0000 UTC", "5", "ECONOMY", "XSMALL", "2", "AVAILABLE", "STANDARD", "2024-06-06 00:00:00.000 +0000 UTC"},
+			},
+			expectedOutput: `
+resource "snowflake_warehouse" "snowflake_generated_warehouse_WH1" {
+  name = "WH1"
```
sfc-gh-jmichalak: Nit: similarly to program test, we should have only required fields in these two cases, the tests are a bit more readable this way.

pkg/scripts/migration_script/mappings_warehouses_test.go
```diff
@@ -0,0 +1,118 @@
...
+				{"true", "450", "80.00", "Test warehouse", "2024-06-06 00:00:00.000 +0000 UTC", "true", "1", "false", "false", "4", "1", "TEST_WH", "0", "ADMIN", "ROLE", "0", "30", "1", "0", "MEMORY_2X", "MONITOR2", "2024-06-06 12:00:00.000 +0000 UTC", "3", "CLASSIC", "MEDIUM", "1", "SUSPENDED", "SNOWPARK-OPTIMIZED", "2024-06-06 00:00:00.000 +0000 UTC", "ACCOUNT", "5", "ACCOUNT", "900", "ACCOUNT", "120"},
+			},
+			expectedOutput: `
+resource "snowflake_warehouse" "snowflake_generated_warehouse_TEST_WH" {
+  name = "TEST_WH"
+  auto_suspend = 450
+  comment = "Test warehouse"
+  enable_query_acceleration = "true"
+  max_cluster_count = 4
```
sfc-gh-jmichalak: Similar values are missing, like in the program tests.

pkg/scripts/migration_script/mappings_warehouses_test.go
```diff
@@ -0,0 +1,118 @@
...
+  auto_suspend = 300
+  comment = "Production warehouse"
+  enable_query_acceleration = "true"
+  max_cluster_count = 5
+  max_concurrency_level = 10
+  min_cluster_count = 2
+  query_acceleration_max_scale_factor = 50
+  resource_monitor = "MONITOR1"
+  statement_queued_timeout_in_seconds = 600
+  statement_timeout_in_seconds = 300
```
sfc-gh-jmichalak: A few are still missing (like warehouse_type).

pkg/scripts/migration_script/converter_warehouses.go
```diff
@@ -0,0 +1,119 @@
...
+			AutoResume:                      row.AutoResume == "true",
+			MinClusterCount:                 minClusterCount,
+			MaxClusterCount:                 maxClusterCount,
+			QueryAccelerationMaxScaleFactor: queryAccelerationMaxScaleFactor,
+			ResourceMonitor:                 sdk.NewAccountObjectIdentifier(row.ResourceMonitor),
+		},
+	}
+
+	if row.Generation != "" {
+		generation := sdk.WarehouseGeneration(row.Generation)
```
sfc-gh-jmichalak: Why not `ToWarehouseGeneration`​ with error handling? Similarly below.

pkg/scripts/migration_script/program_test.go
sfc-gh-jmichalak: Note: The solution now also has fields that have defaults in Snowflake.

pkg/scripts/migration_script/mappings_warehouses.go
sfc-gh-jmichalak: I meant that this should be discussed before the implementation.

pkg/scripts/migration_script/program_test.go
```diff
@@ -64,6 +64,12 @@ object_type represents the type of Snowflake object you want to generate terrafo
 			Warning: currently secondary databases and shared databases are treated as plain databases.
 			Supported resources:
 				- snowflake_database
+		- "warehouses" which expects a converted CSV output from the snowflake_warehouses data source
+			To support object parameters, one should use the SHOW PARAMETERS output, and combine it with the SHOW WAREHOUSES output, so the CSV header looks like "comment","created_on",...,"max_cluster_count","min_cluster_count","name","other",...
```
sfc-gh-jmichalak, line 68: In the docs (here and in the readme), we should mention the default values behavior. E.g. "The script always outputs fields that have non-empty default values in Snowflake." Let's also note that not fields are supported (e.g. initially_suspended) because you can't read the value from SF.

pkg/scripts/migration_script/converter_warehouses.go
sfc-gh-jcieslak: As we discussed, it would be a good idea to add it. Please address it in one of the follow-ups

pkg/scripts/migration_script/mappings_warehouses.go
sfc-gh-jcieslak: I don't think I fully get it. Required fields should always be present; optional fields are included depending on the value retrieved from Snowflake. In this case, you mean that optionals with default values on our side should always be filled? Please explain

pkg/scripts/migration_script/program_test.go
```diff
@@ -232,6 +238,72 @@ import {
...
+"false","1200","80.00","Production warehouse with all fields","2024-06-06 00:00:00.000 +0000 UTC","true","","false","false","4","2","WH_COMPLETE","0","ADMIN","ROLE","0","16","1","0","MEMORY_16X","MONITOR1","2024-06-06 12:00:00.000 +0000 UTC","3","ECONOMY","MEDIUM","1","SUSPENDED","SNOWPARK-OPTIMIZED","2024-06-06 00:00:00.000 +0000 UTC","WAREHOUSE","8","WAREHOUSE","300","WAREHOUSE","86400"
+"true","600","100","Gen2 warehouse","2024-06-06 00:00:00.000 +0000 UTC","false","2","false","false","1","1","WH_GEN2","0","ADMIN","ROLE","0","8","0","0","","","2024-06-06 12:00:00.000 +0000 UTC","0","STANDARD","LARGE","1","AVAILABLE","STANDARD","2024-06-06 00:00:00.000 +0000 UTC","","","","","",""`,
+			expectedOutput: `resource "snowflake_warehouse" "snowflake_generated_warehouse_WH_BASIC" {
+  name = "WH_BASIC"
+  auto_resume = "true"
+  auto_suspend = 600
+  max_cluster_count = 1
+  min_cluster_count = 1
+  query_acceleration_max_scale_factor = 8
+  scaling_policy = "STANDARD"
```
sfc-gh-jmichalak, line 260: No generation/resource constraint?

pkg/scripts/migration_script/mappings_warehouses_test.go
```diff
@@ -0,0 +1,134 @@
...
+  warehouse_type = "STANDARD"
+}
+import {
+  to = snowflake_warehouse.snowflake_generated_warehouse_WAREHOUSE
+  id = "\"WAREHOUSE\""
+}
+`,
+		},
+		{
+			name: "warehouse without parameters",
```
sfc-gh-jmichalak: Nit:

```suggestion
			name: "warehouse with all parameters set on higher level",
```

pkg/scripts/migration_script/mappings_warehouses.go
sfc-gh-jmichalak: How can you know if a value has been set by a user or set by Snowflake (as a default)? E.g., let's assume there's a warehouse with auto_suspend=600 (SF default). You can either:  
1\. Hardcode the SF defaults in the script and omit this value because you know it's the default, or  
2\. Do not hardcode the value. And we come back to my question.

For the resource implementation, we have decided to go with (2), and not (1), which this change originally presented, so that's why I expressed my doubts.

pkg/scripts/migration_script/converter_warehouses.go
sfc-gh-mkowalski: Will address it in the PR with roles

pkg/scripts/migration_script/program_test.go
sfc-gh-mkowalski: This test is meant to have neither; there are 2 other tests with generation and resource constraint respectively.

sfc-gh-jcieslak (approved):

pkg/scripts/migration_script/program_test.go
sfc-gh-jmichalak: We also include that such (optional non-zero default) fields can be removed from the final generated config. Also, the note about initially_suspended is missing.

pkg/scripts/migration_script/program_test.go
sfc-gh-jmichalak: Old warehouses can have both of these empty, so we can leave it.

PR #4250, opened 2025-11-30
chore: Fix SDK generation when nested definition is reused

sfc-gh-asawicki:
Generate SDK correctly when the nested definitions are reused.

The problem lies with the `Field` implementation; the `[]*Field` slice reuses fields in some cases, causing the setting parent to be incorrect and resulting in the generation of incorrect paths (which were previously altered manually).

For now, the implementation has been changed:
- `[]Field` instead of a slice of pointers
- reused nested definitions are wrapped in a function call
- `panic` was added for cases when the parent is being overriden

Later, when we split the definition from the model (as in the rest of the generators), the field structure should be constructed each time, allowing for the reuse of nested definitions.

### DIFF
--- a/pkg/sdk/generator/README.md
+++ b/pkg/sdk/generator/README.md
@@ -202,5 +202,3 @@ find a better solution to solve the issue (add more logic to the templates ?)
 
 ##### Known issues
 - generating two converts when Show and Desc use the same data structure
-- cannot re-generate when client.go is using generated interface
-- spaces in templates (especially nested validations)


pkg/sdk/generator/README.md
```diff
@@ -91,6 +91,9 @@ make generate-sdk-examples SF_TF_GENERATOR_ARGS='--filter-generation-part-names=
 make generate-sdk-examples SF_TF_GENERATOR_ARGS='--help'
 ```
 
+##### Known issues
```
sfc-gh-jmichalak, line 94: Nit: Why is this section separated from the other Known issues below?

sfc-gh-asawicki: The below one is part of the section with disclaimer: `⚠️ Disclaimer: This section may contain the deprecated essentials/improvements/known issues. It will be cleaned up shortly.`

PR #4254, opened 2025-12-02
feat: Migration script roles support

sfc-gh-mkowalski:
Added support for `account_roles` and `database_roles` to the migration script

Moved `ParameterType` handling to the helper function

### DIFF
--- a/pkg/sdk/type_helpers.go
+++ b/pkg/sdk/type_helpers.go
@@ -1,10 +1,7 @@
 package sdk
 
 import (
-	"fmt"
-	"slices"
 	"strconv"
-	"strings"
 )
 
 // String returns a pointer to the given string.
@@ -24,27 +21,6 @@ func ToBool(s string) bool {
 		panic(err)
 	}
 	return b
-}
-
-var AllParameterTypes = []ParameterType{
-	ParameterTypeAccount,
-	ParameterTypeUser,
-	ParameterTypeSession,
-	ParameterTypeObject,
-	ParameterTypeWarehouse,
-	ParameterTypeDatabase,
-	ParameterTypeSchema,
-	ParameterTypeTask,
-	ParameterTypeFunction,
-	ParameterTypeProcedure,
-	ParameterTypeSnowflakeDefault,
-}
-
-func ToParameterType(s string) (ParameterType, error) {
-	if slices.Contains(AllParameterTypes, ParameterType(strings.ToUpper(s))) {
-		return ParameterType(s), nil
-	}
-	return "", fmt.Errorf("invalid parameter type: %s", s)
 }
 
 // Int returns a pointer to the given int.
--- a/pkg/sdk/parameters.go
+++ b/pkg/sdk/parameters.go
@@ -2007,6 +2007,27 @@ const (
 	ParameterTypeProcedure        ParameterType = "PROCEDURE"
 )
 
+var AllParameterTypes = []ParameterType{
+	ParameterTypeAccount,
+	ParameterTypeUser,
+	ParameterTypeSession,
+	ParameterTypeObject,
+	ParameterTypeWarehouse,
+	ParameterTypeDatabase,
+	ParameterTypeSchema,
+	ParameterTypeTask,
+	ParameterTypeFunction,
+	ParameterTypeProcedure,
+	ParameterTypeSnowflakeDefault,
+}
+
+func ToParameterType(s string) (ParameterType, error) {
+	if slices.Contains(AllParameterTypes, ParameterType(strings.ToUpper(s))) {
+		return ParameterType(s), nil
+	}
+	return "", fmt.Errorf("invalid parameter type: %s", s)
+}
+
 type Parameter struct {
 	Key         string
 	Value       string


pkg/scripts/migration_script/README.md
```diff
@@ -97,6 +97,19 @@ where script options are:
 
     Supported resources:
       - snowflake_warehouse
+
+  - `account_roles` which expects a converted CSV output from the snowflake_account_roles data source
+      For more details about using multiple sources, visit the [Multiple sources section](#multiple-sources).
```
sfc-gh-jmichalak: I think these resources don't need multiple sources (data sources use only SHOW). We can just link the SHOW docs, like in the grant docs.

pkg/scripts/migration_script/common.go
```diff
@@ -58,10 +70,16 @@ func (h *parameterHandler) handleBooleanParameter(level sdk.ParameterType, value
 	return nil
 }
 
-func (h *parameterHandler) handleStringParameter(level sdk.ParameterType, value string, setField **string) error {
-	if h.level != level {
+func (h *parameterHandler) handleStringParameter(level string, value string, setField **string) error {
```
sfc-gh-jmichalak, line 63: Nit: now that we have more common logic, I think we can make one function generic, and from the handleXParameter wrappers just call the generic converter with the given conversion function (strconv.ParseBool, etc)

pkg/scripts/migration_script/mappings_account_roles.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func HandleAccountRoles(config *Config, csvInput [][]string) (string, error) {
+	return HandleResources[AccountRoleCsvRow, AccountRoleRepresentation](config, csvInput, MapAccountRoleToModel)
+}
+
+func MapAccountRoleToModel(role AccountRoleRepresentation) (accconfig.ResourceModel, *ImportModel, error) {
+	roleId := sdk.NewAccountObjectIdentifier(role.Name)
+	resourceId := NormalizeResourceId(fmt.Sprintf("account_role_%s", roleId.FullyQualifiedName()))
```
sfc-gh-jmichalak: Nit: we can use `resources.AccountRole`​ and trim `snowflake_`​, in other resources as well.

pkg/sdk/type_helpers.go
```diff
@@ -23,6 +25,27 @@ func ToBool(s string) bool {
...
+	ParameterTypeWarehouse,
+	ParameterTypeDatabase,
+	ParameterTypeSchema,
+	ParameterTypeTask,
+	ParameterTypeFunction,
+	ParameterTypeProcedure,
+	ParameterTypeSnowflakeDefault,
+}
+
+func ToParameterType(s string) (ParameterType, error) {
```
sfc-gh-jmichalak: Nit: we should call upper case before comparing and write unit tests for this (we do it for other enums like this).

pkg/scripts/migration_script/converter_account_roles.go
```diff
@@ -0,0 +1,36 @@
...
+type AccountRoleRepresentation struct {
+	sdk.Role
+}
+
+func (row AccountRoleCsvRow) convert() (*AccountRoleRepresentation, error) {
+	accountRoleRepresentation := &AccountRoleRepresentation{
+		Role: sdk.Role{
+			Name:    row.Name,
+			Comment: row.Comment,
+			Owner:   row.Owner,
```
sfc-gh-jmichalak: Other fields are missing.

pkg/scripts/migration_script/converter_database_roles.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type DatabaseRoleRepresentation struct {
+	sdk.DatabaseRole
+}
+
+func (row DatabaseRoleCsvRow) convert() (*DatabaseRoleRepresentation, error) {
+	databaseRoleRepresentation := &DatabaseRoleRepresentation{
+		DatabaseRole: sdk.DatabaseRole{
+			Name:         row.Name,
```
sfc-gh-jmichalak: Other fields are missing.

pkg/scripts/migration_script/README.md
sfc-gh-jmichalak: Um, there still is `converted CSV output from the snowflake_account_roles data source`​ technically, this is still true, but we should state `The expected input is in the form of SHOW DATABASES output`​ .​

pkg/scripts/migration_script/mappings_account_roles.go
sfc-gh-jmichalak: Nit: this can still be wrapped in a new function, so here we do sth like `	resourceId := ResourceId(resources.AccountRole, roleId.FullyQualifiedName())`​ and inside `ResourceId` we do the trimming etc.

pkg/sdk/type_helpers.go
```diff
@@ -37,6 +61,14 @@ func ToInt(s string) int {
 	return i
 }
 
+func ToIntWithDefault(s string, defaultValue int) int {
```
sfc-gh-jmichalak, line 40: I think it's not used anywhere?

sfc-gh-jmichalak (dismissed):
Resolve conflicts.

pkg/sdk/type_helpers.go
sfc-gh-mkowalski: It's used in the `convert` function in roles.

sfc-gh-jmichalak: Right, the search experience during reviews is not the best.

pkg/scripts/migration_script/mappings.go
```diff
@@ -18,6 +18,10 @@ func NormalizeResourceId(resourceId string) string {
 	return "snowflake_generated_" + string(resourceIdDisallowedCharacters.ReplaceAll([]byte(resourceId), []byte("")))
 }
 
+func ResourceId(resource string, id string) string {
```
sfc-gh-jcieslak, line 21: Why this needed to be introduced when we already have function above?

pkg/sdk/type_helpers.go
```diff
@@ -23,6 +26,27 @@ func ToBool(s string) bool {
 	return b
 }
 
+var AllParameterTypes = []ParameterType{
```
sfc-gh-jcieslak: this and function responsible for parsing should be next to the enum

pkg/scripts/migration_script/mappings.go
sfc-gh-mkowalski: It's meant to wrap the above function to prettify its syntax.

pkg/scripts/migration_script/mappings_account_roles.go
sfc-gh-jmichalak: Nit: we still could take ParameterType in ResourceId and not cast to string in each call.

PR #4255, opened 2025-12-02
chore: Enhance deletion detection in custom diffs

sfc-gh-jcieslak:
The test `TestAcc_SecretWithBasicAuthentication_BasicUseCase` was failing with the following error: `Error: Saved plan is stale`. From initial research, it appears that the error occurs whenever the plan saved at the end of one step doesn't match the plan generated by the refresh step of the next step (before applying).

~~Unfortunately, I cannot prove it, but it seems that generating a destroy plan triggers our custom diffs to recompute output collections in the resource (only marking as computed), which takes place during the next read (actually setting the values from Snowflake), just before applying the destroy step. This creates the error due to the difference in the plans.~~

~~Additionally, this doesn't prove why the test was flaky (only failing sometimes), but it appears to be a race condition between saving the plan and reading it in the next step. I'm planning to run the tests a few times to observe the results, but it seems that the fix actually made the test pass.~~

~~What I'm still not sure of is if we need all the steps in the newly introduced function to be sure we are in the destroy plan. I will conduct a set of tests to check which combination works. Also, maybe I'll be able to create a minimal reproduction step that would prove my theory (but could be impossible if above race condition happens within the testing framework / terraform).~~

It turns out the combination of Destroy + Check in the testing framework triggers an additional Refresh step, which seems to be causing the failure. Unfortunately, no logical proof was found, but after the adjustments, the tests were more stable and no longer failing. Due to changes only in tests, I think it's safe to assume the issue only applies to terraform testing framework, and not production Terraform CLI.

## changes
- Adjust and unskip test adjusted in #4286

### DIFF
--- a/pkg/testacc/resource_secret_with_basic_authentication_acceptance_test.go
+++ b/pkg/testacc/resource_secret_with_basic_authentication_acceptance_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/invokeactionassert"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/objectassert"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceassert"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert"
@@ -191,9 +190,6 @@ func TestAcc_SecretWithBasicAuthentication_BasicUseCase(t *testing.T) {
 			{
 				Destroy: true,
 				Config:  config.FromModels(t, basic),
-				Check: assertThat(t,
-					invokeactionassert.SecretDoesNotExist(t, id),
-				),
 			},
 			// Create - with optionals
 			{


pkg/resources/custom_diffs.go
```diff
@@ -132,6 +140,41 @@ func ComputedIfAnyAttributeChanged(resourceSchema map[string]*schema.Schema, key
...
+	// Check if all config values are null - this indicates a destroy operation.
+	rawConfigMap := rawConfig.AsValueMap()
+	allNull := true
+	for _, v := range rawConfigMap {
+		if !v.IsNull() {
+			allNull = false
+			break
+		}
+	}
+	if allNull && len(rawConfigMap) > 0 {
```
sfc-gh-jmichalak: Q: is it possible to have len(rawConfigMap) == 0? Perhaps the second condition is unnecessary?

pkg/resources/custom_diffs.go
```diff
@@ -132,6 +140,41 @@ func ComputedIfAnyAttributeChanged(resourceSchema map[string]*schema.Schema, key
...
+	// Check if the raw config is null/empty - this indicates a destroy operation
+	rawConfig := diff.GetRawConfig()
+	if rawConfig.IsNull() {
+		log.Printf("[DEBUG] isResourceBeingDestroyed: rawConfig is null, detected destroy")
+		return true
+	}
+
+	// During a destroy operation triggered by the test framework with Destroy: true,
+	// the config might still be present but all values would be null.
+	// Check if all config values are null - this indicates a destroy operation.
```
sfc-gh-jmichalak: Yikes, but I guess we're limited by the sdk itself. Are there any docs stating this?

sfc-gh-jcieslak: That may not be quite true, will look for more places that would prove this. Unfortunately, documentation is not that detailed...

pkg/resources/custom_diffs.go
sfc-gh-jcieslak: Tbh I don't know, this touches the internal implementation, also the GetRawConfig function is documented as experimental function, so it could be changing depending on the version (maybe?)

pkg/resources/custom_diffs.go
```diff
@@ -132,6 +140,41 @@ func ComputedIfAnyAttributeChanged(resourceSchema map[string]*schema.Schema, key
 	})
 }
 
+// isResourceBeingDestroyed detects if the resource is being destroyed by checking
+// if the raw config is null or if all config values are null.
+// During destroy, the raw config will be null for all attributes while the state still has values.
+func isResourceBeingDestroyed(diff *schema.ResourceDiff) bool {
```
sfc-gh-asawicki: question: isnt't there any method already giving us this information on the sdkv2 side (I guess not but I would rather ask)?
question2: did you check how other providers handle it?
questions3: how would it be solve in the terraform plugin framework? it would be enough to document this situation in the epic, so that we check it later

sfc-gh-jcieslak: ad.1. Didn't find one
ad.2. I didn't, as I wasn't even sure what to look for
ad.3. I will prepare a document summarizing my findings, I think this may be terraform testing framework specific, meaning, it doesn't require any changes on the resource-level, only the tests are affected (the real question is why only those 3 tests are affected, and despite searching pretty deeply, I couldn't find anything that would make those resources stand out in any way).

sfc-gh-asawicki: After all it's a testing framework issue.

PR #4259, opened 2025-12-03
chore: Add conditional assertions

sfc-gh-jcieslak:
## Changes
- Introduce a new env that would indicate Snowflake environment where the tests are run (currently: PROD and NON_PROD)
- Add a bit of abstraction over the env (use PROD as default)
- Adjust Docker Compose to use the appropriate NON_PROD Snowflake environment value
- Adjust failing tests due to environment differences
  - Added a new SnowflakeDefaults client in TestClient struct and used in tests

## Next pr
- Adjust environment variables to account locator based (as currently there are a few ways of recognizing in which Snowflake environment we are)


packaging/docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
     environment:
       TEST_SF_TF_GENERATED_RANDOM_VALUE: ${TEST_SF_TF_GENERATED_RANDOM_VALUE}
       TEST_SF_TF_TEST_OBJECT_SUFFIX: ${TEST_SF_TF_TEST_OBJECT_SUFFIX}
+      TEST_SF_TF_SNOWFLAKE_TESTING_ENVIRONMENT: NON_PROD
```
sfc-gh-jmichalak, line 9: Q: Why just not `PREPROD`​? What if preprod and qa are different as well?

pkg/acceptance/bettertestspoc/assert/objectassert/security_integration_snowflake_ext.go
```diff
@@ -0,0 +1,13 @@
+package objectassert
+
+import "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvs"
+
+func (s *SecurityIntegrationAssert) HasEnabledSnowflakeDefault() *SecurityIntegrationAssert {
```
sfc-gh-jmichalak: I don't think this logic should belong to the object assertions package. This package's functionality is to offer a set of simple assertions, and not resolving such values. We should do it in a separate package, and resolve the expected values in acceptance tests.

pkg/acceptance/testenvs/snowflake_environment.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func GetSnowflakeEnvironmentWithProdDefault() SnowflakeEnvironment {
+	env := os.Getenv(string(SnowflakeTestingEnvironment))
+	if env == "" {
+		return SnowflakeProdEnvironment
+	}
+	snowflakeEnvironment, err := parseSnowflakeEnvironment(env)
+	if err != nil {
+		return SnowflakeProdEnvironment
```
sfc-gh-jmichalak, line 40: Nit: log ther error here.

pkg/testacc/resource_oauth_integration_for_custom_clients_acceptance_test.go
```diff
@@ -51,12 +52,18 @@ func TestAcc_OauthIntegrationForCustomClients_BasicUseCase(t *testing.T) {
 		WithPreAuthorizedRoles(preAuthorizedRole.ID()).
 		WithComment(comment)
 
+	snowflakeEnv := testenvs.GetSnowflakeEnvironmentWithProdDefault()
+	enabledSnowflakeDefault := resources.BooleanFalse
+	if snowflakeEnv == testenvs.SnowflakeNonProdEnvironment {
```
sfc-gh-jmichalak: How about moving the values to a separate package? This way we would separate nicely the env differences into one place.

pkg/acceptance/bettertestspoc/assert/objectassert/security_integration_snowflake_ext.go
sfc-gh-asawicki: +1

pkg/acceptance/testenvs/testing_environment_variables.go
```diff
@@ -40,6 +40,7 @@ const (
 	RequireTestObjectsSuffix    env = "TEST_SF_TF_REQUIRE_TEST_OBJECT_SUFFIX"
 	RequireGeneratedRandomValue env = "TEST_SF_TF_REQUIRE_GENERATED_RANDOM_VALUE"
 	GeneratedRandomValue        env = "TEST_SF_TF_GENERATED_RANDOM_VALUE"
+	SnowflakeTestingEnvironment env = "TEST_SF_TF_SNOWFLAKE_TESTING_ENVIRONMENT"
```
sfc-gh-asawicki, line 43: Thought: We already have two options for knowing which environment we are in:
1. `GITHUB_ACTIONS` env, which is always set to true when inside the GH pipeline.
2. account locators
This is yet another abstraction, or 3rd source of truth.

I am generally okay if having this implemented this way as a quick workaround for failing tests but I think it would be great to introduce another solution which will be less error-prone later.

WDYT @sfc-gh-jcieslak @sfc-gh-jmichalak ?

sfc-gh-jcieslak: ad.1. `GITHUB_ACTIONS` can be also set locally, but we may also add other environments (?) requiring more distinction. 
ad.2. As we already have account locator managed env that steers if a given test should be run or not ( `TEST_SF_TF_NON_PROD_MODIFIABLE_ACCOUNT_LOCATOR`) that I use from time to time. Having a simple, known value like PROD or NON_PROD is less annoying than figuring out the account locator is, especially if you are jumping between Snowflake environments. But not sure if you were thinking about a similar abstraction.

sfc-gh-asawicki: ad 1 We don't have more envs than prod and preprod rn, so this is the only distinction we need now.
ad 2 This targets mainly the CI use case; for local, we could have a default meaning basically if empty, then default (prod in our case); for CIs you would always set these values, so there is no annoying; if you want to switch env locally, then it's just a matter of getting this env right in a small number of cases - I wouldn't call it annoying. For sure, it's safer then setting NO_PROD but using the prod credentials.

sfc-gh-jcieslak: As discussed offline the alternative approaches, I would leave it as it is and provide more automatic account locator based during one of the FF (I'll add the implementation details in the FF doc)

packaging/docker-compose.yml
sfc-gh-jcieslak: answered offline

pkg/acceptance/bettertestspoc/assert/objectassert/security_integration_snowflake_ext.go
sfc-gh-jcieslak: Taken out to separate package and used within tests where "normal" assertions are called. For simplicity, I used standalone functions, but I was also thinking about something similar to test client to have the per-object separation as struct fields to call `snowflakeDefaults().StreamOnDirectoryTable.StageOutputFormatting(stage.ID())` instead of `snowflakedefaults.StageIdentifierOutputFormatForStreamOnDirectoryTable(stage.ID())`. I guess for now, the standalone functions are fine, but maybe it would be good to switch (e.g. new pr), wdyt?

pkg/acceptance/testenvs/snowflake_environment.go
sfc-gh-jcieslak: added logs

pkg/testacc/resource_oauth_integration_for_custom_clients_acceptance_test.go
sfc-gh-jcieslak: moved to a separate package

pkg/acceptance/bettertestspoc/assert/objectassert/security_integration_snowflake_ext.go
sfc-gh-asawicki: I guess for now it's enough with that many defaults. We may soon face the problem with another env (gov), so let's not change it for now.

pkg/acceptance/snowflakedefaults/snowflake_environment.go
```diff
@@ -0,0 +1,45 @@
...
+
+func parseSnowflakeEnvironment(environment string) (SnowflakeEnvironment, error) {
+	snowflakeEnvironment := SnowflakeEnvironment(strings.ToUpper(environment))
+	if slices.Contains(allSnowflakeEnvironments, snowflakeEnvironment) {
+		return snowflakeEnvironment, nil
+	}
+	return "", fmt.Errorf("invalid Snowflake environment: %s, valid values are: %v", environment, allSnowflakeEnvironments)
+}
+
+func getSnowflakeEnvironmentWithProdDefault() SnowflakeEnvironment {
```
sfc-gh-asawicki: During our discussion about how, I suggested it could be executed just once when the acceptance/integration tests are initiated (the same way we check role existence, parameter setting, etc before running any test). For a single test run, it shouldnt change, so it would be good to either do it once here, or move this logic to test init and just keep the current environment there/in the test context/in the test client context - the last one allowing to reuse it for both acc and int tests).

pkg/acceptance/snowflakedefaults/snowflake_environment_test.go
```diff
@@ -0,0 +1,66 @@
...
+			"",
+			"dev",
+			"prod-env",
+			"non prod",
+		}
+
+		for _, input := range testCases {
+			input := input
+			t.Run(input, func(t *testing.T) {
+				t.Parallel()
```
sfc-gh-asawicki: We removed all t.Parallel earlier, it would be a first one. ~Also, this function is not safe for parallel tests as it depends on the environment state.~ - not true, it's only the parsing.

sfc-gh-asawicki (changes requested):

pkg/acceptance/snowflakedefaults/snowflake_environment_test.go
sfc-gh-jcieslak: Removed

pkg/acceptance/snowflakedefaults/snowflake_environment.go
sfc-gh-jcieslak: Done, I ultimately resigned from the `snowflakedefaults` package and moved the rest to `testenvs`. I wanted to work on those envs tomorrow to make them account locator-based for easier local testing and a more consistent approach to recognizing in which environment the tests are running.

pkg/acceptance/bettertestspoc/assert/objectassert/security_integration_snowflake_ext.go
sfc-gh-jcieslak: If I'll be working on it tomorrow (see: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4259#discussion_r2606994724) I would take it into account.

pkg/acceptance/snowflakedefaults/snowflake_environment.go
sfc-gh-asawicki: I think, we can work on it later on but let's discuss in the morning.

pkg/acceptance/helpers/snowflake_defaults.go
```diff
@@ -0,0 +1,36 @@
...
+
+func NewSnowflakeDefaultsClient(context *TestClientContext) *SnowflakeDefaultsClient {
+	return &SnowflakeDefaultsClient{
+		context: context,
+	}
+}
+
+func (c *SnowflakeDefaultsClient) EnabledForSnowflakeOauthSecurityIntegration(t *testing.T) bool {
+	t.Helper()
+	if c.context.snowflakeEnvironment == testenvs.SnowflakeNonProdEnvironment {
```
sfc-gh-asawicki, line 24: nit: `return c.context.snowflakeEnvironment == testenvs.SnowflakeNonProdEnvironment` instead of if

pkg/testacc/resource_stream_on_directory_table_acceptance_test.go
```diff
@@ -36,26 +36,28 @@ func TestAcc_StreamOnDirectoryTable_BasicUseCase(t *testing.T) {
 	complete := model.StreamOnDirectoryTable("test", id.DatabaseName(), id.SchemaName(), id.Name(), stage.ID().FullyQualifiedName()).
 		WithComment(comment)
 
+	expectedStageId := testClient().SnowflakeDefaults.StageIdentifierOutputFormatForStreamOnDirectoryTable(t, stage.ID())
```
sfc-gh-jmichalak, line 39: Nit: it's not actually an ID, I'd say it's a `stageName` or `stageString`?

sfc-gh-jmichalak (approved):

pkg/testacc/resource_stream_on_directory_table_acceptance_test.go
sfc-gh-jcieslak: I mean, it's ID formatted as string, will add String suffix in the follow-up

pkg/acceptance/helpers/snowflake_defaults.go
sfc-gh-jcieslak: Will address in a follow-up

pkg/testacc/resource_stream_on_directory_table_acceptance_test.go
sfc-gh-jmichalak: My point was that on a few environments it's _not_ an ID, but only object name, so it may be confusing.

PR #4260, opened 2025-12-03
chore: Fix dev

sfc-gh-jmichalak:
- Remove a field that no longer exists in generators.
- Fix linter findings.
- I ran pre-push locally and it runs without errors.

### DIFF
--- a/pkg/scripts/migration_script/mappings.go
+++ b/pkg/scripts/migration_script/mappings.go
@@ -19,7 +19,7 @@ func NormalizeResourceId(resourceId string) string {
 }
 
 func ResourceId(resource string, id string) string {
-	return NormalizeResourceId(fmt.Sprintf("%s_%s", strings.TrimPrefix(string(resource), "snowflake_"), id))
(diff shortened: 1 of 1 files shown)


pkg/scripts/migration_script/mappings.go
```diff
@@ -19,7 +19,7 @@ func NormalizeResourceId(resourceId string) string {
 }
 
 func ResourceId(resource string, id string) string {
```
sfc-gh-mkowalski, line 21: This change is going to be overwritten by my PR with users:
```
func ResourceId(resource resources.Resource, id string) string {
	return NormalizeResourceId(fmt.Sprintf("%s_%s", strings.TrimPrefix(resource.String(), "snowflake_"), id))
}
```

sfc-gh-jmichalak: Then you can adjust it further in your PR, as we discussed in one of the PRs before. The purpose of this change is to fix the broken dev branch. We should do it quickly because it blocks other people. That's why this is just a linter fix, not a more comprehensive solution to this problem.

PR #4261, opened 2025-12-03
chore: Adjust authentication policy snowflake output handling

sfc-gh-jcieslak:
## Changes
- Adjust authentication output handling to account for new built-in authentication policy
  - Filter out SHOW entries with invalid identifiers
- Re-generate assertions
- Test with newly acceptance test (run only in non prod environment)
- Documented the change in Migration guide

### DIFF
--- a/pkg/sdk/authentication_policies_impl_gen.go
+++ b/pkg/sdk/authentication_policies_impl_gen.go
@@ -4,6 +4,8 @@ package sdk
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"slices"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
@@ -201,14 +203,23 @@ func (r showAuthenticationPolicyDBRow) convert() (*AuthenticationPolicy, error) 
 		Options: r.Options,
 		Comment: r.Comment,
 	}
-	if r.CreatedOn.Valid {
-		policy.CreatedOn = r.CreatedOn.Time
-	}
+	var errs []error
 	if r.DatabaseName.Valid {
 		policy.DatabaseName = r.DatabaseName.String
+	} else {
+		errs = append(errs, fmt.Errorf("Missing database name for authentication policy with name: %s", r.Name))
 	}
 	if r.SchemaName.Valid {
 		policy.SchemaName = r.SchemaName.String
+	} else {
+		errs = append(errs, fmt.Errorf("Missing schema name for authentication policy with name: %s", r.Name))
+	}
+	if len(errs) > 0 {
+		return nil, errors.Join(errs...)
+	}
+
+	if r.CreatedOn.Valid {
+		policy.CreatedOn = r.CreatedOn.Time
 	}
 	if r.Owner.Valid {
 		policy.Owner = r.Owner.String
@@ -216,6 +227,7 @@ func (r showAuthenticationPolicyDBRow) convert() (*AuthenticationPolicy, error) 
 	if r.OwnerRoleType.Valid {
 		policy.OwnerRoleType = r.OwnerRoleType.String
 	}
+
 	return policy, nil
 }
 


pkg/sdk/authentication_policies_gen.go
```diff
@@ -134,31 +135,34 @@ type ShowAuthenticationPolicyOptions struct {
...
+	OwnerRoleType *string
 	Options       string
 }
 
 func (v *AuthenticationPolicy) ID() SchemaObjectIdentifier {
-	return NewSchemaObjectIdentifier(v.DatabaseName, v.SchemaName, v.Name)
+	if v.DatabaseName == nil && v.SchemaName == nil {
+		return NewSchemaObjectIdentifier("", "", v.Name)
+	}
+	return NewSchemaObjectIdentifier(*v.DatabaseName, *v.SchemaName, v.Name)
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: \+1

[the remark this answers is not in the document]
sfc-gh-jcieslak: Hmm, not sure, I would say either both or none, wdyt?

docs/data-sources/authentication_policies.md
```diff
@@ -5,7 +5,7 @@ description: |-
...
 ---
 
-!> **Caution: Preview Feature** This feature is considered a preview feature in the provider, regardless of the state of the resource in Snowflake. We do not guarantee its stability. It will be reworked and marked as a stable feature in future releases. Breaking changes are expected, even without bumping the major version. To use this feature, add the relevant feature name to `preview_features_enabled` field in the [provider configuration](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs#schema). Please always refer to the [Getting Help](https://github.com/snowflakedb/terraform-provider-snowflake?tab=readme-ov-file#getting-help) section in our Github repo to best determine how to get help for your questions.
+!> **Warning** The data source will not run describe query (regardless of the `with_describe` setting) for authentication policies with missing database and schema names in the output (e.g., the `BUILT-IN` authentication policy that is Snowflake default set at the account level, only if no user-defined policy is set).
```
sfc-gh-asawicki: it's still preview, shouldn't be removed

pkg/sdk/authentication_policies_gen.go
```diff
@@ -134,31 +135,34 @@ type ShowAuthenticationPolicyOptions struct {
...
-	OwnerRoleType string
+	Owner         *string
+	OwnerRoleType *string
 	Options       string
 }
 
 func (v *AuthenticationPolicy) ID() SchemaObjectIdentifier {
-	return NewSchemaObjectIdentifier(v.DatabaseName, v.SchemaName, v.Name)
+	if v.DatabaseName == nil && v.SchemaName == nil {
+		return NewSchemaObjectIdentifier("", "", v.Name)
```
sfc-gh-asawicki: I don't think we are handling such identifiers correctly given the current fully qualified name implementation

pkg/sdk/authentication_policies_gen.go
```diff
@@ -134,31 +135,34 @@ type ShowAuthenticationPolicyOptions struct {
...
+	SchemaName    *string
 	Kind          string
-	Owner         string
-	OwnerRoleType string
+	Owner         *string
+	OwnerRoleType *string
 	Options       string
 }
 
 func (v *AuthenticationPolicy) ID() SchemaObjectIdentifier {
```
sfc-gh-asawicki, line 161: if this is modified manually, you should leave a comment

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func TestAcc_AuthenticationPolicies_handling_with_builtin_policy_set_on_current_account(t *testing.T) {
+	basicModel := datasourcemodel.AuthenticationPolicies("test").
+		WithOnAccount()
+	providerModel := providermodel.SnowflakeProvider().
+		WithProfile(testprofiles.Secondary).
+		WithPreviewFeaturesEnabled(string(previewfeatures.AuthenticationPoliciesDatasource))
+
+	policy := secondaryTestClient().AuthenticationPolicy.ShowOnCurrentAccount(t)
```
sfc-gh-asawicki, line 24: we shouldn't have such a test run on the prod account... It should be run on a dedicated non-prod account if yu are unsetting the account authentication policy

sfc-gh-asawicki (changes requested):

docs/data-sources/authentication_policies.md
sfc-gh-jcieslak: Hmm missed it, I changed the subcategory (because it was set to stable), but didn't see it was removed (I copied stable template instead of preview one). Added back the caution.

pkg/sdk/authentication_policies_gen.go
sfc-gh-jcieslak: True, added

pkg/sdk/authentication_policies_gen.go
sfc-gh-jcieslak: At the moment the ID function is only used in Describe for authentication policy data source and in tests. This kind of identifier cannot be used in describe (as far as I checked), because it's broken. I was thinking about going for the route of returning either AccountObjectIdentifier or SchemaObjectIdentifier, but that also breaks other usages and conditional logic is necessary in some places to be sure we are dealing with SchemaObjectIdentifier (as mentioned above mostly tests). That's why I went with this approach, but I'm also not quite sure how such identifiers should be treated in such cases. For now, this approach doesn't seem to be that bad given it will only be broken when dealing with Snowflake object that are created in hacky way and are broken by default / not usable in SQL operations (as far as I checked myself). I can see that in the FullyQualifiedName implementation we would return `"".""."name"` and I would say it's equally bad as `"name"`, so I'm not sure if we can do any better in this case.

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
sfc-gh-jcieslak: added skip

pkg/sdk/authentication_policies_gen.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Adjusted

pkg/sdk/authentication_policies_gen.go
sfc-gh-asawicki: And what if someone tries to attach it to something? e.g. to account or to user? I understand this is an existing Snowflake object (good or not), but we should ensure it will be useful if someone will make it appear in their tf configuration.

MIGRATION_GUIDE.md
```diff
@@ -52,6 +52,28 @@ No changes in configuration are required for existing tasks. You can optionally
...
+(specifically, the `null` values for columns like `database_name`, `schema_name`, `owner`, `owner_role_type`, and `created_on`),
+our internal output conversion was failing.
+
+This issue could cause resources or data sources related to authentication policies, such as snowflake_authentication_policies, to fail with errors similar to the following:
+
+```
+sql: Scan error on column index 0, name "created_on": unsupported Scan, storing driver.Value type <nil> into type *time.Time
+```
+
+The internal implementation for authentication policy has now been updated to skip built-in entity, which lacks the aforementioned field values.
```
sfc-gh-jmichalak: Not exactly. First, we adjust the sdk so the error is gone, and second we skip the desc for that policy.

pkg/schemas/authentication_policy_gen.go
```diff
@@ -51,14 +51,24 @@ var _ = ShowAuthenticationPolicySchema
 
 func AuthenticationPolicyToSchema(authenticationPolicy *sdk.AuthenticationPolicy) map[string]any {
 	authenticationPolicySchema := make(map[string]any)
-	authenticationPolicySchema["created_on"] = authenticationPolicy.CreatedOn.String()
+	if authenticationPolicy.CreatedOn != nil {
```
sfc-gh-jmichalak: I'm confused, wasn't this behavior supposed to be hidden in the SDK? I imagined that the whole point of skipping the BUILT-IN policy was precisely to avoid such changes.

sfc-gh-jmichalak (changes requested):

pkg/sdk/authentication_policies_gen.go
sfc-gh-asawicki: I still don't like these adjustments in the id/struct logic. I would consider changing the SHOW logic entirely in this case, so that the "built-in" row is not being deserialized at all (which is not aligned with the current show implementation with convertRows, but we could adjust it). I shared the details privately.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Adjusted

pkg/schemas/authentication_policy_gen.go
sfc-gh-jcieslak: Adjusted

pkg/sdk/authentication_policies_gen.go
sfc-gh-jcieslak: Adjusted according to the offline discussions

pkg/sdk/generator/defs/authentication_policies_def.go
```diff
@@ -124,14 +124,14 @@ var authenticationPoliciesDef = g.NewInterface(
 	ShowOperation(
 		"https://docs.snowflake.com/en/sql-reference/sql/show-authentication-policies",
 		g.DbStruct("showAuthenticationPolicyDBRow").
-			Time("created_on").
+			OptionalTime("created_on").
```
sfc-gh-asawicki, line 127: Are these changes still neded if we skip deserialization for BUILT-IN?

pkg/sdk/authentication_policies_gen.go
```diff
@@ -134,15 +135,15 @@ type ShowAuthenticationPolicyOptions struct {
...
-	CreatedOn     time.Time `db:"created_on"`
-	Name          string    `db:"name"`
-	Comment       string    `db:"comment"`
-	DatabaseName  string    `db:"database_name"`
-	SchemaName    string    `db:"schema_name"`
-	Kind          string    `db:"kind"`
-	Owner         string    `db:"owner"`
-	OwnerRoleType string    `db:"owner_role_type"`
-	Options       string    `db:"options"`
+	CreatedOn     sql.NullTime   `db:"created_on"`
```
sfc-gh-asawicki, line 138: as above

pkg/sdk/authentication_policies_impl_gen.go
```diff
@@ -191,17 +195,28 @@ func (r *ShowAuthenticationPolicyRequest) toOpts() *ShowAuthenticationPolicyOpti
...
+	policy := &AuthenticationPolicy{
+		Name:    r.Name,
+		Kind:    r.Kind,
+		Options: r.Options,
+		Comment: r.Comment,
+	}
+	if r.CreatedOn.Valid {
+		policy.CreatedOn = r.CreatedOn.Time
+	}
+	if r.DatabaseName.Valid {
```
sfc-gh-asawicki, line 207: as above.

pkg/sdk/authentication_policies_impl_gen.go
```diff
@@ -42,6 +43,9 @@ func (v *authenticationPolicies) Show(ctx context.Context, request *ShowAuthenti
 	if err != nil {
 		return nil, err
 	}
+	dbRows = slices.DeleteFunc(dbRows, func(row showAuthenticationPolicyDBRow) bool {
```
sfc-gh-asawicki, line 48: nit (other PR, not before release): integration test on non-prod (create new account, try to fetch built-in which is the default for new acc)

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
```diff
@@ -0,0 +1,64 @@
...
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders: ExternalProviderWithExactVersion("2.11.0"),
+				Config:            accconfig.FromModels(t, providerModel, basicModel),
+				ExpectError:       regexp.MustCompile("Error: sql: Scan error on column index 0, name \"created_on\""),
+			},
+			{
+				ProtoV6ProviderFactories: secondaryAccountProviderFactory,
```
sfc-gh-asawicki, line 47: nit: secondaryAccountProviderFactory is initiated with all the preview features enabled, so `WithPreviewFeaturesEnabled` is not necessary

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	basicModel := datasourcemodel.AuthenticationPolicies("test").
+		WithOnAccount()
+	providerModel := providermodel.SnowflakeProvider().
+		WithProfile(testprofiles.Secondary).
+		WithPreviewFeaturesEnabled(string(previewfeatures.AuthenticationPoliciesDatasource))
+
+	policy := secondaryTestClient().AuthenticationPolicy.ShowOnCurrentAccount(t)
+
+	// In case a predefined authentication policy is attached, unset it, and set back after the test
```
sfc-gh-asawicki: nit (to consider when adding integration test): as suggested for the integration test, we could here just create a new account instead

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
```diff
@@ -0,0 +1,64 @@
...
+	if policy != nil && policy.Name != "BUILT-IN" {
```
sfc-gh-asawicki: question: how does it work after adjusting the show behavior? We will always have a nil for BUILT-IN because we will filter it out, right?

MIGRATION_GUIDE.md
```diff
@@ -84,6 +84,28 @@ These options can be set in the provider configuration, TOML configuration file,
 
 No changes in configuration are required for existing setups. You can optionally update your configurations to use these new options if you need to connect through a proxy.
 
+### *(bugfix)* authentication policies data sources
+Snowflake recently introduced a new, default authentication policy at the account level, which is applied when no user-defined policy is set.
+
+Due to the structure of this built-in authentication policy
```
sfc-gh-asawicki: Let's remove this short paragraph. And in the impact section, let's replace with something like:
```
The built-in policy does not reside in a database and schema. As it can't be directly interacted with, we decided to treat this default transparently:
- It won't be visible as output for snowflake_authentication_policies; if you query for authentication policies (with either `on.account` or `on.user` filtering options)
and only the built-in policy exists (no user-defined policies), the data source will return empty `show_output` and `describe_output` lists. If you have user-defined authentication policies, they will continue to appear in the output as expected.
- It can't be imported into snowflake_authentication_policy.
```
and also:
`The internal implementation for authentication policy has now been updated to handle and skip built-in entities, which lacks the aforementioned field values.` -> let's remove `which lacks the aforementioned field values`

sfc-gh-asawicki (changes requested):

pkg/sdk/authentication_policies_impl_gen.go
sfc-gh-jcieslak: Noted

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
sfc-gh-jcieslak: noted

pkg/sdk/generator/defs/authentication_policies_def.go
sfc-gh-jcieslak: I would say this is the minimal effort needed to skip the built-in authentication policy in the SHOW output. To skip the deserialization, we would have to adjust how the rows are queried in authentication policy SHOW function. We would have to query raw rows, then filter invalid rows, and at the end deserialize the rows or directly map to final AuthenticationPolicy objects. I checked and there's no less invasive way like adjusting `db` tags to include "skip if null" functionality, so I went with adjusting only the row representation.

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
sfc-gh-jcieslak: Hmm, the logic in the if was wrong now that I look at it. Adjusted, and now it should work with the new behavior.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Adjusted

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
sfc-gh-jcieslak: Now it should work because: whenever error happens during Show it's going to stop the test on the assertion in test client, when the policy is returned but it's null we get an invalid (built-in) policy, otherwise we get user-defined one (will be adjusted with integration test to work on dedicated, newly created account).

pkg/testacc/data_source_authentication_policies_account_level_acceptance_test.go
sfc-gh-jcieslak: Right, then removing the provider model

pkg/sdk/authentication_policies_gen.go
sfc-gh-jcieslak: as above

pkg/sdk/authentication_policies_impl_gen.go
sfc-gh-jcieslak: as above

pkg/sdk/generator/defs/authentication_policies_def.go
sfc-gh-asawicki: Discussed internally, we are adding errors in covnert and keep the struct changes in the db struct.

PR #4262, opened 2025-12-03
feat: Migration script users support

sfc-gh-mkowalski:
Added support for users in the migration script

### DIFF
--- a/pkg/scripts/migration_script/README.md
+++ b/pkg/scripts/migration_script/README.md
@@ -618,10 +618,6 @@ data "snowflake_schemas" "test" {
   in {
     database = "DATABASE"
   }
-
-  # Enable describe and parameters output
-  with_describe   = true
-  with_parameters = true
 }
 
 locals {


pkg/scripts/migration_script/program.go
```diff
@@ -181,6 +183,16 @@ object_type represents the type of Snowflake object you want to generate terrafo
 		- "database_roles" which expects input in the form of [SHOW DATABASE ROLES](https://docs.snowflake.com/en/sql-reference/sql/show-database-roles) output. Can also be obtained as a converted CSV output from the snowflake_database_roles data source.
 			Supported resources:
 				- snowflake_database_role
+		- "users" which expects input in the form of [SHOW USERS](https://docs.snowflake.com/en/sql-reference/sql/show-users) output. Can also be obtained as a converted CSV output from the snowflake_users data source.
```
sfc-gh-jmichalak: Not entirely true, this should be the data source because we're using object parameters here, so let's suggest that. Let's not suggest that the SHOW USERS can be used to omit a subset of the fields (someone may misinterpret).

pkg/scripts/migration_script/README.md
```diff
@@ -643,6 +661,76 @@ go run github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/scripts/migrat
...
 
+#### Users example
+
+Similarly, you can import users with their parameters. The following example shows how to generate a CSV file for users:
+
+```terraform
+terraform {
+  required_providers {
+    snowflake = {
+      source = "Snowflake-Labs/snowflake"
```
sfc-gh-jmichalak:
```suggestion
      source = "snowflakedb/snowflake"
```

pkg/scripts/migration_script/README.md
```diff
@@ -643,6 +661,76 @@ go run github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/scripts/migrat
...
+
+  # Get all unique keys from the first user to create CSV header
+  users_csv_header = join(",", [for key in keys(local.users_flattened[0]) : "\"${key}\""])
+
+  # Convert each user object to CSV row (properly escape quotes for CSV format)
+  users_csv_rows = [
+    for user in local.users_flattened :
+      join(",", [
+        for key in keys(local.users_flattened[0]) :
+        "\"${replace(tostring(lookup(user, key, "")), "\"", "\"\"")}\""
```
sfc-gh-jmichalak: Q: What's the purpose of this section? We already have a similar example for schemas. If the only difference is the \\" replace, we should just update the section above.

pkg/scripts/migration_script/converter_users.go
```diff
@@ -0,0 +1,352 @@
...
+	MinsToBypassMfa                               string `csv:"mins_to_bypass_mfa"`
+	MinsToUnlock                                  string `csv:"mins_to_unlock"`
+	MustChangePassword                            string `csv:"must_change_password"`
+	Name                                          string `csv:"name"`
+	Owner                                         string `csv:"owner"`
+	SnowflakeLock                                 string `csv:"snowflake_lock"`
+	Type                                          string `csv:"type"`
+	AbortDetachedQueryLevel                       string `csv:"abort_detached_query_level"`
+	AbortDetachedQueryValue                       string `csv:"abort_detached_query_value"`
+	ActivePythonProfilerLevel                     string `csv:"active_python_profiler_level"`
```
sfc-gh-jmichalak, line 54: A few of these fields, like this one, are missing from the data source. This is a valid parameter present in SHOW PARAMETERS, but not in the data source schema. Let's discuss if such parameters should be here.

pkg/scripts/migration_script/mappings_users.go
```diff
@@ -0,0 +1,344 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+const (
+	UserTypePerson        = "PERSON"
+	UserTypeService       = "SERVICE"
+	UserTypeLegacyService = "LEGACY_SERVICE"
+)
+
+func HandleUsers(config *Config, csvInput [][]string) (string, error) {
```
sfc-gh-jmichalak, line 19: Why not just call HandleResources?

pkg/scripts/migration_script/mappings_users.go
```diff
@@ -0,0 +1,344 @@
...
+	handleIf(user.MustChangePassword, resourceModel.WithMustChangePassword)
+
+	// Secondary roles
+	secondaryRolesOption := user.GetSecondaryRolesOption()
+	if secondaryRolesOption != sdk.SecondaryRolesOptionDefault {
+		resourceModel.WithDefaultSecondaryRolesOption(string(secondaryRolesOption))
+	}
+
+	// Parameters
+	handleOptionalFieldWithBuilder(user.AbortDetachedQuery, resourceModel.WithAbortDetachedQuery)
```
sfc-gh-jmichalak, line 250: We could extract it to a separate function. resourceModel is of different types for each user type, but we could introduce a simple interface.

pkg/scripts/migration_script/mappings_users_test.go
```diff
@@ -0,0 +1,376 @@
...
+  name = "SVC_USER"
+}
+import {
+  to = snowflake_service_user.snowflake_generated_service_user_SVC_USER
+  id = "\"SVC_USER\""
+}
+`,
+		},
+		{
+			name: "complete service user with all basic attributes",
```
sfc-gh-jmichalak, line 206: Let's add at least one test case covering all parameters.

pkg/scripts/migration_script/program_test.go
```diff
@@ -364,6 +374,252 @@ import {
...
...econds_level","statement_timeout_in_seconds_value","strict_json_output_level","strict_json_output_value","time_input_format_level","time_input_format_value","time_output_format_level","time_output_format_value","timestamp_day_is_always24h_level","timestamp_day_is_always24h_value","timestamp_input_format_level","timestamp_input_format_value","timestamp_l_t_z_output_format_level","timestamp_l_t_z_output_format_value","timestamp_n_t_z_output_format_level","timestamp_n_t_z_output_format_value","timestamp_output_format_level","timestamp_output_format_value","timestamp_t_z_output_format_level","timestamp_t_z_output_format_value","timestamp_type_mapping_level","timestamp_type_mapping_value","timezone_level","timezone_value","trace_level_level","trace_level_value","transaction_abort_on_error_level","transaction_abort_on_error_value","transaction_default_isolation_level_level","transaction_default_isolation_level_value","two_digit_century_start_level","two_digit_century_start_value","unsupported_d_d_l_action_level","unsupported_d_d_l_action_value","use_cached_result_level","use_cached_result_value","week_of_year_policy_level","week_of_year_policy_value","week_start_level","week_start_value"
```
sfc-gh-jmichalak: middle_name is missing

pkg/scripts/migration_script/program_test.go
```diff
@@ -78,6 +78,16 @@ object_type represents the type of Snowflake object you want to generate terrafo
 		- "database_roles" which expects input in the form of [SHOW DATABASE ROLES](https://docs.snowflake.com/en/sql-reference/sql/show-database-roles) output. Can also be obtained as a converted CSV output from the snowflake_database_roles data source.
 			Supported resources:
 				- snowflake_database_role
+		- "users" which expects input in the form of [SHOW USERS](https://docs.snowflake.com/en/sql-reference/sql/show-users) output. Can also be obtained as a converted CSV output from the snowflake_users data source.
```
sfc-gh-jmichalak: Let's also note in the program and in the README that the password is not supported.

pkg/scripts/migration_script/program_test.go
sfc-gh-jmichalak: And public_key fields - they're available only in DESC, so we need to support describe_output as well.

pkg/scripts/migration_script/mappings_users.go
sfc-gh-mkowalski: Unfortunately, the interface won't work due to mismatched return types.

pkg/scripts/migration_script/README.md
```diff
@@ -598,17 +616,23 @@ data "snowflake_schemas" "test" {
   in {
     database = "DATABASE"
   }
+
+  # Enable describe and parameters output
+  with_describe   = true
+  with_parameters = true
```
sfc-gh-jmichalak: They are enabled by default, so there's no need for this.

pkg/scripts/migration_script/README.md
```diff
@@ -620,10 +644,13 @@ locals {
...
   csv_header = join(",", [for key in keys(local.schemas_flattened[0]) : "\"${key}\""])
 
-  # Convert each schema object to CSV row
+  # Convert each schema object to CSV row (properly escape quotes and remove newlines for CSV format)
   csv_rows = [
     for schema in local.schemas_flattened :
-      join(",", [for key in keys(local.schemas_flattened[0]) : "\"${lookup(schema, key, "")}\""])
+      join(",", [
+        for key in keys(local.schemas_flattened[0]) :
+        "\"${replace(replace(tostring(lookup(schema, key, "")), "\n", " "), "\"", "\"\"")}\""
```
sfc-gh-jmichalak: Nit: This replaces with space. Can we escape it properly, like with `"`? Or at least document it as a limitation.

pkg/scripts/migration_script/converter_users.go
sfc-gh-asawicki: So what's the outcome here? I think that if we suggest using the data source to get the csv data and the data source does not have this field then mby it's better to remove it from the csv/

sfc-gh-asawicki (dismissed):

pkg/scripts/migration_script/converter_users.go
sfc-gh-mkowalski: We discussed it, and concluded that it won't hurt to leave them and they might be needed in the future.

pkg/scripts/migration_script/common_test.go
```diff
@@ -0,0 +1,68 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCsvUnescape(t *testing.T) {
```
sfc-gh-jmichalak, line 9: It would be also great to have tests that prove hcl escape + unescape do not change the data (pls add it to the follow up epic).

pkg/scripts/migration_script/converter_users.go
```diff
@@ -0,0 +1,388 @@
...
+		User: sdk.User{
+			Name:                  row.Name,
+			LoginName:             row.LoginName,
+			DisplayName:           row.DisplayName,
+			FirstName:             row.FirstName,
+			LastName:              row.LastName,
+			Email:                 row.Email,
+			MinsToUnlock:          row.MinsToUnlock,
+			DaysToExpiry:          row.DaysToExpiry,
+			Comment:               csvUnescape(row.Comment),
```
sfc-gh-jmichalak, line 282: So, we escape public keys because they contain multiline values, but comment as well? What about other strings? Maybe let's state it as a limitation in the docs.

sfc-gh-jmichalak (approved):

sfc-gh-jmichalak (approved):

PR #4271, opened 2025-12-08
feat: Introduce manual acceptance tests to the migration script

sfc-gh-mkowalski:
Added manual acceptance tests setup guide alongside example test for `grants` object.


pkg/scripts/migration_script/manual_tests/users/datasource.tf
```diff
@@ -0,0 +1,87 @@
+# =============================================================================
+# Users Data Source - Fetch users from Snowflake and generate CSV
+# =============================================================================
+# This file fetches the test users created by user_objects_def.tf and generates
+# a CSV file that can be used to test the migration script.
+#
+# Usage:
+#   1. First create users: terraform apply (with user_objects_def.tf)
+#   2. Generate CSV: terraform apply (this will create users.csv)
```
sfc-gh-jmichalak: Shouldn't 1 and 2 be executed in one step?

pkg/scripts/migration_script/manual_tests/users/datasource.tf
```diff
@@ -0,0 +1,87 @@
...
+  users_csv_rows = length(local.users_flattened) > 0 ? [
+    for user in local.users_flattened :
+      join(",", [
+        for key in keys(local.users_flattened[0]) :
+        "\"${local.csv_escape[user.name][key]}\""
+      ])
+  ] : []
+
+  # Combine header and rows
+  users_csv_content = length(local.users_flattened) > 0 ? join("\n", concat([local.users_csv_header], local.users_csv_rows)) : "# No users found matching MIGRATION_TEST_%"
```
sfc-gh-jmichalak: This way, if there are no users, we output the "# No users..." string to the csv. We should either fail (preferred, as this is a test assertion) or not write such an output.

pkg/scripts/migration_script/manual_tests/users/objects_def.tf
```diff
@@ -0,0 +1,267 @@
...
+
+# 1. Basic PERSON user - minimal attributes
+resource "snowflake_user" "person_basic" {
+  name = "MIGRATION_TEST_PERSON_BASIC"
+}
+
+# 2. PERSON user with all basic attributes
+resource "snowflake_user" "person_complete" {
+  name                         = "MIGRATION_TEST_PERSON_COMPLETE"
+  login_name                   = "migration_person_login"
```
sfc-gh-jmichalak: This (login_name and public keys in other resources) should be randomized. Or get it from the input? In other places as well.

pkg/scripts/migration_script/manual_tests/users/objects_def.tf
```diff
@@ -0,0 +1,267 @@
...
+
+# ------------------------------------------------------------------------------
+# Edge Cases
+# ------------------------------------------------------------------------------
+
+# 14. User with very long comment (tests field handling)
+resource "snowflake_user" "long_comment" {
+  name       = "MIGRATION_TEST_LONG_COMMENT"
+  login_name = "migration_long_comment_login"
+  comment    = "This is a very long comment that tests how the migration script handles comments with many characters. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
```
sfc-gh-jmichalak: We should test the escaping here.

pkg/scripts/migration_script/manual_tests/users/expected_output.tf
```diff
@@ -0,0 +1,306 @@
+# =============================================================================
+# Expected Migration Script Output
+# =============================================================================
+# This file contains the EXPECTED output from running:
+#   go run .. -import=block users < users.csv
+#
+# Use this to compare with actual output:
+#   go run .. -import=block users < users.csv > actual_output.tf
+#   diff users_expected_output.tf actual_output.tf
```
sfc-gh-jmichalak: What about this comment? The first 20 lines will be visible in the `diff` output, right? If so, we should move them to readme.

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+MIGRATION_SCRIPT_DIR="$(dirname "$SCRIPT_DIR")"
+
+# Colors for output
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+CYAN='\033[0;36m'
+NC='\033[0m' # No Color
```
sfc-gh-jmichalak: For the future: please implement basic functionalities first, and after that iterate with new features, as this PR is unnecessarily big.

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+#
+# Usage:
+#   ./run_test.sh <object_type>                    # Run test (create, test, keep resources)
+#   ./run_test.sh <object_type> --destroy          # Only destroy existing resources
+#   ./run_test.sh <object_type> --skip-create      # Skip creation, just run migration test
+#   ./run_test.sh --list                           # List available object types
+#
+# Examples:
+#   ./run_test.sh users
+#   ./run_test.sh account_roles --skip-create
```
sfc-gh-jmichalak: Account roles are not yet supported.

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+    rm -f actual_output.tf
+    log_info "Removed previous actual_output.tf"
+fi
+
+# =============================================================================
+# DESTROY ONLY MODE
+# =============================================================================
+if [ "$DESTROY_ONLY" = true ]; then
+    log_step "Destroying all test resources for $OBJECT_TYPE..."
+    terraform destroy -auto-approve
```
sfc-gh-jmichalak: Nit: mention it in the helper about the auto-approve flag.

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+
+        # Build -target arguments
+        TARGET_ARGS=""
+        for resource in $RESOURCES; do
+            TARGET_ARGS="$TARGET_ARGS -target=$resource"
+        done
+
+        export SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true
+        terraform apply -auto-approve $TARGET_ARGS
+        log_info "Test objects created successfully!"
```
sfc-gh-jmichalak: Why not just `terraform apply`? And add depends_on in the data source.

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+# =============================================================================
+if [ ! -d ".terraform" ]; then
+    log_step "Step 0: Initializing Terraform..."
+    terraform init
+fi
+
+# =============================================================================
+# STEP 1: Create test objects on Snowflake
+# =============================================================================
+if [ "$SKIP_CREATE" = false ]; then
```
sfc-gh-jmichalak: What's the use case of this flag?

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+
+    # Get local_file resources
+    LOCAL_FILES=$(grep -E '^resource\s+"local_file"' datasource.tf | sed 's/resource "local_file" "\([^"]*\)".*/local_file.\1/' | tr '\n' ' ')
+    for lf in $LOCAL_FILES; do
+        DATASOURCE_TARGETS="$DATASOURCE_TARGETS -target=$lf"
+    done
+fi
+
+# Apply only the data source and CSV generation
+export SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true
```
sfc-gh-jmichalak: We shouldn't need this env.

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+# =============================================================================
+# STEP 2: Fetch objects via data source and generate CSV
+# =============================================================================
+log_step "Step 2: Fetching objects via data source and generating CSV..."
+log_info "This retrieves object data including DESCRIBE and PARAMETERS output (if applicable)"
+
+# Get data sources and local_file resources from datasource.tf only
+# This avoids re-applying objects_def.tf resources
+DATASOURCE_TARGETS=""
+if [ -f "datasource.tf" ]; then
```
sfc-gh-jmichalak: Can't we just do the apply, maybe with the depends_on in data source, or with two steps with a cond flag?

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+    exit 1
+fi
+
+if [ ! -f "actual_output.tf" ]; then
+    log_error "Actual output file not found: actual_output.tf"
+    exit 1
+fi
+
+# Normalize and compare (remove comments and empty lines)
+grep -v '^#' expected_output.tf | grep -v '^$' > /tmp/expected_norm.txt
```
sfc-gh-jmichalak: As I wrote before, I think we can do this without such normalization. Also, the file names should be unique (at least per resource type).

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+grep -v '^#' actual_output.tf | grep -v '^$' > /tmp/actual_norm.txt
+
+echo ""
+echo "Differences (if any):"
+echo "====================="
+
+DIFF_RESULT=0
```
sfc-gh-jmichalak: I think the assertions below are unnecessary, as we assert everything (even the order) in the expected_output.tf

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,306 @@
+# Migration Script Manual Tests
+
+This directory contains end-to-end tests for the migration script. Each object type has its own folder with test configuration.
+
+## Quick Start
+
+### Step 0: Set account information in environment variables:
+- SNOWFLAKE_ACCOUNT
```
sfc-gh-jmichalak: Why is this needed? Can't we fallback to the default profile?

pkg/scripts/migration_script/manual_tests/README.md
```diff
@@ -0,0 +1,306 @@
...
+- SNOWFLAKE_HOST
+- SNOWFLAKE_USER
+- SNOWFLAKE_PASSWORD
+- SNOWFLAKE_ROLE
+
+## Steps using script
+
+```bash
+# List available object types
+./run_test.sh --list
```
sfc-gh-jmichalak: We should be able to run (and cleanup) the whole suite with one command.

pkg/scripts/migration_script/manual_tests/README.md
```diff
@@ -0,0 +1,306 @@
...
+├── users/                   # Users test
+│   ├── objects_def.tf       # Creates test users on Snowflake
+│   ├── datasource.tf        # Fetches users, generates CSV
+│   ├── expected_output.tf   # Expected migration script output
+│   ├── objects.csv          # Generated CSV (after terraform apply)
+│   └── actual_output.tf     # Generated output (after test run)
+└── <new_object_type>/       # Add new object types here
+```
+
+## Test Workflow
```
sfc-gh-jmichalak: I think this section is just repeated and doesn't explain anything new.

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,306 @@
...
+}
+```
+
+### Step 5: Test
+
+```bash
+./run_test.sh warehouses
+```
+
+## CSV Format Notes
```
sfc-gh-jmichalak: Again, maybe we should move this section to the script readme.

pkg/scripts/migration_script/manual_tests/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Common Issues
+
+### "Object type not found"
+
+Make sure the folder exists and contains `objects_def.tf`, `datasource.tf`, and `expected_output.tf`.
+
+### "Object already exists" errors
+
+If objects already exist on Snowflake, use `--skip-create`:
```
sfc-gh-jmichalak: Why does this situation happen? Is it because someone may forget destroying the resources? In such case, they can import to TF.

pkg/scripts/migration_script/manual_tests/README.md
```diff
@@ -0,0 +1,306 @@
...
+
+```bash
+./run_test.sh users --skip-create
+```
+
+Or manually skip Step 3 and go directly to fetching.
+
+### Terraform state issues
+
+Each object type folder has its own Terraform state. If state gets corrupted:
```
sfc-gh-jmichalak: Why would it be corrupted? I don't think we should recommend this action, I'm just trying to understand the purpose here.

pkg/scripts/migration_script/manual_tests/README.md
```diff
@@ -0,0 +1,306 @@
...
+```bash
+cd <object_type>
+rm -rf .terraform terraform.tfstate*
+terraform init
+```
+
+### Differences in expected output
+
+Common reasons for differences:
+- **Uppercase**: Snowflake may uppercase some values (e.g., `login_name`)
```
sfc-gh-jmichalak: 1. I don't think this should be a problem for us, as the provider is case-sensitive and should handle the wrapping. 
2. I think I didn't see the unicode escpaing in the limitations. We should add it to the readme.

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,306 @@
+# Migration Script Manual Tests
+
+This directory contains end-to-end tests for the migration script. Each object type has its own folder with test configuration.
```
sfc-gh-asawicki, line 3: we already have a dir for manual tests, I would rather see it there in a migration_script subfolder

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
+#!/bin/bash
+# =============================================================================
+# Migration Script Manual Test Runner
```
sfc-gh-asawicki: The point of having manual tests was not to develop a script runner; ultimately, we will simply include these tests in our acceptance test suite. The goal was to have a short readme, a bunch of tf files, and to run them manually, given the description (like with the other manual tests we have).
This runner is untested, written in bash, inconsistent with other manual tests, would be ultimately removed when switching to acceptance tests, provides a way to modify customers' infrastructure, and is not what we discussed on Thursday.
Let's discuss how we proceed with it. From my perspective, we should remove it.

pkg/scripts/migration_script/manual_tests/users/expected_output.tf
```diff
@@ -0,0 +1,306 @@
+# =============================================================================
+# Expected Migration Script Output
```
sfc-gh-asawicki: The output shouldn't be the last thing validated in these tests; the last step should be applying the  generated resources/imports and simply verifying that there are no changes in plan.

pkg/scripts/migration_script/manual_tests/run_test.sh
```diff
@@ -0,0 +1,319 @@
...
+echo -e "${YELLOW}Resources were NOT destroyed.${NC}"
+echo "To cleanup, run: $0 $OBJECT_TYPE --destroy"
+echo ""
+
+if [ $DIFF_RESULT -eq 0 ]; then
+    log_step "Test completed successfully! ✓"
+else
+    log_step "Test completed with differences. Review the output above."
+    exit 1
+fi
```
sfc-gh-jmichalak: Also, we could have another step: actually run the Terraform apply on the generated config and verify if the plan is empty. Maybe we could even do this instead of diffing the referenced file. wdyt @sfc-gh-jcieslak @sfc-gh-asawicki ?

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/run_test.sh
sfc-gh-mkowalski: File was removed.

pkg/scripts/migration_script/manual_tests/README.md
sfc-gh-mkowalski: This script no longer exists.

pkg/scripts/migration_script/manual_tests/README.md
sfc-gh-mkowalski: Runner script no longer exists.

pkg/scripts/migration_script/manual_tests/users/objects_def.tf
sfc-gh-mkowalski: Needed to replace current tests with grants so this will be addressed in the PR with all the other tests.
The reason for that was moving the tests into the `pkg/manual_tests` folder and using `go run github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/scripts/migration_script@main` instead of a local script, which currently only supports grants.

sfc-gh-jmichalak: Why, though? We could still test it as a local script, right?

pkg/manual_tests/migration_script/grants/objects_def.tf
```diff
@@ -0,0 +1,288 @@
...
+
+# ==============================================================================
+# GRANT DATABASE ROLE TESTS
+# ==============================================================================
+
+# ------------------------------------------------------------------------------
+# Grant Database Role to Database Role
+# ------------------------------------------------------------------------------
+resource "snowflake_grant_database_role" "dbrole_to_dbrole" {
+  database_role_name        = "\"${snowflake_database.test_db.name}\".\"${snowflake_database_role.child_db_role.name}\""
```
sfc-gh-jmichalak:
```suggestion
  database_role_name        = snowflake_database_role.child_db_role.fully_qualified_name
```

in other places as well.

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,200 @@
...
+  #   these are not privilege grants and the migration script can't handle them)
+  test_grants = [
+    for g in local.all_grants : g
+    if (can(regex("MIGRATION_TEST", g.grantee_name)) || can(regex("MIGRATION_TEST", g.name))) &&
+       g.granted_by != "" &&
+       g.privilege != ""
+  ]
+
+  # CSV header - matches the GrantCsvRow struct
+  grants_csv_header = "\"privilege\",\"granted_on\",\"grant_on\",\"name\",\"granted_to\",\"grant_to\",\"grantee_name\",\"grant_option\",\"granted_by\""
```
sfc-gh-jmichalak: This should be compute automatically, like we do it in the readme.

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,200 @@
...
+
+  # CSV escape function
+  csv_escape_grant = {
+    for idx, grant in local.test_grants :
+    idx => {
+      privilege   = replace(replace(replace(tostring(grant.privilege), "\\", "\\\\"), "\n", "\\n"), "\"", "\"\"")
```
sfc-gh-jmichalak: Why is each field listed? The previous solution just iterated on each field.

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,200 @@
...
+      grantee_name = replace(replace(replace(tostring(grant.grantee_name), "\\", "\\\\"), "\n", "\\n"), "\"", "\"\"")
+      grant_option = tostring(grant.grant_option)
+      granted_by  = replace(replace(replace(tostring(grant.granted_by), "\\", "\\\\"), "\n", "\\n"), "\"", "\"\"")
+    }
+  }
+
+  # Convert each grant to CSV row
+  grants_csv_rows = [
+    for idx, grant in local.test_grants :
+    "\"${local.csv_escape_grant[idx].privilege}\",\"${local.csv_escape_grant[idx].granted_on}\",\"${local.csv_escape_grant[idx].grant_on}\",\"${local.csv_escape_grant[idx].name}\",\"${local.csv_escape_grant[idx].granted_to}\",\"${local.csv_escape_grant[idx].grant_to}\",\"${local.csv_escape_grant[idx].grantee_name}\",\"${local.csv_escape_grant[idx].grant_option}\",\"${local.csv_escape_grant[idx].granted_by}\""
```
sfc-gh-jmichalak: Same here.

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,200 @@
...
+  ]
+
+  # Remove duplicates by converting to set and back
+  grants_csv_rows_unique = tolist(toset(local.grants_csv_rows))
```
sfc-gh-jmichalak, line 154: Why are we having duplicates here?

pkg/scripts/migration_script/manual_tests/users/datasource.tf
sfc-gh-jmichalak: It's not obvious, but make sure that when the grants are empty, the HCL assertion detects it. (e.g. use an invalid role as a filter in a separate test Also, we can add this assertion to the readme, it can be useful.

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+This creates objects from `objects_def.tf` and generates `objects.csv` via `datasource.tf`.
+
+### Step 5: Run migration script
+
+```bash
+go run github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/scripts/migration_script@main -import=block grants < objects.csv > generated_output.tf
+```
+
+### Step 6: Apply generated output (imports existing objects)
```
sfc-gh-jmichalak: No, we shouldn't apply it here because Terraform will read all the tf files, so the resources will be declared in two places. We should be able to assert that the plan for the generated output is empty, so we should:
- Import the resources in a new directory (so the state doesn't conflict),
- Run the plan and assert it (I think TF has a flag that causes non-empty plans to fail),
- In the original directory, destroy the previously created objects.

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,204 @@
...
+```bash
+terraform apply -auto-approve
+```
+
+This creates objects from `objects_def.tf` and generates `objects.csv` via `datasource.tf`.
+
+### Step 5: Run migration script
+
+```bash
+go run github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/scripts/migration_script@main -import=block grants < objects.csv > generated_output.tf
```
sfc-gh-jmichalak: Use `go run ../scripts/migration_script/` instead.

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+# Debug outputs
+output "objects_found" {
+  value = length(local.warehouses_flattened)
+}
+```
+
+### Step 4: Test it
+
+```bash
```
sfc-gh-jmichalak: Yes, that's what I wanted :) Let's put this on top (with removed second apply).

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,200 @@
...
+}
+
+# ==============================================================================
+# GRANTS TO DATABASE ROLES
+# ==============================================================================
+
+# Fetch all grants TO the privilege database role
+data "snowflake_grants" "to_priv_db_role" {
+  grants_to {
+    database_role = "\"${snowflake_database.test_db.name}\".\"${snowflake_database_role.priv_db_role.name}\""
```
sfc-gh-jmichalak: Use the fully_qualified_name field - in other places as well.

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,200 @@
...
+    [for g in data.snowflake_grants.of_child_role.grants : g],
+    [for g in data.snowflake_grants.to_priv_db_role.grants : g],
+    [for g in data.snowflake_grants.to_parent_db_role.grants : g],
+    [for g in data.snowflake_grants.of_child_db_role.grants : g],
+    [for g in data.snowflake_grants.of_priv_db_role.grants : g]
+  )
+
+  # Filter to only privilege grants:
+  # - Must contain MIGRATION_TEST in grantee_name or name
+  # - Must have a non-empty granted_by (implicit grants from Snowflake have empty granted_by
```
sfc-gh-jmichalak: Q: could you give some examples? Are you talking about ownership, or something different?

pkg/manual_tests/migration_script/README.md
sfc-gh-asawicki: or `@dev`

pkg/manual_tests/migration_script/README.md
sfc-gh-asawicki: Hmmm, and I guess we should run the import in a separate dir, and run the plan after.

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,204 @@
...
+### Step 3: Initialize Terraform
+
+```bash
+terraform init
+```
+
+### Step 4: Create test objects AND generate CSV
+
+```bash
+terraform apply -auto-approve
```
sfc-gh-asawicki: nit, let's not use auto-approve; also I don't think we need to be so precise with tf initiation and previous state removal

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+**Success criteria:** The plan should show "No changes. Your infrastructure matches the configuration."
+
+If there are changes, the migration script generated incorrect values.
+
+### Step 8: Cleanup (when done)
+
+```bash
+terraform destroy -auto-approve
```
sfc-gh-asawicki: same here, especially here. blinddestruction is never a good idea

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,204 @@
...
+│   ├── datasource.tf        # Fetches users, generates CSV
+│   ├── objects.csv          # Generated CSV (after terraform apply)
+│   └── generated_output.tf  # Generated by migration script
+├── schemas/
+├── warehouses/
+├── grants/
+└── <new_object_type>/       # Add new object types here
+```
+
+## Adding a New Object Type
```
sfc-gh-asawicki: I dont think we need a detailed section for adding the new test object

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-mkowalski: Duplicates can appear, because the code queries grants by multiple conditions and some of the returned sets (that are being concatenated) can overlap.

pkg/manual_tests/migration_script/README.md
sfc-gh-mkowalski: I deleted the entire section according to @sfc-gh-asawicki's comment.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-mkowalski: We want to filter out grants with empty `privilege`, because they are not supported by the migration script.
Filter with `granted_on` turned out to be unnecessary.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-jmichalak: In this case, we should split it into different test cases, as the premise was to test a single data source output.

pkg/manual_tests/migration_script/grants/objects_def.tf
```diff
@@ -0,0 +1,288 @@
...
+    name = "NAME"
+    type = "VARCHAR(100)"
+  }
+}
+
+# ------------------------------------------------------------------------------
+# Warehouse (for account object grants)
+# ------------------------------------------------------------------------------
+resource "snowflake_warehouse" "test_wh" {
+  name           = "MIGRATION_TEST_GRANT_WH"
```
sfc-gh-jmichalak, line 124: I guess we missed this initially, but the names should be unique in general (maybe with a custom suffix, we do such thing in performance tests); please add this to a follow-up ticket.

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,85 @@
...
+terraform init
+terraform apply
+```
+
+This imports the existing Snowflake objects into a fresh state.
+
+### Step 6: Verify plan is empty
+
+```bash
+terraform plan
```
sfc-gh-jmichalak, line 48: Nit: we can fail if it's non empty with a proper flag.

pkg/manual_tests/migration_script/README.md
```diff
@@ -0,0 +1,85 @@
...
+```bash
+cd ..
+terraform destroy
+```
+
+## Test Assertions
+
+The `datasource.tf` includes a **precondition assertion** that fails if no grants are found. This prevents generating an empty CSV that would cause silent failures.
+
+### How it works
```
sfc-gh-jmichalak: Nit: I don't think the `Test Assertions` section is needed.

pkg/manual_tests/migration_script/grants/import/provider.tf
```diff
@@ -0,0 +1,15 @@
+# Provider configuration for import directory
+# This file is kept separate from generated main.tf
+
+terraform {
+  required_providers {
+    snowflake = {
+      source = "snowflakedb/snowflake"
```
sfc-gh-asawicki, line 7: nit (can be in other PR): let's set the version constraint here and in other required provider blocks

pkg/manual_tests/migration_script/grants/test_assertion/main.tf
```diff
@@ -0,0 +1,73 @@
+# =============================================================================
```
sfc-gh-jmichalak: We don't need this.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-jmichalak: This was not addressed. Why do we need to specify the column names here, and it's missing from the readme? The convertion functions should be as generic as possible.

pkg/manual_tests/migration_script/grants/objects_def.tf
```diff
@@ -0,0 +1,288 @@
...
+# ==============================================================================
+# GRANT PRIVILEGES TO ACCOUNT ROLE TESTS
+# ==============================================================================
+
+# ------------------------------------------------------------------------------
+# Grant Privileges on Account
+# ------------------------------------------------------------------------------
+resource "snowflake_grant_privileges_to_account_role" "on_account" {
+  account_role_name = snowflake_account_role.priv_role.name
+  privileges        = ["CREATE DATABASE", "CREATE WAREHOUSE"]
```
sfc-gh-asawicki: nit (can be a separate PR): let's not grant warehouse creation here

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,196 @@
...
+
+  # CSV header - matches the GrantCsvRow struct
+  grants_csv_header = join(",", [for col in local.csv_columns : "\"${col}\""])
+
+  # CSV escape helper - escapes special chars for CSV format
+  csv_escape = {
+    for idx, grant in local.test_grants :
+    idx => {
+      for col in local.csv_columns :
+      col => col == "grant_on" || col == "grant_to" ? "" : (
```
sfc-gh-jmichalak: Why do we put empty strings in certain cases? Especially that you said: `We want to filter out grants with empty privilege, because they are not supported by the migration script.￼Filter with granted_on turned out to be unnecessary.`

sfc-gh-jmichalak (changes requested):

sfc-gh-asawicki (approved):

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,196 @@
...
+    snowflake_grant_account_role.role_to_role,
+  ]
+}
+
+# ==============================================================================
+# GRANTS OF ACCOUNT ROLES (role-to-role and role-to-user grants)
+# ==============================================================================
+
+# Fetch grants OF the child role (shows who it's granted to)
+data "snowflake_grants" "of_child_role" {
```
sfc-gh-jcieslak, line 45: Question: Isn't querying grants **of** child role and grants **to** parent role (data source above) the same thing?

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,196 @@
...
+    [for g in data.snowflake_grants.of_child_role.grants : g],
+    [for g in data.snowflake_grants.to_priv_db_role.grants : g],
+    [for g in data.snowflake_grants.to_parent_db_role.grants : g],
+    [for g in data.snowflake_grants.of_child_db_role.grants : g],
+    [for g in data.snowflake_grants.of_priv_db_role.grants : g]
+  )
+
+  # Filter grants:
+  # - Must contain MIGRATION_TEST in grantee_name or name (test objects only)
+  # - Must have a non-empty privilege (grants_of returns role membership rows with
```
sfc-gh-jcieslak, line 124: I'm confused, you used grants_of data source above, but now we are filtering its output here? Let's either:  
\- show the script can take this output, but won't do anything with it  
\- or remove the data souces that aren't used anyway

let's discuss

pkg/manual_tests/migration_script/grants/datasource.tf
```diff
@@ -0,0 +1,196 @@
...
+  ]
+
+  # CSV column definitions - order matters for output
+  csv_columns = ["privilege", "granted_on", "grant_on", "name", "granted_to", "grant_to", "grantee_name", "grant_option", "granted_by"]
+
+  # CSV header - matches the GrantCsvRow struct
+  grants_csv_header = join(",", [for col in local.csv_columns : "\"${col}\""])
+
+  # CSV escape helper - escapes special chars for CSV format
+  csv_escape = {
```
sfc-gh-jcieslak, line 136: nit: maybe we could simplify it with built-in functions (encode, then decode to JSON maybe would solve this?)

pkg/manual_tests/migration_script/grants/import/provider.tf
sfc-gh-jcieslak, line 7: Question 1: Is it latest version by default?  


Question 2: If it even is the latest version, should we test against latest production version or local ("dev") version? I guess it shouldn't matter that much because here we mostly focus on migration script, but I'm curious what everyone thinks

pkg/manual_tests/migration_script/grants/test_assertion/main.tf
```diff
@@ -0,0 +1,73 @@
...
+provider "snowflake" {}
+
+# Query grants to PUBLIC role (always exists, will have some system grants)
+data "snowflake_grants" "empty_test" {
+  grants_to {
+    account_role = "PUBLIC"
+  }
+}
+
+locals {
```
sfc-gh-jcieslak: nit / idea: we could think about creating a module for this transformation (if possible)

pkg/manual_tests/migration_script/grants/objects_def.tf
sfc-gh-mkowalski: Will be done in a follow-up ticket.

pkg/manual_tests/migration_script/README.md
sfc-gh-mkowalski: We can leave it as is.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-mkowalski: It turned out to be unnecessary.

pkg/manual_tests/migration_script/grants/import/provider.tf
sfc-gh-mkowalski: Added version 2.11.0

pkg/manual_tests/migration_script/grants/objects_def.tf
sfc-gh-mkowalski: Deleted

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-mkowalski: We want to keep all the datasources for universality.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-mkowalski: It seems there are no datasources that allow that, and therefore it's easier to just filter out it this way.
Could be discussed offline.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-mkowalski: It seems there are only functions for JSON, but not for the CSV type.

pkg/manual_tests/migration_script/grants/import/provider.tf
sfc-gh-mkowalski: The version has been set to 2.11.0.
We test stable versions only anyway.

pkg/manual_tests/migration_script/grants/test_assertion/main.tf
sfc-gh-mkowalski: To be discussed in the follow-up.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-mkowalski: We discussed it and decided it's fine to leave as is.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-mkowalski: Removed

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-jmichalak: Not exactly, the premise was to provide basic tests. This could be later split into separate tests. Had we tested only selected data sources, we wouldn't have needed even the row deduplication.

pkg/manual_tests/migration_script/grants/datasource.tf
sfc-gh-jcieslak: This related to https://github.com/snowflakedb/terraform-provider-snowflake/pull/4271#discussion_r2611420074, let's do this later

PR #4276, opened 2025-12-09
feat: Add proxy fields to the provider

sfc-gh-mkowalski:
Added proxy fields supported in the new [gosnowflake](https://github.com/snowflakedb/gosnowflake/blob/1a798e3124a04e945ae4d54242fbed84513b87f1/dsn.go) driver version to the provider:
- ProxyHost
- ProxyPort
- ProxyUser
- ProxyPassword
- ProxyProtocol
- NoProxy

### DIFF
--- a/pkg/acceptance/bettertestspoc/config/providermodel/snowflake_model_ext.go
+++ b/pkg/acceptance/bettertestspoc/config/providermodel/snowflake_model_ext.go
@@ -141,12 +141,12 @@ func (m *SnowflakeModel) AllFields(tmpConfig *helpers.TmpTomlConfig, tmpUser *he
 		WithWorkloadIdentityEntraResource("workload_identity_entra_resource").
 		WithLogQueryText(true).
 		WithLogQueryParameters(true).
-		WithProxyHost(""). // TODO: Remove empty proxy after testing
+		WithProxyHost("").
 		WithProxyPort(443).
 		WithProxyUser("proxy_user").
 		WithProxyPassword("proxy_password").
-		WithProxyProtocol("proxy_protocol").
-		WithNoProxy("no_proxy")
+		WithProxyProtocol("https").
+		WithNoProxy("localhost,snowflake.computing.com")
 }
 
 func PatConfig(h helpers.TmpServiceUserWithPat) *SnowflakeModel {
--- a/pkg/acceptance/helpers/config_toml_creator.go
+++ b/pkg/acceptance/helpers/config_toml_creator.go
@@ -123,8 +123,8 @@ func FullInvalidTomlConfigForServiceUser(t *testing.T, profile string) string {
 		WithProxyPort(443).
 		WithProxyUser("proxy_user").
 		WithProxyPassword("proxy_password").
-		WithProxyProtocol("proxy_protocol").
-		WithNoProxy("no_proxy")
+		WithProxyProtocol("https").
+		WithNoProxy("localhost,snowflake.computing.com")
 	return configDtoToTomlString(t, profile, dto)
 }
 
--- a/pkg/acceptance/helpers/legacy_config_toml_creator.go
+++ b/pkg/acceptance/helpers/legacy_config_toml_creator.go
@@ -48,7 +48,7 @@ func FullLegacyTomlConfigForServiceUser(t *testing.T, profile string, userId sdk
 			WithParams(map[string]*string{"foo": sdk.Pointer("bar")}).
 			WithLogQueryText(true).
 			WithLogQueryParameters(true).
-			WithProxyHost(""). // TODO: Remove empty proxy after testing
+			WithProxyHost("").
 			WithProxyPort(443).
 			WithProxyUser("proxy_user").
 			WithProxyPassword("proxy_password").


sfc-gh-mkowalski:
Acceptance tests were omitted for now, because they require providing a real proxy host with its password which we don't want to hardcode; We are waiting for a response from the drivers team regarding some way to mock these values.

docs/index.md
```diff
@@ -142,11 +143,16 @@ provider "snowflake" {
...
 - `profile` (String) Sets the profile to read from ~/.snowflake/config file. Can also be sourced from the `SNOWFLAKE_PROFILE` environment variable.
 - `protocol` (String) A protocol used in the connection. Valid options are: `http` | `https`. Can also be sourced from the `SNOWFLAKE_PROTOCOL` environment variable.
+- `proxy_host` (String) The host of the proxy to use for the connection. Can also be sourced from the `SNOWFLAKE_PROXY_HOST` environment variable.
+- `proxy_password` (String, Sensitive) The password of the proxy to use for the connection. Can also be sourced from the `SNOWFLAKE_PROXY_PASSWORD` environment variable.
+- `proxy_port` (Number) The port of the proxy to use for the connection. Can also be sourced from the `SNOWFLAKE_PROXY_PORT` environment variable.
+- `proxy_protocol` (String) The protocol of the proxy to use for the connection. Can also be sourced from the `SNOWFLAKE_PROXY_PROTOCOL` environment variable.
```
sfc-gh-jcieslak: I think it should be an enum since there are only two options available (just like the `protocol` field above, and it should be similarly documented listing those options)

pkg/provider/provider.go
```diff
@@ -489,6 +489,43 @@ func GetProviderSchema() map[string]*schema.Schema {
...
+			Sensitive:   true,
+			DefaultFunc: schema.EnvDefaultFunc(snowflakeenvs.ProxyPassword, nil),
+		},
+		"proxy_protocol": {
+			Type:        schema.TypeString,
+			Description: envNameFieldDescription("The protocol of the proxy to use for the connection.", snowflakeenvs.ProxyProtocol),
+			Optional:    true,
+			DefaultFunc: schema.EnvDefaultFunc(snowflakeenvs.ProxyProtocol, nil),
+		},
+		"no_proxy": {
```
sfc-gh-jcieslak, line 524: nit: I know the config is pretty much pass-through to driver, but we could do better here on our side and create an actual list (or Set) of hostnames, domains, and IP addresses. I think, It would be better UX.

pkg/provider/provider.go
```diff
@@ -489,6 +489,43 @@ func GetProviderSchema() map[string]*schema.Schema {
...
+		"proxy_password": {
+			Type:        schema.TypeString,
+			Description: envNameFieldDescription("The password of the proxy to use for the connection.", snowflakeenvs.ProxyPassword),
+			Optional:    true,
+			Sensitive:   true,
+			DefaultFunc: schema.EnvDefaultFunc(snowflakeenvs.ProxyPassword, nil),
+		},
+		"proxy_protocol": {
+			Type:        schema.TypeString,
+			Description: envNameFieldDescription("The protocol of the proxy to use for the connection.", snowflakeenvs.ProxyProtocol),
```
sfc-gh-jcieslak: as mentioned above, let's maybe make it an enum with documented values

examples/additional/provider_config_tf.MD
```diff
@@ -40,6 +40,12 @@ provider "snowflake" {
 	enable_single_use_refresh_tokens = true
 	log_query_text = false
 	log_query_parameters = false
+	proxy_host = "proxy.example.com"
+	proxy_port = 443
+	proxy_user = "username"
+	proxy_password = "****"
```
sfc-gh-jmichalak: Let's use a variable here, like with `var.oauth_client_secret`.

CHANGELOG.md
```diff
@@ -4656,7 +4656,7 @@
 ### BugFixes
 
 * adding in issue link to slackbot ([#1158](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1158)) ([6f8510b](https://github.com/Snowflake-Labs/terraform-provider-snowflake/commit/6f8510b8e8b7c6b415ef6258a7c1a2f9e1b547c4))
-* Deleting a snowflake_user and their associated snowlfake_role_grant causes an error ([#1142](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1142)) ([5f6725a](https://github.com/Snowflake-Labs/terraform-provider-snowflake/commit/5f6725a8d0df2f5924c6d6dc2f62ebeff77c8e14))
+* Deleting a snowflake_user and their associated snowflake_role_grant causes an error ([#1142](https://github.com/Snowflake-Labs/terraform-provider-snowflake/issues/1142)) ([5f6725a](https://github.com/Snowflake-Labs/terraform-provider-snowflake/commit/5f6725a8d0df2f5924c6d6dc2f62ebeff77c8e14))
```
sfc-gh-jmichalak: This file shouldn't have any changes.

pkg/acceptance/helpers/config_toml_creator.go
```diff
@@ -58,7 +58,13 @@ func FullTomlConfigForServiceUser(t *testing.T, profile string, userId sdk.Accou
...
-		WithLogQueryParameters(true),
+		WithLogQueryParameters(true).
+		WithProxyHost("proxy_host").
+		WithProxyPort(443).
+		WithProxyUser("proxy_user").
+		WithProxyPassword("proxy_password").
+		WithProxyProtocol("proxy_protocol").
+		WithNoProxy("no_proxy"),
 	)
 }
```
sfc-gh-jmichalak, line 70: <p class='graphite__hidden'><i>[Re: line +121]</i></p>
They should be added here as well.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/snowflakedb/terraform-provider-snowflake/4276?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/provider/provider.go
sfc-gh-jmichalak: Yes, but let's discuss first. What if a few elements are set in env, a few in TF config, and a few in TOML? Should we merge them or override? Because I believe the TF config overrides the env entirely...

pkg/provider/provider_test.go
```diff
@@ -124,6 +124,12 @@ func TestGetDriverConfigFromTerraform_AllFields(t *testing.T) {
 		"workload_identity_entra_resource": "workload_identity_entra_resource",
```
sfc-gh-jmichalak, line 130: <p class='graphite__hidden'><i>[Re: line +75]</i></p>
They should be added here as well.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/snowflakedb/terraform-provider-snowflake/4276?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/provider/provider.go
```diff
@@ -489,6 +489,43 @@ func GetProviderSchema() map[string]*schema.Schema {
 			Optional:    true,
 			DefaultFunc: schema.EnvDefaultFunc(snowflakeenvs.LogQueryParameters, nil),
 		},
+		"proxy_host": {
```
sfc-gh-jmichalak, line 492: Acceptance tests in `pkg/testacc/provider_acceptance_test.go` are missing. _I_ know that we need to have a proper proxy, but I found that when we skip ProxyHost only and set other Proxy fields, the driver doesn't connect to that proxy. But first, we should make sure this is the correct driver behavior. If it is, we can at least set the other fields and create a follow-up for ProxyHost.

docs/index.md
```diff
@@ -128,6 +128,7 @@ provider "snowflake" {
 - `log_query_text` (Boolean) When set to true, the full query text will be logged. Be aware that it may include sensitive information. Default value is false. Can also be sourced from the `SNOWFLAKE_LOG_QUERY_TEXT` environment variable.
 - `login_timeout` (Number) Login retry timeout in seconds EXCLUDING network roundtrip and read out http response. Can also be sourced from the `SNOWFLAKE_LOGIN_TIMEOUT` environment variable.
 - `max_retry_count` (Number) Specifies how many times non-periodic HTTP request can be retried by the driver. Can also be sourced from the `SNOWFLAKE_MAX_RETRY_COUNT` environment variable.
+- `no_proxy` (String) A comma-separated list of hostnames, domains, and IP addresses to exclude from proxying. Can also be sourced from the `SNOWFLAKE_NO_PROXY` environment variable.
```
sfc-gh-jmichalak, line 131: What about

```
clarify with the driver team if using the new fields is preferred over using the HTTPS_PROXY or NO_PROXY and what is the relation (gosnowflake docs)
```

What if someone uses HTTPS_PROXY or NO_PROXY already? Does it get overridden?

pkg/provider/provider.go
sfc-gh-jcieslak: Imo override. I saw that the values are mostly 1 to 1 mapping so list to string mapping could be a bit more tricky, but just wanted to mention it would better from the configuration perspective.

CHANGELOG.md
sfc-gh-mkowalski: I just corrected a typo that I accidentally spotted.

pkg/provider/provider.go
sfc-gh-mkowalski: To be added in the follow-up.

docs/index.md
sfc-gh-mkowalski: Tests validating this behaviour to be added in the follow-up.
It is expected from the config to override the env variables.

pkg/provider/provider.go
sfc-gh-mkowalski: Added tests with empty `ProxyHost`. Using proper proxy for the follow-up.

sfc-gh-jcieslak (approved):

CHANGELOG.md
sfc-gh-jmichalak: Reverted because this file is autogenerated.

pkg/acceptance/bettertestspoc/config/providermodel/snowflake_model_ext.go
```diff
@@ -140,7 +140,13 @@ func (m *SnowflakeModel) AllFields(tmpConfig *helpers.TmpTomlConfig, tmpUser *he
 		WithWorkloadIdentityProvider("workload_identity_provider").
 		WithWorkloadIdentityEntraResource("workload_identity_entra_resource").
 		WithLogQueryText(true).
-		WithLogQueryParameters(true)
+		WithLogQueryParameters(true).
+		WithProxyHost("").
```
sfc-gh-jmichalak, line 144: nit: TODOs are missing. In general, in each place where we don't test the WithProxyHost, there should be a TODO with a ticket number.

docs/index.md
sfc-gh-jmichalak: Ok, but a TODO annotation is missing.

pkg/provider/provider.go
```diff
@@ -489,6 +489,44 @@ func GetProviderSchema() map[string]*schema.Schema {
...
+		"proxy_password": {
+			Type:        schema.TypeString,
+			Description: envNameFieldDescription("The password of the proxy to use for the connection.", snowflakeenvs.ProxyPassword),
+			Optional:    true,
+			Sensitive:   true,
+			DefaultFunc: schema.EnvDefaultFunc(snowflakeenvs.ProxyPassword, nil),
+		},
+		"proxy_protocol": {
+			Type:             schema.TypeString,
+			Description:      envNameFieldDescription(fmt.Sprintf("The protocol of the proxy to use for the connection. Valid options are: %v.", docs.PossibleValuesListed(allProtocols)), snowflakeenvs.ProxyProtocol),
```
sfc-gh-jmichalak, line 519: nit: the note that the value is case-insensitive is missing.

PR #4280, opened 2025-12-11
feat: Add strict privilege management behavior to grant privileges to account role

sfc-gh-jcieslak:
## Changes
- Add new `strict_privilege_management` field with described limitations
- Adjust `grant_privilege_to_account_role` `Read` operation behavior
  - If the field is set, check the experimental flag (if not set, throw an error)
  - Include the flag whenever it's necessary in the privilege picking logic
  - Adjust the structure of privilege picking logic to be more readable
- Add acceptance tests
  - Add new provider for the strict privilege management tests

## Next prs
- Add example usage
- Add migration guide entry

### DIFF
--- a/pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
+++ b/pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
@@ -2532,12 +2532,13 @@ func TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_Updates(t *testin
 		WithPrivileges(string(sdk.AccountObjectPrivilegeMonitor)).
 		WithOnAccountObject(sdk.ObjectTypeDatabase, database.ID())
 
-	resourceModelWithUpdatedPrivileges := model.GrantPrivilegesToAccountRole("test", role.ID().Name()).
-		WithPrivileges(string(sdk.AccountObjectPrivilegeOperate)).
-		WithOnAccountObject(sdk.ObjectTypeDatabase, database.ID())
-
 	resourceModelWithStrictRoleManagement := model.GrantPrivilegesToAccountRole("test", role.ID().Name()).
 		WithPrivileges(string(sdk.AccountObjectPrivilegeMonitor)).
+		WithOnAccountObject(sdk.ObjectTypeDatabase, database.ID()).
+		WithStrictPrivilegeManagement(true)
+
+	resourceModelWithStrictRoleManagementAndUpdatedPrivileges := model.GrantPrivilegesToAccountRole("test", role.ID().Name()).
+		WithPrivileges(string(sdk.AccountObjectPrivilegeModify)).
 		WithOnAccountObject(sdk.ObjectTypeDatabase, database.ID()).
 		WithStrictPrivilegeManagement(true)
 
@@ -2628,13 +2629,13 @@ func TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_Updates(t *testin
 			},
 			// Confirm that StrictPrivilegeManagement shouldn't influence regular update operations
 			{
-				Config: accconfig.FromModels(t, providerModel, resourceModelWithUpdatedPrivileges),
+				Config: accconfig.FromModels(t, providerModel, resourceModelWithStrictRoleManagementAndUpdatedPrivileges),
 				Check: assertThat(t,
-					resourceassert.GrantPrivilegesToAccountRoleResource(t, resourceModelWithUpdatedPrivileges.ResourceReference()).
+					resourceassert.GrantPrivilegesToAccountRoleResource(t, resourceModelWithStrictRoleManagementAndUpdatedPrivileges.ResourceReference()).
 						HasAccountRoleNameString(role.ID().Name()).
 						HasStrictPrivilegeManagementString("true").
-						HasPrivileges(string(sdk.AccountObjectPrivilegeOperate)),
-					assert.Check(queriedAccountRolePrivilegesEqualTo(t, role.ID(), string(sdk.AccountObjectPrivilegeOperate))),
+						HasPrivileges(string(sdk.AccountObjectPrivilegeModify)),
+					assert.Check(queriedAccountRolePrivilegesEqualTo(t, role.ID(), string(sdk.AccountObjectPrivilegeModify))),
 				),
 			},
 		},


pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -843,36 +870,43 @@ func ReadGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceDat
 	}
 
 	for _, grant := range grants {
```
sfc-gh-jcieslak, line 876: Changes in this `for` I wasn't sure of as ideally we should only add to what's already there to not touch the existing logic, but I still adjusted the `ifs` a bit to make the whole filtering logic more readable.

docs/resources/grant_privileges_to_account_role.md
```diff
@@ -281,6 +281,7 @@ resource "snowflake_grant_privileges_to_account_role" "example" {
...
 - `on_schema_object` (Block List, Max: 1) Specifies the schema object on which privileges will be granted. (see [below for nested schema](#nestedblock--on_schema_object))
 - `privileges` (Set of String) The privileges to grant on the account role. This field is case-sensitive; use only upper-case privileges.
+- `strict_privilege_management` (Boolean) (Default: `false`) If true, the resource will revoke all privileges that are not explicitly defined in the config making it a central source of truth for the privileges granted on an object to an account role. If false, the resource will be only concerned with the privileges that are explicitly defined in the config. The potential privilege removals will be planned only after second `terraform apply` run, when the flag is set in the state. This means, the resource creation and flag update doesn't revoke immediately any externally granted privileges. External privileges will be detected regardless of their grant option. The parameter can be only used when `GRANTS_STRICT_PRIVILEGE_MANAGEMENT` option is specified in provider block in the `experimental_features_enabled` field.
```
sfc-gh-jmichalak: Nit: This paragraph lacks consistency. E.g. the `If false, the resource will be only concerned with the privileges that are explicitly defined in the config.` is in the middle of describing the `true` option.  
`This means, the resource creation and flag update doesn't revoke immediately any externally granted privileges` - why? I would expect that just when I set this flag, the resource should revoke the other grants immediately.

pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -843,36 +870,41 @@ func ReadGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceDat
...
-		// Only consider privileges that are already present in the ID, so we
-		// don't delete privileges managed by other resources.
-		if !slices.Contains(expectedPrivileges, grant.Privilege) {
+
+		// Only consider privileges specified by the configuration, so we don't delete privileges managed by other resources.
+		// The privilege should also match grant option with the one stored in the configuration.
+		//
+		// The strict privilege management skips this check, because we want to take into account all privileges
+		// (not only those specified in the config) and we want to get privileges regardless of grant option setting.
+		if !strictPrivilegeManagement && (!slices.Contains(expectedPrivileges, grant.Privilege) || grant.GrantOption != id.WithGrantOption) {
```
sfc-gh-jmichalak, line 889: So, when the strict flag is set, and the grant option is different, we are skipping? This way the privileges with different grantoptions won't be detected.

pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -843,36 +870,41 @@ func ReadGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceDat
...
-				actualPrivileges = append(actualPrivileges, grant.Privilege)
-			} else if grantedOn == grant.GrantedOn || grantedOn == grant.GrantOn {
-				actualPrivileges = append(actualPrivileges, grant.Privilege)
-			}
+		// If grantedBy is an empty string, it means:
+		// - it's a future grant, or
+		// - it's a predefined Snowflake grant
+		// Thus, we should skip if we are getting future grants when not specified in the SHOW opts, or
+		// we are dealing with predefined object other than the "SNOWFLAKE" database (unsupported feature).
+		if (opts.Future == nil || !*opts.Future) && grant.GrantedBy.Name() == "" && grant.Name.Name() != "SNOWFLAKE" {
```
sfc-gh-jmichalak, line 898: When the strict flag is set, shouldn't the future grants be revoked?

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2413,331 @@ variable "privilege" {
...
+	t.Cleanup(roleCleanup)
+
+	shareExternalId := createSharedDatabaseOnSecondaryAccount(t)
+	databaseFromShare, databaseFromShareCleanup := testClient().Database.CreateDatabaseFromShare(t, shareExternalId)
+	t.Cleanup(databaseFromShareCleanup)
+
+	providerModel := providermodel.SnowflakeProvider().
+		WithExperimentalFeaturesEnabled(experimentalfeatures.GrantsStrictPrivilegeManagement)
+	resourceModel := model.GrantPrivilegesToAccountRole("test", role.ID().Name()).
+		WithPrivileges(string(sdk.AccountObjectPrivilegeImportedPrivileges)).
```
sfc-gh-jmichalak, line 2745: Q: what when the imported privilege is missing from the config? It can't be revoked, so this will error out, right?

sfc-gh-jmichalak (changes requested):

docs/resources/grant_privileges_to_account_role.md
sfc-gh-jcieslak: Well, I'll try to simplify.
Regarding not immediate change, let's say user wants to update this field, this will go through the following stages:
Read (planning) -> Update (changing the state) -> Read (refresh after the update to compare the state).

By modifying the read operation, we will be only able to detect external privileges **after** the Update, because that's where the flag is set. Only then we could go through the same process, but this time, the first read operation will also detect the external privileges and the Update would be able to act right away.

I didn't try it yet, but perhaps using RawConfig could make it better 🤔

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jcieslak: When the flag is set this will be always skipped `if false && (whatever) == don't skip`

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jcieslak: That's a good question, on the second thought, I need to rethink how futures should or could behave with newly added flag and will discuss internally

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: Not sure, If I understand correctly, the SNOWFLAKE database can be granted **only** with `imported privileges` (and I think this is true for other cases where it can be granted), so that's the only option, but when the user would try possibly other predefined object with grants on it, then we could encounter errors, but that applies to the resource itself, not the newly introduced flag.

docs/resources/grant_privileges_to_account_role.md
sfc-gh-asawicki: I think that:
1. `when the flag is set in the state` -> `after setting the flag in config` or something similar. The bottom line here is that we currently don't have a way to react to the flag change with the list of privileges to be removed (there are options that we considered or will condider in the future)
2. I wouldn't mention create. The fact that you can create a resource when the privileges are already granted is a lucky behavior of regranting. If the object existts/grants are granted, the proper way would be through import. And it's normal with import that if the config does not match, it needs apply after import.

pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -843,36 +870,41 @@ func ReadGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceDat
...
-		if grant.GrantTo != sdk.ObjectTypeRole && grant.GrantedTo != sdk.ObjectTypeRole {
+		// Process only (ACCOUNT) ROLE specified in the config
+		if (grant.GrantTo != sdk.ObjectTypeRole && grant.GrantedTo != sdk.ObjectTypeRole) || grant.GranteeName.Name() != id.RoleName.Name() {
 			continue
 		}
-		// Only consider privileges that are already present in the ID, so we
-		// don't delete privileges managed by other resources.
-		if !slices.Contains(expectedPrivileges, grant.Privilege) {
+
+		// Only consider privileges specified by the configuration, so we don't delete privileges managed by other resources.
```
sfc-gh-asawicki: `Only consider privileges specified by the configuration, so we don't delete privileges managed by other resources.` -> what do you mean by that? If the strict flag is set, we MUST delete privileges that are not listed.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2413,331 @@ variable "privilege" {
...
+	resourceModelWithStrictRoleManagement := model.GrantPrivilegesToAccountRole("test", role.ID().Name()).
+		WithPrivileges(string(sdk.AccountObjectPrivilegeMonitor)).
+		WithOnAccountObject(sdk.ObjectTypeDatabase, database.ID()).
+		WithStrictPrivilegeManagement(true)
+
+	resource.Test(t, resource.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
```
sfc-gh-asawicki: if you are creating a new model, you need to use a different provider factory here, e.g. providerFactoryUsingCache("TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_BasicOnCreate")

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2413,331 @@ variable "privilege" {
 }
 `, roleId.Name(), dbId.Name())
 }
+
+func TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_BasicOnCreate(t *testing.T) {
```
sfc-gh-asawicki, line 2420: there are multiple tests failing as this flag is not correctly imported

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2413,331 @@ variable "privilege" {
...
+		Steps: []resource.TestStep{
+			{
+				Config: accconfig.FromModels(t, providerModel, resourceModelWithStrictRoleManagement),
+				Check: assertThat(t,
+					resourceassert.GrantPrivilegesToAccountRoleResource(t, resourceModelWithStrictRoleManagement.ResourceReference()).
+						HasStrictPrivilegeManagementString("true").
+						HasPrivileges(string(sdk.AccountObjectPrivilegeMonitor)),
+					assert.Check(queriedAccountRolePrivilegesEqualTo(t, role.ID(), string(sdk.AccountObjectPrivilegeMonitor), string(sdk.AccountObjectPrivilegeUsage))),
+				),
+				ExpectNonEmptyPlan: true,
```
sfc-gh-asawicki, line 2498: Why the plan is not empty after creation? It shouldn't.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2413,331 @@ variable "privilege" {
...
+	resourceModelWithStrictRoleManagement := model.GrantPrivilegesToAccountRole("test", role.ID().Name()).
+		WithPrivileges(string(sdk.AccountObjectPrivilegeMonitor)).
+		WithOnAccountObject(sdk.ObjectTypeDatabase, database.ID()).
+		WithStrictPrivilegeManagement(true)
+
+	resource.Test(t, resource.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
```
sfc-gh-asawicki: Same as above. Maybe you can create a separate factory like e.g. `viewsProviderFactory`

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2413,331 @@ variable "privilege" {
...
+						HasStrictPrivilegeManagementString("false").
+						HasPrivileges(string(sdk.AccountObjectPrivilegeMonitor)),
+					assert.Check(queriedAccountRolePrivilegesEqualTo(t, role.ID(), string(sdk.AccountObjectPrivilegeMonitor))),
+				),
+			},
+			{
+				PreConfig: func() {
+					// Notice, the grant option is set in the externally granted privilege.
+					// This means, the strict privilege management works regardless of the grant option setting.
+					testClient().Grant.GrantPrivilegesOnDatabaseToAccountRole(t, role.ID(), database.ID(), []sdk.AccountObjectPrivilege{sdk.AccountObjectPrivilegeUsage}, true)
```
sfc-gh-asawicki, line 2567: hmm, but we discussed it should revoke it in such a case, right?

docs/resources/grant_privileges_to_account_role.md
sfc-gh-jcieslak: Adjusted

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jcieslak: after further discussions, the support for future grants will be added in the follow-up prs

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jcieslak: The strict privilege management behavior is explained in second paragraph, but maybe it should be merged for clarity? For now, left as it is.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: In this case, it is not empty because the read after creation detected additional privileges which were granted outside of Terraform knowledge before. Here we created resource with grants on object O to role R privileges A and B. After whole create, we Read that indeed, the A and B privileges are there, but because the `strict_privilege_management` flag is there, we read additional privileges not defined in the configuration, thus we see there's also privilege C granted. This example is shown here.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: It seems this is something worse than just changing the provider factory as we should modify the resource identifier, create a migration function, and test it. I'll add it.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: In this case no, because in this step the flag is set to false, in the step 3 we set it to true (that's why we observe the non empty plan afterwards), and in step 4 the user can actually see the privileges in the plan and that's where they're removed.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-asawicki: wait wait - let's not modify the resource's identifier as this will be a breaking change. let's discuss.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-asawicki: Okay, so we don't have a basic create (so no privileges granted pre-creation)? We should have it.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: chill 😏 adjusted only the import as discussed

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: as above, adjusted the import

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: True, added basic case without external privileges

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: I don't think this is needed, it's failing due to the incorrect import failing, after the fix they should pass

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: Added a new provider as suggested

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-asawicki: Oh, okay, I see that previous comment was rephrased. Let's clarify it by starting with something like: "depending on the strict privilege management flag:" and the if not set (the old), if set (the new skip expl.)

docs/resources/grant_privileges_to_account_role.md
```diff
@@ -281,6 +281,7 @@ resource "snowflake_grant_privileges_to_account_role" "example" {
...
 - `on_schema_object` (Block List, Max: 1) Specifies the schema object on which privileges will be granted. (see [below for nested schema](#nestedblock--on_schema_object))
 - `privileges` (Set of String) The privileges to grant on the account role. This field is case-sensitive; use only upper-case privileges.
+- `strict_privilege_management` (Boolean) (Default: `false`) If true, the resource will revoke all privileges that are not explicitly defined in the config making it a central source of truth for the privileges granted on an object to an account role. If false, the resource will be only concerned with the privileges that are explicitly defined in the config. The potential privilege removals will be planned only after second `terraform apply` run, after setting the flag in resource configuration. This means, the flag update doesn't revoke immediately any externally granted privileges. External privileges will be detected regardless of their grant option. The parameter can be only used when `GRANTS_STRICT_PRIVILEGE_MANAGEMENT` option is specified in provider block in the `experimental_features_enabled` field.
```
sfc-gh-asawicki: Two changes:
1.  "The potential privilege removals will be planned only after second `terraform apply` run" -> let's here say a short why - to be able to list the privileges that will be removed.
2. Let's have a link to `experimental_features_enabled`.

pkg/scripts/migration_script/README.md
```diff
@@ -39,6 +39,15 @@ However, if you're using an older version, you can still utilize the script as l
...
 
+### Different behavior for import options
+
+> **Important:** Read before going further!
+
+The script is designed to work with the latest version of the provider.
+However, if you're using an older version, you can still utilize the script as long as the object types you need are supported and haven't undergone major changes.
+For instance, with grants, you can use the script to transition from [old to new grants](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/guides/grants_redesign_design_decisions#mapping-from-old-grant-resources-to-the-new-ones)
+since they haven't significantly changed from the current provider version, but there may be minor differences like quotes handling in identifiers.
+
```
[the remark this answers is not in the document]
sfc-gh-asawicki: Hmm, it's duplicated, why?

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2416,394 @@ variable "privilege" {
...
+	resourceModelMissingExperiment := model.GrantPrivilegesToAccountRole("test", role.ID().Name()).
+		WithPrivileges(string(sdk.GlobalPrivilegeCreateDatabase)).
+		WithOnAccount(true).
+		WithStrictPrivilegeManagement(true)
+
+	resource.Test(t, resource.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		ProtoV6ProviderFactories: strictPrivilegeManagementGrantProviderFactory,
```
sfc-gh-asawicki: If you use this factory and other test with provider config containing experiment flag is used, then this test won't work. You need to use different factory (e.g. the base one) in this test.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2416,394 @@ variable "privilege" {
...
+		CheckDestroy:             CheckAccountRolePrivilegesRevoked(t),
+		Steps: []resource.TestStep{
+			// Expect empty plan as external privileges match the ones defined within the configuration
```
sfc-gh-asawicki: expect empty plan when? before or after? If before, then add a check for that. If after, then pls adjust the comment.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2416,394 @@ variable "privilege" {
...
+
+func TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_BasicOnCreate(t *testing.T) {
+	role, roleCleanup := testClient().Role.CreateRole(t)
+	t.Cleanup(roleCleanup)
+
+	database, databaseCleanup := testClient().Database.CreateDatabase(t)
+	t.Cleanup(databaseCleanup)
+
+	providerModel := providermodel.SnowflakeProvider().WithExperimentalFeaturesEnabled(experimentalfeatures.GrantsStrictPrivilegeManagement)
+	resourceModelWithStrictRoleManagement := model.GrantPrivilegesToAccountRole("test", role.ID().Name()).
```
sfc-gh-asawicki, line 2428: so as I understand, this privilege already exists, right? In this case, the import should be the proper way. Let's add a test for a privilege that's not added before applying the config.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2402,3 +2416,394 @@ variable "privilege" {
...
+					resourceassert.GrantPrivilegesToAccountRoleResource(t, resourceModel.ResourceReference()).
+						HasPrivileges(string(sdk.AccountObjectPrivilegeImportedPrivileges)).
+						HasStrictPrivilegeManagementString("true"),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_OnSnowflakeDatabase(t *testing.T) {
```
sfc-gh-asawicki, line 2783: question: why we need this test?

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-asawicki: Hmmm, I see it was added after all :D

sfc-gh-asawicki (changes requested):
There are failing tests, pls update with dev and fix 2 new tests

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jcieslak: Sure, adjusted

pkg/scripts/migration_script/README.md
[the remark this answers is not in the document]
sfc-gh-jcieslak: Didn't merge with dev yet. Now merged and resolved the changes

docs/resources/grant_privileges_to_account_role.md
sfc-gh-jcieslak: Done

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: Done

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: Adjust the comment

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: I have BasicOnCreate and BasicOnCreate_WithAdditionalExistingPrivilegesOnSnowflakeSide test to show both versions

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: As the change involved a slight adjustment of read which is pretty complex I wanted the assurance that:
1) The flag can be used with Snowflake application / database
2) The logic for Imported Privileges with the flag enabled works as intended

sfc-gh-jmichalak (commented):
The `TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_Updates` test is failing.

PR #4281, opened 2025-12-11
chore: Adjustments before strict privilege management behavior change in grants

sfc-gh-jcieslak:
## Changes
- Adjust grant privileges to account role assertions
- Add new experimental flag


docs/index.md
```diff
@@ -116,7 +116,7 @@ provider "snowflake" {
...
+- `experimental_features_enabled` (Set of String) A list of experimental features. Similarly to preview features, they are not yet stable features of the provider. Enabling given experiment is still considered a preview feature, even when applied to the stable resource. These switches offer experiments altering the provider behavior. If the given experiment is successful, it can be considered an addition in the future provider versions. This field can not be set with environmental variables. Valid options are: `PARAMETERS_IGNORE_VALUE_CHANGES_IF_NOT_ON_OBJECT_LEVEL` | `WAREHOUSE_SHOW_IMPROVED_PERFORMANCE` | `GRANTS_STRICT_PRIVILEGE_MANAGEMENT`.
```
sfc-gh-jmichalak, line 126: Q: regarding all the valid options, why are they actually in upper case (they are not related to SQL), and why we don't say these values are case-insensitive?

sfc-gh-asawicki: This way we are not committing to their end state, as we are still discussing how we want to handle and document them. IMO, not a question for now.

sfc-gh-asawicki (approved):

PR #4285, opened 2025-12-12
feat: support private facts and metrics in semantic views

sfc-gh-sarur:
Adding support for private facts and metrics.

## Test Plan

* [x] acceptance tests
* [x] integration tests

## References

*

### DIFF
--- a/pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
+++ b/pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
@@ -209,7 +209,6 @@ func (s *SemanticViewModel) WithFacts(facts []sdk.FactDefinition) *SemanticViewM
 			semExpVar["synonym"] = tfconfig.SetVariable(syns...)
 		}
 		m["semantic_expression"] = tfconfig.ListVariable(tfconfig.ObjectVariable(semExpVar))
-		//isPrivateFact := v.isPrivateFact()
 		if v.IsPrivateFact() {
 			m["private_fact"] = tfconfig.StringVariable("PRIVATE")
 		}


sfc-gh-jmichalak:
Hi @sfc-gh-sarur 👋 

Thank you for your contribution! The team will review it next week. For the next PRs, please discuss them with us on our channel first :)

sfc-gh-sarur:
> Hi @sfc-gh-sarur 👋
> 
> Thank you for your contribution! The team will review it next week. For the next PRs, please discuss them with us on our channel first :)

thanks @sfc-gh-jmichalak , I will make sure to discuss further changes with you first. Please let me know if we should discuss this PR as well!

pkg/resources/semantic_view.go
```diff
@@ -182,31 +182,46 @@ var semanticViewsSchema = map[string]*schema.Schema{
...
-					),
-				},
-				"sql_expression": {
-					Type:        schema.TypeString,
-					Required:    true,
-					Description: "The SQL expression used to compute the fact.",
-				},
-				"synonym": {
-					Type:        schema.TypeSet,
+				"semantic_expression": {
```
sfc-gh-jcieslak, line 185: Hmm, this is a breaking change. It can be done as it is a preview feature, but we try to not do them and I don't it's an exception here (let's try to not break the current schema). Also, If we are going to reuse semantic expression, I recommend pulling it out above or below schema for whole semantic view, this way you can reuse the same schema in other fields.

pkg/resources/semantic_view.go
```diff
@@ -355,6 +376,11 @@ var semanticViewsSchema = map[string]*schema.Schema{
 						},
 					},
 				},
+				"private_metric": {
```
sfc-gh-jcieslak, line 379: This should be within semantic expression that is shared between facts and metrics like this:
```go
var semanticViewSchema = ...

var semanticViewSemanticExpressionSchema = ... // this schema is used in the facts and metrics field
```

examples/resources/snowflake_semantic_view/resource.tf
```diff
@@ -25,20 +25,26 @@ resource "snowflake_semantic_view" "complete" {
   comment  = "comment"
 
   dimensions {
-    comment                   = "dimension comment"
-    qualified_expression_name = "\"lt1\".\"d2\""
-    sql_expression            = "\"lt1\".\"a2\""
-    synonym                   = ["dim2"]
+    semantic_expression {
```
sfc-gh-jcieslak, line 28: I don't think this nesting is necessary, internally we can call it semantic expression and share among fields, but here, I think it should be flatten so that fields like comment, synonym, etc. are defined on the dimensions level

examples/resources/snowflake_semantic_view/resource.tf
```diff
@@ -25,20 +25,26 @@ resource "snowflake_semantic_view" "complete" {
...
+      synonym                   = ["dim2"]
+    }
   }
 
   facts {
-    comment                   = "fact comment"
-    qualified_expression_name = "\"lt1\".\"f2\""
-    sql_expression            = "\"lt1\".\"a1\""
-    synonym                   = ["fact2"]
+    private_fact = "PRIVATE"
```
sfc-gh-jcieslak, line 37: I think a better field name would be e.g. `visibility`, `access_level`, `scope`, or something similar

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -173,64 +173,77 @@ func (s *SemanticViewModel) WithMetrics(metrics []sdk.MetricDefinition) *Semanti
...
+			m["private_metric"] = tfconfig.StringVariable(*privateMetric)
+		}
 		maps[i] = tfconfig.ObjectVariable(m)
 	}
 	s.Metrics = tfconfig.ListVariable(maps...)
 	return s
 }
 
-func (s *SemanticViewModel) WithFacts(facts []sdk.SemanticExpression) *SemanticViewModel {
+func (s *SemanticViewModel) WithFacts(facts []sdk.FactDefinition) *SemanticViewModel {
```
sfc-gh-jcieslak, line 186: Could facts and metrics implementation be shared somehow (as the fields they have should be the same)?

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -147,6 +147,63 @@ func TestInt_SemanticView(t *testing.T) {
...
+		// metric related details
+		metricTable := objectassert.NewSemanticViewDetailsMetric(metricName, t1Alias, "TABLE", t1Alias)
+		metricExpression := objectassert.NewSemanticViewDetailsMetric(metricName, t1Alias, "EXPRESSION", `SUM("table1"."first_a")`)
+		metricDataType := objectassert.NewSemanticViewDetailsMetric(metricName, t1Alias, "DATA_TYPE", "NUMBER(38,0)")
+		metricAccessModifier := objectassert.NewSemanticViewDetailsMetric(metricName, t1Alias, "ACCESS_MODIFIER", "PRIVATE")
+
+		assertThatObject(t, objectassert.SemanticViewDetails(t, id).
+			ContainsDetail(factTable).
+			ContainsDetail(factExpression).
+			// TODO [SNOW-2852837]: there is a currently open BCR changing the VARCHAR default size (VARCHAR(16777216) vs VARCHAR(134217728)), uncomment when generally available
```
sfc-gh-jcieslak, line 197: you can use the current default, and our tests will catch the change by failing, we'll adjust it then, but you can leave the comment so that it will be clear why it's failing

pkg/sdk/semantic_views_validations_gen.go
```diff
@@ -21,25 +21,6 @@ func (opts *CreateSemanticViewOptions) validate() error {
 	if everyValueSet(opts.IfNotExists, opts.OrReplace) {
 		errs = append(errs, errOneOf("CreateSemanticViewOptions", "IfNotExists", "OrReplace"))
 	}
-	// Adjusted manually
```
sfc-gh-jcieslak, line 24: accidental removal or doesn't apply now?

pkg/sdk/semantic_views_dto_gen.go
```diff
@@ -112,7 +112,17 @@ type SemanticSqlExpressionRequest struct {
 	SqlExpression string
 }
 
+type FactDefinitionRequest struct {
```
sfc-gh-jcieslak, line 115: Did you use the generator? I don't see any changes to `_def.go` file. Now, the generator is a bit smarter and you can choose the files it should generate making it less cumbersome to use with already present objects, please read: https://github.com/snowflakedb/terraform-provider-snowflake/blob/e0f780f33b80bf2dc208bde543e43cb926b425ca/pkg/sdk/generator/README.md#L11-L10

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: Do you mean facts and dimensions? The metrics fields are significantly different.
The facts and dimensions fields are similar right now, but they are going to diverge as we add fields specific to each of them in upcoming work. This is why I separated their implementations, to make that easier

examples/resources/snowflake_semantic_view/resource.tf
sfc-gh-sarur: ack, I will rename the field to access_modifier to match the DESCRIBE output

pkg/resources/semantic_view.go
sfc-gh-sarur: Metrics can have either semantic expressions or window functions, and they can both me PRIVATE (there is an error in the official documentation regarding this, it will be corrected asap). That's why I set this field outside.
Would it be better to move it inside the semantic expression and add it to the winfow_function field separately?

PR #4287, opened 2025-12-12
feat: Add missing CRL fields

sfc-gh-mkowalski:
Added fields from the new driver version to the provider

- DisableOCSPChecks
- CertRevocationCheckMode
- CrlAllowCertificatesWithoutCrlURL
- CrlInMemoryCacheDisabled
- CrlOnDiskCacheDisabled
- CrlHTTPClientTimeout
- DisableSamlURLCheck


pkg/provider/provider.go
```diff
@@ -527,6 +525,52 @@ func GetProviderSchema() map[string]*schema.Schema {
 			Optional:    true,
 			DefaultFunc: schema.EnvDefaultFunc(snowflakeenvs.NoProxy, nil),
 		},
+		"disable_ocsp_checks": {
+			Type:        schema.TypeBool,
+			Description: envNameFieldDescription("True by default. When set to true, the driver doesn't check certificate revocation status.", snowflakeenvs.DisableOCSPChecks),
```
sfc-gh-jcieslak: Why is it true by default? The driver has it set to false by default afaik (I checked their documentation). It seems like a breaking change.

pkg/provider/provider.go
```diff
@@ -527,6 +525,52 @@ func GetProviderSchema() map[string]*schema.Schema {
...
+		},
+		"crl_on_disk_cache_disabled": {
+			Type:        schema.TypeBool,
+			Description: envNameFieldDescription("False by default. When set to true, the CRL on-disk cache is disabled.", snowflakeenvs.CrlOnDiskCacheDisabled),
+			Optional:    true,
+			DefaultFunc: schema.EnvDefaultFunc(snowflakeenvs.CrlOnDiskCacheDisabled, nil),
+		},
+		"crl_http_client_timeout": {
+			Type:             schema.TypeInt,
+			Description:      envNameFieldDescription("Timeout for HTTP client used to download CRL.", snowflakeenvs.CrlHTTPClientTimeout),
```
sfc-gh-jcieslak: Is it in milliseconds? Let's specify what is accepted here

pkg/provider/provider.go
```diff
@@ -914,6 +957,23 @@ func getDriverConfigFromTerraform(s *schema.ResourceData) (*gosnowflake.Config,
 		handleStringField(s, "proxy_password", &config.ProxyPassword),
 		handleStringField(s, "proxy_protocol", &config.ProxyProtocol),
 		handleStringField(s, "no_proxy", &config.NoProxy),
+		handleBoolField(s, "disable_ocsp_checks", &config.DisableOCSPChecks),
+		// cert_revocation_check_mode
+		func() error {
```
sfc-gh-jcieslak: could be solved by adding a new handle function that accepts mapping function (maybe it's already there?)

pkg/sdk/config.go
```diff
@@ -177,6 +177,16 @@ func (c *ConfigDTO) DriverConfig() (gosnowflake.Config, error) {
 	pointerAttributeSet(c.ProxyPassword, &driverCfg.ProxyPassword)
 	pointerAttributeSet(c.ProxyProtocol, &driverCfg.ProxyProtocol)
 	pointerAttributeSet(c.NoProxy, &driverCfg.NoProxy)
+	pointerAttributeSet(c.DisableOCSPChecks, &driverCfg.DisableOCSPChecks)
+	err = pointerCertRevocationCheckModeAttributeSet(c.CertRevocationCheckMode, &driverCfg.CertRevocationCheckMode)
```
sfc-gh-jcieslak: let's add / use more generic functions, in this case you could pass the parsing function as last parameter

pkg/sdk/config.go
```diff
@@ -347,6 +357,27 @@ func MergeConfig(baseConfig *gosnowflake.Config, mergeConfig *gosnowflake.Config
 	if baseConfig.NoProxy == "" {
 		baseConfig.NoProxy = mergeConfig.NoProxy
 	}
+	if !baseConfig.DisableOCSPChecks {
+		baseConfig.DisableOCSPChecks = mergeConfig.DisableOCSPChecks
+	}
+	if baseConfig.CertRevocationCheckMode == CertRevocationCheckModeEmpty {
```
sfc-gh-jcieslak: shouldn't we check against empty string in this case?

pkg/sdk/config.go
```diff
@@ -590,6 +632,36 @@ func ToDriverLogLevel(s string) (DriverLogLevel, error) {
...
+	case string(CertRevocationCheckModeAdvisory):
+		return gosnowflake.CertRevocationCheckAdvisory, nil
+	case string(CertRevocationCheckModeEnabled):
+		return gosnowflake.CertRevocationCheckEnabled, nil
+	default:
+		return 0, fmt.Errorf("invalid cert revocation check mode: %s", s)
+	}
+}
+
+const CertRevocationCheckModeEmpty = gosnowflake.CertRevocationCheckDisabled
```
sfc-gh-jcieslak: I don't think this should be necessary, check other enums

pkg/provider/provider.go
```diff
@@ -375,8 +375,6 @@ func GetProviderSchema() map[string]*schema.Schema {
 			DefaultFunc:      schema.EnvDefaultFunc(snowflakeenvs.DisableConsoleLogin, resources.BooleanDefault),
 			ValidateDiagFunc: validators.ValidateBooleanStringWithDefault,
 		},
-		// TODO(SNOW-1761318): Add DisableSamlURLCheck.
-		// TODO(SNOW-1917271): Add DisableOCSPChecks.
```
sfc-gh-jcieslak, line 379: This ticket says the insecure mode should be deprecated

pkg/provider/provider.go
sfc-gh-mkowalski: It's my mistake; it's false by default.

pkg/provider/provider.go
sfc-gh-mkowalski: Added 2 new handling functions and applied them to a couple of other fields.

pkg/sdk/config.go
sfc-gh-mkowalski: Added a generic `pointerEnumSet` function.

pkg/sdk/config.go
sfc-gh-mkowalski: I think we should do it the same way as for other enums (eg. `authenticator`).

pkg/sdk/config.go
sfc-gh-mkowalski: I adjusted the naming, but other enums seem to also use the same convention, or I'm missing something.

pkg/provider/provider.go
```diff
@@ -231,6 +231,7 @@ func GetProviderSchema() map[string]*schema.Schema {
 			Type:        schema.TypeBool,
 			Description: envNameFieldDescription("If true, bypass the Online Certificate Status Protocol (OCSP) certificate revocation check. IMPORTANT: Change the default value for testing or emergency situations only.", snowflakeenvs.InsecureMode),
 			Optional:    true,
+			Deprecated:  "This field is deprecated. Use `disable_ocsp_checks` instead.",
```
sfc-gh-jcieslak, line 234: It seems the deprecaed field only is returned when the field is used, let's also modify the description, so this message is visible from the documentation pov

pkg/provider/provider_helpers.go
```diff
@@ -144,3 +144,23 @@ func handleBooleanStringAttribute(d *schema.ResourceData, key string, field *gos
 	}
 	return nil
 }
+
+func handleFieldWithConversion[T any](d *schema.ResourceData, key string, field *T, converter func(string) (T, error)) error {
```
sfc-gh-jcieslak: nit: I propose this to be called `handleFieldWithMappingIfSet` and the one below `handleFieldWithMapping`

pkg/sdk/config.go
```diff
@@ -112,7 +112,7 @@ func (c *ConfigDTO) DriverConfig() (gosnowflake.Config, error) {
 	pointerAttributeSet(c.Passcode, &driverCfg.Passcode)
 	pointerAttributeSet(c.Port, &driverCfg.Port)
 	pointerAttributeSet(c.PasscodeInPassword, &driverCfg.PasscodeInPassword)
-	err := pointerUrlAttributeSet(c.OktaUrl, &driverCfg.OktaURL)
+	err := pointerEnumSet(c.OktaUrl, &driverCfg.OktaURL, url.Parse)
```
sfc-gh-jcieslak: this is not an enum, please adjust, use already existing `pointerUrlAttributeSet` (don't remove it)

pkg/testacc/13_plugin_framework_model_and_schema_gen.go
```diff
@@ -193,9 +193,10 @@ var pluginFrameworkPocProviderSchemaV0 = map[string]schema.Attribute{
 		Sensitive:   false,
 	},
 	"insecure_mode": schema.BoolAttribute{
-		Description: existingSchema["insecure_mode"].Description,
-		Optional:    true,
-		Sensitive:   false,
+		Description:        existingSchema["insecure_mode"].Description,
+		DeprecationMessage: "This field is deprecated. Use `disable_ocsp_checks` instead.", // edited manually
```
sfc-gh-jcieslak, line 197: I don't think this is necessary, as it is only used in tests for plugin framework poc (if I remember correctly)

pkg/sdk/config.go
sfc-gh-jcieslak: Okay, looks good I think

pkg/testacc/13_plugin_framework_model_and_schema_gen.go
sfc-gh-mkowalski: It is added here to satisfy the tests.

pkg/provider/provider.go
```diff
@@ -229,8 +229,9 @@ func GetProviderSchema() map[string]*schema.Schema {
 		},
 		"insecure_mode": {
```
sfc-gh-asawicki, line 230: What's the driver's behavior if both these fields are set to some value?

sfc-gh-asawicki (dismissed):

sfc-gh-jcieslak (dismissed):

pkg/provider/provider.go
sfc-gh-mkowalski: If both are set the `disable_ocsp_checks` takes precedence; if only one is set than it works no matter which one it is.

PR #4300, opened 2025-12-17
feat: Restructure migration script's tests and add users tests

sfc-gh-mkowalski:
Restructured migration script's tests to use `terraform import` instead of `terraform apply`.

Added tests for users.

Addressed comments from the [previous PR](https://github.com/snowflakedb/terraform-provider-snowflake/pull/4271).

### DIFF
--- a/pkg/manual_tests/migration_script/users/import/provider.tf
+++ b/pkg/manual_tests/migration_script/users/import/provider.tf
@@ -5,7 +5,7 @@ terraform {
   required_providers {
     snowflake = {
       source = "snowflakedb/snowflake"
-      version = "= 2.11.0"
+      version = ">= 2.0.0"
     }
   }
 }
--- a/pkg/manual_tests/migration_script/users/objects_def.tf
+++ b/pkg/manual_tests/migration_script/users/objects_def.tf
@@ -2,6 +2,7 @@ terraform {
   required_providers {
     snowflake = {
       source = "snowflakedb/snowflake"
+      version = ">= 2.0.0"
     }
     random = {
       source = "hashicorp/random"
@@ -10,8 +11,7 @@ terraform {
 }
 
 provider "snowflake" {
-  # Configure via environment variables:
-  # SNOWFLAKE_ACCOUNT, SNOWFLAKE_USER, SNOWFLAKE_PASSWORD, etc.
+  # Uses default configuration from ~/.snowflake/config or environment variables
 }
 
 # ------------------------------------------------------------------------------
@@ -250,8 +250,8 @@ resource "snowflake_user" "long_comment" {
 resource "snowflake_user" "unicode" {
   name         = "${local.prefix}_UNICODE"
   login_name   = "${local.test_id}_unicode_login"
-  display_name = "Test User with Unicode"
-  first_name   = "Jose"
-  last_name    = "Garcia"
-  comment      = "Unicode comment test"
+  display_name = "Tëst Üsér wïth Ünïcödé"
+  first_name   = "José"
+  last_name    = "García"
+  comment      = "Unicode comment test: émojis 🎉, accents àéîõü, symbols €£¥, Chinese 中文, Japanese 日本語"
 }
--- a/pkg/scripts/migration_script/converter_users.go
+++ b/pkg/scripts/migration_script/converter_users.go
@@ -34,7 +34,9 @@ type UserCsvRow struct {
 	LockedUntilTime                               string `csv:"locked_until_time"`
 	LoginName                                     string `csv:"login_name"`
 	MiddleName                                    string `csv:"middle_name"`
+	MinsToBypassMfa                               string `csv:"mins_to_bypass_mfa"`
 	MinsToBypassNetworkPolicy                     string `csv:"mins_to_bypass_network_policy"`
+	MinsToUnlock                                  string `csv:"mins_to_unlock"`
 	MustChangePassword                            string `csv:"must_change_password"`
 	Name                                          string `csv:"name"`
 	Owner                                         string `csv:"owner"`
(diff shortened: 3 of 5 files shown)


pkg/manual_tests/migration_script/users/import/provider.tf
```diff
@@ -0,0 +1,16 @@
+# Provider configuration for import directory
+# This file is kept separate from generated main.tf
+
+terraform {
+  required_providers {
+    snowflake = {
+      source = "snowflakedb/snowflake"
+      version = "= 2.11.0"
```
sfc-gh-jcieslak: nit: in grants/objects_def it's version > 2.0.0, should we adjust here?

pkg/manual_tests/migration_script/users/objects_def.tf
```diff
@@ -0,0 +1,257 @@
...
+resource "snowflake_user" "long_comment" {
+  name       = "${local.prefix}_LONG_COMMENT"
+  login_name = "${local.test_id}_long_comment_login"
+  comment    = "This is a very long comment that tests how the migration script handles comments with many characters. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
+}
+
+# 15. User with unicode characters in names
+resource "snowflake_user" "unicode" {
+  name         = "${local.prefix}_UNICODE"
+  login_name   = "${local.test_id}_unicode_login"
```
sfc-gh-jcieslak, line 252: where unicode characters?

pkg/manual_tests/migration_script/users/objects_def.tf
```diff
@@ -0,0 +1,257 @@
...
+      source = "snowflakedb/snowflake"
+    }
+    random = {
+      source = "hashicorp/random"
+    }
+  }
+}
+
+provider "snowflake" {
+  # Configure via environment variables:
```
sfc-gh-jcieslak: Is it suggestion or requirement?

pkg/scripts/migration_script/converter_users.go
```diff
@@ -34,9 +34,7 @@ type UserCsvRow struct {
 	LockedUntilTime                               string `csv:"locked_until_time"`
 	LoginName                                     string `csv:"login_name"`
 	MiddleName                                    string `csv:"middle_name"`
-	MinsToBypassMfa                               string `csv:"mins_to_bypass_mfa"`
```
sfc-gh-jcieslak, line 37: Why these are removed?

pkg/manual_tests/migration_script/users/import/provider.tf
sfc-gh-mkowalski: It shouldn't hurt; I adjusted it to `>= 2.0.0` everywhere.

pkg/manual_tests/migration_script/users/objects_def.tf
sfc-gh-mkowalski: Just a suggestion, adjusted it to be unified in all the files.

pkg/scripts/migration_script/converter_users.go
sfc-gh-mkowalski: I reverted them, however they don't matter as they have incorrect defaults anyway, and will be showed in changes.

pkg/manual_tests/migration_script/users/import/provider.tf
sfc-gh-asawicki: I'm not sure if I like it. We base the scripts input on the specific output from the datasource which may change between minor versions (extend, not reduce). I think it's better to use the precise version here (but this is not blocking the merge).

pkg/manual_tests/migration_script/users/objects_def.tf
```diff
@@ -0,0 +1,257 @@
...
+}
+
+# 5. PERSON user with session parameters set
+resource "snowflake_user" "person_params" {
+  name         = "${local.prefix}_PERSON_PARAMS"
+  login_name   = "${local.test_id}_params_login"
+  display_name = "User with Parameters"
+
+  # Session parameters
+  abort_detached_query                          = true
```
sfc-gh-asawicki, line 91: nit (non blocking): not all parameters are present here IMO

pkg/manual_tests/migration_script/users/objects_def.tf
```diff
@@ -0,0 +1,257 @@
...
+  # <PASTE_RSA_PUBLIC_KEY_2_HERE>
+  # EOT
+}
+
+# 10. SERVICE user with parameters
+resource "snowflake_service_user" "service_params" {
+  name       = "${local.prefix}_SERVICE_PARAMS"
+  login_name = "${local.test_id}_service_params_login"
+
+  statement_timeout_in_seconds = 3600
```
sfc-gh-asawicki, line 193: nit (non blocking): why these 3 parameters?

pkg/scripts/migration_script/converter_users.go
```diff
@@ -277,10 +277,10 @@ func (row UserCsvRow) convert() (*UserRepresentation, error) {
 			FirstName:             row.FirstName,
 			LastName:              row.LastName,
 			Email:                 row.Email,
-			MinsToUnlock:          row.MinsToUnlock,
 			DaysToExpiry:          row.DaysToExpiry,
 			Comment:               csvUnescape(row.Comment),
 			Disabled:              row.Disabled == "true",
+			MinsToUnlock:          row.MinsToUnlock,
```
sfc-gh-asawicki, line 283: nit: we shouldnt change this ordering (yes, it does not matter from the end result perspective but this is not a meanigful change, and also the ordering inside the struct was aligned with this logic - supposedly)

sfc-gh-jmichalak (approved):

PR #4305, opened 2025-12-18
chore: Add note about differences between import blocks and import statements to readme

sfc-gh-mkowalski:
Added a note about differences between the behaviour of import blocks and import statements to the migration script's readme.

### DIFF
--- a/pkg/scripts/migration_script/README.md
+++ b/pkg/scripts/migration_script/README.md
@@ -39,6 +39,29 @@ However, if you're using an older version, you can still utilize the script as l
 For instance, with grants, you can use the script to transition from [old to new grants](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/guides/grants_redesign_design_decisions#mapping-from-old-grant-resources-to-the-new-ones)
 since they haven't significantly changed from the current provider version, but there may be minor differences like quotes handling in identifiers.
 
+
+### Different import behaviors
+
+> **Important:** Read before going further!
+
+The script by default outputs the [import blocks](https://developer.hashicorp.com/terraform/language/block/import). Before deciding which output to use familiarize with the limitations.
+
+The behaviour between [import blocks](https://developer.hashicorp.com/terraform/language/block/import) embedded into hcl and the [`terraform import`](https://developer.hashicorp.com/terraform/cli/commands/import) command differs.
+
+- **Import Blocks** (`-import=block`): When using embedded import blocks, the `terraform apply` command performs two actions: it imports the resource into the state and immediately applies any configuration changes.
+
+  Consequently, this method should be avoided if your primary goal is to preview changes before they are committed.
+
+  Furthermore, `terraform plan` may not provide comprehensive insights at first, as the resources have not yet been formally ingested into the state file during the planning phase.
+
+- **Import Command** (`-import=statement`): When using the `terraform import` command, the import process is decoupled from the plan/apply cycle.
+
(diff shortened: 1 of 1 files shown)


pkg/scripts/migration_script/README.md
```diff
@@ -306,6 +306,14 @@ import {
 }
 ```
 
+#### Caution
```
sfc-gh-jcieslak: 1. Because this is important, let's define it much higher
2. Make sure table of contents above is up-to-date
3. Let's give this a better header name indicating we want to warn about import choices
4. Maybe we should update CLI -h flag to also warn about it or link to this section, wdyt?

pkg/scripts/migration_script/README.md
```diff
@@ -306,6 +306,14 @@ import {
...
+
+The behaviour between import blocks embedded into hcl and the `terraform import [..]` command differs.
```
sfc-gh-jcieslak: let's link to terraform documentation on `import blocks` and `terrafrom import` command

pkg/scripts/migration_script/README.md
```diff
@@ -306,6 +306,14 @@ import {
...
+
+- **Import Blocks** (import blocks in HCL): When using embedded import blocks, the `terraform apply` command performs a dual role: it imports the resource into the state and immediately applies any configuration changes. Consequently, this method should be avoided if your primary goal is to preview changes before they are committed. Furthermore, `terraform plan` may not provide comprehensive insights at first, as the resources have not yet been formally ingested into the state file during the planning phase.
```
sfc-gh-jcieslak: 1. let's remove those `(import blocks in HCL)`, because there will be links above. Same below with `(terraform import CLI)`
2. let's break it into multiple lines if possible
3. `a dual role` could be simplified to `two actions`

pkg/scripts/migration_script/README.md
```diff
@@ -306,6 +306,14 @@ import {
...
+
+- **Import Command** (terraform import CLI): When using the standard CLI command, the import process is decoupled from the plan/apply cycle. Once the command is executed, the resource is immediately added to the state. Subsequent runs of `terraform plan` will accurately reflect the delta between your configuration and the existing infrastructure. In this workflow, `terraform apply` does not handle the ingestion; the `terraform import` command must be executed successfully beforehand.
```
sfc-gh-jcieslak: `standard CLI` -> `terraform import`

pkg/scripts/migration_script/README.md
sfc-gh-mkowalski: Added it also to the migration script's help.

pkg/scripts/migration_script/README.md
sfc-gh-jcieslak: Please, use md preview as the multilines are not rendering correctly

pkg/scripts/migration_script/README.md
sfc-gh-jcieslak: Not sure if this should be here in the main readme, let's discuss. I think it should be even higher, maybe under `Compatibility with Provider Versions` section? If so, the TOC should be updated aswell (you can use GoLand to adjust it automatically it will add other missing headers if there's more)

sfc-gh-asawicki: Id' move it to compatibility section as suggested.
Also, I would add some starter like: the script by default output s the import blocks. Before deciding which output to use familiarize with the limitations. (or something similar)

sfc-gh-jcieslak (approved):

PR #4310, opened 2025-12-18
chore: Add proxy acceptance test to the provider.

sfc-gh-mkowalski:
Added a proxy acceptance test to the provider that is based on the local proxy created with go's `httptest` library.

The proxy test uses the HTTP CONNECT method to tunnel HTTPS traffic to Snowflake. CONNECT proxies work differently from regular HTTP proxies:
1. Client sends CONNECT host:443 to the proxy
2. Proxy opens a TCP connection to the target
3. Proxy responds with 200 Connection Established
4. Proxy becomes a transparent TCP tunnel — just forwarding raw bytes in both directions
5. TLS handshake happens directly between client and Snowflake through the tunnel

The Hijacker interface is required because after step 3, the connection is no longer HTTP — it's a raw TCP stream. Hijack() releases the connection from Go's HTTP server machinery and gives us direct access to the underlying net.Conn, allowing us to perform bidirectional byte copying.

### DIFF
--- a/pkg/testacc/provider_acceptance_test.go
+++ b/pkg/testacc/provider_acceptance_test.go
@@ -1292,17 +1292,30 @@ func TestAcc_Provider_Proxy(t *testing.T) {
 
 	proxyServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		proxyUsed.Store(true)
+
+		// 1. Establish TCP connection to the target server
 		destConn, err := net.Dial("tcp", r.Host)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusBadGateway)
 			return
 		}
-		w.WriteHeader(http.StatusOK)
+		defer destConn.Close()
+
+		// 2. Hijack the client connection to get raw TCP access
 		clientConn, _, err := w.(http.Hijacker).Hijack()
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+		defer clientConn.Close()
+
+		// 3. Send "200 Connection Established" directly to the hijacked connection
+		_, err = clientConn.Write([]byte("HTTP/1.1 200 Connection Established\r\n\r\n"))
+		if err != nil {
+			return
+		}
+
+		// 4. Bidirectional copy: tunnel bytes between client and destination
 		go io.Copy(destConn, clientConn)
 		io.Copy(clientConn, destConn)
 	}))


pkg/testacc/provider_acceptance_test.go
```diff
@@ -1278,6 +1281,51 @@ func TestAcc_Provider_HandlingPromotedFeatures(t *testing.T) {
 	})
 }
 
+func TestAcc_Provider_Proxy(t *testing.T) {
+	tmpServiceUser := testClient().SetUpTemporaryServiceUser(t)
+	tmpServiceUserConfig := testClient().TempTomlConfigForServiceUser(t, tmpServiceUser)
+
+	var proxyUsed atomic.Bool
+
+	server := &http.Server{
```
sfc-gh-asawicki: These are tests, so why not httptest.NewServer (and basically httptest pkg)?

pkg/testacc/provider_acceptance_test.go
```diff
@@ -1278,6 +1281,51 @@ func TestAcc_Provider_HandlingPromotedFeatures(t *testing.T) {
...
+				return
+			}
+			w.WriteHeader(http.StatusOK)
+			clientConn, _, _ := w.(http.Hijacker).Hijack()
+			go io.Copy(destConn, clientConn)
+			io.Copy(clientConn, destConn)
+		}),
+	}
+	go server.ListenAndServe()
+	defer server.Close()
```
sfc-gh-asawicki: we should use t.Cleanup in tests instead

pkg/testacc/provider_acceptance_test.go
```diff
@@ -1278,6 +1281,51 @@ func TestAcc_Provider_HandlingPromotedFeatures(t *testing.T) {
...
+				http.Error(w, err.Error(), http.StatusBadGateway)
+				return
+			}
+			w.WriteHeader(http.StatusOK)
+			clientConn, _, _ := w.(http.Hijacker).Hijack()
+			go io.Copy(destConn, clientConn)
+			io.Copy(clientConn, destConn)
+		}),
+	}
+	go server.ListenAndServe()
```
sfc-gh-asawicki: why do we need to start it in a separate goroutine?

pkg/testacc/provider_acceptance_test.go
```diff
@@ -1278,6 +1281,51 @@ func TestAcc_Provider_HandlingPromotedFeatures(t *testing.T) {
...
+	tmpServiceUser := testClient().SetUpTemporaryServiceUser(t)
+	tmpServiceUserConfig := testClient().TempTomlConfigForServiceUser(t, tmpServiceUser)
+
+	var proxyUsed atomic.Bool
+
+	server := &http.Server{
+		Addr: "127.0.0.1:8070",
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			proxyUsed.Store(true)
+			destConn, err := net.Dial("tcp", r.Host)
```
sfc-gh-asawicki: do we need all this following logic? what would be the bare minimum just to redirect?

pkg/testacc/provider_acceptance_test.go
```diff
@@ -1278,6 +1281,51 @@ func TestAcc_Provider_HandlingPromotedFeatures(t *testing.T) {
...
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				PreConfig: func() {
+					t.Setenv(snowflakeenvs.ConfigPath, tmpServiceUserConfig.Path)
+				},
+				Config: config.FromModels(t, providermodel.SnowflakeProvider().WithProfile(tmpServiceUserConfig.Profile).WithProxyHost("127.0.0.1").WithProxyPort(8070).WithProxyProtocol("http").WithNoProxy(""), datasourceModel()),
+				Check: func(state *terraform.State) error {
```
sfc-gh-asawicki: let's also add a check for the datasource

sfc-gh-asawicki (changes requested):

pkg/testacc/provider_acceptance_test.go
```diff
@@ -1278,6 +1281,51 @@ func TestAcc_Provider_HandlingPromotedFeatures(t *testing.T) {
...
+			}
+			w.WriteHeader(http.StatusOK)
+			clientConn, _, _ := w.(http.Hijacker).Hijack()
+			go io.Copy(destConn, clientConn)
+			io.Copy(clientConn, destConn)
+		}),
+	}
+	go server.ListenAndServe()
+	defer server.Close()
+	time.Sleep(100 * time.Millisecond)
```
sfc-gh-jcieslak: Use assert.Eventually if possible

pkg/testacc/provider_acceptance_test.go
```diff
@@ -1278,6 +1281,51 @@ func TestAcc_Provider_HandlingPromotedFeatures(t *testing.T) {
...
+		ProtoV6ProviderFactories: providerFactoryWithoutCache(),
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				PreConfig: func() {
+					t.Setenv(snowflakeenvs.ConfigPath, tmpServiceUserConfig.Path)
+				},
+				Config: config.FromModels(t, providermodel.SnowflakeProvider().WithProfile(tmpServiceUserConfig.Profile).WithProxyHost("127.0.0.1").WithProxyPort(8070).WithProxyProtocol("http").WithNoProxy(""), datasourceModel()),
```
sfc-gh-jcieslak: Do we have to create another user if we are already using provider without cache?

pkg/testacc/provider_acceptance_test.go
sfc-gh-mkowalski: We do not need it

pkg/testacc/provider_acceptance_test.go
sfc-gh-mkowalski: Simplified the logic to stop unnecessarily validating the connection.

pkg/testacc/provider_acceptance_test.go
sfc-gh-mkowalski: Used `httptest` instead of sleep.

pkg/testacc/provider_acceptance_test.go
sfc-gh-mkowalski: I'm getting `account is empty` error when I'm trying to test without setting up a temporary user.

pkg/testacc/provider_acceptance_test.go
```diff
@@ -1278,6 +1284,72 @@ func TestAcc_Provider_HandlingPromotedFeatures(t *testing.T) {
 	})
 }
 
+func TestAcc_Provider_Proxy(t *testing.T) {
+	tmpServiceUser := testClient().SetUpTemporaryServiceUser(t)
+	tmpServiceUserConfig := testClient().TempTomlConfigForServiceUser(t, tmpServiceUser)
+
+	var proxyUsed atomic.Bool
+
+	proxyServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
sfc-gh-asawicki, line 1293: Let's in the future:
- extract this proxy server to be reusable
- track more (not only a single atomic bool, but e.g. certain requests)
- add more tests with more config options (e.g. HTTPS_PROXY env + tf config)

sfc-gh-asawicki (approved):

PR #4318, opened 2025-12-22
fix: user external change detection

sfc-gh-jcieslak:
## Changes
- Adjust user reading function and test for new change
- Add new migration guide entry

Parameters that are read outside of the `show_output` reading model (because they are coming from `desc` command) are not detecting external changes correctly for string parameters in the following cases:
In state: "value"
In Snowflake null or ""

> The above change won't be detected until the Snowflake value is null or empty

The change includes reading values that are null or empty, which enables external change detection for those values.


sfc-gh-jmichalak (approved):
TestAcc_GrantOwnership_OnTask_Discussion2877 is failing - not related, but we can provide the task object before.

MIGRATION_GUIDE.md
```diff
@@ -26,6 +26,38 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ## v2.13.x ➞ v2.14.0
 
+### *(bugfix)* Fixed external change detection in user resources (`snowflake_user`, `snowflake_service_user`, and `snowflake_legacy_service_user`)
```
sfc-gh-jmichalak, line 29: Hmm, it's a new feature, actually :D

pkg/resources/resource_helpers_read.go
```diff
@@ -19,11 +19,15 @@ func setFromStringProperty(d *schema.ResourceData, key string, property *sdk.Str
 		if err := d.Set(key, property.Value); err != nil {
 			return err
 		}
+	} else {
```
sfc-gh-jmichalak, line 22: Q: what about setFromIntProperty, should we set 0/-1 in such case?

pkg/resources/user.go
```diff
@@ -283,9 +283,9 @@ func GetImportUserFunc(userType sdk.UserType) func(ctx context.Context, d *schem
 
 		err = errors.Join(
 			d.Set("name", id.Name()),
-			setFromStringPropertyIfNotEmpty(d, "login_name", userDetails.LoginName),
-			setFromStringPropertyIfNotEmpty(d, "display_name", userDetails.DisplayName),
-			setFromStringPropertyIfNotEmpty(d, "default_namespace", userDetails.DefaultNamespace),
+			setFromStringPropertyIfNotNullOrEmpty(d, "login_name", userDetails.LoginName),
```
sfc-gh-jmichalak, line 286: Q: These use a different function. Is it because they are always non-empty, or something else? If so, this should be adjusted in the sdk.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Hmm true

pkg/resources/resource_helpers_read.go
sfc-gh-jcieslak: Well, maybe. It's used in one resource (`password_policy`). I'll add it to my list, if I won't have time to test it, I'll modify the stabilization of password policy ticket

pkg/resources/user.go
sfc-gh-jcieslak: Hmm, good point, it's in import, which means both functions would be handled by import in the same way (I think). Will test it and provide a follow-up it it would matter.

PR #4340, opened 2026-01-07
fix: Preserve state after failed schema updates due to insufficient privileges

sfc-gh-jmichalak:
Fixed a bug where the schema resource would enter a corrupted state if an update failed due to insufficient privileges.
- Updated the schema resource to use the `d.Partial` method for all SQL operations during updates, ensuring the state is preserved correctly after failures.
- Added a migration guide entry detailing the fix and instructions for recovering from a corrupted state.
- Introduced a new acceptance test to verify the correct behavior when updates fail.
- Removed failing grants tests.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -29,7 +29,7 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 ## *(bugfix)* Fixed broken state after errors in `terraform apply` in the schema resource
 Previously, when the schema's `with_managed_access` value was changed during the apply, and the Terraform role did not have sufficient privileges, the operation resulted in a corrupted state. The value of such a field was set to `true` in the state, even though the operation returned an error. This behavior could also happen in other fields.
 
-In this release, this bug has been fixed. Now, all SQL operations during the update use the `d.Partial` method. After failing Terraform operations, the state should be preserved correctly.
+In this release, this bug has been fixed. After failing Terraform operations, the state should be preserved correctly.
 
 If you previously ended up in a corrupted state, you can remove the resource from the state and reimport it using `terraform import`.
 


pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -2713,21 +2703,11 @@ func TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_Validation_Confli
 				PlanOnly:    true,
 				ExpectError: regexp.MustCompile(`"strict_privilege_management": conflicts with`),
 			},
-			{
-				Config:      accconfig.FromModels(t, providerModel, resourceModelOnSchemaFuture),
```
sfc-gh-asawicki, line 2717: why we are removing them? Were they passing before? Is this change influencing them?

MIGRATION_GUIDE.md
```diff
@@ -24,6 +24,17 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.12.0 ➞ v2.12.1
+
+## *(bugfix)* Fixed broken state after errors in `terraform apply` in the schema resource
+Previously, when the schema's `with_managed_access` value was changed during the apply, and the Terraform role did not have sufficient privileges, the operation resulted in a corrupted state. The value of such a field was set to `true` in the state, even though the operation returned an error. This behavior could also happen in other fields.
+
+In this release, this bug has been fixed. Now, all SQL operations during the update use the `d.Partial` method. After failing Terraform operations, the state should be preserved correctly.
```
sfc-gh-asawicki: I don;t think that this really deep implementation detail of using d.Partial should be included in the migration guide.

sfc-gh-jmichalak: Adjusted.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jmichalak: Because these validations no longer exist.

sfc-gh-asawicki: But you are not removing them, so:
1. Were they passing before?
2. Is this change influencing them?

sfc-gh-jmichalak: They were removed in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4296/files

PR #4343, opened 2026-01-08
chore: Use UNSET for certain fields in warehouses

sfc-gh-jmichalak:
Previously, Snowflake didn't support `UNSET` for `scaling_policy`, `auto_resume`, and `warehouse_type` in warehouses. As a workaround, the provider used `SET` with default values.
Now, `UNSET` is available in Snowflake, and the provider uses this operation for these fields.
Note that SNOW-1473453 is still open because auto_suspend is still not working as desired.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -34,6 +34,15 @@ Note: `auto_suspend` still uses `SET` with the default value (600) as a workarou
 
 No changes in the configuration is required.
 
+## *(bugfix)* Fixed broken state after errors in `terraform apply` in the schema resource
+Previously, when the schema's `with_managed_access` value was changed during the apply, and the Terraform role did not have sufficient privileges, the operation resulted in a corrupted state. The value of such a field was set to `true` in the state, even though the operation returned an error. This behavior could also happen in other fields.
+
+In this release, this bug has been fixed. After failing Terraform operations, the state should be preserved correctly.
+
+If you previously ended up in a corrupted state, you can remove the resource from the state and reimport it using `terraform import`.
+
+No changes in configuration are required.
+
 ## v2.11.x ➞ v2.12.0
 
 ### *(new feature)* The new `strict_privilege_management` flag in the `snowflake_grant_privileges_to_account_role` resource
--- a/pkg/acceptance/bettertestspoc/config/model/grant_privileges_to_account_role_model_ext.go
+++ b/pkg/acceptance/bettertestspoc/config/model/grant_privileges_to_account_role_model_ext.go
@@ -34,6 +34,13 @@ func (g *GrantPrivilegesToAccountRoleModel) WithOnFutureSchemasInDatabase(id sdk
 	return g
 }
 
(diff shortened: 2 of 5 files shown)


pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -669,70 +672,6 @@ func TestInt_Warehouses(t *testing.T) {
 		require.ErrorContains(t, err, "invalid property combination 'RESOURCE_CONSTRAINT'='STANDARD_GEN_1' and 'GENERATION'='2'")
 	})
 
-	t.Run("alter: prove problems with unset auto suspend", func(t *testing.T) {
```
sfc-gh-asawicki, line 672: From what we discussed the auto suspend is still unsetting to 0 instead of 600, right?

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -337,6 +336,10 @@ func TestInt_Warehouses(t *testing.T) {
 				Comment:                         sdk.Bool(true),
 				EnableQueryAcceleration:         sdk.Bool(true),
 				QueryAccelerationMaxScaleFactor: sdk.Bool(true),
+				WarehouseType:                   sdk.Bool(true),
+				ScalingPolicy:                   sdk.Bool(true),
+				AutoSuspend:                     sdk.Bool(true),
```
sfc-gh-asawicki: AutoSuspend should still go through SET

MIGRATION_GUIDE.md
```diff
@@ -24,6 +24,14 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.12.0 ➞ v2.12.1
+
+### *(improvement)* Using UNSET for certain fields in warehouses
+Previously, Snowflake didn't support `UNSET` for `scaling_policy`, `auto_resume`, and `warehouse_type` in warehouses. As a workaround, the provider used `SET` with default values.
+Now, `UNSET` is available in Snowflake, and the provider uses this operation for these fields.
```
sfc-gh-jcieslak, line 31: nit: maybe we should list those fields that are still going through set when unset should be called (e.g., auto_suspend)?

sfc-gh-jcieslak (dismissed):

MIGRATION_GUIDE.md
sfc-gh-jmichalak: Done.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jmichalak: Yes.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jmichalak: In the resource, we use set still.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-asawicki: The test was reverted, closing.

Issue #3942, opened 2025-08-15
[Feature]: Support Workload identity federation


nakamichiworks:
### Use Cases or Problem Statement

We want to use [workload identity federation](https://docs.snowflake.com/en/user-guide/workload-identity-federation) for more secure authentication of service users, since majority of our applications are running on AWS.

### Category

category:resource

### Object type(s)

resource:service_user

### Proposal

[Workload identity federation](https://docs.snowflake.com/en/user-guide/workload-identity-federation) is now generally available, so please support related parameters in `snowflake_service_user` resource.

### How much impact is this issue causing?

Low

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
Hi @nakamichiworks 👋 

This feature is on our [roadmap](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/ROADMAP.md#oauth2-authorization-code-and-client-credential-flows). But now, we are focused on delivering new OAuth capabilities.

aladeeb:
<img width="1045" height="159" alt="Image" src="https://github.com/user-attachments/assets/a89a3513-898d-41cd-b41a-4e7fbb565c2c" />

Any tentative date when this will be supported? We are migrating the whole org platform codebase to Terraform and this one would be a dealbreaker for us. Appreciate your response, it will help us to plan the migration. :)

sfc-gh-asawicki:
Hey @aladeeb. We are not actively working on it. The soonest we will start WIF-related work is November. The date may vary depending on the complexity of the implementation. We are pretty confident it will be available in December.

sfc-gh-jmichalak:
This issue also tracks support for the WIF authenticator in the provider configuration (moved from #4012).

TimoVink:
I'm currently using this workaround, which seems to work for my purposes. I have not used `snowflake_execute` before so there may be some nuances I'm missing - open to suggestions 😊

```terraform
# Input variables:
#    user_name (str): The name to use for the Snowflake USER
#    iam_role_arn (str): The ARN of the IAM role that should be able to authenticate as our new USER

resource "snowflake_service_user" "this" {
  name = var.user_name
  # ... rest of your config goes here ...
}

resource "snowflake_execute" "workload_identity_federation" {
  execute = <<-SQL
    ALTER USER ${snowflake_service_user.this.name}
      SET WORKLOAD_IDENTITY = (
[...]

PR #4367, opened 2026-01-15
feat: WIF support for legacy service and service users

sfc-gh-jmichalak:
Adds `default_workload_identity` block to `snowflake_service_user` and `snowflake_legacy_service_user` resources, enabling passwordless authentication via cloud workload identities.

Supported providers
- **AWS** - IAM role ARN
- **GCP** - Service account subject  
- **Azure** - Issuer + subject
- **OIDC** - Generic OIDC with optional audience list

## References
#3942


pkg/acceptance/bettertestspoc/assert/objectassert/user_workload_identity_authentication_methods_snowflake_gen.go
```diff
@@ -9,11 +9,12 @@ import (
 	"time"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 )
 
 type UserWorkloadIdentityAuthenticationMethodsAssert struct {
-	*assert.SnowflakeObjectAssert[sdk.UserWorkloadIdentityAuthenticationMethod, sdk.AccountObjectIdentifier]
+	*assert.SnowflakeObjectAssert[sdk.UserWorkloadIdentityAuthenticationMethod, helpers.UserWorkloadIdentityAuthenticationMethodsObjectIdentifier]
```
sfc-gh-asawicki, line 18: was this one adjusted manually? should it be changed in `sdk_object_def.go` instead? If not, then pls at least mark as manually adjusted.

pkg/testacc/resource_service_user_wif_acceptance_test.go
```diff
@@ -0,0 +1,678 @@
...
+						HasOidcAdditionalInfo(sdk.UserWorkloadIdentityAuthenticationMethodsOidcAdditionalInfo{
+							Issuer:       testvars.OidcIssuer,
+							Subject:      subject,
+							AudienceList: []string{testvars.OidcIssuer + "/o/oauth2/auth"},
+						}),
+				),
+			},
+		},
+	})
+}
```
sfc-gh-asawicki, line 646: Let's add a test where:
1. 2.12.0 is used to create the user.
2. We "externally" set the default wif.
3. We use the newest version and expect that wif won't be removed. - This expectation is aligned with the `No changes in configuration are required for existing service users` part of migration guide. If the behavior is different, please adjust the migration guide.

pkg/testacc/resource_service_user_wif_acceptance_test.go
```diff
@@ -0,0 +1,678 @@
...
+							Issuer:       testvars.OidcIssuer,
+							Subject:      subject,
+							AudienceList: []string{testvars.OidcIssuer + "/o/oauth2/auth"},
+						}),
+				),
+			},
+			// ALTER - CHANGE AUDIENCE LIST EXTERNALLY
+			{
+				PreConfig: func() {
+					testClient().User.Alter(t, id, &sdk.AlterUserOptions{
```
sfc-gh-asawicki: let's extract this method to user test client and reuse. The same with other similar methods in this file.

pkg/testacc/resource_service_user_wif_acceptance_test.go
```diff
@@ -0,0 +1,678 @@
...
+					},
+				},
+				// Terraform should detect the type change and re-apply OIDC config
+				Check: assertThat(t,
+					resourceassert.ServiceUserResource(t, userModelWithOIDC.ResourceReference()).
+						HasNameString(id.Name()).
+						HasDefaultWorkloadIdentityOidc(testvars.OidcIssuer, subject, testvars.OidcIssuer+"/o/oauth2/auth"),
+					resourceshowoutputassert.UserShowOutput(t, userModelWithOIDC.ResourceReference()).
+						HasHasWorkloadIdentity(true),
+					objectassert.UserWorkloadIdentityAuthenticationMethods(t, id, "DEFAULT").
```
sfc-gh-asawicki: nit: for now it's only `"DEFAULT"`, right? We could have a helper hiding this detail.

pkg/resources/user_commons.go
```diff
@@ -3,9 +3,128 @@ package resources
...
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"arn": {
+								Type:        schema.TypeString,
+								Required:    true,
+								Description: "The ARN of the AWS IAM role to use for workload identity federation.",
+							},
+						},
+					},
+					ExactlyOneOf: []string{
```
sfc-gh-asawicki, line 39: I may have missed it but I think I havent seen the validation test for these?

pkg/resources/user.go
```diff
@@ -717,3 +759,108 @@ var DeleteUser = ResourceDeleteContextFunc(
 	sdk.ParseAccountObjectIdentifier,
 	func(client *sdk.Client) DropSafelyFunc[sdk.AccountObjectIdentifier] { return client.Users.DropSafely },
 )
+
+// parseWorkloadIdentityConfig parses the default_workload_identity block from ResourceData
+// and returns the SDK representation for use in Create/Update operations.
+func parseWorkloadIdentityConfig(v any) (sdk.UserObjectWorkloadIdentityProperties, error) {
```
sfc-gh-asawicki, line 778: it never return an error and we have a logic on the calling side to handle the error:
1. Should we omit this error?
2. Should we return error as the last return or in a different case?

pkg/resources/user.go
```diff
@@ -521,6 +538,26 @@ func GetReadUserFunc(userType sdk.UserType, withExternalChangesMarking bool) sch
...
+					}
+
+					defaultWIF, err := collections.FindFirst(wifMethods, func(m sdk.UserWorkloadIdentityAuthenticationMethod) bool {
+						return m.Name == "DEFAULT"
+					})
+					switch {
+					case errors.Is(err, collections.ErrObjectNotFound):
+						errs = errors.Join(errs, d.Set("default_workload_identity", nil))
+						// TODO(SNOW-3003261): Handle AWS type externally
+					case err == nil && defaultWIF.Type != sdk.WIFTypeAWS:
```
sfc-gh-asawicki, line 561: what if err ==nil && defaultWIF.Type == sdk.WIFTypeAWS (I mean from the business perspective, currently we will return nil)?

sfc-gh-asawicki (changes requested):

pkg/acceptance/bettertestspoc/assert/objectassert/user_workload_identity_authentication_methods_snowflake_gen.go
sfc-gh-jmichalak: Added a note.

pkg/testacc/resource_service_user_wif_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/user.go
sfc-gh-jmichalak: Added a note - it's done for compatibility. All the validations are set in the schema.

pkg/testacc/resource_service_user_wif_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/user_commons.go
sfc-gh-jmichalak: Added.

pkg/resources/user.go
sfc-gh-jmichalak: Discussed offline and decided to leave this behavior, but hide the whole feature behind a feature flag.

pkg/testacc/resource_service_user_wif_acceptance_test.go
sfc-gh-jmichalak: Discussed offline. We decided to place this behind a feature flag. Added acceptance tests and updated the migrationg uide.

pkg/acceptance/helpers/user_client.go
```diff
@@ -260,3 +260,110 @@ func (c *UserClient) ShowUserWorkloadIdentityAuthenticationMethodOptions(t *test
...
+func (c *UserClient) SetOidcWorkloadIdentity(t *testing.T, userId sdk.AccountObjectIdentifier, issuer, subject string, audienceList ...string) {
+	t.Helper()
+	ctx := context.Background()
+
+	audiences := make([]sdk.StringListItemWrapper, len(audienceList))
+	for i, v := range audienceList {
+		audiences[i] = sdk.StringListItemWrapper{Value: v}
+	}
+
+	err := c.client().Alter(ctx, userId, &sdk.AlterUserOptions{
```
sfc-gh-asawicki, line 274: nit: the common part could be further extracted (in the dedicated funcs create given subtype and modify  UserObjectWorkloadIdentityProperties and pass it to the common part) - but let's not do it now

pkg/resources/user.go
```diff
@@ -576,7 +583,11 @@ func GetReadUserFunc(userType sdk.UserType, withExternalChangesMarking bool) sch
 
 func GetUpdateUserFunc(userType sdk.UserType) func(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	return func(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-		client := meta.(*provider.Context).Client
+		providerCtx := meta.(*provider.Context)
+		client := providerCtx.Client
+		if ok := d.HasChange("default_workload_identity"); ok && !experimentalfeatures.IsExperimentEnabled(experimentalfeatures.UserEnableDefaultWorkloadIdentity, providerCtx.EnabledExperiments) {
+			return diag.Errorf("to use `default_workload_identity`, you need to first specify the `USER_ENABLE_DEFAULT_WORKLOAD_IDENTITY` feature in the `experimental_features_enabled` field at the provider level")
```
sfc-gh-asawicki, line 589: nit: we could extract helper for this error

JESCHO99:
Hey @sfc-gh-asawicki, do you alread have an estimation when a new version of the provider will be released which will add this functionality? Would be great to get this information before starting to implement a workaround until it is released. Thanks a lot.

sfc-gh-jmichalak:
Hi @JESCHO99, we're planning to release it later this week.

PR #4368, opened 2026-01-15
chore: Propose usage of service users in migration tests

sfc-gh-jmichalak:
We want to replace legacy service users with service users for old provider versions (<0.97.0). The problem is, the old versions don't support setting private keys in TOML. That's why in this proposal, we use a provider config with values sourced from TF variables. We can't use the regular `SNOWFLAKE_...` envs here to avoid clashes with provider config acceptance tests.

TODO (next PRs)
- Use this setup in all other tests using the legacy service.
- Remove the previous v097 compatible config from GH, and SetV097CompatibleConfigPathEnv from the test helpers.
- Remove legacy service users on our envs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -111,11 +111,11 @@ jobs:
           chmod 0600 ./config ./config_legacy ./config_v097_compatible ./config_v097_compatible_with_service_user
 
       - name: Run unit tests
-        run: echo ok
+        run: make test-unit
         if: ${{ !cancelled() && steps.create_config.conclusion == 'success' }}
 
       - name: Run integration tests
-        run: echo ok
+        run: make test-integration
         if: ${{ !cancelled() && steps.create_config.conclusion == 'success' }}
         env:
           SNOWFLAKE_BUSINESS_CRITICAL_ACCOUNT: ${{ secrets.SNOWFLAKE_BUSINESS_CRITICAL_ACCOUNT }}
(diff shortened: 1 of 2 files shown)


pkg/acceptance/testvars/provider.go
```diff
@@ -0,0 +1,22 @@
+package testvars
+
+func V097CompatibleProvider() string {
```
sfc-gh-asawicki, line 3: nit: it should be fully possible to model this using the existing model builders (take a look at  `SecretStringVariableModelWithConfigVariables` and `config.VariableReference`) and place it under `providermodel` pkg

.github/workflows/tests.yml
```diff
@@ -152,6 +153,9 @@ jobs:
           TEST_SF_TF_AZURE_EXTERNAL_SAS_TOKEN: ${{ secrets.TEST_SF_TF_AZURE_EXTERNAL_SAS_TOKEN }}
           TEST_SF_TF_AZURE_EXTERNAL_TENANT_ID: ${{ secrets.TEST_SF_TF_AZURE_EXTERNAL_TENANT_ID }}
           TEST_SF_TF_GCS_EXTERNAL_BUCKET_URL: ${{ secrets.TEST_SF_TF_GCS_EXTERNAL_BUCKET_URL }}
+          TF_VAR_V097_COMPATIBLE_PRIVATE_KEY: ${{ secrets.TF_VAR_V097_COMPATIBLE_PRIVATE_KEY }}
```
sfc-gh-asawicki, line 156: q: do I understand correctly that these wiul need to be additional secrets in our pipeline, if so, let's chat

sfc-gh-asawicki (changes requested):

.github/workflows/tests.yml
sfc-gh-jmichalak: Discussed.

pkg/acceptance/testvars/provider.go
sfc-gh-jmichalak: This introduces an import cycle, so I'll leave it this way.

sfc-gh-asawicki: Discussed, leaving it open

PR #4378, opened 2026-01-21
chore: Use a new file format in stages SDK

sfc-gh-jmichalak:
- Introduce new StageFileFormatRequest struct and generated SDK files for improved file format handling in stage operations
- Rename FileFormats interface and implementation to LegacyFileFormats to prepare for future API improvements
- Rename FileFormatTypeOptions to LegacyFileFormatTypeOptions for consistency
- Update stage definitions to use new StageFileFormatRequest with proper validation

### DIFF
--- a/pkg/internal/genhelpers/struct_details_extractor_test.go
+++ b/pkg/internal/genhelpers/struct_details_extractor_test.go
@@ -114,7 +114,7 @@ func Test_ExtractStructDetails(t *testing.T) {
 		assertFieldExtracted(structDetails.Fields[29], "unexportedIdentifierSlice", "[]sdk.SchemaObjectIdentifier", "slice")
 
 		assertFieldExtracted(structDetails.Fields[30], "unexportedInterface", "sdk.ObjectIdentifier", "interface")
-		assertFieldExtracted(structDetails.Fields[31], "unexportedStruct", "sdk.FileFormatTypeOptions", "struct")
+		assertFieldExtracted(structDetails.Fields[31], "unexportedStruct", "sdk.LegacyFileFormatTypeOptions", "struct")
 	})
 }
 


pkg/sdk/generator/defs/file_formats_def.go
```diff
@@ -0,0 +1,137 @@
...
+
+func fileFormatDef() *g.QueryStruct {
+	return g.NewQueryStruct("FileFormatOptions").
+		OptionalQueryStructField(
+			"CsvOptions",
+			g.NewQueryStruct("FileFormatCsvOptions").
+				PredefinedQueryStructField("formatType", "string", g.StaticOptions().SQL("TYPE = CSV")).
+				OptionalAssignment("COMPRESSION", g.KindOfTPointer[sdkcommons.CsvCompression](), g.ParameterOptions().NoQuotes()).
+				OptionalQueryStructField("RecordDelimiter", stageFileFormatStringOrNone(), g.ListOptions().NoParentheses().SQL("RECORD_DELIMITER =")).
+				OptionalQueryStructField("FieldDelimiter", stageFileFormatStringOrNone(), g.ListOptions().NoParentheses().SQL("FIELD_DELIMITER =")).
```
sfc-gh-asawicki, line 30: We need to tackle definitions like `.SQL("FIELD_DELIMITER =")` (containing equals sign in SQL). I don't want to block this PR as there are more places in our codebase where we use this hack. I will add this to our internal sdk generation improvements tracker.

pkg/sdk/file_formats_validations_gen.go
```diff
@@ -0,0 +1,17 @@
...
+var _ validatable = new(DummyOperationFileFormatOptions)
+
+func (opts *DummyOperationFileFormatOptions) validate() error {
+	if opts == nil {
+		return ErrNilOptions
+	}
+	var errs []error
+	// adjusted manually
+	if valueSet(opts.FileFormat) {
+		errs = append(errs, opts.FileFormat.validate())
```
sfc-gh-asawicki, line 14: why the whole validate was extracted in the ext file?

sfc-gh-jmichalak: Mainly because this method validates the whole `DummyOperationFileFormatOptions`, but we want to validate only the `FileFormat` struct. On the other hand, we could leave the validation code here (so, no manual changes in the end) and have a duplicate in `ext`.

pkg/sdk/generator/defs/file_formats_def.go
sfc-gh-jmichalak: 👍

sfc-gh-asawicki: Added to generator readme (locally, I will push it with other changes)

PR #4380, opened 2026-01-22
feat: support private facts and metrics in semantic views

sfc-gh-sarur:
Adding support for private facts and metrics.

## Test Plan

* [x] acceptance tests
* [x] integration tests


pkg/testacc/data_source_semantic_views_acceptance_test.go
sfc-gh-sarur, line 1: there seems to be an issue with the data source tests, I could not run them on the dev branch either.

sfc-gh-jmichalak:
Hi @sfc-gh-sarur  👋 
Thanks for your contribution! I'll review it on Monday.

sfc-gh-sarur:
Hi @sfc-gh-jmichalak ! Did you get a chance to review this PR yet?

docs/resources/semantic_view.md
```diff
@@ -64,13 +64,15 @@ resource "snowflake_semantic_view" "complete" {
   }
 
   facts {
+    is_private                = true
```
sfc-gh-jmichalak: Please also add an entry to the migration guide.

pkg/sdk/generator/defs/semantic_views_def.go
```diff
@@ -55,8 +55,8 @@ var semanticViewsDef = g.NewInterface(
 			Name().
 			PredefinedQueryStructField("logicalTables", "[]LogicalTable", g.ParameterOptions().Required().Parentheses().NoEquals().SQL("TABLES")).
 			PredefinedQueryStructField("semanticViewRelationships", "[]SemanticViewRelationship", g.ParameterOptions().Parentheses().NoEquals().SQL("RELATIONSHIPS")).
-			PredefinedQueryStructField("semanticViewFacts", "[]SemanticExpression", g.ParameterOptions().Parentheses().NoEquals().SQL("FACTS")).
-			PredefinedQueryStructField("semanticViewDimensions", "[]SemanticExpression", g.ParameterOptions().Parentheses().NoEquals().SQL("DIMENSIONS")).
+			PredefinedQueryStructField("semanticViewFacts", "[]FactDefinition", g.ParameterOptions().Parentheses().NoEquals().SQL("FACTS")).
```
sfc-gh-jmichalak, line 58: Why are these two changed?

pkg/sdk/semantic_views_ext.go
```diff
@@ -58,6 +66,30 @@ func (m *MetricDefinition) SetWindowFunctionMetricDefinition(windowFunc *WindowF
 	m.windowFunctionMetricDefinition = windowFunc
 }
 
+func (f *FactDefinition) GetIsPrivate() *bool {
+	return f.isPrivate
+}
+
+func (f *FactDefinition) SetIsPrivate(isPrivate *bool) {
```
sfc-gh-jmichalak: Please remove the Set... methods here. In dto_builders_gen, you have autogenerated `With...` methods.

pkg/sdk/testint/semantic_view_integration_test.go
```diff
@@ -147,6 +147,63 @@ func TestInt_SemanticView(t *testing.T) {
 		)
 	})
 
+	t.Run("create: with private facts and metrics", func(t *testing.T) {
```
sfc-gh-jmichalak: This should be tested in one of the already existing cases - feel free to just add the PRIVATE field to one of the requests, and add an appropriate assertion.

pkg/testacc/data_source_semantic_views_acceptance_test.go
```diff
@@ -33,7 +33,7 @@ func TestAcc_SemanticViews_Basic(t *testing.T) {
 	logicalTable1 := model.LogicalTableWithProps("lt1", table1.ID(), []sdk.SemanticViewColumn{{Name: "a1"}}, [][]sdk.SemanticViewColumn{{{Name: "a2"}}, {{Name: "a3"}, {Name: "a4"}}}, nil, "logical table 1")
 	semExp1 := model.SemanticExpressionWithProps(`"lt1"."se1"`, `SUM("lt1"."a1")`, []sdk.Synonym{{Synonym: "sem1"}, {Synonym: "baseSem"}}, "semantic expression 1")
 
-	metric1 := model.MetricDefinitionWithProps(semExp1, nil)
+	metric1 := model.MetricDefinitionWithProps(semExp1, nil, false)
```
sfc-gh-jmichalak: This function should not be changed. Instead, the new parameter should be added like `WithPrivate...`.

pkg/testacc/data_source_semantic_views_acceptance_test.go
sfc-gh-jmichalak: They are passing on dev branch. If you have any errors, please paste them here.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -303,17 +309,37 @@ func WindowFunctionMetricDefinitionWithProps(
 	return windowFunctionMetricDefinition
 }
 
-func MetricDefinitionWithProps(semExp *sdk.SemanticExpression, windowFunc *sdk.WindowFunctionMetricDefinition) *sdk.MetricDefinition {
+func MetricDefinitionWithProps(semExp *sdk.SemanticExpression, windowFunc *sdk.WindowFunctionMetricDefinition, isPrivate bool) *sdk.MetricDefinition {
```
sfc-gh-jmichalak: Please remove these funcs with isPrivate and DimensionDefinitionWithProps. Use WithIsPrivate directly in tests.

pkg/resources/semantic_view.go
```diff
@@ -182,6 +182,12 @@ var semanticViewsSchema = map[string]*schema.Schema{
 		Description: externalChangesNotDetectedFieldDescription("The list of facts in the semantic view."),
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
+				"is_private": {
+					Type:        schema.TypeBool,
```
sfc-gh-jmichalak: Use our stringBool solution, check other resources like saml2's "enabled' field.

pkg/resources/semantic_view.go
```diff
@@ -620,7 +631,8 @@ func getMetricDefinitionRequest(from any) (*sdk.MetricDefinitionRequest, error)
 				semExpRequest = semExpRequest.WithSynonyms(sRequest)
 			}
 		}
-		return metricDefinitionRequest.WithSemanticExpression(*semExpRequest), nil
+		return metricDefinitionRequest.WithSemanticExpression(*semExpRequest).
+			WithIsPrivate(c["is_private"].(bool)), nil
```
sfc-gh-jmichalak: WithIsPrivate should be used only if it's set. Check `booleanStringAttributeCreate`

pkg/resources/semantic_view.go
```diff
@@ -644,10 +656,86 @@ func getMetricDefinitionRequest(from any) (*sdk.MetricDefinitionRequest, error)
...
-		return metricDefinitionRequest.WithWindowFunctionMetricDefinition(*windowFuncRequest), nil
+		return metricDefinitionRequest.WithWindowFunctionMetricDefinition(*windowFuncRequest).
+			WithIsPrivate(c["is_private"].(bool)), nil
 	default:
-		return nil, fmt.Errorf("either semantic expression or window function is required")
+		return nil, fmt.Errorf("either semantic expression or window function is required in metric definition")
+	}
+}
+
+func getFactDefinitionRequest(from any) (*sdk.FactDefinitionRequest, error) {
```
sfc-gh-jmichalak, line 675: These two look very similar. Please extract some common code.

pkg/resources/semantic_view.go
```diff
@@ -644,10 +656,86 @@ func getMetricDefinitionRequest(from any) (*sdk.MetricDefinitionRequest, error)
...
+
+	sqlExpression := c["sql_expression"].(string)
+	if sqlExpression == "" {
+		return nil, fmt.Errorf("sql_expression is required")
+	}
+	sqlExpRequest := sdk.NewSemanticSqlExpressionRequest().
+		WithSqlExpression(sqlExpression)
+	semExpRequest := sdk.NewSemanticExpressionRequest(qualifiedExpNameRequest, sqlExpRequest)
+
+	if c["comment"] != nil && c["comment"].(string) != "" {
```
sfc-gh-jmichalak: Our helpers in resource_helpers_create and resource_helpers_update can handle it.

pkg/resources/semantic_view.go
sfc-gh-asawicki: +1

pkg/resources/semantic_view.go
```diff
@@ -259,9 +265,14 @@ var semanticViewsSchema = map[string]*schema.Schema{
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				// TODO [SNOW-2396311]: update the SDK with the newly added/updated fields for semantic expressions, then add them here
-				// TODO [SNOW-2396371]: add PUBLIC/PRIVATE field
 				// TODO [SNOW-2398097]: add table_alias
 				// TODO [SNOW-2398097]: add fact_or_metric
+				"is_private": {
```
sfc-gh-asawicki, line 271: same as above. The rationale in https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/v1-preparations/CHANGES_BEFORE_V1.md#default-values

pkg/resources/semantic_view.go
sfc-gh-asawicki: +1

pkg/resources/semantic_view.go
sfc-gh-asawicki: +1

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-asawicki: +1

pkg/sdk/semantic_views_ext.go
sfc-gh-asawicki: +1

pkg/testacc/data_source_semantic_views_acceptance_test.go
sfc-gh-asawicki: +1

docs/resources/semantic_view.md
sfc-gh-sarur: done

pkg/resources/semantic_view.go
sfc-gh-sarur: done

pkg/resources/semantic_view.go
sfc-gh-sarur: done

pkg/resources/semantic_view.go
sfc-gh-sarur: done

pkg/resources/semantic_view.go
sfc-gh-sarur: done

pkg/sdk/generator/defs/semantic_views_def.go
sfc-gh-sarur: Facts can have the private field, but dimensions can't. So I wrapped the semantic expression object within FactDefinition and DimensionDefinition and added the is_private field for facts only. I did not strictly need to wrap the dimensions right now, but I believe it will get some more fields in the near future which won't apply to facts. So I just did the prep work for it now. It does not change anything for the users.

pkg/sdk/testint/semantic_view_integration_test.go
sfc-gh-sarur: done

pkg/sdk/semantic_views_ext.go
sfc-gh-sarur: done

pkg/testacc/data_source_semantic_views_acceptance_test.go
sfc-gh-sarur: The WithIsPrivate function is on *MetricDefinitionRequest, but the MetricDefinitionWithProps gives us a MetricDefinition object. In this case, how can this be done?

sfc-gh-jmichalak: Right, I didn't see this uses the sdk request functions. We can leave it for now, but we could just inline `*sdk.MetricDefinition` here.

sfc-gh-sarur: OK, I will leave is as is for now. That means I will need to leave the SetIsPrivate() methods as well, since the isPrivate field is unexported and there is no other way to modify it

MIGRATION_GUIDE.md
```diff
@@ -332,6 +332,12 @@ No changes in configuration are required.
 
 References: [#1630](https://github.com/snowflakedb/terraform-provider-snowflake/issues/1630)
 
+### *(new feature)* Private Facts and Metrics support in Semantic Views
```
sfc-gh-jmichalak: 1. This is stale now - move to top under `v2.14.0`
2. nit: let's not document the public behavior, as we don't want to rely on SF defaults in general. Maybe just link to the semantic view docs.

pkg/sdk/semantic_views_ext.go
sfc-gh-jmichalak: The methods are still here.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-jmichalak: This is not addressed. We should use structs from dto_builders, as they have public methods and fields. And then, remove the custom `Set` methods for the structs in `gen` file.

pkg/resources/semantic_view.go
sfc-gh-jmichalak: I believe this is incorrect. This should be set only if `v != BooleanDefault` - so you can use `booleanStringAttributeCreate` as I stated above.

pkg/testacc/data_source_semantic_views_acceptance_test.go
sfc-gh-jmichalak: I wrote it in the other comment - we should use other types for this. But it can be done in a follow-up. Also, we should test what happens when we don't set the is_private field. And I believe this could fail due to incorrect handling - I added a comment in the resource implementation.

pkg/resources/semantic_view.go
```diff
@@ -620,6 +633,14 @@ func getMetricDefinitionRequest(from any) (*sdk.MetricDefinitionRequest, error)
 				semExpRequest = semExpRequest.WithSynonyms(sRequest)
 			}
 		}
+		if v := c["is_private"]; v != nil {
+			isPrivate, err := booleanStringToBool(c["is_private"].(string))
+			if err != nil {
+				return nil, err
+			}
+			return metricDefinitionRequest.WithSemanticExpression(*semExpRequest).
```
sfc-gh-jmichalak: Let's define metricDefinitionRequest outside, and here don't return (this will get returned below). The reason is that when you add support for a new type, this would be incorrect or would have to be rewritten. Same in other places. The pattern is like

```
req := sdk.NewXRequest()
if field is set {
  req.WithField(v)
}
return req
```

pkg/resources/semantic_view.go
```diff
@@ -806,3 +866,35 @@ func getRelationshipRequests(from any) ([]sdk.SemanticViewRelationshipRequest, e
 	}
 	return to, nil
 }
+
+func getFactDefinitionRequests(from any) ([]sdk.FactDefinitionRequest, error) {
+	cols, ok := from.([]any)
+	if !ok {
+		return nil, fmt.Errorf("type assertion failure")
```
sfc-gh-jmichalak: Nit: let's improve error messages here or in the callers. Right now, this error doesn't say much.

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: I am a bit confused about this, we did not add any structs in the semantic_views_dto_builders_gen.go, are you referring to a different file?

sfc-gh-asawicki: Hmmm. I don't know, maybe it was marked in one of your older PRs but all these methods here in this file that return `*sdk.<Something>` should be removed from here..

The model package is to create the terraform representation (so methods returning model SemanticViewModel). What you are building here is not the model but the definition. So it should be built using the builders from the sdk like @sfc-gh-jmichalak suggested.

pkg/resources/semantic_view.go
sfc-gh-sarur: Makes sense, done

pkg/sdk/semantic_views_ext.go
sfc-gh-jmichalak: This is still unanswered. These helpers should not be needed, as we should use `Request` structs, which already have these helpers generated - see `semantic_views_dto_builders_gen.go`.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: We released v2.14.0, so this entry should be under 2.15.0

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
```diff
@@ -173,15 +173,18 @@ func (s *SemanticViewModel) WithMetrics(metrics []sdk.MetricDefinition) *Semanti
 			}
 			m["window_function"] = tfconfig.ListVariable(tfconfig.ObjectVariable(windFuncVar))
 		}
+		m["is_private"] = tfconfig.BoolVariable(v.GetIsPrivate())
```
sfc-gh-jmichalak: I think this way it will always be set. Let's set this map key only if the property in `v` is non-nil. Same in other places.

pkg/resources/semantic_view.go
```diff
@@ -644,10 +664,43 @@ func getMetricDefinitionRequest(from any) (*sdk.MetricDefinitionRequest, error)
...
+
+	semExpRequest, err := getSemanticExpressionRequest(from)
+	if err != nil {
+		return nil, err
+	}
+
+	factDefinitionRequest := sdk.NewFactDefinitionRequest().
+		WithSemanticExpression(*semExpRequest)
+
+	if v := c["is_private"]; v != nil {
```
sfc-gh-jmichalak: `&& v.(string) != BooleanDefault` is missing here. This will error in booleanStringToBool if this field is not set. I believe it was not caught in the tests because the `is_private` is always set in the current configuration.

sfc-gh-sarur: done

pkg/acceptance/bettertestspoc/config/model/semantic_view_model_ext.go
sfc-gh-sarur: Done

docs/resources/semantic_view.md
```diff
@@ -64,13 +64,15 @@ resource "snowflake_semantic_view" "complete" {
   }
 
   facts {
+    is_private                = "true"
```
sfc-gh-jcieslak, line 67: nit: not sure if we want to have them in quotes or not. Generally, the fields are strings, but HCL has built-in type coercion which allows for putting boolean `true` or `false` keywords in string fields, which will transform them into string representations `"true"` and `"false"`.

MIGRATION_GUIDE.md
```diff
@@ -26,6 +26,10 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ## v2.14.x ➞ v2.15.0
 
+### *(new feature)* Private Facts and Metrics support in Semantic Views
+
+We have added support for Private Facts and Metrics in the Semantic Views resource.
```
sfc-gh-jcieslak, line 31: nit: would be useful to give a bit more context (e.g. examples)

sfc-gh-sarur:
The acceptance tests failed on this test 
```
=== RUN   TestAcc_Provider_testFixedDefaultAuthenticatorForToken
    provider_acceptance_test.go:774: Generated config:
        data "snowflake_database" "t" {
          name = "acc_test_db_AT_AD592FC37EC1DB7CD6C231E6026F64B0F296D984"
        }
    provider_acceptance_test.go:782: TestStep 1/2 running init: exit status 1
        
        Error: Failed to install provider
        
        Error while installing snowflakedb/snowflake v2.8.0: could not query provider
        registry for registry.terraform.io/snowflakedb/snowflake: failed to retrieve
        authentication checksums for provider: the request failed after 2 attempts,
        please try again later: 502 Bad Gateway returned from github.com
        
--- FAIL: TestAcc_Provider_testFixedDefaultAuthenticatorForToken (1.38s)
```

The Semantic View resource and datasource tests passed successfully.

PR #4381, opened 2026-01-23
chore: Add missing WIF integration tests

sfc-gh-jmichalak:
- Add missing integration tests for AWS, Azure, and GCP workload identity federation (WIF) authentication methods
- Update the examples in manual tests

### DIFF
--- a/pkg/sdk/testint/users_integration_test.go
+++ b/pkg/sdk/testint/users_integration_test.go
@@ -576,16 +576,59 @@ func TestInt_Users(t *testing.T) {
 		subject string
 	}
 
+	awsWifConfig := func(arn string) *sdk.UserObjectWorkloadIdentityProperties {
+		return &sdk.UserObjectWorkloadIdentityProperties{
+			AwsType: &sdk.UserObjectWorkloadIdentityAws{
+				Arn: sdk.String(arn),
+			},
+		}
+	}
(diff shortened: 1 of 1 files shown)


pkg/sdk/testint/users_integration_test.go
```diff
@@ -1299,6 +1502,178 @@ func TestInt_Users(t *testing.T) {
...
+			wifConfig: func(data any) *sdk.UserObjectWorkloadIdentityProperties {
+				return gcpWifConfig(data.(gcpWifData).subject)
+			},
+			wifAssertion: func(t *testing.T, assertion *objectassert.UserWorkloadIdentityAuthenticationMethodsAssert, data any) {
+				gcpWifAssertion(t, assertion, data.(gcpWifData).subject)
+			},
+		},
+	}
+
+	for _, tt := range wifAlterTestCases {
```
sfc-gh-asawicki, line 1617: I'm unsure if we need to test the whole "matrix" of options here. Maybe it would be enough to test all the "providers" for one of the user types? My main concern is the explosion of test cases and execution time. It's not blocking this PR, but this is something we should discuss internally soon.

sfc-gh-jmichalak: Sure, we can discuss our approach, because it affects many other resources.

PR #4394, opened 2026-01-30
chore: Fix a warehouse integration test

sfc-gh-jmichalak:
The warehouse integration test had a potential race condition in which the warehouse state and type were checked against stale data.

- Move warehouse `ShowByID` call inside the `Eventually` block.
- Combine state and type validation into a single atomic check.

### DIFF
--- a/pkg/sdk/testint/warehouses_integration_test.go
+++ b/pkg/sdk/testint/warehouses_integration_test.go
@@ -42,6 +42,13 @@ func TestInt_Warehouses(t *testing.T) {
 		time.Sleep(3 * time.Second)
 	}
 
+	warehouseCondition := func(id sdk.AccountObjectIdentifier, pred func(r *sdk.Warehouse) bool) func() bool {
+		return func() bool {
+			r, e := client.Warehouses.ShowByID(ctx, id)
+			return e == nil && pred(r)
+		}
+	}
+
 	t.Run("show: without options", func(t *testing.T) {
 		warehouses, err := client.Warehouses.Show(ctx, nil)
 		require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -414,10 +414,14 @@ func TestInt_Warehouses(t *testing.T) {
 		})
 		t.Cleanup(warehouseCleanup)
 
-		returnedWarehouse, err := client.Warehouses.ShowByID(ctx, warehouse.ID())
-		require.NoError(t, err)
-		assert.Equal(t, sdk.WarehouseTypeStandard, returnedWarehouse.Type)
-		require.Eventually(t, func() bool { return sdk.WarehouseStateStarted == returnedWarehouse.State }, 5*time.Second, time.Second)
+		var returnedWarehouse *sdk.Warehouse
+		var err error
```
sfc-gh-jcieslak: Is it necessary to save the error (I don't see it being used in a later assertion)?
I think it would be sufficient to maybe log it inside Eventually assert, wdyt?
edit: oh, okay it's needed because you need to reassign `=` instead of assign `:=` (then maybe let's make this error local to the function used in the assert?

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -414,10 +414,14 @@ func TestInt_Warehouses(t *testing.T) {
 		})
 		t.Cleanup(warehouseCleanup)
 
-		returnedWarehouse, err := client.Warehouses.ShowByID(ctx, warehouse.ID())
-		require.NoError(t, err)
-		assert.Equal(t, sdk.WarehouseTypeStandard, returnedWarehouse.Type)
-		require.Eventually(t, func() bool { return sdk.WarehouseStateStarted == returnedWarehouse.State }, 5*time.Second, time.Second)
+		var returnedWarehouse *sdk.Warehouse
```
sfc-gh-asawicki: The same problem was in `alter: suspend with a long running-query`, can we solve it the same way here for consistency? so extract a func and not save wh/err outside of the func - as not needed.

sfc-gh-asawicki (changes requested):

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jmichalak: Yes, it was needed because we are checking other errors later. Note that this pattern is almost everywhere, i.e. we don't have scoped errors.

sfc-gh-asawicki (approved):

PR #4404, opened 2026-02-02
feat: Add the internal_stage resource

sfc-gh-jmichalak:
## Changes
The existing snowflake_stage resource handles both internal and external stages, making it complex to maintain and extend. By splitting internal stages into their own dedicated resource, we improve clarity, maintainability, and align with Snowflake's latest documentation and provider conventions.
- Add new snowflake_internal_stage resource as a preview feature
- Add StageType enum in SDK for type-safe stage type handling
- Update SDK Comment field in alter operations to StringAllowEmpty to support "unsetting" comments

## References

### DIFF
--- a/pkg/acceptance/bettertestspoc/assert/resourceassert/internal_stage_resource_ext.go
+++ b/pkg/acceptance/bettertestspoc/assert/resourceassert/internal_stage_resource_ext.go
@@ -8,11 +8,13 @@ import (
 )
 
 func (i *InternalStageResourceAssert) HasDirectoryEnableString(expected string) *InternalStageResourceAssert {
+	i.AddAssertion(assert.ValueSet("directory.#", "1"))
 	i.AddAssertion(assert.ValueSet("directory.0.enable", expected))
 	return i
 }
 
 func (i *InternalStageResourceAssert) HasDirectoryAutoRefreshString(expected string) *InternalStageResourceAssert {
+	i.AddAssertion(assert.ValueSet("directory.#", "1"))
 	i.AddAssertion(assert.ValueSet("directory.0.auto_refresh", expected))
 	return i
 }
--- a/pkg/resources/internal_stage.go
+++ b/pkg/resources/internal_stage.go
@@ -117,7 +117,7 @@ func ImportInternalStage(ctx context.Context, d *schema.ResourceData, meta any) 
 	if err != nil {
 		return nil, err
 	}
-	if _, err := ImportName[sdk.SchemaObjectIdentifier](context.Background(), d, nil); err != nil {
+	if _, err := ImportName[sdk.SchemaObjectIdentifier](ctx, d, nil); err != nil {
 		return nil, err
 	}
 	stageDetails, err := client.Stages.Describe(ctx, id)
--- a/pkg/testacc/resource_internal_stage_acceptance_test.go
+++ b/pkg/testacc/resource_internal_stage_acceptance_test.go
@@ -22,7 +22,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
-func TestAcc_InternalStage_basic(t *testing.T) {
+func TestAcc_InternalStage_BasicUseCase(t *testing.T) {
 	id := testClient().Ids.RandomSchemaObjectIdentifier()
 	newId := testClient().Ids.RandomSchemaObjectIdentifier()
 	comment, changedComment := random.Comment(), random.Comment()
@@ -295,7 +295,7 @@ func TestAcc_InternalStage_basic(t *testing.T) {
 	})
 }
 
-func TestAcc_InternalStage_complete(t *testing.T) {
+func TestAcc_InternalStage_CompleteUseCase(t *testing.T) {
 	id := testClient().Ids.RandomSchemaObjectIdentifier()
 	comment := random.Comment()
 


pkg/resources/internal_stage.go
```diff
@@ -0,0 +1,320 @@
...
+
+		Schema: internalStageSchema,
+		Importer: &schema.ResourceImporter{
+			StateContext: TrackingImportWrapper(resources.InternalStage, ImportInternalStage),
+		},
+		Timeouts: defaultTimeouts,
+	}
+}
+
+func ImportInternalStage(ctx context.Context, d *schema.ResourceData, meta any) ([]*schema.ResourceData, error) {
```
sfc-gh-jcieslak, line 114: Shouldn't we validate that the stage imported here is an internal stage and not an external one?

pkg/acceptance/bettertestspoc/assert/resourceassert/internal_stage_resource_ext.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"strconv"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func (i *InternalStageResourceAssert) HasDirectoryEnableString(expected string) *InternalStageResourceAssert {
+	i.AddAssertion(assert.ValueSet("directory.0.enable", expected))
```
sfc-gh-jcieslak, line 12: Shouldn't this and the assert below also check that the directory is set (`directory.# == 1`)?

pkg/resources/internal_stage.go
```diff
@@ -0,0 +1,320 @@
...
+	client := meta.(*provider.Context).Client
+	id, err := sdk.ParseSchemaObjectIdentifier(d.Id())
+	if err != nil {
+		return nil, err
+	}
+	if _, err := ImportName[sdk.SchemaObjectIdentifier](context.Background(), d, nil); err != nil {
```
sfc-gh-jcieslak: can use context passed to the Import function

pkg/testacc/resource_internal_stage_acceptance_test.go
```diff
@@ -3,46 +3,377 @@
...
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	r "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/tfversion"
 )
 
-func TestAcc_InternalStage(t *testing.T) {
+func TestAcc_InternalStage_basic(t *testing.T) {
```
sfc-gh-jcieslak: Shouldn't it be BasicUseCase / CompleteUseCase to be visible in other tests?

examples/resources/snowflake_internal_stage/resource.tf
```diff
@@ -0,0 +1,24 @@
...
+}
+
+# complete resource
+resource "snowflake_internal_stage" "complete" {
+  name     = "complete_stage"
+  database = "my_database"
+  schema   = "my_schema"
+
+  encryption {
+    snowflake_full {}
```
sfc-gh-jcieslak, line 15: I'm guessing those are blocks, just in case there will be encryption dependent options available in the future (?)

pkg/resources/internal_stage.go
sfc-gh-jmichalak: Done.

pkg/testacc/resource_internal_stage_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/resources/internal_stage.go
sfc-gh-jmichalak: This will show up in the plan after import. I'll add tests covering this in the next PRs.

pkg/acceptance/bettertestspoc/assert/resourceassert/internal_stage_resource_ext.go
sfc-gh-jmichalak: Done.

examples/resources/snowflake_internal_stage/resource.tf
sfc-gh-jmichalak: Yes.

pkg/resources/internal_stage.go
sfc-gh-jcieslak: Okay, I was just referring to a similar case in grants where we decided import should validate correctness, and maybe they should in this case too? Can be discussed later on (but before the release)

sfc-gh-jmichalak: Right, but the grants case is different itself due to a different "conceptual model". We don't validate e.g. ForceNew fields in other resources anyway, so let's keep it consistent.

sfc-gh-asawicki: Let's discuss this as a future guideline. There is an added value in failing fast for such cases (such cases meaning: objects we split into multiple resources, like stages, storage integrations, users, etc; this way we ensure our users use the correct import without the forced recreation later on).

pkg/acceptance/bettertestspoc/assert/resourceassert/internal_stage_resource_ext.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"strconv"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func (i *InternalStageResourceAssert) HasDirectoryEnableString(expected string) *InternalStageResourceAssert {
+	i.AddAssertion(assert.ValueSet("directory.#", "1"))
```
sfc-gh-asawicki, line 11: I don't think it's needed here to always add the size assertion, it doesn't prevent second assertion from hapenning

PR #4407, opened 2026-02-03
feat: Add the external_azure_stage resource

sfc-gh-jmichalak:
## Changes
The old `snowflake_stage` resource handled both internal and external stages, making it complex to maintain and extend. This PR introduces dedicated stage resources aligned with Snowflake's documentation and the provider's new conventions, providing clearer separation of concerns and better support for cloud-specific options.

- Add `snowflake_external_azure_stage` resource.
- Add SDK enhancements for stage types and cloud providers with type-safe enums
- Extract common stage functionality (schema, rename, directory table handling) into a shared file.
- Add custom diff functions for detecting external stage type/cloud changes.
- Adjust internal stages to handle the directory properly.

## TODO
- File formats
- Other stage types
- Stages data source
- Fill the follow-up descriptions

## References

### DIFF
--- a/pkg/resources/internal_stage.go
+++ b/pkg/resources/internal_stage.go
@@ -71,7 +71,7 @@ var internalStageSchema = func() map[string]*schema.Schema {
 						ValidateDiagFunc: validateBooleanString,
 						Optional:         true,
 						Description:      "Specifies whether Snowflake should automatically refresh the directory table metadata when new or updated data files are available on the internal named stage.",
-						DiffSuppressFunc: ignoreNestedBooleanDefault,
+						DiffSuppressFunc: IgnoreChangeToCurrentSnowflakePlainValueInOutput("describe_output.0.directory_table", "auto_refresh"),
 					},
 				},
 			},
--- a/pkg/resources/diff_suppressions.go
+++ b/pkg/resources/diff_suppressions.go
@@ -310,7 +310,3 @@ func ignoreCaseAndTrimSpaceSuppressFunc(_, old, new string, _ *schema.ResourceDa
 func ignoreAlwaysSuppressFunc(_, old, new string, _ *schema.ResourceData) bool {
 	return true
 }
-
-func ignoreNestedBooleanDefault(_, old, new string, _ *schema.ResourceData) bool {
-	return old == "" && new == BooleanDefault
-}


pkg/acceptance/helpers/stage_client.go
```diff
@@ -61,6 +61,11 @@ func (c *StageClient) CreateStage(t *testing.T) (*sdk.Stage, func()) {
 	return c.CreateStageInSchema(t, c.ids.SchemaId())
 }
 
+func (c *StageClient) CreateStageWithId(t *testing.T, id sdk.SchemaObjectIdentifier) (*sdk.Stage, func()) {
```
sfc-gh-jcieslak: Shouldn't we keep it consistent with Alter commands and specify that it's Internal or External stage? (I have a feeling I left similar comment somewhere but forgot where ;_;)

pkg/resources/custom_diffs.go
```diff
@@ -99,6 +99,13 @@ func ForceNewIfChangeToEmptyString(key string) schema.CustomizeDiffFunc {
 	})
 }
 
+// ForceNewIfNotDefault sets a ForceNew for a string field which was set to a default value.
```
sfc-gh-jcieslak: `a default value` -> `a non-default value`

pkg/resources/custom_diffs.go
```diff
@@ -302,6 +314,23 @@ func RecreateWhenStreamIsStale() schema.CustomizeDiffFunc {
 	}
 }
 
+func RecreateWhenCredentialsAreSetOnExternalStageWithStorageIntegration() schema.CustomizeDiffFunc {
+	return func(_ context.Context, diff *schema.ResourceDiff, _ interface{}) error {
```
sfc-gh-jcieslak, line 324: I don't get this function, shouldn't it be just checking the change as you did with the credentials, but for storage_integration it should be just `Get`, I don't think checking changes there is necessary 🤔 (the second `if` could be removed imo)

sfc-gh-jcieslak: or maybe I'm just confused by the function name and a few words of function description would help

pkg/resources/internal_stage.go
```diff
@@ -61,28 +61,21 @@ var internalStageSchema = func() map[string]*schema.Schema {
...
 					},
 					"auto_refresh": {
 						Type:             schema.TypeString,
 						Default:          BooleanDefault,
 						ValidateDiagFunc: validateBooleanString,
 						Optional:         true,
-						ForceNew:         true,
 						Description:      "Specifies whether Snowflake should automatically refresh the directory table metadata when new or updated data files are available on the internal named stage.",
-						DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInDescribe("directory_table.0.auto_refresh"),
+						DiffSuppressFunc: ignoreNestedBooleanDefault,
```
sfc-gh-jcieslak: Why wasn't it working with IgnoreChangeToCurrentSnowflakeValueInDescribe? ignoreNestedBooleanDefault feels hacky as I think there are everywhere similar cases and functions, like the one used previously, worked just fine

sfc-gh-jcieslak: Ok, I think I see. There's the `ForceNewIfNotDefault("directory.0.auto_refresh")`, but if this field is ForceNew, shouldn't it be reguarly marked and not with the custom diff? Now, as far as I understand, there will be no changes from value to default, which is not always correct, because the default may be different from the value before. Please clarify if something different is happening here, because I don't understand why new behavior (new suppress func and custom diff func) needs to be added.

pkg/sdk/stages_enums.go
```diff
@@ -74,3 +74,25 @@ var AcceptableStageTypes = map[StageType][]StageType{
 	StageTypeInternal: {StageTypeInternal, StageTypeInternalNoCse},
 	StageTypeExternal: {StageTypeExternal},
 }
+
+type StageCloud string
+
+var (
+	StageCloudAzure StageCloud = "AZURE"
+	StageCloudAws   StageCloud = "AWS"
```
sfc-gh-jcieslak, line 82: nit: Shouldn't there also be aws-compatible?
Probably would have to be added in the next prs anyway

pkg/testacc/resource_internal_stage_acceptance_test.go
```diff
@@ -105,6 +108,11 @@ func TestAcc_InternalStage_BasicUseCase(t *testing.T) {
 			},
 			// Set optionals (complete)
 			{
+				ConfigPlanChecks: resource.ConfigPlanChecks{
```
sfc-gh-jcieslak, line 111: nit: generally, when writing those tests, I was separating force new fields from those that can update (because now, having update on force new fields, we don't test the update where we are actually setting the fields). This could be mitigated by having one additional step: the first step be an update (without force new fields, which would test update), and the second also update but with changed force new fields (here we are testing force new is triggered as expected).

templates/resources/internal_stage.md.tmpl
```diff
@@ -13,6 +13,8 @@ description: |-
 
 -> **Note** Temporary stages are not supported because they result in per-session objects.
 
+-> **Note** External changes detection on the `encryption` field is not supported due to Snowflake and Terraform limitations.
```
sfc-gh-jcieslak: Question: What exact limitations?

pkg/resources/external_azure_stage.go
```diff
@@ -0,0 +1,478 @@
...
+		}); err != nil {
+			return nil, err
+		}
+	}
+	if details.PrivateLink != nil {
+		if err := d.Set("use_privatelink_endpoint", booleanStringFromBool(details.PrivateLink.UsePrivatelinkEndpoint)); err != nil {
+			return nil, err
+		}
+	} else {
+		if err := d.Set("use_privatelink_endpoint", BooleanFalse); err != nil {
```
sfc-gh-jcieslak: I think, we should not set anything and let the default value be there

pkg/resources/external_azure_stage.go
```diff
@@ -0,0 +1,478 @@
...
+							Schema: map[string]*schema.Schema{},
+						},
+					},
+				},
+			},
+		},
+		"use_privatelink_endpoint": {
+			Type:             schema.TypeString,
+			Optional:         true,
+			Default:          BooleanDefault,
```
sfc-gh-jcieslak, line 92: Diff suppress is missing here

pkg/resources/external_azure_stage.go
```diff
@@ -0,0 +1,478 @@
...
+		ReadContext:   PreviewFeatureReadContextWrapper(string(previewfeatures.ExternalAzureStageResource), TrackingReadWrapper(resources.ExternalAzureStage, ReadExternalAzureStageFunc(true))),
+		UpdateContext: PreviewFeatureUpdateContextWrapper(string(previewfeatures.ExternalAzureStageResource), TrackingUpdateWrapper(resources.ExternalAzureStage, UpdateExternalAzureStage)),
+		DeleteContext: DeleteStage(previewfeatures.ExternalAzureStageResource, resources.ExternalAzureStage),
+		Description:   "Resource used to manage external Azure stages. For more information, check [external stage documentation](https://docs.snowflake.com/en/sql-reference/sql/create-stage#external-stage-parameters-externalstageparams).",
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.ExternalAzureStage, customdiff.All(
+			ComputedIfAnyAttributeChanged(externalAzureStageSchema, ShowOutputAttributeName, "name", "comment", "url", "storage_integration", "encryption"),
+			ComputedIfAnyAttributeChanged(externalAzureStageSchema, FullyQualifiedNameAttributeName, "name"),
```
sfc-gh-jcieslak, line 156: I think ComputedIfAnyAttributeChanged is missing for DescOutput (if it's there, if not it should be added to the common schema or in this schema)

pkg/sdk/stages_enums.go
sfc-gh-jmichalak: The s3-compatible returns AWS - see integration tests.

pkg/resources/internal_stage.go
sfc-gh-jmichalak: The nested fields are much trickier than that (convoluting this with other places using the booleanDefaults...), so I was experimenting a bit. In the end, I found a solution to get rid of it and reuse IgnoreChange...

pkg/resources/custom_diffs.go

pkg/acceptance/helpers/stage_client.go
sfc-gh-jmichalak: Done.

templates/resources/internal_stage.md.tmpl
sfc-gh-jmichalak: We don't usually document the implementation details like this, but added.

pkg/resources/custom_diffs.go
sfc-gh-jmichalak: Adjusted name and description. We need to check this one both ways.

pkg/resources/external_azure_stage.go
sfc-gh-jmichalak: Added.

pkg/resources/external_azure_stage.go
sfc-gh-jmichalak: Added.

pkg/testacc/resource_internal_stage_acceptance_test.go
sfc-gh-jmichalak: See

```
		// Alter (update comment, directory.enable)
        // Unset comment and rename
```

pkg/resources/external_azure_stage.go
sfc-gh-jmichalak: Done.

pkg/acceptance/helpers/stage_client.go
sfc-gh-jcieslak: There are other function e.g. CreateStageOnS3 etc. but can be addressed later on (e.g. separate pr with just renames)

pkg/resources/custom_diffs.go
sfc-gh-jcieslak: Much better, thx

templates/resources/internal_stage.md.tmpl
sfc-gh-jcieslak: Just wanted to know the limitations, didn't expect anything to be changed, but I guess I'm also fine with the change 😅

PR #4409, opened 2026-02-03
feat: Add the new external_gcs_stage resource

sfc-gh-jmichalak:
## Changes
The old `snowflake_stage` resource handled both internal and external stages, making it complex to maintain and extend. This PR introduces dedicated stage resources aligned with Snowflake's documentation and the provider's new conventions, providing clearer separation of concerns and better support for cloud-specific options.

- Add `snowflake_external_gcs_stage` resource.
- Adjust previous directory table implementation.

## TODO
- File formats
- Other stage types
- Stages data source
- Fill the follow-up descriptions

## References

### DIFF
--- a/pkg/testacc/resource_external_gcs_stage_acceptance_test.go
+++ b/pkg/testacc/resource_external_gcs_stage_acceptance_test.go
@@ -13,9 +13,7 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config/model"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers/random"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/ids"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/importchecks"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvs"
-	resourcehelpers "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeroles"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
 	r "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/resources"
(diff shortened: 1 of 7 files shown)


pkg/resources/external_gcs_stage.go
```diff
@@ -0,0 +1,429 @@
...
+						ValidateDiagFunc: validateBooleanString,
+						Optional:         true,
+						Description:      "Specifies whether Snowflake should enable triggering automatic refreshes of the directory table metadata.",
+						DiffSuppressFunc: IgnoreChangeToCurrentSnowflakePlainValueInOutput("describe_output.0.directory_table", "auto_refresh"),
+					},
+					"notification_integration": {
+						Type:             schema.TypeString,
+						Optional:         true,
+						ForceNew:         true,
+						Description:      "Specifies the name of the notification integration used to automatically refresh the directory table metadata.",
```
sfc-gh-jcieslak: can use identifier validation and description about allowed characters (some in other identifier fields)

pkg/resources/external_gcs_stage.go
```diff
@@ -0,0 +1,429 @@
...
+	database := d.Get("database").(string)
+	schemaName := d.Get("schema").(string)
+	id := sdk.NewSchemaObjectIdentifier(database, schemaName, name)
+
+	url := d.Get("url").(string)
+	storageIntegrationRaw := d.Get("storage_integration").(string)
+	storageIntegrationId := sdk.NewAccountObjectIdentifier(storageIntegrationRaw)
+
+	externalStageParams := sdk.NewExternalGCSStageParamsRequest(url).
+		WithStorageIntegration(storageIntegrationId)
```
sfc-gh-jcieslak, line 203: shouldn't storage id be a part of the NewExternalGCSStageParamsRequest since it's required parameter?

pkg/resources/external_gcs_stage.go
```diff
@@ -0,0 +1,429 @@
...
+
+	err := errors.Join(
```
sfc-gh-jcieslak, line 205: join two `errors.Join`, I don't think there's a reason to keep them separate

pkg/resources/external_gcs_stage.go
```diff
@@ -0,0 +1,429 @@
...
+			directoryTable := []any{
+				map[string]any{
+					"enable":       details.DirectoryTable.Enable,
+					"auto_refresh": details.DirectoryTable.AutoRefresh,
+				},
+			}
+			directoryTableToSet := []any{
+				map[string]any{
+					"enable":       details.DirectoryTable.Enable,
+					"auto_refresh": booleanStringFromBool(details.DirectoryTable.AutoRefresh),
```
sfc-gh-jcieslak, line 277: What about notification integration? I think it was also missing from the import

pkg/resources/external_gcs_stage.go
```diff
@@ -0,0 +1,429 @@
...
+		return diag.FromErr(err)
+	}
+
+	id, err = handleStageRename(ctx, client, d, id)
+	if err != nil {
+		d.Partial(true)
+		return diag.FromErr(err)
+	}
+
+	err = handleStageDirectoryTable(ctx, client, d, id)
```
sfc-gh-jcieslak, line 328: I think it doesn't handle notification integration (and auto_refresh), right?

pkg/testacc/resource_external_gcs_stage_acceptance_test.go
```diff
@@ -0,0 +1,517 @@
...
+					assert.Check(resource.TestCheckResourceAttr(modelUpdated.ResourceReference(), "describe_output.0.directory_table.0.auto_refresh", "false")),
+				),
+			},
+			// External change detection
+			{
+				PreConfig: func() {
+					testClient().Stage.DropStageFunc(t, id)()
+					testClient().Stage.CreateStageOnGCSWithId(t, id, gcsUrl)
+				},
+				Config: accconfig.FromModels(t, modelUpdated),
```
sfc-gh-jcieslak, line 320: let's add plan check here

pkg/testacc/resource_external_gcs_stage_acceptance_test.go
```diff
@@ -0,0 +1,517 @@
...
+						HasOwner(snowflakeroles.Accountadmin.Name()).
+						HasCreatedOnNotEmpty(),
+					assert.Check(resource.TestCheckResourceAttr(modelComplete.ResourceReference(), "describe_output.0.directory_table.0.enable", "true")),
+					assert.Check(resource.TestCheckResourceAttr(modelComplete.ResourceReference(), "describe_output.0.directory_table.0.auto_refresh", "false")),
+				),
+			},
+			{
+				Config:       accconfig.FromModels(t, modelComplete),
+				ResourceName: modelComplete.ResourceReference(),
+				ImportState:  true,
```
sfc-gh-jcieslak: can't we just use import state verify? This checks only two fields' import correctness (same in other test)

sfc-gh-jmichalak: We can, but we would skip the missing fields anyway.

pkg/resources/external_gcs_stage.go
sfc-gh-jmichalak: What about it? We don't have it in SHOW/DESC. I'll follow up with this field, as it requires more resources to test properly.

sfc-gh-jcieslak: Then I think it should be documented as it current is with encryption that we don't support the external changes for it

pkg/testacc/resource_external_gcs_stage_acceptance_test.go
sfc-gh-jcieslak: I don't get it, ImportStateVerify would validate every field (unless specified in the ignore list)

sfc-gh-jmichalak (commented):
Answered.

pkg/resources/external_gcs_stage.go
sfc-gh-jmichalak: Added.

pkg/resources/external_gcs_stage.go
sfc-gh-jmichalak: Yes, will do in a followup.

pkg/resources/external_gcs_stage.go
sfc-gh-jmichalak: They operate on different objects (externalStageParams vs request).

pkg/resources/external_gcs_stage.go
sfc-gh-jmichalak: No, as they are ForceNew

pkg/testacc/resource_external_gcs_stage_acceptance_test.go
sfc-gh-jmichalak: Done, but we had it in different objects.

pkg/resources/external_gcs_stage.go
sfc-gh-jcieslak: nit: could also be documented on the field doc level

Issue #4403, opened 2026-02-02
[Bug]: snowflake_tag_association errors on update for object_type FUNCTION


sudo-make-me-a-coffee:
### Terraform CLI Version

1.10.5

### Terraform Provider Version

2.1.1

### Terraform Configuration

```terraform
resource "snowflake_tag_association" "test_env_tag" {
  object_identifiers =[snowflake_function_python.test_function.fully_qualified_name]
  object_type        = "FUNCTION"
  tag_id             = "MQ_TAGS.CANONICAL_TAGS.ENV"
  tag_value          = "qa"
  skip_validation    = false
}
```

### Category

category:resource

### Object type(s)

resource:tag_association

### Expected Behavior

Updating `tag_value` for a `snowflake_tag_association`, `object_type` FUNCTION results in a successful update of the resource as defined in the plan:

```terraform
  # snowflake_tag_association.test_env_tag will be updated in-place
  ~ resource "snowflake_tag_association" "test_env_tag" {
        id                 = "\"LC_TAGS\".\"CANONICAL_TAGS\".\"ENV\"|qa|FUNCTION"
      ~ tag_value          = "qa" -> "dev"
        # (4 unchanged attributes hidden)
    }
```

### Actual Behavior

The apply fails with a parsing error of the functions `fully_qualified_name` :

```terraform
Error: unable to read identifier: "TESTDB"."LC_NEW_SCHEMA"."DD_SUBMIT_COUNT_METRICS"(VARCHAR), err = parse error on line 1, column 50: extraneous or missing " in quoted-field

  with snowflake_tag_association.test_env_tag,
  on main.tf line 277, in resource "snowflake_tag_association" "test_env_tag":
 277: resource "snowflake_tag_association" "test_env_tag" {
```

### Steps to Reproduce

1. Create a `snowflake_function_python` resource and a `snowflake_tag_association` setting the object_identifiers as the functions fully qualified name. 
2. Once the function and tag are created update the tag value of the `snowflake_tag_association`
(Plan will look good and show an update that will fail on apply)

### How much impact is this issue causing?

Low

### Additional Information

Tested manually creating the fully qualified name omitting the method signature:

```terraform
format("\"%s\".\"%s\".\"%s\"", snowflake_function_python.metrics.database, snowflake_function_python.metrics.schema, snowflake_function_python.metrics.name)
```

[...]

PR #4412, opened 2026-02-05
fix: tag association identifier parsing during update

sfc-gh-jcieslak:
## Changes
Adjust incorrect identifier parsing in tag association for function and procedure object types.
Add a test that validates the fix.
References: #4403


pkg/resources/tag_association.go
```diff
@@ -228,17 +228,23 @@ func UpdateContextTagAssociation(ctx context.Context, d *schema.ResourceData, me
 		newIdsFQN := collections.Map(newIds, sdk.ObjectIdentifier.FullyQualifiedName)
```
sfc-gh-jmichalak, line 228: The migration guide entry is missing.

sfc-gh-jcieslak: that's true, added

PR #4421, opened 2026-02-08
feat: Prepare new storage integration resources

sfc-gh-asawicki:
Prepare the schemas for all types of storage integrations:
- based on the old resource (some fields removed/modified/added)
- these schema will most probably slightly change during the implementation
- resource logic is NOT part of this change

Generate all code for storage integrations:
- schemas for each details output
- documentation with examples
- model builders
- resource assertions
- resource describe output assertions

Next PRs:
- add logic to each storage integration and add acceptance tests
- update data source
- add migration guide
- (nit) extract common schema/logic part
- (later) handle reaction to external type/provider change


pkg/provider/previewfeatures/preview_features.go
```diff
@@ -86,6 +86,9 @@ const (
 	StageResource                                 feature = "snowflake_stage_resource"
 	StagesDatasource                              feature = "snowflake_stages_datasource"
 	StorageIntegrationResource                    feature = "snowflake_storage_integration_resource"
+	StorageIntegrationAwsResource                 feature = "snowflake_storage_integration_aws_resource"
```
sfc-gh-jmichalak, line 89: Nit: unit tests are missing.

sfc-gh-asawicki: Hmmm, I see there are a lot of values missed there. The question is: do we want to add all values to this test? I will add them in one of the PRs and we can discuss later.

PR #4426, opened 2026-02-09
feat: Implement AWS storage integration resource

sfc-gh-asawicki:
- Add implementation to previously prepared resource
- Adjust schema (allowed/blocked locations list -> set)
- Add acceptance tests
- Add missing assertions
- Add describe assertions quick workaround
- Fill out ID in aws/azure/gcs describe details

### DIFF
--- a/pkg/resources/storage_integration_aws.go
+++ b/pkg/resources/storage_integration_aws.go
@@ -5,11 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"strings"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
-	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider/validators"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/previewfeatures"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/schemas"
@@ -40,7 +38,8 @@ var storageIntegrationAwsSchema = map[string]*schema.Schema{
 		Type:             schema.TypeString,
 		Required:         true,
 		ForceNew:         true,
-		ValidateDiagFunc: validators.NormalizeValidation(sdk.ToS3Protocol),
+		DiffSuppressFunc: NormalizeAndCompare(sdk.ToS3Protocol),
+		ValidateDiagFunc: sdkValidation(sdk.ToS3Protocol),
 		Description:      fmt.Sprintf("Specifies the storage provider for the integration. Valid options are: %s", possibleValuesListed(sdk.AllStorageProviders)),
 	},
 	"storage_allowed_locations": {
@@ -192,7 +191,7 @@ func GetReadStorageIntegrationAwsFunc(withExternalChangesMarking bool) schema.Re
 		if withExternalChangesMarking {
 			if err = handleExternalChangesToObjectInFlatDescribe(d,
 				outputMapping{"external_id", "storage_aws_external_id", awsDetails.ExternalId, awsDetails.ExternalId, nil},
-				outputMapping{"use_privatelink_endpoint", "use_privatelink_endpoint", awsDetails.UsePrivatelinkEndpoint, awsDetails.UsePrivatelinkEndpoint, nil},
+				outputMapping{"use_privatelink_endpoint", "use_privatelink_endpoint", awsDetails.UsePrivatelinkEndpoint, booleanStringFromBool(awsDetails.UsePrivatelinkEndpoint), nil},
 			); err != nil {
 				return diag.FromErr(err)
 			}
@@ -228,7 +227,7 @@ func CreateStorageIntegrationAws(ctx context.Context, d *schema.ResourceData, me
 	id := sdk.NewAccountObjectIdentifierFromFullyQualifiedName(name)
 	enabled := d.Get("enabled").(bool)
 	storageAllowedLocations, _ := parseLocations(d.Get("storage_allowed_locations").(*schema.Set).List())
-	storageProvider := strings.ToUpper(d.Get("storage_provider").(string))
+	storageProvider := d.Get("storage_provider").(string)
 	s3Protocol, err := sdk.ToS3Protocol(storageProvider)
 	if err != nil {
 		return diag.FromErr(err)


pkg/resources/resource_helpers_update.go
```diff
@@ -16,6 +16,13 @@ func stringAttributeUpdate(d *schema.ResourceData, key string, setField **string
 	return nil
 }
 
+func stringAttributeUpdateSetOnlyNotEmpty(d *schema.ResourceData, key string, setField **string) error {
```
sfc-gh-jmichalak, line 19: Nit: maybe this one should be `stringAttributeUpdateSetOnly`, and the one below `stringAttributeUpdateSetOnlyAllowEmpty`. Of course, we can leave this for now, and take a broader a look during some kind of a refactor, because now we have a few established patterns.

pkg/resources/storage_integration_aws.go
```diff
@@ -33,18 +40,18 @@ var storageIntegrationAwsSchema = map[string]*schema.Schema{
 		Type:             schema.TypeString,
 		Required:         true,
 		ForceNew:         true,
-		ValidateDiagFunc: StringInSlice(sdk.AllStorageProviders, true),
+		ValidateDiagFunc: validators.NormalizeValidation(sdk.ToS3Protocol),
 		Description:      fmt.Sprintf("Specifies the storage provider for the integration. Valid options are: %s", possibleValuesListed(sdk.AllStorageProviders)),
```
sfc-gh-jmichalak, line 43: Diff suppress is missing.

pkg/resources/storage_integration_aws.go
```diff
@@ -59,27 +66,26 @@ var storageIntegrationAwsSchema = map[string]*schema.Schema{
...
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInDescribe("use_privatelink_endpoint"),
+		Description:      booleanStringFieldDescription("Specifies whether to use outbound private connectivity to harden the security posture."),
 	},
 	"storage_aws_role_arn": {
-		Type:        schema.TypeString,
-		Required:    true,
-		Description: "Specifies the Amazon Resource Name (ARN) of the AWS identity and access management (IAM) role that grants privileges on the S3 bucket containing your data files.",
+		Type:         schema.TypeString,
+		Required:     true,
+		ValidateFunc: validation.StringIsNotEmpty,
```
sfc-gh-jmichalak, line 74: Q: why do we need this validator, if this field is required? Is it for `field = ""`? Doesn't Terraform treat this as non-set?

pkg/resources/storage_integration_aws.go
```diff
@@ -109,22 +116,233 @@ func StorageIntegrationAws() *schema.Resource {
...
+
+		awsDetails, err := client.StorageIntegrations.DescribeAwsDetails(ctx, id)
+		if err != nil {
+			return diag.FromErr(fmt.Errorf("could not describe aws storage integration (%s), err = %w", d.Id(), err))
+		}
+
+		if withExternalChangesMarking {
+			if err = handleExternalChangesToObjectInFlatDescribe(d,
+				outputMapping{"external_id", "storage_aws_external_id", awsDetails.ExternalId, awsDetails.ExternalId, nil},
+				outputMapping{"use_privatelink_endpoint", "use_privatelink_endpoint", awsDetails.UsePrivatelinkEndpoint, awsDetails.UsePrivatelinkEndpoint, nil},
```
sfc-gh-jmichalak: I think the first `awsDetails.UsePrivatelinkEndpoint` needs to be converted to a string type - to match the type in the schema.

pkg/resources/storage_integration_aws.go
```diff
@@ -109,22 +116,233 @@ func StorageIntegrationAws() *schema.Resource {
...
-func DummyStorageIntegrationAws(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
-	return nil
+func CreateStorageIntegrationAws(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+
+	name := d.Get("name").(string)
+	id := sdk.NewAccountObjectIdentifierFromFullyQualifiedName(name)
+	enabled := d.Get("enabled").(bool)
+	storageAllowedLocations, _ := parseLocations(d.Get("storage_allowed_locations").(*schema.Set).List())
+	storageProvider := strings.ToUpper(d.Get("storage_provider").(string))
```
sfc-gh-jmichalak: Nit: the sdk enum parsers already use `strings.ToUpper`.

pkg/acceptance/bettertestspoc/assert/resourceassert/storage_integration_aws_resource_ext.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func (s *StorageIntegrationAwsResourceAssert) HasStorageAllowedLocations(expected ...sdk.StorageLocation) *StorageIntegrationAwsResourceAssert {
+	s.AddAssertion(assert.ValueSet("storage_allowed_locations.#", strconv.FormatInt(int64(len(expected)), 10)))
+	for i, v := range expected {
+		s.AddAssertion(assert.ValueSet(fmt.Sprintf("storage_allowed_locations.%d", i), v.Path))
+	}
+	return s
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: \+1

pkg/resources/resource_helpers_update.go
sfc-gh-asawicki: I didn't want to rename, affecting all the files where it's used.

We will revisit these helpers either way, this or the next quarter, while working on improved generativeness. Leaving for now.

pkg/resources/storage_integration_aws.go
sfc-gh-asawicki: Added

pkg/resources/storage_integration_aws.go
sfc-gh-asawicki: It is for `field = ""`, you can check `TestAcc_StorageIntegrationAws_Validations`. I was hesitant to add it, as we are not doing it in other places. But there shouldn't be a harm in it.

pkg/resources/storage_integration_aws.go
sfc-gh-asawicki: that's true, however, somehow it's working without it. I'll adjust it either way.

pkg/resources/storage_integration_aws.go
sfc-gh-asawicki: True, adjusted

pkg/acceptance/bettertestspoc/assert/resourceassert/storage_integration_aws_resource_ext.go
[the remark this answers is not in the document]
sfc-gh-asawicki: This is already done but only for describe assertions in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4430 (and was done with https://github.com/snowflakedb/terraform-provider-snowflake/pull/4427 and https://github.com/snowflakedb/terraform-provider-snowflake/pull/4428 for other resources).

I think that for resource assertions (so based on the user config), they will be deterministic. Let's leave it like this and if any test fails because of it, then we'll remove it.

PR #4429, opened 2026-02-10
feat: Rework storage integrations data source

sfc-gh-asawicki:
- Add aggregated describe details to SDK
- Add aggregated describe details schema
- Rework storage integrations data source
- Add acceptance tests


pkg/sdk/generator/defs/storage_integrations_def.go
```diff
@@ -209,4 +209,20 @@ var storageIntegrationsDef = g.NewInterface(
 			Text("Comment").
 			Bool("UsePrivatelinkEndpoint").
 			Text("ServiceAccount"),
+		g.PlainStruct("StorageIntegrationAllDetails").
```
sfc-gh-jmichalak, line 212: Is this one just a merge of the three cloud providers? I was thinking we could e.g. nest the fields for a given provider. So that gcs integrations don't have the `use_privatelink_endpoint`, for example.

sfc-gh-jmichalak (dismissed):

pkg/sdk/generator/defs/storage_integrations_def.go
sfc-gh-asawicki: it is. I was considering a few options here, left a comment in `storage_integrations_ext.go` to `extract common mapping logic` - it includes considering struct changes. On the other hand, these are simple structs, and we generate them from them currently, so maybe it's good enough (at least for now).

PR #4431, opened 2026-02-11
feat: Final changes in stage resources

sfc-gh-jmichalak:
- Rename all stage resources from `snowflake_external_<cloud>_stage` / `snowflake_internal_stage` to `snowflake_stage_external_<cloud>` / `snowflake_stage_internal` for consistent naming
- Add `file_format` attribute to all four external stage resources (Azure, GCS, S3, S3-compatible) with support for named file formats and inline CSV options
- Move `file_format` schema from internal-stage-specific code to `stageCommonSchema` so it is shared across all stage types
- Extract common file format test model builders and resource assertions into shared `stage_common.go` files
- Add acceptance tests for file format CRUD operations on each external stage resource (create with format name, update to inline CSV, import)
- Update documentation, examples, migration guide, and preview feature constants to reflect the new resource names


pkg/resources/external_s3_stage.go
```diff
@@ -209,6 +209,10 @@ func ExternalS3Stage() *schema.Resource {
 			ForceNewIfChangeToEmptyString("aws_access_point_arn"),
 			RecreateWhenStageTypeChangedExternally(sdk.StageTypeExternal),
 			RecreateWhenStageCloudChangedExternally(sdk.StageCloudAws),
+			// This is the same configuration as for external S3 stage, but the additional differences are:
```
sfc-gh-jcieslak, line 212: This is the same comment as in `s3 compat`,and according to it, the force new fields are not in this resource for those fields. Adjust the comment or remove it.

sfc-gh-jmichalak: Right, I'll adjust it in the data source PR.

Issue #4445, opened 2026-02-16
[Bug]: Timestamp parsing error in snowflake_stage_external_s3   during state refresh


coopawhite:
### Terraform CLI Version

1.12.1

### Terraform Provider Version

2.12.0

### Terraform Configuration

```terraform
resource "snowflake_stage_external_s3" "events" {
  count = var.snowflake_external_id != null ? 1 : 0

  database            = var.snowflake_database
  schema              = var.snowflake_schema
  name                = upper("${local.snowflake_object_name}_s3_stage")
  url                 = local.s3_bucket_url
  storage_integration = snowflake_storage_integration_aws.events.name

  comment = "Stage for ${var.service_name} events from ${var.environment} environment"
  lifecycle {
    ignore_changes = [directory]
  }
}
```

### Category

category:resource

### Expected Behavior

  - `terraform plan` should successfully refresh the state of
  existing `snowflake_stage_external_s3` resources
  - The provider should correctly parse timestamps returned by
  Snowflake's API in either "Z" (UTC) or offset format

### Actual Behavior

  - `terraform plan` fails with timestamp parsing error
  - Unable to manage existing stage resources through normal
  Terraform workflows
  - Forces use of workarounds to bypass state refresh

### Steps to Reproduce

  ```hcl
  resource "snowflake_storage_integration" "example" {
    name    = "example_integration"
    type    = "EXTERNAL_STAGE"
    enabled = true

    storage_provider         = "S3"
    storage_aws_role_arn     =
  "arn:aws:iam::123456789012:role/example-role"
    storage_allowed_locations = ["s3://example-bucket/"]
  }

  resource "snowflake_stage_external_s3" "example" {
    name                = "example_stage"
    url                 = "s3://example-bucket/"
    storage_integration =
  snowflake_storage_integration.example.name
  }
```
  After initial creation, running terraform plan triggers the
  error.

  Workarounds

  The following workarounds have been identified, though none are
  ideal:

  1. Skip state refresh: Use terraform apply -refresh=false to
  bypass state refresh
  2. Drop and recreate: Manually drop the stage and recreate it in
  the same apply
  3. Conditional creation: Use count to defer stage creation,
  though this doesn't help with existing resources

### How much impact is this issue causing?

High

### Logs

```
**Plan: 0 to add, 1 to change, 0 to destroy.
╷
[...]

PR #4448, opened 2026-02-16
fix: Fix timestamp parsing in stage resources

sfc-gh-jmichalak:
## Changes
The stage resources introduced in v2.13.0 parsed the `LAST_REFRESHED_ON` directory table timestamp using a rigid Go time format, but Snowflake can return variant formats (e.g. `"Z"` instead of `"-0700"`). This caused `DESCRIBE STAGE` reads to fail, tainting resources and potentially leading to unwanted recreation.

- Change `StageDirectoryTable.LastRefreshedOn` from `*time.Time` to `*string` to store the raw timestamp without parsing
- Expose `last_refreshed_on` as a new computed field in `directory_table` within `describe_output` for all stage types
- Add state upgraders to all five stage resources (S3, Azure, GCS, S3-compatible, Internal) to re-populate `describe_output` on provider upgrade
- Extract shared `directoryTableOutputMapping` helper to deduplicate directory table handling across all stage Read functions
- Add regression test covering the v2.13.0 → v2.14.0 upgrade path with directory tables enabled
- Add migration guide entry documenting the fix and the `terraform untaint` workaround

Reference: [#4445](https://github.com/snowflakedb/terraform-provider-snowflake/issues/4445).

### DIFF
--- a/pkg/testacc/resource_external_s3_stage_account_level_acceptance_test.go
+++ b/pkg/testacc/resource_external_s3_stage_account_level_acceptance_test.go
@@ -32,7 +32,10 @@ func TestAcc_ExternalS3Stage_DirectoryTableTimestampParsing(t *testing.T) {
 	awsUrl := testenvs.GetOrSkipTest(t, testenvs.AwsExternalBucketUrl)
 	storageIntegrationId := ids.PrecreatedS3StorageIntegration
 
-	revertParameter := secondaryTestClient().Parameter.UpdateAccountParameterTemporarily(t, sdk.AccountParameterTimestampOutputFormat, "YYYY-MM-DD HH24:MI:SS")
+	incompatibleFormat := "YYYY-MM-DD HH24:MI:SS"
+	compatibleFormat := "YYYY-MM-DD HH24:MI:SS.FF3 TZHTZM"
+
+	revertParameter := secondaryTestClient().Parameter.UpdateAccountParameterTemporarily(t, sdk.AccountParameterTimestampOutputFormat, incompatibleFormat)
 	t.Cleanup(revertParameter)
 
 	providerModel := providermodel.SnowflakeProvider().WithProfile(testprofiles.Secondary).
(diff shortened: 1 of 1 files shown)


pkg/resources/stages_common_state_upgraders.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+	"fmt"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/schemas"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func v2_14_0_StageStateUpgrader(describeToSchema func(sdk.StageDetails) (map[string]any, error)) schema.StateUpgradeFunc {
```
sfc-gh-asawicki, line 13: Q: do we need this upgrader? It will be repopulated during the refresh phase either way, right? Or is there some problem we are solving here?

MIGRATION_GUIDE.md
```diff
@@ -43,6 +43,31 @@ No configuration changes are required. Existing configurations will continue to
...
 Reference: [#3956](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3956), [#4437](https://github.com/snowflakedb/terraform-provider-snowflake/issues/4437)
 
+### *(bugfix)* Fixed timestamp parsing in stage resources
+
+The new stage resources introduced in v2.13.0 parsed the `last_refreshed_on` timestamp column. However, due to flexibility of the time formats, this could fail with errors like
+```
+│ Error: parsing time "2026-02-15 23:59:47.000 Z" as "2006-01-02 15:04:05.000 -0700": cannot parse "Z" as "-0700"
+```
+
+This caused Terraform to detect a diff on the next plan and taint the resource, potentially leading to an unwanted recreation.
```
sfc-gh-asawicki, line 53: So why the recreation could happen here in the first place when the `last_refreshed_on` could not be parsed? Is it because one of the other customized diffs was executed then? If so, then this logic needs to be fixed as error like this one shouldn't taint the resource.

sfc-gh-jmichalak: Terraform automatically tains a resource when it fails durring terraform apply. Here, it failed because the time.Parse failed. This causes the recreation plan.

pkg/resources/stages_common_state_upgraders.go
sfc-gh-jmichalak: Yes. Otherwise, the resource is updated due to changes in handling directory_table. I altered acceptance tests to be more informative.

PR #4459, opened 2026-02-17
feat: Network rule rework

sfc-gh-jmichalak:
## Changes

The rework of `snowflake_network_rule`.

- Migrate resource ID from pipe-separated format to fully qualified Snowflake identifiers
- Add `show_output` and `describe_output` computed attributes to the resource schema
- Make `ValueList` optional in `NetworkRuleSetRequest` SDK definition
- Adjust Network Rule set.
- Add error when generator filter matches no objects.

### DIFF
--- a/docs/resources/network_rule.md
+++ b/docs/resources/network_rule.md
@@ -41,7 +41,7 @@ resource "snowflake_network_rule" "rule" {
 - `name` (String) Specifies the identifier for the network rule; must be unique for the database and schema in which the network rule is created.
 - `schema` (String) The schema in which to create the network rule.
 - `type` (String) Specifies the type of network identifiers being allowed or blocked. A network rule can have only one type. Allowed values are determined by the mode of the network rule; see https://docs.snowflake.com/en/sql-reference/sql/create-network-rule#required-parameters for details. Valid values are (case-insensitive): `IPV4` | `AWSVPCEID` | `AZURELINKID` | `GCPPSCID` | `HOST_PORT` | `PRIVATE_HOST_PORT`.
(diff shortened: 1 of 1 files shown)


pkg/sdk/network_rules_gen.go
```diff
@@ -73,17 +73,16 @@ type ShowNetworkRuleOptions struct {
...
-	OwnerRoleType      string `db:"owner_role_type"`
+	CreatedOn          time.Time `db:"created_on"`
+	Name               string    `db:"name"`
+	DatabaseName       string    `db:"database_name"`
+	SchemaName         string    `db:"schema_name"`
+	Owner              string    `db:"owner"`
+	Comment            string    `db:"comment"`
+	Type               string    `db:"type"`
+	Mode               string    `db:"mode"`
+	EntriesInValuelist int       `db:"entries_in_valuelist"`
```
sfc-gh-jmichalak, line 84: This is on purpose.

pkg/resources/network_rule.go
```diff
@@ -45,7 +46,7 @@ var networkRuleSchema = map[string]*schema.Schema{
 		DiffSuppressFunc: NormalizeAndCompare(sdk.ToNetworkRuleType),
 	},
 	"value_list": {
-		Type:        schema.TypeSet,
+		Type:        schema.TypeList,
```
sfc-gh-asawicki: q: can this collection be empty? (i.e. can I create with empty list of allowed values?)

sfc-gh-asawicki: Let's consult this change (list to set) internally. If the returned ordering is guaranteed, that's fine. If not, let's keep it a set to avoid constant drifts.

sfc-gh-jmichalak: Reverted - still waiting for one small verification.

sfc-gh-jmichalak: Yes.

PR #4460, opened 2026-02-17
feat: Add network rules data source

sfc-gh-jmichalak:
Add the missing network rule data source (preview).

### DIFF
--- a/pkg/testacc/data_source_network_rules_acceptance_test.go
+++ b/pkg/testacc/data_source_network_rules_acceptance_test.go
@@ -143,6 +143,7 @@ func TestAcc_NetworkRules_CompleteUseCase(t *testing.T) {
 			{
 				Config: config.DatasourceFromModel(t, withDescribe),
 				Check: assertThat(t,
+					assert.Check(resource.TestCheckResourceAttr(withDescribe.DatasourceReference(), "network_rules.#", "1")),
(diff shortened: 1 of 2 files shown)


pkg/testacc/data_source_network_rules_acceptance_test.go
```diff
@@ -0,0 +1,159 @@
...
+						HasName(nr.Name).
+						HasDatabaseName(id.DatabaseName()).
+						HasSchemaName(id.SchemaName()).
+						HasComment(comment).
+						HasType(sdk.NetworkRuleTypeIpv4).
+						HasMode(sdk.NetworkRuleModeIngress).
+						HasEntriesInValueList(2),
+
+					assert.Check(resource.TestCheckResourceAttr(withoutDescribe.DatasourceReference(), "network_rules.#", "1")),
+					assert.Check(resource.TestCheckResourceAttr(withoutDescribe.DatasourceReference(), "network_rules.0.describe_output.#", "0")),
```
sfc-gh-asawicki, line 140: q: what about describe output assertions?

sfc-gh-jmichalak: This step doesn't include describe_output, so we assert len=0. Added more assertions, but to the second step.

PR #4462, opened 2026-02-18
feat: Add import validation for snowflake_grant_privileges_to_account_role

sfc-gh-jmichalak:
## Changes

Importing grant resources with specific privileges can silently succeed even when the specified privileges don't actually exist in Snowflake, leading to state drift discovered only on the next plan/apply. This adds an opt-in validation step during import to catch mismatches early.

- Add `GRANTS_IMPORT_VALIDATION` experimental feature that validates privileges during import of `snowflake_grant_privileges_to_account_role`
- When enabled, importing with a fixed set of privileges checks that those privileges actually exist in Snowflake with the correct `with_grant_option` setting, erroring immediately on mismatch
- Extract `computePrivileges` helper from `ReadGrantPrivilegesToAccountRole` and reuse it in import validation
- Refactor `ImportGrantPrivilegesToAccountRole` from closure to direct function
- Add acceptance tests covering mismatched privileges, disabled validation, valid import, and interaction with strict privilege management
- Update migration guide and experimental features documentation

The validation logic reuses the existing `computePrivileges` and `prepareShowGrantsRequestForAccountRole` functions from the Read path, ensuring consistent privilege computation. The feature is gated behind the `GRANTS_IMPORT_VALIDATION` experiment so it has no impact on existing users. It intentionally passes `strictPrivilegeManagement=false` to `computePrivileges` during import validation, since the goal is to verify the specified privileges exist regardless of extra grants.

## References
- https://github.com/snowflakedb/terraform-provider-snowflake/issues/3944

### DIFF
--- a/pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
+++ b/pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
@@ -3300,10 +3300,11 @@ func TestAcc_GrantPrivilegesToAccountRole_ImportValidation_Disabled(t *testing.T
 			},
 			// Import without experiment enabled - should succeed (default behavior preserved)
 			{
-				Config:                  accconfig.FromModels(t, resourceModel),
-				ResourceName:            resourceModel.ResourceReference(),
-				ImportState:             true,
-				ImportStateVerify:       true,
+				Config:            accconfig.FromModels(t, resourceModel),
+				ResourceName:      resourceModel.ResourceReference(),
+				ImportState:       true,
+				ImportStateVerify: true,
+				// Privileges are not verified because the config is not matching the current state, and this field is overridden by the read function.
 				ImportStateVerifyIgnore: []string{"account_role_name", "on_account_object.0.object_name", "privileges"},
 			},
 		},


pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -3220,3 +3220,207 @@ func TestAcc_GrantPrivilegesToAccountRole_StrictPrivilegeManagement_OnFuture_Non
...
+		CheckDestroy:             CheckAccountRolePrivilegesRevoked(t),
+		Steps: []resource.TestStep{
+			{
+				Config:        accconfig.FromModels(t, providerModel, resourceModel),
+				ResourceName:  "snowflake_grant_privileges_to_account_role.test",
+				ImportState:   true,
+				ImportStateId: importId,
+				ExpectError:   regexp.MustCompile("privileges granted in Snowflake do not match the expected privileges"),
+			},
+			// grant the privilege, but with incorrect grant option
```
sfc-gh-jcieslak: Should this step fail? I know there's the monitor privilege granted to the role, but the required set of privileges used within the resource is there 🤔

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -3220,3 +3220,207 @@ func TestAcc_GrantPrivilegesToAccountRole_StrictPrivilegeManagement_OnFuture_Non
...
+		WithWithGrantOption(false)
+
+	resource.Test(t, resource.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+		CheckDestroy:             CheckAccountRolePrivilegesRevoked(t),
+		Steps: []resource.TestStep{
+			{
```
sfc-gh-jcieslak, line 3296: Do we need this step? I think the 2nd step is sufficient in this case (but have to adjust the import to pass resource id as in the test above

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -3220,3 +3220,207 @@ func TestAcc_GrantPrivilegesToAccountRole_StrictPrivilegeManagement_OnFuture_Non
...
+			},
+			// Import with matching ID should succeed with experiment enabled
+			{
+				Config:                  accconfig.FromModels(t, providerModel, resourceModel),
+				ResourceName:            resourceModel.ResourceReference(),
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"account_role_name", "on_account_object.0.object_name", "strict_privilege_management"},
+			},
+			// Import with matching ID should succeed with experiment enabled
```
sfc-gh-jcieslak: I'm a bit confused. The `TestAcc_GrantPrivilegesToAccountRole_ImportValidation_MismatchedPrivilege` test fails due to a privilege mismatch, even though the privileges are not defined in the config, whereas this one passes. Is this test case incorrect?

pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -347,105 +347,139 @@ func GrantPrivilegesToAccountRole() *schema.Resource {
...
-		if err := d.Set("with_grant_option", id.WithGrantOption); err != nil {
-			return nil, err
-		}
-		if err := d.Set("always_apply", id.AlwaysApply); err != nil {
-			return nil, err
-		}
-		if err := d.Set("all_privileges", id.AllPrivileges); err != nil {
+func ImportGrantPrivilegesToAccountRole(ctx context.Context, d *schema.ResourceData, m any) ([]*schema.ResourceData, error) {
+	id, err := ParseGrantPrivilegesToAccountRoleId(d.Id())
+	if err != nil {
```
sfc-gh-asawicki, line 352: nit: errors.Join?

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-asawicki: +1, I thought that this step would succeed, it's stricter than discussed earlier (or maybe I misunderstood the earlier discussions)

**EDIT**: the mislieading part is the monitor privilege, let's improve this test so that additional privilege is not a factor but only the with_grant_option state

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-jmichalak: Done.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jmichalak: This one passes because the privileges are matching (removed PreConfig, as it was a leftover). That one doesn't pass because with grant option doesn't match.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jmichalak: I'm not sure what you mean. It works like this:  
1\. grant an X privilege  
2\. step1: import with Y, expect failure because Y is not granted at all  
3\. preconfig step2: grant Y, but with wrong grant option  
4\. step2: import with Y, expect failure because Y is granted with wrong grant option  
  
Anyway, added a revoke call before the second step.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jmichalak: Yes. Even after setting ImportStatePersist, the state is empty for some reason, and the revoke fails. In other places where we use ImportStatePersist, we have apply steps next, so maybe that's why.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -3220,3 +3220,206 @@ func TestAcc_GrantPrivilegesToAccountRole_StrictPrivilegeManagement_OnFuture_Non
...
+		WithPrivileges(string(sdk.AccountObjectPrivilegeMonitor)).
+		WithOnAccountObject(sdk.ObjectTypeDatabase, database.ID()).
+		WithStrictPrivilegeManagement(true).
+		WithWithGrantOption(false)
+
+	resource.Test(t, resource.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		ProtoV6ProviderFactories: grantsImportValidationProviderFactory,
```
sfc-gh-asawicki: if you turn on two experiments (differently than in other tests), you need a different factory here

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-asawicki: I meant, that if we are testing, we need to validate these in isolation. The isolated use case is: there is only one privilege but with different with_grant_option, so exactly what you added.

I went through the tests and I dont see (if I missed it please point me to it), a test, without strict handling in which: there are 2 privileges on SF, and we try importing with only one of them (correct with_grant option). My expectation is, that this import should be successful, but I think that with current implementation it will fail.

sfc-gh-jmichalak: Added - it passes.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jmichalak: Done.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak, line 3308: Question: Shouldn't it be empty? Ignore == mismatch, and privileges and other values should be the same, right? I can see other tests do similar thing, but that just caught my eye

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jcieslak: Still, I think we don't have a test where two privileges are granted on the Snowflake side, and we would like to import a resource that handles one of them, right? It would show that we don't validate privileges that are not in the resource config

sfc-gh-asawicki: We do now (that's what I asked for in my above comment). From my perspective, this thread could be resolved.

sfc-gh-jcieslak: Yeah, I can see the step in _Valid test

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-jmichalak: This is the case that has been reported. So here, we don't validate the privileges. They actually don't get imported because they are cleared during Read. (or, they can be validated with this flag we're adding)

sfc-gh-jcieslak (approved):

PR #4463, opened 2026-02-18
chore: Rework external volumes SDK

sfc-gh-jmichalak:
The external volumes SDK had several gaps compared to the Snowflake SQL reference — missing S3-compatible storage location support, missing optional fields for S3 and Azure providers, and helper functions that lived outside the SDK package. Meanwhile, the network rule resource hardcoded enum values and rejected valid newer options. This PR addresses both.

External Volumes SDK rework:
- Add S3-compatible (S3COMPAT) storage location support across the SDK.
- Add missing S3 and Azure fields
- Rename encryption struct field from `Type` to `EncryptionType` to avoid confusion with the SQL column name
- Move some helpers, and parsing/validation logic from `pkg/helpers` into `pkg/sdk`
- Regenerate SDK files with the updated generator definition
- Fix bug in `external_volume_model_ext.go` where Azure storage location read `StorageBaseUrl` from GCS params
- Update integration tests to use `sdk.ParseExternalVolumeDescribed` instead of `helpers.ParseExternalVolumeDescribed`

## TODO
- Rework resource
- Improve integration tests
- Add data source

### DIFF
--- a/pkg/sdk/external_volumes_ext.go
+++ b/pkg/sdk/external_volumes_ext.go
@@ -23,12 +23,14 @@ func CopySentinelStorageLocation(
 	case StorageProviderS3, StorageProviderS3GOV:
 		tempNameStorageLocation = ExternalVolumeStorageLocation{
 			S3StorageLocationParams: &S3StorageLocationParams{
-				Name:                 newName,
-				StorageProvider:      storageLocation.S3StorageLocationParams.StorageProvider,
-				StorageBaseUrl:       storageLocation.S3StorageLocationParams.StorageBaseUrl,
-				StorageAwsRoleArn:    storageLocation.S3StorageLocationParams.StorageAwsRoleArn,
-				StorageAwsExternalId: storageLocation.S3StorageLocationParams.StorageAwsExternalId,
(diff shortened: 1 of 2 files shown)


pkg/sdk/external_volumes_ext.go
```diff
@@ -0,0 +1,235 @@
...
+		return StorageProviderS3Compatible, nil
+	default:
+		return "", fmt.Errorf("Invalid storage location")
+	}
+}
+
+// Returns the index of the last matching elements in the list
+// e.g. [1,2,3] [1,3,2] -> 0, [1,2,3] [1,2,4] -> 1
+// -1 is returned if there are no common prefixes in the list
+func CommonPrefixLastIndex(a []ExternalVolumeStorageLocation, b []ExternalVolumeStorageLocation) (int, error) {
```
sfc-gh-asawicki, line 116: nit: should it be part of our collection helpers?

sfc-gh-jmichalak: Maybe, I'll take a closer look in later PRs. For now, I just wanted to move existing code to the proper location.

PR #4467, opened 2026-02-19
chore: Fix convert guard template

sfc-gh-jmichalak:
## Summary
- Fixed `convert_guard.tmpl` to include a trailing newline so generated `var _ convertibleRow[...]` declarations are separated by blank lines
- This prevents `gofumpt` from grouping consecutive `var` declarations into a `var()` block when saving generated `_impl_gen.go` files
- Regenerated Stages and Notebooks as verification (in stages, the format was different by mistake).


sfc-gh-jmichalak:
1. By aligning the generator output with gofumpt’s expectations, we eliminate diffs for anyone who has 'Format on Save' enabled, regardless of their IDE. Right now, the generator is producing code that gofumpt considers unformatted.
2. Asking every developer to change their global IDE settings (or the IDE itself) or remember to disable formatting for specific files is suboptimal.
3. This change doesn't affect the logic; it's only about formatting, so there's basically zero risk.

sfc-gh-asawicki:
We have `make fmt` (part of `make pre-push`) which is basically `tools/bin/gofumpt -l -w .` and it doesn't do the reformatting in this case. We shouldn't adjust how we generate files based on the single IDE making the changes whereas our pipelines/makefile/linters set up for the project are not messing it.

The purpose of the project-level formatting is that no matter what IDE you use, you should achieve the same results.

It's a bit like we would have proper build instructions in our Makefile, and then someone would say that building it in the IDE (with a different command) is failing, so we need to adjust the project.

Imagine this:
1. We make this change.
1. Then someone refactors the templates.
1. Our `make fmt` passes.
1. We push it.
1. You start working and see changes in tens of files because your formatting rules differ from the project-wide ones.

Let's do it differently.
1. One of the problems here is that `tools/bin/gofumpt -l -w .` and `./tools/bin/gofumpt -l -w pkg/sdk/notebooks_impl_gen.go` behave differently in terms of this var formatting (I checked that adding a stupid whitespace is recognized in both options, but vars are formatted only if this file is specified).
1. The main source of the problem is that the generated files are programmatically formatted using gofmt ([here](https://github.com/snowflakedb/terraform-provider-snowflake/blob/26478e583eab77b8cabce7e446afca79004051f9/pkg/internal/genhelpers/util.go#L65)) and not by gofumpt as the rest of the project. I see that gofumpt offers a similar method ([here](https://github.com/mvdan/gofumpt/blob/d802ec05c9dad106e8eb09e980986152026f55ff/format/format.go#L126)) we can add to our genutils.

This way, we solve this problem in an IDE-agnostic way, the generated code follows the gofumpt used in the rest of the project, and we can be a bit sloppier with the templates. I'll send a PR for that.

sfc-gh-jmichalak:
> tools/bin/gofumpt -l -w . and it doesn't do the reformatting in this case

It does when you specify a filename in this command. 

> single IDE making the changes

It's not a single IDE. It's the behavior of gofumpt. Having `.vscode` configuration in the repo forces me to use repository settings. On the one hand, VSCode is kinda supported by having this file (especially with forcing gofumpt on single files); on the other hand, the files are not compatible with it. That's why I proposed an alternative PR that removes the requirement to use gofumpt. Also, VSCode is literally the most popular IDE, and we have external contributors, so just saying "use another IDE" isn't the best solution here.

> It's a bit like we would have proper build instructions in our Makefile, and then someone would say that building it in the IDE (with a different command) is failing, so we need to adjust the project.

No, I'm not saying this. Imagine we have a configuration in the repo that changes `go build` or other arguments for some reason, and it breaks code compilation in an IDE. In that case, I also would change such a configuration. Besides, we already have these `gotype` annotations, which are Goland-specific, I guess.

I'm happy about the change in generator formatting, though. Closing this PR, as this has been adjusted in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4470

PR #4470, opened 2026-02-19
chore: Format generated code using gofumpt

sfc-gh-asawicki:
References: #4467 

Format the generated code the same way as we format manually written code (gofumpt)
- Use gofumpt formatting in genhelpers
- Verify by:
  - Regenerating SDK (without unit tests regeneration): `make generate-sdk SF_TF_GENERATOR_ARGS='--filter-generation-part-names=default,impl,dto,dto_builders,validations'`
  - Regenerating model builders: `make pre-push`

Next PRs:
- check in assertion generation

### DIFF
--- a/pkg/sdk/api_integrations_impl_gen.go
+++ b/pkg/sdk/api_integrations_impl_gen.go
@@ -10,8 +10,10 @@ import (
 
 var _ ApiIntegrations = (*apiIntegrations)(nil)
 
-var _ convertibleRow[ApiIntegration] = new(showApiIntegrationsDbRow)
-var _ convertibleRow[ApiIntegrationProperty] = new(descApiIntegrationsDbRow)
+var (
+	_ convertibleRow[ApiIntegration]         = new(showApiIntegrationsDbRow)
+	_ convertibleRow[ApiIntegrationProperty] = new(descApiIntegrationsDbRow)
+)
 
 type apiIntegrations struct {
 	client *Client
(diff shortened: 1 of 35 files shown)


pkg/internal/genhelpers/util.go
```diff
@@ -67,6 +69,11 @@ func WriteCodeToFile(buffer *bytes.Buffer, fileName string) error {
 		return fmt.Errorf("writing code to file %s failed with err: %w", fileName, err)
```
sfc-gh-jmichalak, line 69: <p class='graphite__hidden'><i>[Re: line +67]</i></p>
gofumpt is stricter than gofmt, so this call can be removed.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/snowflakedb/terraform-provider-snowflake/4470?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/internal/genhelpers/util.go
```diff
@@ -67,6 +69,11 @@ func WriteCodeToFile(buffer *bytes.Buffer, fileName string) error {
...
 	}
 
+	src, err = gofumptformat.Source(src, gofumptformat.Options{})
+	if err != nil {
+		return fmt.Errorf("writing code to file %s failed with err: %w", fileName, err)
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: \+1

pkg/internal/genhelpers/util.go
sfc-gh-asawicki: I left in on purpose. It's stricter but wasn't sure if it will format the same way, so to avoid more changes left two. It should be removed in the future, closer to when we have fewer manual changes. I'll add a TODO for it.

PR #4471, opened 2026-02-19
chore: Regenerate, fix, and adjust resource assertions

sfc-gh-asawicki:
Regenerate, fix, and adjust:
- resource assertions (`make clean-resource-assertions generate-resource-assertions`)
  - for stages assertions: removed `HasFileFormatEmpty` from ext as it is being generated
  - rename `ExternalS3CompatStageResourceAssert` to `ExternalS3CompatibleStageResourceAssert` (this is how it's generated; adjusted feature enum too)
  - rename notebook method so that it does not conflict with the generated one
- resource show output assertions (` make clean-resource-show-output-assertions generate-resource-show-output-assertions`)
  - add TODO for improper `sdk.` classification (not all such objects are enums - causing improper generation)
  - add missing network rule show output assertions
  - add missing user workload identity authentication method show output assertions
  - adjust cloud assertion for stages
- resource parameter assertions (`make clean-resource-parameters-assertions generate-resource-parameters-assertions`)
  - no changes (there are still two files with manual adjustments)


pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/network_rule_show_output_gen.go
```diff
@@ -0,0 +1,143 @@
+// Code generated by resource show output assertions generator (v0.1.0); DO NOT EDIT.
```
sfc-gh-jmichalak, line 1: This file is added in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4459/changes

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/user_workload_identity_authentication_method_show_output_gen.go
```diff
@@ -0,0 +1,137 @@
+// Code generated by resource show output assertions generator (v0.1.0); DO NOT EDIT.
```
sfc-gh-jmichalak, line 1: Ultimately, this file shouldn't be added because we don't have a resource for WIFs. Let's add a TODO to iterate resources, not objects in this generator.

sfc-gh-jmichalak (dismissed):

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/network_rule_show_output_gen.go
sfc-gh-asawicki: yup

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/user_workload_identity_authentication_method_show_output_gen.go
sfc-gh-asawicki: The use of struct is on purpose; it allows us to create asserts using objects/enums. It's currently a big limitation on the normal resource assertions using tf schemas as a base. Ultimately, I would like to move every generation to structs (schemas will be also generated out of them, and this is the world in which we already generate all the stuff based on the object definition).

Currently, the problem is that both object assert gen and show output use the same list. I will filter it out in the next PR.

sfc-gh-jmichalak: I'm saying that now we assume that each SDK object added to the list in objectassert has a related resource, which is not true in this case. That's why we have this file, but in the end, we shouldn't.

sfc-gh-jmichalak (approved):

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/user_workload_identity_authentication_method_show_output_gen.go
sfc-gh-asawicki: Yes:
> Currently, the problem is that both object assert gen and show output use the same list. I will filter it out in the next PR.

Issue #4398, opened 2026-01-30
[Bug]: Updating Share results in a TEMP_ db error


prae04:
### Terraform CLI Version

opentofu 1.10.5

### Terraform Provider Version

2.12.0

### Terraform Configuration

```terraform
resource "snowflake_share" "my-share" {
  name     = "MY_SHARE"
  accounts = ["1234.my-account"]
  provider = snowflake.tofu_svc
}
```

### Category

category:preview

### Object type(s)

resource:share

### Expected Behavior

An existing Share *only* is being updated from an empty list of `accounts` to a non-empty list of 1 account (1234.my-account)

The Share should be updated in-place, containing the existing databases/tables.  I did not provide those grant resources since they are already applied, successfully in a previous deployment.

### Actual Behavior

Error during apply:
```
module.share-accts["demo_acct"].snowflake_share.my_share: Modifying... [id=MY_SHARE]

│ Error: error granting privilege to share (MY_SHARE) err = 003033 (0A000): SQL compilation error:
│ Database 'TEMP_MY_DB_1769788429' does not belong to the database that is being shared.
│ 
│   with module.share-accts["demo_acct"].snowflake_share.my-share,
│   on ../../modules/share-acct/main.tf line 55, in resource "snowflake_share" "my-share":
│   55: resource "snowflake_share" "my-share" {
│ 
```

### Steps to Reproduce

1. An existing `share` deploys successfully, with an empty `accounts` value.
2. `grants_to_share` deploy successfully (USAGE on 1 db and 1 schema)
3. A second plan/apply with ONLY the Accounts changing from empty to a list of 1 value.

### How much impact is this issue causing?

High

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

prae04:
I read the Note [here](https://github.com/snowflakedb/terraform-provider-snowflake/blob/b91c1b29cd2a777e8ac988b9faab313bca045b99/templates/resources/share.md.tmpl#L14C177-L14C204)
The Share does have database grants immediately after creation, so looks like this `TEMP_` db and the existing database granted to the Share are clashing?

prae04:
Given the ERROR message, I'm not clear on how a database can `belong` to another database?

sfc-gh-jcieslak:
Hey @prae04 👋 
[...]

PR #4473, opened 2026-02-20
fix: issue 4398

sfc-gh-jcieslak:
## Changes
- Fix the Update logic in the share resource to only use the temporary database workaround whenever no other database is granted to a given share
- Add the test proving it now works, and the previous behavior of using the workaround on Update (when no database is assigned to the share) still works

References: #4398

### DIFF
--- a/pkg/resources/share.go
+++ b/pkg/resources/share.go
@@ -230,7 +230,25 @@ func UpdateShare(ctx context.Context, d *schema.ResourceData, meta any) diag.Dia
 			}
 		} else {
 			accountIdentifiers := accountIdentifiersFromSlice(newAccounts)
-			err := setShareAccounts(ctx, client, id, accountIdentifiers)
+
+			share, err := client.Shares.ShowByID(ctx, id)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/resources/share.go
```diff
@@ -230,7 +230,25 @@ func UpdateShare(ctx context.Context, d *schema.ResourceData, meta any) diag.Dia
...
+			share, err := client.Shares.ShowByID(ctx, id)
+			if err != nil {
+				return diag.FromErr(fmt.Errorf("error reading share (%v) err = %w", d.Id(), err))
+			}
+
+			// If a database is already granted to the share, we can add accounts directly.
+			// The setShareAccounts workaround (which creates a temp database) is only needed
+			// when no database has been granted yet — Snowflake requires at least one database
+			// on a share before accounts can be added.
+			if share.DatabaseName.Name() != "" {
```
sfc-gh-jmichalak, line 243: Nit: we can simplify this logic a bit

```
if share has no database {
    provide the share and register the cleanup
}
alter the share

```

Also, a general note that during the rework, if this workaround is still needed, we could read the show output's value here (now it's missing entirely).

sfc-gh-jmichalak (approved):

PR #4482, opened 2026-02-23
chore: External volumes SDK part 2

sfc-gh-jmichalak:
## Changes

- Promote `Name` to the shared `ExternalVolumeStorageLocation` struct, eliminating redundancy across all 4 storage provider types
- Introduce `ExternalVolumeStorageLocationItem` wrapper for correct DDL parenthesized rendering (nested parentheses).
- Add `ALTER ... UPDATE STORAGE_LOCATION` support for updating S3Compat credentials
- Extract `CommonPrefixLastIndex` from SDK-specific code into a generic collection helper with full unit tests
- Rename types for consistency: `ParsedExternalVolumeDescribed` → `ExternalVolumeDetails`, `ExternalVolumeStorageLocationJson` → `ExternalVolumeStorageLocationDetails`
- Add `ExternalVolumeDetailsAssert` for describe-level object assertions in integration tests
- Add `Describe` method to the external volume test client helper
- Convert generator definitions from global variables to functions to prevent accidental mutation
- Remove `GetStorageLocationName`.
- Update unit tests, integration tests, and resource code to use the new model

## TODO
- Adjust the JSON file def. Now we have an exported JSON struct with flattened values, but we want to group them by storage provider, something like in the stages.
- In ParseExternalVolumeDescribed, we assume the locations are in order. We don't parse the indexes.
- Rework integration tests - mainly adjust assertions.
- Rework resource and data source.

### DIFF
--- a/pkg/sdk/external_volumes_dto_builders_gen.go
+++ b/pkg/sdk/external_volumes_dto_builders_gen.go
@@ -80,14 +80,12 @@ func (s *AlterExternalVolumeSetRequest) WithComment(comment string) *AlterExtern
 	return s
 }
 
-func NewExternalVolumeStorageLocationItemRequest() *ExternalVolumeStorageLocationItemRequest {
+func NewExternalVolumeStorageLocationItemRequest(
+	externalVolumeStorageLocation ExternalVolumeStorageLocationRequest,
+) *ExternalVolumeStorageLocationItemRequest {
 	s := ExternalVolumeStorageLocationItemRequest{}
-	return &s
-}
-
-func (s *ExternalVolumeStorageLocationItemRequest) WithExternalVolumeStorageLocation(externalVolumeStorageLocation ExternalVolumeStorageLocationRequest) *ExternalVolumeStorageLocationItemRequest {
 	s.ExternalVolumeStorageLocation = externalVolumeStorageLocation
-	return s
+	return &s
 }
 
 func NewExternalVolumeStorageLocationRequest(
--- a/pkg/sdk/external_volumes_dto_gen.go
+++ b/pkg/sdk/external_volumes_dto_gen.go
@@ -34,7 +34,7 @@ type AlterExternalVolumeSetRequest struct {
 }
 
 type ExternalVolumeStorageLocationItemRequest struct {
-	ExternalVolumeStorageLocation ExternalVolumeStorageLocationRequest
+	ExternalVolumeStorageLocation ExternalVolumeStorageLocationRequest // required
 }
 
 type ExternalVolumeStorageLocationRequest struct {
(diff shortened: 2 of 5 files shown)


pkg/resources/external_volume.go
```diff
@@ -365,13 +367,12 @@ func UpdateContextExternalVolume(ctx context.Context, d *schema.ResourceData, me
 		// can be added back. The storage locations lower than index 5 don't need to be modified.
 		// The removal process could be done without the above recreation, but it handles this case
 		// too so it's used for both actions.
-		commonPrefixLastIndex, err := sdk.CommonPrefixLastIndex(newLocations, oldLocations)
-		if err != nil {
-			return diag.FromErr(err)
-		}
+		commonPrefixLastIndex := collections.CommonPrefixLastIndex(newLocations, oldLocations, func(a, b sdk.ExternalVolumeStorageLocationItem) bool {
```
sfc-gh-jcieslak, line 370: nit: Can we use function reference in this case? `collections.CommonPrefixLastIndex(newLocations, oldLocations, reflect.DeepEqual)`

sfc-gh-jcieslak (approved):

pkg/acceptance/helpers/external_volume_client.go
```diff
@@ -63,6 +63,20 @@ func (c *ExternalVolumeClient) Show(t *testing.T, id sdk.AccountObjectIdentifier
...
 }
 
+func (c *ExternalVolumeClient) Describe(t *testing.T, id sdk.AccountObjectIdentifier) (*sdk.ExternalVolumeDetails, error) {
+	t.Helper()
+	ctx := context.Background()
+	props, err := c.client().Describe(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	details, err := sdk.ParseExternalVolumeDescribed(props)
```
sfc-gh-asawicki, line 73: nit: why we call here the Describe and parse it, instead of adding the additional method to ExternalVolumes interface and do it once in one place?

sfc-gh-asawicki (approved):

pkg/resources/external_volume.go
sfc-gh-jmichalak: Yup, next PR.

pkg/acceptance/helpers/external_volume_client.go
sfc-gh-jmichalak: There is a drawback - another manual change in the generated functions. Eventually, I think this should be moved to `convert` with adjusted desc outputs, but this is another story.

pkg/resources/external_volume.go
sfc-gh-jmichalak: Ugh, actually not. CommonPrefixLastIndex expects a cmp parameter of `type func(T, T) bool` (where T is sdk.ExternalVolumeStorageLocationItem), but reflect.DeepEqual has the signature `func(x any, y any) bool`. In Go, function types must match exactly.

pkg/acceptance/helpers/external_volume_client.go
sfc-gh-asawicki: But this is a really simple change in the interface, impl goes to ext, and that way it's reusable everywhere. Please adjust in the next PRs.

PR #4483, opened 2026-02-23
feat: Add catalog integrations to SDK

sfc-gh-kwasilewski:
## Changes
- Add definition and generate SDK code.
- There are several limitations:
  - UNSET COMMENT does not work - instead we use SET COMMENT = ''.
  - UNSET REFRESH_INTERVAL_SECONDS does not work - instead we'll use SET REFRESH_INTERVAL_SECONDS = NULL. A wrapper will be implemented in the follow-up PR.
  - Many other properties are not supported by the ALTER command. We'll use CREATE OR REPLACE as a workaround.

## TODO
 - Add integration tests.

### DIFF
--- a/pkg/sdk/catalog_integrations_dto_gen.go
+++ b/pkg/sdk/catalog_integrations_dto_gen.go
@@ -87,7 +87,7 @@ type SapBdcParamsRequest struct {
 
 type SapBdcRestConfigRequest struct {
 	SapBdcInvitationLink string // required
-	AccessDelegationMode *string
+	AccessDelegationMode *CatalogIntegrationAccessDelegationMode
 }
 
 type AlterCatalogIntegrationRequest struct {
--- a/pkg/sdk/catalog_integrations_enums.go
+++ b/pkg/sdk/catalog_integrations_enums.go
@@ -77,7 +77,7 @@ func ToCatalogIntegrationRestAuthenticationType(s string) (CatalogIntegrationRes
 type CatalogIntegrationAccessDelegationMode string
 
 const (
-	CatalogIntegrationAccessDelegationModeVendedCredentials         CatalogIntegrationAccessDelegationMode = "VENDED_CREDENTIALS"
+	CatalogIntegrationAccessDelegationModeVendedCredentials         CatalogIntegrationAccessDelegationMode = "VENDED_CREDENTIALS" //nolint:gosec
 	CatalogIntegrationAccessDelegationModeExternalVolumeCredentials CatalogIntegrationAccessDelegationMode = "EXTERNAL_VOLUME_CREDENTIALS"
 )
 
@@ -102,7 +102,7 @@ const (
 	CatalogIntegrationCatalogApiTypeAwsApiGateway        CatalogIntegrationCatalogApiType = "AWS_API_GATEWAY"
 	CatalogIntegrationCatalogApiTypeAwsPrivateApiGateway CatalogIntegrationCatalogApiType = "AWS_PRIVATE_API_GATEWAY"
 	CatalogIntegrationCatalogApiTypeAwsGlue              CatalogIntegrationCatalogApiType = "AWS_GLUE"
-	CatalogIntegrationCatalogApiTypeAwsPrivateGlue       CatalogIntegrationCatalogApiType = "AWS_PRIVATE_GLAUE"
+	CatalogIntegrationCatalogApiTypeAwsPrivateGlue       CatalogIntegrationCatalogApiType = "AWS_PRIVATE_GLUE"
 )
 
 var AllCatalogIntegrationCatalogApiTypes = []CatalogIntegrationCatalogApiType{
--- a/pkg/sdk/catalog_integrations_gen_test.go
+++ b/pkg/sdk/catalog_integrations_gen_test.go
@@ -332,7 +332,7 @@ func TestCatalogIntegrations_Create(t *testing.T) {
 	t.Run("all options - SAP Business Data Cloud", func(t *testing.T) {
 		opts := defaultOptsSapBdc()
 		opts.IfNotExists = Bool(true)
-		opts.SapBdcCatalogSourceParams.RestConfig.AccessDelegationMode = String("VENDED_CREDENTIALS")
+		opts.SapBdcCatalogSourceParams.RestConfig.AccessDelegationMode = Pointer(CatalogIntegrationAccessDelegationModeVendedCredentials)
 		opts.Enabled = false
 		opts.RefreshIntervalSeconds = Int(60)
 		opts.Comment = String("test comment")
--- a/pkg/sdk/generator/defs/catalog_integrations_def.go
+++ b/pkg/sdk/generator/defs/catalog_integrations_def.go
@@ -21,7 +21,7 @@ var icebergRestRestConfigDef = g.NewQueryStruct("IcebergRestRestConfig").
 
 var sapBdcRestConfigDef = g.NewQueryStruct("SapBdcRestConfig").
 	TextAssignment("SAP_BDC_INVITATION_LINK", g.ParameterOptions().SingleQuotes().Required()).
-	OptionalTextAssignment("ACCESS_DELEGATION_MODE", g.ParameterOptions().NoQuotes())
+	OptionalAssignment("ACCESS_DELEGATION_MODE", g.KindOfT[sdkcommons.CatalogIntegrationAccessDelegationMode](), g.ParameterOptions().NoQuotes())
 
 var oAuthRestAuthenticationDef = g.NewQueryStruct("OAuthRestAuthentication").
 	SQLWithCustomFieldName("restAuthType", "TYPE = OAUTH").
--- a/pkg/sdk/catalog_integrations_dto_builders_gen.go
+++ b/pkg/sdk/catalog_integrations_dto_builders_gen.go
@@ -244,7 +244,7 @@ func NewSapBdcRestConfigRequest(
 	return &s
 }
 
-func (s *SapBdcRestConfigRequest) WithAccessDelegationMode(accessDelegationMode string) *SapBdcRestConfigRequest {
+func (s *SapBdcRestConfigRequest) WithAccessDelegationMode(accessDelegationMode CatalogIntegrationAccessDelegationMode) *SapBdcRestConfigRequest {
 	s.AccessDelegationMode = &accessDelegationMode
 	return s
 }
--- a/pkg/sdk/catalog_integrations_enums_test.go
+++ b/pkg/sdk/catalog_integrations_enums_test.go
@@ -170,7 +170,7 @@ func Test_ToCatalogIntegrationCatalogApiType(t *testing.T) {
 		{input: "AWS_API_GATEWAY", want: CatalogIntegrationCatalogApiTypeAwsApiGateway},
 		{input: "AWS_PRIVATE_API_GATEWAY", want: CatalogIntegrationCatalogApiTypeAwsPrivateApiGateway},
 		{input: "AWS_GLUE", want: CatalogIntegrationCatalogApiTypeAwsGlue},
-		{input: "AWS_PRIVATE_GLAUE", want: CatalogIntegrationCatalogApiTypeAwsPrivateGlue},
+		{input: "AWS_PRIVATE_GLUE", want: CatalogIntegrationCatalogApiTypeAwsPrivateGlue},
 	}
 
 	invalid := []test{
--- a/pkg/sdk/catalog_integrations_gen.go
+++ b/pkg/sdk/catalog_integrations_gen.go
@@ -110,8 +110,8 @@ type SapBdcParams struct {
 }
 
 type SapBdcRestConfig struct {
-	SapBdcInvitationLink string  `ddl:"parameter,single_quotes" sql:"SAP_BDC_INVITATION_LINK"`
-	AccessDelegationMode *string `ddl:"parameter,no_quotes" sql:"ACCESS_DELEGATION_MODE"`
+	SapBdcInvitationLink string                                  `ddl:"parameter,single_quotes" sql:"SAP_BDC_INVITATION_LINK"`
+	AccessDelegationMode *CatalogIntegrationAccessDelegationMode `ddl:"parameter,no_quotes" sql:"ACCESS_DELEGATION_MODE"`
 }
 
 // AlterCatalogIntegrationOptions is based on https://docs.snowflake.com/en/sql-reference/sql/alter-catalog-integration.


pkg/sdk/object_types.go
```diff
@@ -105,7 +105,8 @@ const (
 	ObjectTypeListingDetails ObjectType = "LISTING DETAILS"
 	// ObjectTypeApiIntegration is a pseudo-object, only used in object and invoke action assertions.
 	// For actual Snowflake operations where object type is needed, ObjectTypeIntegration should be used.
-	ObjectTypeApiIntegration ObjectType = "API INTEGRATION"
+	ObjectTypeApiIntegration     ObjectType = "API INTEGRATION"
+	ObjectTypeCatalogIntegration ObjectType = "CATALOG INTEGRATION"
 )
 
 func (o ObjectType) String() string {
```
sfc-gh-jmichalak, line 116: It should be also added here, in objectTypeSingularToPluralMap, as a PluralObjectType and in `object_types_test.go`.

pkg/sdk/generator/gen/sdkcommons/enums.go
```diff
@@ -74,3 +76,17 @@ const (
 	StorageProviderGCS   StorageProvider = "GCS"
 	StorageProviderAzure StorageProvider = "AZURE"
 )
+
+const (
```
sfc-gh-jmichalak: 1. These enums should be defined in pkg/sdk/catalog_integrations_enums.go - but leave typedefs here.
2. They should also have converter functions and unit tests - check e.g. `ToStageCloud`.

pkg/sdk/catalog_integrations_impl_gen.go
```diff
@@ -0,0 +1,210 @@
...
+
+func (r *ShowCatalogIntegrationRequest) toOpts() *ShowCatalogIntegrationOptions {
+	opts := &ShowCatalogIntegrationOptions{
+		Like: r.Like,
+	}
+	return opts
+}
+
+func (r showCatalogIntegrationsDbRow) convert() (*CatalogIntegration, error) {
+	// TODO: Mapping
```
sfc-gh-jmichalak: These two convert functions should be filled manually, unfortunately

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+	PredefinedQueryStructField("restAuthType", "string", g.StaticOptions().SQL(fmt.Sprintf("TYPE = %s", sdkcommons.RestAuthenticationTypeBearer))).
+	TextAssignment("BEARER_TOKEN", g.ParameterOptions().SingleQuotes().Required())
+
+var sigV4RestAuthenticationDef = g.NewQueryStruct("SigV4RestAuthentication").
+	PredefinedQueryStructField("restAuthType", "string", g.StaticOptions().SQL(fmt.Sprintf("TYPE = %s", sdkcommons.RestAuthenticationTypeSigV4))).
+	TextAssignment("SIGV4_IAM_ROLE", g.ParameterOptions().SingleQuotes().Required()).
+	OptionalTextAssignment("SIGV4_SIGNING_REGION", g.ParameterOptions().SingleQuotes()).
+	OptionalTextAssignment("SIGV4_EXTERNAL_ID", g.ParameterOptions().SingleQuotes())
+
+var catalogIntegrationsDef = g.NewInterface(
```
sfc-gh-jmichalak, line 44: Catalog integrations should be registered to `pkg/sdk/client.go`, similarly to other objects.

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+						"SetOAuthRestAuthentication",
+						g.NewQueryStruct("SetOAuthRestAuthentication").
+							TextAssignment("OAUTH_CLIENT_SECRET", g.ParameterOptions().SingleQuotes().Required()),
+						g.ListOptions().SQL("REST_AUTHENTICATION =").Parentheses().NoComma()).
+					OptionalQueryStructField(
+						"SetBearerRestAuthentication",
+						g.NewQueryStruct("SetBearerRestAuthentication").
+							TextAssignment("BEARER_TOKEN", g.ParameterOptions().SingleQuotes().Required()),
+						g.ListOptions().SQL("REST_AUTHENTICATION =").Parentheses().NoComma()).
+					OptionalBooleanAssignment("ENABLED", g.ParameterOptions()).
```
sfc-gh-jmichalak, line 149: It's missing from the docs, if it works, it should be reported to the doc team. Also, according to docs `REST_AUTHENTICATION` is required (lacks []), this is probably incorrect as well.

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+			OptionalQueryStructField(
+				"SapBdcCatalogSourceParams",
+				g.NewQueryStruct("SapBdcParams").
+					PredefinedQueryStructField("catalogSource", "string", g.StaticOptions().SQL(fmt.Sprintf("CATALOG_SOURCE = %s", sdkcommons.CatalogSourceTypeSAPBusinessDataCloud))).
+					QueryStructField(
+						"RestConfig",
+						sapBdcRestConfigDef,
+						g.ListOptions().SQL("REST_CONFIG =").Parentheses().NoComma()),
+				g.KeywordOptions()).
+			TextAssignment("TABLE_FORMAT", g.ParameterOptions().NoQuotes().Required()).
```
sfc-gh-jmichalak: It looks like it's type-specific, e.g.

- glue, open catalog, iceberg rest -> iceberg
- object storage -> iceberg/delta
- sap -> delta

I think we just should move this field to type-specific sub structs, and restrict the values.

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+		g.NewQueryStruct("CreateCatalogIntegration").
+			Create().
+			OrReplace().
+			SQL("CATALOG INTEGRATION").
+			IfNotExists().
+			Name().
+			OptionalQueryStructField(
+				"AwsGlueCatalogSourceParams",
+				g.NewQueryStruct("AwsGlueParams").
+					PredefinedQueryStructField("catalogSource", "string", g.StaticOptions().SQL(fmt.Sprintf("CATALOG_SOURCE = %s", sdkcommons.CatalogSourceTypeAWSGlue))).
```
sfc-gh-jmichalak: We can just have `SQL()` with the enum value conversions. The same for other fields like this one.

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+					OptionalTextAssignment("GLUE_REGION", g.ParameterOptions().SingleQuotes()).
+					OptionalTextAssignment("CATALOG_NAMESPACE", g.ParameterOptions().SingleQuotes()),
+				g.KeywordOptions()).
+			OptionalQueryStructField(
+				"ObjectStorageCatalogSourceParams",
+				g.NewQueryStruct("ObjectStorageParams").
+					PredefinedQueryStructField("catalogSource", "string", g.StaticOptions().SQL(fmt.Sprintf("CATALOG_SOURCE = %s", sdkcommons.CatalogSourceTypeObjectStorage))),
+				g.KeywordOptions()).
+			OptionalQueryStructField(
+				"PolarisCatalogSourceParams",
```
sfc-gh-jmichalak: Q: In docs, this section is called "Open Catalog". We should think later what resource name would be appropriate here.

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+	"fmt"
+
+	g "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/generator/gen"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/generator/gen/sdkcommons"
+)
+
+var polarisRestConfigDef = g.NewQueryStruct("PolarisRestConfig").
+	TextAssignment("CATALOG_URI", g.ParameterOptions().SingleQuotes().Required()).
+	OptionalTextAssignment("CATALOG_API_TYPE", g.ParameterOptions().NoQuotes()).
```
sfc-gh-jmichalak: Use `OptionalSQL` instead.

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+	TextAssignment("CATALOG_NAME", g.ParameterOptions().SingleQuotes().Required()).
+	OptionalTextAssignment("ACCESS_DELEGATION_MODE", g.ParameterOptions().NoQuotes())
```
sfc-gh-jmichalak: In such cases, we usually use enums. E.g., see sdkcommons.ComputePoolInstanceFamily in compute pool defs.

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+				"ObjectStorageCatalogSourceParams",
+				g.NewQueryStruct("ObjectStorageParams").
+					PredefinedQueryStructField("catalogSource", "string", g.StaticOptions().SQL(fmt.Sprintf("CATALOG_SOURCE = %s", sdkcommons.CatalogSourceTypeObjectStorage))),
+				g.KeywordOptions()).
+			OptionalQueryStructField(
+				"PolarisCatalogSourceParams",
+				g.NewQueryStruct("PolarisParams").
+					PredefinedQueryStructField("catalogSource", "string", g.StaticOptions().SQL(fmt.Sprintf("CATALOG_SOURCE = %s", sdkcommons.CatalogSourceTypePolaris))).
+					OptionalTextAssignment("CATALOG_NAMESPACE", g.ParameterOptions().SingleQuotes()).
+					QueryStructField(
```
sfc-gh-jmichalak, line 79: Hmm, in docs we have two types: private and public. It looks like the only other difference is with `OAUTH_TOKEN_URI`  
\- Is it the Snowflake behavior, or just the docs are missing it for one type?  
\- In general, in cases like this, we provide separate structs, treating them as "different types".

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+	OptionalTextAssignment("CATALOG_API_TYPE", g.ParameterOptions().NoQuotes()).
+	TextAssignment("CATALOG_NAME", g.ParameterOptions().SingleQuotes().Required()).
+	OptionalTextAssignment("ACCESS_DELEGATION_MODE", g.ParameterOptions().NoQuotes())
+
+var icebergRestRestConfigDef = g.NewQueryStruct("IcebergRestRestConfig").
+	TextAssignment("CATALOG_URI", g.ParameterOptions().SingleQuotes().Required()).
+	OptionalTextAssignment("PREFIX", g.ParameterOptions().SingleQuotes()).
+	OptionalTextAssignment("CATALOG_NAME", g.ParameterOptions().SingleQuotes()).
+	OptionalTextAssignment("CATALOG_API_TYPE", g.ParameterOptions().NoQuotes()).
+	OptionalTextAssignment("ACCESS_DELEGATION_MODE", g.ParameterOptions().NoQuotes())
```
sfc-gh-jmichalak: These two should be enums

pkg/sdk/catalog_integrations_gen_test.go
```diff
@@ -0,0 +1,396 @@
...
+		opts := defaultOptsAws()
+		opts.IfNotExists = Bool(true)
+		opts.AwsGlueCatalogSourceParams.GlueRegion = String(glueRegion)
+		opts.AwsGlueCatalogSourceParams.CatalogNamespace = String(catalogNamespace)
+		opts.Enabled = false
+		opts.RefreshIntervalSeconds = Int(refreshIntervalSeconds)
+		opts.Comment = String(comment)
+		assertOptsValidAndSQLEquals(t, opts, "CREATE CATALOG INTEGRATION IF NOT EXISTS %s CATALOG_SOURCE = GLUE "+
+			"GLUE_AWS_ROLE_ARN = '%s' GLUE_CATALOG_ID = '%s' GLUE_REGION = '%s' CATALOG_NAMESPACE = '%s' "+
+			"TABLE_FORMAT = ICEBERG ENABLED = false REFRESH_INTERVAL_SECONDS = %d COMMENT = '%s'",
```
sfc-gh-jmichalak: I missed this one in def, but we should preserve the field order in SQL. Surprisingly, it matters sometimes.

pkg/sdk/catalog_integrations_gen_test.go
```diff
@@ -0,0 +1,396 @@
...
+		opts.RefreshIntervalSeconds = Int(refreshIntervalSeconds)
+		opts.Comment = String(comment)
+		assertOptsValidAndSQLEquals(t, opts, "CREATE CATALOG INTEGRATION IF NOT EXISTS %s CATALOG_SOURCE = POLARIS "+
+			"CATALOG_NAMESPACE = '%s' REST_CONFIG = (CATALOG_URI = '%s' CATALOG_API_TYPE = PUBLIC CATALOG_NAME = '%s' ACCESS_DELEGATION_MODE = VENDED_CREDENTIALS) "+
+			"REST_AUTHENTICATION = (TYPE = OAUTH OAUTH_TOKEN_URI = '%s' OAUTH_CLIENT_ID = '%s' OAUTH_CLIENT_SECRET = '%s' OAUTH_ALLOWED_SCOPES = ('%s')) "+
+			"TABLE_FORMAT = ICEBERG ENABLED = false REFRESH_INTERVAL_SECONDS = %d COMMENT = '%s'",
+			id.FullyQualifiedName(), catalogNamespace, polarisCatalogUri, catalogName, oAuthTokenUri, oAuthClientId, oAuthClientSecret, oAuthAllowedScope, refreshIntervalSeconds, comment)
+	})
+
+	t.Run("all options - Iceberg REST", func(t *testing.T) {
```
sfc-gh-jmichalak: Let's add more test cases for:  
\- bearer token auth for iceberg rest  
\- oauth auth for iceberg rest  
\- basic objects - for all types we should have options covering only required arguments, something like in `basic` tc

pkg/sdk/catalog_integrations_gen_test.go
```diff
@@ -0,0 +1,396 @@
...
+		opts.Set = new(CatalogIntegrationSet)
+		assertOptsInvalidJoinedErrors(t, opts, errAtLeastOneOf("AlterCatalogIntegrationOptions.Set", "SetOAuthRestAuthentication", "SetBearerRestAuthentication", "Enabled", "RefreshIntervalSeconds", "Comment"))
+	})
+
+	t.Run("basic", func(t *testing.T) {
+		opts := defaultOpts()
+		assertOptsValidAndSQLEquals(t, opts, "ALTER CATALOG INTEGRATION %s SET ENABLED = true", id.FullyQualifiedName())
+	})
+
+	t.Run("all options", func(t *testing.T) {
```
sfc-gh-jmichalak: Nit: a separate case for oauth auth

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Yeah, documentation is wrong, `REST_AUTHENTICATION` is definitely optional in ALTER.  Let me report these two issues.

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-jcieslak: Also, the order of the fields sometimes matters in SQL, there is possibility that TABLE_FORMAT may not even be accepted by Snowflake at this position.

pkg/sdk/catalog_integrations_gen_test.go
```diff
@@ -0,0 +1,396 @@
...
+	oAuthClientId          = "my_client_id"
+	oAuthClientSecret      = "my_client_secret"
+	oAuthAllowedScope      = "PRINCIPAL_ROLE:ALL"
+	oAuthTokenUri          = "https://api.tabular.io/ws/v1/oauth/tokens"
+	sigV4IamRole           = "arn:aws:iam::123456789012:role/my-role"
+	sigV4SignignRegion     = "us-west-2"
+	sigV4ExternalId        = "external_id"
+	refreshIntervalSeconds = 60
+	comment                = "test comment"
+)
```
sfc-gh-jcieslak, line 24: imo, if possible, we should avoid global context, but if you would like to use it I propose having a common prefix that would clearly suggest those are **test values** and they are used in **catalog integrations**

pkg/sdk/object_types.go
sfc-gh-jcieslak: Also, a fun fact, as the above comment suggests, all integrations are of object type `INTEGRATION` in Snowflake, so this one is also a pseudo object type we're just adding for convenience. Maybe we should adjust this comment to include all integration types under it as for now it only points at ObjectTypeApiIntegration

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Why? It's an assignment (e.g. `CATALOG_API_TYPE = PUBLIC`) and part of the CREATE operation.

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: We could do this:
* replace `PolarisCatalogSourceParams` with `OpenCatalogCatalogSourceParams`
* replace `PolarisParams` with `OpenCatalogParams`
* replace `PolarisRestConfig` with `OpenCatalogRestConfig`

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: I successfully created a private integration with `OAUTH_TOKEN_URI` property on my preprod account. `DESCRIBE` also included this property in the response. So it's a valid SQL. I'm not sure why the property isn't in the docs, but maybe it's connected to this notice:

> If you’re using External OAuth with private connectivity (CATALOG_API_TYPE=PRIVATE), Snowflake routes the token requests for External OAuth over the public internet.

Basically, Snowflake is saying "this is allowed, but since the request goes through the public internet, it might not be what you want".

pkg/sdk/generator/gen/sdkcommons/enums.go
sfc-gh-kwasilewski: Converter functions won't be needed until I start implementing the resource, right? If so, I can add them later

sfc-gh-asawicki: Yes, but then the resource PR is unnecessarily big. We try to make SDK-related changes in this one, as it contains Snowflake object's behavior, and not the TF-related logic.

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-asawicki: Technically, it's not an assignment from our perspective, since you have only one mandatory value. So, for such cases, we use the OptionalSQL.

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-asawicki: +1

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-asawicki: Let's confirm with the proper team what is the reason for this being missing from the docs (we can keep it with a follow-up TODO that the confirmation is needed and we may remove it).

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-asawicki: +1

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-asawicki: +1

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+			OptionalQueryStructField(
+				"Set",
+				g.NewQueryStruct("CatalogIntegrationSet").
+					OptionalQueryStructField(
+						"SetOAuthRestAuthentication",
+						g.NewQueryStruct("SetOAuthRestAuthentication").
+							TextAssignment("OAUTH_CLIENT_SECRET", g.ParameterOptions().SingleQuotes().Required()),
+						g.ListOptions().SQL("REST_AUTHENTICATION =").Parentheses().NoComma()).
+					OptionalQueryStructField(
+						"SetBearerRestAuthentication",
```
sfc-gh-asawicki, line 145: question: is this option available to all ctalog integrations? I dont think all of them have the bearer token (its only iceberg that has it?)

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+						g.ListOptions().SQL("REST_AUTHENTICATION =").Parentheses().NoComma()).
+					OptionalBooleanAssignment("ENABLED", g.ParameterOptions()).
+					OptionalNumberAssignment("REFRESH_INTERVAL_SECONDS", g.ParameterOptions().NoQuotes()).
+					// TODO(SNOW-3121221): use COMMENT in unset and here use OptionalComment
+					OptionalAssignment("COMMENT", "StringAllowEmpty", g.ParameterOptions()).
+					WithValidation(g.ConflictingFields, "SetOAuthRestAuthentication", "SetBearerRestAuthentication").
+					WithValidation(g.AtLeastOneValueSet, "SetOAuthRestAuthentication", "SetBearerRestAuthentication", "Enabled", "RefreshIntervalSeconds", "Comment"),
+				g.KeywordOptions().SQL("SET"),
+			).
+			OptionalSetTags().
```
sfc-gh-asawicki, line 157: q: Are tags undocumented?

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,209 @@
...
+			Time("CreatedOn"),
+		g.NewQueryStruct("ShowCatalogIntegration").
+			Show().
+			SQL("CATALOG INTEGRATIONS").
+			OptionalLike(),
+	).
+	ShowByIdOperationWithFiltering(
+		g.ShowByIDLikeFiltering,
+	).
+	DescribeOperation(
```
sfc-gh-asawicki, line 195: note for later (not for this PR): this format of describe output will need to be additionally mapped to the dedicated struct (check e.g. recent stages/storage integrations)

pkg/sdk/catalog_integrations_gen_test.go
```diff
@@ -0,0 +1,396 @@
...
+	// Minimal valid CreateCatalogIntegrationOptions for SAP Business Data Cloud
+	defaultOptsSapBdc := func() *CreateCatalogIntegrationOptions {
+		return &CreateCatalogIntegrationOptions{
+			name: id,
+			SapBdcCatalogSourceParams: &SapBdcParams{
+				RestConfig: SapBdcRestConfig{
+					SapBdcInvitationLink: sapBdcInvitationLink,
+				},
+			},
+			TableFormat: "DELTA",
```
sfc-gh-asawicki: table format is tighlty coupled with each type, so it shouldnt be settable (it's already in some other comment on def)

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Yes

sfc-gh-kwasilewski: At least that's the case in the ALTER CATALOG INTERNATION page. It's documented elsewhere: https://docs.snowflake.com/en/user-guide/object-tagging/introduction#supported-objects

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Yes, only Iceberg REST supports bearer tokens. Should we do anything with that? We could define multiple alter operations (one for each catalog integration type), but at the end of the day, generated SQL will be the same.

sfc-gh-asawicki: I would consider extracting icebergparams, similarly to create, to distinguish which option is supported for which type, but I'd say in this case is nice-to-have, not a must-have.

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: > Since you have only one mandatory value

Not true, the user chooses one value from multiple allowed values:
* For Open Catalog: PUBLIC, PRIVATE
* For Iceberg REST: PUBLIC, PRIVATE, AWS_API_GATEWAY, AWS_PRIVATE_API_GATEWAY, AWS_GLUE, AWS_PRIVATE_GLUE

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-kwasilewski: True, except for Open Catalog type, where two table formats are allowed

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-kwasilewski: It seems we don't have any established conventions, see [api_integrations_gen_test.go](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/sdk/api_integrations_gen_test.go) or [notification_integrations_gen_test.go](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/sdk/notification_integrations_gen_test.go).

I removed some constants that were more generic, but I'd like to keep the rest (especially these with long URLs)

pkg/sdk/catalog_integrations_impl_gen.go
sfc-gh-kwasilewski: Done

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Done

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Done

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Done, I used `SQLWithCustomFieldName`

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Done

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Done

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Done

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-kwasilewski: Done

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-kwasilewski: Done

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-kwasilewski: Done

pkg/sdk/object_types.go
sfc-gh-kwasilewski: Done

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-asawicki: Ah true, I missed that there are two separate sections for OpenCatalog.

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Hmm, I think it can't be similar to create, because you provide a type (the `CATALOG_SOURCE` property) in the CREATE command, but not in the ALTER command. In fact, you don't know the type of integration you're altering.

pkg/sdk/catalog_integrations_dto_gen.go
```diff
@@ -0,0 +1,128 @@
...
+	OauthAllowedScopes []StringListItemWrapper // required
+}
+
+type IcebergRestParamsRequest struct {
+	CatalogNamespace         *string
+	RestConfig               IcebergRestRestConfigRequest
+	OAuthRestAuthentication  *OAuthRestAuthenticationRequest
+	BearerRestAuthentication *BearerRestAuthenticationRequest
+	SigV4RestAuthentication  *SigV4RestAuthenticationRequest
+}
```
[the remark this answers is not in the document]
sfc-gh-kwasilewski: The "RestConfig" field is a value type because it's not marked as optional in the definition. That's how the SDK generator works.

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: Added a TODO, will ask the team about that.

pkg/sdk/generator/gen/sdkcommons/enums.go
sfc-gh-kwasilewski: Added converter functions + unit tests

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-jmichalak: Yeah, in such cases, we define enums and use them with e.g. Assignment. LGTM.

pkg/sdk/catalog_integrations_enums.go
```diff
@@ -0,0 +1,123 @@
...
+	if !slices.Contains(AllCatalogIntegrationRestAuthenticationTypes, CatalogIntegrationRestAuthenticationType(s)) {
+		return "", fmt.Errorf("invalid rest authentication type: %s", s)
+	}
+	return CatalogIntegrationRestAuthenticationType(s), nil
+}
+
+type CatalogIntegrationAccessDelegationMode string
+
+const (
+	CatalogIntegrationAccessDelegationModeVendedCredentials         CatalogIntegrationAccessDelegationMode = "VENDED_CREDENTIALS"
```
sfc-gh-jmichalak: Note: this is marked as insecure in pre push (we don't have pipelines for it). Just add `//nolint:gosec` at the end.

pkg/sdk/catalog_integrations_enums.go
```diff
@@ -0,0 +1,123 @@
...
+
+type CatalogIntegrationCatalogApiType string
+
+const (
+	CatalogIntegrationCatalogApiTypePublic               CatalogIntegrationCatalogApiType = "PUBLIC"
+	CatalogIntegrationCatalogApiTypePrivate              CatalogIntegrationCatalogApiType = "PRIVATE"
+	CatalogIntegrationCatalogApiTypeAwsApiGateway        CatalogIntegrationCatalogApiType = "AWS_API_GATEWAY"
+	CatalogIntegrationCatalogApiTypeAwsPrivateApiGateway CatalogIntegrationCatalogApiType = "AWS_PRIVATE_API_GATEWAY"
+	CatalogIntegrationCatalogApiTypeAwsGlue              CatalogIntegrationCatalogApiType = "AWS_GLUE"
+	CatalogIntegrationCatalogApiTypeAwsPrivateGlue       CatalogIntegrationCatalogApiType = "AWS_PRIVATE_GLAUE"
```
sfc-gh-jmichalak:
```suggestion
	CatalogIntegrationCatalogApiTypeAwsPrivateGlue       CatalogIntegrationCatalogApiType = "AWS_PRIVATE_GLUE"
```

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -0,0 +1,212 @@
...
+var icebergRestRestConfigDef = g.NewQueryStruct("IcebergRestRestConfig").
+	TextAssignment("CATALOG_URI", g.ParameterOptions().SingleQuotes().Required()).
+	OptionalTextAssignment("PREFIX", g.ParameterOptions().SingleQuotes()).
+	OptionalTextAssignment("CATALOG_NAME", g.ParameterOptions().SingleQuotes()).
+	OptionalAssignment("CATALOG_API_TYPE", g.KindOfT[sdkcommons.CatalogIntegrationCatalogApiType](), g.ParameterOptions().NoQuotes()).
+	OptionalAssignment("ACCESS_DELEGATION_MODE", g.KindOfT[sdkcommons.CatalogIntegrationAccessDelegationMode](), g.ParameterOptions().NoQuotes())
+
+var sapBdcRestConfigDef = g.NewQueryStruct("SapBdcRestConfig").
+	TextAssignment("SAP_BDC_INVITATION_LINK", g.ParameterOptions().SingleQuotes().Required()).
+	OptionalTextAssignment("ACCESS_DELEGATION_MODE", g.ParameterOptions().NoQuotes())
```
sfc-gh-jmichalak: I believe this should also be an enum, can be adjusted in a follow-up.

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-asawicki: You know as a caller (especially from the tf context when the type is a given). This way, we are making it easier on the callers side to know which params are available - I would try to set aws params for azure integration, etc.
But as I said, it's a single attribute, we can leave it like this.

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-asawicki: True, there are no conventions, we should discuss how we'd like to have test data like this one.

Just a note: the two examples are more than 2 years old, so they are not the freshest.

sfc-gh-asawicki (dismissed):

sfc-gh-jcieslak (dismissed):

pkg/sdk/catalog_integrations_enums.go
sfc-gh-kwasilewski: Ehh. Good catch, thanks!

pkg/sdk/generator/defs/catalog_integrations_def.go
sfc-gh-kwasilewski: That's right. I've already adjusted it.

pkg/sdk/catalog_integrations_enums.go
sfc-gh-kwasilewski: Done

sfc-gh-jmichalak (approved):

PR #4484, opened 2026-02-23
chore: adjust tests for BCR 2026_01

sfc-gh-jcieslak:
Merge after BCR 2026_01 is enabled by default on production environments

## Changes
- Adjust tests for changes that come in BCR 2026_01, no changes in production code were needed
- Remove unused assertions
- "inline" `testClient().SnowflakeDefaults.StageIdentifierOutputFormatForStreamOnDirectoryTable` as it contained no Snowflake environment-dependent logic after BCR change

### DIFF
--- a/pkg/testacc/resource_saml2_integration_acceptance_test.go
+++ b/pkg/testacc/resource_saml2_integration_acceptance_test.go
@@ -52,15 +52,15 @@ func TestAcc_Saml2Integration_basic(t *testing.T) {
 		WithSaml2SnowflakeAcsUrl(acsURL).
 		WithSaml2SnowflakeIssuerUrl(issuerURL).
 		WithSaml2SpInitiatedLoginPageLabel("foo")
-		// TODO: allowed_user_domains and allowed_email_patterns temporarily disabled due to Snowflake-side changes
-		// WithAllowedEmailPatterns("^(.+dev)@example.com$").
-		// WithAllowedUserDomains("example.com")
+	// TODO(SNOW-3200188): allowed_user_domains and allowed_email_patterns temporarily disabled due to Snowflake-side changes
+	// WithAllowedEmailPatterns("^(.+dev)@example.com$").
+	// WithAllowedUserDomains("example.com")
 	recreatesModel := model.Saml2SecurityIntegrationVar("test", id.Name(), issuer, string(sdk.Saml2SecurityIntegrationSaml2ProviderCustom), validUrl, temporaryVariableName).
 		WithSaml2SnowflakeAcsUrl(acsURL).
 		WithSaml2SnowflakeIssuerUrl(issuerURL).
 		WithSaml2SpInitiatedLoginPageLabel("foo")
-		// WithAllowedEmailPatterns("^(.+dev)@example.com$").
-		// WithAllowedUserDomains("example.com")
+	// WithAllowedEmailPatterns("^(.+dev)@example.com$").
+	// WithAllowedUserDomains("example.com")
 
 	resource.Test(t, resource.TestCase{
 		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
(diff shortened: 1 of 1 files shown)


pkg/sdk/testint/streams_gen_integration_test.go
```diff
@@ -152,8 +152,8 @@ func TestInt_Streams(t *testing.T) {
 			HasComment("some comment").
 			HasSourceType(sdk.StreamSourceTypeStage).
 			HasMode(sdk.StreamModeDefault).
-			HasStageName(stage.ID().Name()).
-			HasBaseTablesPartiallyQualified(stage.ID().Name()),
+			HasTableId(stage.ID()).
+			HasBaseTables(fmt.Sprintf(`"%s"."%s".%s`, stage.ID().DatabaseName(), stage.ID().SchemaName(), stage.ID().Name())),
```
sfc-gh-kwasilewski, line 156: I understand we can't use the `FullyQualifiedName` method because the stage name is unquoted, is that so?

pkg/acceptance/bettertestspoc/assert/objectassert/stream_snowflake_ext.go
```diff
@@ -1,9 +1,7 @@
 package objectassert
```
sfc-gh-jmichalak, line 1: We should also add an entry to the BCR migration guide and an entry regarding the fix in streams on directory tables to the main migration guide.

pkg/sdk/testint/streams_gen_integration_test.go
sfc-gh-jmichalak, line 156: I think we can just use FullyQualifiedName, right? In other places as well.

pkg/testacc/data_source_streams_acceptance_test.go
```diff
@@ -267,9 +267,9 @@ func TestAcc_StreamOnDirectoryTable(t *testing.T) {
 						HasDatabaseName(id.DatabaseName()).
 						HasSchemaName(id.SchemaName()).
 						HasOwner(snowflakeroles.Accountadmin.Name()).
-						HasTableName(stage.ID().Name()).
+						HasTableName(fmt.Sprintf(`"%s"."%s".%s`, stage.ID().DatabaseName(), stage.ID().SchemaName(), stage.ID().Name())).
```
sfc-gh-kwasilewski, line 270: This new format is used in multiple places, should we have a helper function?

pkg/sdk/testint/streams_gen_integration_test.go
```diff
@@ -152,8 +152,8 @@ func TestInt_Streams(t *testing.T) {
 			HasComment("some comment").
 			HasSourceType(sdk.StreamSourceTypeStage).
```
sfc-gh-jmichalak, line 153: Please also adjust the resource (remove diff suppress and adjust the description) - SNOW-1733130.

pkg/sdk/testint/streams_gen_integration_test.go
sfc-gh-asawicki: the last part doesnt have quotes

sfc-gh-asawicki (approved):

pkg/sdk/testint/streams_gen_integration_test.go
sfc-gh-jcieslak: +1, our ID generator (at least for the last part) generates all upper-case mix of letters and numbers, which makes it return an identifier with mixed quotes from the Snowflake side. We could address this by having at least one lower case letter in our schema object identifiers, but that's another change. For now, I think this could be good enough, wdyt?

sfc-gh-jcieslak: Yes, the answer is here: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4484#discussion_r2854606076

pkg/testacc/data_source_streams_acceptance_test.go
sfc-gh-jcieslak: Well, we could have something like `schemaObjId.FullyQualifiedNameWithLastPartUnquoted` or something similar or more global helper in `sdk.SomeFunc(schemaObjId)`. We have similar formatting in other places, so maybe we could extract this as a task that applies globally?
cc: @sfc-gh-jmichalak @sfc-gh-asawicki

pkg/acceptance/bettertestspoc/assert/objectassert/stream_snowflake_ext.go
sfc-gh-jcieslak: will address in next pr (the change ended up being too big)

pkg/sdk/testint/streams_gen_integration_test.go
sfc-gh-jcieslak: will address in next pr (the change ended up being too big)

pkg/sdk/testint/streams_gen_integration_test.go
sfc-gh-jcieslak: I think in the follow up this going to change to FQN

pkg/testacc/data_source_streams_acceptance_test.go
sfc-gh-kwasilewski: Since this will be changed to FQN in the follow-up PR (https://github.com/snowflakedb/terraform-provider-snowflake/pull/4484#discussion_r2870942916), I'm ok with not introducing a helper function.

PR #4485, opened 2026-02-24
feat: Support DECFLOAT datatype

sfc-gh-asawicki:
Add DECFLOAT to our SDK datatype abstraction:
- new type
- support for all datatype methods
- unit tests
- integration test

Next PR:
- verify usage of this type for objects like functions, procedures, tables (TODOs left in this PR)

### DIFF
--- a/pkg/sdk/testint/data_types_integration_test.go
+++ b/pkg/sdk/testint/data_types_integration_test.go
@@ -12,7 +12,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TODO [this PR]: add decfloat to below tests
 func TestInt_DataTypes(t *testing.T) {
 	client := testClient(t)
 	ctx := testContext(t)
@@ -39,6 +38,12 @@ func TestInt_DataTypes(t *testing.T) {
 		"NUMBER(x)",
 		"INT()",
 		"NUMBER(36, 5, 7)",
+	}
+	incorrectDecfloatDatatypes := []string{
+		"DECFLOAT()",
+		"DECFLOAT(x)",
+		"DECFLOAT(38, 2)",
+		"DECFLOAT(35)",
 	}
 	incorrectTextDatatypes := []string{
 		"VARCHAR()",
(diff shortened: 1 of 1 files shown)


pkg/sdk/datatypes/decfloat.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+const DefaultDecfloatPrecision = 38
+
+// DecfloatDataType is based on https://docs.snowflake.com/en/sql-reference/data-types-numeric#decfloat
+// It doesn't have synonyms. It does have optional precision attribute.
+// Currently, the only allowed value is 38. To be consistent with other data types we are not validating this value (e.g., it's possible to set number's precision to 40 and it will fail on Snowflake).
+// Precision can be known or unknown.
+type DecfloatDataType struct {
+	precision      int
```
sfc-gh-kwasilewski, line 16: WDYT? Does it make sense?

```suggestion
	precision      uint
```

pkg/sdk/testint/procedures_integration_test.go
```diff
@@ -2246,6 +2246,7 @@ def filter_by_role(session, table_name, role):
 		assert.ErrorIs(t, err, sdk.ErrObjectNotFound)
 	})
 
+	// TODO [SNOW-3151661]: add decfloat to below tests
```
sfc-gh-kwasilewski, line 2249: DECFLOAT is not yet supported in sprocs, see: https://docs.snowflake.com/en/sql-reference/data-types-numeric#limitations-for-the-decfloat-data-type. The comment is unnecessary

pkg/sdk/datatypes/decfloat.go
sfc-gh-asawicki: All other datatypes use `int` and this is output of parsing, we've seen strange things, so it's safer to stick with signed.

sfc-gh-kwasilewski: Fair enough, thanks

pkg/sdk/testint/procedures_integration_test.go
sfc-gh-asawicki: This limitation (`The DECFLOAT data type isn’t supported in stored procedures or user-defined functions (UDFs) written in a language other than SQL`) can be understood in two ways:
1. It's not supported at all in sprocs and only supported in SQL UDFs.
2. It's supported for both sprocs and UDFs, but only for the SQL type.

I think it's the latter, but we'll see in the follow-up PR with tests.

sfc-gh-kwasilewski: I know for sure that back in the previous quarter, it was the first case, but maybe they have added support for SQL sprocs since then. Let's see in the follow-up PR.

sfc-gh-asawicki: It was indeed 1. (https://github.com/snowflakedb/terraform-provider-snowflake/pull/4489).

PR #4486, opened 2026-02-24
chore: Test nullable lists in Terraform

sfc-gh-jcieslak:
## Changes
- Add two resources testing the Terraform SDKv2 ability to differentiate between filled, empty, and null lists:
  - Just regular resource with list
  - Resource with additional computed field which aids in empty <-> null transitions
- Add a set of acceptance tests that check for state transitions in both resources, and checks how the external changes are handled

### DIFF
--- a/pkg/testfunctional/1_sdk_v2_functional_tests_provider_test.go
+++ b/pkg/testfunctional/1_sdk_v2_functional_tests_provider_test.go
@@ -30,11 +30,11 @@ func sdkV2FunctionalTestsProvider() *schema.Provider {
 
 func testResources() map[string]*schema.Resource {
 	return map[string]*schema.Resource{
-		"snowflake_test_resource_object_renaming":              testfunctional.TestResourceObjectRenamingListsAndSets(),
-		"snowflake_test_resource_data_type_diff_handling":      testfunctional.TestResourceDataTypeDiffHandling(),
-		"snowflake_test_resource_data_type_diff_handling_list": testfunctional.TestResourceDataTypeDiffHandlingList(),
-		"snowflake_test_resource_list_null_value_logic":                testfunctional.TestResourceListNullValueLogic(),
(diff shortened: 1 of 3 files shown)


pkg/testfunctional/test_resource_list_null_value_logic_acceptance_test.go
```diff
@@ -0,0 +1,480 @@
...
+			Steps: []resource.TestStep{
+				{
+					PreConfig: func() { t.Setenv(envName, "") },
+					Config:    configWithEmpty,
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckResourceAttr(ref, "nullable_list.#", "0"),
+					),
+				},
+				{
+					PreConfig: func() { t.Setenv(envName, "__NULL__") },
```
sfc-gh-asawicki, line 436: nit: can we use extracted values for "__NULL__" etc?

pkg/acceptance/testenvs/testing_environment_variables.go
```diff
@@ -45,6 +45,7 @@ const (
 
 	SimplifiedIntegrationTestsSetup env = "TEST_SF_TF_SIMPLIFIED_INTEGRATION_TESTS_SETUP"
 
+	TestResourceNullListHandlingEnv     env = "TEST_SF_TF_TEST_RESOURCE_DATA_NULL_LIST_HANDLING_ENV"
```
sfc-gh-asawicki, line 48: nit: why env and not existing http server helper?

sfc-gh-jcieslak: tbh, forgot that we could use it here, I was looking at other cases where env-way was still used

pkg/testfunctional/test_resource_list_null_value_logic_acceptance_test.go
sfc-gh-jcieslak: You mean constants I defined in the resource file?

sfc-gh-jcieslak:
Merging, but can address nits follow-up (if needed)

Issue #4238, opened 2025-11-24
[Bug]: Unsetting allowed_values property on TAG resource


leonardo-henriques-deel:
### Terraform CLI Version

2.4.0

### Terraform Provider Version

2.4.0

### Terraform Configuration

```terraform
resource "snowflake_tag" "test_tag" {
  name           = "TEST_TAG"
  database       = var.database_name
  schema         = var.schema_name
  allowed_values = ["VAL1", "VAL2"]
}

# After changes

resource "snowflake_tag" "test_tag" {
  name     = "TEST_TAG"
  database = var.database_name
  schema   = var.schema_name

  # `allowed_values` removed; setting it to null also has no effect
}
```

### Category

category:resource

### Expected Behavior

When the `allowed_values` attribute is removed from the Terraform resource (or set to `null`), the provider should restore the tag to an unrestricted state.

Instead of removing individual values using:

```sql
ALTER TAG <tag_name> DROP ALLOWED_VALUES ...;
```

the provider should issue:

```sql
ALTER TAG <tag_name> UNSET ALLOWED_VALUES;
```

This correctly clears the `allowed_values` property according to Snowflake’s documentation and allows any value to be assigned to the tag.

### Actual Behavior

When the `allowed_values` attribute is removed from the Terraform resource, the provider generates a statement of the form:

```sql
ALTER TAG <tag_name> DROP ALLOWED_VALUES 'VAL1', 'VAL2';
```

This removes the specified values but leaves the tag with an empty `allowed_values` list (`[]`). Snowflake interprets this as *no allowed values*, which prevents any tag value from being set.

As a result, the tag becomes unusable instead of reverting to an unrestricted state.

### Steps to Reproduce

1. Create a new tag with allowed values defined, for example:

   ```hcl
   resource "snowflake_tag" "test_tag" {
     name           = "TEST_TAG"
     database       = "DB"
     schema         = "SCHEMA"
     allowed_values = ["VAL1", "VAL2"]
   }
   ```

2. Apply the configuration:

   ```bash
   terraform apply
   ```

3. Remove the `allowed_values` attribute from the resource:

   ```hcl
   resource "snowflake_tag" "test_tag" {
     name     = "TEST_TAG"
     database = "DB"
     schema   = "SCHEMA"
     # allowed_values removed
   }
   ```

4. Apply again:

   ```bash
   terraform apply
   ```

5. Observe the SQL generated by the provider (via debug logs or Snowflake query history).
[...]

PR #4498, opened 2026-02-25
chore: Introduce a new behavior for tags

sfc-gh-jcieslak:
## Changes
- Introduce a new experimental flag for the new allowed values behavior
- Add new `no_allowed_values` field in tag resource
- Adjust resource logic to support new behavior whenever the flag is set, otherwise keep the old logic
  - where the logic may diverge is the change in create, where alters after resource creation are not causing failures, but warnings instead. This aligns with other resources that need to be altered after an object is created.
  - note: I thought about matching old behavior with new one a bit more to keep the code lean, but in the end decided to keep clean separation with `if` on the flag
- Add tests proving the new behavior works, old behavior remains, and enabling the flag is seamless
- Add/Update documentation for the tag resource

References: #4238

### DIFF
--- a/pkg/sdk/testint/notification_integrations_gen_integration_test.go
+++ b/pkg/sdk/testint/notification_integrations_gen_integration_test.go
@@ -339,7 +339,7 @@ func TestInt_NotificationIntegrations(t *testing.T) {
 		details, err := client.NotificationIntegrations.Describe(ctx, integration.ID())
 		require.NoError(t, err)
 
-		assert.Contains(t, details, sdk.NotificationIntegrationProperty{Name: "ENABLED", Type: "Boolean", Value: "true", Default: "true"})
+		assert.Contains(t, details, sdk.NotificationIntegrationProperty{Name: "ENABLED", Type: "Boolean", Value: "false", Default: "true"})
 		assert.Contains(t, details, sdk.NotificationIntegrationProperty{Name: "ALLOWED_RECIPIENTS", Type: "List", Value: "jan.cieslak@snowflake.com", Default: "[]"})
 		assert.Contains(t, details, sdk.NotificationIntegrationProperty{Name: "COMMENT", Type: "String", Value: "changed comment", Default: ""})
 
@@ -355,7 +355,7 @@ func TestInt_NotificationIntegrations(t *testing.T) {
 		details, err = client.NotificationIntegrations.Describe(ctx, integration.ID())
 		require.NoError(t, err)
 
-		assert.Contains(t, details, sdk.NotificationIntegrationProperty{Name: "ENABLED", Type: "Boolean", Value: "true", Default: "true"})
+		assert.Contains(t, details, sdk.NotificationIntegrationProperty{Name: "ENABLED", Type: "Boolean", Value: "false", Default: "true"})
 		assert.Contains(t, details, sdk.NotificationIntegrationProperty{Name: "ALLOWED_RECIPIENTS", Type: "List", Value: "", Default: "[]"})
 		assert.Contains(t, details, sdk.NotificationIntegrationProperty{Name: "COMMENT", Type: "String", Value: "", Default: ""})
 	})


pkg/provider/experimentalfeatures/experimental_features.go
```diff
@@ -18,7 +18,8 @@ const (
 	UserEnableDefaultWorkloadIdentity              ExperimentalFeature = "USER_ENABLE_DEFAULT_WORKLOAD_IDENTITY"
 	GrantsImportValidation                         ExperimentalFeature = "GRANTS_IMPORT_VALIDATION"
 	// TODO [SNOW-2739299]: Discuss having an additional ParametersNoOutput experiment
-	ParametersReducedOutput ExperimentalFeature = "PARAMETERS_REDUCED_OUTPUT"
+	ParametersReducedOutput             ExperimentalFeature = "PARAMETERS_REDUCED_OUTPUT"
+	TagNewTriValueAllowedValuesBehavior ExperimentalFeature = "TAG_NEW_TRI_VALUE_ALLOWED_VALUES_BEHAVIOR"
```
sfc-gh-jmichalak: Nit: I think the TriValue is a bit misleading. How about `TAG_NEW_EMPTY_ALLOWED_VALUES_BEHAVIOR`.

pkg/resources/tag.go
```diff
@@ -153,21 +166,59 @@ func CreateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
...
 	}
 	d.SetId(helpers.EncodeResourceIdentifier(id))
+
+	var updateAfterCreationDiags diag.Diagnostics
+
 	if v, ok := d.GetOk("masking_policies"); ok {
 		ids, err := parseSchemaObjectIdentifierSet(v)
 		if err != nil {
-			return diag.FromErr(err)
+			updateAfterCreationDiags = append(updateAfterCreationDiags, diag.Diagnostic{
```
sfc-gh-jmichalak, line 179: Why is this behavior changed?

pkg/resources/tag.go
```diff
@@ -153,21 +166,59 @@ func CreateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
...
+				Summary:  "Failed to set masking policies on the tag",
+				Detail:   fmt.Sprintf("Unable to alter tag %s, err = %s", id.FullyQualifiedName(), err),
+			})
 		}
 	}
-	return ReadContextTag(ctx, d, meta)
+
+	if experimentalfeatures.IsExperimentEnabled(experimentalfeatures.TagNewTriValueAllowedValuesBehavior, providerCtx.EnabledExperiments) {
+		if v, ok := d.GetOk("no_allowed_values"); ok && v.(bool) {
+			// We have to temporarily add and remove allowed value for Snowflake to make the tag block any value.
```
sfc-gh-jmichalak, line 198: Q: Is there a follow-up to this? I think the object should support it eventually.

pkg/resources/tag.go
```diff
@@ -153,21 +166,59 @@ func CreateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
...
+			if err := client.Tags.Alter(ctx, sdk.NewAlterTagRequest(id).WithAdd([]string{tempTagAllowedValue})); err != nil {
+				updateAfterCreationDiags = append(updateAfterCreationDiags, diag.Diagnostic{
+					Severity: diag.Warning,
+					Summary:  "Failed to set masking policies on the tag",
```
sfc-gh-jmichalak: These two summaries are incorrect.

pkg/resources/tag.go
```diff
@@ -243,25 +305,74 @@ func UpdateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
...
+							return diag.FromErr(err)
+						}
+					} else {
+						// Some values are still in the config, use DROP to avoid disrupting other users who depend on the remaining allowed values
+						if err := client.Tags.Alter(ctx, sdk.NewAlterTagRequest(id).WithDrop(removedItems)); err != nil {
+							return diag.FromErr(err)
+						}
+					}
+				}
 			}
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: \+1

pkg/resources/tag.go
```diff
@@ -153,21 +166,59 @@ func CreateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
...
+			updateAfterCreationDiags = append(updateAfterCreationDiags, diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary:  "Failed to set masking policies on the tag",
+				Detail:   fmt.Sprintf("Unable to alter tag %s, err = %s", id.FullyQualifiedName(), err),
+			})
 		}
 	}
-	return ReadContextTag(ctx, d, meta)
+
+	if experimentalfeatures.IsExperimentEnabled(experimentalfeatures.TagNewTriValueAllowedValuesBehavior, providerCtx.EnabledExperiments) {
```
sfc-gh-jmichalak: We can also fail when no_allowed_values is set and the experiment is not enabled.

pkg/resources/tag.go
```diff
@@ -243,25 +305,74 @@ func UpdateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
...
+			addedItems, removedItems := ListDiff(oldAllowedValues, newAllowedValues)
+			noAllowedValues := d.Get("no_allowed_values").(bool)
 
-		if len(addedItems) > 0 {
-			if err := client.Tags.Alter(ctx, sdk.NewAlterTagRequest(id).WithAdd(addedItems)); err != nil {
-				return diag.FromErr(err)
+			if d.HasChange("no_allowed_values") && noAllowedValues {
+				if len(oldAllowedValues) == 0 {
+					// If no values where previously set, add a new temporary one and drop it to block any value on the Snowflake side
+					if err := errors.Join(
```
sfc-gh-jmichalak, line 325: This logic is similar to the create and can be extracted.

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -308,3 +310,476 @@ resource "snowflake_tag" "test" {
...
+					objectassert.Tag(t, id).
+						HasName(id.Name()).
+						HasAllowedValuesNil(), // After initial creation, allowed values are null in Snowflake
+
+					resourceassert.TagResource(t, withNoAllowedValues.ResourceReference()).
+						HasNameString(id.Name()).
+						HasAllowedValuesEmpty(),
+				),
+			},
+			// Update - set no_allowed_values to true, but experiment is off so it should be ignored
```
sfc-gh-jmichalak: Nit: this description is the same as below? This step uses a config without no_allowed_values.

docs/resources/tag.md
```diff
@@ -9,6 +9,8 @@ description: |-
...
 
+~> **Note** A new `TAG_NEW_TRI_VALUE_ALLOWED_VALUES_BEHAVIOR` experimental feature is available for this resource. When enabled, it improves how `allowed_values` are handled — removing `allowed_values` from the configuration correctly reverts the tag to accepting any value, and a new `no_allowed_values` field allows you to explicitly block any value from being set on the tag. Without the flag, the behavior is unchanged from previous versions and the `no_allowed_values` field has no effect. To enable it, add `TAG_NEW_TRI_VALUE_ALLOWED_VALUES_BEHAVIOR` to the [`experimental_features_enabled`](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs#experimental_features_enabled-1) provider field. See the [migration guide](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#new-feature-improved-allowed_values-handling-in-snowflake_tag) for more details.
```
sfc-gh-asawicki: `TAG_NEW_TRI_VALUE_ALLOWED_VALUES_BEHAVIOR` -> `TAGS_ALLOW_EMPTY_ALLOWED_VALUES` ?

docs/index.md
```diff
@@ -949,3 +949,10 @@ Enables import validation for the `snowflake_grant_privileges_to_account_role` r
 When enabled, importing a grant resource with a fixed set of privileges (`privileges` field) will validate that the specified privileges actually exist in Snowflake with the correct `with_grant_option` setting, and error immediately if they don't match.
 
 This feature works independently of the `GRANTS_STRICT_PRIVILEGE_MANAGEMENT` flag.
+
+#### TAG_NEW_TRI_VALUE_ALLOWED_VALUES_BEHAVIOR
+Enables behavior changes for the `allowed_values` field in the `snowflake_tag` resource.
+
+When enabled, the three possible states in Snowflake for allowed values will be supported: `nil` (any value is allowed; whenever `allowed_values` are empty), `empty` (no value is allowed; handled by the `no_allowed_values` field), and `set` (all values defined in `allowed_values` are allowed).
```
sfc-gh-asawicki, line 956: I think this description is too technical. I think that for customers it should be clear that:
- you can restrict it to certain values
- you can set the tag to allow every value
- you can set it to no accepted values

And we could add how to achieve each of these options, and that's it (do not mention nils etc)

pkg/provider/experimentalfeatures/experimental_features.go
sfc-gh-asawicki: My proposal for the name: `TAGS_ALLOW_EMPTY_ALLOWED_VALUES`

pkg/resources/tag.go
sfc-gh-asawicki: +1 do we have confirmation from the team, that it cannot be achieved currently as oneshot operation? (e.g. passing  empty paretheses `()`)

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -308,3 +310,476 @@ resource "snowflake_tag" "test" {
...
+					objectassert.Tag(t, id).
+						HasName(id.Name()).
+						HasAllowedValuesNil(),
+
+					resourceassert.TagResource(t, basic.ResourceReference()).
+						HasNameString(id.Name()).
+						HasAllowedValuesEmpty(),
+				),
+			},
+			// Update - set no_allowed_values to true, but experiment is off so it should be ignored
```
sfc-gh-asawicki: q: should we ignore or throw validation error that it can be only used with the experiment enabled?

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -308,3 +310,476 @@ resource "snowflake_tag" "test" {
 }
 `, id.DatabaseName(), id.SchemaName(), id.Name())
 }
+
+func TestAcc_Tag_NoAllowedValues_WithoutExperimentFlag(t *testing.T) {
```
sfc-gh-asawicki, line 315: Could we have a test where, in the old behavior, the list is externally set to no tag allowed to verify how we react (with both empty and non-empty list)?

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -308,3 +310,476 @@ resource "snowflake_tag" "test" {
...
+						HasNameString(id.Name()).
+						HasAllowedValuesEmpty().
+						HasNoAllowedValuesString("true"),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_Tag_TransitionToExperimentFlag_NullAllowedValues(t *testing.T) {
```
sfc-gh-asawicki, line 764: let's do a similar test but where the list is restricted on 2.14 (so we have create+drop) and check that after migration the plan won't be empty (because we should in such a case clear it)

pkg/resources/tag.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: +1 addressed and tested

pkg/provider/experimentalfeatures/experimental_features.go
sfc-gh-jcieslak: Adjusted to `TAGS_ALLOW_EMPTY_ALLOWED_VALUES`

docs/resources/tag.md
sfc-gh-jcieslak: Sure, I changed the experiment a few times because I couldn't decide which name was better. Adjusted.

pkg/resources/tag.go
sfc-gh-jcieslak: I wrote about it in pr description, not sure yet about this change, because in some sense it's breaking change, but on the other hand it fixes problem that we address in other and new resources, where create is failing after Create command and setting resource id, which results in strange terraform state in the end (I think ultimately in this case the resource is not saved to state but object is created on the Snowflake side and has to be deleted manually; I may be wrong, but that's what I recall).

pkg/resources/tag.go
sfc-gh-jcieslak: Asked the tag team. They provided examples of how to transition between states, which is my current approach, but I may have mistakenly assumed there's no other way.

pkg/resources/tag.go
sfc-gh-jcieslak: yup, adjusted

pkg/resources/tag.go
sfc-gh-jcieslak: I can validate it here, but import is always a workaround for such validations, because there we don't have access to config values. Added validation.

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: Adjusted

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: I went with the ignore path, but maybe we should go with the validation error path. I'll adjust it to the validation path as it's handled this way in other resources (e.g. user and grants)

pkg/resources/tag.go
sfc-gh-jcieslak: Tag team confirmed it has to be done in two steps (adding and removing allowed values). We could "hack" this by using transactions or flow operator, but already discussed with @sfc-gh-asawicki that we should aim for simplicity first, and we can improve it later on (e.g. when we'll have flow operator abstraction in place).

pkg/resources/tag.go
sfc-gh-jcieslak: The error handling is different (even if extracted function would return error that would be differently handled, I'm not sure if it's worth to pull out 2 line common logic as it's only add + drop functions)

docs/index.md
sfc-gh-jcieslak: Got it, rephrased a bit

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: Sure, added TestAcc_Tag_ExternalChanges_WithoutExperimentFlag

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: Good point, added TestAcc_Tag_TransitionToExperimentFlag_EmptyAllowedValuesWithBasicConfig

pkg/provider/experimentalfeatures/experimental_features.go
sfc-gh-jmichalak: "allow ... allowed" but ok

docs/index.md
sfc-gh-asawicki: I still think it's written froim the technical perspective and not user perspective. Instead of naming this `nil`/`empty`/`set`, please focus on what it means for the user. The first paragraph IMO should be more something like this:
```
This option enables three possible ways to set allowed values:
- any value is allowed (`allowed_values` should be left empty);
- given values are allowed (`allowed_values` should be set to the desired values);
- no value is allowed (`no_allowed_values` field set to true and `allowed_values` absent in the config).
```
Similarly the next one.

pkg/resources/tag.go
sfc-gh-asawicki: If this is the only way available, let's document it on the registry level, so that it's clear flor the customer why they may get this strange value in the meantime.

sfc-gh-asawicki (changes requested):

pkg/resources/tag.go
sfc-gh-jcieslak: Done

pkg/resources/tag.go
sfc-gh-jmichalak: OK, I think we usually solve this with `d.Partial`, and optionally, we could mention it as an improvement in the migration guide.

sfc-gh-jcieslak: Hmm, I think Partial is Update only thing, but will check and write todo for myself to mention it in the migration gudie

PR #4499, opened 2026-02-25
chore: unskip connection tests for preprod environment

sfc-gh-jcieslak:
## Changes
- Unkipped connection tests for preprod environment (only for NON_PROD environment)
- Adjusted some of the assertions
- Added new account that lives on azure cloud
- Adjusted test initialization for acceptance tests to reduce clutter


pkg/testacc/resource_secondary_connection_acceptance_test.go
```diff
@@ -24,21 +25,23 @@ import (
...
+	}, "SNOW-3198924: Missing azure configuration on all testing environments")
 
 	// create primary connection
-	connection, connectionCleanup := secondaryTestClient().Connection.Create(t)
+	connection, connectionCleanup := azureTestClient().Connection.Create(t)
 	t.Cleanup(connectionCleanup)
 
 	accountId := testClient().Account.GetAccountIdentifier(t)
-	secondaryTestClient().Connection.Alter(t, sdk.NewAlterConnectionRequest(connection.ID()).
+	azureTestClient().Connection.Alter(t, sdk.NewAlterConnectionRequest(connection.ID()).
```
sfc-gh-jmichalak, line 41: If this is only for acceptance tests, can't we just define a new provider config with azure profile? We're polluting the test objects now.

pkg/testacc/0_setup_test.go
```diff
@@ -194,25 +175,61 @@ func (atc *acceptanceTestContext) initialize() error {
...
 }
 
-func cleanup() {
-	accTestLog.Printf("[INFO] Running acceptance tests cleanup")
-	if atc.databaseCleanup != nil {
-		defer atc.databaseCleanup()
-	}
-	if atc.schemaCleanup != nil {
-		defer atc.schemaCleanup()
+func (atc *acceptanceTestContext) initializeSnowflakeEnvironment(
```
sfc-gh-jmichalak, line 144: Nit: I don't think we need to pass these params, as they are fields of `acceptanceTestContext`.

pkg/testacc/0_setup_test.go
```diff
@@ -97,94 +103,69 @@ func (atc *acceptanceTestContext) initialize() error {
...
+			secondaryTestClient().EnsureQuotedIdentifiersIgnoreCaseIsSetToFalse(ctx),
 			secondaryTestClient().EnsureEssentialRolesExist(ctx),
-		)
-		if errs != nil {
+		); errs != nil {
 			return errs
 		}
+
+		// TODO(SNOW-3198924): For now, tests requiring multiple Snowflake instances on other clouds are done only on non-prod environment
+		if testenvs.GetSnowflakeEnvironmentWithProdDefault() == testenvs.SnowflakeNonProdEnvironment {
```
sfc-gh-jmichalak, line 122: Nit: It's a bit unclear why we must use Azure when the expected env does not match. Can you explain?

sfc-gh-jmichalak (dismissed):

pkg/testacc/0_setup_test.go
sfc-gh-asawicki: +1

pkg/testacc/0_setup_test.go
```diff
@@ -194,25 +175,61 @@ func (atc *acceptanceTestContext) initialize() error {
...
-		defer atc.secondaryWarehouseCleanup()
+	atc.cleanups = append(atc.cleanups, warehouseCleanup)
+	*warehouseField = warehouse
+
+	return nil
+}
+
+func cleanup() {
+	accTestLog.Printf("[INFO] Running acceptance tests cleanup")
+	for _, cleanupFunc := range slices.Backward(atc.cleanups) {
```
sfc-gh-asawicki, line 229: this changes the previous behavior with the defered cleanups (when we use defer, all the cleanups are run, even if any of them panics - it's different with this implem entation)

pkg/testacc/0_setup_test.go
```diff
@@ -33,26 +34,31 @@ var (
...
+
+	database  *sdk.Database
+	schema    *sdk.Schema
+	warehouse *sdk.Warehouse
+
+	secondaryDatabase  *sdk.Database
+	secondarySchema    *sdk.Schema
+	secondaryWarehouse *sdk.Warehouse
+
+	azureDatabase  *sdk.Database
```
sfc-gh-asawicki, line 59: if we are adding a new profile, we should update the CONTRIBUTING.md (there is a section with profiles)

pkg/testacc/0_setup_test.go
sfc-gh-asawicki, line 151: shouldnt it also be hidden behind simplified tests?

sfc-gh-asawicki (changes requested):

pkg/testacc/0_setup_test.go
sfc-gh-jcieslak: It's in the simplified tests "if", so it will be only initialized when: simplified tests are turned off and snowflake environment is set to non_prod (preprod).

sfc-gh-jcieslak: I'm not sure what "why we must use Azure when the expected env does not match" means. It's only initialized when: simplified tests are turned off and snowflake environment is set to non_prod (preprod).

pkg/testacc/0_setup_test.go
sfc-gh-jcieslak: Done

pkg/testacc/0_setup_test.go
sfc-gh-jcieslak: Well, the setup is the same, but the target fields are different. That's why they're passed as pointers. I simplified this to a single field, which encapsulates all fields necessary for a new testing environment profile. Note, I also considered an enum approach where an enum would be passed and based on its value, a given field would be set, but this way is easier and less error-prone (e.g., missing or wrong enum to field mapping).

pkg/testacc/0_setup_test.go
sfc-gh-jcieslak: Adjusted the behavior to match previous behavior. Now in case of panicking cleanup other cleanups won't be affected

pkg/testacc/resource_secondary_connection_acceptance_test.go
sfc-gh-jcieslak: Discussed with Artur, let's leave it as a new profile. Having ad-hoc setting creates more burden that we would have to resolve later on when more tests would require this profile. Let's introduce this as new profile, but let's hide its initialization behind flags, so that current CI setup and development won't be affected that much.

sfc-gh-asawicki (dismissed):

pkg/testacc/resource_primary_connection_acceptance_test.go
```diff
@@ -25,10 +26,12 @@ import (
 )
 
 func TestAcc_PrimaryConnection_Basic(t *testing.T) {
-	testenvs.SkipTestIfValueIn(t, testenvs.SnowflakeTestingEnvironment, []string{
-		string(testenvs.SnowflakeProdEnvironment),
-		string(testenvs.SnowflakePreProdGovEnvironment),
-	}, "SNOW-3198924: Missing azure configuration on all testing environments")
+	if slices.Contains([]testenvs.SnowflakeEnvironment{
```
sfc-gh-asawicki, line 29: nit (for next PR): It would be good to adjust the function/add another one, so that this logic is contained (like the previous execution SkipTestIfValueIn).

sfc-gh-asawicki (approved):

PR #4504, opened 2026-03-02
chore: External volume resource adjustments

sfc-gh-jmichalak:
## Summary
- Replace flat property-based `describe_output` (parent/name/type/value/default) with a structured, typed schema that exposes per-provider storage location details (S3, S3GOV, GCS, Azure, S3-compatible)
- Add state upgrader (v0 → v1) to migrate existing state to the new describe_output schema
- Refactor Create/Read/Update to use shared resource helpers (`stringAttributeCreateBuilder`, `booleanStringAttributeUnsetFallbackUpdate`, `errors.Join` for `d.Set`)
- Deduplicate `storageLocationDetailsToStateMaps` (previously duplicated in Import and Read)
- Simplify `CopySentinelStorageLocationItem` by copying struct fields directly instead of per-provider switch
- Add `ExternalVolumeDescribeOutputAssert` for structured describe_output assertions in acceptance tests
- Adjust acceptance tests for gov environment compatibility

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -24,12 +24,41 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.14.x ➞ v2.15.0
+
+### *(enhancement)* Rework of `snowflake_external_volume` resource
+#### *(breaking change)* `snowflake_external_volume` resource `describe_output` schema changed
+
+The `describe_output` attribute on the `snowflake_external_volume` resource has been restructured.
+Previously it was a flat list of property rows with `parent`, `name`, `type`, `value`, and `default` fields.
+It is now a single structured object with `active`, `comment`, `allow_writes`, and `storage_locations` fields,
+where `storage_locations` contains typed, per-provider sub-objects (`s3_storage_location`, `gcs_storage_location`,
+`azure_storage_location`, `s3_compat_storage_location`).
+
+A state upgrader handles the migration automatically. No changes to your resource configuration are required,
+but if you reference `describe_output` in other parts of your Terraform config (e.g. `output` blocks or `local` values),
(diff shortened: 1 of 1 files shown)


pkg/testacc/resource_external_volume_acceptance_test.go
```diff
@@ -114,27 +116,26 @@ func externalVolumeMultiple(s3StorageLocations config.Variable, gcsStorageLocati
 }
 
 // Test volume with s3 storage locations
-func TestAcc_External_Volume_S3(t *testing.T) {
+func TestAcc_ExternalVolume_BasicUseCase_S3(t *testing.T) {
+	if testenvs.GetSnowflakeEnvironmentWithProdDefault() == testenvs.SnowflakePreProdGovEnvironment {
+		t.Skip("Skipping test - Snowflake error 393962 (42601): External volumes in government deployments cannot use non-government S3 storage locations. Storage type S3 is not allowed. Please use S3GOV instead.")
```
sfc-gh-jcieslak, line 121: I guess should have a task to create one and unskip this

pkg/testacc/resource_external_volume_acceptance_test.go
```diff
@@ -2302,36 +2165,69 @@ func TestAcc_External_Volume_External_Changes(t *testing.T) {
...
+								EncryptionType:  azureEncryptionTypeNone,
+								AzureStorageLocation: &sdk.StorageLocationAzureDetails{
+									AzureTenantId: azureTenantId,
+								},
+							},
+						}),
 				),
 			},
-			// externally update comment
+			// add new azure storage location to position 3
```
sfc-gh-jcieslak, line 2221: nit: I know they were here before, but idk if we that many list transitions in tests

pkg/testacc/resource_external_volume_acceptance_test.go
sfc-gh-jmichalak: Hmm, I don't think we should unskip it someday, as it's an SF limitation. Still, this test can be run in gov environments.

pkg/testacc/resource_external_volume_acceptance_test.go
sfc-gh-jmichalak: Yeah, maybe a few are redundant, but still, I reduced time from ~220s to ~150s, it's not that terrible :D

PR #4505, opened 2026-03-03
feat: Add SDK integration tests for catalog integrations

sfc-gh-kwasilewski:
## Changes
- Add integration tests with autogenerated object assertions
- Add support for nested arrays in ParseCommaSeparatedStringArray function
- Adjust definition file as "ACCESS_DELEGATION_MODE = VENDED_CREDENTIALS" is the only allowed option for SAP_BDC integrations

### DIFF
--- a/pkg/sdk/testint/catalog_integrations_integration_test.go
+++ b/pkg/sdk/testint/catalog_integrations_integration_test.go
@@ -44,7 +44,7 @@ func TestInt_CatalogIntegrations(t *testing.T) {
 		t.Helper()
 		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "CATALOG_SOURCE", Type: "String", Value: string(catalogSource), Default: ""})
 		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "TABLE_FORMAT", Type: "String", Value: string(tableFormat), Default: ""})
-		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "ENABLED", Type: "Boolean", Value: fmt.Sprintf("%t", enabled), Default: "false"})
+		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "ENABLED", Type: "Boolean", Value: fmt.Sprintf("%t", enabled), Default: "true"})
 		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "REFRESH_INTERVAL_SECONDS", Type: "Integer", Value: fmt.Sprintf("%d", refreshIntervalSeconds), Default: "30"})
 		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "COMMENT", Type: "String", Value: comment, Default: ""})
 	}
@@ -353,14 +353,14 @@ func TestInt_CatalogIntegrations(t *testing.T) {
 
 		err := client.CatalogIntegrations.Alter(ctx, sdk.NewAlterCatalogIntegrationRequest(id).WithSet(*sdk.NewCatalogIntegrationSetRequest().
 			WithComment(sdk.StringAllowEmpty{Value: "new comment"}).
-			WithEnabled(true).
+			WithEnabled(false).
 			WithRefreshIntervalSeconds(120)))
 		require.NoError(t, err)
 
 		details, err := client.CatalogIntegrations.Describe(ctx, id)
 		require.NoError(t, err)
 
-		assertSharedProperties(t, details, sdk.CatalogIntegrationCatalogSourceTypeObjectStorage, sdk.CatalogIntegrationTableFormatDelta, true, 120, "new comment")
+		assertSharedProperties(t, details, sdk.CatalogIntegrationCatalogSourceTypeObjectStorage, sdk.CatalogIntegrationTableFormatDelta, false, 120, "new comment")
 	})
 
 	t.Run("alter catalog integration: bearer token", func(t *testing.T) {


pkg/sdk/catalog_integrations_gen_test.go
```diff
@@ -184,7 +184,7 @@ func TestCatalogIntegrations_Create(t *testing.T) {
 		assertOptsValidAndSQLEquals(t, opts, "CREATE CATALOG INTEGRATION %s "+
 			"CATALOG_SOURCE = SAP_BDC "+
 			"TABLE_FORMAT = DELTA "+
-			"REST_CONFIG = (SAP_BDC_INVITATION_LINK = '%s') "+
+			"REST_CONFIG = (SAP_BDC_INVITATION_LINK = '%s' ACCESS_DELEGATION_MODE = VENDED_CREDENTIALS) "+
```
sfc-gh-jmichalak, line 187: Hmm, I think it shouldn't be added automatically, as it's an optional field? We should allow this to be not set. Maybe use `OptionalSQL` in the def file.

pkg/sdk/catalog_integrations_ext.go
```diff
@@ -0,0 +1,286 @@
...
+				params.BearerRestAuthentication = bearerRestAuth
+				params.SigV4RestAuthentication = sigV4RestAuth
+			}
+		}
+	}
+	return params, errors.Join(errs...)
+}
+
+func parseSapBdcProperties(properties []CatalogIntegrationProperty) (*SapBdcParams, error) {
+	params := &SapBdcParams{RestConfig: SapBdcRestConfig{}}
```
sfc-gh-jmichalak, line 140: Q: Is this empty on purpose? Doesn't snowflake return anything meaningful? :cry:

pkg/sdk/catalog_integrations_ext.go
```diff
@@ -0,0 +1,286 @@
...
+		case "SIGV4_SIGNING_REGION":
+			// SIGV4_SIGNING_REGION is always returned, even if unset
+			restAuthentication.Sigv4SigningRegion = String(v)
+		}
+		// SIGV4_EXTERNAL_ID not returned
+	}
+	return restAuthentication, nil
+}
+
+func parseCommaSeparatedEnumMap(property CatalogIntegrationProperty) []string {
```
sfc-gh-jmichalak, line 275: Nit: can we have some basic unit tests for either this function or parse... functions? At least to see what the typical outputs look like.

pkg/sdk/parsers.go
```diff
@@ -19,17 +19,19 @@ func ParseTimestampWithOffset(s string, dateTimeFormat string) (string, error) {
...
-// ParseCommaSeparatedStringArray can be used to parse Snowflake output containing a list in the format of "[item1, item2, ...]",
-// the assumptions are that:
-// 1. The list is enclosed by [] brackets, and they shouldn't be a part of any item's value
-// 2. Items are separated by commas, and they shouldn't be a part of any item's value
+// ParseCommaSeparatedStringArray can be used to parse Snowflake output containing a list in the format of "[item1, item2, ...]".
+// It also supports nested lists like "[[item1, item2], [item3, item4]]".
+// The assumptions are that:
+// 1. The list may be enclosed by outer [] brackets which are stripped
+// 2. Items are separated by top-level commas (commas inside nested brackets are preserved)
 // 3. Items can have as many spaces in between, but after separation they will be trimmed and shouldn't be a part of any item's value
```
sfc-gh-jmichalak: 1. When the list is nested, trimming quotes doesn't really work, please add this as a limitation.
2. What happens now if we run this with e.g. `"a", "b", "c[d"`? I think it can break, though having such output from SF is rather unlikely. But let's add some more tests for weird formatting and document the limitations.

pkg/sdk/catalog_integrations_ext.go
```diff
@@ -0,0 +1,286 @@
...
+
+func (v *catalogIntegrations) DescribeSapBdcParams(ctx context.Context, id AccountObjectIdentifier) (*SapBdcParams, error) {
+	properties, err := v.Describe(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	return parseSapBdcProperties(properties)
+}
+
+func parseAwsGlueProperties(properties []CatalogIntegrationProperty) (*AwsGlueParams, error) {
```
sfc-gh-jmichalak, line 56: Q: What about REFRESH_INTERVAL_SECONDS? Is it not returned?

pkg/sdk/catalog_integrations_ext.go
sfc-gh-kwasilewski: Yes. The only SAP-specific property, `SAP_BDC_INVITATION_LINK`, is not in the DESCRIBE output.

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-kwasilewski: It's actually a required field, as the command fails without it:

> Missing option(s): REST_CONFIG.ACCESS_DELEGATION_MODE

Either the docs are wrong, or there's a bug in implementation.

pkg/sdk/parsers.go
sfc-gh-kwasilewski: 2. The last element will be skipped entirely. Thanks, I'll try to add more test cases

pkg/sdk/catalog_integrations_ext.go
sfc-gh-kwasilewski: It's not a part of `AwsGlueParams`, so it isn't returned here. It's tested in the `assertSharedProperties` helper.

pkg/sdk/testint/catalog_integrations_integration_test.go
```diff
@@ -0,0 +1,588 @@
...
+	assertSharedProperties := func(t *testing.T, details []sdk.CatalogIntegrationProperty, catalogSource sdk.CatalogIntegrationCatalogSourceType, tableFormat sdk.CatalogIntegrationTableFormat, enabled bool, refreshIntervalSeconds int, comment string) {
+		t.Helper()
+		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "CATALOG_SOURCE", Type: "String", Value: string(catalogSource), Default: ""})
+		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "TABLE_FORMAT", Type: "String", Value: string(tableFormat), Default: ""})
+		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "ENABLED", Type: "Boolean", Value: fmt.Sprintf("%t", enabled), Default: "false"})
+		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "REFRESH_INTERVAL_SECONDS", Type: "Integer", Value: fmt.Sprintf("%d", refreshIntervalSeconds), Default: "30"})
+		assert.Contains(t, details, sdk.CatalogIntegrationProperty{Name: "COMMENT", Type: "String", Value: comment, Default: ""})
+	}
+
+	cleanupCatalogIntegrationProvider := func(id sdk.AccountObjectIdentifier) func() {
```
sfc-gh-jcieslak: nit: We usually use `testClientHelper().object.DropFunc`

sfc-gh-jcieslak: \+ most of the functions below could be extracted to the test client

sfc-gh-jcieslak (approved):

pkg/sdk/parsers.go
sfc-gh-kwasilewski: I extracted a separate parsing function and added extra test cases

pkg/sdk/catalog_integrations_ext.go
sfc-gh-kwasilewski: I added basic tests for parseCommaSeparatedEnumMap with a small number of test cases - I didn't want to duplicate existing ParseCommaSeparatedStringArray tests

pkg/sdk/testint/catalog_integrations_integration_test.go
sfc-gh-kwasilewski: Replaced with `testClientHelper().object.DropFunc`. I didn't extract more functions because these create operations require a lot of parameters, and I think it's easier to populate them here in the test file.

pkg/sdk/catalog_integrations_gen_test.go
sfc-gh-jmichalak: OK, please report it to the docs team.

pkg/sdk/testint/catalog_integrations_integration_test.go
sfc-gh-jcieslak: Sure, as you have already done it this way, we can proceed. We created helper functions earlier, as you did here, but then moved to the test client first approach, as there's an additional benefit (by adding helper functions now, testing other objects that have catalog integration as a dependency will be much easier, and we keep the integration tests smaller).

pkg/sdk/parsers_test.go
```diff
@@ -7,113 +7,144 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type commaSeparatedStringArrayTestCase struct {
```
sfc-gh-jcieslak: nit: to not "pollute" package context we usually create anonymous structs like this:
```go
var testCases = []struct{
   FieldA string
}{
  { FieldA: "some string" }
}
```

sfc-gh-kwasilewski: Fixed

sfc-gh-jcieslak (approved):

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
```diff
@@ -200,6 +200,42 @@ var allStructs = []SdkObjectDef{
 		IdType:       "sdk.AccountObjectIdentifier",
 		ObjectStruct: sdk.StorageLocationS3CompatDetails{},
 	},
+	{
+		IdType:       "sdk.AccountObjectIdentifier",
+		ObjectStruct: sdk.CatalogIntegration{},
+	},
+	{
+		IdType:       "sdk.AccountObjectIdentifier",
```
sfc-gh-asawicki, line 208: We are not adding these sub-structures here but in .ext files instead. Adding them here requires manual alterations in the generated files, which adds another thing that needs to be done when we want to make these run without any manual alterations.

PR #4508, opened 2026-03-03
chore: Support adaptive warehouses in SDK

sfc-gh-jmichalak:
Snowflake introduced a new "adaptive" warehouse type that behaves differently from standard and Snowpark-optimized warehouses — many fields (size, cluster counts, scaling policy, auto-suspend, query acceleration, etc.) are not applicable and return NULL from `SHOW WAREHOUSES`. The SDK and resource need to handle these nullable fields to avoid parsing errors and incorrect state.

- Add `ADAPTIVE` warehouse type to the SDK with a dedicated `CreateAdaptive` operation and `CreateAdaptiveWarehouseOptions` struct
- Change multiple `Warehouse` struct fields from value types to pointer types (`Size`, `MinClusterCount`, `MaxClusterCount`, `StartedClusters`, `Running`, `Queued`, `AutoSuspend`, `EnableQueryAcceleration`, `QueryAccelerationMaxScaleFactor`, `ScalingPolicy`) to handle NULL values from Snowflake
- Update `warehouseDBRow` to use `sql.NullString`, `sql.NullInt64`, `sql.NullBool` for nullable columns
- Add `ToWarehouseTypeUserSettable` to prevent users from setting `ADAPTIVE` type through the standard warehouse resource (must use the dedicated create operation)
- Add helper functions (`setOptionalValue`, `optionalStringOutputMapping`, `optionalBooleanStringOutputMapping`, `optionalIntOutputMapping`, `mapNullInt`) for handling optional/nullable fields
- Update warehouse resource import, read, and update operations to handle pointer types
- Add `HasNo*` assertion methods for all newly-nullable warehouse fields
- Update show output schema to conditionally set nullable fields
- Add adaptive warehouse helpers and integration tests (minimal and complete creation)
- Update migration script to reject adaptive warehouses and handle pointer types
- Update all existing tests to use pointer comparisons

## TODO

- Add support for the new columns in SHOW.
- Add missing BRSF field once it's ready.
- Add new resource for adaptive warehouses.
- Add tests for handling adaptive warehouses by the regular resource and vice-versa.
- Add a test for data source with adaptive warehouses.

### DIFF
--- a/pkg/sdk/testint/warehouses_integration_test.go
+++ b/pkg/sdk/testint/warehouses_integration_test.go
@@ -292,7 +292,7 @@ func TestInt_Warehouses(t *testing.T) {
 		id := testClientHelper().Ids.RandomAccountObjectIdentifier()
 		err := client.Warehouses.CreateAdaptive(ctx, id, &sdk.CreateAdaptiveWarehouseOptions{
 			Comment:                         sdk.String("test adaptive warehouse"),
-			MaxStatementSize: sdk.Pointer(sdk.MaxStatementSizeMedium),
+			MaxStatementSize:                sdk.Pointer(sdk.MaxStatementSizeMedium),
 			StatementQueuedTimeoutInSeconds: sdk.Int(30),
 			StatementTimeoutInSeconds:       sdk.Int(60),
 		})
@@ -452,9 +452,19 @@ func TestInt_Warehouses(t *testing.T) {
 		})
 		require.NoError(t, err)
 
-		returnedWarehouse, err := client.Warehouses.ShowByID(ctx, warehouse.ID())
-		require.NoError(t, err)
-		assert.Equal(t, *warehouse, *returnedWarehouse)
+		assertThatObject(t, objectassert.Warehouse(t, warehouse.ID()).
+			HasType(sdk.WarehouseTypeStandard).
+			HasSize(sdk.WarehouseSizeMedium).
+			HasGeneration(sdk.WarehouseGenerationStandardGen1).
+			HasNoResourceConstraint().
+			HasMaxClusterCount(1).
+			HasMinClusterCount(1).
+			HasScalingPolicy(sdk.ScalingPolicyStandard).
+			HasAutoSuspend(600).
+			HasAutoResume(true).
+			HasEnableQueryAcceleration(false).
+			HasQueryAccelerationMaxScaleFactor(8),
+		)
 	})
 
 	t.Run("alter: set and unset parameters", func(t *testing.T) {
--- a/pkg/scripts/migration_script/program_test.go
+++ b/pkg/scripts/migration_script/program_test.go
@@ -72,7 +72,7 @@ object_type represents the type of Snowflake object you want to generate terrafo
 			To support object parameters, one should use the SHOW PARAMETERS output, and combine it with the SHOW WAREHOUSES output, so the CSV header looks like "comment","created_on",...,"max_cluster_count","min_cluster_count","name","other",...
 			When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "WAREHOUSE".
 			The script always outputs fields that have non-empty default values in Snowflake (they can be removed from the output)
-			Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid, initially_suspended)
+			Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid, initially_suspended). Adaptive warehouses are also not supported.
 			For more details about using multiple sources, visit https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/scripts/migration_script/README.md#multiple-sources
 			Supported resources:
 				- snowflake_warehouse


pkg/sdk/warehouses.go
```diff
@@ -632,46 +731,48 @@ type warehouseDBRow struct {
...
+	mapNullBool(&wh.AutoResume, row.AutoResume)
+	mapNullBool(&wh.EnableQueryAcceleration, row.EnableQueryAcceleration)
+	mapNullInt(&wh.QueryAccelerationMaxScaleFactor, row.QueryAccelerationMaxScaleFactor)
+	mapNullStringWithMapping(&wh.ScalingPolicy, row.ScalingPolicy, ToScalingPolicy)
+	if row.Size.Valid {
+		if size, err := ToWarehouseSize(row.Size.String); err != nil {
+			return nil, err
+		} else {
+			wh.Size = &size
+		}
```
sfc-gh-kwasilewski, line 778: `mapNullStringWithMapping`? Similarly to what you've done to Generation

pkg/sdk/warehouses.go
```diff
@@ -325,6 +375,55 @@ func (c *warehouses) Create(ctx context.Context, id AccountObjectIdentifier, opt
...
+
+	// Object properties
+	Comment              *string              `ddl:"parameter,single_quotes" sql:"COMMENT"`
+	TargetStatementSize  *TargetStatementSize `ddl:"parameter,single_quotes" sql:"TARGET_STATEMENT_SIZE"`
+	WarehouseCreditLimit *int                 `ddl:"parameter" sql:"WAREHOUSE_CREDIT_LIMIT"`
+
+	// Object params
+	StatementQueuedTimeoutInSeconds *int             `ddl:"parameter" sql:"STATEMENT_QUEUED_TIMEOUT_IN_SECONDS"`
+	StatementTimeoutInSeconds       *int             `ddl:"parameter" sql:"STATEMENT_TIMEOUT_IN_SECONDS"`
+	Tag                             []TagAssociation `ddl:"keyword,parentheses" sql:"TAG"`
```
sfc-gh-kwasilewski, line 397: According to docs, tags are listed higher than object params in the CREATE command.

pkg/sdk/warehouses.go
```diff
@@ -581,16 +680,16 @@ type Warehouse struct {
...
+	MaxClusterCount                 *int
+	StartedClusters                 *int
+	Running                         *int
+	Queued                          *int
 	IsDefault                       bool
 	IsCurrent                       bool
-	AutoSuspend                     int
-	AutoResume                      bool
+	AutoSuspend                     *int
+	AutoResume                      *bool
```
sfc-gh-kwasilewski: Is this property nullable? It's set to true even when creating an adaptive warehouse without options, which your integration test demonstrates.

pkg/sdk/warehouses.go
sfc-gh-kwasilewski, line 682: What about TargetStatementSize and WarehouseCreditLimit? They are present in the SHOW output.

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -260,21 +261,81 @@ func TestInt_Warehouses(t *testing.T) {
...
+			HasStatementTimeoutInSeconds(172800),
+		)
+	})
+
+	t.Run("create adaptive: complete", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomAccountObjectIdentifier()
+		err := client.Warehouses.CreateAdaptive(ctx, id, &sdk.CreateAdaptiveWarehouseOptions{
+			Comment:                         sdk.String("test adaptive warehouse"),
+			TargetStatementSize:             sdk.Pointer(sdk.TargetStatementSizeMedium),
+			WarehouseCreditLimit:            sdk.Int(100),
```
sfc-gh-kwasilewski: We should have assertions for these

sfc-gh-jmichalak: Yeah. For now, I did not add these columns (next PR will add them to sdk, to assertions, the data source, proving DS works for all types). We should also discuss if we want to modify the standard warehouse show output or keep the old one, as all the new columns are null.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: As above - new columns will be addressed in the next PR.

pkg/resources/resource_helpers_read.go
```diff
@@ -76,6 +76,10 @@ func setOptionalValueWithMapping[T, R any](d *schema.ResourceData, key string, v
 }
 
 func setOptionalFromStringPtr(d *schema.ResourceData, key string, ptr *string) error {
+	return setOptionalValue(d, key, ptr)
+}
+
+func setOptionalValue[T any](d *schema.ResourceData, key string, ptr *T) error {
```
sfc-gh-jcieslak: Hmm, not particularly regarding this PR, but I think this function may be wrong. Sometimes, we should set a given key to null whenever the pointer is null (I'm not sure, but in this case we could just call `d.Set(key, ptr)`). We should separate cases where it doesn't matter and when it does. In the case of warehouses, we always set values from SHOW in Read, but now we do this only whenever the values are not-null, which I guess follows the previous behavior, but just flagging this as potential source of bugs.

pkg/scripts/migration_script/README.md
```diff
@@ -114,7 +114,7 @@ where script options are:
       When the additional columns are present, the resulting resource will have the parameters values, if the parameter level is set to "WAREHOUSE".
       The script always outputs fields that have non-empty default values in Snowflake (they can be removed from the output)
 
-      Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid, initially_suspended)
+      Caution: Some of the fields are not supported (actives, pendings, failed, suspended, uuid, initially_suspended). Adaptive warehouses are also not supported.
```
sfc-gh-jcieslak, line 117: I think the `-h` flag should be also updated, right?

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -336,18 +397,79 @@ func TestInt_Warehouses(t *testing.T) {
...
+
+		// Change warehouse type from standard to adaptive
+		err := client.Warehouses.Alter(ctx, warehouse.ID(), &sdk.AlterWarehouseOptions{
+			Set: &sdk.WarehouseSet{WarehouseType: sdk.Pointer(sdk.WarehouseTypeAdaptive)},
+		})
+		require.NoError(t, err)
+
+		returnedWarehouse, err := client.Warehouses.ShowByID(ctx, warehouse.ID())
+		require.NoError(t, err)
+		assertThatObject(t, objectassert.WarehouseFromObject(t, returnedWarehouse).
```
sfc-gh-jcieslak: nit: Can just use `objectassert.Warehouse` to automatically query the warehouse

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -336,18 +397,79 @@ func TestInt_Warehouses(t *testing.T) {
...
+			Set: &sdk.WarehouseSet{
+				WarehouseType: sdk.Pointer(sdk.WarehouseTypeStandard),
+				WarehouseSize: sdk.Pointer(sdk.WarehouseSizeMedium),
+			},
+		})
+		require.NoError(t, err)
+
+		returnedWarehouse, err = client.Warehouses.ShowByID(ctx, warehouse.ID())
+		require.NoError(t, err)
+		assertThatObject(t, objectassert.WarehouseFromObject(t, returnedWarehouse).
```
sfc-gh-jcieslak: nit: we could just compare returnedWarehouse with warehouse you created at the beginning

pkg/sdk/warehouses.go
```diff
@@ -261,6 +279,38 @@ func (s WarehouseGeneration) ToWarehouseResourceConstraint() (WarehouseResourceC
...
 
+type TargetStatementSize string
+
+const (
+	TargetStatementSizeXSmall   TargetStatementSize = "XSMALL"
+	TargetStatementSizeSmall    TargetStatementSize = "SMALL"
+	TargetStatementSizeMedium   TargetStatementSize = "MEDIUM"
+	TargetStatementSizeLarge    TargetStatementSize = "LARGE"
+	TargetStatementSizeXLarge   TargetStatementSize = "XLARGE"
+	TargetStatementSizeXXLarge  TargetStatementSize = "XXLARGE"
```
sfc-gh-jcieslak: question: Are synonyms accepted like `2XLARGE`?

pkg/sdk/warehouses.go
```diff
@@ -325,6 +375,55 @@ func (c *warehouses) Create(ctx context.Context, id AccountObjectIdentifier, opt
 	return err
 }
 
+// CreateAdaptiveWarehouseOptions is based on https://docs.snowflake.com/en/LIMITEDACCESS/adaptive-warehouses
+type CreateAdaptiveWarehouseOptions struct {
+	create            bool                    `ddl:"static" sql:"CREATE"`
+	OrReplace         *bool                   `ddl:"keyword" sql:"OR REPLACE"`
+	adaptiveWarehouse bool                    `ddl:"static" sql:"ADAPTIVE WAREHOUSE"`
```
sfc-gh-jcieslak: strange, shouldn't it be create warehouse type = 'adaptive'?

pkg/sdk/warehouses.go
```diff
@@ -325,6 +375,55 @@ func (c *warehouses) Create(ctx context.Context, id AccountObjectIdentifier, opt
...
+	IfNotExists       *bool                   `ddl:"keyword" sql:"IF NOT EXISTS"`
+	name              AccountObjectIdentifier `ddl:"identifier"`
+
+	// Object properties
+	Comment              *string              `ddl:"parameter,single_quotes" sql:"COMMENT"`
+	TargetStatementSize  *TargetStatementSize `ddl:"parameter,single_quotes" sql:"TARGET_STATEMENT_SIZE"`
```
sfc-gh-jcieslak: Not sure If i'm reading it right, but shouldn't there be max_statement_size param?

pkg/sdk/warehouses.go
sfc-gh-jmichalak: There are two ways to create them. It seems that the type is preferred, so I changed it.

pkg/scripts/migration_script/README.md
sfc-gh-jmichalak: Done.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jmichalak: Done.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: It seems like they are, but they are undocumented. Let's leave this for now.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jmichalak: We can't, as there are fields that change, e.g. updated_on.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: Adjusted.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: Yes, this resource is changing a lot, there are more changes on our side.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: These helpers don't return errors, and I wanted to keep compatibility here.

pkg/sdk/warehouses.go
sfc-gh-jmichalak: Reverted.

pkg/resources/resource_helpers_read.go
sfc-gh-jmichalak: We don't always set resource_constraint and generation, because they are handled by withExternalChangesMarking block. These fields are mutually exclusive in warehouses.

pkg/acceptance/bettertestspoc/assert/objectassert/warehouse_snowflake_ext.go
```diff
@@ -99,8 +99,8 @@ func (w *WarehouseAssert) HasNoAutoSuspend() *WarehouseAssert {
 func (w *WarehouseAssert) HasNoAutoResume() *WarehouseAssert {
```
sfc-gh-jcieslak, line 99: nit: this could be removed (in generated asserts we have `HasAutoResume(false)`)

pkg/resources/resource_helpers_read.go
sfc-gh-jcieslak: Right, just pointing out that this helper should be removed since we don't need it.

pkg/sdk/warehouses.go
sfc-gh-kwasilewski: Was it? I don't see any changes to ordering

sfc-gh-jmichalak: Hmm, for some reason, it wasn't checked in. I'll do this in the next PR

pkg/schemas/warehouse_gen.go
```diff
@@ -134,15 +134,36 @@ func WarehouseToSchema(warehouse *sdk.Warehouse) map[string]any {
 	warehouseSchema["name"] = warehouse.Name
 	warehouseSchema["state"] = string(warehouse.State)
 	warehouseSchema["type"] = string(warehouse.Type)
-	warehouseSchema["size"] = string(warehouse.Size)
-	warehouseSchema["min_cluster_count"] = warehouse.MinClusterCount
-	warehouseSchema["max_cluster_count"] = warehouse.MaxClusterCount
-	warehouseSchema["started_clusters"] = warehouse.StartedClusters
-	warehouseSchema["running"] = warehouse.Running
-	warehouseSchema["queued"] = warehouse.Queued
+	// Adjusted manually
```
sfc-gh-asawicki, line 137: q: are these needed manually? We have the support for nullable fields in this generator.

sfc-gh-jmichalak: You're right that the generator handles nullable fields in the template. However, there's a gap: the mapper still receives the pointer itself (e.g. warehouse.Size) rather than the dereferenced value, producing invalid Go for pointer-to-enum fields:

```
// generator produces (doesn't compile):
if warehouse.Size != nil {
    warehouseSchema["size"] = string(warehouse.Size) // *WarehouseSize, not WarehouseSize
}
```

For plain *int/*bool fields, the Identity mapper doesn't need a cast, so those are fine. I think the fix should be rather straightforward in MapToSchemaField - we could use `sdk.Pointer` for nullable enums, or just dereference optional fields.

sfc-gh-asawicki: But for *int/*bool fields we could also dereference, as we are apparently not doing it. Nevertheless, the fields marked as manual are not only the enum ones; let's have a follow-up on that, so that it's clarified and fixed where needed.

PR #4511, opened 2026-03-04
chore: stream changes for bcr 2026 01

sfc-gh-jcieslak:
A follow-up for #4484 
Addresses SNOW-1733130 (after BCR 2026_01, the stage identifiers are returned as FQNs)

## Changes
- Adjust stream to return identifiers for base tables and table names. Confirmed internally, they are always non-null.
- Handle edge cases where the dependent object is missing, or the querying user/role doesn't have enough privileges to see it
- Adjust resources to use FQN (after confirmation, we can assume they are non-null; otherwise error will be returned by SHOW
- Adjust the diff suppressor used in a previously invalid stream resource. It was "upgraded" and left for backward compatibility reasons.
- Adjust test for above chagnes / Add tests for above changes
- Migration guide


pkg/resources/stream_on_directory_table.go
```diff
@@ -112,17 +111,7 @@ func ReadStreamOnDirectoryTable(withDirectoryChangesMarking bool) schema.ReadCon
...
-			return diag.Diagnostics{
-				diag.Diagnostic{
-					Severity: diag.Error,
-					Summary:  "Could not parse stage id",
-					Detail:   fmt.Sprintf("stream name: %s", id.FullyQualifiedName()),
-				},
-			}
-		}
-		if err := d.Set("stage", *stream.TableName); err != nil {
+		if err := d.Set("stage", stream.TableName.FullyQualifiedName()); err != nil {
```
sfc-gh-jcieslak, line 126: Table name is still nullable, but we've got a confirmation that it won't ever be null (it can only be null when calling SHOW with the TERSE option). It's either proper identifier or placeholder that we handle (added in this pr).

pkg/sdk/generator/defs/streams_def.go
```diff
@@ -46,17 +46,18 @@ var (
...
-				Field("TableName", "*string").
-				Field("SourceType", "*StreamSourceType").
-				Field("BaseTables", "[]string").
-				Field("Type", "*string").
-				Field("Stale", "bool").
-				Field("Mode", "*StreamMode").
-				Field("StaleAfter", "*time.Time").
-				Field("InvalidReason", "*string").
-				Field("OwnerRoleType", "*string")
+		// Below fields have to be nullable as they are not returned by SHOW TERSE streams
```
sfc-gh-jcieslak, line 49: Except stale as it was treated in the conversion as `null/false -> false` and `true -> true` (of course null != false, but on the other hand we don't use terse queries). Idk, can adjust this as well, but I would rather do it in the follow-up.

sfc-gh-jmichalak: Hmm. On the other hand, if we had terse operations separated (I think we don't use them anyway), we could have also separate structs for these results, leading to a simpler implementation. Just food for thought for the next SDK objects.

pkg/testacc/data_source_streams_acceptance_test.go
```diff
@@ -267,9 +267,8 @@ func TestAcc_StreamOnDirectoryTable(t *testing.T) {
 						HasDatabaseName(id.DatabaseName()).
 						HasSchemaName(id.SchemaName()).
 						HasOwner(snowflakeroles.Accountadmin.Name()).
-						HasTableName(fmt.Sprintf(`"%s"."%s".%s`, stage.ID().DatabaseName(), stage.ID().SchemaName(), stage.ID().Name())).
+						HasTableName(stage.ID().FullyQualifiedName()).
 						HasSourceType(sdk.StreamSourceTypeStage).
-						HasBaseTablesPartiallyQualified(fmt.Sprintf(`"%s"."%s".%s`, stage.ID().DatabaseName(), stage.ID().SchemaName(), stage.ID().Name())).
 						HasType("DELTA").
```
sfc-gh-jmichalak, line 273: HasBaseTables assertion is now missing.

pkg/testacc/resource_stream_on_directory_table_acceptance_test.go
```diff
@@ -87,20 +84,19 @@ func TestAcc_StreamOnDirectoryTable_BasicUseCase(t *testing.T) {
...
 			HasDatabaseName(id.DatabaseName()).
 			HasSchemaName(id.SchemaName()).
-			HasTableName(expectedStageId).
+			HasTableName(stage.ID().FullyQualifiedName()).
 			HasMode(sdk.StreamModeDefault).
 			HasComment(comment).
 			HasOwner(testClient().Context.CurrentRole(t).Name()).
 			HasSourceType(sdk.StreamSourceTypeStage).
-			HasBaseTablesPartiallyQualified(expectedStageId).
 			HasType("DELTA").
```
sfc-gh-jmichalak, line 102: Same here.

pkg/testacc/resource_stream_on_directory_table_acceptance_test.go
```diff
@@ -450,3 +446,93 @@ func TestAcc_StreamOnDirectoryTable_ExternalStreamTypeChange(t *testing.T) {
...
+						HasDatabaseString(id.DatabaseName()).
+						HasSchemaString(id.SchemaName()).
+						HasStageString(stage.ID().FullyQualifiedName()),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_StreamOnDirectoryTable_migrateFromV2_14_0_NoPlanWithoutConfigChanges(t *testing.T) {
```
sfc-gh-jmichalak: These tests seem identical.

pkg/resources/diff_suppressions.go
```diff
@@ -240,20 +240,22 @@ func suppressIdentifierQuoting(_, oldValue, newValue string, _ *schema.ResourceD
 	return slices.Equal(oldId, newId)
 }
 
-func suppressIdentifierQuotingPartiallyQualifiedName(_, oldValue, newValue string, _ *schema.ResourceData) bool {
+func suppressPartiallyQualifiedSchemaObjectIdentifier(_, oldValue, newValue string, _ *schema.ResourceData) bool {
```
sfc-gh-jmichalak: I think we don't need this function anymore. We can just have a state upgrader for this version, updating this field value, mitigating the diff.

pkg/schemas/stream.go
```diff
@@ -91,34 +88,12 @@ func StreamDescriptionToSchema(stream sdk.Stream) map[string]any {
...
-		} else {
-			tableId, err := sdk.ParseSchemaObjectIdentifier(*stream.TableName)
-			if err != nil {
-				log.Printf("[DEBUG] could not parse table ID: %v", err)
-			} else {
-				streamSchema["table_name"] = tableId.FullyQualifiedName()
-			}
-		}
-	}
+	streamSchema["table_name"] = stream.TableName.FullyQualifiedName()
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: \+1

sfc-gh-jmichalak (dismissed):

pkg/schemas/stream.go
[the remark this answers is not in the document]
sfc-gh-jcieslak: Done

pkg/sdk/generator/defs/streams_def.go
sfc-gh-jcieslak: Good idea, options like TERSE shouldn't impact layers above SDK that much, separation is a good option to avoid it.

pkg/resources/diff_suppressions.go
sfc-gh-jcieslak: Yeah, I initially removed it, but then brought it back because it would produce plans, but I think you may be right that the upgrader would solve this. The state would be different, but no plan would be produced (because of the other diff suppressors). I'll adjust it.

pkg/testacc/data_source_streams_acceptance_test.go
sfc-gh-jcieslak: Added

pkg/testacc/resource_stream_on_directory_table_acceptance_test.go
sfc-gh-jcieslak: Added

pkg/testacc/resource_stream_on_directory_table_acceptance_test.go
sfc-gh-jcieslak: Yeah, removed one as it was unnecessary

pkg/resources/diff_suppressions.go
sfc-gh-jcieslak: Not sure if the state upgrader is even needed in this case, but I'll leave it just in case

pkg/resources/stream_on_directory_table_state_upgraders.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func v2_14_0_StreamOnDirectoryTableStateUpgrader(ctx context.Context, state map[string]any, meta any) (map[string]any, error) {
+	if state == nil {
+		return state, nil
+	}
+
+	stageId, err := sdk.ParseSchemaObjectIdentifier(state["stage"].(string))
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: stage is required param

pkg/resources/stream_on_directory_table_state_upgraders.go
```diff
@@ -0,0 +1,22 @@
+package resources
+
+import (
+	"context"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+)
+
+func v2_14_0_StreamOnDirectoryTableStateUpgrader(ctx context.Context, state map[string]any, meta any) (map[string]any, error) {
```
sfc-gh-asawicki, line 9: nit: v2_15_0?

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

PR #4516, opened 2026-03-06
chore: translate the semantic view describe output into a struct

sfc-gh-sarur:
Restructured the Semantic View describe output into a struct in the sdk. We are now using this new struct as input for the describe output. Also updated the validation code to use the new struct. Integration tests and acceptance tests successful.

## Test Plan
* [x] acceptance tests
* [x] integration tests


pkg/acceptance/bettertestspoc/assert/objectassert/semantic_view_describe_snowflake_ext.go
```diff
@@ -49,13 +49,35 @@ func (s *SemanticViewDetailsAssert) HasDetailsCount(expected int) *SemanticViewD
 func (s *SemanticViewDetailsAssert) ContainsDetail(expected sdk.SemanticViewDetails) *SemanticViewDetailsAssert {
```
sfc-gh-jmichalak, line 49: This is a step in the right direction. Ultimately, we'd like to have more static types in `SemanticViewDescribeDetails` fields, instead of the current generic list. So, the resulting struct should also be parsed from `SemanticViewDetails`. Please see stages/storage integrations for reference, we don't just handle the top-level generic fields like here, but also nested fields.



The end goal will allow us to remove this method and just assert the `SemanticViewDescribeDetails`'s fields (add it to `objectassert/gen/sdk_object_def.go` and run the generators to get the assertions.

PR #4525, opened 2026-03-11
fix(sdk): map CORTEX_AGENT_SERVER to MCP SERVER in grant show

landiinii:
Added new mappings for the new "MCP Server" object type that returns in the `SHOW GRANTS` query as a different type: "CORTEX_AGENT_SERVER"

## Test Plan
* [X] acceptance tests - included an acceptance test that creates a grant, applies, the re-applies and asserts the plan returns empty

* [X] Ran a local dev override on my actual implementation, worked like a charm

* [X] Also ran unit tests, lint, and pre push

## References
Resolves issue https://github.com/snowflakedb/terraform-provider-snowflake/issues/4524

This fix is almost identical to the fix made for models: https://github.com/snowflakedb/terraform-provider-snowflake/issues/3050
Also similar to the fix made for Volumes: https://github.com/snowflakedb/terraform-provider-snowflake/pull/2538

### DIFF
--- a/pkg/sdk/grants.go
+++ b/pkg/sdk/grants.go
@@ -239,6 +239,9 @@ func (row grantRow) convert() (*Grant, error) {
 	if row.GrantedOn == "MODULE" {
 		grantedOn = ObjectTypeModel
 	}
+	if row.GrantedOn == "CORTEX_AGENT_SERVER" {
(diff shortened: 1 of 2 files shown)


landiinii:
Getting fixed in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4532 instead

PR #4528, opened 2026-03-12
chore: Rework catalog integration describe

sfc-gh-kwasilewski:
## Summary

Avoid a second DESCRIBE when building the full picture of a catalog integration. Previously, type-specific data (e.g. AWS Glue, Iceberg REST) and common data were fetched via separate describe paths, so a resource could end up issuing two DESCRIBE queries.

### DIFF
--- a/pkg/acceptance/bettertestspoc/assert/objectassert/catalog_integration_iceberg_rest_details_snowflake_gen.go
+++ b/pkg/acceptance/bettertestspoc/assert/objectassert/catalog_integration_iceberg_rest_details_snowflake_gen.go
@@ -27,7 +27,7 @@ func CatalogIntegrationIcebergRestDetails(t *testing.T, id sdk.AccountObjectIden
 func CatalogIntegrationIcebergRestDetailsFromObject(t *testing.T, catalogIntegrationIcebergRestDetails *sdk.CatalogIntegrationIcebergRestDetails) *CatalogIntegrationIcebergRestDetailsAssert {
 	t.Helper()
 	return &CatalogIntegrationIcebergRestDetailsAssert{
-		assert.NewSnowflakeObjectAssertWithObject(sdk.ObjectType("CatalogIntegrationwIcebergRestDetails"), catalogIntegrationIcebergRestDetails.Id, catalogIntegrationIcebergRestDetails),
+		assert.NewSnowflakeObjectAssertWithObject(sdk.ObjectType("CatalogIntegrationIcebergRestDetails"), catalogIntegrationIcebergRestDetails.Id, catalogIntegrationIcebergRestDetails),
 	}
 }
 
--- a/pkg/sdk/catalog_integrations_ext.go
+++ b/pkg/sdk/catalog_integrations_ext.go
@@ -10,63 +10,6 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
 )
 
-type CatalogIntegrationAwsGlueDetails struct {
-	Id                     AccountObjectIdentifier
-	CatalogSource          CatalogIntegrationCatalogSourceType
-	TableFormat            CatalogIntegrationTableFormat
-	Enabled                bool
-	RefreshIntervalSeconds int
-	Comment                string
-	GlueAwsRoleArn         string
-	GlueCatalogId          string
-	GlueRegion             string
-	CatalogNamespace       string
-}
-
-type CatalogIntegrationObjectStorageDetails struct {
-	Id                     AccountObjectIdentifier
-	CatalogSource          CatalogIntegrationCatalogSourceType
-	TableFormat            CatalogIntegrationTableFormat
-	Enabled                bool
-	RefreshIntervalSeconds int
-	Comment                string
-}
-
-type CatalogIntegrationOpenCatalogDetails struct {
-	Id                     AccountObjectIdentifier
-	CatalogSource          CatalogIntegrationCatalogSourceType
-	TableFormat            CatalogIntegrationTableFormat
-	Enabled                bool
-	RefreshIntervalSeconds int
-	Comment                string
-	CatalogNamespace       string
-	RestConfig             OpenCatalogRestConfig
-	RestAuthentication     OAuthRestAuthentication
-}
-
-type CatalogIntegrationIcebergRestDetails struct {
-	Id                       AccountObjectIdentifier
-	CatalogSource            CatalogIntegrationCatalogSourceType
-	TableFormat              CatalogIntegrationTableFormat
-	Enabled                  bool
-	RefreshIntervalSeconds   int
-	Comment                  string
-	CatalogNamespace         string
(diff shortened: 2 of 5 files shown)


pkg/sdk/catalog_integrations_ext.go
```diff
@@ -4,140 +4,288 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
 )
 
+type CatalogIntegrationAwsGlueDetails struct {
```
sfc-gh-kwasilewski: I could define these structs in the .def file and generate them, but I wouldn't be able to use enums

pkg/sdk/parsers_test.go
```diff
@@ -107,7 +107,8 @@ func TestParseCommaSeparatedStringArray(t *testing.T) {
 }
 
 func TestParseOuterCommaSeparatedStringArray(t *testing.T) {
-	testCases := append(commaSeparatedStringArrayTestCases, []struct {
+	testCases := commaSeparatedStringArrayTestCases
```
sfc-gh-kwasilewski, line 110: This silences a warning reported by go-critic:

> appendAssign: append result not assigned to the same slice (gocritic)

pkg/sdk/testint/catalog_integrations_integration_test.go
```diff
@@ -497,11 +538,12 @@ func TestInt_CatalogIntegrations(t *testing.T) {
 	t.Run("describe catalog integration: AWS Glue", func(t *testing.T) {
```
sfc-gh-kwasilewski, line 538: describe tests now become almost identical to create tests, which also use describe. For now, I've kept them, but let me know what you think.

pkg/sdk/catalog_integrations_ext.go
sfc-gh-jmichalak: I think this can be done with `Field("FieldName", "EnumType")`. But we can leave this IMO. Although, we have a few repeated fields, like Id, CatalogSource, TableFormat, Enabled, RefreshIntervalSeconds, Comment.

pkg/acceptance/bettertestspoc/assert/objectassert/catalog_integration_iceberg_rest_details_snowflake_gen.go
```diff
@@ -0,0 +1,144 @@
...
+		assert.NewSnowflakeObjectAssertWithTestClientObjectProvider(sdk.ObjectType("CatalogIntegrationIcebergRestDetails"), id, func(testClient *helpers.TestClient) assert.ObjectProvider[sdk.CatalogIntegrationIcebergRestDetails, sdk.AccountObjectIdentifier] {
+			return testClient.CatalogIntegration.DescribeIcebergRest
+		}),
+	}
+}
+
+func CatalogIntegrationIcebergRestDetailsFromObject(t *testing.T, catalogIntegrationIcebergRestDetails *sdk.CatalogIntegrationIcebergRestDetails) *CatalogIntegrationIcebergRestDetailsAssert {
+	t.Helper()
+	return &CatalogIntegrationIcebergRestDetailsAssert{
+		assert.NewSnowflakeObjectAssertWithObject(sdk.ObjectType("CatalogIntegrationwIcebergRestDetails"), catalogIntegrationIcebergRestDetails.Id, catalogIntegrationIcebergRestDetails),
```
sfc-gh-jmichalak:
```suggestion
		assert.NewSnowflakeObjectAssertWithObject(sdk.ObjectType("CatalogIntegrationIcebergRestDetails"), catalogIntegrationIcebergRestDetails.Id, catalogIntegrationIcebergRestDetails),
```

pkg/sdk/catalog_integrations_ext.go
```diff
@@ -4,140 +4,288 @@ import (
...
+type CatalogIntegrationIcebergRestDetails struct {
+	Id                       AccountObjectIdentifier
+	CatalogSource            CatalogIntegrationCatalogSourceType
+	TableFormat              CatalogIntegrationTableFormat
+	Enabled                  bool
+	RefreshIntervalSeconds   int
+	Comment                  string
+	CatalogNamespace         string
+	RestConfig               IcebergRestRestConfig
+	OAuthRestAuthentication  *OAuthRestAuthentication
```
sfc-gh-jmichalak: Nit: this works well here, but in principle we should use dedicated objects for SHOW, not reusign the request builders.

pkg/sdk/testint/catalog_integrations_integration_test.go
sfc-gh-jmichalak: Yeah, that's what usually happens, we can leave it.

pkg/sdk/catalog_integrations_ext.go
sfc-gh-kwasilewski: > Although, we have a few repeated fields

That's true. I could extract a base struct and embed it, but our assertion generator doesn't handle it well enough (it creates only a single assertion)

pkg/sdk/catalog_integrations_ext.go
sfc-gh-kwasilewski: I'll keep it for now and decide later, while working on other resources.

pkg/acceptance/bettertestspoc/assert/objectassert/catalog_integration_iceberg_rest_details_snowflake_gen.go
sfc-gh-kwasilewski: Fixed

sfc-gh-kwasilewski:
Two tests failed, but they don't seem to be related to my change: 
- TestAcc_OauthIntegrationForCustomClients_BasicUseCase
- TestAcc_OauthIntegrationForCustomClients_CompleteUseCase

pkg/sdk/generator/defs/catalog_integrations_def.go
```diff
@@ -209,4 +210,52 @@ var catalogIntegrationsDef = g.NewInterface(
...
+		g.PlainStruct("CatalogIntegrationIcebergRestDetails").
+			AccountObjectIdentifier().
+			Field("CatalogSource", g.KindOfT[sdkcommons.CatalogIntegrationCatalogSourceType]()).
+			Field("TableFormat", g.KindOfT[sdkcommons.CatalogIntegrationTableFormat]()).
+			Bool("Enabled").
+			Number("RefreshIntervalSeconds").
+			Text("Comment").
+			Text("CatalogNamespace").
+			Field("RestConfig", g.KindOfT[sdk.IcebergRestRestConfig]()).
+			OptionalField("OAuthRestAuthentication", g.KindOfT[sdk.OAuthRestAuthentication]()).
```
sfc-gh-asawicki, line 251: nit: other places use `Field` and `g.KindOfTPointer[sdkcommons.AccountObjectIdentifier]()`

Issue #4524, opened 2026-03-11
MCP Server Grants cause infinite updates


landiinii:
### Terraform CLI Version

1.14.6

### Terraform Provider Version

2.14.0

### Terraform Configuration

```terraform
resource "snowflake_grant_privileges_to_database_role" "read_future_mcp_in_database" {
  privileges         = ["USAGE"]
  database_role_name = snowflake_database_role.database_role.fully_qualified_name
  on_schema_object {
    future {
      object_type_plural = "MCP Servers"
      in_database        = var.database_name
    }
  }
}
```

### Category

category:grants

### Object type(s)

resource:grant_privileges_to_database_role

### Expected Behavior

@sfc-gh-jmichalak added "MCP Server" types to grant objects a few weeks ago in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4478 (much thanks). We implemented them pretty quickly for a project that happened to line up soon after. We expected to be able to grant USAGE permissions on the object and for it to apply once, then for the terraform state to track that it has already been applied and not touch it after.

### Actual Behavior

After the initial creation of the grant, subsequent terraform applies infinitely recreated the same grants on the same database roles. The terraform state was not properly recognizing the existing grants.

We noticed that the output that Snowflake provides in the "SHOW FUTURE GRANTS" response that the provider uses to track state changes does not match the object type identifier that is defined inside of this provider. Where the provider expects to see "MCP_SERVER" snowflake returns "CORTEX_AGENT_SERVER". So the check for the grant returns empty, and the grant reapplies.

This is also the case for "Models" and "Volumes" types as well which had to be mapped in https://github.com/snowflakedb/terraform-provider-snowflake/pull/3070 and https://github.com/snowflakedb/terraform-provider-snowflake/pull/2538 respectively. I believe the correction here (if it is not decided to just change the output of show grants) is to also add a mapping for the MCP Server -> Cortex Agent Server shift.

### Steps to Reproduce

Create a grant resource as shown in the configuration
Apply it
Apply it again

### How much impact is this issue causing?

Low

### Additional Information

These blocks on lines 242 and 257 of `pkg/sdk/grants.go` resolved this issue for me on local:
```
if row.GrantedOn == "CORTEX_AGENT_SERVER" {
    grantedOn = ObjectTypeMcpServer
}
...
if row.GrantOn == "CORTEX_AGENT_SERVER" {
    grantOn = ObjectTypeMcpServer
}
```

[...]

PR #4532, opened 2026-03-12
fix: Adjust mcp server object type in grants

sfc-gh-jcieslak:
## Changes
- follow-up for #4529 but for MCP SERVER object type

References: #4524


pkg/sdk/grants.go
```diff
@@ -242,6 +242,9 @@ func (row grantRow) convert() (*Grant, error) {
 	if row.GrantedOn == "CORTEX_AGENT" {
 		grantedOn = ObjectTypeAgent
 	}
+	if row.GrantedOn == "CORTEX_AGENT_SERVER" {
```
sfc-gh-jmichalak, line 245: Nit: at this point, we can have a map :D

sfc-gh-jcieslak: True, will adjust it in a followup

sfc-gh-kwasilewski (approved):

PR #4536, opened 2026-03-12
feat: Make storage_aws_external_id optional on external volume resource

sfc-gh-jmichalak:
## Summary

- Change `storage_aws_external_id` in `storage_location` from a `Computed` (read-only) field to `Optional`, allowing users to explicitly configure it for S3/S3GOV storage locations.
- Add a state upgrader (v2.14.0) that clears previously Snowflake-populated `storage_aws_external_id` values so existing state does not drift after the schema change.
- Refactor `Read` to preserve user-configured `storage_aws_external_id` across refreshes (since DESCRIBE does not return it) and detect external changes via `describe_output`.
- Add validation that `storage_aws_external_id` is rejected for non-S3 providers (GCS, Azure, S3COMPAT).
- Update acceptance tests: basic/complete S3 use cases cover the new field; validation tests cover unsupported provider combinations; state migration test verifies the upgrader clears the old computed value without drift.
- Update migration guide and docs.

### DIFF
--- a/docs/resources/external_volume.md
+++ b/docs/resources/external_volume.md
@@ -148,15 +148,12 @@ Optional:
 - `encryption_kms_key_id` (String) Specifies the ID for the KMS-managed key used to encrypt files.
 - `encryption_type` (String) Specifies the encryption type used.
 - `storage_aws_access_point_arn` (String) Specifies the access point ARN for the S3 bucket containing your data files. Only applicable for S3 and S3GOV storage providers.
+- `storage_aws_external_id` (String) External ID that Snowflake uses to establish a trust relationship with AWS.
 - `storage_aws_key_id` (String) Specifies the AWS key ID for the S3-compatible storage location. Only applicable for S3COMPAT storage provider.
 - `storage_aws_role_arn` (String) Specifies the case-sensitive Amazon Resource Name (ARN) of the AWS identity and access management (IAM) role that grants privileges on the S3 bucket containing your data files.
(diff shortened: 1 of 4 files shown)


MIGRATION_GUIDE.md
```diff
@@ -86,7 +86,11 @@ New fields in `storage_location`:
 - `storage_aws_key_id` - AWS key ID for S3COMPAT storage locations.
 - `storage_aws_secret_key` - AWS secret key for S3COMPAT storage locations (sensitive).
 
-No changes to existing configurations are required. If you are using S3-compatible storage, you can now set `storage_provider = "S3COMPAT"` with the new fields above.
+#### *(breaking change)* `storage_aws_external_id` changed from computed to optional
+
+Previously, `storage_aws_external_id` in `storage_location` was a computed (read-only) field populated by Snowflake. It is now an optional user-configurable field. A state upgrader clears the previously computed value automatically, so no changes to existing configurations are required and `terraform plan` will show no drift after upgrading.
+
+If you previously referenced `storage_location.*.storage_aws_external_id` (e.g. in `output` blocks or `local` values), note that it will now be empty unless you explicitly set it. The Snowflake-generated external ID remains accessible via `describe_output.0.storage_locations.*.s3_storage_location.0.storage_aws_external_id`.
```
sfc-gh-jcieslak, line 93: Can't we / Shouldn't we try to make the `storage_aws_external_id` Optional, but also fill it up with Snowflake computed ID? Parameter fields in some resources are already like this

sfc-gh-jmichalak: I think it works for parameters because we can check the parameter level. Here, the problem occurs when you leave an empty config, fill the value from Snowflake, which causes a diff.

PR #4539, opened 2026-03-13
feat: add snowflake_catalog_integration_aws_glue resource

sfc-gh-kwasilewski:
Added new snowflake_catalog_integration_aws_glue resource.


examples/resources/snowflake_catalog_integration_aws_glue/import.sh
```diff
@@ -0,0 +1 @@
+terraform import snowflake_catalog_integration_aws_glue.example name
```
sfc-gh-jmichalak: Nit: this is case-sensitive, so we usually note this as:

```bash
terraform import snowflake_catalog_integration_aws_glue.example '"<name>"'
```

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/catalog_integration_aws_glue_details_show_output_gen.go
```diff
@@ -0,0 +1,144 @@
+// Code generated by resource show output assertions generator (v0.1.0); DO NOT EDIT.
```
sfc-gh-jmichalak: Nit: this file should be named `catalog_integration_aws_glue_details_output_gen.go` - without `show`.

pkg/resources/catalog_integration_aws_glue.go
```diff
@@ -0,0 +1,240 @@
...
+		Type:     schema.TypeBool,
+		Required: true,
+		Description: joinWithSpace("Specifies whether the catalog integration is available for use for Iceberg tables.",
+			"`true` allows users to create new Iceberg tables that reference this integration. Existing Iceberg tables that reference this integration function normally.",
+			"`false` prevents users from creating new Iceberg tables that reference this integration. Existing Iceberg tables that reference this integration cannot access the catalog in the table definition."),
+	},
+	"refresh_interval_seconds": {
+		Type:         schema.TypeInt,
+		Optional:     true,
+		Default:      30,
```
sfc-gh-jmichalak: We should not use defaults - read more [here](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/v1-preparations/CHANGES_BEFORE_V1.md).

pkg/resources/catalog_integration_aws_glue.go
```diff
@@ -0,0 +1,240 @@
...
+		Description:  "Specifies the number of seconds to wait between attempts to poll the external Iceberg catalog for metadata updates for automated refresh.",
+		ValidateFunc: validation.All(validation.IntAtLeast(30), validation.IntAtMost(86400)),
```
sfc-gh-jmichalak: Please don't use validators with concrete values, as they don't apply to all accounts. For ints, we can just validate that the number is positive/non-negative.

pkg/resources/catalog_integration_aws_glue.go
```diff
@@ -0,0 +1,240 @@
...
+		ValidateFunc: validation.StringIsNotEmpty,
+	},
+	"glue_region": {
+		Type:     schema.TypeString,
+		Optional: true,
+		ForceNew: true,
+		Description: joinWithSpace("Specifies the AWS region of your AWS Glue Data Catalog.",
+			"You must specify a value for this attribute if your Snowflake account is not hosted on AWS.",
+			"Otherwise, the default region is the Snowflake deployment region for the account."),
+		DiffSuppressFunc: IgnoreAfterCreation,
```
sfc-gh-jmichalak: Q: why this diff suppress? This way we won't detect external changes, and the user won't be able to change it.

pkg/resources/catalog_integration_aws_glue.go
```diff
@@ -0,0 +1,240 @@
...
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.CatalogIntegrationAwsGlueResource), TrackingDeleteWrapper(resources.CatalogIntegrationAwsGlue, deleteFunc)),
+		Description:   "Resource used to manage AWS Glue catalog integration objects. For more information, check [catalog integration documentation](https://docs.snowflake.com/en/sql-reference/sql/create-catalog-integration-glue).",
+
+		Schema: catalogIntegrationAwsGlueSchema,
+		Importer: &schema.ResourceImporter{
+			StateContext: TrackingImportWrapper(resources.CatalogIntegrationAwsGlue, ImportCatalogIntegrationAwsGlue),
+		},
+		Timeouts: defaultTimeouts,
+		CustomizeDiff: customdiff.All(
+			ComputedIfAnyAttributeChanged(catalogIntegrationAwsGlueSchema, ShowOutputAttributeName, "enabled", "comment"),
```
sfc-gh-jmichalak, line 119: Nit: the ForceNew attributes don't need to be here and below. Basically, this function says: "if `enabled` or `comment` is changed, the `show_output` should also show a change". But when they are ForceNew, the whole resource is recreated anyway.

pkg/resources/catalog_integration_aws_glue.go
```diff
@@ -0,0 +1,240 @@
...
+
+	details, err := client.CatalogIntegrations.DescribeAwsGlueDetails(ctx, id)
+	if err != nil {
+		return diag.FromErr(fmt.Errorf("could not describe AWS Glue catalog integration (%s), err = %w", d.Id(), err))
+	}
+
+	errs := errors.Join(
+		d.Set("name", details.Id.Name()),
+		d.Set("enabled", details.Enabled),
+		d.Set("refresh_interval_seconds", details.RefreshIntervalSeconds),
```
sfc-gh-jmichalak: I think this field should be handled in a different way:  
\- It is optional with a default value in Snowflake,  
\- Usually, we handle these fields with `withExternalChangesMarking` flag.  
\- Check `auto_suspend` in warehouses for reference. Note: setStateToValuesFromConfig is not needed after all, I believe.  
\- This way, we can decouple the default value in the schema definition.

pkg/testacc/resource_catalog_integration_aws_glue_acceptance_test.go
```diff
@@ -0,0 +1,437 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_CatalogIntegrationAwsGlue_BasicUseCase(t *testing.T) {
+	id := testClient().Ids.RandomAccountObjectIdentifier()
+
+	providerModel := providermodel.SnowflakeProvider().
```
sfc-gh-jmichalak: This shouldn't be needed. You can set `SF_TF_ACC_TEST_ENABLE_ALL_PREVIEW_FEATURES=true`. A custom provider model is used only for migration tests.

pkg/testacc/resource_catalog_integration_aws_glue_acceptance_test.go
```diff
@@ -0,0 +1,437 @@
...
+
+	alteredProperties := []assert.TestCheckFuncProvider{
+		resourceassert.CatalogIntegrationAwsGlueResource(t, ref).
+			HasName(id.Name()).
+			HasEnabledString(r.BooleanTrue).
+			HasComment(newComment).
+			HasRefreshIntervalSeconds(newRefreshIntervalSeconds).
+			HasGlueAwsRoleArn(glueAwsRoleArn).
+			HasGlueCatalogId(glueCatalogId).
+			HasGlueCatalogId(glueCatalogId).
```
sfc-gh-jmichalak: Nit: duplicated line.

pkg/testacc/resource_catalog_integration_aws_glue_acceptance_test.go
```diff
@@ -0,0 +1,437 @@
...
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionCreate),
+					},
+				},
+				Config: config.FromModels(t, providerModel, catalogIntegrationAwsGlueAllAttributes),
+				Check:  assertThat(t, completeAssertions...),
+			},
+			// Change props externally
+			{
+				PreConfig: func() {
+					alterRequest := sdk.NewAlterCatalogIntegrationRequest(id).WithSet(*sdk.NewCatalogIntegrationSetRequest().
```
sfc-gh-jmichalak, line 366: Here, we assert only 3 fields. We should test detecting external changes on all fields. You can just call create or replace with different values.

pkg/testacc/resource_catalog_integration_aws_glue_acceptance_test.go
```diff
@@ -0,0 +1,437 @@
...
+
+	comment := random.Comment()
+	newComment := random.Comment()
+	externalComment := random.Comment()
+
+	refreshIntervalSeconds := random.IntRange(30, 86400)
+	newRefreshIntervalSeconds := random.IntRange(30, 86400)
+	externalRefreshIntervalSeconds := random.IntRange(30, 86400)
+
+	glueRegion := "us-east-1"
```
sfc-gh-jmichalak, line 45: Q: can we test changed glue region?

sfc-gh-jmichalak (changes requested):

pkg/resources/catalog_integration_aws_glue.go
sfc-gh-kwasilewski: Because the default value can vary depending on the current region. I changed it to `IgnoreChangeToCurrentSnowflakeValueInDescribe`

pkg/resources/catalog_integration_aws_glue.go
sfc-gh-jmichalak: If only positive numbers are valid, we can skip the -1 default. This is only needed for fields when 0 is a valid value, and we need to detect if it's unset or set to 0.

pkg/testacc/resource_catalog_integration_aws_glue_acceptance_test.go
```diff
@@ -0,0 +1,538 @@
...
+func TestAcc_CatalogIntegrationAwsGlue_Validations(t *testing.T) {
+	id := testClient().Ids.RandomAccountObjectIdentifier()
+
+	glueAwsRoleArn := "arn:aws:iam::123456789012:role/sqsAccess"
+	glueCatalogId := random.NumericN(15)
+
+	catalogIntegrationAwsGlueRefreshIntervalTooSmall := model.CatalogIntegrationAwsGlue("t", id.Name(), false, glueAwsRoleArn, glueCatalogId).
+		WithRefreshIntervalSeconds(0)
+
+	catalogIntegrationAwsGlueRefreshIntervalTooBig := model.CatalogIntegrationAwsGlue("t", id.Name(), false, glueAwsRoleArn, glueCatalogId).
```
sfc-gh-jmichalak: Nit: this one is now not needed.

pkg/resources/catalog_integration_aws_glue.go
```diff
@@ -0,0 +1,261 @@
...
+	awsGlueRequest := sdk.NewAwsGlueParamsRequest(glueAwsRoleArn, glueCatalogId)
+	errs := errors.Join(
+		stringAttributeCreateBuilder(d, "comment", request.WithComment),
+		stringAttributeCreateBuilder(d, "glue_region", awsGlueRequest.WithGlueRegion),
+		stringAttributeCreateBuilder(d, "catalog_namespace", awsGlueRequest.WithCatalogNamespace),
+	)
+	if errs != nil {
+		return diag.FromErr(errs)
+	}
+	if v := d.Get("refresh_interval_seconds").(int); v != IntDefault {
```
sfc-gh-jmichalak: Nit: we have `intAttributeWithSpecialDefaultCreateBuilder` for this.

sfc-gh-jmichalak (dismissed):

pkg/resources/catalog_integration_aws_glue.go
sfc-gh-kwasilewski: This will remove `Default: fallback to Snowflake default` note from the docs - is it fine?

sfc-gh-jmichalak: Yes, this note appears only for -1 and "default" because they may be more surprising to the users.

sfc-gh-jmichalak (dismissed):

sfc-gh-jmichalak (approved):

PR #4543, opened 2026-03-18
chore: Add RevokePrivilegesFromAccountRoleSafely for graceful revoke on missing objects

sfc-gh-jmichalak:
## Summary
- Adds `SafeRevokePrivileges` helper in `pkg/sdk/operation_helpers.go` that wraps revoke calls and suppresses `ErrObjectNotExistOrAuthorized`, allowing `terraform destroy` to succeed when grant-target objects (tables, views, schemas, databases, warehouses) or roles are already deleted.
- Exposes `RevokePrivilegesFromAccountRoleSafely` on the `Grants` interface so resource implementations can use it directly.
- Adds comprehensive integration tests covering: existing objects, never-granted privileges, missing schema objects, missing schemas, missing databases, missing account objects, missing roles, and insufficient privileges.


pkg/sdk/operation_helpers.go
```diff
@@ -144,6 +144,18 @@ func SafeShowById[T any, ID ObjectIdentifierConstraint](
 	return result, nil
 }
 
+// SafeRevokePrivileges is a helper function that wraps a revoke function
+// and handles the case where the target object or role no longer exists.
+// When the referenced object (e.g. a view, table, schema, database, warehouse) or the role itself
+// does not exist, Snowflake returns ErrObjectNotExistOrAuthorized.
+func SafeRevokePrivileges(revoke func() error) error {
+	err := revoke()
+	if errors.Is(err, ErrObjectNotExistOrAuthorized) {
```
sfc-gh-jcieslak, line 153: Can we add different object types? I think in errors_integration_test we have a list where different object types return different "missing" errors and I'm curious if that would be also the case here

sfc-gh-kwasilewski: REVOKE command might behave differently from DROP and other commands tested by errors_integration_test, but yeah, it'd be nice to test some other object types, even manually.

sfc-gh-jmichalak: Added - also for SHOW GRANTS.

sfc-gh-jcieslak (approved):

sfc-gh-kwasilewski (approved):

PR #4554, opened 2026-03-23
fix: tag association object type mapping

sfc-gh-jcieslak:
## Changes
* Follow-up for #4412, where the change involved the usage of GetOnObjectIdentifier, which, although it supported functions and procedures, didn't support the remaining possible object types, which resulted in degradation of the tag_association resource. To fix this, missing object type cases had to be added to the mapping within the GetOnObjectIdentifier function.


pkg/resources/grant_ownership.go
```diff
@@ -405,54 +405,88 @@ func ReadGrantOwnership(ctx context.Context, d *schema.ResourceData, meta any) d
 // TODO(SNOW-1229218): Make sdk.ObjectType + string objectName to sdk.ObjectIdentifier mapping available in the sdk (for all object types).
 func GetOnObjectIdentifier(objectType sdk.ObjectType, objectName string) (sdk.ObjectIdentifier, error) {
```
sfc-gh-jmichalak, line 406: Nit: this is not scoped properly, as we use it in tag associations, so this helper should be moved.

pkg/resources/grant_ownership.go
```diff
@@ -405,54 +405,88 @@ func ReadGrantOwnership(ctx context.Context, d *schema.ResourceData, meta any) d
...
+		sdk.ObjectTypeCatalogIntegration,
+		sdk.ObjectTypeComputePool,
+		sdk.ObjectTypeConnection,
 		sdk.ObjectTypeDatabase,
 		sdk.ObjectTypeExternalVolume,
 		sdk.ObjectTypeFailoverGroup,
 		sdk.ObjectTypeIntegration,
+		sdk.ObjectTypeListing,
 		sdk.ObjectTypeNetworkPolicy,
+		sdk.ObjectTypeSessionPolicy,
```
sfc-gh-jmichalak: I think this should be schema-level.

sfc-gh-jmichalak (changes requested):

pkg/resources/grant_ownership.go
sfc-gh-kwasilewski: +1, session policy is a schema-level entity.

sfc-gh-jcieslak: Hmm, I moved it because in the docs it says "Identifier for the session policy; must be unique for your account," which usually means it's an account-level object... I'll put it back

pkg/resources/grant_ownership.go
sfc-gh-jcieslak: We already have a dedicated task for taking this out, so I decided not to do it now

pkg/resources/grant_ownership.go
sfc-gh-kwasilewski: Docs are misleading, for example, [here](https://docs.snowflake.com/en/sql-reference/sql/show-session-policies) you can see that the SHOW command returns the database and schema where the given policy is placed in.

sfc-gh-kwasilewski (approved):

sfc-gh-jmichalak (approved):

PR #4555, opened 2026-03-23
feat: add snowflake_catalog_integration_open_catalog resource

sfc-gh-kwasilewski:
## Changes
- Added new snowflake_catalog_integration_open_catalog resource:
  - Show output schema has been implemented manually, as the generator doesn't handle nested attributes correctly.
  - Resource show output assertions, resource output assertions, and model have been partially generated. Everything else that couldn't be easily generated was implemented in *ext.go files.
- Introduced *Details structs for catalog integration DESCRIBE output (e.g. OpenCatalogRestConfigDetails, IcebergRestRestConfigDetails), replacing types that were previously reused to hold parsed properties.


pkg/resources/catalog_integration_open_catalog.go
```diff
@@ -0,0 +1,373 @@
...
+					},
+					"oauth_client_secret": {
+						Type:         schema.TypeString,
+						Required:     true,
+						Sensitive:    true,
+						Description:  externalChangesNotDetectedFieldDescription("Specifies the secret of the OAuth2 credential associated with your Open Catalog service connection."),
+						ValidateFunc: validation.StringIsNotEmpty,
+					},
+					"oauth_allowed_scopes": {
+						Type:        schema.TypeList,
```
sfc-gh-jmichalak, line 106: Nit: use TypeSet instead, unless the order really matters.

pkg/resources/catalog_integration_open_catalog.go
```diff
@@ -0,0 +1,373 @@
...
+	output, ok := d.GetOk(DescribeOutputAttributeName + ".0.rest_config")
+	if !ok {
+		return nil
+	}
+	outputList := output.([]any)
+	result := outputList[0].(map[string]any)
+
+	catalogApiType := string(details.RestConfig.CatalogApiType)
+	accessDelegationMode := string(details.RestConfig.AccessDelegationMode)
+	if result["catalog_api_type"] == catalogApiType && result["access_delegation_mode"] == accessDelegationMode {
```
sfc-gh-jmichalak: What if these values match, but e.g. catalog_uri is different? We won't detect such an external change, similarly below. BTW, we already have helpers in `show_and_describe_handlers.go` - see ReadExternalS3StageFunc as an example.

pkg/testacc/resource_catalog_integration_open_catalog_acceptance_test.go
```diff
@@ -0,0 +1,794 @@
...
+			HasOauthAllowedScopes(oAuthAllowedScope),
+	}
+
+	basicAssertionsWithRefreshIntervalZero := append(
+		[]assert.TestCheckFuncProvider{
+			resourceassert.CatalogIntegrationOpenCatalogResource(t, ref).
+				HasName(id.Name()).
+				HasEnabledString(r.BooleanFalse).
+				HasCommentEmpty().
+				HasRefreshIntervalSeconds(0).
```
sfc-gh-jmichalak, line 167: Why do we have a refresh interval of 0 here, but in the show output, it's 30?

pkg/testacc/resource_catalog_integration_open_catalog_acceptance_test.go
```diff
@@ -0,0 +1,794 @@
...
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionCreate),
+					},
+				},
+				Config: config.FromModels(t, allAttributes),
+				Check:  assertThat(t, completeAssertions...),
+			},
+			// Import
+			{
```
sfc-gh-jmichalak, line 495: Nit: this step is not needed - we test it in step 2.

pkg/testacc/resource_catalog_integration_open_catalog_acceptance_test.go
```diff
@@ -0,0 +1,794 @@
...
+					testClient().CatalogIntegration.CreateFunc(t, createRequest)
+				},
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionDestroyBeforeCreate),
+						planchecks.ExpectDrift(ref, "rest_config.0.catalog_uri", sdk.String(privateCatalogUri), sdk.String(catalogUri)),
+						planchecks.ExpectChange(ref, "rest_config.0.catalog_uri", tfjson.ActionDelete, sdk.String(catalogUri), sdk.String(privateCatalogUri)),
+						planchecks.ExpectDrift(ref, "rest_config.0.catalog_name", sdk.String(newCatalogName), sdk.String(catalogName)),
+						planchecks.ExpectChange(ref, "rest_config.0.catalog_name", tfjson.ActionDelete, sdk.String(catalogName), sdk.String(newCatalogName)),
+						planchecks.ExpectDrift(ref, "rest_config.0.catalog_api_type", sdk.String(string(sdk.CatalogIntegrationCatalogApiTypePrivate)), sdk.String(string(sdk.CatalogIntegrationCatalogApiTypePublic))),
```
sfc-gh-jmichalak, line 584: In relation to my comment before, this test doesn't detect it because catalog_api_type is changed, so d.Set is called anyway. Similar for the rest_authentication steps below.

sfc-gh-jmichalak (dismissed):
Looks good overall. Let's address the limitations of external change detection here or in the next PR.

pkg/testacc/resource_catalog_integration_open_catalog_acceptance_test.go
sfc-gh-kwasilewski: It's 30 in the describe output because that's the Snowflake default. And here we have 0 because the key exists in the schema (we did import few steps earlier), but doesn't exist in the configuration (we unset everything in this step), so SDK returns the default value for the type.

We have the same assertion in acc tests for previous catalog integration resources as well.

pkg/resources/catalog_integration_open_catalog.go
sfc-gh-kwasilewski: I will keep TypeList because Snowflake allows duplicates, and because of that, it will be easier to manage this attribute as a list.

pkg/testacc/resource_catalog_integration_open_catalog_acceptance_test.go
sfc-gh-kwasilewski: It was supposed to be an import step with a complete model. I'll fix it.

pkg/resources/catalog_integration_open_catalog.go
sfc-gh-kwasilewski: Fixed. I couldn't use a helper for rest_authentication because it contains a field that is a list, and `reflect.DeepEqual` returns false due to different types (`[]string` vs `[]interface{}`)

pkg/testacc/resource_catalog_integration_open_catalog_acceptance_test.go
sfc-gh-kwasilewski: That's true, but I'd rather not add separate test steps for each of these fields because it would make the tests much longer (instead of 2 steps, we'd have 7). Is it acceptable?

pkg/testacc/resource_catalog_integration_open_catalog_acceptance_test.go
sfc-gh-jmichalak: Ah yes, this comment was supposed to be deleted.

pkg/testacc/resource_catalog_integration_open_catalog_acceptance_test.go
sfc-gh-jmichalak: Yes! In such cases, we still group them to keep the tests reasonably short. I wrote this to show that this case in particular is tricky, as it's easy to miss.

pkg/resources/catalog_integration_open_catalog.go
```diff
@@ -0,0 +1,374 @@
...
+	oAuthTokenUri := details.RestAuthentication.OauthTokenUri
+	if result["oauth_token_uri"] == oAuthTokenUri &&
+		result["oauth_client_id"] == details.RestAuthentication.OauthClientId &&
+		areSlicesEqual(details.RestAuthentication.OauthAllowedScopes, result["oauth_allowed_scopes"].([]any)) {
+		return nil
+	}
+
+	return d.Set("rest_authentication", []any{map[string]any{
+		"oauth_token_uri":      oAuthTokenUri,
+		"oauth_client_id":      details.RestAuthentication.OauthClientId,
```
sfc-gh-jmichalak, line 360: Nit: we can skip oauth_client_secret here because all other fields are ForceNew which causes the recreation anyway, let's comment this.

pkg/resources/catalog_integration_open_catalog.go
sfc-gh-jmichalak: Yeah, this is annoying. In other resources we convert the whole objects to have the same types, but this approach looks fine as well.

pkg/resources/catalog_integration_open_catalog.go
```diff
@@ -359,15 +349,26 @@ func handleExternalChangesToRestAuthentication(d *schema.ResourceData, details *
...
-		"oauth_client_secret":  oAuthClientSecret,
 		"oauth_allowed_scopes": details.RestAuthentication.OauthAllowedScopes,
 	}})
 }
+
+func areSlicesEqual(a []string, b []any) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	bs := make([]string, len(b))
```
sfc-gh-kwasilewski, line 369: I've just noticed I could use `expandStringList`. I can do it in a follow-up PR.

sfc-gh-jmichalak (approved):

pkg/resources/catalog_integration_open_catalog.go
sfc-gh-kwasilewski: I already removed oauth_client_secret from here

PR #4556, opened 2026-03-23
chore: Adaptive warehouse SDK outputs

sfc-gh-jmichalak:
## Changes

- Add `MaxStatementSize` and `MaxBurstRateCredits` fields to SDK `Warehouse` struct and `warehouseDBRow`
- Split `ShowWarehouseSchema` into common + regular parts; expose `ShowRegularWarehouseSchema` for resource and data source
- Reorder `CreateAdaptiveWarehouseOptions` so `Tag` appears before object params (fixes SQL generation order)

## TODO
- Adjust the SDK after changes in Snowflake.
- Implement adaptive warehouse resource
- Verify that adaptive warehouses are supported in the warehouses data source - update schemas and conversions.

### DIFF
--- a/pkg/schemas/warehouse_gen.go
+++ b/pkg/schemas/warehouse_gen.go
@@ -3,12 +3,15 @@
 package schemas
 
 import (
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-// ShowWarehouseSchema represents output of SHOW query for the single Warehouse.
-var ShowWarehouseSchema = map[string]*schema.Schema{
+// Adjusted manually: split into common, regular, and adaptive schemas.
+
+// showWarehouseSchemaCommon contains fields present for all warehouse types.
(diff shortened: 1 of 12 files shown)


pkg/schemas/warehouse_gen.go
```diff
@@ -3,12 +3,15 @@
...
 
 import (
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-// ShowWarehouseSchema represents output of SHOW query for the single Warehouse.
-var ShowWarehouseSchema = map[string]*schema.Schema{
+// Adjusted manually: split into common, regular, and adaptive schemas.
```
sfc-gh-kwasilewski, line 11: [nit] Maybe it should no longer be a `_gen` file at this point. These changes aren't simple deletions or replacements.

sfc-gh-kwasilewski (approved):

pkg/schemas/warehouse_gen.go
sfc-gh-jmichalak: True. Also, we can support such splits in the generator. I'll add this to the readme.

PR #4551, opened 2026-03-20
feat: add missing authentication policy features

sfc-gh-jcieslak:
## Changes
- Add `client_policy` to sdk and resource
- Add `OTP` option to `allowed_methods`
- Add `require_role_restriction_for_service_users` to `pat_policy`
- Add integration and acceptance tests

## Follow-up
- Make parsers for DESC in Read not fail on invalid output

## References
- #4552

PR #4577, opened 2026-03-27
chore: Adjust authentication policy describe parameter parsing

sfc-gh-jcieslak:
Follow-up for #4551:
- Add optional describe property parsing (don't error out on missing property; instead, return a warning and proceed with the read operation)


pkg/resources/authentication_policy.go
```diff
@@ -447,38 +446,81 @@ func ReadContextAuthenticationPolicy(withExternalChangesMarking bool) schema.Rea
...
 		authenticationPolicyDescriptions := sdk.AuthenticationPolicyDetails(authenticationPolicyDescriptionsRaw)
+		diags := make(diag.Diagnostics, 0)
+
 		if withExternalChangesMarking {
+			describeMappings := make([]outputMapping, 0)
+
 			authenticationMethods, err := authenticationPolicyDescriptions.GetAuthenticationMethods()
 			if err != nil {
-				return diag.FromErr(err)
+				diags = append(diags, diag.Diagnostic{
```
sfc-gh-jmichalak, line 458: 1. It would be nice to extract it somehow and use only helpers here. We can even make the summary more descriptive: that a specific row is missing. The eariler error was very vague.
2. Should we also do this for all other resources parsing DESC rows? If such changes can be done outside of BCR, I'm in favor of it.

We can do both in follow-up(s).

sfc-gh-jmichalak (dismissed):

sfc-gh-asawicki (approved):

PR #4580, opened 2026-03-30
chore: Configure generation part per object

sfc-gh-asawicki:
Our generator framework allows filtering by generation parts globally across all objects generated in the given run. The ability to regenerate only with the subset of parts, settable per object, should make it easier to introduce changes when reworking/extending generators.

Per-object resolution logic:
1. If object implements `HasObjectGenerationSettings` with non-empty `AllowedGenerationParts` → intersect those with `globalParts`.
2. Otherwise → return `globalParts` unchanged (backward-compatible default; falls back to all parts when no CLI filter is set).

Example Usage:
```go
DatabaseRoles = g.NewInterface("DatabaseRoles", "DatabaseRole", "DatabaseObjectIdentifier").
    WithAllowedGenerationParts("default", "dto", "dto_builders", "validations").
    CreateOperation(...).
    AlterOperation(...)
```

Iterative Adoption:
- Existing generators (objectassert, objectparametersassert, etc.) require zero changes.
- SDK generator `Interface` objects without `WithAllowedGenerationParts` keep their current behavior.
- New per-object settings can be added one object at a time as needed.


pkg/sdk/generator/gen/interface.go
```diff
@@ -42,6 +42,17 @@ type Interface struct {
 	IdentifierKind string
 
 	*genhelpers.PreambleModel
+	*genhelpers.ObjectGenerationSettings
+}
+
+// WithAllowedGenerationParts restricts this object to only the specified generation parts.
+// Parts not listed here will be skipped during generation, even if enabled globally.
+func (i *Interface) WithAllowedGenerationParts(parts ...string) *Interface {
```
sfc-gh-jcieslak, line 50: Could we make "parts" into an enum? Would be easier to choose from rather than searching docs for available values

sfc-gh-jcieslak (approved):

pkg/sdk/generator/gen/interface.go
sfc-gh-asawicki: It would have to be an enum per generator, but it's fine as the SDK generator is currently the only one using generator parts besides the default one. I'll adjust in one of the next PRs.

sfc-gh-jmichalak (approved):

PR #4581, opened 2026-03-30
feat: Add GRANTS_SAFE_DESTROY experiment for safe revoke in grant privileges

sfc-gh-jmichalak:
## Summary

- Introduces a new `GRANTS_SAFE_DESTROY` experimental feature flag that enables safe revoke behavior during Terraform destroy of `snowflake_grant_privileges_to_account_role` resources
- When enabled, `DeleteGrantPrivilegesToAccountRole` uses a safe revoke path that silently ignores `ErrObjectNotExistOrAuthorized` errors, preventing failures when the grant target (table, view, schema, etc.) has already been dropped before the grant resource
- Adds migration guide entry and docs index update explaining the new experiment

Ref. https://github.com/snowflakedb/terraform-provider-snowflake/issues/3412


docs/index.md
```diff
@@ -123,7 +123,7 @@ provider "snowflake" {
...
+- `experimental_features_enabled` (Set of String) A list of experimental features. Similarly to preview features, they are not yet stable features of the provider. Enabling given experiment is still considered a preview feature, even when applied to the stable resource. These switches offer experiments altering the provider behavior. If the given experiment is successful, it can be considered an addition in the future provider versions. This field can not be set with environmental variables. Check more details in the [experimental features section](#experimental-features). Active experiments are: `WAREHOUSE_SHOW_IMPROVED_PERFORMANCE` | `GRANTS_STRICT_PRIVILEGE_MANAGEMENT` | `PARAMETERS_IGNORE_VALUE_CHANGES_IF_NOT_ON_OBJECT_LEVEL` | `PARAMETERS_REDUCED_OUTPUT` | `USER_ENABLE_DEFAULT_WORKLOAD_IDENTITY` | `GRANTS_IMPORT_VALIDATION` | `TAGS_ALLOW_EMPTY_ALLOWED_VALUES` | `IMPORT_BOOLEAN_DEFAULT` | `SAFE_DESTROY`.
```
sfc-gh-asawicki: nit: safe destroying is already available in all the non-grant, non-attachment resources. Let's make it clear that it touches only subset. Let's maybe consider making the name a bit more explicit like SAFE_DESTROY_GRANT_AND_ATTACHMENTS, or, to be more consistent with other experiments: GRANTS_SAFE_DESTROY and ATTACHMENTS_SAFE_DESTROY (we can have two similar experiments, why not).

sfc-gh-asawicki (dismissed):

docs/index.md
sfc-gh-jmichalak: Done.

sfc-gh-asawicki (approved):

PR #4582, opened 2026-03-31
feat: Add RevokePrivilegeFromShareSafely to SDK Grants interface

sfc-gh-jmichalak:
## Summary

- Adds `RevokePrivilegeFromShareSafely` to the `Grants` interface and implementation, following the same pattern as `RevokePrivilegesFromAccountRoleSafely` (PR #4543)
- The safe variant wraps `RevokePrivilegeFromShare` with `SafeRevokePrivileges`, which swallows `ErrObjectNotExistOrAuthorized` — the error Snowflake returns when the share or grant target object no longer exists
- Adds integration tests covering non-existing shares, databases, schemas, tables, and views (including fully non-existing hierarchy)

### DIFF
--- a/pkg/sdk/testint/safe_grant_handlers_integration_test.go
+++ b/pkg/sdk/testint/safe_grant_handlers_integration_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestInt_SafeRevokePrivilegesFromAccountRole(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/sdk/testint/safe_grant_handlers_integration_test.go
```diff
@@ -331,3 +332,77 @@ func TestInt_ShowFutureGrantsInNonExistingContainer(t *testing.T) {
...
+
+	t.Run("non-existing database", func(t *testing.T) {
+		err := client.Grants.RevokePrivilegeFromShareSafely(ctx, []sdk.ObjectPrivilege{sdk.ObjectPrivilegeUsage}, &sdk.ShareGrantOn{
+			Database: NonExistingAccountObjectIdentifier,
+		}, share.ID())
+		require.NoError(t, err)
+	})
+}
+
+func TestInt_SafeRevokeFromShareOnNonExistingSchemaLevelObjects(t *testing.T) {
```
sfc-gh-kwasilewski, line 372: [nit] You could merge this test function with the previous one so that you don't create a share twice. Plus, you could add `"non-existing database"` test case to the parameterized set.

sfc-gh-kwasilewski (approved):

pkg/sdk/testint/safe_grant_handlers_integration_test.go
sfc-gh-jmichalak: I'll do this in the next PR.

PR #4586, opened 2026-03-31
feat: Add missing tag features

sfc-gh-jcieslak:
## Changes
- Adjust tags SDK
  - Add propagate and on_conflict options and adjust validation for those fields
  - Adjust the request builder style closer to the auto-generated one
  - Add unit and integration tests for new fields

## TODO
- Make the propagation method required in alter statements
- Handle output for standard Snowflake accounts correctly
- Migration guide
- Resource
  - Follow-up regarding `on_conflict = ALLOWED_VALUES_SEQUENCE` behavior
  - Make sure any changes regarding propagation (e.g. in allowed_values) are followed up by an alter tag with propagation properties

PR #4587, opened 2026-03-31
feat: Add missing tag features resource follow up

sfc-gh-jcieslak:
Follow up for #4586 (use new SDK behavior in tag resource).

## Changes
- Add new propagation fields to tag resource
- Add acceptance tests that validate field and propagation behavior
- Regenerate show output schema, models, assertions (plus add some custom extensions)
- Turn propagation into a request and make the propagation method required in the SDK
- Add migration guide
- Make sure any changes regarding propagation (e.g., in allowed_values) are followed up by an alter tag with propagation properties (validated by acceptance tests that check the actual propagated value to confirm the resource properly refreshes the propagation state).

## TODO
- Follow-up regarding `on_conflict = ALLOWED_VALUES_SEQUENCE` behavior
- Run the tests and, if necessary, adjust tag propagation for standard Snowflake accounts
- Address follow-ups from https://github.com/snowflakedb/terraform-provider-snowflake/pull/4586
- Extend resource documentation about propagation fields


pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -991,3 +1000,460 @@ func TestAcc_Tag_NoAllowedValues_DisableExperimentFlag(t *testing.T) {
 		},
 	})
 }
+
+func TestAcc_Tag_PropagateAndOnConflict(t *testing.T) {
```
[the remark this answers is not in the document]
sfc-gh-jcieslak: Can be replaced with already existing tests

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -991,3 +1000,460 @@ func TestAcc_Tag_NoAllowedValues_DisableExperimentFlag(t *testing.T) {
...
+					resourceassert.TagResource(t, withCustomConflict.ResourceReference()).
+						HasPropagateEnum(sdk.TagPropagationOnDependency).
+						HasOnConflictCustomValue("FAIL"),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_Tag_PropagateWithAllowedValuesSequence(t *testing.T) {
```
sfc-gh-jcieslak: Should be used with assertions on propagated tag

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -991,3 +1000,460 @@ func TestAcc_Tag_NoAllowedValues_DisableExperimentFlag(t *testing.T) {
...
+						HasName(id.Name()).
+						HasPropagateEnum(sdk.TagPropagationOnDependency),
+
+					resourceassert.TagResource(t, withCustomConflict.ResourceReference()).
+						HasPropagateEnum(sdk.TagPropagationOnDependency).
+						HasAllowedValuesEmpty().
+						HasOnConflictCustomValue("FAIL"),
+				),
+			},
+			// Update - remove everything (propagate, on_conflict)
```
sfc-gh-jcieslak: Drop this step

docs/resources/tag.md
```diff
@@ -55,6 +55,8 @@ resource "snowflake_tag" "tag" {
...
 - `no_allowed_values` (Boolean) When set to true, the tag explicitly disallows any value from being assigned. This is different from omitting `allowed_values`, which means any value is accepted. Available only when the `TAGS_ALLOW_EMPTY_ALLOWED_VALUES` experiment is enabled. Conflicts with `allowed_values`.
+- `on_conflict` (Block List, Max: 1) Specifies what happens when there is a conflict between the values of [propagated tags](https://docs.snowflake.com/en/user-guide/object-tagging/propagation). External changes for this field won't be detected. In case you want to apply external changes, you can re-create the resource manually using "terraform taint". (see [below for nested schema](#nestedblock--on_conflict))
+- `propagate` (String) Specifies that the tag will be automatically propagated from source objects to target objects. See more about tag propagation in the [official documentation](https://docs.snowflake.com/en/user-guide/object-tagging/propagation). Valid options are: `NONE` | `ON_DEPENDENCY` | `ON_DATA_MOVEMENT` | `ON_DEPENDENCY_AND_DATA_MOVEMENT`
```
sfc-gh-jcieslak, line 73: One thing that we can limit is allowing to set `NONE` as it's not documented (discussing with tag team). Removing the param (`ALTER TAG UNSET PROPAGATE`) effectively should be the same as setting it to none (`ALTER TAG SET PROPAGATE = NONE`), which is the default after unset (or if not set during create).

pkg/resources/tag.go
```diff
@@ -60,6 +62,48 @@ var tagSchema = map[string]*schema.Schema{
...
+	},
+	"on_conflict": {
+		Type:         schema.TypeList,
+		Optional:     true,
+		Description:  externalChangesNotDetectedFieldDescription("Specifies what happens when there is a conflict between the values of [propagated tags](https://docs.snowflake.com/en/user-guide/object-tagging/propagation)."),
+		MaxItems:     1,
+		RequiredWith: []string{"propagate"},
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"allowed_values_sequence": {
```
sfc-gh-asawicki, line 93: important: we can't release this resource without addressing the follow-up of dealing with the allowed values ordering (currently not enforced, as we use set instead of list). I am not sure if we shouldn't keep this PR off dev branch, until it's discussed and implemented (we could add another PR on top of this one).

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -991,3 +1002,225 @@ func TestAcc_Tag_NoAllowedValues_DisableExperimentFlag(t *testing.T) {
...
+	tagId := testClient().Ids.RandomSchemaObjectIdentifier()
+
+	var viewId sdk.SchemaObjectIdentifier
+
+	withSeqInitial := model.TagBase("test", tagId).
+		WithPropagateEnum(sdk.TagPropagationOnDependency).
+		WithAllowedValues("confidential", "internal", "public").
+		WithOnConflictAllowedValuesSequence()
+
+	withSeqReordered := model.TagBase("test", tagId).
```
sfc-gh-asawicki, line 1019: WithAllowedValues is a set, so how does it currently work (i.e. the ordering is NOT changed).

sfc-gh-asawicki (changes requested):

pkg/sdk/tags_validations.go
```diff
@@ -54,8 +54,8 @@ func (v *TagPropagate) validate() error {
 		return errors.Join(ErrNilOptions)
 	}
 	var errs []error
-	if !valueSet(v.PropagationMethod) && !valueSet(v.OnConflict) {
-		errs = append(errs, errAtLeastOneOf("TagPropagate", "PropagationMethod", "OnConflict"))
+	if !valueSet(v.PropagationMethod) {
```
sfc-gh-jmichalak: Nit: this is now required, so do we still need this validation?

pkg/resources/tag.go
sfc-gh-jcieslak: Yes, we shouldn't merge it, let's just make sure we agree with those changes, and the follow-up will be pointing to this branch. After merging the follow-up, we will merge this PR.

pkg/resources/tag.go
```diff
@@ -60,6 +62,48 @@ var tagSchema = map[string]*schema.Schema{
 		Description:   "When set to true, the tag explicitly disallows any value from being assigned. This is different from omitting `allowed_values`, which means any value is accepted. Available only when the `TAGS_ALLOW_EMPTY_ALLOWED_VALUES` experiment is enabled. Conflicts with `allowed_values`.",
 		ConflictsWith: []string{"allowed_values"},
 	},
+	"propagate": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: fmt.Sprintf("Specifies that the tag will be automatically propagated from source objects to target objects. See more about tag propagation in the [official documentation](https://docs.snowflake.com/en/user-guide/object-tagging/propagation). Valid options are: %s", docs.PossibleValuesListed(sdk.AllTagPropagationValues)),
+		DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
sfc-gh-jmichalak: Why do we need this? How is it different from handling object default values?

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: The test is currently skipped as it won't work due to the fact you mentioned. In the follow-up, after addressing the allowed_values order, the test should pass. Here, I just prepared the test case I think should be possible from the user's pov.

pkg/resources/tag.go
```diff
@@ -392,6 +467,58 @@ func UpdateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
...
+		// If on_conflict is configured with allowed_values_sequence and allowed_values changed,
+		// re-send propagation to re-evaluate conflict resolution on dependent objects.
+		if v, ok := d.GetOk("on_conflict.0.allowed_values_sequence"); ok && v.(bool) {
+			shouldPropagate = true
+		}
+	}
+
+	if shouldPropagate {
+		if v, ok := d.GetOk("propagate"); ok {
+			tagPropagation, err := sdk.ToTagPropagation(v.(string))
```
sfc-gh-jmichalak: Nit: reuse from Create

sfc-gh-jmichalak (changes requested):
\+1 to other comments.

pkg/resources/tag.go
```diff
@@ -310,6 +377,14 @@ func UpdateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
 		}
 	}
 
+	if d.HasChange("on_conflict") {
+		if len(d.Get("on_conflict").([]any)) == 0 {
+			if err := client.Tags.Alter(ctx, sdk.NewAlterTagRequest(id).WithUnset(*sdk.NewTagUnsetRequest().WithOnConflict(true))); err != nil {
```
sfc-gh-jmichalak, line 393: Nit: this one should be "merged" with the handling below to reduce queries.

MIGRATION_GUIDE.md
```diff
@@ -26,6 +26,21 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 
 ## v2.14.x ➞ v2.15.0
 
+### *(new feature)* snowflake_tag resource: propagation and conflict resolution support
+
+We added support for [tag propagation](https://docs.snowflake.com/en/user-guide/object-tagging/propagation) to the `snowflake_tag` resource. The following new fields are now available:
+
+- `propagate` - Specifies the propagation method for the tag. Valid values are: `ON_DEPENDENCY`, `ON_DATA_MOVEMENT`, `ON_DEPENDENCY_AND_DATA_MOVEMENT`.
```
sfc-gh-jmichalak: Nit: `NONE` is missing here, but it's present in the resource documentation.

sfc-gh-jmichalak (changes requested):

pkg/sdk/tags_validations.go
sfc-gh-jcieslak: I don't think we do, removed

pkg/resources/tag.go
sfc-gh-jcieslak: We shouldn't, it was added when I thought Snowflake sometimes returns empty string in propagated, but i was bad output mapping on my part. Forgot, about removing this part after the fix.

pkg/resources/tag.go
sfc-gh-jcieslak: I wouldn't touch it as the order of sets and unsets matters, e.g., this unset is specifically needed here before allowed values because there are scenarios this unset has to go first, and after the allowed values "block", there are sets and unsets that can only happen afterward. Merging could make certain scenarios break. From the docs: https://docs.snowflake.com/en/sql-reference/sql/alter-tag, there's not much we could merge anyway. As the logic is brittle, let's leave it as it is.

MIGRATION_GUIDE.md
sfc-gh-jcieslak: True, added

pkg/resources/tag.go
sfc-gh-jcieslak: Done

pkg/resources/tag.go
sfc-gh-jcieslak: Nvm, it's needed as None == null (removed from config). Adjusted a bit, should be cleaner.

sfc-gh-asawicki (approved):

PR #4588, opened 2026-03-31
fix: Remove file_format from ComputedIfAnyAttributeChanged on stage resources

sfc-gh-jmichalak:
## Summary

- Fixes `describe_output` permadiff on `snowflake_stage_external_s3` (and all other stage resources) when `file_format.0.format_name` uses unquoted identifier notation.
- Removes `"file_format"` from the `ComputedIfAnyAttributeChanged` calls for `describe_output` on all five stage resources. The `ComputedIfAnyAttributeChanged` function cannot respect `DiffSuppressFunc` on nested attributes (documented at `custom_diffs.go:119`), so `diff.HasChange("file_format")` returns `true` when `format_name` differs only in quoting, eagerly marking `describe_output` as `(known after apply)`.
- Adds a regression test that reproduces the bug on v2.14.1 and verifies a clean plan on the current version.

Ref. https://github.com/snowflakedb/terraform-provider-snowflake/issues/4514

### DIFF
--- a/pkg/testacc/resource_external_s3_stage_acceptance_test.go
+++ b/pkg/testacc/resource_external_s3_stage_acceptance_test.go
@@ -967,6 +967,7 @@ func TestAcc_ExternalS3Stage_DescribeOutputPermadiff(t *testing.T) {
 			{
 				ExternalProviders:  ExternalProviderWithExactVersion("2.14.1"),
 				Config:             accconfig.FromModels(t, providerModel, stageModelQuotedFileFormatId),
+				PlanOnly:           true,
 				ExpectNonEmptyPlan: true,
 			},
 			// Step 3: Upgrade to current version — plan must be empty (proves the fix).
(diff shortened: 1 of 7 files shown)


sfc-gh-kwasilewski (dismissed):

sfc-gh-kwasilewski (approved):

pkg/testacc/resource_external_s3_stage_acceptance_test.go
```diff
@@ -920,3 +922,60 @@ func TestAcc_ExternalS3Stage_Validations(t *testing.T) {
...
 	})
 }
+
+// TestAcc_ExternalS3Stage_DescribeOutputPermadiff verifies that the describe_output
+// permadiff caused by identifier quoting on file_format.0.format_name is fixed.
+// In v2.14.1, ComputedIfAnyAttributeChanged was triggered on file_format changes,
+// causing describe_output to be marked
+// as (known after apply) on every plan when a nested field had changes which were suppressed.
+// See: https://github.com/snowflakedb/terraform-provider-snowflake/issues/4514
+func TestAcc_ExternalS3Stage_DescribeOutputPermadiff(t *testing.T) {
```
sfc-gh-asawicki, line 932: nit: let's document (if it's undocumented) that we could implement the recursive-traversal computedIf... func to try all diffsuppresses

sfc-gh-asawicki (approved):

pkg/testacc/resource_external_s3_stage_acceptance_test.go
sfc-gh-jmichalak: SNOW-3322958

PR #4590, opened 2026-04-02
chore: Dereference optional fields in show output schema generator

sfc-gh-jmichalak:
## Summary

Fix `to_schema_mapper.tmpl` to dereference pointer fields before passing them to mappers.
Previously, the template generated `string(warehouse.Size)` for `*WarehouseSize` (invalid) and
`schemaMap["running"] = warehouse.Running` storing a raw `*int` instead of `int`.

The fix wraps pointer accesses in `(*field)` so the generated code is e.g.
`string((*warehouse.Size))` and `(*warehouse.Running)`.

Addresses: #4508 (review comment [r3009756221](https://github.com/snowflakedb/terraform-provider-snowflake/pull/4508#discussion_r3009756221))

## Regenerated files

- `warehouse_gen.go` — removed all `// Adjusted manually` nil-check annotations; added missing `max_statement_size` and `max_burst_rate_credits` fields
- `resource_monitor_gen.go` — fixed `SuspendAt`/`SuspendImmediateAt` stored as raw `*int` pointers
- `account_gen.go` — removed `// Manually modified, please don't re-generate` guard; fixed pointer fields stored without dereference
- `organization_account_gen.go` — removed 8 `// adjusted manually` annotations; now fully auto-generated

### DIFF
--- a/pkg/schemas/account_gen.go
+++ b/pkg/schemas/account_gen.go
@@ -135,79 +135,78 @@ func AccountToSchema(account *sdk.Account) map[string]any {
 	accountSchema["account_name"] = account.AccountName
 	accountSchema["snowflake_region"] = account.SnowflakeRegion
 	if account.RegionGroup != nil {
-		accountSchema["region_group"] = account.RegionGroup
+		accountSchema["region_group"] = (*account.RegionGroup)
 	}
 	if account.Edition != nil {
-		// Manually modified, please don't re-generate
-		accountSchema["edition"] = string(*account.Edition)
+		accountSchema["edition"] = string((*account.Edition))
 	}
 	if account.AccountURL != nil {
-		accountSchema["account_url"] = account.AccountURL
+		accountSchema["account_url"] = (*account.AccountURL)
 	}
 	if account.CreatedOn != nil {
-		accountSchema["created_on"] = account.CreatedOn.String()
+		accountSchema["created_on"] = (*account.CreatedOn).String()
 	}
 	if account.Comment != nil {
-		accountSchema["comment"] = account.Comment
+		accountSchema["comment"] = (*account.Comment)
 	}
 	accountSchema["account_locator"] = account.AccountLocator
 	if account.AccountLocatorUrl != nil {
-		accountSchema["account_locator_url"] = account.AccountLocatorUrl
+		accountSchema["account_locator_url"] = (*account.AccountLocatorUrl)
 	}
 	if account.ManagedAccounts != nil {
-		accountSchema["managed_accounts"] = account.ManagedAccounts
+		accountSchema["managed_accounts"] = (*account.ManagedAccounts)
 	}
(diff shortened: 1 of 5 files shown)


pkg/schemas/account_gen.go
```diff
@@ -135,79 +135,78 @@ func AccountToSchema(account *sdk.Account) map[string]any {
 	accountSchema["account_name"] = account.AccountName
 	accountSchema["snowflake_region"] = account.SnowflakeRegion
 	if account.RegionGroup != nil {
-		accountSchema["region_group"] = account.RegionGroup
+		accountSchema["region_group"] = (*account.RegionGroup)
```
sfc-gh-jcieslak, line 138: Is this additional set of parentheses needed?

sfc-gh-jcieslak (approved):

pkg/schemas/account_gen.go
sfc-gh-jmichalak: No, it's for simplicity. It's needed when calling additional methods.

pkg/schemas/gen/templates/to_schema_mapper.tmpl
```diff
@@ -7,7 +7,7 @@ func {{ .Name }}ToSchema({{ $nameLowerCase }} *{{ .SdkType }}) map[string]any {
 {{- range .SchemaFields }}
     {{ if .IsOriginalTypePointer -}}
         if {{ $nameLowerCase }}.{{ .OriginalName }} != nil {
-        {{ $schemaName }}["{{ .Name }}"] = {{ RunMapper .Mapper $nameLowerCase "." .OriginalName }}
+        {{ $schemaName }}["{{ .Name }}"] = {{ RunMapper .Mapper "(*" $nameLowerCase "." .OriginalName ")" }}
```
sfc-gh-asawicki, line 10: There is a logic in `schema_field_mapper.go` that was supposed to handle the case with enums. It's using the underlying type to check if this is an enum (without ptr). Because of that, it uses cast no matter what, where it should either cast (non-null) or dereference and cast (null). It's also handling ints incorrectly (if it is a pointer, that dereference mapper should be used instead).
This solution works, but I think we should target the incorrect logic in schema_field_mapper (so that this logic is in one place).

sfc-gh-asawicki (approved):

pkg/schemas/gen/templates/to_schema_mapper.tmpl
sfc-gh-jmichalak: Yes, we may do this when reworking MapToSchemaField. This solution does not add any logic, but improves the "formatting". I think that other features in MapToSchemaField should be quite simple; maybe I'll work on them right after.

sfc-gh-asawicki: The formatting logic is achieved currently through these mappers, so now it's in two places instead of one.

PR #4591, opened 2026-04-02
chore: Rework pre-push-check command

sfc-gh-kwasilewski:
The reworked pre-push-check command runs the same set of validations as pre-push, but without modifying any files.
The Documentation And Code Formatting GitHub workflow now runs `make pre-push-check`, so CI enforces the same checks as local pre-push-check.

## Changes
- Add fmt-check and terraform-fmt-check commands
- In mod-check, use -diff flag that causes tidy not to modify go.mod or go.sum but instead print the necessary changes as a diff
- After failed git diff in docs and generate* checks, restore paths and exit with the original status

## TODO
- Skip tests if docs-and-fmt workflow fails
- Add a pre-commit configuration

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,14 +26,21 @@ docs: generate-docs-additional-files ## generate docs
 	tools/bin/tfplugindocs generate --provider-name=terraform-provider-snowflake
 
 docs-check: docs ## check that docs have been generated
-	git diff --exit-code -- docs
+	git diff --exit-code -- docs || ( status=$$?; git restore -- docs; exit "$$status" )
 
 fmt: terraform-fmt ## Run terraform fmt and gofumpt
 	tools/bin/gofumpt -l -w .
 
+fmt-check: terraform-fmt-check ## Run terraform fmt and gofumpt without modifying any files
+	tools/bin/gofumpt -d .
+
 terraform-fmt: ## Run terraform fmt
 	terraform fmt -recursive ./examples/
 	terraform fmt -recursive ./pkg/testacc/testdata/
+
+terraform-fmt-check: ## check if all Terraform configuration files are correctly formatted
+	terraform fmt -check -diff -recursive ./examples/
+	terraform fmt -check -diff -recursive ./pkg/testacc/testdata/
 
 help:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,15 +26,22 @@ docs: generate-docs-additional-files ## generate docs
...
 
+fmt-check: terraform-fmt-check ## Run terraform fmt and gofumpt without modifying any files
+	tools/bin/gofumpt -d .
+
 terraform-fmt: ## Run terraform fmt
 	terraform fmt -recursive ./examples/
 	terraform fmt -recursive ./pkg/testacc/testdata/
 
+terraform-fmt-check: ## check if all Terraform configuration files are correctly formatted
+	terraform fmt -check -diff -recursive ./examples/
```
sfc-gh-asawicki, line 46: nit: I had to browse the docs https://developer.hashicorp.com/terraform/cli/commands/fmt and https://go.dev/ref/mod#go-mod-tidy to verify that non-zero exit status is returned in the fmt and mod checks. I would consider leaving short comments like: `-check causes non-zero exit status to be returned when the input is not properly formatted (source: https://developer.hashicorp.com/terraform/cli/commands/fmt)`, so that it's clearer to someone unfamiliar with the syntax.

Makefile
```diff
@@ -147,7 +154,7 @@ generate-docs-additional-files: ## generate docs additional files
 	go run ./pkg/internal/tools/doc-gen-helper/ $$PWD
 
 generate-docs-additional-files-check: generate-docs-additional-files ## check that docs additional files have been generated
-	git diff --exit-code -- examples/additional
+	git diff --exit-code -- examples/additional || ( status=$$?; git restore -- examples/additional; exit "$$status" )
```
sfc-gh-asawicki: nit: Two thoughts here:
1. Do we need to restore the git? This is a separate pipeline, if it's not cleaned, it shouldnt matter, unless it affets other commands.
2. If we need to clean it, this part is pretty repetitive, we could probably extract it as a separate rule.

sfc-gh-asawicki (dismissed):

Makefile
sfc-gh-kwasilewski: Yeah, it doesn't matter from the GitHub workflow's perspective. I wanted to prevent unexpected behavior for any developer (or agent) who uses this command locally, as "check" usually implies that no files are modified, and that would happen without this cleanup.

sfc-gh-asawicki (approved):

PR #4594, opened 2026-04-02
feat: Add snowflake_warehouse_adaptive resource

sfc-gh-jmichalak:
## Changes

Introduces the `snowflake_warehouse_adaptive` resource for managing Snowflake adaptive warehouses.

## TODO
- Add missing columns and acceptance tests to the warehouses data source.
- Support detecting external type changes.


docs/resources/warehouse_adaptive.md
```diff
@@ -0,0 +1,139 @@
...
+## Schema
+
+### Required
+
+- `name` (String) Identifier for the adaptive warehouse; must be unique for your account. Due to technical limitations (read more [here](../guides/identifiers_rework_design_decisions#known-limitations-and-identifier-recommendations)), avoid using the following characters: `|`, `.`, `"`.
+
+### Optional
+
+- `comment` (String) Specifies a comment for the adaptive warehouse.
+- `max_query_performance_level` (String) Specifies the maximum query performance level for the adaptive warehouse. Determines the initial compute capacity. Valid values are (case-insensitive): `XSMALL` | `SMALL` | `MEDIUM` | `LARGE` | `XLARGE` | `XXLARGE` | `XXXLARGE` | `X4LARGE`.
```
sfc-gh-asawicki, line 49: question: do we want to explicitly list the synonyms? (we do it for the warehouse sizes)

pkg/datasources/warehouses.go
```diff
@@ -114,7 +114,7 @@ func ReadWarehouses(ctx context.Context, d *schema.ResourceData, meta any) diag.
 		}
 
 		flattenedWarehouses[i] = map[string]any{
-			resources.ShowOutputAttributeName:     []map[string]any{schemas.WarehouseToSchema(&warehouse)},
+			resources.ShowOutputAttributeName:     []map[string]any{schemas.RegularWarehouseToSchema(&warehouse)},
```
sfc-gh-asawicki: nit: I think that a more standard distinction would be leaving it as `WarehouseToSchema` and naming the new one `AdaptiveWarehouseToSchema` (AFAIK Regular is not a domain name, we only use it in this showoutput schema context)

pkg/resources/warehouse_adaptive.go
```diff
@@ -0,0 +1,302 @@
...
+			return client.Warehouses.DropSafely
+		},
+	)
+
+	return &schema.Resource{
+		CreateContext: PreviewFeatureCreateContextWrapper(string(previewfeatures.WarehouseAdaptiveResource), TrackingCreateWrapper(resources.WarehouseAdaptive, CreateWarehouseAdaptive)),
+		ReadContext:   PreviewFeatureReadContextWrapper(string(previewfeatures.WarehouseAdaptiveResource), TrackingReadWrapper(resources.WarehouseAdaptive, ReadWarehouseAdaptiveFunc(true))),
+		UpdateContext: PreviewFeatureUpdateContextWrapper(string(previewfeatures.WarehouseAdaptiveResource), TrackingUpdateWrapper(resources.WarehouseAdaptive, UpdateWarehouseAdaptive)),
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.WarehouseAdaptiveResource), TrackingDeleteWrapper(resources.WarehouseAdaptive, deleteFunc)),
+		Description:   "Resource used to manage adaptive warehouse objects. Adaptive warehouses automatically scale compute resources based on workload. For more information, check [adaptive warehouse documentation](https://docs.snowflake.com/en/LIMITEDACCESS/adaptive-warehouses).",
```
sfc-gh-asawicki: nit: the documentation reference will most probably change soon. Let's discuss internally with the owning team, so that we link it future-proof

pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
```diff
@@ -0,0 +1,315 @@
...
+func TestAcc_WarehouseAdaptive_BasicUseCase(t *testing.T) {
+	warehouseId := testClient().Ids.RandomAccountObjectIdentifier()
+	newWarehouseId := testClient().Ids.RandomAccountObjectIdentifier()
+	comment := random.Comment()
+	newComment := random.Comment()
+
+	warehouseModel := model.WarehouseAdaptiveWithId(warehouseId)
+	warehouseModelWithOptionals := model.WarehouseAdaptiveWithId(warehouseId).
+		WithComment(comment).
+		WithQueryThroughputMultiplier(2).
```
sfc-gh-asawicki, line 35: question: QueryThroughputMultiplier uses the special int default (-1), so I assume that setting 0 is possible. If yes, let's add a test step for that.

sfc-gh-asawicki (dismissed):

docs/resources/warehouse_adaptive.md
sfc-gh-jmichalak: Well, the warehouse sizes have an explicit synonym table: https://docs.snowflake.com/en/sql-reference/sql/create-warehouse, whether adaptive warehouses do not, so I decided not to do this. The current approach allows only the values from the docs, but we also support these "synonyms" in the converter because Snowflake returns these synonyms actually.

pkg/datasources/warehouses.go
sfc-gh-jmichalak: Yes, it's not, but it's used only as an internal name. Renamed.

pkg/resources/warehouse_adaptive.go
sfc-gh-jmichalak: That's true, I removed the link for now. We could merge it as-is and add the link later once we have it.

pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
sfc-gh-jmichalak: Done.

docs/resources/warehouse_adaptive.md
sfc-gh-asawicki: This feature is still PrPr so it is understandable that not everything is documented. It's not blocking this PR, I expect we may get asked to support the synonyms also from the config perspective later on. For now, closing.

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

PR #4596, opened 2026-04-03
chore: Migrate to generated assertions

sfc-gh-jcieslak:
## Changes
- Migrate database role and pipes integration tests, as well as external OAuth integration acceptance tests, to the custom, builder-style assertions from our bettertestspoc package.


sfc-gh-jmichalak (approved):

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
```diff
@@ -180,6 +180,10 @@ var allStructs = []SdkObjectDef{
 		IdType:       "sdk.SchemaObjectIdentifier",
 		ObjectStruct: sdk.NetworkRule{},
 	},
+	{
+		IdType:       "sdk.SchemaObjectIdentifier",
+		ObjectStruct: sdk.Pipe{},
```
sfc-gh-asawicki, line 185: I don't know if we should address the old-style tests/assertions with the preview objects that have not been reworked yet. They can dramatically change during the rework (both resource and SDK).

sfc-gh-asawicki (approved):

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
sfc-gh-jcieslak: Well, good point. Next migrations would be more focused on stable objects. The pipes were just small enough to test the migration.

PR #4601, opened 2026-04-07
fix: Functional tests for object renaming experiments

sfc-gh-jcieslak:
## Changes
- A fix for TestAcc_SdkV2Functional_ListNameUpdate - The test was skipped. Removing the skip required adding a new column "name", needed in list item comparisons/assertions. Other tests and assertions had to be adjusted.


pkg/testfunctional/test_resource_object_renaming_lists_and_sets_acceptance_test.go
```diff
@@ -454,9 +452,9 @@ func TestAcc_SdkV2Functional_ListsWithDuplicatedItems(t *testing.T) {
 				}),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					assert.ContainsExactlyInAnyOrder("snowflake_test_resource_object_renaming.test", "list", []map[string]string{
-						{"string": "111", "int": "111"},
-						{"string": "222", "int": "222"},
-						{"string": "333", "int": "333"},
+						{"name": "", "string": "111", "int": "111"},
```
sfc-gh-jmichalak: So, why are we asserting the name here with the new filtering? We could:

- (1) keep the assertion strict and the mechanical additions (current behavior could be simpler without the filter),
- (2) keep the filter and remove the "name": "" entries from tests that don't care about name.

I'm for (1) because the schema we use isn't that complex, and we avoid situations where someone forgets to assert a field. And in other such tests, like acceptance tests, we usually assert all possible fields anyway.

pkg/testfunctional/test_resource_object_renaming_lists_and_sets.go
```diff
@@ -541,7 +565,7 @@ func ignoreListOrderAfterFirstApply(parentKey string) schema.SchemaDiffSuppressF
 		// Raw state is not null after first apply
 		if !d.GetRawState().IsNull() {
 			// Parse item index from the key
-			keyParts := strings.Split(strings.TrimLeft(key, parentKey+"."), ".")
+			keyParts := strings.Split(strings.TrimPrefix(key, parentKey+"."), ".")
```
sfc-gh-jmichalak, line 568: Nit: this can be fixed in other places as well - L630 and L666.

sfc-gh-jmichalak (dismissed):

pkg/acceptance/bettertestspoc/assert/commons.go
```diff
@@ -160,20 +160,40 @@ func ContainsExactlyInAnyOrder(resourceKey string, attributePath string, expecte
 			}
 		}
 
+		// Determine which keys to compare: only those that appear in at least one expected item.
+		// This allows callers to omit schema fields they don't care about (e.g. optional fields that
+		// default to an empty string in state) without causing spurious mismatches.
+		keysToCompare := make(map[string]bool)
```
sfc-gh-asawicki: Filtering out specific keys here can be misleading for someone using this method. I would prefer us to be strict and verify the full schema. We can have a separate less restrictive method but let's make it separate if needed.

sfc-gh-asawicki (changes requested):

pkg/testfunctional/test_resource_object_renaming_lists_and_sets_acceptance_test.go
sfc-gh-asawicki: +1

pkg/testfunctional/test_resource_object_renaming_lists_and_sets.go
sfc-gh-jcieslak: Done

pkg/testfunctional/test_resource_object_renaming_lists_and_sets_acceptance_test.go
sfc-gh-jcieslak: removed the empty name assertions, and adjusted the assertion as discussed

pkg/acceptance/bettertestspoc/assert/commons.go
sfc-gh-jcieslak: Adjusted. Took out the common part, and create two assertions: ContainsAtLeastInAnyOrder and ContainsExactlyInAnyOrder

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

PR #4602, opened 2026-04-07
feat: extend session policies SDK definition

sfc-gh-kwasilewski:
## Changes
- Add missing AllowedSecondaryRoles and BlockedSecondaryRoles properties and wire them into CREATE and into ALTER SET / UNSET.
- SHOW SESSION POLICIES: add optional LIKE, IN, ON, STARTS WITH, and LIMIT.
- DESCRIBE SESSION POLICY: use property-row slice mapping (SessionPolicyProperty) instead of a single structured row.
- Reorder SHOW row fields so owner_role_type precedes options.
- SQL builder: For ddl:list structs rendered as sqlListClause, when there are no inner clauses and the field uses must_parentheses, emit () so SQL like ... = () is valid (fixes empty AllowedSecondaryRoles / BlockedSecondaryRoles lists).

## TODO
- Parse session policy properties into a structured output
- Extend integration tests

### DIFF
--- a/pkg/sdk/generator/defs/session_policies_def.go
+++ b/pkg/sdk/generator/defs/session_policies_def.go
@@ -7,9 +7,10 @@ import (
 )
 
 var sessionPolicySecondaryRoles = g.NewQueryStruct("SessionPolicySecondaryRoles").
-	OptionalSQLWithCustomFieldName("All", "'ALL'").
-	List("Roles", "AccountObjectIdentifier", g.ListOptions()).
-	WithValidation(g.ConflictingFields, "All", "Roles")
+	OptionalSQLWithCustomFieldName("All", "('ALL')").
+	OptionalSQLWithCustomFieldName("None", "()").
+	List("Roles", g.KindOfT[sdkcommons.AccountObjectIdentifier](), g.ListOptions().Parentheses()).
+	WithValidation(g.ExactlyOneValueSet, "All", "None", "Roles")
 
 var sessionPoliciesDef = g.NewInterface(
 	"SessionPolicies",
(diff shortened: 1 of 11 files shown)


pkg/sdk/generator/defs/session_policies_def.go
```diff
@@ -6,6 +6,11 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/generator/gen/sdkcommons"
 )
 
+var sessionPolicySecondaryRoles = g.NewQueryStruct("SessionPolicySecondaryRoles").
+	OptionalSQLWithCustomFieldName("All", "'ALL'").
+	List("Roles", "AccountObjectIdentifier", g.ListOptions()).
+	WithValidation(g.ConflictingFields, "All", "Roles")
```
sfc-gh-asawicki: I think they should rather be exactly one (if both are absent but AllowedSecondaryRoles or BlockedSecondaryRoles is provided). + tests for empty variant

pkg/sdk/sql_builder.go
```diff
@@ -578,6 +578,9 @@ type sqlListClause struct {
 
 func (v sqlListClause) String() string {
 	if len(v.clauses) == 0 {
+		if v.pm == MustParentheses {
```
sfc-gh-asawicki: question: does this improvement help some other objects (e.g. that were implemented differently because of this limitation)? We can follow-up, it's not blocking this chagne.

sfc-gh-asawicki (changes requested):

pkg/sdk/generator/defs/session_policies_def.go
sfc-gh-kwasilewski: exactlyOneValueSet will return an error for an empty slice (`SessionPolicySecondaryRoles{Roles: []AccountObjectIdentifier{}}`), as it considers slices as not set if len is 0: https://github.com/snowflakedb/terraform-provider-snowflake/blob/625590e092b5b5d8e961fe42da778adade83e3ab/pkg/sdk/validations.go#L84

That's not what I want because an empty slice is a valid use case.

I can add a test for empty struct (`SessionPolicySecondaryRoles{}`)

sfc-gh-kwasilewski: Nevermind, I found another solution that allows me to use exactlyOneValueSet. On top of that, sql_builder changes are no longer necessary. PTAL.

pkg/sdk/sql_builder.go
sfc-gh-kwasilewski: This has been reverted

sfc-gh-jmichalak (dismissed):

sfc-gh-asawicki (approved):

PR #4604, opened 2026-04-07
feat: Extend GRANTS_SAFE_DESTROY experiment to snowflake_grant_privileges_to_database_role

sfc-gh-jmichalak:
## Changes

Extends the existing `GRANTS_SAFE_DESTROY` experiment (introduced in #4581) to the `snowflake_grant_privileges_to_database_role` resource. Also, reuse the cached provider between the related grants tests.

### DIFF
--- a/pkg/testacc/resource_grant_privileges_safe_destroy_experimental_features_acceptance_test.go
+++ b/pkg/testacc/resource_grant_privileges_safe_destroy_experimental_features_acceptance_test.go
@@ -32,7 +32,7 @@ func TestAcc_Experimental_GrantPrivilegesToAccountRole_SafeDestroy_MissingWareho
 
 	experimentProviderModel := providermodel.SnowflakeProvider().
 		WithExperimentalFeaturesEnabled(experimentalfeatures.GrantsSafeDestroy)
-	experimentFactory := providerFactoryUsingCache("TestAcc_Experimental_GrantPrivilegesToAccountRole_SafeDestroy_MissingWarehouse")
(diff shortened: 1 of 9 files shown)


pkg/testacc/resource_grant_privileges_safe_destroy_experimental_features_acceptance_test.go
```diff
@@ -32,7 +32,7 @@ func TestAcc_Experimental_GrantPrivilegesToAccountRole_SafeDestroy_MissingWareho
 
 	experimentProviderModel := providermodel.SnowflakeProvider().
 		WithExperimentalFeaturesEnabled(experimentalfeatures.GrantsSafeDestroy)
-	experimentFactory := providerFactoryUsingCache("TestAcc_Experimental_GrantPrivilegesToAccountRole_SafeDestroy_MissingWarehouse")
+	experimentFactory := providerFactoryUsingCache("TestAcc_Experimental_GrantPrivileges_SafeDestroy")
```
sfc-gh-jcieslak, line 35: Since now the factories are the same (same key), shouldn't we just put them in `1_setup_provider_test.go` where other shared provider cache factories live?

sfc-gh-jcieslak (approved):

pkg/testacc/resource_grant_privileges_safe_destroy_experimental_features_acceptance_test.go
sfc-gh-jmichalak: They are used only in this file, but we can do it.

PR #4609, opened 2026-04-08
chore: Adjust warehouse SDK

sfc-gh-jcieslak:
## Changes
- Follow-up after BCR 2026_02 analysis
  - Adjust warehouse SDK
  - Add a new entry about URL changes to BCR_MIGRATION_GUIDE

### DIFF
--- a/pkg/sdk/testint/warehouses_integration_test.go
+++ b/pkg/sdk/testint/warehouses_integration_test.go
@@ -146,7 +146,6 @@ func TestInt_Warehouses(t *testing.T) {
 			MaxConcurrencyLevel:             sdk.Int(10),
 			StatementQueuedTimeoutInSeconds: sdk.Int(2000),
 			StatementTimeoutInSeconds:       sdk.Int(3000),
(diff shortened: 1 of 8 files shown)


pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -146,7 +146,6 @@ func TestInt_Warehouses(t *testing.T) {
 			MaxConcurrencyLevel:             sdk.Int(10),
```
sfc-gh-jmichalak, line 146: Nit: it was like this before, but & generation should be removed from the name.

sfc-gh-jmichalak (dismissed):

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-jcieslak: Addressed

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

PR #4612, opened 2026-04-08
feat: Handle allowed_values_sequence tag conflict resolution

sfc-gh-jcieslak:
## Changes
- Follow-up for #4586 and #4587
- Add a new `ordered_allowed_values` field that handles `allowed_values` in order by using TypeList
- Deprecate `allowed_values` in favor of the newly added one
- Adjust examples, resource documentation, and MIGRATION_GUIDE
- Address comments from the previous PRs
  - Manually ran the tag tests on the Standard Snowflake account to validate correct resource handling
  - Remove unnecessary validation https://github.com/snowflakedb/terraform-provider-snowflake/pull/4586#discussion_r3021651948, https://github.com/snowflakedb/terraform-provider-snowflake/pull/4587#discussion_r3027429030
  - Use StringAllowEmpty instead of AllowedValue helper struct https://github.com/snowflakedb/terraform-provider-snowflake/pull/4586#discussion_r3021519216
  - Extract helper for creating tables with AsSelect and use in a helper that creates conflicting tag values https://github.com/snowflakedb/terraform-provider-snowflake/pull/4586#discussion_r3021586992

## Follow-up
This will be a separate PR as get_tag does not contain this information, but tag_references does. We don't have it in the SDK, so adding and using it will be in the next PR, as it could bloat this PR a bit:
- Update GetTag to contain propagation level and adjust assertions to check concrete level https://github.com/snowflakedb/terraform-provider-snowflake/pull/4586#discussion_r3027011194


MIGRATION_GUIDE.md
```diff
@@ -26,15 +26,47 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
...
+resource "snowflake_tag" "example" {
+  # ...
+  ordered_allowed_values = ["production", "staging", "development"]
+}
+```
+
+After switching, run `terraform plan` — Terraform will show an update moving the values from `allowed_values` to `ordered_allowed_values`.
+The tag's allowed values in Snowflake remain unchanged (what only may be altered is the order of the values).
+
+**Import behavior:** When importing a `snowflake_tag` resource, values are always populated into the `allowed_values` field (because during import there is no configuration to indicate which field to use).
```
sfc-gh-asawicki: question: should we change it to the new field instead importing into the deprecated one?

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -1195,6 +1225,177 @@ func TestAcc_Tag_Propagation_CustomOnConflictValue(t *testing.T) {
...
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionUpdate),
+					},
+				},
+				Config: config.FromModels(t, providerModel, withOrderedModified),
+				Check: assertThat(t,
+					objectassert.Tag(t, id).HasAllowedValues("x", "b", "c"),
+					resourceassert.TagResource(t, ref).HasOrderedAllowedValues("x", "b", "c"),
+				),
+			},
+			// Switch to allowed_values (unordered).
```
sfc-gh-asawicki, line 1324: question: I understand that a different filed is used but if we keep the same values then semantically, there is no change needed. Shouldn't we suppress such a change?

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -1195,6 +1225,177 @@ func TestAcc_Tag_Propagation_CustomOnConflictValue(t *testing.T) {
...
+			{
+				Config: config.FromModels(t, providerModel, withAllowedValues),
+				Check: assertThat(t,
+					objectassert.Tag(t, id).HasAllowedValuesUnordered("a", "b", "c"),
+					resourceassert.TagResource(t, ref).
+						HasAllowedValues("a", "b", "c").
+						HasOrderedAllowedValuesEmpty(),
+				),
+			},
+			// Switch to ordered_allowed_values - triggers Update, verify Snowflake order.
```
sfc-gh-asawicki, line 1267: question: what if the Snowflake ordering is already good? Shouldn't we suppress such a change?

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -1195,6 +1225,177 @@ func TestAcc_Tag_Propagation_CustomOnConflictValue(t *testing.T) {
...
+				Config: config.FromModels(t, providerModel, withNoAllowedValues),
+				Check: assertThat(t,
+					objectassert.Tag(t, id).HasAllowedValuesEmpty(),
+					resourceassert.TagResource(t, ref).
+						HasNoAllowedValuesString("true").
+						HasAllowedValuesEmpty().
+						HasOrderedAllowedValuesEmpty(),
+				),
+			},
+			// Switch from no_allowed_values to allowed_values.
```
sfc-gh-asawicki: question: isnt it covered with some previously existing tests when this logic was added?

sfc-gh-asawicki (changes requested):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: I left it for backward compatibility, as it would change the import behavior. Maybe let's leave it like this and, depending on demand, adjust it or put it behind a feature flag?

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: Should be, I'll remove `no_allowed_values` <-> `allowed_values` transitions

MIGRATION_GUIDE.md
```diff
@@ -26,15 +26,47 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
...
 - `propagate` - Specifies the propagation method for the tag. Valid values are: `ON_DEPENDENCY`, `ON_DATA_MOVEMENT`, `ON_DEPENDENCY_AND_DATA_MOVEMENT`.
 - `on_conflict` - Configures how conflicting tag values from multiple source objects are resolved during propagation. Requires `propagate` to be set. Supports two mutually exclusive options:
-  - `on_conflict.0.allowed_values_sequence` - Resolves conflicts using the order defined in the tag's `allowed_values`.
+  - `on_conflict.0.allowed_values_sequence` - Resolves conflicts using the order defined in the tag's `ordered_allowed_values`. Requires `ordered_allowed_values` to be set.
   - `on_conflict.0.custom_value` - Resolves conflicts by using a custom string value.
 
+#### New `ordered_allowed_values` field
+
+A new `ordered_allowed_values` field (TypeList) has been added to the `snowflake_tag` resource.
+It replaces the existing `allowed_values` field (TypeSet) and preserves the order you specify — which is required when using `on_conflict.allowed_values_sequence` for tag propagation conflict resolution,
```
sfc-gh-jmichalak: Nit: it doesn't replace it, but this field is preferred.

MIGRATION_GUIDE.md
sfc-gh-jmichalak: \+1 for importing into the new field. If someone's importing the resource, the state is already empty, so we should steer towards using the new field.

pkg/resources/tag.go
```diff
@@ -385,84 +411,80 @@ func UpdateContextTag(ctx context.Context, d *schema.ResourceData, meta any) dia
...
-			if err := errors.Join(
-				client.Tags.Alter(ctx, sdk.NewAlterTagRequest(id).WithAdd([]string{tempTagAllowedValue})),
-				client.Tags.Alter(ctx, sdk.NewAlterTagRequest(id).WithDrop([]string{tempTagAllowedValue})),
-			); err != nil {
-				return diag.FromErr(err)
+			} else {
+				// Without experiment: removing values = blocking state (legacy behavior).
+				// DROP old values; if already blocking (no old values), do nothing.
+				oldUnorderedRaw, _ := d.GetChange("allowed_values")
+				oldValues := expandStringListAllowEmpty(oldUnorderedRaw.(*schema.Set).List())
```
sfc-gh-jmichalak, line 478: question: why AllowEmpty only here, and not anywhere else?

docs/resources/tag.md
```diff
@@ -7,9 +7,9 @@ description: |-
...
 
-~> **Current limitations** Recently, the tags propagation was introduced (check [2025-05-14-tag-propagation](https://docs.snowflake.com/en/release-notes/2025/other/2025-05-14-tag-propagation)). This resource is not currently supporting the `ON_CONFLICT` attribute and the tag allowed values ordering. If needed, use the [`snowflake_execute`](./execute) for the time-being. This limitation will be addressed in the next versions of the provider.
+~> **Deprecation notice** The `allowed_values` field (unordered set) is deprecated and will be removed in the next major version. Use `ordered_allowed_values` (ordered list) instead. `ordered_allowed_values` preserves the order you specify, which is required when using `on_conflict.allowed_values_sequence` for [tag propagation](https://docs.snowflake.com/en/user-guide/object-tagging/propagation) conflict resolution. See the [migration guide](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/MIGRATION_GUIDE.md#new-ordered_allowed_values-field) for details.
```
sfc-gh-jmichalak, line 10: Nit: please report the import limitation.

sfc-gh-jmichalak (approved):
\+1 for other comms

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: I think it's possible, but is it worth suppressing the change?
For the field switch, the user would be expecting to see it in the plan.
Making it show in the plan later could be more confusing (e.g. the user is already using new field for some time, and some value changes; the user would expect only changes planned on the new field, but because of this diff suppress the user, as far as I understand, would also see changes on `allowed_values` because its removal was previously suppressed so it stayed in the state for this time).

pkg/resources/tag.go
sfc-gh-jcieslak: Good point, empty values are allowed, so changing to `expandStringListAllowEmpty`

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: same as below: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4612#discussion_r3057512479

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Adjusted implementation, tests, and docs for the new import target field for allowed values

docs/resources/tag.md
sfc-gh-jcieslak: As there's no import limitation after the change (it goes to the new field), I don't think we should add anything else

pkg/testacc/resource_tag_acceptance_test.go
sfc-gh-jcieslak: As discussed, will be tested and addressed in a follow-up

pkg/testacc/resource_tag_acceptance_test.go
```diff
@@ -1195,6 +1228,146 @@ func TestAcc_Tag_Propagation_CustomOnConflictValue(t *testing.T) {
...
+				Config:            config.FromModels(t, providerModel, withOrderedModified),
+				ResourceName:      ref,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			// Apply after import - values already in ordered_allowed_values, no reconciliation needed.
+			{
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionNoop),
```
sfc-gh-jmichalak, line 1350: Nit: I guess this is now not needed, since we know the plan is empty after the previous step.

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

Issue #3869, opened 2025-07-20
[Bug]: issue on removing the tag_association on a table Column


Tharun-K-Git:
### Terraform CLI Version

1.12.2

### Terraform Provider Version

2.0.0

### Terraform Configuration

```terraform
resource "snowflake_tag_association" "tag_association" {
  object_identifiers = var.object_identifiers
  object_type        = var.object_type
  tag_id             = var.tag_id
  tag_value          = var.tag_value
  skip_validation    = var.skip_validation
}

module "column_tag_associations_new" {
  for_each = {
    for assoc in var.column_tag_associations_new :
    "${join("_", assoc.object_identifiers)}_${assoc.tag_id}_${assoc.tag_value}" => assoc
  }
  source = "../modules/snowflake_tag_association"

  object_identifiers = each.value.object_identifiers
  object_type        = each.value.object_type
  tag_id             = each.value.tag_id
  tag_value          = each.value.tag_value
  skip_validation    = lookup(each.value, "skip_validation", true)
}
```

### Category

category:stable

### Expected Behavior

Expected to remove the tag on the column, after removing the entry, but getting failed with the error below

`SQL compilation error:
Object does not exist, or operation cannot be performed.

Error: [errors.go:22] object does not exist, or operation cannot be performed
module.column_tag_associations_new["dev.sch.TEST_TABLE_KT_3.Age_dev.sch.TEST_TABLE_KT_3.Department_dev.sch.TEST_TABLE_KT_3.JoinDate_dev.sch.TEST_TABLE_KT_3.tedt_db.COMMON.TAG_Test_BO"].snowflake_tag_association.tag_association: Creation complete after 1s [id="tedt_db"."COMMON"."TAG_Test"|BO|COLUMN]`

When checked the logs, it is attempting to query like `SHOW TABLES LIKE 'TEST_TABLE_KT_3 IN SCHEMA db.sch`

> Does the removing tag on column need access to the target table and schema?

Removing the tag on table works totally fine. Only issue with 

### Actual Behavior

Should successfully remove the tag on column. Currently working totally fine for the table level.

### Steps to Reproduce

1. Associate tag on table column
2. run terraform apply
3. remove the tag on column
4. run terraform apply

### How much impact is this issue causing?

High

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jcieslak:
Hey @Tharun-K-Git 👋 
[...]

PR #4616, opened 2026-04-08
feat: Add TAG_ASSOCIATION_SAFE_DESTROY experiment for safe tag unset

sfc-gh-jmichalak:
## Summary

- Adds a new `TAG_ASSOCIATION_SAFE_DESTROY` experimental feature for `snowflake_tag_association`
- When enabled, destroy operations silently succeed when the tagged object (or its parent hierarchy) no longer exists, instead of failing with `does not exist or not authorized`
- Ref #3869: destroying a column-level tag association when the parent table/schema was dropped externally now succeeds (previously `skipColumnIfDoesNotExist` propagated `ErrDoesNotExistOrOperationCannotBePerformed`)

Mirrors the pattern introduced in #4581 (`GRANTS_SAFE_DESTROY`).


pkg/resources/tag_association.go
```diff
@@ -328,7 +334,7 @@ func skipColumnIfDoesNotExist(ctx context.Context, client *sdk.Client, id sdk.Ob
 		return false, errors.New("invalid column identifier")
```
sfc-gh-jmichalak, line 350: <p class='graphite__hidden'><i>[Re: line +321]</i></p>
Nit: I think we also have to use UnsetSafely here. The reason is that an object can be removed after Read (though unlikely), so here, if the column is still in the state, we could call UnsetSafely and ignore skipColumnIfDoesNotExist when the experiment is enabled.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/snowflakedb/terraform-provider-snowflake/4616?utm_source=unchanged-line-comment">Graphite</a>.</p>

sfc-gh-kwasilewski: I'm worried about ignoring skipColumnIfDoesNotExist. If a tag is associated with two columns, and only one column is removed after Read (unlikely, but possible), we will call UnsetSafely for the remaining column, and that will break (ALTER COLUMN doesn't support IF EXISTS).

sfc-gh-jmichalak: UnssetSafely won't fail in this case - see TestInt_SafeUnsetTagFromColumn.

sfc-gh-jmichalak (commented):
LGTM

sfc-gh-asawicki (dismissed):
+1 for the open comments

pkg/resources/tag_association.go
sfc-gh-kwasilewski: Done, I added it to the Update and Delete functions

sfc-gh-jmichalak (commented):
LGTM

sfc-gh-asawicki (approved):

sfc-gh-kwasilewski (approved):

PR #4623, opened 2026-04-09
feat: Detect and restore external warehouse type changes for adaptive and standard warehouses

sfc-gh-jmichalak:
## Changes

The `snowflake_warehouse_adaptive` resource had no mechanism to detect when the warehouse type was externally changed (e.g., via the Snowflake UI or SQL). This left Terraform unaware of drift, causing the declared type to silently diverge from the actual type.

- Add `warehouse_type` computed attribute to `snowflake_warehouse_adaptive` to track the actual warehouse type
- Add `HandleWarehouseExternalTypeChange` custom diff that detects type drift and schedules an UPDATE to restore the declared type (without forcing recreation)
- Add acceptance tests for both resources to confirm the behavior.
- Remove the stale documentation note that external type detection was unsupported

### DIFF
--- a/pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
+++ b/pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
@@ -325,3 +325,46 @@ func TestAcc_WarehouseAdaptive_Validations(t *testing.T) {
 		},
 	})
 }
+
+func TestAcc_WarehouseAdaptive_ExternalTypeChange(t *testing.T) {
+	id := testClient().Ids.RandomAccountObjectIdentifier()
+	warehouseModel := model.WarehouseAdaptiveWithId(id)
+	ref := warehouseModel.ResourceReference()
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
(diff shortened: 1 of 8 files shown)


pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
```diff
@@ -325,3 +325,46 @@ func TestAcc_WarehouseAdaptive_Validations(t *testing.T) {
...
+			},
+			{
+				PreConfig: func() {
+					testClient().Warehouse.UpdateWarehouseType(t, id, sdk.WarehouseTypeStandard)
+				},
+				Config: accconfig.FromModels(t, warehouseModel),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionUpdate),
+						planchecks.ExpectChange(ref, "warehouse_type", tfjson.ActionUpdate, sdk.String(string(sdk.WarehouseTypeStandard)), sdk.String(string(sdk.WarehouseTypeAdaptive))),
```
sfc-gh-jcieslak, line 358: nit: this check is missing in another test

sfc-gh-jcieslak (approved):

pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
sfc-gh-jmichalak: Right, I'll add this later.

sfc-gh-asawicki (commented):
LGTM

PR #4629, opened 2026-04-10
chore: Add aliases to SDK enum generation

sfc-gh-asawicki:
- Support generation of enums with aliases.
- Propose a generic enum conversion test.
- Enable conditional part generation (to avoid having empty files generated when no enum is defined for the object).

Next PRs:
- Move all enum tests to the generic test if accepted
- Generate the enum test addition in the unit test generation part
- Move all enums to generator


pkg/sdk/generator/example/defs/enum_examples_def.go
```diff
@@ -0,0 +1,31 @@
...
+	"ExampleStatuses",
+	"ACTIVE", "INACTIVE", "EXPIRED",
+)
+
+// ExampleSizeDef demonstrates an enum with aliases, similar to WarehouseSize.
+var ExampleSizeDef = g.NewEnum(
+	"ExampleSize",
+	"ExampleSizes",
+	"XSMALL", "SMALL", "MEDIUM", "LARGE", "XLARGE",
+).WithAliases("XSMALL", "X-SMALL").
```
sfc-gh-kwasilewski, line 23: I'd add one example with multiple aliases pointing to the same element to double-check that the template correctly supports multiple aliases.

pkg/sdk/generator/gen/enum.go
```diff
@@ -3,43 +3,61 @@ package gen
...
 		NamePlural: namePlural,
 		Values:     values,
+		Aliases:    make(map[string][]string),
 	}
 }
 
+// WithAliases adds aliases for a canonical enum value.
+// E.g. for canonical value "XSMALL" with aliases "X-SMALL".
+func (e *Enum) WithAliases(value string, aliases ...string) *Enum {
+	e.Aliases[value] = append(e.Aliases[value], aliases...)
```
sfc-gh-kwasilewski, line 33: I think we should fail if the value is not in the map

pkg/sdk/0_enum_conversion_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// allEnumConversionTests keeps all enums which conversion needs to be tested
+// Currently, it's populated from the list inside Test_AllEnumConversions.
+// Ultimately, it will be populated also through init method in the object's test file from each enum existing in the definition.
+var allEnumConversionTests []enumTestProvider
```
sfc-gh-kwasilewski, line 14: I wonder whether we need to test ALL enum conversions if they're all generated the same way.

sfc-gh-jmichalak (dismissed):
\+1 to some other comments.

pkg/sdk/generator/gen/enum.go
sfc-gh-asawicki: Friendly error handling is missing in all the generators (it should be improved overall). If you browse through this and other generators, you'll see that it misses it too, sometimes the panic is used to mark some obvious situations.
For now, I'll add a quick panic like in some other places. And I will add the better error handling to generator helpers readme (if its missing).

pkg/sdk/0_enum_conversion_test.go
sfc-gh-asawicki: This was the concern even before this change. We could extend this concern to other sdk unit tests (e.g. the show sql creation which is usually the same and reuses the same dsl methods).
The good parts are:
- the unit tests are super fast and cheap
- the enum conversion tests won't add a lot of code when using thius approach.

For these reasons, I would leave them for now.

WDYT?

pkg/sdk/generator/gen/enum.go
sfc-gh-kwasilewski: Fair enough, +1 to adding a panic

pkg/sdk/generator/example/defs/enum_examples_def.go
sfc-gh-asawicki: Done

pkg/sdk/generator/gen/enum.go
sfc-gh-asawicki: Done

pkg/sdk/0_enum_conversion_test.go
sfc-gh-kwasilewski: It's true that the unit tests are fast, that's why it's not a major concern. Because you've already implemented the generation, we can keep it.

sfc-gh-kwasilewski (approved):

sfc-gh-jmichalak (approved):

sfc-gh-asawicki:
Failing `TestAcc_Tags_CompleteUseCase ` is unrelated and is being diagnosed as a separate effort.

PR #4645, opened 2026-04-14
feat: add snowflake_session_policy resource

sfc-gh-kwasilewski:
Added new snowflake_session_policy resource.

## TODO
- Due to SDK limitations, it's not possible to disallow all secondary roles by specifying an empty set for the allowed_secondary_roles property. This will be reworked in a follow-up PR.


pkg/resources/session_policy.go
```diff
@@ -0,0 +1,315 @@
...
+		Description: joinWithSpace("Specifies the allowed secondary roles for a session policy, if any.",
+			"Use a single-element set whose only value is `all` (case-insensitive) to allow all secondary roles, equivalent to `('ALL')` in Snowflake."),
+		DiffSuppressFunc: SuppressIfAny(
+			IgnoreChangeToCurrentSnowflakeValueInDescribe("allowed_secondary_roles"),
+			NormalizeAndCompareIdentifiersInSet("allowed_secondary_roles"),
+		),
+		Elem: &schema.Schema{
+			Type:             schema.TypeString,
+			ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
+			DiffSuppressFunc: suppressIdentifierQuoting,
```
sfc-gh-jmichalak, line 68: Sets with elem-scoped diffsuppress need to have NormalizeAndCompareIdentifiersInSet diff suppressor.

pkg/resources/session_policy.go
```diff
@@ -0,0 +1,315 @@
...
+	if !reflect.DeepEqual(*unsetRequest, *sdk.NewSessionPolicyUnsetRequest()) {
+		if err := client.SessionPolicies.Alter(ctx, sdk.NewAlterSessionPolicyRequest(id).WithUnset(*unsetRequest)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	return ReadSessionPolicyFunc(false)(ctx, d, meta)
+}
+
+func secondaryRolesAttributeCreateBuilder[T any](d *schema.ResourceData, key string, setValue func(request sdk.SessionPolicySecondaryRolesRequest) *T) error {
```
sfc-gh-jmichalak, line 269: Nit: instead of using these two, we could use attributeMappedValueCreateBuilder and setValueUpdate. If not, let's have more generic wrappers/alternatives to these ones.

pkg/resources/session_policy.go
```diff
@@ -0,0 +1,315 @@
...
+			StateContext: TrackingImportWrapper(resources.SessionPolicy, ImportName[sdk.SchemaObjectIdentifier]),
+		},
+		Timeouts: defaultTimeouts,
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.SessionPolicy, customdiff.All(
+			ComputedIfAnyAttributeChanged(sessionPolicySchema, ShowOutputAttributeName, "name", "schema", "database", "comment"),
+			ComputedIfAnyAttributeChanged(sessionPolicySchema, DescribeOutputAttributeName, "name", "schema", "database", "session_idle_timeout_mins", "session_ui_idle_timeout_mins", "allowed_secondary_roles", "blocked_secondary_roles", "comment"),
+			ComputedIfAnyAttributeChanged(sessionPolicySchema, FullyQualifiedNameAttributeName, "name", "schema", "database"),
+		)),
+	}
+}
```
sfc-gh-jmichalak, line 138: The migration guide is missing.

pkg/resources/session_policy.go
```diff
@@ -0,0 +1,315 @@
...
+	roles := expandStringList(rawRoles)
+	request := sdk.NewSessionPolicySecondaryRolesRequest()
+	var errs []error
+	switch {
+	case len(roles) == 0:
+		request.WithNone(true)
+	case len(roles) == 1 && strings.EqualFold("ALL", roles[0]):
+		request.WithAll(true)
+	default:
+		request.WithRoles(collections.Map(roles, func(elem string) sdk.AccountObjectIdentifier {
```
sfc-gh-jmichalak: Nit: use MapErr instead.

pkg/resources/session_policy.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	return nil
+}
+
+func buildSessionPolicySecondaryRolesRequest(rawRoles []any) (*sdk.SessionPolicySecondaryRolesRequest, error) {
+	roles := expandStringList(rawRoles)
+	request := sdk.NewSessionPolicySecondaryRolesRequest()
+	var errs []error
+	switch {
+	case len(roles) == 0:
```
sfc-gh-jmichalak, line 300: So here we set None if the list is empty. However, in create and update, if the key doesn't exist in the config, we won't even call this function, right? I know this will be handled in the next PR, but I'm just flagging this for myself.

sfc-gh-jmichalak (changes requested):

pkg/resources/session_policy.go
sfc-gh-kwasilewski: uu, nice

pkg/resources/session_policy.go
sfc-gh-kwasilewski: > if the key doesn't exist in the config, we won't even call this function, right?

That's right, and this behavior is fine. The problem is that it won't be called if the key exists in the config, but it is set to an empty set.

pkg/resources/session_policy.go
sfc-gh-kwasilewski: I think it already does - am I missing something?

pkg/resources/session_policy.go
sfc-gh-kwasilewski: Fixed

pkg/resources/session_policy.go
sfc-gh-kwasilewski: Done

pkg/resources/session_policy.go
sfc-gh-kwasilewski: Let's adjust it in a follow-up PR, as this code will likely change anyway

pkg/resources/session_policy.go
sfc-gh-jmichalak: Ah, yes, mb

sfc-gh-jmichalak (approved):

docs/resources/session_policy.md
```diff
@@ -0,0 +1,114 @@
+---
```
sfc-gh-asawicki, line 1: nit: there is a way to add the resource just to our acceptance tests provider. This way we don't need to upfront add all docs, examples, migration guide, etc. it's purely in the development state. That way we can improve it as we go (here are some follow-ups that we need to do before the resource is ready). The `dev` branch stays "clean" and in case of urgent release, we dont need to care about such in-progress objects.

This is just informational, close after reading.

pkg/testacc/resource_session_policy_acceptance_test.go
```diff
@@ -0,0 +1,315 @@
...
+	t.Cleanup(role2Cleanup)
+	role3, role3Cleanup := testClient().Role.CreateRole(t)
+	t.Cleanup(role3Cleanup)
+
+	basic := model.SessionPolicy("t", id.DatabaseName(), id.SchemaName(), id.Name())
+
+	altered := model.SessionPolicy("t", newId.DatabaseName(), newId.SchemaName(), newId.Name()).
+		WithSessionIdleTimeoutMins(sessionIdleTimeoutMins).
+		WithSessionUiIdleTimeoutMins(sessionUiIdleTimeoutMins).
+		// TODO: make it possible to provide an empty list
```
sfc-gh-asawicki, line 55: nit: usually we leave a SNOW-xxx in brackets to know what issue the TODO is tied to. In this case, you will do it in the next PR, so the current SNOW-xxx could be used.

It's only infromational, close after reading.

sfc-gh-asawicki (approved):

pkg/testacc/resource_session_policy_acceptance_test.go
sfc-gh-kwasilewski: That's right, it's not meant to be a long-living TODO comment, so I didn't add a ticket ID.

PR #4655, opened 2026-04-16
chore: Migrate majority of db and plain struct definitions to paired structs approach

sfc-gh-asawicki:
Migrate the majority db/plain struct definitions to the paired struct approach, which is essential before the `convert` generation
- SDK was regenerated to validate no changes in these structs
- new helpers were introduced
- add missing comments (`manually adjusted`)
- notably the columns with `?` are now natively supported without problems

Follow-ups:
- migrate the structs used in custom operations
- decide what about the db/plain structs that have the varying attributes length (e.g. functions, procedures)

### DIFF
--- a/pkg/sdk/generator/defs/api_integrations_def.go
+++ b/pkg/sdk/generator/defs/api_integrations_def.go
@@ -123,22 +123,15 @@ var apiIntegrationsDef = g.NewInterface(
 			Name().
 			WithValidation(g.ValidIdentifier, "name"),
 	).
-	ShowOperation(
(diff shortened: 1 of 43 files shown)


sfc-gh-kwasilewski:
If columns with ? are now natively supported without problems, how about adjusting the def file for hybrid tables, where it should be `null?`, but now we have `null`?

sfc-gh-asawicki:
We discussed that hybrid tables will be adjusted after the tests are merged in the dedicated PR.

sfc-gh-kwasilewski (approved):

PR #4658, opened 2026-04-16
feat: nested none/all/roles for allowed and blocked secondary roles

sfc-gh-kwasilewski:
Follow-up to 0f0ca926. Removes the TODO about session_policy resource limits: an empty allowed_secondary_roles set could not express “disallow all” secondary roles.

Model each side as a single nested block with exactly one of none, all, or roles set; update docs and examples accordingly.

### DIFF
--- a/pkg/resources/session_policy.go
+++ b/pkg/resources/session_policy.go
@@ -54,35 +54,22 @@ var sessionPolicySchema = map[string]*schema.Schema{
 		ValidateFunc:     validation.IntAtLeast(1),
 	},
 	"allowed_secondary_roles": {
-		Type:     schema.TypeSet,
-		Optional: true,
-		Description: joinWithSpace("Specifies the allowed secondary roles for a session policy, if any.",
-			"Use a single-element set whose only value is `all` (case-insensitive) to allow all secondary roles, equivalent to `('ALL')` in Snowflake."),
-		DiffSuppressFunc: SuppressIfAny(
-			IgnoreChangeToCurrentSnowflakeValueInDescribe("allowed_secondary_roles"),
-			NormalizeAndCompareIdentifiersInSet("allowed_secondary_roles"),
-		),
-		Elem: &schema.Schema{
-			Type:             schema.TypeString,
-			ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
-			DiffSuppressFunc: suppressIdentifierQuoting,
+		Type:        schema.TypeList,
+		Optional:    true,
+		MaxItems:    1,
+		Description: "Specifies the allowed secondary roles for a session policy, if any.",
+		Elem: &schema.Resource{
+			Schema: sessionPolicySecondaryRolesSchema("allowed_secondary_roles"),
 		},
 	},
 	"blocked_secondary_roles": {
-		Type:     schema.TypeSet,
+		Type:     schema.TypeList,
 		Optional: true,
+		MaxItems: 1,
 		Description: joinWithSpace("Specifies the blocked secondary roles for a session policy, if any.",
-			"Blocked secondary roles take precedence over allowed secondary roles.",
-			"Use a single-element set whose only value is `all` (case-insensitive) to disallow all secondary roles, equivalent to `('ALL')` in Snowflake.",
-		),
-		DiffSuppressFunc: SuppressIfAny(
-			IgnoreChangeToCurrentSnowflakeValueInDescribe("blocked_secondary_roles"),
-			NormalizeAndCompareIdentifiersInSet("blocked_secondary_roles"),
-		),
-		Elem: &schema.Schema{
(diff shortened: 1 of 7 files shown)


pkg/resources/session_policy.go
```diff
@@ -54,35 +54,22 @@ var sessionPolicySchema = map[string]*schema.Schema{
...
-		Elem: &schema.Schema{
-			Type:             schema.TypeString,
-			ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
-			DiffSuppressFunc: suppressIdentifierQuoting,
+		Type:        schema.TypeList,
+		Optional:    true,
+		MaxItems:    1,
+		Description: "Specifies the allowed secondary roles for a session policy, if any.",
+		Elem: &schema.Resource{
+			Schema: sessionPolicySecondaryRolesSchema("allowed_secondary_roles"),
```
sfc-gh-kwasilewski, line 62: NormalizeAndCompareIdentifiersInSet only works for top-level TypeSet fields, it would no longer work for the nested `roles` field, so I removed it.

pkg/testacc/resource_session_policy_acceptance_test.go
```diff
@@ -267,12 +264,16 @@ func TestAcc_SessionPolicy_BasicUseCase(t *testing.T) {
...
+							planchecks.ExpectDrift(ref, "allowed_secondary_roles.0.none", sdk.String("false"), sdk.String("true")),
+							planchecks.ExpectDrift(ref, "blocked_secondary_roles.0.roles", blockedSecondaryRoles, sdk.String("[]")),
+							planchecks.ExpectDrift(ref, "blocked_secondary_roles.0.all", sdk.String("false"), sdk.String("true")),
 							planchecks.ExpectDrift(ref, "comment", sdk.String(comment), sdk.String(externalComment)),
 							planchecks.ExpectChange(ref, "session_idle_timeout_mins", tfjson.ActionUpdate, externalSessionIdle, sessionIdle),
 							planchecks.ExpectChange(ref, "session_ui_idle_timeout_mins", tfjson.ActionUpdate, externalSessionUiIdle, sessionUiIdle),
-							planchecks.ExpectChange(ref, "blocked_secondary_roles", tfjson.ActionUpdate, sdk.String("[ALL]"), blockedSecondaryRoles),
+							planchecks.ExpectChange(ref, "allowed_secondary_roles.0.none", tfjson.ActionUpdate, sdk.String("true"), nil),
+							planchecks.ExpectChange(ref, "blocked_secondary_roles.0.roles", tfjson.ActionUpdate, sdk.String("[]"), blockedSecondaryRoles),
+							planchecks.ExpectChange(ref, "blocked_secondary_roles.0.all", tfjson.ActionUpdate, sdk.String("true"), nil),
```
sfc-gh-asawicki, line 324: let's add a test verifying more transitions between the 3 options (all, none, roles).

sfc-gh-asawicki (changes requested):

pkg/testacc/resource_session_policy_acceptance_test.go
sfc-gh-kwasilewski: Added one additional test step. Here's an up-to-date list of all potential transitions and which test step covers each of them:

1. Roles -> None -- external drift, step 9 
2. Roles -> All -- external drift, step 9
3. Roles -> Empty
4. All -> None -- OK, step 4 
5. All -> Roles -- OK, step 9
6. All -> Empty -- OK, step 5
7. None -> All -- OK, step 4
8. None -> Roles -- OK, step 9
9. None -> Empty -- OK, step 5
10. Empty -> Roles
11. Empty -> All -- OK, step 3
12. Empty -> None -- OK, step 3

I want to keep these tests relatively short, so not all transitions are covered, specifically 3) and 10). Also, let me know if you would like me to add more tests for external changes detection

sfc-gh-asawicki (approved):

pkg/testacc/resource_session_policy_acceptance_test.go
sfc-gh-asawicki: Let's ad a few more external drift detection checks (in a follow-up PR). The rationale would be that this is a complex field, and better to be safe than sorry later. 

Question:
`Roles -> Empty` should it be allowed at all? For empty, `None` should be used, right?

sfc-gh-kwasilewski: By `Empty` I mean that the field isn't present in the config. The effect differs across fields; e.g., for allowed_secondary_roles, it means all roles are allowed, while for blocked_secondary_roles, it means none are allowed.

sfc-gh-kwasilewski: I'll add more external drift detection checks in a follow-up

Issue #3896, opened 2025-08-01
[Bug]: Adding a comment to a stream in snowflake_stream_on_table causes the stream to be recreated


y3yokota:
### Terraform CLI Version

1.3.1

### Terraform Provider Version

2.1.0

### Company Name

Nomura Research Institute, Ltd.

### Terraform Configuration

```terraform
1. create stream
resource "snowflake_stream_on_table" "test_stream" {
  schema            = "schema"
  database          = "database"
  name              = "TEST_STREAM"
  append_only       = true
  comment          = "default"
  show_initial_rows = true
}
# show_initial_rows can be true or false or unspecified

2. create new tfstate file ( terraform import )
show_initial_rows is changed "null" in tfstate file

3. comment update ( terraform apply )
plan have only update, but stream is recreated
show_initial_rows is changed "default" in tfstate file
```

### Category

category:resource

### Object type(s)

resource:stream_on_table

### Expected Behavior

stream is not recreated, or I can see in the plan that show_initial_rows will be changed from null to default.

### Actual Behavior

stream is recreated.
When importing an existing stream, show_initial_rows becomes null (as per the specification).
After that, if you apply just adding a comment, the stream will be recreated.
Since the plan only shows the addition of a comment, it is not clear that the stream will be recreated.
When simply adding a comment, rewriting while show_initial_rows is hidden can lead to serious accidents.

### Steps to Reproduce

1. create stream
2. terraform import ( make new statefile ) & refresh
3. comment update
4. terraform apply

### How much impact is this issue causing?

Medium

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
Hi @y3yokota 👋 

I think this behavior depends on the import. When I ran these steps with the `terraform import` command (1), I got the following plan (independently of the `comment` field, there is a change on `show_initial_rows`:
```
  ~ resource "snowflake_stream_on_table" "test_stream" {
      ~ comment              = "default" -> "defult"
      ~ describe_output      = [
          - {
              - base_tables     = [
                  - "\"JMICHALAK\".\"PUBLIC\".\"TEST\"",
                ]
              - comment         = "default"
              - created_on      = "2025-08-01 03:17:57.32 -0700 PDT"
              - database_name   = "JMICHALAK"
              - invalid_reason  = "N/A"
              - mode            = "APPEND_ONLY"
              - name            = "TEST_STREAM"
[...]

PR #4659, opened 2026-04-16
fix: Set show_initial_rows on stream import

sfc-gh-jmichalak:
## Summary
- Importing `snowflake_stream_on_table` or `snowflake_stream_on_view` left `show_initial_rows` as `null` in state, causing silent stream recreation on the next apply.
- Adds a v0→v1 state upgrader (`v2_15_0_StreamStateUpgrader`) for both resources that backfills `show_initial_rows` with the default value.
- Fixes the import functions to explicitly set `show_initial_rows` during import.
- Adds an acceptance test that reproduces the bug with v2.15.0 and verifies the state upgrader and import fix resolve it.
- Adds migration guide entry.

[#3896](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3896)


pkg/testacc/2_provider_helpers_test.go
```diff
@@ -39,6 +39,32 @@ func ExternalProviderWithExactVersion(version string) map[string]resource.Extern
...
+//	| Could not retrieve the list of available versions for provider
+//	| hashicorp/snowflake: provider registry registry.terraform.io does not have a
+//	| provider named registry.terraform.io/hashicorp/snowflake
+//	|
+//	| All modules should specify their required_providers so that external
+//	| consumers will get the correct providers when using a module. To see which
+//	| modules are currently depending on hashicorp/snowflake, run the following
+//	| command:
+//	|     terraform providers
+func requiredProvidersBlock(version string) string {
```
sfc-gh-asawicki: It doesn't seem to be used anywhere, is it? Also, this is not a deterministic error. I am unsure that this would be the solution.

sfc-gh-asawicki (changes requested):

pkg/testacc/2_provider_helpers_test.go
sfc-gh-jmichalak: It fixed that. This is now removed.

sfc-gh-asawicki (approved):

PR #4661, opened 2026-04-16
chore: Add custom interface method to SDK gen

sfc-gh-asawicki:
Allow defining custom methods in SDK interfaces
The methods need to be manually implemented in the `ext` files

Follow-up:
- Use this new definition for all existing custom methods


pkg/sdk/generator/gen/templates/interface.tmpl
```diff
@@ -30,4 +30,12 @@ type {{ .Name }} interface {
 			{{ .Name }}(ctx context.Context, request *{{ .OptsField.DtoDecl }}) error
 		{{- end -}}
 	{{ end }}
+	{{- range .CustomMethods }}
+		{{- if .Doc }}
+			{{- range splitLines .Doc }}
+			// {{ . }}
+			{{- end }}
+		{{- end }}
+		{{ .Name }}(ctx context.Context{{ range .Parameters }}, {{ .Name }} {{ .Kind }}{{ end }}){{ if eq (len .ReturnTypes) 0 }}{{ else if eq (len .ReturnTypes) 1 }} {{ index .ReturnTypes 0 }}{{ else }} ({{ join .ReturnTypes ", " }}){{ end }}
```
sfc-gh-jcieslak, line 39: nit: return type could be simplified by checking if ReturnTypes == 1 or > 1 (instead of == 0, ==1, else)

sfc-gh-jcieslak (approved):

pkg/sdk/generator/gen/templates/interface.tmpl
sfc-gh-jcieslak: or we could have a helper function so we could call shorter `{{ formatReturns .Returns }}`

sfc-gh-asawicki: I'll adjust in the follow up with usage

PR #4664, opened 2026-04-17
fix: Add an experiment for granting account roles to the PUBLIC role

sfc-gh-jmichalak:
## Summary
- Adds a new `GRANT_ACCOUNT_ROLE_SAFE_PUBLIC_ROLE` experimental feature that makes `snowflake_grant_account_role` treat granting the PUBLIC role as a silent no-op
- Snowflake implicitly grants PUBLIC to every role and user, so `GRANT ROLE PUBLIC` is always a no-op at the SQL level and `SHOW GRANTS` never lists it — this caused an inconsistent-result error in the provider
- With the experiment enabled, Create, Read, and Delete all short-circuit for PUBLIC role grants, avoiding the broken `SHOW GRANTS` lookup

https://github.com/snowflakedb/terraform-provider-snowflake/issues/3001


pkg/resources/grant_account_role.go
```diff
@@ -87,44 +88,66 @@ func GrantAccountRole() *schema.Resource {
...
 // CreateGrantAccountRole implements schema.CreateFunc.
 func CreateGrantAccountRole(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-	client := meta.(*provider.Context).Client
+	providerCtx := meta.(*provider.Context)
+	client := providerCtx.Client
 	roleName := d.Get("role_name").(string)
 	roleIdentifier := sdk.NewAccountObjectIdentifierFromFullyQualifiedName(roleName)
+
+	safePublicRole := experimentalfeatures.IsExperimentEnabled(experimentalfeatures.GrantAccountRoleSafePublicRole, providerCtx.EnabledExperiments) &&
+		strings.EqualFold(roleIdentifier.Name(), snowflakeroles.Public.Name())
```
sfc-gh-kwasilewski, line 97: It's actually possible to have a user-defined `"public"` role alongside the `PUBLIC` role, we need to have an exact, case-sensitive match here.

sfc-gh-kwasilewski (changes requested):

pkg/resources/grant_account_role.go
sfc-gh-jmichalak: It is possible, but this is already handled by the ID parsing functions.

sfc-gh-kwasilewski: How is that? `NewAccountObjectIdentifierFromFullyQualifiedName` strips quotes but preserves case, so "public" becomes public, and because of EqualFold, public == PUBLIC.

sfc-gh-jmichalak: Right, I didn't think you meant the strings.EqualFold part. Done.

sfc-gh-kwasilewski (approved):

PR #4665, opened 2026-04-17
chore: Move manual enums to SDK generator part5

sfc-gh-asawicki:
- Move compute_pools, data_metric_function_references, listings, network_rules, organization_accounts, services, streams, views enums to sdk generator
- Rename the values to match the generated ones


pkg/sdk/data_metric_function_references_enums.go
```diff
@@ -5,12 +5,6 @@ import (
 	"strings"
 )
 
-type DataMetricFunctionRefEntityDomainOption string
-
-const (
-	DataMetricFunctionRefEntityDomainView DataMetricFunctionRefEntityDomainOption = "VIEW"
-)
-
 type DataMetricScheduleStatusOption string
```
sfc-gh-jcieslak, line 8: Q: Is there a reason enums like those are not migrated as well? Same as with ComputePoolInstanceFamily

pkg/sdk/data_metric_function_references_enums_gen.go
```diff
@@ -0,0 +1,28 @@
+// Code generated by SDK builder generator (v0.1.0); DO NOT EDIT.
+
+package sdk
+
+import (
+	"fmt"
+	"strings"
+)
+
+type DataMetricFunctionRefEntityDomainOption string
```
sfc-gh-jcieslak, line 10: nit: Option suffix could be removed

pkg/sdk/network_rules_enums_gen.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+type NetworkRuleType string
+
+const (
+	NetworkRuleTypeIpv4            NetworkRuleType = "IPV4"
+	NetworkRuleTypeAwsvpceid       NetworkRuleType = "AWSVPCEID"
```
sfc-gh-jcieslak, line 14: nit: Due to value mapping directly to name, it lost a bit of readability. We could use a custom name in cases like those.

sfc-gh-jcieslak (approved):

pkg/sdk/data_metric_function_references_enums.go
sfc-gh-asawicki: `DataMetricScheduleStatusOption` can be migrated too, however it have multiple `All` collections, it can be done in the follow-up.
ComputePoolInstanceFamily has names like `GPU_GCP_NV_L4_1_24G` which will be changed to something like: `ComputePoolInstanceFamilyGpuGcpNvL4124G` (no separators between digits). For these, it would be better to improve the generator to allow custom name overrides (like with AWSVPCEID below).

pkg/sdk/data_metric_function_references_enums_gen.go
sfc-gh-asawicki: There are a few enums using the Option suffix (the result of hand-writing the enums before). I didn't want to do a lot of renaming with the changes but otherwise I agrre that the Option suffix could be remvoed.

pkg/sdk/network_rules_enums_gen.go
sfc-gh-asawicki: I agree, I'll document the need for the addition of such value renames.

PR #4666, opened 2026-04-17
feat: Add snowflake_user_session_policy_attachment resource

sfc-gh-kwasilewski:
Attach a session policy to a user via ALTER USER ... SET SESSION POLICY, behind a preview feature flag. This resource follows the same attachment convention as `snowflake_user_authentication_policy_attachment` and `snowflake_user_password_policy_attachment`.

## Changes

- Add `user_session_policy_attachment` resource.
- Extend resource and data source picklists in `.github/ISSUE_TEMPLATE/` issue forms.


examples/resources/snowflake_user_session_policy_attachment/import.sh
```diff
@@ -0,0 +1 @@
+terraform import snowflake_user_session_policy_attachment.example 'USER_NAME|"<database_name>"."<schema_name>"."<session_policy_name>"'
```
sfc-gh-jcieslak: nit: for consistency it should be `<user_name>`

pkg/resources/user_session_policy_attachment.go
```diff
@@ -0,0 +1,149 @@
...
+
+func ReadUserSessionPolicyAttachment(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+
+	parts := helpers.ParseResourceIdentifier(d.Id())
+	if len(parts) != 2 {
+		return diag.FromErr(fmt.Errorf("required id format '<user_identifier>|<session_policy_fqn>', but got: '%s'", d.Id()))
+	}
+
+	userName := sdk.NewAccountObjectIdentifierFromFullyQualifiedName(parts[0])
```
sfc-gh-jcieslak: use sdk.ParseAccountObjectIdentifier over those constructor fuctions

pkg/resources/user_session_policy_attachment.go
```diff
@@ -0,0 +1,149 @@
...
+		return diag.Diagnostics{
+			diag.Diagnostic{
+				Severity: diag.Warning,
+				Summary:  "Failed to find user's session policy. Marking the resource as removed.",
+				Detail:   fmt.Sprintf("User id: %s", userName.Name()),
+			},
+		}
+	}
+
+	if err := d.Set("user_name", userName.Name()); err != nil {
```
sfc-gh-jcieslak, line 118: I don't think setting user_name is necessary and could be removed

pkg/resources/user_session_policy_attachment.go
```diff
@@ -0,0 +1,149 @@
...
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		Description:      "User name of the user you want to attach the session policy to.",
+		ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
+	},
+	"session_policy_name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
```
sfc-gh-jcieslak, line 28: Can't session policy be non-force new param? Effectively, altering is the same as drop + create, but maybe we should do a quick check to see if there are any pros and cons of one approach over the other. Even if drop + create is present in other attachment resources.

pkg/resources/user_session_policy_attachment.go
```diff
@@ -0,0 +1,149 @@
...
+
+	parts := helpers.ParseResourceIdentifier(d.Id())
+	if len(parts) != 2 {
+		return diag.FromErr(fmt.Errorf("required id format '<user_identifier>|<session_policy_fqn>', but got: '%s'", d.Id()))
+	}
+
+	userName := sdk.NewAccountObjectIdentifierFromFullyQualifiedName(parts[0])
+	policyReferences, err := client.PolicyReferences.GetForEntity(ctx, sdk.NewGetForEntityPolicyReferenceRequest(userName, sdk.PolicyEntityDomainUser))
+	if err != nil {
+		if errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) {
```
sfc-gh-jcieslak, line 83: Should we, in this case, query the user to confirm it's not there? I'm guessing `sdk.ErrObjectNotExistOrAuthorized` comes from this fact, but I'm wondering if we should assume based on the error or explicitly check 🤔 It shouldn't happen that often, so additional check shouldn't cost much, but we can discuss

pkg/resources/user_session_policy_attachment.go
```diff
@@ -0,0 +1,149 @@
...
+	client := meta.(*provider.Context).Client
+
+	userName := sdk.NewAccountObjectIdentifierFromFullyQualifiedName(d.Get("user_name").(string))
+
+	err := client.Users.Alter(ctx, userName, &sdk.AlterUserOptions{
+		Unset: &sdk.UserUnset{
+			SessionPolicy: sdk.Bool(true),
+		},
+	})
+	if err != nil {
```
sfc-gh-jcieslak, line 146: I don't think this follows the concept of "SafeDelete," as I can imagine a case where this could possibly fail and block the deletion. I think if `errors.Is(err, sdk.ErrObjectNotExistOrAuthorized) or an additional user existence check should improve this a bit. It's still low risk as it would block only in some "unlucky" cases with race condition (read -> user removed after resource read passes -> delete fails because user is not there).

docs/index.md
```diff
@@ -150,7 +150,7 @@ provider "snowflake" {
 - `passcode_in_password` (Boolean) False by default. Set to true if the MFA passcode is embedded to the configured password. Can also be sourced from the `SNOWFLAKE_PASSCODE_IN_PASSWORD` environment variable.
```
sfc-gh-jcieslak, line 150: not related to this file, but migration guide entry is missing

pkg/testacc/resource_user_session_policy_attachment_acceptance_test.go
```diff
@@ -0,0 +1,148 @@
...
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionDestroy),
+					},
+				},
+				Config:  config.FromModels(t, newPolicy),
+				Destroy: true,
+			},
+			{
+				Config: config.FromModels(t, basic),
+			},
+			// Drop user externally and remove attachment from config - expect empty plan
```
sfc-gh-jcieslak, line 118: There's at least one test case that we're missing: drop user when the config is there.

pkg/testacc/resource_user_session_policy_attachment_acceptance_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/resourceassert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config/model"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_UserSessionPolicyAttachment(t *testing.T) {
```
sfc-gh-jcieslak: If possible use BasicUseCase or CompleteUseCase postfix to new tests for resources or data sources as this postfix is used for choosing preprod tests (see makefile: `test-main-terraform-use-cases`). It's unfortunately still waiting for me to document it :(

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: > I'm guessing sdk.ErrObjectNotExistOrAuthorized comes from this fact

Yes, that's right.

> Should we, in this case, query the user to confirm it's not there? 

Maybe, but this is only one problem. For example, according to [usage notes](https://docs.snowflake.com/en/sql-reference/functions/policy_references#usage-notes), it can also fail if the role that executes a query is not authorized to view any policy on the object. Should we have yet another explicit check for that case? We could, but that would already be 3 queries in one Read method - quite a lot. Because of that performance concern, I'd rather not add any additional checks. 
WDYT?

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: session_policy_name is encoded into the resource ID, so we'd need to figure out another way to construct the ID so it's always unique. It's doable, but requires more time to investigate.

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: Hm, import tests are starting to fail due to missing attributes after import when I remove it.

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: Yeah. I can add IF EXISTS to the ALTER statement to solve this problem.

pkg/testacc/resource_user_session_policy_attachment_acceptance_test.go
sfc-gh-kwasilewski: Isn't it more or less the same thing to test? This test shows that external change detection works correctly, so I think we can be sure that it would work if the config were there as well, even without an additional test.

examples/resources/snowflake_user_session_policy_attachment/import.sh
sfc-gh-kwasilewski: Fixed

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: Done

pkg/testacc/resource_user_session_policy_attachment_acceptance_test.go
sfc-gh-kwasilewski: Done, I added BasicUseCase suffix

docs/index.md
sfc-gh-kwasilewski: Added

pkg/resources/user_session_policy_attachment.go
sfc-gh-jcieslak: You also missed constructor usage in L138, can be changed in a follow-up if you want

pkg/resources/user_session_policy_attachment.go
sfc-gh-jcieslak: Because Import works by going into the Import method first (if defined) and then goes to the Read method (if defined). As we don't have an import method here, the Read method "takes its job". If we were to define a simple import method with this line (setting `user_name`), it would work again (even when you remove this line from Read). I would say this is a nit, because it works as intended even through Read. It's just unnecessarily sets the same thing on every read.

pkg/resources/user_session_policy_attachment.go
sfc-gh-jcieslak: I wouldn't say it's a blocking comment, but it's worth documenting for future exploration. Identifiers can be changed, e.g. you can look up how different resources handle renames, this could be also applied here. I would call it a nit, so it's up to you which option you choose.

pkg/resources/user_session_policy_attachment.go
sfc-gh-jcieslak: I wouldn't call it a performance concern, as additional calls would only occur in those special cases where the infrastructure resources were cleaned up manually without Terraform "knowledge". In this case, the user could just run `terraform state rm` to take the most performant path of removing the resource (since the user knows it is already removed from Snowflake). You can proceed with the change, but we could discuss what should be done in this case as we already started the movement of safely removing and handling attachment resources.

pkg/testacc/resource_user_session_policy_attachment_acceptance_test.go
sfc-gh-jcieslak: You're right, the case I'm describing would most likely fail on Create operation

sfc-gh-jcieslak (approved):

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: To be done in a follow-up

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: To be done in a follow-up

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: Let's discuss this further at the meeting.

pkg/resources/user_session_policy_attachment.go
sfc-gh-kwasilewski: Let's discuss this further at the meeting.

pkg/resources/user_session_policy_attachment.go
sfc-gh-asawicki: My take: ALTER is usually cleaner; we can update the resource ID (like we do with the renames). On the other hand, dropping before recreation is not that destructive; we only remove the association between the entities. If proceeding with alter path, please verify `d.Partial()`.

pkg/resources/user_session_policy_attachment.go
sfc-gh-asawicki: Let's discuss and maybe do it already with a `safely`-like approach

pkg/resources/user_session_policy_attachment.go
sfc-gh-asawicki: +1 for a dedicated import (we could also have verification in import that user, policy, and the connection all exist)

sfc-gh-asawicki (approved):

PR #4667, opened 2026-04-19
fix: Improve table data type parsing

sfc-gh-asawicki:
Fix TABLE data type parsing and improve data type handling in functions and procedures
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
  - Fix TABLE with parametrized column types (e.g. TABLE(col NUMBER(38,0), status VARCHAR(256))) parsing. The comma inside the type parameter was incorrectly treated as a column separator.                                                                                                                                                                                                                   
  - Prevent spurious plan diffs for functions and procedures when Snowflake returns a normalized data type that differs only in default parameter values (e.g. config has NUMBER, Snowflake returns NUMBER(38, 0)). This aligns function and procedure behavior with masking policy and row access policy, which already handled this correctly.                                                                                                                                                        
  - Add migration tests verifying that resources created with v2.15.0 produce no drift after upgrading to the current version.                                                                                                                                                                                                                                                                                                                                                                          
  - Add a migration guide entry for the TABLE parsing fix.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -48,6 +48,21 @@ To fix this, enable the `IMPORT_BOOLEAN_DEFAULT` experimental feature in the pro
 
 References: [#3896](https://github.com/snowflakedb/terraform-provider-snowflake/issues/3896)
 
+### *(bugfix)* TABLE data type parsing with parametrized column types
+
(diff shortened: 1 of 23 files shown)


sfc-gh-jcieslak (approved):

sfc-gh-asawicki:
The failing tests are not connected with this change and addressed in other PRs.

PR #4668, opened 2026-04-20
fix: tag propagation field

sfc-gh-jcieslak:
## Changes
- Changes `propagate` column type from non-null to nullable, as with lower Snowflake editions, it can be missing, which makes the tag resource fail. Update all places that use this field (assertions, tag resources, show_schema, etc.)

## References

* https://github.com/snowflakedb/terraform-provider-snowflake/issues/4651

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -46,6 +46,8 @@ The `propagate` field is now treated as nullable: when the column is absent from
 
 No changes in configuration are required. Users on standard accounts who experienced a crash on plan or apply should be able to use the `snowflake_tag` resource normally after upgrading.
 
+References: [#4651](https://github.com/snowflakedb/terraform-provider-snowflake/issues/4651)
+
 ## v2.14.x ➞ v2.15.0
 
 ### **IMPORTANT** *(improvement)* Go driver bumped to v2
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -36,13 +36,13 @@ No changes are required for existing configurations unless you want to manage se
 
 ### *(bugfix)* Fixed `snowflake_tag` crash on Standard accounts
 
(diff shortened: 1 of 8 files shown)


sfc-gh-kwasilewski (commented):
Should we add a MIGRATION_GUIDE entry?

MIGRATION_GUIDE.md
```diff
@@ -34,15 +34,19 @@ This feature will be marked as stable in future releases. To use it, add `snowfl
 
 No changes are required for existing configurations unless you want to manage session policies with Terraform.
 
-## v2.15.x ➞ v2.15.1
+### *(bugfix)* Fixed `snowflake_tag` crash on Standard accounts
 
-### *(bug fix)* `snowflake_stream_on_table` and `snowflake_stream_on_view` import fix
+In v2.15.0 we added the `propagate` field to the `snowflake_tag` resource. On [Standard accounts](https://docs.snowflake.com/en/user-guide/intro-editions#standard-edition), Snowflake does not return a `propagate` column in `SHOW TAGS`. The provider attempted to scan a SQL `NULL` into a non-nullable Go string, resulting in a fatal error whenever any `snowflake_tag` resource was refreshed on such accounts:
```
sfc-gh-kwasilewski: nit

> On [Standard accounts](https://docs.snowflake.com/en/user-guide/intro-editions#standard-edition), Snowflake does not return a `propagate` column in `SHOW TAGS`.

To be more precise, the column is present in the output, but the value is NULL.

sfc-gh-kwasilewski (dismissed):

MIGRATION_GUIDE.md
sfc-gh-jcieslak: That's true, I think the previous behavior was that it was missing, but as far as I checked now it's there and it is null, so adjusting the desc.

sfc-gh-kwasilewski (approved):

PR #4675, opened 2026-04-22
feat: sfpg sdk support

sfc-gh-jsouthard:
This commit adds full Go SDK support for Snowflake Postgres Instances to the Terraform provider. It introduces a CRUD interface following the provider's code-generation patterns and the documented SQL for performing these operations for Postgres clusters.
  
**Key changes**

  New SDK interface (`postgres_instances_gen.go`): Defines the PostgresInstances interface with these operations:

  • **Create** — provisions a new Postgres instance with options for compute family, storage size, Postgres version, HA, network policy, authentication authority, storage integration, settings, comments, tags, and forking from an existing instance (with point-in-time via AT/BEFORE timestamps or offsets)
  • **Alter** — rename, set/unset properties (network policy, auth authority, comment, HA, compute family, storage, version, maintenance window, settings), suspend, resume, reset access, and manage tags
  • **Drop** — with IF EXISTS support
  • **Show** — list instances with LIKE, STARTS WITH, and LIMIT filtering, plus ShowByID helpers
  • **Describe** — returns key/value property pairs for an instance

  **Enums** (`postgres_instances_enums.go`):

  • **PostgresInstanceState** — 10 lifecycle states (CREATING, RESTORING, STARTING, REPLAYING, FINALIZING, READY, RESTARTING, RESUMING, SUSPENDING, SUSPENDED)
  • **PostgresInstanceAuthenticationAuthority** — POSTGRES, POSTGRES_OR_SNOWFLAKE

  **Generated code** (following the provider's SDK builder generator pattern):

  • `postgres_instances_dto_gen.go` / `postgres_instances_dto_builders_gen.go` — request DTOs and builder methods
  • `postgres_instances_impl_gen.go` — implementation wiring (SQL construction, row-to-struct mapping)
  • `postgres_instances_validations_gen.go` — input validation (valid identifiers, exactly-one-of constraints, at-least-one-of constraints)
  • `postgres_instances_gen_test.go` — unit tests for all operations (411 lines)

  **Integration points**:

  • `client.go` — registers PostgresInstances on the SDK client
  • `object_types.go` — adds ObjectTypePostgresInstance ("POSTGRES INSTANCE")
  • `0_init.go` — registers the generator definition
  • `sdkcommons/enums.go` — registers the PostgresInstanceAuthenticationAuthority enum type for the generator


pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+	"https://docs.snowflake.com/en/sql-reference/sql/drop-postgres-instance",
+	g.NewQueryStruct("DropPostgresInstance").
+		Drop().
+		SQL("POSTGRES INSTANCE").
+		IfExists().
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).ShowOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/show-postgres-instances",
+	g.DbStruct("postgresInstancesRow").
```
sfc-gh-jcieslak, line 140: nit: Could use StructPair (as in: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4655)

pkg/sdk/postgres_instances_enums_gen.go
```diff
@@ -0,0 +1,66 @@
+package sdk
+
+import (
+	"fmt"
+	"strings"
+)
+
+type PostgresInstanceState string
```
sfc-gh-jcieslak, line 10: nit: Enums can be also generated (as in: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4654)

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/generator/gen/sdkcommons"
+)
+
+var postgresInstancesDef = g.NewInterface(
+	"PostgresInstances",
+	"PostgresInstance",
+	g.KindOfT[sdkcommons.AccountObjectIdentifier](),
+).CreateOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/create-postgres-instance",
```
sfc-gh-jcieslak, line 26: Currently, this includes parameters for both `CREATE` versions. In case of multiple paths for commands, we create separate Create operations. An example could be `procedure_def.go` where we have separate operations per language (e.g. `CreateForJavaScript`, `CreateForJava`, etc.) from sdk usage this gives better separation and mental burden of checking which version of create we're using and what parameters are allowed. What's also important is that often those different versions end up as multiple resources to continue this separation on the resource level (again, procedures do it by separating by language, e.g. https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/procedure_java)

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+	g.NewQueryStruct("CreatePostgresInstance").
+		Create().
+		SQL("POSTGRES INSTANCE").
+		Name().
+		OptionalIdentifier("Fork", g.KindOfT[sdkcommons.AccountObjectIdentifier](), g.IdentifierOptions().SQL("FORK")).
```
sfc-gh-jcieslak: After the split we could make it required (and make compute_family, storage_size_gb, authentication_authority required in the other create operation)

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+		OptionalTags().
+		WithValidation(g.ValidIdentifier, "name"),
+).AlterOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/alter-postgres-instance",
+	g.NewQueryStruct("AlterPostgresInstance").
+		Alter().
+		SQL("POSTGRES INSTANCE").
+		IfExists().
+		Name().
+		Identifier("RenameTo", g.KindOfTPointer[sdkcommons.AccountObjectIdentifier](), g.IdentifierOptions().SQL("RENAME TO")).
```
sfc-gh-jcieslak, line 86: Should be optional as it's "one of" options in the alter like this:
```go
OptionalIdentifier("RenameTo", g.KindOfT[sdkcommons.AccountObjectIdentifier](), g.IdentifierOptions().SQL("RENAME TO")).

```

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+		OptionalQueryStructField(
+			"Set",
+			g.NewQueryStruct("PostgresInstanceSet").
+				OptionalTextAssignment("NETWORK_POLICY", g.ParameterOptions().SingleQuotes()).
```
sfc-gh-jcieslak, line 90: nit: Technically could be identifier instead of plain text (similarly to how RenameTo is defined)

pkg/sdk/postgres_instances_enums_gen.go
sfc-gh-asawicki, line 10: nit (can be done in a follow-up PR): we have added the enum generation in the meantime, check e.g. https://github.com/snowflakedb/terraform-provider-snowflake/pull/4635

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+		WithValidation(g.ExactlyOneValueSet, "RenameTo", "Set", "Unset", "Suspend", "Resume", "ResetAccess", "SetTags", "UnsetTags"),
+).DropOperation(
+	"https://docs.snowflake.com/en/sql-reference/sql/drop-postgres-instance",
+	g.NewQueryStruct("DropPostgresInstance").
+		Drop().
+		SQL("POSTGRES INSTANCE").
+		IfExists().
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).ShowOperation(
```
sfc-gh-asawicki, line 138: nit (can be fixed in the follow up PR): in the meantime we added a new, preferred paired struct syntax (because it will allow us simpler convert methods generation), check e.g. https://github.com/snowflakedb/terraform-provider-snowflake/pull/4655

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+		Number("RetentionTime").
+		Field("State", "PostgresInstanceState").
+		OptionalText("Comment"),
+	g.NewQueryStruct("ShowPostgresInstances").
+		Show().
+		SQL("POSTGRES INSTANCES").
+		OptionalLike().
+		OptionalStartsWith().
+		OptionalLimitFrom(),
+).ShowByIdOperationWithFiltering(
```
sfc-gh-asawicki, line 184: nit: this will be not needed after https://github.com/snowflakedb/terraform-provider-snowflake/pull/4674 is merged

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+			g.KeywordOptions().SQL("SET"),
+		).
+		OptionalQueryStructField(
+			"Unset",
+			g.NewQueryStruct("PostgresInstanceUnset").
+				OptionalSQL("COMMENT").
+				OptionalSQL("POSTGRES_SETTINGS").
+				OptionalSQL("NETWORK_POLICY").
+				OptionalSQL("MAINTENANCE_WINDOW_START").
+				OptionalSQL("STORAGE_INTEGRATION").
```
sfc-gh-jcieslak, line 114: In the documentation, does `[ , ... ]` mean that you can unset multiple Unset parameters at once (for example, `ALTER ... UNSET one_param, second_param`), or does it mean that any parameter allowed in SET can also be used with UNSET (for example, `ALTER ... UNSET one_param, postgres_settings`, where `postgres_settings` is listed under SET but not explicitly under UNSET)? If it’s the latter, then we’re missing some OptionalSQL entries in the UNSET option that are present in the SET option.

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+				WithValidation(g.AtLeastOneValueSet, "Comment", "PostgresSettings", "NetworkPolicy", "MaintenanceWindowStart", "StorageIntegration"),
+			g.ListOptions().NoParentheses().SQL("UNSET"),
+		).
+		OptionalSQL("SUSPEND").
+		OptionalSQL("RESUME").
+		OptionalQueryStructField(
+			"ResetAccess",
+			g.NewQueryStruct("PostgresInstanceResetAccess").
+				OptionalTextAssignment("FOR", g.ParameterOptions().NoEquals().SingleQuotes()),
```
sfc-gh-jcieslak, line 123: nit: For limited options like here we could define enum and use it here, like in procedures:
```go
OptionalAssignment("AUTO_EVENT_LOGGING", g.KindOfTPointer[sdkcommons.AutoEventLogging](), g.ParameterOptions().SingleQuotes()).
```
Note: With generated enums the kind (2nd parameter) will look a bit different ([reference](https://github.com/snowflakedb/terraform-provider-snowflake/pull/4675/changes#r3122738400))

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+	g "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/generator/gen"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/generator/gen/sdkcommons"
+)
+
+var postgresInstancesDef = g.NewInterface(
+	"PostgresInstances",
+	"PostgresInstance",
+	g.KindOfT[sdkcommons.AccountObjectIdentifier](),
+).CreateOperation(
```
sfc-gh-asawicki, line 25: Two variants of creation have different required/optional/allowed attributes. It's not expressed in the current definition. There are two ways to move forward:
1. (preferred) Split into two create operations: first is the default, the second one is a custom one for fork (check e.g. functions_def with multiple create variants)
2. Add validations to conflict fields (e.g. if for, then network policy cannot be set)

We prefer the first one as it is a cleaner split, this is generated, so we don't mint the duplication of common attributes.

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+	"https://docs.snowflake.com/en/sql-reference/sql/create-postgres-instance",
+	g.NewQueryStruct("CreatePostgresInstance").
+		Create().
+		SQL("POSTGRES INSTANCE").
+		Name().
+		OptionalIdentifier("Fork", g.KindOfT[sdkcommons.AccountObjectIdentifier](), g.IdentifierOptions().SQL("FORK")).
+		OptionalQueryStructField(
```
sfc-gh-asawicki, line 54: At and Before should have the validation added (only one of them can be set simultaneously)

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -0,0 +1,172 @@
...
+		OptionalText("Comment"),
+	g.NewQueryStruct("ShowPostgresInstances").
+		Show().
+		SQL("POSTGRES INSTANCES").
+		OptionalLike().
+		OptionalStartsWith().
+		OptionalLimitFrom(),
+).ShowByIdOperationWithFiltering(
+	g.ShowByIDLikeFiltering,
+).DescribeOperation(
```
sfc-gh-asawicki, line 186: This is a comment for the next PRs. When the describe output is in a form of key value rows, we add a special dedicated method translating to the dedicated struct. You can check the recent storage integrations or catalog integrations for inspiration. This is not a needed change in this PR but will be required before releasing the resource (to have a flat describe_output). We can discuss it separately.

pkg/sdk/postgres_instances_impl_gen.go
```diff
@@ -0,0 +1,214 @@
...
+func (r *ShowPostgresInstanceRequest) toOpts() *ShowPostgresInstanceOptions {
+	opts := &ShowPostgresInstanceOptions{
+		Like:       r.Like,
+		StartsWith: r.StartsWith,
+		Limit:      r.Limit,
+	}
+	return opts
+}
+
+func (r postgresInstancesRow) convert() (*PostgresInstance, error) {
```
sfc-gh-asawicki, line 182: nit: please add the comment whenever the generated code is manually edited (this is a temporary way of making our work easier; ultimately, no changes in the generated files will be allowed but we are slowly getting there).

pkg/sdk/postgres_instances_gen_test.go
```diff
@@ -0,0 +1,411 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("basic", func(t *testing.T) {
+		opts := defaultOpts()
+		assertOptsValidAndSQLEquals(t, opts, "DESCRIBE POSTGRES INSTANCE %s", id.FullyQualifiedName())
+	})
+}
+
+func Test_PostgresInstance_ToPostgresInstanceState(t *testing.T) {
```
sfc-gh-asawicki: nit: enum tests are added now with enum generation, so you can remove them after moving to the generated enum

pkg/sdk/postgres_instances_gen_test.go
```diff
@@ -0,0 +1,411 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
+	})
+
+	t.Run("validation: valid identifier for [opts.name]", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.name = emptyAccountObjectIdentifier
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("validation: exactly one field from [opts.At.Timestamp opts.At.Offset] should be present", func(t *testing.T) {
```
sfc-gh-asawicki: nit: We will later on improve the tests generation, we now add manually the test when in "exacty one" case, both options are present too

pkg/sdk/postgres_instances_gen_test.go
```diff
@@ -0,0 +1,411 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("CreatePostgresInstanceOptions.At", "Timestamp", "Offset"))
+	})
+
+	t.Run("validation: exactly one field from [opts.Before.Timestamp opts.Before.Offset] should be present", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.Before = &PostgresInstanceForkBefore{}
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("CreatePostgresInstanceOptions.Before", "Timestamp", "Offset"))
+	})
+
+	t.Run("basic", func(t *testing.T) {
```
sfc-gh-asawicki, line 36: The resulting SQL is not a correct SQL (there are missing required fields - both variants). That's why I left the commetn about stricter validation, or, preferrably, splitting into multiple variants. In the latter case, both variants need to have a basic/full test.

pkg/sdk/generator/defs/postgres_instances_def.go
sfc-gh-jcieslak, line 186: nit: For prop-based outputs, we usually prepare parser that would return struct with mapped properties (e.g. `ParseStageDetails` function in SDK).

pkg/sdk/postgres_instances_gen_test.go
sfc-gh-jcieslak: nit: usually, we add an additional case where more than one field is added to test the validation

pkg/sdk/postgres_instances_gen_test.go
```diff
@@ -0,0 +1,411 @@
...
+		opts := defaultOpts()
+		opts.Fork = &forkId
+		opts.Before = &PostgresInstanceForkBefore{
+			Timestamp: Pointer("2025-01-15 12:00:00"),
+		}
+		assertOptsValidAndSQLEquals(t, opts,
+			`CREATE POSTGRES INSTANCE %s FORK %s BEFORE (TIMESTAMP => 2025-01-15 12:00:00)`,
+			id.FullyQualifiedName(), forkId.FullyQualifiedName())
+	})
+}
```
sfc-gh-jcieslak, line 187: nit: tests could be a bit simpler with separate create operations (basic + complete per create operation)

sfc-gh-asawicki (changes requested):

pkg/sdk/postgres_instances_impl_gen.go
```diff
@@ -0,0 +1,214 @@
...
+		UpdatedOn:               r.UpdatedOn,
+		Type:                    r.Type,
+		ComputeFamily:           r.ComputeFamily,
+		AuthenticationAuthority: r.AuthenticationAuthority,
+		StorageSize:             r.StorageSize,
+		PostgresVersion:         r.PostgresVersion,
+		IsHa:                    r.IsHa == "true",
+		RetentionTime:           r.RetentionTime,
+	}
+	if r.Origin.Valid {
```
sfc-gh-jcieslak, line 197: nit: we could use map helper functions (see pkg/sdk/mapping_helpers.go and usage of functions inside like `mapNullString`

sfc-gh-jcieslak (changes requested):

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -16,7 +28,29 @@ var postgresInstancesDef = g.NewInterface(
...
+		OptionalNumberAssignment("POSTGRES_VERSION", g.ParameterOptions()).
+		OptionalTextAssignment("NETWORK_POLICY", g.ParameterOptions().SingleQuotes()).
+		OptionalBooleanAssignment("HIGH_AVAILABILITY", g.ParameterOptions()).
+		OptionalTextAssignment("STORAGE_INTEGRATION", g.ParameterOptions().SingleQuotes()).
+		OptionalTextAssignment("POSTGRES_SETTINGS", g.ParameterOptions().SingleQuotes()).
+		OptionalTextAssignment("COMMENT", g.ParameterOptions().SingleQuotes()).
+		OptionalTags().
+		WithValidation(g.ValidIdentifier, "name"),
+).CustomOperation(
+	"Fork",
```
sfc-gh-jcieslak, line 47: nit: could be more descriptive (e.g. `CreateFromFork`)

pkg/sdk/generator/defs/postgres_instances_def.go
sfc-gh-asawicki: #4674 was merged, please remove:
```
.ShowByIdOperationWithFiltering(
     g.ShowByIDLikeFiltering,
)
```
and regenerate with `make generate-sdk SF_TF_GENERATOR_ARGS='--filter-generation-part-names=default,impl --filter-object-names=PostgresInstances'` to prove that ShowByID methods are still properly visible in the interface and impl file.

It can be done in the follow-up.

sfc-gh-asawicki (dismissed):
I left only comments for the follow-up PR; closed others and approved this one.

sfc-gh-jcieslak (dismissed):

sfc-gh-asawicki (approved):

pkg/sdk/postgres_instances_enums_gen.go
sfc-gh-jsouthard: I believe this was resolved with the enum change?

pkg/sdk/generator/defs/postgres_instances_def.go
sfc-gh-jsouthard: I believe this was fixed during the creation split work.

pkg/sdk/generator/defs/postgres_instances_def.go
sfc-gh-jsouthard: Semantically, it's creating a fork from an existing cluster. "Fork" as an operations is common terminology, but CreateFork would be comparable and correct if preferred.

PR #4682, opened 2026-04-23
feat: handle external catalog source type drift

sfc-gh-kwasilewski:
Compare describe_output.0.catalog_source to the expected source type and trigger replacement through fully_qualified_name, because SDKv2 ResourceDiff only allows SetNew and ForceNew on top-level computed attributes.

Register RecreateWhenCatalogSourceChangedExternally on all four catalog integration resources.


pkg/resources/custom_diffs.go
```diff
@@ -271,17 +271,23 @@ func RecreateWhenSecretTypeChangedExternally(secretType sdk.SecretType) schema.C
...
 // RecreateWhenStreamTypeChangedExternally recreates a stream when argument streamType is different than in the state.
 func RecreateWhenStreamTypeChangedExternally(streamType sdk.StreamSourceType) schema.CustomizeDiffFunc {
-	return RecreateWhenResourceTypeChangedExternally("stream_type", streamType, sdk.ToStreamSourceType)
+	return RecreateWhenResourceTypeChangedExternally("stream_type", "stream_type", streamType, sdk.ToStreamSourceType)
+}
+
+// RecreateWhenCatalogSourceChangedExternally recreates a catalog integration when argument catalogSource is different
+// than in the state.
+func RecreateWhenCatalogSourceChangedExternally(catalogSource sdk.CatalogIntegrationCatalogSourceType) schema.CustomizeDiffFunc {
+	return RecreateWhenResourceTypeChangedExternally(DescribeOutputAttributeName+".0.catalog_source", FullyQualifiedNameAttributeName, catalogSource, sdk.ToCatalogIntegrationCatalogSourceType)
```
sfc-gh-asawicki: In other resources we've usually been adding a dedicated computed attribute with type. It's a bit cleaner to show the forced replacement based on the type attribute instead of the fully_qualified_name.

sfc-gh-asawicki (changes requested):

sfc-gh-kwasilewski:
I don't think we need a state upgrader - the new field is set in the Read function. I added an acc test (`TestAcc_CatalogIntegrationAwsGlue_NewCatalogSourceComputedField`) that demonstates the plan is empty after upgrade - let me know if we need it

pkg/testacc/resource_catalog_integration_aws_glue_acceptance_test.go
```diff
@@ -480,3 +488,44 @@ func TestAcc_CatalogIntegrationAwsGlue_ImportValidation(t *testing.T) {
...
+				Config: config.FromModels(t, provider, catalogIntegrationAwsGlue),
+			},
+			{
+				ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+				Config: config.FromModels(t, catalogIntegrationAwsGlue),
```
sfc-gh-asawicki, line 527: nit: we could have an assertion for this type being set

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

PR #4686, opened 2026-04-27
feat: postgres instance sdk integration tests

sfc-gh-jsouthard:
Change summary:

1. APPLY keyword support for ALTER SET — Added APPLY IMMEDIATELY and APPLY ON '<timestamp>' options to ALTER POSTGRES INSTANCE ... SET, allowing callers to control when settings changes take effect. This touches the generator def, SDK structs, DTOs, builders, impl, and validation (exactly-one-of Immediately/On).

  2. Type-safe ResetAccess role enum — The For field on PostgresInstanceResetAccess was changed from a raw string to a new PostgresInstanceResetAccessRole enum (snowflake_admin, application), with the standard ToPostgresInstanceResetAccessRole converter.

  3. Row-conversion cleanup — Replaced manual if r.Field.Valid null-string handling in postgresInstancesRow.convert() with mapNullString / mapStringWithMapping helpers, removing the fmt import.

  4. Extension file — New postgres_instances_ext.go adding a GetName() accessor on CreatePostgresInstanceRequest, needed by the acceptance test helper.

  5. Acceptance test helper — New PostgresInstanceClient in pkg/acceptance/helpers/ (Create, CreateWithRequest, DropFunc, Show) wired into TestClient.

  6. Integration tests — 
    * All 9 SDK interface methods (Create, Fork, Alter, Drop, DropSafely, Show, ShowByID, ShowByIDSafely, Describe)
    *  All 10 ALTER SET fields
    * All 5 ALTER UNSET fields
    * All 8 optional CREATE parameters
    * Both Fork temporal modes (At/Before) with both Timestamp/Offset
    * All SHOW filter options (Like, StartsWith, Limit, LimitFrom)
    * Both authentication authorities
    * Both reset access roles
    * Error paths for non-existing objects across all operations
    * Invalid state transitions (suspend/resume)
    * Name collision on rename
    * Field-level assertions on SHOW results including nullable fields
  
7. Unit test additions — New test cases for set with apply immediately, set with apply on timestamp, APPLY validation, and both reset access role variants.


pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -101,6 +101,14 @@ var postgresInstancesDef = g.NewInterface(
 				OptionalNumberAssignment("POSTGRES_VERSION", g.ParameterOptions()).
 				OptionalNumberAssignment("MAINTENANCE_WINDOW_START", g.ParameterOptions()).
 				OptionalTextAssignment("POSTGRES_SETTINGS", g.ParameterOptions().SingleQuotes()).
+				OptionalQueryStructField(
+					"Apply",
+					g.NewQueryStruct("PostgresInstanceApply").
+						OptionalSQL("IMMEDIATELY").
+						OptionalTextAssignment("ON", g.KeywordOptions().SingleQuotes()).
```
sfc-gh-jcieslak: KeywordOptions -> ParameterOptions

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+				WithHighAvailability(true)))
+		require.NoError(t, err)
+
+		assertThatObject(t, objectassert.PostgresInstance(t, postgresInstance.ID()).
+			HasIsHa(true),
+		)
+	})
+
+	// Doc example: ALTER POSTGRES INSTANCE my_postgres SET AUTHENTICATION_AUTHORITY = POSTGRES_OR_SNOWFLAKE
+	t.Run("alter: set authentication_authority", func(t *testing.T) {
```
sfc-gh-jcieslak: nit: set and unset for all properties could be bundled into one test case and all properties with `apply [ immediately ]` be another test case

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+		require.NoError(t, err)
+
+		// Forked instances should have an origin referencing the source
+		assert.NotNil(t, forkedInstance.Origin)
+		if forkedInstance.Origin != nil {
+			assert.Contains(t, *forkedInstance.Origin, sourceInstance.Name)
+		}
+
+		// Verify type field is populated
+		assert.NotEmpty(t, forkedInstance.Type)
```
sfc-gh-jcieslak: nit: we could assert which type is set

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+		require.NoError(t, err)
+		require.NotEmpty(t, properties)
+
+		// Verify expected properties are present per the docs output table
+		propertyMap := make(map[string]string)
+		for _, p := range properties {
+			propertyMap[p.Property] = p.Value
+		}
+
+		assert.Contains(t, propertyMap, "name")
```
sfc-gh-jcieslak: nit: there's a way to define custom fluent assertions for property-based describe output, e.g., in stages: https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/sdk/testint/stages_gen_integration_test.go#L59.

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-asawicki: +1, please bundle all settable/unsettable fields in as few invocations as possible (this is also a good test proving that altering SET/UNSET can be done with multiple attrobutes specified - we had problems with commas, lack of commas etc).

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-asawicki: +1

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-asawicki: +1, but let's remember that the dedicated method for structured describe output should be added: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4675#discussion_r3122952281

pkg/sdk/generator/defs/postgres_instances_def.go
sfc-gh-asawicki: +1

pkg/acceptance/bettertestspoc/assert/objectassert/postgres_instance_snowflake_gen.go
```diff
@@ -0,0 +1,321 @@
+package objectassert
```
sfc-gh-asawicki, line 3: this file should be generated: https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/acceptance/bettertestspoc/README.md#adding-new-snowflake-object-assertions

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+			HasAuthenticationAuthority("POSTGRES").
+			HasPostgresVersion("17").
+			HasIsHa(true).
+			HasComment(comment).
+			HasCreatedOnNotEmpty().
+			HasUpdatedOnNotEmpty(),
+		)
+	})
+
+	t.Run("create - with network_policy", func(t *testing.T) {
```
sfc-gh-asawicki: nit: why do we need this test? the network policy is covered in the complete above

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+		require.NoError(t, err)
+		propertyMap := make(map[string]string)
+		for _, p := range properties {
+			propertyMap[p.Property] = p.Value
+		}
+		assert.Equal(t, networkPolicy.Name, propertyMap["network_policy"])
+	})
+
+	// Doc example: POSTGRES_SETTINGS = '{"postgres:work_mem" = "128MB", "pgbouncer:default_pool_size" = "200"}'
+	t.Run("create - with postgres settings", func(t *testing.T) {
```
sfc-gh-asawicki: shouldn't this be a part of the "complete" above?

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+				WithPostgresSettings(true)))
+		require.NoError(t, err)
+
+		assertThatObject(t, objectassert.PostgresInstance(t, id).
+			HasNoComment(),
+		)
+	})
+
+	// Doc example: ALTER POSTGRES INSTANCE my_postgres SUSPEND;
+	t.Run("alter: suspend", func(t *testing.T) {
```
sfc-gh-asawicki: nit: merge into suspend and resume test (setup for resume does suspend first, so there is a reason to merge them)

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+		// State may be RESUMING, STARTING, CREATING, or READY depending on timing
+		assert.Contains(t, []sdk.PostgresInstanceState{
+			sdk.PostgresInstanceStateResuming,
+			sdk.PostgresInstanceStateStarting,
+			sdk.PostgresInstanceStateCreating,
+			sdk.PostgresInstanceStateReady,
+		}, result.State)
+	})
+
+	t.Run("alter: suspend already suspended instance", func(t *testing.T) {
```
sfc-gh-asawicki: add this to the suspend/resume test

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+			WithSuspend(true))
+		require.NoError(t, err)
+
+		// Suspend again - expect error due to invalid state
+		err = client.PostgresInstances.Alter(ctx, sdk.NewAlterPostgresInstanceRequest(postgresInstance.ID()).
+			WithSuspend(true))
+		assert.Error(t, err)
+	})
+
+	t.Run("alter: resume already running instance", func(t *testing.T) {
```
sfc-gh-asawicki: as above

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+		postgresInstance, cleanup := testClientHelper().PostgresInstance.Create(t)
+		t.Cleanup(cleanup)
+
+		instances, err := client.PostgresInstances.Show(ctx, sdk.NewShowPostgresInstanceRequest())
+		require.NoError(t, err)
+		require.NotEmpty(t, instances)
+
+		found := false
+		for _, inst := range instances {
+			if inst.Name == postgresInstance.Name {
```
sfc-gh-asawicki, line 542: nit: verify the fully qualified name, not a name only

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+		}
+		assert.Equal(t, "3", propertyMap["maintenance_window_start"])
+
+		// Unset
+		err = client.PostgresInstances.Alter(ctx, sdk.NewAlterPostgresInstanceRequest(postgresInstance.ID()).
+			WithUnset(*sdk.NewPostgresInstanceUnsetRequest().WithMaintenanceWindowStart(true)))
+		require.NoError(t, err)
+	})
+
+	t.Run("alter: set and unset storage_integration", func(t *testing.T) {
```
sfc-gh-asawicki: this should be a part of a single the set/unset test (if possible)

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+			WithSet(*sdk.NewPostgresInstanceSetRequest().WithStorageIntegration(storageIntegration.Name)))
+		require.NoError(t, err)
+
+		// Unset
+		err = client.PostgresInstances.Alter(ctx, sdk.NewAlterPostgresInstanceRequest(postgresInstance.ID()).
+			WithUnset(*sdk.NewPostgresInstanceUnsetRequest().WithStorageIntegration(true)))
+		require.NoError(t, err)
+	})
+
+	t.Run("alter: set and unset network_policy", func(t *testing.T) {
```
sfc-gh-asawicki: this should be a part of a single the set/unset test (if possible)

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -0,0 +1,1126 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert/objectassert"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers/random"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/testenvs"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeroles"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInt_PostgresInstances(t *testing.T) {
```
sfc-gh-asawicki, line 17: please make the following changes:
1. Order the methods, so that create/alter/show are placed next to each other - curently, there is a mix of tests throughout the file
2. Bundle the tests into single ones (e.g. resume+suspend with options in one test, all set/unset in one test)
3. Remove the create with tests that are parts of complete setup.

sfc-gh-asawicki (dismissed):

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-jsouthard: I had created a set of tests that were a 1:1 mapping between the test and the documentation examples as a starting point. I can certainly remove it as a case being covered by a broader test, but I appreciated having explicit coverage of documentation examples.
I'll collapse the test cases down to reduce redundancy.

sfc-gh-asawicki: In other objects, we don't follow the documentation examples (which can change and evolve in time), but we cover all the use cases we find to be useful from the provider side in the similar way: create with no optional, create with all optionals, set/unset everything, handle parameters, handle special operatiions (e.g. suspend/resume), tag handling, show variants, etc. These tests will be pretty extensive and should cover majority (if not all) these use cases.

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -538,53 +609,64 @@ func TestInt_PostgresInstances(t *testing.T) {
...
+		result, err := client.PostgresInstances.ShowByID(ctx, postgresInstance.ID())
 		require.NoError(t, err)
-		require.LessOrEqual(t, len(instances), 1)
 
-		_ = postgresInstance
+		assertThatObject(t, objectassert.PostgresInstanceFromObject(t, result).
+			HasName(postgresInstance.Name).
+			HasOwner(snowflakeroles.Accountadmin.Name()).
+			HasOwnerRoleType("ROLE").
+			HasType("PRIMARY").
```
sfc-gh-jcieslak, line 571: nit: If all types are known this could be an enum

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -76,76 +82,221 @@ func TestInt_PostgresInstances(t *testing.T) {
...
+		assert.Error(t, err)
 	})
 
-	// Doc example: ALTER POSTGRES INSTANCE my_postgres SET COMMENT = '...'
-	t.Run("alter: set and unset comment", func(t *testing.T) {
+	// ==================
+	// Alter
+	// ==================
+
+	t.Run("alter: set and unset properties", func(t *testing.T) {
```
sfc-gh-jcieslak, line 293: This kind of test could also bundle all properties into single set and unset. In the resource we would be doing potentitally multiple properties in both set and unset and this set should ensure both are possible with the whole set of `set` and `unset` properties (sometimes we can see in other objects that some properties cannot be set together because of some internal validation/constraints).

Generally, we want to test the extremes (full set of properties in set, basic in create with no props, full in create) rather than in-betweens (e.g. like in this case test for each property).

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-jcieslak: to be applied (or I cannot see this test)

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-jcieslak: to be applied (or I cannot see this test)

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-jcieslak: 1. Seems to be done
2. Tests could be bundled a bit more
3. Could `complete` test for create contain more properties like tags, storage_integration, etc.? Then we could remove tests like `create - with storage_integration` which test only singular options in create operation, which would be then repetitive with `complete` test and could be removed.

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -593,13 +675,55 @@ func TestInt_PostgresInstances(t *testing.T) {
...
+		assert.Contains(t, propertyMap, "updated_on")
+		assert.Contains(t, propertyMap, "type")
+		assert.Contains(t, propertyMap, "host")
+		assert.Contains(t, propertyMap, "compute_family")
+		assert.Contains(t, propertyMap, "storage_size_gb")
+		assert.Contains(t, propertyMap, "postgres_version")
+		assert.Contains(t, propertyMap, "high_availability")
+		assert.Contains(t, propertyMap, "authentication_authority")
+		assert.Contains(t, propertyMap, "state")
+		assert.Contains(t, propertyMap, "retention_time")
```
sfc-gh-jcieslak, line 650: We should assert for values, not existance (at least for the fields we know value of, `created_on` and similar fields can stay as they are right now). Additionally for basic and complete cases (and for fork tests) in create we could add describe assertions to validate the values are populating as we expect. You can check other tests, assertions are usually defined as helper functions on top, e.g.:
```go
func Test_XYZ(t *testing.T) {
  assertDesc := func(t *testing.T, arguments...) {
    t.Helper()
    // bundled assertions...
  }
}
```

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-jcieslak: This has to be still addressed

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-jcieslak: Then as it's a step before PR with resource changes, we should apply this in this PR

pkg/acceptance/bettertestspoc/assert/objectassert/postgres_instance_snowflake_gen.go
sfc-gh-jsouthard: This is now generated, can this be resolved?

pkg/sdk/generator/gen/templates/enum.tmpl
```diff
@@ -32,7 +32,11 @@ var All{{ .NamePlural }}String = []string{
 }
 {{ end }}
 func To{{ .Name }}(s string) ({{ .Name }}, error) {
-	s = strings.ToUpper(s)
+  {{- if .IsLowercase }}
```
sfc-gh-jcieslak, line 35: It seems the postgres instance enum implementation wasn't regenerated to include those changes

pkg/sdk/generator/gen/templates/enum.tmpl
```diff
@@ -32,7 +32,11 @@ var All{{ .NamePlural }}String = []string{
...
+  s = strings.ToLower(s)
+  {{- else }}
+  s = strings.ToUpper(s)
+  {{- end }}
```
sfc-gh-jcieslak, line 39: nit: For the team to be applied later on, I think the generation in this case should normalize both ends so that casing doesn't matter in the comparison. The correct enum value casing would be encoded in the enum value itself.

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -293,32 +291,34 @@ func TestInt_PostgresInstances(t *testing.T) {
...
-		require.NoError(t, err)
-
-		assertThatObject(t, objectassert.PostgresInstance(t, postgresInstance.ID()).
-			HasComment(comment),
-		)
-
-		err = client.PostgresInstances.Alter(ctx, sdk.NewAlterPostgresInstanceRequest(postgresInstance.ID()).
-			WithUnset(*sdk.NewPostgresInstanceUnsetRequest().WithComment(true)))
+			WithSet(*sdk.NewPostgresInstanceSetRequest().
+				WithComment(comment).
```
sfc-gh-jcieslak, line 304: We could still add storage integration and PostgreSQL version. Can be applied in a follow-up

sfc-gh-jcieslak (approved):

PR #4688, opened 2026-04-27
chore: Add instance method invocation to SDK builder, generator, and add a budget PoC

sfc-gh-asawicki:
This PoC introduces a mechanism for instance method invocation within the SQL builder and SDK generator.

It focuses on:
- providing SQL builder extensions supporting `!` syntax and `SYSTEM$REFERENCE`
- providing helpers to the SDK generator DSL to simplify instance methods creation (e.g. budgets have ~40 instance methods - https://docs.snowflake.com/en/sql-reference/classes/budget#budget-methods)
- provide a simple example
- provide a base for the budget implementation with integration tests proving this functionality in action

Follow-ups:
- Finish integration tests (two left with TODO)
- Implement all budget instance methods
- Solve added TODOs/reassign them


pkg/sdk/generator/defs/budgets_def.go
```diff
@@ -0,0 +1,96 @@
+//go:build sdk_generation
```
sfc-gh-jmichalak: Hmm, I saw this once before, we don't use such tag besides password policies. Is it just a hallucination?

pkg/sdk/generator/example/sdk_definitions.go
```diff
@@ -63,6 +63,10 @@ func randomDatabaseObjectIdentifier() DatabaseObjectIdentifier {
 	return DatabaseObjectIdentifier{}
 }
 
+func randomAccountObjectIdentifier() AccountObjectIdentifier {
```
sfc-gh-jmichalak, line 66:
```suggestion
func emptyAccountObjectIdentifier() AccountObjectIdentifier {
```

Nit, similar for others.

pkg/sdk/generator/defs/budgets_def.go
```diff
@@ -0,0 +1,96 @@
...
+		WithValidation(g.ConflictingFields, "OrReplace", "IfNotExists"),
+).DropOperation(
+	"https://docs.snowflake.com/en/sql-reference/classes/budget/commands/drop-budget",
+	g.NewQueryStruct("DropBudgetOptions").
+		Drop().
+		SQLWithCustomFieldName("snowflakeCoreBudget", "SNOWFLAKE.CORE.BUDGET").
+		IfExists().
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).InstanceMethodOperationScalar(
```
sfc-gh-jmichalak, line 63: Nit: for future, let's have this list sorted alphabetically for easier reviews.

pkg/sdk/generator/defs/budgets_def.go
```diff
@@ -0,0 +1,96 @@
...
+var setCycleStartActionArgs = g.NewQueryStruct("SetCycleStartActionArgs").
+	Identifier("Procedure", g.KindOfT[sdkcommons.SchemaObjectIdentifier](), g.IdentifierOptions().Required().SystemReference("PROCEDURE")).
+	List("Arguments", "string", g.ListOptions().Required())
+
+var getCycleStartActionResult = g.StructPair(
+	"getCycleStartActionRow",
+	"BudgetCycleStartAction",
+).Text("action_uuid").
+	Text("procedure_fqn").
+	Text("procedure_args").
```
sfc-gh-jmichalak: Q: According to the docs, it's an array of arguments. Why is it Text?

pkg/sdk/generator/defs/budgets_def.go
```diff
@@ -0,0 +1,96 @@
...
+		Drop().
+		SQLWithCustomFieldName("snowflakeCoreBudget", "SNOWFLAKE.CORE.BUDGET").
+		IfExists().
+		Name().
+		WithValidation(g.ValidIdentifier, "name"),
+).InstanceMethodOperationScalar(
+	"https://docs.snowflake.com/en/sql-reference/classes/budget/method/set_spending_limit",
+	"SET_SPENDING_LIMIT",
+	setSpendingLimitArgs,
+	"string",
```
sfc-gh-jmichalak, line 67: Why not StructPair? We could return the parsed values.

pkg/sdk/generator/defs/budgets_def.go
```diff
@@ -0,0 +1,96 @@
...
+
+var getNotificationIntegrationsResult = g.StructPair(
+	"getNotificationIntegrationsRow",
+	"BudgetNotificationIntegration",
+).Text("integration_name").
+	Number("last_notification_time").
+	Time("added_date")
+
+var setCycleStartActionArgs = g.NewQueryStruct("SetCycleStartActionArgs").
+	Identifier("Procedure", g.KindOfT[sdkcommons.SchemaObjectIdentifier](), g.IdentifierOptions().Required().SystemReference("PROCEDURE")).
```
sfc-gh-jmichalak, line 27: Nit: let's have comments that reference_statement and array_construct_statement are redundant from our POV (I assume).

pkg/sdk/generator/gen/field_transformers.go
```diff
@@ -271,12 +272,22 @@ func (v *IdentifierTransformer) Equals() *IdentifierTransformer {
...
+
 func (v *IdentifierTransformer) Transform(f *Field) *Field {
 	addTagIfMissing(f.Tags, "ddl", "identifier")
 	if v.required {
 		f.Required = true
 	}
-	addTagIfMissing(f.Tags, "sql", v.sqlPrefix)
+	if v.systemReference != "" {
+		addTagIfMissing(f.Tags, "ddl", "system_reference")
+		addTagIfMissing(f.Tags, "sql", v.systemReference)
```
sfc-gh-jmichalak, line 287: Nit: let's have a validation for setting the prefix and system reference at the same time.

pkg/sdk/generator/gen/operation.go
```diff
@@ -27,6 +27,14 @@ const (
 	ShowMappingKindSlice       ShowMappingKind = "slice"
 )
 
+type InstanceMethodKind string
+
+const (
+	InstanceMethodKindSingleValue InstanceMethodKind = "single_value"
+	InstanceMethodKindSlice       InstanceMethodKind = "slice"
+	InstanceMethodKindScalar      InstanceMethodKind = "scalar"
```
sfc-gh-jmichalak: This is not used anywhere.

sfc-gh-jmichalak (changes requested):

pkg/sdk/generator/defs/budgets_def.go
sfc-gh-asawicki: Missed that, removed.

pkg/sdk/generator/example/sdk_definitions.go
sfc-gh-asawicki: It's correct. The unit tests generate invocation to random*Identifier like randomAccountObjectIdentifier, so we need it defined. No changes.

pkg/sdk/generator/defs/budgets_def.go
sfc-gh-asawicki: I thought about having them in the same order as in the docs (https://docs.snowflake.com/en/sql-reference/classes/budget#budget-methods) and they appear to be alphabetic. I will adjust during addition of missing methods.

pkg/sdk/generator/defs/budgets_def.go
sfc-gh-asawicki: Changed to StringList

pkg/sdk/generator/defs/budgets_def.go
sfc-gh-asawicki: It does not return the tabular (columnar) output, this is the scalar return type

pkg/sdk/generator/defs/budgets_def.go
sfc-gh-asawicki: Yes. Added a comment.

pkg/sdk/generator/gen/field_transformers.go
sfc-gh-asawicki: Not sure what you mean. We currently do not validate the constructed defs, it's an improvement for the future.

pkg/sdk/generator/gen/operation.go
sfc-gh-asawicki: Removing for now.

pkg/sdk/generator/defs/budgets_def.go
sfc-gh-jmichalak: Yes. My point is that SF returns string, and we could extract the value from there and return an int from the SDK. It would be a nice improvement to have separate types for that, especially since we already do this for SHOW/DESC.

pkg/sdk/generator/example/sdk_definitions.go
sfc-gh-jmichalak: Hmm, it's a symptom of spreading testing helpers across multiple packages. Even more, we have separate methods for acceptance tests, duplicating the logic.

sfc-gh-jmichalak (approved):

pkg/sdk/generator/defs/budgets_def.go
sfc-gh-asawicki: The problem with these methods is that it is not purely an int response. In this case, it's `The spending limit has been updated to <n> credits.`. Sure, we could parse it, but:
- we don't have control over how this output may change in the future (it won't be a BCR like when the tabular output changes)
- we don't really get much from parsing it and keeping an int, since on the Terraform layer we still depend on the read step, which in this case would invoke GET_SPENDING_LIMIT, which in turn returns an int without any additional text.

We can further discuss it next week.

PR #4702, opened 2026-05-05
feat: Add Cortex agents to SDK

sfc-gh-kwasilewski:
## Changes
- Add definition and generate SDK code.
- Limitations:
  - UNSET COMMENT not supported - we'll use SET COMMENT = '' instead.
  - UNSET PROFILE not supported - we'll use SET PROFILE = '{}' instead (as SET PROFILE = '' doesn't work).


pkg/sdk/cortex_agents_impl_gen.go
```diff
@@ -0,0 +1,160 @@
...
+	return &CortexAgentDetails{
+		Name:               r.Name,
+		DatabaseName:       r.DatabaseName,
+		SchemaName:         r.SchemaName,
+		Owner:              r.Owner,
+		Comment:            r.Comment,
+		Profile:            r.Profile,
+		AgentSpec:          r.AgentSpec,
+		CreatedOn:          r.CreatedOn,
+		DefaultVersionName: r.DefaultVersionName,
```
sfc-gh-jcieslak: nit / question: Those three fields are undocumented, did you check if they're always present (doesn't contain any conditional that would remove them from output)?

sfc-gh-jcieslak (dismissed):

pkg/sdk/cortex_agents_impl_gen.go
```diff
@@ -144,17 +149,28 @@ func (r *DescribeCortexAgentRequest) toOpts() *DescribeCortexAgentOptions {
...
-	}, nil
+	details := &CortexAgentDetails{
+		Name:         r.Name,
+		DatabaseName: r.DatabaseName,
+		SchemaName:   r.SchemaName,
+		Owner:        r.Owner,
+		AgentSpec:    r.AgentSpec,
+		CreatedOn:    r.CreatedOn,
+	}
+	if r.Comment.Valid {
```
sfc-gh-jcieslak, line 160: nit: we could use mapping functions (e.g. mapNullString)

sfc-gh-jcieslak (approved):

PR #4704, opened 2026-05-05
fix: integration test improvements found through localized testing

sfc-gh-jsouthard:
All changes are focused on fixing Postgres instance integration tests: correcting an invalid compute family identifier (STANDARD_1 → STANDARD_M), resolving test failures found during manual runs, and adding timing/wait logic to handle asynchronous operations in the integration test suite.


pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -159,16 +171,45 @@ func TestInt_PostgresInstances(t *testing.T) {
 	t.Run("fork - basic", func(t *testing.T) {
 		sourceInstance, sourceCleanup := testClientHelper().PostgresInstance.Create(t)
 		t.Cleanup(sourceCleanup)
+		testClientHelper().PostgresInstance.WaitForReady(t, sourceInstance.ID(), 5*time.Minute)
```
sfc-gh-jcieslak: This could be replaced by testify function like `(assert / require).Eventually`. We can use it within the WaitForReady helper as I can see we use it in multiple places

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -159,16 +171,45 @@ func TestInt_PostgresInstances(t *testing.T) {
...
 
 		forkId := testClientHelper().Ids.RandomAccountObjectIdentifier()
 		request := sdk.NewForkPostgresInstanceRequest(forkId, sourceInstance.ID())
 
-		err := client.PostgresInstances.Fork(ctx, request)
+		// Retry fork operation — instance may not be internally fork-ready despite being in READY state
+		var err error
+		deadline := time.Now().Add(2 * time.Minute)
+		for {
+			err = client.PostgresInstances.Fork(ctx, request)
```
sfc-gh-jcieslak: Can't we somehow ensure the state when a fork can be created instead of blindly trying to fork it until we are able to? If not, let's at least use `(assert / require).Eventually`

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -159,16 +171,45 @@ func TestInt_PostgresInstances(t *testing.T) {
...
 
-		forkedInstance, err := client.PostgresInstances.ShowByID(ctx, forkId)
-		require.NoError(t, err)
+		// Wait for fork to reach READY state — metadata (origin, type) populates after provisioning completes
+		testClientHelper().PostgresInstance.WaitForReady(t, forkId, 5*time.Minute)
+
+		// Poll until origin and type metadata are populated (fail if not received within timeout)
+		var forkedInstance *sdk.PostgresInstance
+		metadataDeadline := time.Now().Add(5 * time.Minute)
+		for {
```
sfc-gh-jcieslak: also the use case for assert / require Eventually

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -291,30 +333,94 @@ func TestInt_PostgresInstances(t *testing.T) {
...
+			}
+			if time.Now().After(deadline) {
+				break
+			}
+			time.Sleep(5 * time.Second)
+		}
+		require.NoError(t, err)
+
+		// Wait for the instance to return to READY after compute/storage changes complete
+		testClientHelper().PostgresInstance.WaitForReady(t, postgresInstance.ID(), 10*time.Minute)
```
sfc-gh-jcieslak: I think a 10-minute timeout is a bit too generous, or is it not? I don't think we had any object that required this much of a timeout. We need to find the least timeout that we know would work. Unless you have already chosen the least amount, then we need to discuss internally what do we want to do with those tests (2-minute timeout is also pretty big as it may greatly contribute to overall test run times, which are already too high for our liking)

sfc-gh-jsouthard: Agreed, I ran some additional tests to evaluate the sensitivity of this time value and have reduced it for the update.

This test ends up being long in aggregate because of the settings that cannot be set concurrently and needing to wait for confirmation.

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-jsouthard: Unfortunately, there is not currently a method for knowing when a cluster is fork-ready after the `READY` state transition.

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -19,6 +20,86 @@ func TestInt_PostgresInstances(t *testing.T) {
...
 	ctx := testContext(t)
 
+	// Pre-provision shared instances in parallel to minimize total wait time.
+	// sharedInstance: used by read-only tests (show, describe, ShowByID) and as fork source.
+	// alterInstance: used by "alter: set and unset properties" (destructive).
+	// suspendInstance: used by "alter: suspend and resume" (destructive).
+	// resetAccessInstance: used by "alter: reset access" (destructive).
+	// forkSourceInstance: used exclusively by "fork - basic" — forked during setup so the fork is
+	//   already provisioning by the time the test runs, saving ~8 minutes of in-test waiting.
+	type provisionedInstance struct {
```
sfc-gh-jcieslak, line 30: Usually, we utilize a test step file for predefining objects (see https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/sdk/testint/setup_test.go#L113), but as those objects (at least a few of them) are meant to be altered, maybe it would make sense to make those objects more local. Even if the object creation were local, I would still pull it out of TestInt_PostgresInstances, since the object precreation only happens when we enter the test. It's not ideal, as we still need to wait for the objects to be created (on the other hand, it should already be a significant gain, as it happens in parallel). In this case, I think we could go with either:
- Extending the existing (above linked) `setup_test` with new predefined PostgreSQL instance objects and using them here
- Pull this code above into the `init` function that would create object instances whenever the `testint` package is tested

In both approaches, an additional parameter should be introduced to steer the object pre-creation. By default, those objects should not be created, as we don't want to create PostgreSQL instances when we run integration tests for other objects. You can see the `testenvs` package and how it's used in the initialization function: https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/sdk/testint/setup_test.go#L113. We could enable this parameter in our GH [pipelines](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/.github/workflows/tests.yml#L126) or in [Makefile for test-integration](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/Makefile#L91) recipe (I think the latter would be better as there were some plans to split the recipes used by devs and pipelines anyway).

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -169,59 +250,32 @@ func TestInt_PostgresInstances(t *testing.T) {
...
-		require.NoError(t, err)
-		t.Cleanup(testClientHelper().PostgresInstance.DropFunc(t, forkId))
+		// Uses the pre-forked instance created during setup to avoid ~8 minutes of in-test waiting.
+		// The fork was initiated in parallel with other instance provisioning.
+		forkWg.Wait() // Ensure the fork goroutine has completed its Fork call
+		forkedId := preForkedInstance.instance.ID()
 
 		// Wait for fork to reach READY state — metadata (origin, type) populates after provisioning completes
-		testClientHelper().PostgresInstance.WaitForReady(t, forkId, 5*time.Minute)
+		testClientHelper().PostgresInstance.WaitForReady(t, forkedId, 5*time.Minute)
```
sfc-gh-jcieslak, line 260: By pulling the object precreation a level above as mentioned in the other comment, we would be able to get rid of such asserts, or just make them smaller (in seconds, just in case, rather than minutes).

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-asawicki: Let's see how long these take, and if they dont run crazy long, it's enough to merge them in the current shape. We will discuss the endgame, but our team can handle it later.

sfc-gh-jsouthard:
Unless I'm misreading the test results, I didn't see any failures tied to the postgres integration tests.

sfc-gh-asawicki (approved):

PR #4707, opened 2026-05-06
feat: add SDK integration tests for Cortex agents

sfc-gh-kwasilewski:
## Changes
- Integration tests exercise create, alter (SET + modify live version), drop, filtered show, and describe for Cortex agents.
- Helpers decode profile and agent_spec JSON into structs and maps for easier validation.
- Generator definition updates:
  - ALTER SET now separates SET properties with commas; omitting them caused Snowflake compilation errors.
  - COMMENT uses StringAllowEmpty.


pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
```diff
@@ -271,6 +271,15 @@ var allStructs = []SdkObjectDef{
 		IdType:       "sdk.AccountObjectIdentifier",
 		ObjectStruct: sdk.PostgresInstance{},
 	},
+	{
```
sfc-gh-jmichalak, line 274: I suppose this whole change was not gofmt-ed.

pkg/acceptance/bettertestspoc/assert/objectassert/cortex_agent_snowflake_ext.go
```diff
@@ -0,0 +1,72 @@
...
+		t.Helper()
+		if o.CreatedOn == (time.Time{}) {
+			return fmt.Errorf("expected created_on to be not empty")
+		}
+		return nil
+	})
+	return c
+}
+
+func (c *CortexAgentAssert) HasCortexAgentProfile(expected *sdk.CortexAgentProfile) *CortexAgentAssert {
```
sfc-gh-jmichalak: Nit: I think we should take a struct instead of a pointer here, and in assertCortexAgentProfileJsonEqual, and the same for spec map handling. To assert nil (if we need this), let's have a separate HasNo... To be consistent with other objects/fields.

pkg/sdk/cortex_agents_ext.go
```diff
@@ -0,0 +1,40 @@
...
+	if err := json.Unmarshal([]byte(profileAsJson), &profile); err != nil {
+		return nil, err
+	}
+
+	return &profile, nil
+}
+
+// UnmarshalCortexAgentSpec parses agent specification JSON (as returned by DESCRIBE AGENT) into a map.
+func UnmarshalCortexAgentSpec(agentSpec string) (map[string]any, error) {
+	data := strings.TrimSpace(agentSpec)
```
sfc-gh-jmichalak, line 31: Nit: I don't think this is necessary.

pkg/sdk/testint/cortex_agents_integration_test.go
```diff
@@ -0,0 +1,283 @@
...
+					"tokens":  float64(16000),
+				},
+			},
+			"instructions": map[string]any{
+				"response": response,
+			},
+		}
+	}
+
+	cleanupCortexAgentFunc := func(id sdk.SchemaObjectIdentifier) func() {
```
sfc-gh-jmichalak: Nit: Instead of this helper method, we should use the helper client's method. Similarly for specForResponse, it should be hidden there. We already do this for service object.

pkg/sdk/testint/cortex_agents_integration_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInt_CortexAgents(t *testing.T) {
+	client := testClient(t)
+	ctx := testContext(t)
+
+	specForResponse := func(response string) string {
+		return "orchestration:\n" +
```
sfc-gh-jmichalak: Nit: you can use multiline Go strings with fmt.Sprintf.

sfc-gh-jmichalak (dismissed):

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
sfc-gh-kwasilewski: Both `make lint` and `make fmt-check` return no errors.

pkg/sdk/cortex_agents_ext.go
sfc-gh-kwasilewski: It is. The "whitespace only" test case fails with `unexpected end of JSON input` if I remove it.

pkg/acceptance/bettertestspoc/assert/objectassert/cortex_agent_snowflake_ext.go
sfc-gh-kwasilewski: Done for profile. `HasCortexAgentSpec` didn't use a pointer.

pkg/sdk/testint/cortex_agents_integration_test.go
sfc-gh-kwasilewski: Done

pkg/sdk/testint/cortex_agents_integration_test.go
sfc-gh-kwasilewski: Removed

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
sfc-gh-jmichalak: Hmm, it looks like it's a rendering issue in Graphite, as it looks fine in GH.

pkg/sdk/cortex_agents_ext.go
sfc-gh-jmichalak: It shouldn't ([example](https://go.dev/play/p/KXncWrqJoeN)), but hard to say without specifics. We can leave this.

sfc-gh-jmichalak (approved):

pkg/sdk/cortex_agents_ext.go
sfc-gh-kwasilewski: It will fail if you remove empty brackets from the string in your example.

sfc-gh-jmichalak: Yes, but also fails when this string is empty, so we have a workaround here. It would be great if SF returned a valid JSON.

pkg/sdk/cortex_agents_ext_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func TestUnmarshalCortexAgentProfile(t *testing.T) {
+	validCases := []struct {
+		name     string
+		json     string
+		expected *CortexAgentProfile
+	}{
+		{
+			name: "all fields present",
+			json: `{"display_name":"My Assistant","avatar":"assistant.png","color":"blue"}`,
```
sfc-gh-asawicki, line 18: Q: what about nested lists, fields, etc? Will this impl work correctly?

sfc-gh-asawicki: Are there any in either cortex agent or mcp server?

sfc-gh-asawicki (approved):

pkg/sdk/cortex_agents_ext_test.go
sfc-gh-kwasilewski: No more fields except those I listed. The structure is well-described in the docs: https://docs.snowflake.com/en/sql-reference/sql/create-agent#optional-parameters. 

If a new field is ever added, the implementation will still work, see: https://go.dev/play/p/smKIHaAW0VI

sfc-gh-asawicki: Okay for the profile but for the spec, I see that e.g. list of sample_questions is possible, and the next test covers only the nested object. Can we add a list too?

sfc-gh-kwasilewski: Sure, let me do this in a follow-up.

PR #4712, opened 2026-05-07
feat: implement resources for snowflake postgres instance and snowflake postgres fork

sfc-gh-jsouthard:
Initial implementation, acceptance testing modified with some fixes used in SDK integration testing, but test timing remains an issue given the provisioning time required.


examples/resources/snowflake_postgres_fork/resource.tf
```diff
@@ -0,0 +1,23 @@
...
+resource "snowflake_postgres_fork" "basic" {
+  name      = "my_postgres_fork"
+  fork_from = "my_source_postgres_instance"
+}
+
+# fork with point-in-time recovery
+resource "snowflake_postgres_fork" "pitr" {
+  name         = "my_postgres_fork_pitr"
+  fork_from    = "my_source_postgres_instance"
+  at_timestamp = "2025-01-15 12:00:00"
```
sfc-gh-jcieslak, line 11: check how handled in other cases

pkg/acceptance/bettertestspoc/assert/objectassert/postgres_instance_desc_snowflake_gen.go
```diff
@@ -0,0 +1,281 @@
+// Code generated by object assertions generator (v0.1.0); DO NOT EDIT.
```
sfc-gh-jcieslak, line 1: As the model seems outdated or not generated with our generator, please make sure this file is properly generated. Here's more info on generator: https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/acceptance/bettertestspoc/README.md

pkg/acceptance/bettertestspoc/config/model/postgres_instance_model_gen.go
```diff
@@ -0,0 +1,216 @@
+// Code generated by resource model builder generator (v0.1.0); DO NOT EDIT.
```
sfc-gh-jcieslak, line 1: It seems this file was not generated with our generator. Please make sure all _gen.go files are properly generated with our generator commands: https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/acceptance/bettertestspoc/README.md.

pkg/acceptance/bettertestspoc/config/model/postgres_fork_model_ext.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	tfconfig "github.com/hashicorp/terraform-plugin-testing/config"
+)
+
+type PostgresForkModel struct {
```
sfc-gh-jcieslak: Why wasn't it generated?

pkg/provider/previewfeatures/preview_features.go
```diff
@@ -78,6 +78,8 @@ const (
 	PasswordPolicyResource                        feature = "snowflake_password_policy_resource"
 	PipeResource                                  feature = "snowflake_pipe_resource"
 	PipesDatasource                               feature = "snowflake_pipes_datasource"
+	PostgresForkResource                          feature = "snowflake_postgres_fork_resource"
```
sfc-gh-jcieslak, line 81: we always update preview features tests when adding new entries here, see: https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/pkg/provider/previewfeatures/preview_features_test.go

MIGRATION_GUIDE.md
```diff
@@ -70,6 +70,14 @@ We have added a new preview data source for password policies: [snowflake_passwo
 
 This feature will be marked as stable in future releases. To use it, add `snowflake_password_policies_datasource` to the `preview_features_enabled` field in the provider configuration.
 
+## v2.15.x ➞ v2.16.0
```
sfc-gh-jcieslak, line 73: we're currently at 2.16.0, so this needs to be updated

MIGRATION_GUIDE.md
```diff
@@ -70,6 +70,14 @@ We have added a new preview data source for password policies: [snowflake_passwo
...
+
+### *(new feature)* Postgres instance preview feature
```
sfc-gh-jcieslak, line 75: You forgot to mention the fork resource, which requires another preview_feature flag to be enabled. See other migration guide entries with different resource "versions" for inspiration.

pkg/acceptance/bettertestspoc/assert/resourceassert/postgres_fork_resource_ext.go
```diff
@@ -0,0 +1,117 @@
+package resourceassert
+
+import (
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+)
+
+type PostgresForkResourceAssert struct {
```
sfc-gh-jcieslak: This file should also be generated

pkg/acceptance/bettertestspoc/assert/resourceassert/postgres_instance_resource_ext.go
```diff
@@ -0,0 +1 @@
+package resourceassert
```
sfc-gh-jcieslak, line 1: This file is empty

pkg/acceptance/bettertestspoc/assert/resourceassert/postgres_instance_resource_gen.go
```diff
@@ -0,0 +1,314 @@
+// Code generated by resource assertions generator (v0.1.0); DO NOT EDIT.
+
+package resourceassert
```
sfc-gh-jcieslak, line 3: As with other files, make sure it's properly generated (re-generations in this case should produce no diffs)

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+var postgresInstanceSchema = map[string]*schema.Schema{
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      blocklistedCharactersFieldDescription("Specifies the identifier for the Postgres instance; must be unique for your account."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
```
sfc-gh-jcieslak, line 26: For identifier fields we usually add `IsValidIdentifier` validator

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+	},
+	"compute_family": {
+		Type:        schema.TypeString,
+		Required:    true,
+		Description: "Specifies the compute family for the Postgres instance (e.g. STANDARD_M).",
```
sfc-gh-jcieslak, line 31: It is an enum, for those we add validator and normalization function, e.g.: https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/resources/warehouse.go#L41

Edit: Now, I can see you did the similar thing in authentication_authority (the same should be done here)

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+		Type:             schema.TypeString,
+		Required:         true,
+		ValidateDiagFunc: sdkValidation(sdk.ToPostgresInstanceAuthenticationAuthority),
+		DiffSuppressFunc: NormalizeAndCompare(sdk.ToPostgresInstanceAuthenticationAuthority),
+		Description:      fmt.Sprintf("Specifies the authentication authority for the Postgres instance. Valid values are (case-insensitive): %s.", possibleValuesListed(sdk.AllPostgresInstanceAuthenticationAuthorities)),
+	},
+	"postgres_version": {
+		Type:             schema.TypeInt,
+		Optional:         true,
+		Computed:         true,
```
sfc-gh-jcieslak, line 49: This shouldn't be computed. To suppress any changes that would appear, when user didn't specify this field, you should follow this pattern: https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/resources/warehouse.go#L56. IgnoreChangeToCurrentSnowflakeValueInShow uses show_output that is computed value in every new resource (it's basically a Snowflake output cache), which I can see you added at the bottom of the schema.

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+		ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(1)),
+		Description:      "Specifies the Postgres version for the instance.",
+	},
+	"network_policy": {
+		Type:        schema.TypeString,
+		Optional:    true,
```
sfc-gh-jcieslak, line 55: validator for identifier (`IsValidIdentifier`) and also diff suppress for quoting and potential value when it's not set in config `DiffSuppressFunc: SuppressIfAny(suppressIdentifierQuoting, IgnoreChangeToCurrentSnowflakeValueInShow("network_policy"))` (for more than one we use `SuppressIfAny` wrapper)

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+		Description: "Specifies the network policy to associate with the Postgres instance.",
+	},
+	"high_availability": {
+		Type:        schema.TypeBool,
```
sfc-gh-jcieslak, line 59: For bool values we use TypeString, e.g.: https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/resources/warehouse.go#L74. More info about it in: https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/v1-preparations/CHANGES_BEFORE_V1.md#empty-values

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+		Optional:    true,
+		Description: "Specifies whether the Postgres instance should be configured for high availability.",
+	},
+	"storage_integration": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies the storage integration for the Postgres instance.",
```
sfc-gh-jcieslak, line 66: Identifier validation and normalization as in other cases

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+		setOptionalFromPtr(d, "postgres_settings", postgresSettings),
+		d.Set("postgres_version", details.PostgresVersion),
+		setOptionalFromNonEmptyStringPtr(d, "network_policy", details.NetworkPolicy),
+		setOptionalFromNonEmptyStringPtr(d, "storage_integration", details.StorageIntegration),
+	)
+	if errs != nil {
+		return nil, errs
+	}
+
+	if details.MaintenanceWindowStart != 0 {
```
sfc-gh-jcieslak, line 166: I'm doubting if this import function is needed. I think we could use only `ImportName` helper function as in other cases.

And regarding `maintenance_window_start`, because of the [empty values issue](https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/v1-preparations/CHANGES_BEFORE_V1.md#empty-values) I think this field should be of TypeString, because 0 is a valid value. In case when Snowflake default is set e.g. to 6 underneath, we would be able to set 0. By having this check and having this field of TypeInt, we are not able to distinguish if 0 means set or not set. I'm talking here about the resource field.

On the sdk level, imo, this field should be nullable int as value `None` returned from Snowflake mean we didn't set this value and we rely on some default (I didn't find what's the default).

I would like to hear other team members' opinions about it, @snowflakedb/terraform-provider

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+
+func ReadPostgresInstanceFunc(withExternalChangesMarking bool) schema.ReadContextFunc {
+	return func(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+		client := meta.(*provider.Context).Client
+		id, err := sdk.ParseAccountObjectIdentifier(d.Id())
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		pi, err := client.PostgresInstances.ShowByIDSafely(ctx, id)
```
sfc-gh-jcieslak, line 223: Question: would this work? In integration tests reading after creation could fail due to the missing metadata, right? In this case we need to use retry functionality as in `account.go` resource. But in this case, I think the retry won't be in create but here as alter operations may also propagate with the delay. We may introduce a param that would skip those retries as you previously suggested, but let's start simple, and make it work

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		if withExternalChangesMarking {
+			if err = handleExternalChangesToObjectInShow(d,
+				outputMapping{"compute_family", "compute_family", pi.ComputeFamily, pi.ComputeFamily, nil},
+				outputMapping{"storage_size", "storage_size_gb", pi.StorageSize, pi.StorageSize, nil},
+				outputMapping{"authentication_authority", "authentication_authority", pi.AuthenticationAuthority, pi.AuthenticationAuthority, nil},
+				outputMapping{"is_ha", "high_availability", pi.IsHa, pi.IsHa, nil},
```
sfc-gh-jcieslak, line 248: We prefer explicit namings, even if it's returned as `is_ha` the `IsHighlyAvailable` or `IsHighAvailable` reads better

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+			); err != nil {
+				return diag.FromErr(err)
+			}
+			var networkPolicy, storageIntegration string
```
sfc-gh-jcieslak, line 252: both should be of type `sdk.AccountObjectIdentifier` instead of string

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+
+	errs := errors.Join(
+		stringAttributeUpdate(d, "comment", &set.Comment, &unset.Comment),
+		stringAttributeUpdate(d, "network_policy", &set.NetworkPolicy, &unset.NetworkPolicy),
+		stringAttributeUpdate(d, "storage_integration", &set.StorageIntegration, &unset.StorageIntegration),
+		stringAttributeUpdate(d, "postgres_settings", &set.PostgresSettings, &unset.PostgresSettings),
+		intAttributeUpdateSetOnly(d, "storage_size_gb", &set.StorageSizeGb),
+		intAttributeUpdateSetOnly(d, "postgres_version", &set.PostgresVersion),
+		intAttributeUpdate(d, "maintenance_window_start", &set.MaintenanceWindowStart, &unset.MaintenanceWindowStart),
+		booleanAttributeUpdateSetOnly(d, "high_availability", &set.HighAvailability),
```
sfc-gh-jcieslak, line 336: This is the place where you should split alters into multiple, not every parameter can be used with other ones (as we proved with integration tests)

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,377 @@
...
+func normalizePostgresSettings(s *string) *string {
+	if s == nil || *s == "{}" {
+		return nil
+	}
+	return s
+}
+
+// setOptionalFromNonEmptyStringPtr sets a key in resource data only if the
+// pointer is non-nil and the pointed-to string is non-empty.
+func setOptionalFromNonEmptyStringPtr(d *schema.ResourceData, key string, ptr *string) error {
```
sfc-gh-jcieslak, line 372: This should be moved to other similar helpers.

pkg/helpers/helpers.go
```diff
@@ -33,7 +33,7 @@ func EncodeSnowflakeID(attributes ...interface{}) string {
 	// is attribute already an object identifier?
 	if len(attributes) == 1 {
 		if id, ok := attributes[0].(sdk.ObjectIdentifier); ok {
-			if val := reflect.ValueOf(id); val.Kind() == reflect.Ptr && val.IsNil() {
+			if val := reflect.ValueOf(id); val.Kind() == reflect.Pointer && val.IsNil() {
```
sfc-gh-jcieslak, line 36: Most likely already addressed in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4725

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,338 @@
...
+		Description:      blocklistedCharactersFieldDescription("Specifies the identifier for the forked Postgres instance; must be unique for your account."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"fork_from": {
+		Type:        schema.TypeString,
+		Required:    true,
+		ForceNew:    true,
+		Description: "Specifies the identifier of the source Postgres instance to fork from.",
+	},
+	"at_timestamp": {
```
sfc-gh-jcieslak, line 34: To be consistent with other resources, please take a look at `stream_common.go` where we also have `at/before offset/timestamp` modeled in a more structural way.

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,338 @@
+package resources
```
sfc-gh-jcieslak, line 1: Most comments from postgres instance resource also apply here

pkg/schemas/gen/main/main.go
```diff
@@ -17,6 +17,7 @@ const (
 )
 
 func main() {
+	// Generate Show schemas (existing behavior)
```
sfc-gh-jcieslak, line 20: This PR contains too many changes. Two resources, generator changes, and changes from previous changes (e.g., integration tests). As I haven't reviewed generator changes yet, let's move them to their own PR, since this change can be done separately. You can leave the generated output though.

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,258 @@
...
+			HasNoBeforeOffset().
+			HasFullyQualifiedNameString(forkId.FullyQualifiedName()),
+		resourceshowoutputassert.PostgresInstanceShowOutput(t, modelBasic.ResourceReference()).
+			HasCreatedOnNotEmpty().
+			HasName(forkId.Name()).
+			HasOwner(snowflakeroles.Accountadmin.Name()).
+			HasOwnerRoleType("ROLE"),
+	}
+
+	modelWithComment := model.PostgresFork("test", forkId.Name(), sourceId.Name()).
```
sfc-gh-jcieslak, line 55: We usually use two models in BasicUseCase test:
- Minimal model without optionals
- Model with all optionals that are not marked with ForceNew

Also, regarding the steps in those tests, you can take a look at `TestAcc_ExternalOauthIntegration_BasicUseCase` and just copy the test steps we're doing there.

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,258 @@
...
+				Destroy: true,
+				Config:  accconfig.FromModels(t, modelBasic),
+				Check: assertThat(t,
+					invokeactionassert.PostgresForkDoesNotExist(t, forkId),
+				),
+			},
+		},
+	})
+}
+
```
sfc-gh-jcieslak, line 138: We usually also add CompleteUseCase test where we have two steps (create and import) where we have one model with all optionals set (with those that are also set as ForceNew), so this way we test the create path of fully set object.

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,258 @@
...
+func TestAcc_PostgresFork_WithAtTimestamp(t *testing.T) {
```
sfc-gh-jcieslak, line 139: This test would be then covered by CompleteUseCase and could be removed

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,258 @@
...
+					resourceshowoutputassert.PostgresInstanceShowOutput(t, modelFork.ResourceReference()).
+						HasCreatedOnNotEmpty().
+						HasName(forkId.Name()),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_PostgresFork_UpdateAfterFork(t *testing.T) {
```
sfc-gh-jcieslak, line 183: This is already covered by BasicUseCase test and can be removed

pkg/testacc/resource_postgres_instance_acceptance_test.go
```diff
@@ -0,0 +1,247 @@
...
+						plancheck.ExpectResourceAction(modelComplete.ResourceReference(), plancheck.ResourceActionCreate),
+					},
+				},
+				Check: assertThat(t, assertComplete...),
+			},
+		},
+	})
+}
+
+func TestAcc_PostgresInstance_Rename(t *testing.T) {
```
sfc-gh-jcieslak, line 184: Generally, rename can also be covered by BasicUseCase (you can add it as e.g. last step or just complete model could have different name)

pkg/testacc/resource_postgres_instance_acceptance_test.go
```diff
@@ -0,0 +1,247 @@
+//go:build account_level_tests
```
sfc-gh-jcieslak, line 1: Same comments from pkg/testacc/resource_postgres_fork_acceptance_test.go apply here

sfc-gh-jsouthard:
Superceded by #4759

PR #4714, opened 2026-05-07
chore: add timeout support in model builders

sfc-gh-jcieslak:
## Changes
- Adds a timeout block to the model builders


pkg/acceptance/bettertestspoc/config/resource_model.go
```diff
@@ -26,6 +26,7 @@ type ResourceModelMeta struct {
 	name      string
 	resource  resources.Resource
 	dependsOn []string
+	timeouts  map[string]string
```
sfc-gh-jmichalak: Nit: can we have a struct instead?

sfc-gh-jmichalak (dismissed):

pkg/acceptance/bettertestspoc/config/resource_model.go
sfc-gh-jcieslak: Let's leave it as a map for now. I checked that there's nothing more in both SDKv2 and plugin framework for timeouts, so it should be safe to keep it like this

sfc-gh-asawicki: I would prefer a struct too:
- we have only a predefined list of keys
- we could have crisper syntax of WithTimeout(timeout) (and have predefined timeouts we can reuse)

sfc-gh-asawicki (dismissed):

sfc-gh-jmichalak (approved):

PR #4731, opened 2026-05-12
fix: snowflake_schema empty default_ddl_collation override

sfc-gh-jmichalak:
## Summary
- Fix `snowflake_schema` so that `default_ddl_collation = ""` correctly overrides a non-empty value inherited from the parent database. Previously the update was a silent no-op because Terraform SDKv2 treats `""` on a `Computed + Optional` field as "no value provided", suppressing the diff before `handleParameterUpdate` ran.
- In `ParameterValueComputedIf`: force `SetNewComputed` whenever a parameter is inherited from a higher level and the config has a value — regardless of whether the configured value equals the inherited value.
- In `handleParameterUpdateWithMapping`: read the target value from `d.GetRawConfig()` (via a new generic `ctyValueToGo[T]` helper) instead of `d.Get`, which falls back to the state value when the plan is unknown after `SetNewComputed`.
- Add `TestAcc_Schema_EmptyParameterAsDefaultValue_WithDatabaseLevel` covering the parent-database-inheritance case, extend the existing `TestAcc_Schema_EmptyParameterAsDefaultValue` with plan checks, add unit tests for `ctyValueToGo`, and document the fix in `MIGRATION_GUIDE.md`.


pkg/resources/common.go
```diff
@@ -56,6 +57,24 @@ func suppressQuoting(_, oldValue, newValue string, _ *schema.ResourceData) bool
 	}
 }
 
+// ctyValueToGo converts a non-null cty.Value to the requested Go type.
+func ctyValueToGo[T any](v cty.Value) (T, error) {
```
sfc-gh-jcieslak: I think we should constrain the `T` if we know which type variants it supports, e.g.:
```go
type constraint interface {
	~int | ~string | ~bool
}

func Function[T constraint]() {}

func Caller() {
	Some[int]()
	Some[string]()
	Some[bool]()
	Some[any]() // errors (not in constraint)
}
```
It would move runtime errors to compilation time or even before (lsp errors)

pkg/resources/resource_parameters_commons_test.go
```diff
@@ -17,3 +18,60 @@ func Test_enrichWithReferenceToParameterDocs(t *testing.T) {
 		require.Equal(t, description+" "+"For more information, check [ABORT_DETACHED_QUERY docs](https://docs.snowflake.com/en/sql-reference/parameters#abort-detached-query).", enrichedDescription)
 	})
 }
+
+func Test_ctyValueToGo(t *testing.T) {
+	t.Run("converts non-empty string", func(t *testing.T) {
```
sfc-gh-jcieslak: nit: table-based test? Would be easier to read, add new cases, etc

sfc-gh-jcieslak (dismissed):

pkg/resources/custom_diffs.go
```diff
@@ -44,10 +44,15 @@ func ParameterValueComputedIf[T ~string](key string, parameters []*sdk.Parameter
...
-		if ok && !configValue.IsNull() && parameter.Level != objectParameterLevel && parameter.Value == valueToString(d.Get(key)) {
+		// When the value is set in the configuration but the parameter is not set on the object level,
+		// the parameter is inherited from a higher level (account/database/etc.). In this case we must
+		// "forcefully" mark the value as computed so the update can set the parameter on the correct level.
+		// This covers two sub-cases:
+		//   1. The configured value equals the inherited value: we still need the SET to lock it in at the object level.
+		//   2. The configured value differs from the inherited value (e.g. empty string overriding a non-empty parent
+		//      value): SDKv2 would otherwise suppress the diff because the field is Computed+Optional and "" is treated
+		//      as "no value", so d.HasChange returns false during update. SetNewComputed forces the update path to run.
+		if ok && !configValue.IsNull() && parameter.Level != objectParameterLevel {
```
sfc-gh-asawicki, line 55: Q: there is a test for parameters in the plugin framework. As the nulls vs empty are handled better there, I assume it's not a problem but it would be good to have the check for it too (maybe a separate PR?).

pkg/resources/resource_parameters_commons_test.go
sfc-gh-asawicki: +1

sfc-gh-asawicki (dismissed):

pkg/resources/common.go
sfc-gh-jmichalak: Done.

pkg/resources/custom_diffs.go
sfc-gh-jmichalak: Extracted to SNOW-3530282.

pkg/resources/resource_parameters_commons_test.go
sfc-gh-jmichalak: Done.

sfc-gh-jcieslak (dismissed):

sfc-gh-jcieslak (approved):

sfc-gh-asawicki (commented):
LGTM

PR #4733, opened 2026-05-12
fix: handle unknown trigger values in ComputedIfAnyAttributeChanged

### LINKED ISSUE #4188: [Bug]: Adding resource monitor to existing warehouse gives error
### Terraform CLI Version

1.9.8

### Terraform Provider Version

1.20.1

### Company Name

Traveler Indemnity Company

### Terraform Configuration

```terraform
locals {
  test_warehouse_name = "TEST_WAREHOUSE"
}

variable "add_monitor" {
  type    = bool
  default = false
}

resource "snowflake_warehouse" "test" {
  name             = local.test_warehouse_name
  resource_monitor = var.add_monitor ? snowflake_resource_monitor.monitor[0].fully_qualified_name : null
}

resource "snowflake_resource_monitor" "monitor" {
  count    = var.add_monitor ? 1 : 0

  name         = "${local.test_warehouse_name}_monitor"
  credit_quota = 100

  notify_triggers           = [50,75]
  suspend_trigger           = 100
  suspend_immediate_trigger = 150
}
```

### Category

category:resource

### Object type(s)

resource:warehouse

### Expected Behavior

Resource monitor to be added to existing warehouse without error.

### Actual Behavior

Resource monitor is not added to the warehouse and Terraform reports a provider error:

> │ Error: Provider produced inconsistent final plan
> │ 
> │ When expanding the plan for snowflake_warehouse.test to include new values learned so far during apply, provider "registry.terraform.io/snowflakedb/snowflake" produced an invalid new value for .show_output: was
> │ known, but now unknown.
> │ 
> │ This is a bug in the provider, which should be reported in the provider's own issue tracker.
> ╵

### Steps to Reproduce

Using the above configuration run the following steps:

1. `terraform init`
2. `terraform apply`
3. `terraform apply -var add_monitor=true`

A subsequent run of step 3 adds the monitor successfully.

If the monitor is attached when the warehouse is created no error is shown.

Medium

### Logs

_No response_

### Additional Information

_No response_

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
## Summary

- Adding a `resource_monitor` reference to an existing `snowflake_warehouse` in the same apply that creates the monitor used to error with `Provider produced inconsistent final plan` on `.show_output` (`was known, but now unknown`).
- Root cause: `ComputedIfAnyAttributeChanged` read the new value of a trigger field via `diff.GetChange`, which coerces an unknown config value (a reference to a not-yet-created resource) to the zero value. The trigger's `DiffSuppressFunc` was then called against that empty string and could match it against the current `show_output` value, so the computed output was left marked known at plan time. When the real value arrived at apply, the output flipped to unknown and tripped Terraform's plan/apply consistency check.
- Fix: short-circuit when `diff.NewValueKnown` returns false — mark the dependent computed field computed without consulting the suppressor. The helper is shared across ~170 call sites (`warehouse`, `account`, `account_role`, storage/catalog integrations, and many more), so the fix applies broadly.
- Added a regression acceptance test (`TestAcc_Warehouse_AddResourceMonitorAfterCreate`) that mirrors the original reproduction.
- Added a migration guide entry under `v2.15.x ➞ v2.16.0`.

 #4188

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -54,6 +54,20 @@ This feature will be marked as stable in future releases. To use this feature, a
 
 No changes are required for existing configurations unless you want to adopt this preview feature with Terraform.
 
+### *(bug fix)* "Provider produced inconsistent final plan" when adding a reference to a not-yet-created resource
+
+Previously, updating an existing resource to reference an attribute of another Terraform-managed resource that was being created in the same apply (for example, setting `resource_monitor = snowflake_resource_monitor.foo.fully_qualified_name` on an existing `snowflake_warehouse`) could fail with:
+
+> Provider produced inconsistent final plan [...] produced an invalid new value for .show_output: was known, but now unknown.
+
+The internal `ComputedIfAnyAttributeChanged` helper invoked each trigger field's `DiffSuppressFunc` against an empty string that was silently substituted for values still unknown at plan time, which caused computed outputs (like `show_output`) to be left marked as known. Once the referenced resource was created and the real value arrived at apply time, those outputs flipped to unknown — violating Terraform's plan/apply contract.
+
+The helper now short-circuits whenever the new value of a trigger field is unknown and marks the dependent computed field as computed, without consulting the suppressor. The fix applies to every resource that uses this helper (including `snowflake_warehouse`, `snowflake_account`, `snowflake_account_role`, and many others), so references across resources now plan and apply correctly.
+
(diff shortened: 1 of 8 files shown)


sfc-gh-kwasilewski (approved):
LGTM, but since it's a broad change affecting multiple resources, let's wait for tests to be completed.

sfc-gh-asawicki (approved):

PR #4734, opened 2026-05-13
fix: job fork status for tests

sfc-gh-jcieslak:
## Changes
- Fix fork PR status mirroring for repository_dispatch workflows.
- Stop trying to update GitHub-owned job check runs and instead mirror results to a synthetic check run owned by the workflow.
- Reuse the synthetic check across reruns with a stable external_id to avoid duplicate statuses on PRs.
- Apply the same behavior to both regular integration tests and account-level tests.

Addresses [the current failures](https://github.com/snowflakedb/terraform-provider-snowflake/actions/runs/25807738289/job/75815171830)

### DIFF
--- a/.github/workflows/tests-account-level.yml
+++ b/.github/workflows/tests-account-level.yml
@@ -179,7 +179,6 @@ jobs:
         id: update_check_run
         env:
           pr_number: ${{ github.event.client_payload.pull_request.number }}
-          number: ${{ github.event.client_payload.pull_request.number }}
           check_run_id: ${{ job.check_run_id }}
           # Conveniently, job.status maps to https://developer.github.com/v3/checks/runs/#update-a-check-run
           conclusion: ${{ job.status }}
@@ -203,7 +202,7 @@ jobs:
             const externalId = `fork-pr-${process.env.pr_number}-${jobName}`;
             const output = {
               title: `Tests ${conclusion}`,
-              summary: `Integration tests ${conclusion}. See the [workflow run](https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}) for details.`
+              summary: `Account-level tests ${conclusion}. See the [workflow run](https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}) for details.`
             };
 
             const { data: existingRuns } = await github.rest.checks.listForRef({
@@ -219,8 +218,8 @@ jobs:
                 run.external_id === externalId
               )
               .sort((left, right) => {
-                const leftDate = new Date(left.started_at || left.completed_at || 0).getTime();
-                const rightDate = new Date(right.started_at || right.completed_at || 0).getTime();
+                const leftDate = Math.max(new Date(left.started_at || 0).getTime(), new Date(left.completed_at || 0).getTime());
+                const rightDate = Math.max(new Date(right.started_at || 0).getTime(), new Date(right.completed_at || 0).getTime());
                 return rightDate - leftDate;
               })[0];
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tests-account-level.yml
```diff
@@ -170,16 +170,15 @@ jobs:
...
+      # but the job check runs are tied to the default branch, not the fork PR SHA.
+      # This step mirrors the job result onto the fork PR head SHA using a dedicated
+      # synthetic check run that we own, reusing it across reruns to avoid duplicates.
       - name: Set fork job status
         uses: actions/github-script@v6
         if: ${{ always() }}
         id: update_check_run
         env:
+          pr_number: ${{ github.event.client_payload.pull_request.number }}
           number: ${{ github.event.client_payload.pull_request.number }}
```
sfc-gh-jmichalak: Nit: this is duplicated. The same in the other workflow.

.github/workflows/tests-account-level.yml
```diff
@@ -200,20 +199,52 @@ jobs:
...
-              console.log(`No matching check run found for "${jobName}" on ${ref}`);
-              return;
+
+            const existingRun = existingRuns.check_runs
+              .filter((run) =>
+                run.head_sha === ref &&
+                run.name === jobName &&
+                run.external_id === externalId
+              )
+              .sort((left, right) => {
```
sfc-gh-jmichalak, line 220: Nit: we can use `Math.max` here. The same in the other workflow.

.github/workflows/tests-account-level.yml
```diff
@@ -200,20 +199,52 @@ jobs:
             const jobName = currentCheck.name;
             console.log(`Current job check run name: ${jobName}`);
             const ref = process.env.sha;
-            const { data: checks } = await github.rest.checks.listForRef({
+            const conclusion = process.env.conclusion;
+            const externalId = `fork-pr-${process.env.pr_number}-${jobName}`;
+            const output = {
+              title: `Tests ${conclusion}`,
+              summary: `Integration tests ${conclusion}. See the [workflow run](https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}) for details.`
```
sfc-gh-jmichalak:
```suggestion
              summary: `Account-level tests ${conclusion}. See the [workflow run](https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}) for details.`
```

.github/workflows/tests-account-level.yml
```diff
@@ -200,20 +199,52 @@ jobs:
...
+            };
+
+            const { data: existingRuns } = await github.rest.checks.listForRef({
```
sfc-gh-jmichalak, line 208: Note: the default paging is 30 results per page. It should be enough, just marking it.

sfc-gh-jmichalak (changes requested):

.github/workflows/tests-account-level.yml
sfc-gh-jcieslak: Removed duplication

.github/workflows/tests-account-level.yml
sfc-gh-jcieslak: Done, little bit longer, but easier to understand

.github/workflows/tests-account-level.yml
sfc-gh-jcieslak: Adjusted

sfc-gh-jmichalak (approved):

.github/workflows/tests-account-level.yml
sfc-gh-jmichalak: I meant using it instead of sorting. Additionally, I'm not sure if this sorting approach will work. I expected here to just get the run that had been started the latest (assuming only one running) or use check_run_id, but we can observe.

sfc-gh-jmichalak (approved):

PR #4739, opened 2026-05-14
feat: add snowflake_cortex_agent resource

sfc-gh-kwasilewski:
## Changes

- Added new snowflake_cortex_agent data source, behind a preview feature flag. The resource uses normalized comparison on specification so YAML in Terraform can be diffed against JSON returned by Snowflake without formatting-only changes.
- Added a direct dependency on github.com/goccy/go-yaml.


pkg/sdk/cortex_agents_ext.go
```diff
@@ -26,19 +32,35 @@ func UnmarshalCortexAgentProfile(profileAsJson string) (*CortexAgentProfile, err
 	return &profile, nil
 }
 
-// UnmarshalCortexAgentSpec parses agent specification JSON (as returned by DESCRIBE AGENT) into a map.
-func UnmarshalCortexAgentSpec(agentSpec string) (map[string]any, error) {
-	data := strings.TrimSpace(agentSpec)
+// MarshalCortexAgentProfile serializes a CortexAgentProfile to JSON.
+func MarshalCortexAgentProfile(profile CortexAgentProfile) (string, error) {
```
sfc-gh-jmichalak, line 32: Nit: this could be inlined.

pkg/resources/cortex_agent.go
```diff
@@ -0,0 +1,286 @@
...
+	"comment": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies a comment for the Cortex agent.",
+	},
+	"profile": {
+		Type:        schema.TypeList,
+		Optional:    true,
+		MaxItems:    1,
+		Description: "Specifies agent profile information, such as display_name, avatar, and color.",
```
sfc-gh-jmichalak, line 59: We can use `AtLeastOneOf` to force setting at least one attribute.

pkg/resources/cortex_agent.go
```diff
@@ -0,0 +1,286 @@
...
+		return diag.FromErr(err)
+	}
+
+	details, err := client.CortexAgents.Describe(ctx, id)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if details.Comment != nil {
+		if err := d.Set("comment", *details.Comment); err != nil {
```
sfc-gh-jmichalak: We also need to handle a case when it's empty - set an empty value.

pkg/resources/cortex_agent.go
```diff
@@ -0,0 +1,286 @@
...
+			return diag.FromErr(err)
+		}
+	}
+
+	profileList := make([]any, 0)
+	if details.Profile != nil {
```
sfc-gh-jmichalak, line 188: If we're checking agains nil here, why in UnmarshalCortexAgentProfile do we assume it can be empty? Does Snowflake return both of these options? If not, let's simplify this.

pkg/sdk/cortex_agents_ext.go
```diff
@@ -26,19 +32,35 @@ func UnmarshalCortexAgentProfile(profileAsJson string) (*CortexAgentProfile, err
...
+	if err != nil {
+		return "", err
+	}
+	return string(b), nil
+}
+
+// NormalizeCortexAgentSpecification parses YAML or JSON agent specifications into a canonical
+// JSON string (sorted object keys) so Terraform can compare user YAML with Snowflake JSON responses
+// without spurious formatting diffs.
+func NormalizeCortexAgentSpecification(spec string) (string, error) {
```
sfc-gh-jmichalak, line 43: Why not separate functions for JSON and YAML, and when we need to compare these two, we just compare the unmarshaled maps?

pkg/testacc/resource_cortex_agent_acceptance_test.go
```diff
@@ -0,0 +1,334 @@
...
+	partialProfile := sdk.CortexAgentProfile{
+		Color: sdk.String("green"),
+	}
+	externalProfile := sdk.CortexAgentProfile{
+		DisplayName: sdk.String("My Friendly Assistant"),
+		Color:       sdk.String("red"),
+	}
+
+	basic := model.CortexAgent("t", id.DatabaseName(), id.SchemaName(), id.Name(), "").
+		WithSpecificationValue(config.UnquotedWrapperVariable(hclSpec))
```
sfc-gh-jmichalak: The spec should be passed in CortexAgent calls, this applies to all model defs.

sfc-gh-jmichalak (changes requested):

pkg/resources/cortex_agent.go
sfc-gh-kwasilewski: Don't we already do that? If `*details.Comment` is an empty string, we set an empty string.

pkg/sdk/cortex_agents_ext.go
sfc-gh-kwasilewski: I'd rather not, because I would lose the ability to unit test it.

pkg/resources/cortex_agent.go
sfc-gh-kwasilewski: Yeah, it's not necessary in the end. Snowflake cannot return an empty string. To be removed.

pkg/testacc/resource_cortex_agent_acceptance_test.go
sfc-gh-kwasilewski: It wouldn't work, because the generated config would be `specification = "yamlencode({...})"`, which isn't valid.

sfc-gh-kwasilewski: We have to prevent double quotes from being added around yamlencode.

pkg/sdk/cortex_agents_ext.go
sfc-gh-kwasilewski: > Why not separate functions for JSON and YAML

It's easier that way; go-yaml can handle both formats.

> when we need to compare these two

> we just compare the unmarshaled maps

We don't directly compare JSON with YAML; we first normalize the input (whether YAML or JSON) to a canonical string representation and then compare those representations for diff detection.

pkg/resources/cortex_agent.go
sfc-gh-jmichalak: Yes, but if it's nil, this line won't execute. The problem is similar to the one below. We should model empty values with nil OR empty strings, not both. I think we already had bugs like this before.

pkg/testacc/resource_cortex_agent_acceptance_test.go
sfc-gh-jmichalak: Right, so pls add a comment here, and note this limitation in the acceptance test package readme if not already there.

pkg/sdk/cortex_agents_ext.go
sfc-gh-jmichalak: We can still unit test it with json.Marshal. Not blocking.

sfc-gh-jmichalak (changes requested):

pkg/acceptance/helpers/agent_client.go
```diff
@@ -74,3 +83,28 @@ instructions:
   response: "%s"
 `, response)
 }
+
+// SampleSpecAsYamlencodeHCL returns a multiline Terraform/HCL expression for use with
+// model.CortexAgent(...).WithSpecificationValue(config.UnquotedWrapperVariable(...)).
+func (c *AgentClient) SampleSpecAsYamlencodeHCL(t *testing.T, response string) string {
```
sfc-gh-asawicki: This shouldn't be in the agent_client:
- it adds dependency between the helpers package and config; the helpers are meant to be used as Snowflake gate from both acceptance AND integration tests; this package shouldn't know about terraform
- it should be moved e.g. to the model ext file

pkg/testacc/resource_cortex_agent_acceptance_test.go
sfc-gh-asawicki: Let's have a constructor added to ext which hides this from the invocation place:
- this workaround will be then implemented once and commented once
- in other objects we add dedicated constructors too so it will be consistent

pkg/resources/cortex_agent.go
```diff
@@ -0,0 +1,289 @@
...
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.CortexAgentResource), TrackingDeleteWrapper(resources.CortexAgent, deleteFunc)),
+		Description:   "Resource used to manage Cortex agent objects. For more information, check [Cortex agent documentation](https://docs.snowflake.com/en/sql-reference/sql/create-agent).",
+
+		Schema: cortexAgentSchema,
+		Importer: &schema.ResourceImporter{
+			StateContext: TrackingImportWrapper(resources.CortexAgent, ImportName[sdk.SchemaObjectIdentifier]),
+		},
+		Timeouts: defaultTimeouts,
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.CortexAgent, customdiff.All(
+			ComputedIfAnyAttributeChanged(cortexAgentSchema, ShowOutputAttributeName, "comment", "profile"),
```
sfc-gh-asawicki, line 123: question: is it working correctly for the object list fields? I remember that pretty recently there were an issue resulting from incorrect handling of nested objects. @sfc-gh-jmichalak @sfc-gh-jcieslak do you remember it?

pkg/resources/cortex_agent.go
sfc-gh-asawicki: Alternative: We could handle it as external change on show/describe, and not set it through normal set at all here.

pkg/resources/cortex_agent.go
```diff
@@ -0,0 +1,289 @@
...
+					Severity: diag.Warning,
+					Summary:  "Failed to query Cortex agent. Marking the resource as removed.",
+					Detail:   fmt.Sprintf("Cortex agent id: %s, Err: %s", id.FullyQualifiedName(), err),
+				},
+			}
+		}
+		return diag.FromErr(err)
+	}
+
+	details, err := client.CortexAgents.Describe(ctx, id)
```
sfc-gh-asawicki, line 177: nit: we could have a dedicated method already "translating" the describe to a struct that can be consumed here (like with a few other objects).

pkg/resources/cortex_agent.go
```diff
@@ -0,0 +1,289 @@
...
+	var modifyLiveVersionSetRequest *sdk.CortexAgentModifyLiveVersionSetRequest
+	setRequest := sdk.NewCortexAgentSetRequest()
+
+	if d.HasChange("specification") {
+		modifyLiveVersionSetRequest = sdk.NewCortexAgentModifyLiveVersionSetRequest(d.Get("specification").(string))
+	}
+
+	if err := errors.Join(
+		// TODO [SNOW-3530838]: UNSET not implemented
+		stringAttributeUpdateSetOnly(d, "comment", &setRequest.Comment),
```
sfc-gh-asawicki, line 240: q: if there is a difference between nil and empty on SF (usually it's not but from other discussion I understand it is in this case), will all the transitions (nil-> empty, empty-> nil, both internal/external) be handled correctly?

sfc-gh-asawicki (changes requested):

pkg/resources/cortex_agent.go
sfc-gh-jmichalak: I think it happened because one of the nested fields had a custom diff; in this case, it does not.

sfc-gh-kwasilewski: It works for some object list fields but not others. Session policy resource also has list fields (e.g. `blocked_secondary_roles`), and in that case, it didn't work, maybe because they contain an inner list. Here we only have simple string fields.

pkg/resources/cortex_agent.go
sfc-gh-kwasilewski: > We could handle it as external change on show/describe, and not set it through normal set at all here.

I think this overcomplicates things (and may also affect the import flow). Luckily, the resource doesn't need special external change detection; let's keep that. I'll follow Kuba's advice.

pkg/resources/cortex_agent.go
sfc-gh-kwasilewski: Added some tests

pkg/testacc/resource_cortex_agent_acceptance_test.go
sfc-gh-kwasilewski: Done

pkg/resources/cortex_agent.go
sfc-gh-kwasilewski: Done

pkg/acceptance/helpers/agent_client.go
sfc-gh-kwasilewski: Done

pkg/resources/cortex_agent.go
sfc-gh-kwasilewski: To be done in a follow-up.

sfc-gh-asawicki (dismissed):

sfc-gh-jmichalak (dismissed):

sfc-gh-jmichalak (approved):

PR #4740, opened 2026-05-14
chore: Generate convert implementation for the SDK objects

sfc-gh-asawicki:
Continuation of https://github.com/snowflakedb/terraform-provider-snowflake/pull/4655.
                                                                                                                                                                                                                                                    
  Introduces WithConvertGeneration() to the SDK generator, which generates a fully-typed `convert()` method body for a `PairedStruct` instead of a `// TODO: Mapping` placeholder.
                                                                                                                                                                                                                                                    
  - Add `AssignmentKind` enum and boolean predicate methods on `FieldPair` to classify each db→plain field mapping                                                                                                                                
  - Add `AssignmentKindFuncMap()` derived from the constants, so `convert.tmpl` requires no manual `FuncMap` maintenance                                                                                                                                   
  - Add centralized identifier type list to `genhelpers` (`sdk_identifier_types.go`)                                                                                                                                                                    
  - Adjust `convert.tmpl` template                                                                                                                        
  - Wire `WithConvertGeneration()` through `ShowOperationWithPairedStructs`, `DescribeOperationWithPairedStructs`, and `InstanceMethodOperation`                                                                                              
  - Extend example defs (`paired_struct_def.go`) to cover all supported assignment kinds with generated output and regenerate                                                                                                                                        
  - Enable `WithConvertGeneration()` for first production def: notebooks
                                                                                                                                                                                                                                                    
  Follow-ups:                                                                                                                                                                                                                                       
  - Enable `WithConvertGeneration()` for the remaining defs (adjusting the convert capabilities as we go)                                                                                                                            
  - Replace `PlainField(col, enumKind)` with `EnumField`/`OptionalEnumField` to unblock partially-ready defs (compute_pools, image_repositories, network_rules, organization_accounts, stages, streams, services, tasks, listings, user_programmatic_access_tokens)                                                                                                                                                                                                                                                                                    
  - Migrate remaining old-style defs (DbStruct/PlainStruct) to paired structs: external_functions, file_formats, functions, procedures, postgres_instances, data_metric_function_references


pkg/sdk/notebooks_impl_gen.go
```diff
@@ -175,23 +174,20 @@ func (r NotebookDetailsRow) convert() (*NotebookDetails, error) {
...
-	mapNullString(&n.LastVersionSourceLocationUri, r.LastVersionSourceLocationUri)
-	mapNullString(&n.LastVersionGitCommitHash, r.LastVersionGitCommitHash)
-	mapNullString(&n.LiveVersionLocationUri, r.LiveVersionLocationUri)
-
-	return n, nil
+	mapNullString(&result.Title, r.Title)
+	mapNullStringWithMapping(&result.QueryWarehouse, r.QueryWarehouse, ParseAccountObjectIdentifier)
+	mapNullString(&result.UserPackages, r.UserPackages)
+	mapNullString(&result.RuntimeName, r.RuntimeName)
+	mapNullStringWithMapping(&result.ComputePool, r.ComputePool, ParseAccountObjectIdentifier)
```
sfc-gh-jcieslak, line 181: Kind of related: the current helper mapping functions were done when convert was not returning errors. Now, we are able to bubble up the any errors, which means the helper function could be adjusted and the converter should take those errors into account (e.g. by wrapping them in errors.Join block)

pkg/sdk/generator/example/defs/paired_struct_def.go
```diff
@@ -60,7 +60,14 @@ func pairedStructExampleAllOptions(dbName, plainName string) *g.PairedStructs {
 		// db string, plain DatabaseObjectIdentifier; plain name overridden (default would be "Id")
 		DatabaseObjectIdentifier("database_object_id", g.WithPlainFieldName("DatabaseObjectId")).
 		// db string, plain DatabaseObjectIdentifier; plain name overridden
-		DatabaseObjectIdentifier("second_database_object_id", g.WithPlainFieldName("OverriddenDatabaseObjectId"))
+		DatabaseObjectIdentifier("second_database_object_id", g.WithPlainFieldName("OverriddenDatabaseObjectId")).
+		// db string, plain ExampleStatus enum; auto-derived field name
+		EnumField("status", "ExampleStatus").
+		// db sql.NullString, plain *ExampleStatus; optional enum with plain name override
+		OptionalEnumField("optional_status", "ExampleStatus", g.WithPlainFieldName("OptionalStatus")).
+		// db string, JSON-decoded into []PairedStructExampleMetadata
```
sfc-gh-jcieslak, line 68: Q: What happens if the root is an object `{}`, and not an array `[]`?

pkg/sdk/generator/gen/templates/sub_templates/convert.tmpl
```diff
@@ -1,6 +1,48 @@
...
+    {{- range .FieldPairs }}
+        {{- if .AssignmentKindDirect }}
+            {{ .PlainFieldName }}: r.{{ .DbFieldName }},
+        {{- else if .AssignmentKindStringToBool }}
+            {{ .PlainFieldName }}: r.{{ .DbFieldName }} == "Y",
+        {{- end }}
+    {{- end }}
+    }
+    {{- range .FieldPairs }}
+        {{- if .AssignmentKindDirect }}
```
sfc-gh-jcieslak, line 15: nit: unused kinds could be removed

pkg/sdk/generator/gen/templates/sub_templates/convert.tmpl
```diff
@@ -1,6 +1,48 @@
...
+            {{ .NullMapFunc }}(&result.{{ .PlainFieldName }}, r.{{ .DbFieldName }})
+        {{- else if .AssignmentKindNullableToRequired }}
+            {{ .NullRequiredMapFunc }}(&result.{{ .PlainFieldName }}, r.{{ .DbFieldName }})
+        {{- else if .AssignmentKindNullableToIdentifier }}
+            mapNullStringWithMapping(&result.{{ .PlainFieldName }}, r.{{ .DbFieldName }}, Parse{{ TypeWithoutPointer .PlainKind }})
+        {{- else if .AssignmentKindNullableToEnum }}
+            mapNullStringWithMapping(&result.{{ .PlainFieldName }}, r.{{ .DbFieldName }}, To{{ TypeWithoutPointer .PlainKind }})
+        {{- else if .AssignmentKindNullableToStringArray }}
+            if r.{{ .DbFieldName }}.Valid {
+            	result.{{ .PlainFieldName }} = ParseCommaSeparatedStringArray(r.{{ .DbFieldName }}.String, false)
```
sfc-gh-jcieslak, line 37: nit: In a follow-up or as limitation, we should document that it only supports perser without quotes (always set to false for now)

sfc-gh-jcieslak (dismissed):

pkg/sdk/notebooks_impl_gen.go
sfc-gh-asawicki: Good point. There is a dedicated one for that: SNOW-3108659. I didn't want to do both simultaneously and I believe that doing it in order: generate first, adjust later is more desirable (as lack of generation for `convert` is the biggest gap in the SDK generation).

pkg/sdk/generator/gen/templates/sub_templates/convert.tmpl
sfc-gh-asawicki: they are here on purpose (there is an else at the bottom).

pkg/sdk/generator/gen/templates/sub_templates/convert.tmpl
sfc-gh-asawicki: We'll see with other converts. If any of the existing ones requires the quotes variant, we will adjust it on the structspair definition level and distinguish here.

pkg/sdk/generator/example/defs/paired_struct_def.go
sfc-gh-asawicki: The type is defined in the structpair def, so you need to point to/create an object appropriate for the given expected output in this column in dbStruct. The generated conversion should be exactly the same.

sfc-gh-jcieslak (approved):

PR #4746, opened 2026-05-15
chore: Change map to struct for model builder timeouts

sfc-gh-jcieslak:
## Changes
A follow-up for #4714 (addressing a comment regarding converting the Timeout map into a struct, and re-generating the model builders).


sfc-gh-jmichalak (approved):
Approved, but why not change it directly in the original PR? Now, we need to review more PRs, and we'll need to reapprove the original one anyway.

PR #4747, opened 2026-05-15
chore: Add basic iceberg tables SDK definition

sfc-gh-jmichalak:
## Summary

- Adds complete SDK definition for Snowflake Iceberg Tables in `pkg/sdk/generator/defs/iceberg_tables_def.go`
- Generates all SDK artifacts: interface (`iceberg_tables_gen.go`), implementation (`iceberg_tables_impl_gen.go`), DTOs and builders (`iceberg_tables_dto_gen.go`, `iceberg_tables_dto_builders_gen.go`), enums (`iceberg_tables_enums_gen.go`), validations (`iceberg_tables_validations_gen.go`)
- Adds generated unit tests (`iceberg_tables_gen_test.go`) covering SQL statement building for all operations
- Adds extension file (`iceberg_tables_ext.go`) for custom helpers

## TODO
- Add out of line and inline constraints
- Add join, projection policy
- Add integration tests, the resource and data source.


pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+)
+
+var icebergTableColumn = g.NewQueryStruct("IcebergTableColumn").
+	Text("Name", g.KeywordOptions().Required().DoubleQuotes()).
+	PredefinedQueryStructField("ColumnType", "datatypes.DataType", g.ParameterOptions().NoQuotes().NoEquals().Required()).
+	PredefinedQueryStructField("DefaultValue", g.KindOfTPointer[sdkcommons.ColumnDefaultValue](), g.KeywordOptions()).
+	OptionalSQL("NOT NULL").
+	// TODO(next PR): add inline constraint support
+	// OptionalQueryStructField("InlineConstraint", icebergTableColumnInlineConstraint, g.KeywordOptions()).
+	OptionalTags().
```
sfc-gh-jcieslak, line 44: What about masking and projection policies?

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+		Create().
+		OrReplace().
+		OptionalSQL("TRANSIENT").
+		SQL("ICEBERG TABLE").
+		IfNotExists().
+		Name().
+		QueryStructField("ColumnsAndConstraints", icebergTableColumnsAndConstraints, g.ListOptions().Parentheses()).
+		ListQueryStructField("PartitionBy", icebergTablePartitionExpression, g.KeywordOptions().Parentheses().SQL("PARTITION BY")).
+		OptionalAssignment("PATH_LAYOUT", IcebergTablePathLayoutEnumDef.KindPtr(), g.ParameterOptions().NoQuotes()).
+		PredefinedQueryStructField("ClusterBy", "[]Column", g.KeywordOptions().Parentheses().SQL("CLUSTER BY")).
```
sfc-gh-jcieslak: This could be either column name or column expression (in this case, I'm not sure if quoting in Column struct allows specifying the expression that Snowflake would accept).

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+		OptionalIdentifier("ExternalVolume", g.KindOfT[sdkcommons.AccountObjectIdentifier](), g.IdentifierOptions().SQL("EXTERNAL_VOLUME")).
+		OptionalSQLWithCustomFieldName("CatalogSnowflake", "CATALOG = 'SNOWFLAKE'").
```
sfc-gh-jcieslak: I'm wondering if this should be a boolean or just an enum with one value possible 🤔 In case of extending values, the enum version would be safer and wouldn't require us to write much migration logic. Because of it, maybe we should adjust it?

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+
+var icebergTableAddColumnAction = g.NewQueryStruct("IcebergTableAddColumnAction").
+	SQL("ADD COLUMN").
+	OptionalSQL("IF NOT EXISTS").
+	Text("Name", g.KeywordOptions().Required().DoubleQuotes()).
+	PredefinedQueryStructField("ColumnType", "datatypes.DataType", g.ParameterOptions().NoQuotes().NoEquals().Required()).
+	// TODO(next PR): add inline constraint support
+	// OptionalQueryStructField("InlineConstraint", icebergTableColumnInlineConstraint, g.KeywordOptions()).
+	PredefinedQueryStructField("DefaultValue", g.KindOfTPointer[sdkcommons.ColumnDefaultValue](), g.KeywordOptions()).
+	OptionalTags()
```
sfc-gh-jcieslak, line 134: Image

I didn't see anything in the docs about tags during the add column action, but it seems we're missing collate here, right?

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+
+var icebergTableAlterColumnAction = g.NewQueryStruct("IcebergTableAlterColumnAction").
+	SQL("ALTER COLUMN").
+	Text("ColumnName", g.KeywordOptions().Required().DoubleQuotes()).
+	OptionalSQL("SET NOT NULL").
```
sfc-gh-jcieslak, line 139: I think we could model this slighly differently as currently only one column modification is possible. By having ALTER ( <list_item> ) where `list_item` is a struct with all column alter options, we would be able to perform multiple alter operations at one time (covering Snowflake capabilities), which could come in handy in resource when Update contains more column changes in one go. Right now, it seems one column action is possible per call (at least that's what it looks like by looking at the docs; I can see we're using `ListQueryStructField` to wrap `icebergTableAlterColumnAction`, but not sure if that works as expected due to multiple ALTER keyword appearences)

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+	g.NewQueryStruct("AlterIcebergTable").
+		Alter().
+		SQL("ICEBERG TABLE").
+		IfExists().
+		Name().
+		OptionalQueryStructField(
+			"AddColumnAction",
+			icebergTableAddColumnAction,
+			g.KeywordOptions(),
+		).
```
sfc-gh-jcieslak, line 218: It seems we are missing rename and drop column actions

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+	OptionalAssignment("TARGET_FILE_SIZE", IcebergTableTargetFileSizeEnumDef.KindPtr(), g.ParameterOptions().SingleQuotes()).
+	PredefinedQueryStructField("Contact", "[]TableContact", g.KeywordOptions().Parentheses().SQL("CONTACT")).
+	OptionalAssignment("LOG_EVENT_LEVEL", IcebergTableLogEventLevelEnumDef.KindPtr(), g.ParameterOptions().NoQuotes()).
+	OptionalBooleanAssignment("ERROR_LOGGING", g.ParameterOptions()).
+	OptionalBooleanAssignment("ENABLE_DATA_COMPACTION", g.ParameterOptions()).
+	OptionalBooleanAssignment("ENABLE_ICEBERG_MERGE_ON_READ", g.ParameterOptions()).
+	OptionalTextAssignment("COMMENT", g.ParameterOptions().SingleQuotes()).
+	WithValidation(g.AtLeastOneValueSet, "ReplaceInvalidCharacters", "CatalogSync", "DataRetentionTimeInDays", "AutoRefresh", "TargetFileSize", "Contact", "LogEventLevel", "ErrorLogging", "EnableDataCompaction", "EnableIcebergMergeOnRead", "Comment")
+
+var icebergTableUnsetProperties = g.NewQueryStruct("IcebergTableUnsetProperties").
```
sfc-gh-jcieslak, line 113: Q: Did you check if unset is correct and not every property from set has its counterpart in unset?

pkg/sdk/generator/defs/tables_def.go
```diff
@@ -0,0 +1,43 @@
+package defs
+
+import (
+	g "github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/generator/gen"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/generator/gen/sdkcommons"
+)
+
```
sfc-gh-jcieslak, line 7: Q: is it separted because those are general table actions that we are able to apply to iceberg (and potentially other) tables?

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+			"Unset",
+			icebergTableUnsetProperties,
+			g.KeywordOptions().SQL("UNSET"),
+		).
+		OptionalSetTags().
+		OptionalUnsetTags().
+		PredefinedQueryStructField("AddRowAccessPolicy", "*ViewAddRowAccessPolicy", g.KeywordOptions()).
+		PredefinedQueryStructField("DropRowAccessPolicy", "*ViewDropRowAccessPolicy", g.KeywordOptions()).
+		PredefinedQueryStructField("DropAndAddRowAccessPolicy", "*ViewDropAndAddRowAccessPolicy", g.ListOptions().NoParentheses()).
+		OptionalSQL("DROP ALL ROW ACCESS POLICIES").
```
sfc-gh-jcieslak, line 274: It seems we're also missing searchOptimizationActions and what about other alter commands like https://docs.snowflake.com/en/sql-reference/sql/alter-iceberg-table-refresh?

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+		IfExists().
+		Name().
+		OptionalSQL("CASCADE").
+		OptionalSQL("RESTRICT").
+		WithValidation(g.ValidIdentifier, "name").
+		WithValidation(g.ConflictingFields, "Cascade", "Restrict"),
+).ShowOperationWithPairedStructs(
+	"https://docs.snowflake.com/en/sql-reference/sql/show-iceberg-tables",
+	g.StructPair("icebergTableRow", "IcebergTable").
+		Time("created_on").
```
sfc-gh-jcieslak, line 294: Q: when defining nullability of the columns did you take into account what columns are returned with `TERSE` option? As it seems there are column defined here which are non-nullable and are not returned on `TERSE` output

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -0,0 +1,346 @@
...
+		WithConvertGeneration(),
+	g.NewQueryStruct("ShowIcebergTables").
+		Show().
+		Terse().
+		SQL("ICEBERG TABLES").
+		OptionalLike().
+		OptionalIn().
+		OptionalStartsWith().
+		OptionalLimitFrom(),
+	// NOTE: TYPE=COLUMNS returns the same data as omitting the TYPE parameter.
```
sfc-gh-jcieslak: I think this should be a bit lower, next to the desc command, right?

pkg/sdk/iceberg_tables_gen_test.go
```diff
@@ -0,0 +1,703 @@
...
+			{
+				Hour: &IcebergTablePartitionHour{
+					Args: IcebergTablePartitionTimeArgs{Column: "C5"},
+				},
+			},
+		}
+		assertOptsValidAndSQLEquals(t, opts,
+			`CREATE ICEBERG TABLE %s PARTITION BY (TRUNCATE (10, "C1"), YEAR ("C2"), MONTH ("C3"), DAY ("C4"), HOUR ("C5")) AGGREGATION POLICY %s`,
+			id.FullyQualifiedName(), aggregationPolicyId.FullyQualifiedName())
+	})
```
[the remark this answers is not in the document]
sfc-gh-jmichalak: Already addressed — `opts.AggregationPolicy` is set in the current `iceberg_tables_gen_test.go` (lines 164–165), and the standalone "partition by truncate, year, month, day, hour" subtest the bot referenced no longer exists in this branch. Resolving.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Deferred to a follow-up PR. Added a `// TODO(next PR): add inline masking and projection policy support` comment in `icebergTableColumn` to track it (mirroring the existing `inline constraint support` TODO). Inline column-level policies will land alongside inline constraints.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Good catch. Switched `ClusterBy` from `[]Column` to `[]string` for both the CREATE clause and the ALTER clustering action. Quoting is now the caller's responsibility, which lets users pass either bare column names (`"col1"`) or arbitrary expressions (`SUBSTRING(c1, 1, 5)`) — matching what Snowflake actually accepts. Same pattern as `hybrid_tables_def.go`.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Agreed — switched to a single-value enum (`IcebergTableCatalogEnumDef` with one value `SNOWFLAKE`). Replaced the `OptionalSQLWithCustomFieldName("CatalogSnowflake", "CATALOG = 'SNOWFLAKE'")` with a regular `OptionalAssignment("CATALOG", IcebergTableCatalogEnumDef.KindPtr(), …)`. Adding new catalog values later is now non-breaking.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Re: COLLATE — Iceberg tables don't support collated columns (Snowflake errors with `Iceberg tables cannot be created with collated columns.`), so I left it out. Re: tags on ADD COLUMN — kept `OptionalTags()` since it works in practice on iceberg tables despite not being prominent in the docs.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Refactored. Changed `icebergTableAlterColumnAction.SQL("ALTER COLUMN")` to `SQL("COLUMN")`, and wrapped the list with `g.KeywordOptions().SQL("ALTER")` on the `ListQueryStructField`. The generator emits a single leading `ALTER` followed by a comma-separated list of `COLUMN <name> <action>` items, so the rendered SQL is now e.g. `ALTER COLUMN "c1" SET NOT NULL, COLUMN "c2" DROP NOT NULL` — multiple per-column actions work in one statement. Updated tests to match.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Deferred to a follow-up PR. Added a `// TODO(next PR): add RENAME COLUMN and DROP COLUMN actions` comment near the alter operation.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Verified each property against live SQL. Live behavior:

| Property | UNSET works |
|---|---|
| ERROR_LOGGING | ✅ kept (despite not being in the docs) |
| DATA_RETENTION_TIME_IN_DAYS | ✅ added |
| MAX_DATA_EXTENSION_TIME_IN_DAYS | ✅ added (also added to SET — was missing) |
| TARGET_FILE_SIZE | ✅ added |
| COMMENT | ✅ added |
| AUTO_REFRESH | ❌ skipped — Snowflake returns `invalid value 'null' for property 'AUTO_REFRESH'` |
| CONTACT | ❌ skipped — Snowflake returns `invalid property 'CONTACT' for 'TABLE'` |

So the asymmetry is real but smaller than it looked. Updated SET to include `MAX_DATA_EXTENSION_TIME_IN_DAYS`; updated UNSET to support 10 properties total. Tests updated.

pkg/sdk/generator/defs/tables_def.go
sfc-gh-jmichalak: Yes, exactly — these column-policy and tag actions (`ALTER COLUMN ... SET MASKING POLICY`, `SET PROJECTION POLICY`, `SET/UNSET TAG`) share the same SQL shape between regular tables and iceberg tables, so they live in `tables_def.go` to be reused. Iceberg references them in `iceberg_tables_def.go` via `tableSetColumnMaskingPolicy` etc.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Both deferred to follow-up PRs. Added `// TODO(next PR): add search optimization actions (ADD/DROP SEARCH OPTIMIZATION)` and `// TODO(next PR): add ALTER ICEBERG TABLE ... REFRESH (separate operation; see https://docs.snowflake.com/en/sql-reference/sql/alter-iceberg-table-refresh)` near the alter operation. Search optimization will likely reuse the `searchOptimization` struct from `event_tables_def.go`; REFRESH is a distinct command, so it'll be modeled as its own operation.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Removed `.Terse()` from the show operation. Keeping the show row schema strict (all documented non-null columns stay `Text`/`Number`/`Time`) avoids the partial-column problem entirely. Users who need TERSE output can fall back to raw SQL for the rare case.

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Right — moved the NOTE next to the `OptionalAssignmentWithFieldName("TYPE", IcebergTableDescribeTypeEnumDef.Kind(), …)` line in `DescribeIcebergTable`, where the `TYPE=COLUMNS`/`TYPE=STAGE` reference belongs.

sfc-gh-jcieslak (approved):

PR #4758, opened 2026-05-20
chore: add stale error repo test

sfc-gh-jcieslak:
## Changes
- Added a functional test that reproduces "Stale plan" error, which we observed during delete steps
- Observations: By running `go get github.com/hashicorp/terraform-plugin-testing@v1.14.0 && go mod tidy` I was able to 100% reproduce the error, and on the currently used testing library, as the release notes suggested, the error does not appear anymore, and the test passes.

### DIFF
--- a/pkg/testfunctional/test_resource_stale_error_repro.go
+++ b/pkg/testfunctional/test_resource_stale_error_repro.go
@@ -1,0 +1,116 @@
+package testfunctional
+
+import (
+	"context"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/testfunctional/common"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+var _ resource.ResourceWithConfigure = &staleErrorReproResource{}
+
+func NewStaleErrorReproResource() resource.Resource {
+	return &staleErrorReproResource{
+		HttpServerEmbeddable: *common.NewHttpServerEmbeddable[StaleReproRead]("stale_error_repro"),
+	}
+}
+
+type staleErrorReproResource struct {
+	common.HttpServerEmbeddable[StaleReproRead]
+}
+
+// StaleReproRead is the response type returned by the HTTP server handler.
+// The handler generates a new random value on every GET, so successive Read()
+// calls always return different values — the condition that makes the stale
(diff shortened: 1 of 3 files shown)


pkg/testfunctional/test_resource_stale_error_repro.go
```diff
@@ -0,0 +1,116 @@
...
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			// random_int is re-fetched from the HTTP server on every Read call.
+			// The server generates a new value each time, so any two consecutive reads
+			// produce different state — guaranteeing the Refresh() inserted between
+			// CreatePlan and Apply by terraform-plugin-testing v1.14.0 (triggered when
+			// a destroy step has a non-nil Check) always writes a new state serial,
+			// making the plan stale on every single run.
+			"random_int": schema.Int64Attribute{
```
sfc-gh-jmichalak, line 49: I think calling an HTTP server is overkill. Why don't we just set the random value inlined?

sfc-gh-jmichalak (dismissed):

pkg/testfunctional/test_resource_stale_error_repro_acceptance_test.go
```diff
@@ -0,0 +1,104 @@
...
+// randomIntHandler returns a brand-new random int64 on every GET request.
+// This guarantees that the state after Refresh() differs from the state embedded
+// in the destroy plan, making "Saved plan is stale" 100% reproducible.
+type randomIntHandler struct{}
+
+func (h *randomIntHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	switch r.Method {
+	case http.MethodGet:
+		w.WriteHeader(http.StatusOK)
+		_ = json.NewEncoder(w).Encode(testfunctional.StaleReproRead{RandomInt: rand.Int64()}) // #nosec G404
```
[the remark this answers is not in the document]
sfc-gh-asawicki: It's currently handled like this in other helpers too, and this is fine (in other places we have comment why

pkg/testfunctional/test_resource_stale_error_repro.go
sfc-gh-asawicki: +1

pkg/testfunctional/test_resource_stale_error_repro_acceptance_test.go
```diff
@@ -0,0 +1,104 @@
+package testfunctional_test
+
+// TestAcc_TerraformPluginFrameworkFunctional_StaleErrorRepro reproduces the
```
sfc-gh-asawicki, line 3: the only question/doubt I have, is that the referenced test is not using the plugin framework resource but sdkv2. I know we are proving the testing framework behavior (which is common for both) but a closer repro would be with an sdkv2 resource. I think we can leave it as is (i.e. in plugin framework).

sfc-gh-asawicki (dismissed):

pkg/testfunctional/test_resource_stale_error_repro.go
sfc-gh-jcieslak: I was able to drop the HTTP server and get a consistent error on the v1.14.0 version of the testing library.

sfc-gh-jmichalak (approved):

PR #4759, opened 2026-05-21
feat: implement resources for snowflake postgres instance and snowflake postgres fork (test resources)

sfc-gh-jsouthard:
Supercedes #4712 - restructures as a rebasing off of dev to strip away merge bubbles and have a more parsable history

Implements feedback from #4712, future feedback can be more readily managed in this branch.


MIGRATION_GUIDE.md
```diff
@@ -148,50 +160,6 @@ No configuration changes are required.
 
 ## v2.15.x ➞ v2.16.0
 
-### *(improvement)* snowflake_password_policy resource rework
```
sfc-gh-jcieslak, line 433: incorrectly resolved conflicts, please don't remove old migration guide entries

pkg/acceptance/bettertestspoc/config/model/postgres_fork_model_gen.go
```diff
@@ -0,0 +1,187 @@
...
+	Before             tfconfig.Variable `json:"before,omitempty"`
+	Comment            tfconfig.Variable `json:"comment,omitempty"`
+	ComputeFamily      tfconfig.Variable `json:"compute_family,omitempty"`
+	ForkFrom           tfconfig.Variable `json:"fork_from,omitempty"`
+	FullyQualifiedName tfconfig.Variable `json:"fully_qualified_name,omitempty"`
+	HighAvailability   tfconfig.Variable `json:"high_availability,omitempty"`
+	PostgresSettings   tfconfig.Variable `json:"postgres_settings,omitempty"`
+	StorageSizeGb      tfconfig.Variable `json:"storage_size_gb,omitempty"`
+
+	DynamicBlock *config.DynamicBlock `json:"dynamic,omitempty"`
```
sfc-gh-jcieslak, line 25: New changes to the model builder were merged that add timeout support. As model builder generation is checked in CI you should merge with latest dev changes and run `make pre-push` to ensure basic checks pass.

pkg/acceptance/helpers/postgres_instance_client.go
```diff
@@ -70,13 +70,31 @@ func (c *PostgresInstanceClient) WaitForReady(t *testing.T, id sdk.AccountObject
...
+// an additional settling period. Snowflake reports READY before the instance
+// is actually available for fork operations.
+func (c *PostgresInstanceClient) WaitForForkReady(t *testing.T, id sdk.AccountObjectIdentifier, timeout time.Duration) *sdk.PostgresInstance {
 	t.Helper()
-	instance, cleanup := c.Create(t)
-	instance = c.WaitForReady(t, instance.ID(), 5*time.Minute)
-	return instance, cleanup
+	instance := c.WaitForReady(t, id, timeout)
+	// The backend needs additional time after reporting READY before it accepts fork requests.
+	time.Sleep(30 * time.Second)
```
sfc-gh-jcieslak: It's a shame we don't have a dedicated status for this, but I guess we don't have any other way to wait for it. :/

pkg/acceptance/bettertestspoc/assert/resourceassert/postgres_instance_resource_ext.go
```diff
@@ -0,0 +1 @@
+package resourceassert
```
sfc-gh-jcieslak: Empty file

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
```diff
@@ -180,10 +180,6 @@ var allStructs = []SdkObjectDef{
 		IdType:       "sdk.SchemaObjectIdentifier",
 		ObjectStruct: sdk.NetworkRule{},
 	},
-	{
-		IdType:       "sdk.SchemaObjectIdentifier",
-		ObjectStruct: sdk.Pipe{},
-	},
```
sfc-gh-jcieslak, line 199: I don't think this file has correctly resolved conflicts

examples/resources/snowflake_postgres_instance/resource.tf
```diff
@@ -0,0 +1,18 @@
...
+}
+
+# complete resource
+resource "snowflake_postgres_instance" "complete" {
+  name                     = "my_postgres_instance"
+  compute_family           = "STANDARD_1"
+  storage_size_gb          = 10
+  authentication_authority = "POSTGRES"
+  postgres_version         = 16
+  high_availability        = true
```
sfc-gh-jcieslak: in complete example we usually include all optional fields (I can see we're missing the fields such as storage_integration or postgres_settings, and other)

examples/resources/snowflake_postgres_fork/resource.tf
```diff
@@ -0,0 +1,27 @@
...
+resource "snowflake_postgres_fork" "complete" {
+  name      = "my_postgres_fork_complete"
+  fork_from = "my_source_postgres_instance"
+  at {
+    offset = "-3600"
+  }
+  compute_family    = "STANDARD_1"
+  storage_size_gb   = 20
+  high_availability = true
+  comment           = "Forked for testing"
```
sfc-gh-jcieslak, line 45: here, let's also add missing optional fields

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+		return nil, err
+	}
+
+	pi, err := client.PostgresInstances.ShowByID(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set fork_from from the origin field if available
+	forkFrom := ""
```
sfc-gh-jcieslak: Wouldn't it mean the postgres instance is not a fork and the user is trying to import the object into wrong resource? If origin should be there for forked postgres instance, we should prevent users from being able to import objects without it.

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+		return diag.FromErr(err)
+	}
+
+	request := sdk.NewForkPostgresInstanceRequest(id, forkFromId)
+
+	// Handle AT time travel
+	if v := d.Get("at").([]any); len(v) > 0 {
+		atConfig := v[0].(map[string]any)
+		at := sdk.NewPostgresInstanceForkAtRequest()
+		if ts := atConfig["timestamp"].(string); len(ts) > 0 {
```
sfc-gh-jcieslak: if the map value is of type `any` and a value for a given key won't be there, wouldn't we panic on trying to do `(nil).(string)`? I think we should use the following syntax `value, ok := conf[key]; ok && len(value) > 0=

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+			at.WithOffset(offset)
+		}
+		request.WithAt(*at)
+	}
+
+	// Handle BEFORE time travel
+	if v := d.Get("before").([]any); len(v) > 0 {
+		beforeConfig := v[0].(map[string]any)
+		before := sdk.NewPostgresInstanceForkBeforeRequest()
+		if ts := beforeConfig["timestamp"].(string); len(ts) > 0 {
```
sfc-gh-jcieslak: same as in `at` case

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+
+	// Handle optional fork-time parameters
+	errs := errors.Join(
+		stringAttributeCreateBuilder(d, "compute_family", request.WithComputeFamily),
+		booleanStringAttributeCreateBuilder(d, "high_availability", request.WithHighAvailability),
+		stringAttributeCreateBuilder(d, "postgres_settings", request.WithPostgresSettings),
+		stringAttributeCreateBuilder(d, "comment", request.WithComment),
+	)
+	if v, ok := d.GetOk("storage_size_gb"); ok {
+		request.WithStorageSizeGb(v.(int))
```
sfc-gh-jcieslak: nit: Why is this not a part of the above builder helpers group? I think int helper should be there or we could add one.

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+	if errs != nil {
+		return diag.FromErr(errs)
+	}
+
+	if err := client.PostgresInstances.Fork(ctx, request); err != nil {
+		return diag.FromErr(fmt.Errorf("error creating forked Postgres instance %s: %w", id.FullyQualifiedName(), err))
+	}
+
+	if err := util.Retry(5, 3*time.Second, func() (error, bool) {
```
sfc-gh-jcieslak: As it tests, this should be most likely a much larger timeout, or not in this case?

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+
+func ReadPostgresForkFunc(withExternalChangesMarking bool) schema.ReadContextFunc {
+	return func(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+		client := meta.(*provider.Context).Client
+		id, err := sdk.ParseAccountObjectIdentifier(d.Id())
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		pi, err := client.PostgresInstances.ShowByIDSafely(ctx, id)
```
sfc-gh-jcieslak, line 284: After creating or updating, this should be done or retry or we should have a retry logic in Create and Update, as the delays are strictly related to Create or Update operations? I think the latter, as refreshes without any updates would be as fast as other, regular resource, which means only creation or update would contribute to long application times.

sfc-gh-jcieslak: @sfc-gh-jmichalak we should discuss this as it's a pretty uncommon case in the provider

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+	)
+	if errs != nil {
+		return diag.FromErr(errs)
+	}
+
+	if err := client.PostgresInstances.Create(ctx, request); err != nil {
+		return diag.FromErr(fmt.Errorf("error creating Postgres instance %s: %w", id.FullyQualifiedName(), err))
+	}
+
+	if err := util.Retry(5, 3*time.Second, func() (error, bool) {
```
sfc-gh-jcieslak, line 205: Some of the comments (if not all) from the fork resource will also apply here. Here it depends on what the timeouts for both should be (I'm guessing both may contain different timeouts due to the amount of job they have to do).

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+			return diag.FromErr(fmt.Errorf("error unsetting Postgres instance properties: %w", err))
+		}
+	}
+
+	return ReadPostgresInstanceFunc(false)(ctx, d, meta)
+}
+
+// normalizePostgresSettings returns nil if the postgres_settings value is
+// an empty JSON object ("{}"), treating it as unset.
+func normalizePostgresSettings(s *string) *string {
```
sfc-gh-jcieslak: We usually move such handling to the SDK level (e.g., treating `{}` output as nil during the conversion from sql to data structs; in the `convert()` function)

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+	if s == nil || *s == "{}" {
+		return nil
+	}
+	return s
+}
+
+// setOptionalFromNonEmptyStringPtr sets a key in resource data only if the
+// pointer is non-nil and the pointed-to string is non-empty.
+func setOptionalFromNonEmptyStringPtr(d *schema.ResourceData, key string, ptr *string) error {
```
sfc-gh-jcieslak: this and below helper have their place in one of the common helper files (`helpers.go` I think) among other similar helpers

pkg/resources/validators.go
```diff
@@ -138,3 +139,20 @@ func IsValidListingName(i any, path cty.Path) diag.Diagnostics {
 
 	return nil
 }
+
+func isIntBetween(min, max int) schema.SchemaValidateDiagFunc {
```
sfc-gh-jcieslak: I think there's already existing `validation.IntBetween()` provided by HashiCorp

pkg/sdk/postgres_instances_ext.go
```diff
@@ -1,9 +1,133 @@
...
+				}
+			}
+		case "comment":
+			details.Comment = String(prop.Value)
+		case "network_policy":
+			if prop.Value != "" {
+				details.NetworkPolicy = Pointer(NewAccountObjectIdentifier(prop.Value))
+			}
+		case "postgres_settings":
+			details.PostgresSettings = String(prop.Value)
```
sfc-gh-jcieslak, line 113: Here you could handle `{}` postgres settings

pkg/sdk/postgres_instances_identifier_test.go
```diff
@@ -0,0 +1,219 @@
+package sdk
```
sfc-gh-jcieslak: This file is a duplicate of pkg/sdk/postgres_instances_gen_test.go. If any of these tests introduce any value over what's in pkg/sdk/postgres_instances_gen_test.go, they should be moved there. Every object contains one file for all unit tests, let's not divide them into multiple files for postgres.

pkg/testacc/2_check_destroy_test.go
```diff
@@ -162,6 +162,10 @@ func asId[T supportedIdentifierTypes](id sdk.ObjectIdentifier) (*T, error) {
 	}
 }
 
+var showPostgresInstance = func(ctx context.Context, client *sdk.Client, id sdk.ObjectIdentifier) error {
```
sfc-gh-jcieslak: nit: this kind of repetition is not that bad, I would inline it for consistency within the file.

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+// createSourceForFork creates a postgres instance suitable for forking and registers cleanup.
+// It waits for the instance to be fully ready to accept fork operations.
+func createSourceForFork(t *testing.T) sdk.AccountObjectIdentifier {
```
sfc-gh-jcieslak: This could be moved to the test client helper as CreateSourceForFork, then we wouldn't have to introduce WaitForForkReady which is a bit confusing and we would be able to reuse this helper in both integration and acceptance tests.

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,203 @@
...
+	assertBasic := []assert.TestCheckFuncProvider{
+		resourceassert.PostgresForkResource(t, modelBasic.ResourceReference()).
+			HasNameString(forkId.Name()).
+			HasForkFromString(sourceId.Name()).
+			HasNoComment().
+			HasFullyQualifiedNameString(forkId.FullyQualifiedName()),
+		postgresShowOutputBaseAssert(t, modelBasic.ResourceReference(), forkId.Name()),
+	}
+
+	modelWithComment := model.PostgresFork("test", forkId.Name(), sourceId.Name()).
```
sfc-gh-jcieslak: as second modle we usually choose resource with all optional (and non-ForceNew) fields set. All optional and ForceNew fields should be a part of `CompleteUseCase` test

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,203 @@
...
+	t.Helper()
+	sourceId := testClient().Ids.RandomAccountObjectIdentifier()
+	_, sourceCleanup := testClient().PostgresInstance.CreateWithRequest(t,
+		sdk.NewCreatePostgresInstanceRequest(sourceId, "STANDARD_M", 10, sdk.PostgresInstanceAuthenticationAuthorityPostgres))
+	t.Cleanup(sourceCleanup)
+	testClient().PostgresInstance.WaitForForkReady(t, sourceId, 5*time.Minute)
+	return sourceId
+}
+
+func TestAcc_PostgresFork_Basic(t *testing.T) {
```
sfc-gh-jcieslak: `Basic` -> `BasicUseCase` the naming has to be specific for our test runner (we choose some tests based on the name)

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,203 @@
...
+				Config:  accconfig.FromModels(t, modelBasic),
+				Check: assertThat(t,
+					invokeactionassert.PostgresInstanceDoesNotExist(t, forkId),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_PostgresFork_Complete(t *testing.T) {
```
sfc-gh-jcieslak: `Complete` -> `CompleteUseCase`

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,203 @@
...
+	comment := random.Comment()
+
+	// Use a recent timestamp (1 hour ago) — Snowflake requires fork timestamps within the past 10 days
+	recentTimestamp := time.Now().Add(-1 * time.Hour).UTC().Format("2006-01-02 15:04:05")
+
+	modelComplete := model.PostgresFork("test", forkId.Name(), sourceId.Name()).
+		WithComment(comment).
+		WithAtTimestamp(recentTimestamp).
+		WithHighAvailability(false).
+		WithPostgresSettings(`{"work_mem": "64MB"}`)
```
sfc-gh-jcieslak, line 171: include all possible fields

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,203 @@
...
+	sourceId := createSourceForFork(t)
+	forkId := testClient().Ids.RandomAccountObjectIdentifier()
+	comment := random.Comment()
+
+	// Use a recent timestamp (1 hour ago) — Snowflake requires fork timestamps within the past 10 days
+	recentTimestamp := time.Now().Add(-1 * time.Hour).UTC().Format("2006-01-02 15:04:05")
+
+	modelComplete := model.PostgresFork("test", forkId.Name(), sourceId.Name()).
+		WithComment(comment).
+		WithAtTimestamp(recentTimestamp).
```
sfc-gh-jcieslak: nit: It could be more ergonomic if the `WithAtTimestamp` would handle those utc + format conversions. The function could take in just time.Time

pkg/testacc/resource_postgres_instance_acceptance_test.go
```diff
@@ -0,0 +1,288 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+// postgresShowOutputBaseAssert returns a show output assertion with the common fields
+// that every postgres instance (including forks) should have.
+func postgresShowOutputBaseAssert(t *testing.T, ref string, name string) *resourceshowoutputassert.PostgresInstanceShowOutputAssert {
```
sfc-gh-jcieslak: let's not do this as it "pollutes" the package context, let's just inline this, the repetition of this kind is more readable

pkg/testacc/resource_postgres_instance_acceptance_test.go
```diff
@@ -0,0 +1,288 @@
...
+	t.Helper()
+	return resourceshowoutputassert.PostgresInstanceShowOutput(t, ref).
+		HasCreatedOnNotEmpty().
+		HasName(name).
+		HasOwner(snowflakeroles.Accountadmin.Name()).
+		HasOwnerRoleType("ROLE")
+}
+
+func TestAcc_PostgresInstance_Basic(t *testing.T) {
```
sfc-gh-jcieslak: apply all the comments as in other acceptance test file.

pkg/testacc/resource_postgres_instance_acceptance_test.go
```diff
@@ -0,0 +1,288 @@
...
+			{
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(modelBasic.ResourceReference(), plancheck.ResourceActionUpdate),
+					},
+				},
+				Config: accconfig.FromModels(t, modelBasic),
+				Check:  assertThat(t, assertAfterUnset...),
+			},
+			// Step 5: External change detection
```
sfc-gh-jcieslak: Please, look at other files that contain `BasicUseCase` and `CompleteUseCase` to understand what operations are required (what's tested) in each. You can read this PR's description to find more info about them: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4113

sfc-gh-jcieslak (changes requested):

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+	"postgres_settings": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies custom Postgres settings as a JSON string.",
+	},
+	"maintenance_window_start": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ValidateDiagFunc: isIntBetween(0, 23),
+		Description:      "Specifies the hour (0-23 UTC) at which the maintenance window starts.",
```
sfc-gh-jcieslak, line 92: You're missing a DiffSuppressFunc (`IgnoreChangeToCurrentSnowflakeValueInDesc`)

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+			setOptionalFromAccountObjectIdentifierPtr(d, "storage_integration", details.StorageIntegration),
+			d.Set(ShowOutputAttributeName, []map[string]any{schemas.PostgresInstanceToSchema(pi)}),
+			d.Set(DescribeOutputAttributeName, []map[string]any{schemas.PostgresInstanceDetailsToSchema(details)}),
+			d.Set(FullyQualifiedNameAttributeName, id.FullyQualifiedName()),
+		)
+		if errs != nil {
+			return diag.FromErr(errs)
+		}
+
+		if details.MaintenanceWindowStart != 0 || d.Get("maintenance_window_start").(string) != "" {
```
sfc-gh-jcieslak: remove this as the maintenance_window_start is already handled in line 247 in `if withExternalChangesMarking { }` block

sfc-gh-jcieslak (dismissed):

pkg/acceptance/bettertestspoc/assert/resourceassert/postgres_fork_resource_ext.go
```diff
@@ -0,0 +1,10 @@
+package resourceassert
+
+import (
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/assert"
+)
+
+func (p *PostgresForkResourceAssert) HasAuthenticationAuthorityString(expected string) *PostgresForkResourceAssert {
```
sfc-gh-jmichalak: This should be removed - the fork resource doesn't have this field.

pkg/acceptance/helpers/postgres_instance_client.go
sfc-gh-jmichalak: @sfc-gh-jsouthard do you have a ticket to track this? How can we be sure that 30s is enough? This solution has a direct customer impact. Also, the current implementation will significantly increase pipeline time, and it's already quite long.

pkg/sdk/postgres_instances_ext.go
```diff
@@ -1,9 +1,133 @@
...
+	"strconv"
+)
+
 func (r *CreatePostgresInstanceRequest) GetName() AccountObjectIdentifier {
 	return r.name
 }
 
+// PostgresInstanceDetails represents the parsed result of DESCRIBE POSTGRES INSTANCE
+type PostgresInstanceDetails struct {
+	Name                         string
```
sfc-gh-jmichalak, line 19: The typing can be a bit better here:

- Name -> AccountObjectIdentifier
- Type, ComputeFamily, State -> probably a separate enum
- CreatedOn, UpdatedOn -> should remain string, as in DESC, SF does not have strict format
- UpdatedOn -> if it can be null, it should be *string.

pkg/sdk/postgres_instances_ext.go
sfc-gh-jmichalak: We should unmarshal it, as this field can be set in this object, and suppress the diff similarly to the recent cortex agent resource.

pkg/sdk/postgres_instances_ext.go
```diff
@@ -1,9 +1,133 @@
...
+
+func (v *postgresInstances) DescribeDetails(ctx context.Context, id AccountObjectIdentifier) (*PostgresInstanceDetails, error) {
+	properties, err := v.Describe(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	return ParsePostgresInstanceDetails(properties)
+}
+
+func (v *PostgresInstance) IsHighlyAvailable() bool {
```
sfc-gh-jmichalak: Nit: this method is unnecessary. Optionally, we can rename the field from IsHa to IsHighlyAvailable.

pkg/sdk/postgres_instances_identifier_test.go
```diff
@@ -0,0 +1,219 @@
...
+			{Property: "storage_integration", Value: ""},
+		}
+		details, err := ParsePostgresInstanceDetails(properties)
+		require.NoError(t, err)
+		// After the change, empty string should not create an identifier
+		assert.Nil(t, details.StorageIntegration)
+	})
+}
+
+func TestPostgresInstances_DtoBuilders_IdentifierTypes(t *testing.T) {
```
sfc-gh-jmichalak: This is unnecessary.

pkg/sdk/postgres_instances_identifier_test.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+		assertOptsValidAndSQLEquals(t, opts,
+			`ALTER POSTGRES INSTANCE %s SET NETWORK_POLICY = "my_policy" AUTHENTICATION_AUTHORITY = POSTGRES_OR_SNOWFLAKE`+
+				` COMMENT = '%s' HIGH_AVAILABILITY = true COMPUTE_FAMILY = 'STANDARD_M' STORAGE_SIZE_GB = 100`+
+				` STORAGE_INTEGRATION = "my_integration" POSTGRES_VERSION = 18 POSTGRES_SETTINGS = '{}'`,
+			id.FullyQualifiedName(), comment)
+	})
+}
+
+func TestPostgresInstances_ParseDetails_IdentifierTypes(t *testing.T) {
```
sfc-gh-jmichalak: It's good that we test the ParseDetails func, but not all data is tested. We can have a single test with all properties and assert that all suported properties are parsed correctly.

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -251,39 +169,45 @@ func TestInt_PostgresInstances(t *testing.T) {
...
-		// }, 5*time.Minute, 3*time.Second)
-		//
-		// assertThatObject(t, objectassert.PostgresInstanceFromObject(t, forkedInstance).
-		//   HasName(forkedId.Name()).
-		//   HasOriginContaining(forkSourceInstance.instance.Name),
-		// )
-
-		forkedInstance, showErr := client.PostgresInstances.ShowByID(ctx, forkedId)
-		require.NoError(t, showErr)
+		testClientHelper().PostgresInstance.WaitForReady(t, forkId, 5*time.Minute)
```
sfc-gh-jmichalak: Not WaitForForkReady?

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -251,39 +169,45 @@ func TestInt_PostgresInstances(t *testing.T) {
...
 	})
 
 	// Doc example: CREATE POSTGRES INSTANCE my_fork FORK my_source_instance AT (TIMESTAMP => '2025-01-15 12:00:00'::TIMESTAMP_NTZ);
 	t.Run("fork - with time travel options", func(t *testing.T) {
-		sourceInstance := sharedInstance.instance
+		sourceInstance, sourceCleanup := testClientHelper().PostgresInstance.Create(t)
+		t.Cleanup(sourceCleanup)
 
 		// AT with timestamp
 		forkId1 := testClientHelper().Ids.RandomAccountObjectIdentifier()
```
sfc-gh-jmichalak, line 307: <p class='graphite__hidden'><i>[Re: line +217]</i></p>
So, maybe we shouldn't have a fixed timestamp, but just select a date inside our account retention time days? And expect this operation to succeed. Same in the other ones.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/snowflakedb/terraform-provider-snowflake/4759?utm_source=unchanged-line-comment">Graphite</a>.</p>

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -432,24 +358,24 @@ func TestInt_PostgresInstances(t *testing.T) {
 		}, 3*time.Minute, 5*time.Second)
 		require.NoError(t, err)
 
-		// Set postgres settings separately (cannot be combined with COMPUTE_FAMILY/STORAGE_SIZE_GB/HIGH_AVAILABILITY)
-		// Retry because a previous operation may still be in progress
+		// Set HIGH_AVAILABILITY separately (cannot be combined with COMPUTE_FAMILY/STORAGE_SIZE_GB)
```
sfc-gh-jmichalak: These fields should be mutually exclusive in the SDK - use `WithValidation...`

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+		Type:             schema.TypeString,
+		Optional:         true,
+		Description:      "Specifies the storage integration for the Postgres instance.",
+		ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"postgres_settings": {
+		Type:        schema.TypeString,
+		Optional:    true,
+		Description: "Specifies custom Postgres settings as a JSON string.",
```
sfc-gh-jmichalak: As I noted above, this should have a diff suppress to ignore changes on whitespaces, etc.

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+		},
+	},
+	DescribeOutputAttributeName: {
+		Type:        schema.TypeList,
+		Computed:    true,
+		Description: "Outputs the result of `DESCRIBE POSTGRES INSTANCE` for the given Postgres instance.",
+		Elem: &schema.Resource{
+			Schema: schemas.DescribePostgresInstanceSchema,
+		},
+	},
```
sfc-gh-jmichalak, line 116: Q: What about detecting external changes on postgres types? I mean, when you import a forked instance into this resource, what is the expected behavior?

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+			"compute_family",
+			"high_availability",
+		}); err != nil {
+			return diag.FromErr(err)
+		}
+
+		postgresSettings := normalizePostgresSettings(pi.PostgresSettings)
+		errs := errors.Join(
+			d.Set("name", pi.Name),
+			d.Set("compute_family", pi.ComputeFamily),
```
sfc-gh-jmichalak: Some fields are already set in withExternalChangesMarking block and should be removed from here.

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,418 @@
...
+
+	if !reflect.DeepEqual(highAvailabilitySet, &sdk.PostgresInstanceSetRequest{}) {
+		alterReq := sdk.NewAlterPostgresInstanceRequest(id).WithSet(*highAvailabilitySet)
+		if err := client.PostgresInstances.Alter(ctx, alterReq); err != nil {
+			return diag.FromErr(fmt.Errorf("error setting Postgres instance high_availability: %w", err))
+		}
+	}
+
+	// Unset operations (non-conflicting, can be sent together)
+	if !reflect.DeepEqual(pgSettingsUnset, &sdk.PostgresInstanceUnsetRequest{}) {
```
sfc-gh-jmichalak: Nit: if they aren't conflicting, let's actually make one call, not two.

pkg/resources/postgres_instance.go
sfc-gh-jmichalak: Yes, take a look at diff suppression implementation for cortex agent resource.

pkg/resources/postgres_instance.go
sfc-gh-jmichalak: The setOptionalFromNonEmptyStringPtr func is not even used.

pkg/resources/postgres_fork.go
sfc-gh-jmichalak: I think the retries in Read are unnecessary. I assume they are only needed to wait for an effect after modifying a resource, so they should be done in Create/Update. Optionally, we could extract the `d.Set` calls and just pass the successful result from Create/Update's ShowByID, but maybe later.

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+		if err = setStateToValuesFromConfig(d, postgresForkSchema, []string{
+			"compute_family",
+			"high_availability",
+		}); err != nil {
+			return diag.FromErr(err)
+		}
+
+		errs := errors.Join(
+			d.Set("name", pi.Name),
+			d.Set("compute_family", pi.ComputeFamily),
```
sfc-gh-jmichalak, line 323: Ditto.

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+
+	if !reflect.DeepEqual(highAvailabilitySet, &sdk.PostgresInstanceSetRequest{}) {
+		alterReq := sdk.NewAlterPostgresInstanceRequest(id).WithSet(*highAvailabilitySet)
+		if err := client.PostgresInstances.Alter(ctx, alterReq); err != nil {
+			return diag.FromErr(fmt.Errorf("error setting Postgres instance high_availability: %w", err))
+		}
+	}
+
+	// Unset operations
+	if !reflect.DeepEqual(pgSettingsUnset, &sdk.PostgresInstanceUnsetRequest{}) {
```
sfc-gh-jmichalak, line 456: Ditto.

pkg/testacc/resource_postgres_fork_acceptance_test.go
sfc-gh-jmichalak: Or even better: extend the SDK interface with `CreateSafely` or something, that would encapsulate the reading logic. So, we would have the retry logic in one place.

pkg/testacc/resource_postgres_instance_acceptance_test.go
```diff
@@ -0,0 +1,288 @@
...
+						HasComputeFamilyString("STANDARD_M").
+						HasStorageSizeGbString("10").
+						HasAuthenticationAuthorityString("POSTGRES"),
+					postgresShowOutputBaseAssert(t, modelBasic.ResourceReference(), id.Name()).
+						HasComputeFamily("STANDARD_M").
+						HasAuthenticationAuthority("POSTGRES"),
+				),
+			},
+			// rename object - Snowflake backend does not currently support
+			// ALTER POSTGRES INSTANCE ... RENAME TO (returns SQL compilation error)
```
sfc-gh-jmichalak, line 283: Do you have the timeline for supporting this? If this is not yet available, this field should be marked as ForceNew in both resources.

examples/resources/snowflake_postgres_instance/resource.tf
```diff
@@ -0,0 +1,18 @@
...
+  name                     = "my_postgres_instance"
+  compute_family           = "STANDARD_1"
+  storage_size_gb          = 10
+  authentication_authority = "POSTGRES"
+}
+
+# complete resource
+resource "snowflake_postgres_instance" "complete" {
+  name                     = "my_postgres_instance"
+  compute_family           = "STANDARD_1"
```
sfc-gh-jmichalak: `STANDARD_1` family is missing from the docs. Please adjust all examples to use a real compute_family.

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,441 @@
...
+	pgSettingsSet := sdk.NewPostgresInstanceSetRequest()
+	pgSettingsUnset := sdk.NewPostgresInstanceUnsetRequest()
+	errs := errors.Join(
+		stringAttributeUpdate(d, "postgres_settings", &pgSettingsSet.PostgresSettings, &pgSettingsUnset.PostgresSettings),
+	)
+	if errs != nil {
+		return diag.FromErr(errs)
+	}
+
+	if !reflect.DeepEqual(pgSettingsSet, &sdk.PostgresInstanceSetRequest{}) {
```
sfc-gh-jmichalak, line 377: Q: Shouldn't all of these alters be wrapped by this retry logic? In integration tests, they are. Same for the other resource.

sfc-gh-jmichalak (changes requested):

pkg/resources/helpers.go
```diff
@@ -301,3 +301,25 @@ type PlanCheckFunc func(ctx context.Context, req plancheck.CheckPlanRequest, res
...
+	normalized, err := sdk.NormalizePostgresSettings(*s)
+	if err != nil || normalized == "" {
+		return nil
+	}
+	return &normalized
+}
+
+// setOptionalFromAccountObjectIdentifierPtr sets a key in resource data only if the
+// pointer is non-nil.
+func setOptionalFromAccountObjectIdentifierPtr(d *schema.ResourceData, key string, ptr *sdk.AccountObjectIdentifier) error {
```
sfc-gh-jcieslak: nit: this one seems like we could move (or already have in one of our resource_helpers_x.go files)

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,400 @@
...
+		Type:             schema.TypeString,
+		Optional:         true,
+		Description:      "Specifies the storage integration for the Postgres instance.",
+		ValidateDiagFunc: IsValidIdentifier[sdk.AccountObjectIdentifier](),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"postgres_settings": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		DiffSuppressFunc: NormalizeAndCompare(sdk.NormalizePostgresSettings),
```
sfc-gh-jcieslak, line 84: nit: we could add validation that would call sdk.NormalizePostgresSettings (like in other cases where enums are used)

pkg/resources/postgres_instance.go
```diff
@@ -0,0 +1,400 @@
...
+			"storage_size_gb",
+			"postgres_version",
+			"network_policy",
+			"storage_integration",
+			"maintenance_window_start",
+		}); err != nil {
+			return diag.FromErr(err)
+		}
+
+		postgresSettings := normalizePostgresSettings(pi.PostgresSettings)
```
sfc-gh-jcieslak: nit: I think that sdk.NormalizePostgresSettings should be used here. The normalizePostgresSettings should be removed and the sdk function should be adjusted to account for cases normalizePostgresSettings covers.

pkg/sdk/postgres_instances_ext.go
```diff
@@ -1,5 +1,153 @@
...
+	PostgresSettings             *string
+	StorageIntegration           *AccountObjectIdentifier
+}
+
+// ParsePostgresInstanceDetails parses []PostgresInstanceProperty into PostgresInstanceDetails
+func ParsePostgresInstanceDetails(properties []PostgresInstanceProperty) (*PostgresInstanceDetails, error) {
+	details := &PostgresInstanceDetails{}
+	var errs []error
+	for _, prop := range properties {
+		switch prop.Property {
```
sfc-gh-jcieslak: Hmm usually property names are uppercase, could we just make sure the casing would match and call ToLower on prop.Property to ensure we always match no matter the casing convention

pkg/sdk/postgres_instances_gen_test.go
```diff
@@ -452,3 +456,73 @@ func TestPostgresInstances_Describe(t *testing.T) {
 		assertOptsValidAndSQLEquals(t, opts, "DESCRIBE POSTGRES INSTANCE %s", id.FullyQualifiedName())
 	})
 }
+
+func TestPostgresInstances_ParseDetails(t *testing.T) {
```
sfc-gh-jcieslak: nit: Those could be extracted to ext_test.go file to avoid regenerations

MIGRATION_GUIDE.md
sfc-gh-jcieslak: Your entry should be moved up as it's in the section for already released version, you can bring it up as first entry for the new version

pkg/resources/postgres_fork.go
sfc-gh-jcieslak: It seems the retry was removed, is it intended?

pkg/acceptance/helpers/postgres_instance_client.go
sfc-gh-jsouthard: This approach has been replaced by having the create perform retries similar to the update operation.
The lack of availability to perform a fork is independent of time and is reflective of whether the backend can perform it at the request time. The only way to check is to attempt the fork and see if it fails.

pkg/resources/helpers.go
sfc-gh-jsouthard: Actually, it's unused, so it's been removed.

pkg/resources/postgres_fork.go
sfc-gh-jsouthard: What is this comment left to address? It says ditto, but I can't see the related thing it is dittoing.

pkg/resources/postgres_fork.go
sfc-gh-jsouthard: I believe this is resolved in a8645f

pkg/resources/postgres_fork.go
sfc-gh-jsouthard: I'm unsure what this ditto is covering, from proximity it looks like retries on alters, which was addressed in a8645f

pkg/resources/postgres_fork.go
sfc-gh-jsouthard: I'm not certain what the request is here? Are they tied to old logic that went away in newer commits?

pkg/resources/postgres_instance.go
sfc-gh-jsouthard: I believe this comment was originally for the comment field which now has a DiffSuppressFunc.
This doesn't seem appropriate for maintenance_window_start

pkg/resources/postgres_instance.go
sfc-gh-jsouthard: A guard against importing a fork as an instance was added in 21af3f

pkg/resources/postgres_instance.go
sfc-gh-jsouthard: moved functionality to the SDK layer, but a direct swap doesn't quite work due to the value/pointer mechanics of other dependent call sites

pkg/sdk/testint/postgres_instances_gen_integration_test.go
sfc-gh-jsouthard: This has been updated to calculate a timestamp in the valid window instead of a fixed timestamp

pkg/sdk/postgres_instances_ext.go
sfc-gh-jsouthard: Is this a critical change for initial release of the resources?

pkg/sdk/postgres_instances_ext.go
sfc-gh-jsouthard: updated

pkg/sdk/postgres_instances_ext.go
sfc-gh-jsouthard: The approach taken was similar to cortex agent with the diff suppression and normalization calling

pkg/sdk/postgres_instances_gen_test.go
sfc-gh-jsouthard: This and other non-generated helpers have been moved.

pkg/testacc/resource_postgres_fork_acceptance_test.go
sfc-gh-jcieslak: nit: do we need something similar for update?

pkg/resources/postgres_fork.go
```diff
@@ -254,8 +254,19 @@ func CreatePostgresFork(ctx context.Context, d *schema.ResourceData, meta any) d
 		return diag.FromErr(errs)
 	}
 
-	if err := client.PostgresInstances.Fork(ctx, request); err != nil {
-		return diag.FromErr(fmt.Errorf("error creating forked Postgres instance %s: %w", id.FullyQualifiedName(), err))
+	var lastForkErr error
+	if retryErr := util.Retry(24, 5*time.Second, func() (error, bool) {
```
sfc-gh-jcieslak, line 258: In the integration tests, as far as I remember, we were also waiting after ALTER statements, so we should probably wait in the Update method as well.

pkg/sdk/postgres_instances_gen.go
```diff
@@ -19,6 +19,10 @@ type PostgresInstances interface {
 	ShowByIDSafely(ctx context.Context, id AccountObjectIdentifier) (*PostgresInstance, error)
 	Describe(ctx context.Context, id AccountObjectIdentifier) ([]PostgresInstanceProperty, error)
 	DescribeDetails(ctx context.Context, id AccountObjectIdentifier) (*PostgresInstanceDetails, error)
+	// CreateSafely creates the instance and polls until it reaches READY state.
+	// The caller should set a deadline on ctx via context.WithTimeout.
+	// Implemented in postgres_instances_ext.go.
+	CreateSafely(ctx context.Context, request *CreatePostgresInstanceRequest) (*PostgresInstance, error)
```
sfc-gh-jcieslak, line 25: It should be used in resources. Ideally, I think we should have: CreateSafely, ForkSafely, and UpdateSafely that would implement this timeout logic inside.

MIGRATION_GUIDE.md
sfc-gh-jsouthard: I think that is another casualty of the merge fix, will address

pkg/resources/postgres_instance.go
sfc-gh-jcieslak: This one still applies. It applies to every field that in Read function is handled within `if withExternalChangesMarking {` block.

pkg/sdk/postgres_instances_gen_test.go
```diff
@@ -206,160 +142,69 @@ func TestPostgresInstances_Alter(t *testing.T) {
...
-				Value: "value1",
-			},
-		}
-		assertOptsValidAndSQLEquals(t, opts, `ALTER POSTGRES INSTANCE %s SET TAG "tag1" = 'value1'`, id.FullyQualifiedName())
+		opts.Suspend = Bool(true)
+		assertOptsValidAndSQLEquals(t, opts, "ALTER POSTGRES INSTANCE %s SUSPEND", id.FullyQualifiedName())
 	})
 
-	t.Run("unset tags", func(t *testing.T) {
+	t.Run("all options", func(t *testing.T) {
```
sfc-gh-jmichalak, line 178: Other operations are missing - reset access, rename...

pkg/sdk/generator/defs/postgres_instances_def.go
```diff
@@ -20,6 +20,12 @@ var (
 		"PostgresInstanceResetAccessRole", "PostgresInstanceResetAccessRoles",
 		"snowflake_admin", "application",
 	)
+	PostgresInstanceComputeFamilyEnumDef = g.NewEnum(
+		"PostgresInstanceComputeFamily", "PostgresInstanceComputeFamilies",
+		"STANDARD_M", "STANDARD_L", "STANDARD_XL", "STANDARD_2XL", "STANDARD_4XL", "STANDARD_8XL", "STANDARD_12XL", "STANDARD_24XL",
+		"HIGHMEM_L", "HIGHMEM_XL", "HIGHMEM_2XL", "HIGHMEM_4XL", "HIGHMEM_8XL", "HIGHMEM_12XL", "HIGHMEM_16XL", "HIGHMEM_24XL", "HIGHMEM_32XL", "HIGHMEM_48XL",
+		"BURST_XS", "BURST_S", "BURST_M",
```
sfc-gh-jmichalak, line 27: Please update the [docs](https://docs.snowflake.com/en/user-guide/snowflake-postgres/postgres-instance-sizes), as BURST_XS is missing.

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -394,6 +395,17 @@ func TestInt_PostgresInstances(t *testing.T) {
 		assert.Error(t, err)
 	})
 
+	t.Run("fork - validation: At and Before are mutually exclusive", func(t *testing.T) {
```
sfc-gh-jmichalak, line 398: This should be moved to unit tests.

pkg/sdk/postgres_instances_ext.go
```diff
@@ -1,5 +1,201 @@
...
+		default:
+		}
+		instance, err := doShowByID()
+		if err != nil {
+			return nil, err
+		}
+		if instance.State == PostgresInstanceStateReady {
+			return instance, nil
+		}
+		time.Sleep(3 * time.Second)
```
sfc-gh-jmichalak, line 160: Nits:

- doCreate and doShowByID should take a ctx as a parameter
- time.Sleep ignores context cancellation

pkg/sdk/postgres_instances_ext_test.go
```diff
@@ -0,0 +1,195 @@
...
+		require.NoError(t, err)
+		assert.Equal(t, PostgresInstanceStateReady, instance.State)
+		assert.Equal(t, 3, stub.showIdx)
+	})
+
+	t.Run("returns error when context is canceled before READY", func(t *testing.T) {
+		stub := &stubPostgresInstances{
+			showStates: []PostgresInstanceState{PostgresInstanceStateCreating},
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Millisecond)
```
sfc-gh-jmichalak, line 181: This should be done with `context.WithCancel`.

pkg/resources/postgres_fork.go
```diff
@@ -0,0 +1,495 @@
...
+	},
+	"high_availability": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ValidateDiagFunc: validateBooleanString,
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("is_ha"),
+		Description:      booleanStringFieldDescription("Specifies whether the Postgres instance should be configured for high availability."),
+		Default:          BooleanDefault,
+	},
+	"postgres_settings": {
```
sfc-gh-jmichalak, line 110: Nit: a number of these fields are common and should be extracted.

pkg/resources/postgres_fork.go
sfc-gh-jmichalak: The compute_family, storage size, and high_availability are already handled above. They should be removed from here.

pkg/resources/postgres_fork.go
sfc-gh-jmichalak: We are unsetting twice, when in the regular resource we can do it in one call - please verify.

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/snowflakeroles"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_PostgresFork_BasicUseCase(t *testing.T) {
+	t.Skip("TODO: Fix failing test")
```
sfc-gh-jmichalak, line 28: What about this TODO?

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,261 @@
...
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionUpdate),
+					},
+				},
+				Check: assertThat(t, assertWithComment...),
+			},
+			// Step 4: External drift — alter comment externally, Terraform reverts to configured value
+			{
+				PreConfig: func() {
+					testClient().PostgresInstance.Alter(t, sdk.NewAlterPostgresInstanceRequest(forkId).WithSet(
+						*sdk.NewPostgresInstanceSetRequest().
```
sfc-gh-jmichalak, line 128: Please verify all fields here that we can detect.

pkg/testacc/resource_postgres_fork_acceptance_test.go
```diff
@@ -0,0 +1,261 @@
+//go:build account_level_tests
```
sfc-gh-jmichalak, line 1: Why account-level tests? We don't modify the account here, so it should be non-account level. Same in the other resource.

sfc-gh-jmichalak (dismissed):
This is added only to the test provider. Before releasing it, we must:
- unskip and fix all integration and acceptance provider tests
- fix the renaming issue - I saw the failure is asserted. Instead, we should just use ForceNew on this field.
The remaining comments are optional and can be addressed on our end.

sfc-gh-jmichalak (approved):

PR #4763, opened 2026-05-21
chore: Add inline and out-of-line constraints to iceberg tables SDK

sfc-gh-jmichalak:
## Summary

- Adds inline constraint support to `IcebergTableColumn` and `IcebergTableAddColumnAction` (UNIQUE / PRIMARY KEY, FOREIGN KEY, CHECK), with the full set of trailing modifiers (ENFORCED, DEFERRABLE, INITIALLY, ENABLE/DISABLE, VALIDATE, RELY, and their negations).
- Adds out-of-line constraint support to `IcebergTableColumnsAndConstraints` with the same `UniquePK | FK | CH` shape, plus `COMMENT` on UNIQUE/PK/FK variants.
- Replaces the previous flat `TableOutOfLineConstraint` (with `ConstraintType` + `ForeignKey` fields) with the oneOf shape, mirroring the inline form.
- Adds `TableColumnInlineConstraintFromRequest` / `TableOutOfLineConstraintFromRequest` helpers in `iceberg_tables_ext.go`; `toOpts` for both CREATE and ALTER ADD COLUMN call the helper.
- Fix a bug in generator which produced unnecessary test cases for custom validatios


pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -38,18 +38,15 @@ var icebergTableColumn = g.NewQueryStruct("IcebergTableColumn").
 	PredefinedQueryStructField("ColumnType", "datatypes.DataType", g.ParameterOptions().NoQuotes().NoEquals().Required()).
 	PredefinedQueryStructField("DefaultValue", g.KindOfTPointer[sdkcommons.ColumnDefaultValue](), g.KeywordOptions()).
 	OptionalSQL("NOT NULL").
```
sfc-gh-jcieslak, line 40: nit: in docs `not null` is after inline constraints, but that may not matter

pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -129,8 +126,7 @@ var icebergTableAddColumnAction = g.NewQueryStruct("IcebergTableAddColumnAction"
 	OptionalSQL("IF NOT EXISTS").
 	Text("Name", g.KeywordOptions().Required().DoubleQuotes()).
 	PredefinedQueryStructField("ColumnType", "datatypes.DataType", g.ParameterOptions().NoQuotes().NoEquals().Required()).
-	// TODO(next PR): add inline constraint support
-	// OptionalQueryStructField("InlineConstraint", icebergTableColumnInlineConstraint, g.KeywordOptions()).
+	OptionalQueryStructField("InlineConstraint", tableColumnInlineConstraint(), g.KeywordOptions()).
 	PredefinedQueryStructField("DefaultValue", g.KindOfTPointer[sdkcommons.ColumnDefaultValue](), g.KeywordOptions()).
```
sfc-gh-jcieslak, line 132: Didn't remember the full story of this action, but I'm assuming the MaskingPolicy and ProjectionPolicy work, although they're missing from the docs, and the default value can be after inline constraints? Also, we're missing collation here.

pkg/sdk/iceberg_tables_validations_gen.go
```diff
@@ -23,6 +23,130 @@ func (opts *CreateIcebergTableOptions) validate() error {
 	if everyValueSet(opts.OrReplace, opts.IfNotExists) {
 		errs = append(errs, errOneOf("CreateIcebergTableOptions", "OrReplace", "IfNotExists"))
 	}
+	// Adjusted manually: Columns is a slice, validate inline constraint per element
```
sfc-gh-jcieslak: nit: Adjust according to the newly introduced convention from https://github.com/snowflakedb/terraform-provider-snowflake/pull/4772

pkg/sdk/iceberg_tables_gen_test.go
```diff
@@ -218,6 +437,484 @@ func TestIcebergTables_Create(t *testing.T) {
 			aggregationPolicyId.FullyQualifiedName(),
 			tagId1.FullyQualifiedName(), tagId2.FullyQualifiedName(),
 		)
+		// TODO: can it be all types at once?
```
sfc-gh-jcieslak: ?

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Yes, but that depends on which docs :) See also https://docs.snowflake.com/en/sql-reference/sql/create-table-constraint

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: We're missing it because it doesn't work:

```
SQL compilation error: error line 0 at position -1
Iceberg tables cannot be created with collated columns.
```

Masking policies and projection policies do work, this will be reported to the appropriate team.

pkg/sdk/iceberg_tables_validations_gen.go
sfc-gh-jmichalak: Done. I also fixed a bug where it generated a test expecting ReturnedError method.

pkg/sdk/iceberg_tables_gen_test.go
sfc-gh-jmichalak: Removed.

pkg/sdk/iceberg_tables_ext.go
```diff
@@ -2,6 +2,230 @@ package sdk
...
+				errs = append(errs, errOneOf(chPath, "EnableValidate", "EnableNovalidate"))
+			}
+		}
+	}
+	return JoinErrors(errs...)
+}
+
+func (opts *AlterIcebergTableOptions) additionalValidations() error {
+	var errs []error
+	// Adjusted manually: AlterColumnAction is a slice, validate each element
```
sfc-gh-jcieslak, line 141: nit: it's _ext file so we don't have to mark those as "adjusted manually"

pkg/sdk/iceberg_tables_gen.go
```diff
@@ -34,26 +34,25 @@ type CreateIcebergTableOptions struct {
 	PathLayout            *IcebergTablePathLayout           `ddl:"parameter,no_quotes" sql:"PATH_LAYOUT"`
 	ClusterBy             []string                          `ddl:"keyword,parentheses" sql:"CLUSTER BY"`
 	// Adjusted manually
```
sfc-gh-jcieslak, line 36: Q: what's exactly manually adjusted here?

sfc-gh-jcieslak (approved):

pkg/sdk/iceberg_tables_gen.go
sfc-gh-jmichalak: The type because this identifier needs to be wrapped in `'` additionally.

pkg/sdk/iceberg_tables_ext.go
sfc-gh-jmichalak: I'll remove the note, but IMO we can keep the explanation itself - to understand why this is not generated.

PR #4765, opened 2026-05-22
feat: refresh sdk for api integration

sfc-gh-jcieslak:
## Changes
- Update api integration DEF
  - Migrate Enums to auto-generated
  - Add missing Create options (Git and Mcp with their sub-types)
  - Add missing Alter options (Git and Mcp)
  - Generate convert
- Define parsers for all Describe api providers
- Adjust `0_defs.go` as it wrongly decides based on the field name if a given struct should be added to the "structs to generate" collection or not (it should decide based on type/kind, as the same struct could be referenced in two places with different names, causing the struct to be generated twice)
- Regenerate SDK and adjust unit tests (moved global test values to be test-scoped, which is why I also had to modify notification integration tests)

## Next PRs
- Validate some of the undocumented functionalities (e.g., possible alter commands without being documented)
- Extract the struct for ApiUserAuthentication in the describe outputs, and it contains JSON-like syntax.
- (for later PRs) moving some of the SDK definitions to be reusable by other integrations


pkg/sdk/generator/defs/api_integrations_def.go
```diff
@@ -103,8 +284,12 @@ var apiIntegrationsDef = g.NewInterface(
 					OptionalSQL("API_KEY").
 					OptionalSQL("ENABLED").
 					OptionalSQL("API_BLOCKED_PREFIXES").
+					OptionalSQL("ALLOWED_AUTHENTICATION_SECRETS").
+					OptionalSQL("TLS_TRUSTED_CERTIFICATES").
+					// TODO(next pr): Make unset params for each type? The unset private link should most likely be defined only for types using it
```
sfc-gh-jcieslak: @snowflakedb/terraform-provider wdyt?

pkg/sdk/generator/defs/api_integrations_def.go
```diff
@@ -46,14 +93,95 @@ var apiIntegrationsDef = g.NewInterface(
...
+					SQLWithCustomFieldName("apiProvider", "API_PROVIDER = external_mcp").
+					QueryStructField(
+						"ApiUserAuthentication",
+						apiIntegrationDynamicClientMcpAuthDef,
+						g.ListOptions().SQL("API_USER_AUTHENTICATION =").Parentheses().NoComma(),
+					),
+				g.KeywordOptions(),
+			).
 			ListAssignment("API_ALLOWED_PREFIXES", "ApiIntegrationEndpointPrefix", g.ParameterOptions().Parentheses().Required()).
+			// TODO(next pr): API_BLOCKED_PREFIXES is not supported for Amazon API Gateway, github and maybe other types; in next PR test the compatibility of common params for each api_provider and sub-type
```
sfc-gh-jmichalak, line 177: 1. This should be cleared with the appropriate team.
2. In the current implementation, the field order would be different from that in the docs.

pkg/sdk/generator/defs/api_integrations_def.go
```diff
@@ -6,14 +6,61 @@ import (
...
+		"ApiIntegrationOauthClientAuthMethod", "ApiIntegrationOauthClientAuthMethods",
+		"CLIENT_SECRET_BASIC",
+		"CLIENT_SECRET_POST",
+	)
+)
+
+var apiIntegrationEndpointPrefixDef = g.NewQueryStruct("ApiIntegrationEndpointPrefix").Text("Path", g.KeywordOptions().SingleQuotes().Required())
+
+var apiIntegrationAllowedAuthSecretsDef = g.NewQueryStruct("ApiIntegrationAllowedAuthenticationSecrets").
+	OptionalSQLWithCustomFieldName("AllSecrets", "ALL").
```
sfc-gh-jmichalak, line 33: Q: in the docs, they are lowercase, I guess it doesn't matter?

pkg/sdk/generator/defs/api_integrations_def.go
```diff
@@ -46,14 +93,95 @@ var apiIntegrationsDef = g.NewInterface(
...
+						apiIntegrationAllowedAuthSecretsDef,
+						g.KeywordOptions().SQL("ALLOWED_AUTHENTICATION_SECRETS ="),
+					),
+				g.KeywordOptions(),
+			).
+			OptionalQueryStructField(
+				"GitHttpsApiGithubAppProviderParams",
+				g.NewQueryStruct("GitHttpsApiGithubAppParams").
+					SQLWithCustomFieldName("apiProvider", "API_PROVIDER = git_https_api").
+					SQLWithCustomFieldName("apiUserAuthentication", "API_USER_AUTHENTICATION = (TYPE = SNOWFLAKE_GITHUB_APP)"),
```
sfc-gh-jmichalak: Nit: Let's make it a separate struct to make it more future proof.

pkg/sdk/generator/defs/api_integrations_def.go
```diff
@@ -6,14 +6,61 @@ import (
...
+
+var apiIntegrationOauth2GitAuthDef = g.NewQueryStruct("OAuth2GitUserAuthentication").
+	SQLWithCustomFieldName("authType", "TYPE = OAUTH2").
+	TextAssignment("OAUTH_AUTHORIZATION_ENDPOINT", g.ParameterOptions().SingleQuotes().Required()).
+	TextAssignment("OAUTH_TOKEN_ENDPOINT", g.ParameterOptions().SingleQuotes().Required()).
+	TextAssignment("OAUTH_CLIENT_ID", g.ParameterOptions().SingleQuotes().Required()).
+	TextAssignment("OAUTH_CLIENT_SECRET", g.ParameterOptions().SingleQuotes().Required()).
+	OptionalNumberAssignment("OAUTH_ACCESS_TOKEN_VALIDITY", g.ParameterOptions().NoQuotes()).
+	OptionalNumberAssignment("OAUTH_REFRESH_TOKEN_VALIDITY", g.ParameterOptions().NoQuotes()).
+	ListAssignment("OAUTH_ALLOWED_SCOPES", "ApiIntegrationScope", g.ParameterOptions().Parentheses()).
```
sfc-gh-jmichalak: This is unrestricted now, but it should be a list of enums.

pkg/sdk/generator/defs/api_integrations_def.go
```diff
@@ -46,14 +93,95 @@ var apiIntegrationsDef = g.NewInterface(
...
 			ListAssignment("API_ALLOWED_PREFIXES", "ApiIntegrationEndpointPrefix", g.ParameterOptions().Parentheses().Required()).
+			// TODO(next pr): API_BLOCKED_PREFIXES is not supported for Amazon API Gateway, github and maybe other types; in next PR test the compatibility of common params for each api_provider and sub-type
 			ListAssignment("API_BLOCKED_PREFIXES", "ApiIntegrationEndpointPrefix", g.ParameterOptions().Parentheses()).
 			BooleanAssignment("ENABLED", g.ParameterOptions().Required()).
 			OptionalComment().
 			WithValidation(g.ValidIdentifier, "name").
 			WithValidation(g.ConflictingFields, "IfNotExists", "OrReplace").
-			WithValidation(g.ExactlyOneValueSet, "AwsApiProviderParams", "AzureApiProviderParams", "GoogleApiProviderParams"),
-		ApiIntegrationEndpointPrefixDef,
+			WithValidation(g.ConflictingFields, "OrReplace", "ExternalMcpOAuth2ProviderParams", "ExternalMcpDynamicClientProviderParams").
```
sfc-gh-jmichalak, line 183: Why do we need this line? The MCP server types support OR REPLACE, and the mutual exclusivity is handled below anyway.

pkg/sdk/generator/defs/api_integrations_def.go
```diff
@@ -88,13 +216,66 @@ var apiIntegrationsDef = g.NewInterface(
 							TextAssignment("GOOGLE_AUDIENCE", g.ParameterOptions().SingleQuotes().Required()),
 						g.KeywordOptions(),
 					).
+					OptionalQueryStructField(
+						"GitHttpsApiTokenBasedParams",
```
sfc-gh-jmichalak, line 238: Q: Are all fields supported in SET? Some are missing from the docs.

pkg/sdk/generator/defs/api_integrations_def.go
sfc-gh-jmichalak: IMO, This should be handled similarly to SET.

sfc-gh-jmichalak (changes requested):

pkg/sdk/generator/defs/api_integrations_def.go
sfc-gh-jcieslak: Yup, I was checking and all fields here are possible, the only exception was one field that I only discovered during integration testing (will be visible in the next PR). Will confirm those options with the appropriate team, but for now going with what's possible in Snowflake

pkg/sdk/generator/defs/api_integrations_def.go
sfc-gh-jcieslak: They don't support it, will be confirmed with the appropriate team, but for now I'm following the Snowflake behavior

pkg/sdk/generator/defs/api_integrations_def.go
sfc-gh-jcieslak: It doesn't, confirmed by integration tests in the follow-up pr

pkg/sdk/generator/defs/api_integrations_def.go
sfc-gh-jcieslak: 1. Yes, it will, but for now, I have confirmed manually that it works (the SQL passes, and the appropriate values are returned by Snowflake.)
2. The tests confirmed it doesn't matter (and it would be impossible from docs standpoint because `API_BLOCKED_PREFIXES` is sometimes before and sometimes after api provider specific options)

pkg/sdk/generator/defs/api_integrations_def.go
sfc-gh-jcieslak: Done

pkg/sdk/generator/defs/api_integrations_def.go
sfc-gh-jcieslak: Done

pkg/sdk/generator/defs/api_integrations_def.go
sfc-gh-jcieslak: done

sfc-gh-jmichalak (approved):

PR #4771, opened 2026-05-22
chore: Generate convert for more SDK objects - part 3

sfc-gh-asawicki:
Continuation of https://github.com/snowflakedb/terraform-provider-snowflake/pull/4762.

Configures config generation and regenerates for multiple objects. Note that still some manual changes are necessary, and they will be handled as follow-ups.

Follow-ups:
- Continue with generating `convert` for other objects
- Address the manual changes present
- Handle errors properly for all converts


pkg/testacc/resource_cortex_search_service_acceptance_test.go
```diff
@@ -74,8 +74,7 @@ func TestAcc_CortexSearchService_basic(t *testing.T) {
 					resource.TestCheckResourceAttr(resourceName, "describe_output.0.target_lag", "2 minutes"),
 					resource.TestCheckResourceAttr(resourceName, "describe_output.0.warehouse", TestWarehouseName),
 					resource.TestCheckResourceAttr(resourceName, "describe_output.0.search_column", "SOME_TEXT"),
-					resource.TestCheckResourceAttr(resourceName, "describe_output.0.attribute_columns.#", "1"),
-					resource.TestCheckResourceAttr(resourceName, "describe_output.0.attribute_columns.0", ""),
+					resource.TestCheckResourceAttr(resourceName, "describe_output.0.attribute_columns.#", "0"),
```
sfc-gh-jcieslak, line 77: It's a slight change in parsing behavior that affects the output. Should we add a migration guide entry about it?

sfc-gh-jcieslak (approved):

pkg/testacc/resource_cortex_search_service_acceptance_test.go
sfc-gh-asawicki: IMO the previous one was incorrect as we were receiving the 1-element list with empty string element. I don't think any logic could work on that but we can add the migration guide entry more as an FYI. I will do that in a separate PR.

PR #4772, opened 2026-05-22
chore: Extract manual validation blocks from SDK validations gen files

sfc-gh-asawicki:
Part of the ongoing effort to reduce manual changes in SDK generation.

The `*_validations_gen.go` files are code-generated but historically accumulated manually-added validation blocks (marked with comments like `// added manually`, `// modified manually`, `// adjusted manually`, etc.). These blocks are invisible to the generator and are lost or corrupted on regeneration (as demonstrated by hybrid tables, where a slice-iteration loop was replaced with incorrect direct field access).                                                                              

This change extracts all manually-marked validation blocks from `*_validations_gen.go` files into `additionalValidations()` methods in the corresponding `*_ext.go files`, replacing the inline code with a single invocation:
```go
  errs = append(errs, opts.additionalValidations()) // invocation added manually                                                                                                                                                                                                                                                                                                                                                                                                                        
```
The next step is to add an option to the SDK generator to emit these invocation lines automatically. This will eliminate the need to manually re-add validation logic after regeneration and will allow missing validations to be added directly to `*_ext.go` files in a discoverable, maintainable way. Later, the validations will be iteratively added to SDK generator if they are repeatable.

Affected objects: tasks, services, compute pools, stages, views, notebooks, security integrations, notification integrations, semantic views, applications, streamlits, iceberg tables, api integrations, user programmatic access tokens, external volumes, file formats, functions, procedures, hybrid tables.

### DIFF
--- a/pkg/sdk/stages_ext.go
+++ b/pkg/sdk/stages_ext.go
@@ -7,6 +7,16 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
 )
 
+func (f *StageFileFormat) additionalValidations() error {
+	if f == nil {
+		return nil
+	}
+	if valueSet(f.FileFormatOptions) {
+		return f.FileFormatOptions.validate()
+	}
+	return nil
+}
+
 var AcceptableStageTypes = map[StageType][]StageType{
 	StageTypeInternal: {StageTypeInternal, StageTypeInternalNoCse},
 	StageTypeExternal: {StageTypeExternal},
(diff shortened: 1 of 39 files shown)


pkg/sdk/stages_ext.go
```diff
@@ -7,6 +7,16 @@ import (
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
 )
 
+func (f *StageFileFormat) additionalValidations() error {
+	if f == nil {
```
sfc-gh-jcieslak, line 11: nit: value set already handles this

sfc-gh-jcieslak (approved):

pkg/sdk/stages_ext.go
sfc-gh-asawicki: I will adjust in the next PR

PR #4780, opened 2026-05-26
chore: Generate convert for more SDK objects - part 4

sfc-gh-asawicki:
Continuation of: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4771

Configures config generation and regenerates for multiple objects. Note that still some manual changes are necessary, and they will be handled as follow-ups.
Notable changes:
- improved boolean conversions (string comparison, bool parsing)
- custom conversions

Follow-ups:
- Continue with generating convert for other objects
- Address the manual changes present
- Handle errors properly for all converts

### DIFF
--- a/pkg/sdk/compute_pools_impl_gen.go
+++ b/pkg/sdk/compute_pools_impl_gen.go
@@ -2,7 +2,6 @@
 
 package sdk
 
-// imports adjusted manually
 import (
 	"context"
 	"fmt"
@@ -145,8 +144,7 @@ func (r *ShowComputePoolRequest) toOpts() *ShowComputePoolOptions {
 }
 
 func (r computePoolsRow) convert() (*ComputePool, error) {
-	// Added manually
-	cp := &ComputePool{
+	result := &ComputePool{
 		Name:            r.Name,
 		MinNodes:        r.MinNodes,
 		MaxNodes:        r.MaxNodes,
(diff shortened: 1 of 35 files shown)


pkg/sdk/generator/defs/compute_pools_def.go
```diff
@@ -89,7 +89,7 @@ var computePoolsDef = g.NewInterface(
 		Enum("state", ComputePoolStateEnumDef).
 		Number("min_nodes").
 		Number("max_nodes").
-		PlainField("instance_family", "ComputePoolInstanceFamily").
+		PlainField("instance_family", "ComputePoolInstanceFamily", g.WithCustomParser("ToComputePoolInstanceFamily")).
```
sfc-gh-jmichalak, line 92: Nit: how about we omit the argument in WithCustomParser and just use "To" + PlainField's 2nd argument. Are there any cases where it doesn't work?

pkg/sdk/views_impl_gen.go
```diff
@@ -254,41 +254,25 @@ func (r *ShowViewRequest) toOpts() *ShowViewOptions {
...
-		view.Owner = r.Owner.String
-	}
-	if r.Comment.Valid {
-		view.Comment = r.Comment.String
-	}
+	mapNullStringToNonNullableField(&result.Kind, r.Kind)
+	mapNullStringToNonNullableField(&result.Reserved, r.Reserved)
+	mapNullStringToNonNullableField(&result.Owner, r.Owner)
+	mapNullStringToNonNullableField(&result.Comment, r.Comment)
+	// TODO [this PR]: allow metadata trimming
```
sfc-gh-jmichalak, line 267: I guess in the next PR?

pkg/sdk/compute_pools_impl_gen.go
```diff
@@ -163,30 +161,15 @@ func (r computePoolsRow) convert() (*ComputePool, error) {
...
-	if err != nil {
-		return nil, fmt.Errorf("error converting compute pool instance family: %w", err)
-	} else {
-		cp.InstanceFamily = instanceFamily
-	}
-	state, err := ToComputePoolState(r.State)
-	if err != nil {
-		return nil, fmt.Errorf("error converting compute pool state: %w", err)
+	mapStringWithMapping(&result.State, r.State, ToComputePoolState)
+	if v, err := ToComputePoolInstanceFamily(r.InstanceFamily); err == nil {
```
sfc-gh-jmichalak, line 165: Nit: it's more idiomatic to handle an error first.

sfc-gh-jmichalak (approved):

pkg/sdk/generator/defs/compute_pools_def.go
sfc-gh-asawicki: We need to know when we want to do the custom parsing, and currently, not all PlainFields require it. But, I think it's still early, as definitions still need improvements, e.g. this one should be later on represented by en enum, which has its own conversion.
I will leave it as an idea to be verified later.

pkg/sdk/views_impl_gen.go
sfc-gh-asawicki: heh, yes.

pkg/sdk/compute_pools_impl_gen.go
sfc-gh-asawicki: True. It was done based on other places where the ordering was like that. I will adjust later on.

PR #4784, opened 2026-05-28
fix: non empty plan after importing catalog integration iceberg rest resource

sfc-gh-jcieslak:
## Changes
- Adjust the catalog integration import function to include the force new fields (the read after import won't populate those fields in the state, as show and desc outputs are empty, and thus state setting for them are skipped).
- Add an acceptance test that validates the previously observed re-creation plan is no longer there, and we see the update plan instead (due to the secrets being missing from Snowflake outputs).
- Adjust migration guide and resource documentation to include info about import behavior and remaining limitations


pkg/testacc/resource_catalog_integration_iceberg_rest_acceptance_test.go
```diff
@@ -723,6 +707,229 @@ func TestAcc_CatalogIntegrationIcebergRest_BasicUseCaseOAuth(t *testing.T) {
...
+					testClient().CatalogIntegration.CreateFunc(t, createRequest)
+				},
+				Config:             config.FromModels(t, allAttributes),
+				ResourceName:       ref,
+				ImportState:        true,
+				ImportStateId:      id.FullyQualifiedName(),
+				ImportStatePersist: true,
+			},
+			// After import, all ForceNew fields (rest_config, sigv4_rest_authentication, etc.) must already
+			// match the config. sigv4_external_id is copied from configuration into state during import because
```
sfc-gh-jmichalak, line 917: 1. It's not copied in the import implementation.
2. It contradicts the step itself - sigv4_external_id causes the recreation.

pkg/resources/catalog_integration_iceberg_rest.go
```diff
@@ -461,3 +468,39 @@ func handleExternalChangesToSigV4RestAuthentication(d *schema.ResourceData, deta
...
+		"prefix":                 details.RestConfig.Prefix,
+		"catalog_name":           details.RestConfig.CatalogName,
+		"catalog_api_type":       string(details.RestConfig.CatalogApiType),
+		"access_delegation_mode": string(details.RestConfig.AccessDelegationMode),
+	}})
+}
+
+func setIcebergRestAuthInState(d *schema.ResourceData, details *sdk.CatalogIntegrationIcebergRestDetails) error {
+	if details.OAuthRestAuthentication != nil {
+		return errors.Join(
```
sfc-gh-jmichalak, line 484: Nit: here and below, errors.Join is unnecessary.

sfc-gh-jmichalak (approved):

PR #4785, opened 2026-05-28
chore: Adjust or skip failing tests

sfc-gh-jmichalak:
## Summary
- Skip two `TestInt_PostgresInstances` subtests (`create - with storage_integration` and `alter: set and unset storage_integration`) pending investigation in SNOW-3580377. The failures stem from an `invalid value [...] for parameter 'STORAGE_INTEGRATION (must be of type POSTGRES_EXTERNAL_STORAGE)'` SQL compilation error.
- Update `TestAcc_Stages_CompleteUseCase` assertions for `stages.0.describe_output.0.privatelink.0.use_privatelink_endpoint` from `TestCheckNoResourceAttr` to `TestCheckResourceAttr(..., "false")` to match the actual returned value. The change in Snowflake is expected, as now the s3-compat stages support privatelink endpoint.

## References
- SNOW-3580377


sfc-gh-asawicki (approved):

pkg/sdk/testint/postgres_instances_gen_integration_test.go
```diff
@@ -225,7 +225,17 @@ func TestInt_PostgresInstances(t *testing.T) {
 		)
 	})
 
+	// TODO(SNOW-3580377): Investigate and unskip:
```
sfc-gh-kwasilewski, line 228: Can we just paste a link to a failing GitHub Actions job with the logs instead?

sfc-gh-jmichalak: I prefer this way because:
1. The logs disappear after a few days (2 weeks, IIRC). We should resolve this sooner, but still.
2. The GitHub Actions sometimes load very slowly, so it's faster like this.

sfc-gh-kwasilewski: Fair enough, I thought they were kept longer than that.

PR #4786, opened 2026-05-28
feat: normalize cortex agent show/describe output representations

sfc-gh-kwasilewski:
## Changes
* Use non-nil strings for comment and CortexAgentProfile for profile.
* Normalize agent_spec in SDK describe conversion to stable JSON.
* Update schemas, docs, and assertions to use profile as a nested block.
* Adjust acceptance/integration tests to assert typed profile and spec.
* Bump terraform-plugin-testing to v1.16.0.


pkg/sdk/generator/defs/cortex_agents_def.go
```diff
@@ -58,17 +58,24 @@ var cortexAgentsDef = g.NewInterface(
 			Name().
 			WithValidation(g.ValidIdentifier, "name"),
 	).
-	ShowOperationWithPairedStructs(
+	ShowOperation(
 		"https://docs.snowflake.com/en/sql-reference/sql/show-agents",
-		g.StructPair("showCortexAgentDBRow", "CortexAgent").
+		g.DbStruct("showCortexAgentDBRow").
```
sfc-gh-jmichalak: This does not follow the latest enhancements in the sdk generator, why?

sfc-gh-jmichalak (dismissed):

pkg/sdk/generator/defs/cortex_agents_def.go
sfc-gh-kwasilewski: Do you mean WithConvertGeneration? It only works for paired structs

pkg/sdk/generator/defs/cortex_agents_def.go
```diff
@@ -79,10 +86,10 @@ var cortexAgentsDef = g.NewInterface(
 		g.ShowByIDLikeFiltering,
 		g.ShowByIDExtendedInFiltering,
 	).
-	DescribeOperationWithPairedStructs(
+	DescribeOperation(
 		g.DescriptionMappingKindSingleValue,
 		"https://docs.snowflake.com/en/sql-reference/sql/desc-agent",
-		g.StructPair("cortexAgentDetailsRow", "CortexAgentDetails").
```
sfc-gh-asawicki, line 85: You can express it through Struct Pair already (custom parser). We shouldn't move back from struct pair.

pkg/sdk/generator/defs/cortex_agents_def.go
sfc-gh-asawicki: You can express it through Struct Pair already (custom parser). We shouldn't move back from struct pair.

sfc-gh-asawicki (changes requested):

sfc-gh-asawicki (approved):

pkg/sdk/generator/defs/cortex_agents_def.go
sfc-gh-kwasilewski: Done

Issue #4793, opened 2026-05-29
snowflake_notebook - Is this a Legacy or Project Notebook?


WilMorland:
### Company Name

Serios Group

### Documentation Link

https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/notebook

### Description

It is unclear from the documentation is creating a notebook via the `snowflake_notebook` resource creates a 'Legacy' notebook (deprecated) or a project notebook.

Bonus! The 'Object type(s)' drop down on the issues form does not seem to contain this resource and is not searchable - might be worth looking to improve that!

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-kwasilewski:
Hi @WilMorland 👋 

The `snowflake_notebook` resource creates a legacy notebook. The provider doesn't currently support notebook projects. We will update the documentation to clarify this limitation and add the missing drop-down entries shortly. Thank you for bringing this to our attention.

sfc-gh-kwasilewski:
Hi @WilMorland, the updated documentation is now live and available [here](https://registry.terraform.io/providers/snowflakedb/snowflake/2.18.0/docs/resources/notebook). We have also added all the missing dropdown entries to the forms.

PR #4797, opened 2026-06-01
docs: clarify snowflake_notebook manages legacy notebooks

sfc-gh-kwasilewski:
Clarify in the resource/data source descriptions that snowflake_notebook manages Legacy Notebooks (CREATE NOTEBOOK), not Notebooks in Workspaces / Notebook Project Objects (CREATE NOTEBOOK PROJECT).

Also add the missing notebook and cortex_agent entries to the issue-form "Object type(s)" dropdowns and the repository labels list.

References: #4793


pkg/scripts/issues/labels.go
```diff
@@ -66,6 +67,7 @@ var RepositoryLabels = []string{
 	"resource:network_policy",
 	"resource:network_policy_attachment",
 	"resource:network_rule",
+	"resource:notebook",
```
sfc-gh-jmichalak, line 73: Nit: we should also add other new resources, such as stages, storage integrations, etc.

sfc-gh-jmichalak (dismissed):

pkg/scripts/issues/labels.go
sfc-gh-kwasilewski: Done

sfc-gh-jmichalak: These labels should be created for each resource, so the types are separated.

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

pkg/scripts/issues/labels.go
sfc-gh-kwasilewski: To be done in a follow-up.

Issue #3050, opened 2024-09-06
[Bug]: MODEL grants producing repeated plan after apply


halcarrigan-carvana:
### Terraform CLI Version

1.9.2

### Terraform Provider Version

0.95.0

### Terraform Configuration

```terraform
resource "snowflake_role" "role_test_models" {
  name    = "ROLE_TEST_MODELS"
  provider = snowflake.useradmin
}

resource "snowflake_grant_account_role" "role_grant" {
  role_name        = snowflake_role.role_test_models.name
  parent_role_name = "SYSADMIN"
  provider = snowflake.securityadmin
}

resource "snowflake_grant_privileges_to_account_role" "grant_schema" {
  account_role_name = snowflake_role.role_test_models.name
  privileges         = ["USAGE"]
  with_grant_option  = false

  on_schema_object {
    future {
      object_type_plural = "MODELS"
      in_database        = "DB_TEST_MODELS"
    }
  }
  provider = snowflake.securityadmin
}
```

### Category

category:grants

### Object type(s)

resource:grant_privileges_to_account_role

### Expected Behavior

The above configuration creates a role, grants the role to SYSADMIN, and grants USAGE on FUTURE MODELS to that role. After applying USAGE on MODELS to role, subsequent plans should recommend no changes.

### Actual Behavior

Plan does not converge. Each run suggests adding the usage privilege again.

```
Terraform will perform the following actions:

  # snowflake_grant_privileges_to_account_role.grant_schema will be updated in-place
  ~ resource "snowflake_grant_privileges_to_account_role" "grant_schema" {
        id                = "\"_ROLE_TEST_MODELS\"|false|false|USAGE|OnSchemaObject|OnFuture|MODELS|InDatabase|\"SHARED\""
      ~ privileges        = [
          + "USAGE",
        ]
        # (5 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

This may be due to snowflake reporting "grant_on" as 'MODULE' in `SHOW FUTURE GRANTS IN DATABASE`.

### Steps to Reproduce

1. Copy configuration
2. Run *Terraform apply*
3. Confirm in snowflake future grant was applied. `SHOW FUTURE GRANTS TO ROLE ROLE_TEST_MODELS;`
4. Repeat *Terraform apply*

### How much impact is this issue causing?

Low

### Would you like to implement a fix?

[...]

PR #4806, opened 2026-06-02
fix: Adjust model monitor object type in grants

sfc-gh-dszmolka:
## Changes

- Map `QUALITY_MONITOR` to the `MODEL MONITOR` object type in `grantRow.convert()` for both `granted_on` (current grants) and `grant_on` (future grants)
- Follow-up for #4529 (AGENT) and #4532 (MCP SERVER) — same class of Snowflake object type name mismatch

## Problem

Snowflake returns `QUALITY_MONITOR` in the `grant_on` column of `SHOW FUTURE GRANTS` output for grants on future `MODEL MONITORS`. The provider writes the grant using `MODEL MONITORS` but reads back `QUALITY_MONITOR` during state reconciliation — a name it does not recognize as equivalent — causing it to treat the grant as missing and re-apply on every plan.
Example perpetual drift:

```hcl
resource "snowflake_grant_privileges_to_database_role" "model_monitors_future_usage" {
  privileges         = ["USAGE"]
  database_role_name = snowflake_database_role.schema_role.fully_qualified_name
  on_schema_object {
    future {
      object_type_plural = "MODEL MONITORS"
      in_schema          = local.schema_fully_qualified_name
    }
  }
}
~ resource "snowflake_grant_privileges_to_database_role" "model_monitors_future_usage" {
  ~ privileges = [
      + "USAGE",
    ]
  }
```

## Root cause

In pkg/sdk/grants.go, grantRow.convert() applies strings.ReplaceAll(row.GrantOn, "_", " ") which turns QUALITY_MONITOR into QUALITY MONITOR. The provider then compares this against the expected MODEL MONITOR (derived from the HCL object_type_plural = "MODEL MONITORS" via PluralObjectType.Singular()). The comparison fails, so the grant is treated as absent.

Special-case mappings already existed for VOLUME → EXTERNAL VOLUME, MODULE → MODEL, CORTEX_AGENT → AGENT, and CORTEX_AGENT_SERVER → MCP SERVER. The mapping for QUALITY_MONITOR → MODEL MONITOR was missing.

## Fix

Added QUALITY_MONITOR → ObjectTypeModelMonitor mapping in grantRow.convert(), following the exact same pattern as the four existing mappings.

## Test plan

- Added TestAcc_GrantPrivilegesToAccountRole_OnSchemaObject_OnFuture_ModelMonitors_InDatabase_v2_14_1_NonEmptyPlan
    - Step 1: v2.14.1 external provider → ExpectNonEmptyPlan: true (bug present)
    - Step 2: current provider → ExpectEmptyPlan() (bug fixed)
- Added TestAcc_GrantPrivilegesToDatabaseRole_OnFutureModelMonitors_InDatabase_v2_14_1_NonEmptyPlan
Same pattern for database role grants (matches the reported support case)
- go build ./... passes
- Unit test against real grantRow.convert() confirms QUALITY_MONITOR → MODEL MONITOR mapping works for both grant_on and granted_on

## References

- customer support case (SNOW-3595226)
- Related: #4524 (MCP SERVER same class of bug), #3050 (MODEL same class of bug)
- Prior fixes: #4529 (AGENT), #4532 (MCP SERVER)


pkg/acceptance/helpers/test_client.go
```diff
@@ -153,6 +154,7 @@ func NewTestClient(
 		MaskingPolicy:                NewMaskingPolicyClient(context, idsGenerator),
 		MaterializedView:             NewMaterializedViewClient(context, idsGenerator),
 		McpServer:                    NewMcpServerClient(context, idsGenerator),
+		ModelMonitor:                 NewModelMonitorClient(context, idsGenerator),
```
sfc-gh-jmichalak: This does not compile, and it's not needed. In this case, we are only testing ALL/FUTURE grants, so there's no need to use a model monitor client. Same above.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -3622,6 +3622,37 @@ func TestAcc_GrantPrivilegesToAccountRole_OnSchemaObject_OnFuture_McpServers_InD
...
+	databaseName := testClient().Ids.DatabaseId().Name()
+
+	resource.Test(t, resource.TestCase{
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		CheckDestroy: CheckAccountRolePrivilegesRevoked(t),
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders:  ExternalProviderWithExactVersion("2.14.1"),
```
sfc-gh-jmichalak: Nit: this bug has been reported in v2.14.1, but it's also present in 2.17.0. We usually use the last released version with the bug. This is optional, but if you update, please also update the function name (same in the other file).

pkg/testacc/resource_grant_privileges_to_database_role_acceptance_test.go
```diff
@@ -1402,6 +1402,33 @@ func TestAcc_GrantPrivilegesToDatabaseRole_OnFutureModels_issue3050(t *testing.T
...
+		CheckDestroy: CheckAccountRolePrivilegesRevoked(t),
+		Steps: []resource.TestStep{
+			{
+				ExternalProviders:  ExternalProviderWithExactVersion("2.14.1"),
+				Config:             grantPrivilegesToDatabaseRoleOnFutureInDatabaseConfig(databaseRoleId, []string{"USAGE"}, sdk.PluralObjectTypeModelMonitors, databaseRoleId.DatabaseName()),
+				ExpectNonEmptyPlan: true,
+			},
+			{
+				ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+				Config:                   grantPrivilegesToDatabaseRoleOnFutureInDatabaseConfig(databaseRoleId, []string{"USAGE"}, sdk.PluralObjectTypeModelMonitors, databaseRoleId.DatabaseName()),
```
sfc-gh-jmichalak, line 1421: Nit: the tests look good, but it would be nice to also test other grant type - see OnObject test in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4529/changes.

sfc-gh-jmichalak (changes requested):

pkg/testacc/resource_grant_privileges_to_database_role_acceptance_test.go
sfc-gh-dszmolka: we discussed this would require leaving ModelMonitor client after all which was asked to be removed in commnt#1 , and that we should continue without it for now

pkg/acceptance/helpers/test_client.go
sfc-gh-dszmolka: this has been removed

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
sfc-gh-dszmolka: transformed them into reference to 2_17_0

sfc-gh-jmichalak (approved):

PR #4811, opened 2026-06-03
chore: Reconfigure GH Actions CI for faster PR feedback

sfc-gh-kwasilewski:
Run only fast unit tests on push/PR via a new unit-tests workflow, and move the full and account-level suites off pull_request to a weekday daily schedule plus on-demand /ok-to-test dispatch.


.github/workflows/tests-account-level.yml
```diff
@@ -20,39 +20,25 @@ concurrency:
...
-  pull_request:
+  schedule:
+    - cron: '0 6 * * 1-5'
 
 jobs:
   account-level-tests:
     environment: test
     env:
       ## adding additional suffix to don't have conflict in non-account level tests and their pre-created objects
-      TEST_SF_TF_TEST_OBJECT_SUFFIX: ${{ format('{0}{1}', ((github.event_name == 'repository_dispatch' && github.event.client_payload.pull_request.head.sha) || (github.event_name == 'pull_request' && github.event.pull_request.head.sha)), 'AL') }}
```
sfc-gh-asawicki, line 30: We still should require them on the dev -> main PR.

Because of that, maybe it would be good to keep the new daily runs separated and have the old ones configured only for the dev->main?

sfc-gh-asawicki (changes requested):

.github/workflows/tests-account-level.yml
sfc-gh-kwasilewski: You're right.

Let's do that. This file would become too messy if it had the new schedule trigger in addition to the existing ones.

sfc-gh-asawicki (dismissed):

sfc-gh-jmichalak (dismissed):

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

PR #4689, opened 2026-04-28
feat: Add snowflake_hybrid_table resource (preview)

sfc-gh-moczko:
## Summary
- Implements `snowflake_hybrid_table` Terraform resource (PR 3 of 4: SDK → integration tests → **resource** → data source)
- Full CRUD with column management, out-of-line constraints (PK/UNIQUE/FK), property SET/UNSET, drift detection, and import
- Gated behind `snowflake_hybrid_table_resource` preview feature

## Breaking change (preview resource)

`snowflake_hybrid_table` is a **preview** resource. Recent commits in this PR remove the `-1` sentinel behavior for `data_retention_time_in_days` and `max_data_extension_time_in_days` (now handled via the standard Snowflake parameter plumbing — `ParametersCustomDiff`):

- Before: setting `= -1` meant "use the Snowflake default".
- After: omit the field to inherit the default; any non-negative integer is accepted, and `-1` is rejected.

Users with existing configs that set either field to `-1` must remove the field from their config to keep the previous (default) behavior.

## Known limitations (preview → GA gap)

- `name` rename works via `ALTER TABLE … RENAME TO` (no `ForceNew`).
- `database` and `schema` have `ForceNew` — moving a table between databases/schemas is not yet supported by this provider. Tracked as future work.
- AUTOINCREMENT/IDENTITY column defaults, inline constraints per-column, and out-of-line indexes (`CREATE INDEX`) are not yet implemented. Planned for GA follow-up PRs.
- Constraint names (PK/UNIQUE/FK) are not read back from Snowflake during Read — if Snowflake auto-generates a name, it will not appear in state. Planned for GA via `INFORMATION_SCHEMA.TABLE_CONSTRAINTS` (covers all constraint kinds, not only PRIMARY KEY).
- `collate` on non-text columns is passed through to Snowflake (which will return a SQL error). Fail-fast validation via `CustomizeDiff` is deferred to a follow-up PR.
- `ImportStateVerifyIgnore` still skips `column` and `primary_key` because DESCRIBE-derived state can diverge from raw config (type-attribute normalization, synthesized constraint names). Tracked for a follow-up PR — diagnosing the precise divergences requires running `TF_ACC=1 TF_LOG=DEBUG` against a live Snowflake account, which is out of scope here.

## Follow-ups for next PR

Items deferred from this review round, tracked for a follow-up PR:

[...]

PR #4834, opened 2026-06-11
docs(hybrid table): explain d.Partial(true) usage in Update

sfc-gh-moczko:
## Summary

- Adds a comment before the first \`d.Partial(true)\` call in \`UpdateHybridTable\` explaining the SDKv2 partial-state contract and its interaction with \`Computed: true\` fields
- No logic changes

## Context — item 9 from the post-#4689 follow-up plan

**\`d.Partial(true)\`** is called ~10 times in \`UpdateHybridTable\`. In SDKv2, when \`Update\` returns an error the framework writes the *proposed config* into state — even without any \`d.Set\` calls — replacing known-good prior state with unverified values. \`d.Partial(true)\` prevents this by preserving the previous state.

**\`Computed: true\` interaction**: \`show_output\` and \`describe_output\` are only refreshed when \`Update\` succeeds (via the trailing \`ReadHybridTable\` call). On any early-return error path, those fields are safely preserved because (a) they are \`Computed: true\` so Terraform never sources them from config, and (b) \`d.Partial(true)\` preserves all prior state. The comment connects these two concepts in one place.

The comment cites \`terraform-plugin-sdk/v2 (helper/schema/resource_data.go: Partial)\`, consistent with the existing citation style at lines 527 and 1091 of the file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/resources/hybrid_table.go
+++ b/pkg/resources/hybrid_table.go
@@ -743,6 +743,13 @@ func UpdateHybridTable(ctx context.Context, d *schema.ResourceData, meta any) di
 		)
 
 		if err := client.HybridTables.Alter(ctx, sdk.NewAlterHybridTableRequest(id).WithNewName(newId)); err != nil {
+			// d.Partial(true) preserves the resource's previous state on error. Without it, SDKv2 writes
+			// the proposed config into state when Update returns an error — even without any d.Set calls —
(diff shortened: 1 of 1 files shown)


sfc-gh-moczko:
Consolidated into #4840.

PR #4840, opened 2026-06-11
feat(hybrid table): constraint name read-back, inline index block, and generator follow-ups

sfc-gh-moczko:
## Summary

Follow-up improvements to the hybrid table resource addressing reviewer feedback from #4689.

**Constraint name read-back**
- Added `GetConstraints` to the SDK using `SHOW PRIMARY KEYS` / `SHOW UNIQUE KEYS` / `SHOW IMPORTED KEYS` (metadata-only, no warehouse required)
- `primary_key.name`, `unique_constraint.name`, and `foreign_key.name` are now `Computed` — Snowflake auto-assigns names, so we read them back rather than requiring user input
- Name-independent set hashes (`uniqueConstraintHash`, `foreignKeyHash`) prevent spurious ForceNew when the server-assigned name differs from config

**Inline index block**
- New create-only `index` TypeSet block: `name` (required), `columns` (TypeList, ordered), `include_columns` (TypeSet, unordered)
- Inline `INDEX … INCLUDE …` clause emitted in `CREATE HYBRID TABLE` DDL
- `ShowIndexes` read-back filtered to user secondary indexes only: FK-backing indexes excluded by FK name cross-reference and `SYS_INDEX_` prefix; SHOW INDEXES failure is skip-with-warning

**SDK generator migration**
- `ShowPrimaryKeys`, `ShowUniqueKeys`, `ShowImportedKeys` added to the generator definition; hand-rolled SQL helpers deleted

**Other**
- Generator-definition corrections, schema adjustments, test consolidations, and assert-helper restructuring
- `ImportStateVerifyIgnore` trimmed to `"column"` only

## Notes

- FK constraints create a system-managed backing index visible in `SHOW INDEXES` (`is_unique = N`). These are filtered out in two ways: named FK-backing indexes share the FK constraint name (excluded via `GetConstraints` cross-reference); anonymous ones match the `SYS_INDEX_..._FOREIGN_KEY_...` prefix
- Standalone `snowflake_hybrid_table_index` resource is out of scope: `CreateIndex`/`DropIndex` generate fully-qualified identifiers but Snowflake requires bare names — blocked on an SDK fix

## Test Plan

- [x] `go build ./...` — clean
- [x] `go vet ./pkg/resources/... ./pkg/sdk/...` — clean
- [x] `go test ./pkg/resources/ -run 'TestUnit_HybridTable|Test_build|Test_parse'` — passes offline
- [x] `TF_ACC=1 go test -tags non_account_level_tests ./pkg/sdk/testint/ -run TestInt_HybridTableShowIndexes -v -timeout 20m` — pass
- [x] `TF_ACC=1 go test -tags non_account_level_tests ./pkg/sdk/testint/ -run TestInt_HybridTableConstraints -v -timeout 20m` — pass
- [x] `TF_ACC=1 go test -tags non_account_level_tests ./pkg/testacc/ -run TestAcc_HybridTable_ConstraintNameReadback -v -timeout 30m` — pass
- [x] `TF_ACC=1 go test -tags non_account_level_tests ./pkg/testacc/ -run TestAcc_HybridTable_Index -v -timeout 30m` — pass
- [x] `TF_ACC=1 go test -tags non_account_level_tests ./pkg/testacc/ -run TestAcc_HybridTable -v -timeout 90m` — all 17 tests pass


pkg/sdk/testint/hybrid_table_constraints_integration_test.go
```diff
@@ -0,0 +1,104 @@
...
+		},
+	)
+	t.Cleanup(parentCleanup)
+
+	// Create the child table via raw SQL so we can control exact constraint names,
+	// including an anonymous UNIQUE which Snowflake will assign a SYS_CONSTRAINT_* name.
+	childId := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+	t.Cleanup(testClientHelper().HybridTable.DropFunc(t, childId))
+
+	createChildSQL := fmt.Sprintf(
```
sfc-gh-jcieslak: We should be able to do with SDK, let's not use raw SQL

pkg/sdk/testint/hybrid_table_indexes_integration_test.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// TestInt_HybridTableShowIndexes proves the existing ShowIndexes method returns the
+// shapes the resource Read path relies on (settled in the design's step-1 e2e capture):
+//   - user secondary indexes report is_unique = "N" (IsUnique points to false);
+//   - the PK- and UNIQUE-backed indexes report "Y" (IsUnique points to true);
+//   - columns / included_columns are bracketed, uppercase, comma-space lists.
+//
+// No SDK production code is exercised beyond the already-generated ShowIndexes.
+func TestInt_HybridTableShowIndexes(t *testing.T) {
```
sfc-gh-jcieslak: Those tests could be placed in already existing file containing integration tests for hybrid tables (having multiple files per interface is confusing or just cumbersome if someone is looking for a particular test)

pkg/sdk/hybrid_table_constraints.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"sort"
+)
+
+// HybridTableConstraint is the merged, ordered view of one table constraint.
+type HybridTableConstraint struct {
```
sfc-gh-jcieslak: I think this should be possible and should be as part of _def.go file. We should avoid at any cost the manually written SQL abstractions.

pkg/sdk/hybrid_table_constraints.go
```diff
@@ -0,0 +1,147 @@
...
+	KeySequence    int    `db:"key_sequence"`
+	PkDatabaseName string `db:"pk_database_name"`
+	PkSchemaName   string `db:"pk_schema_name"`
+	PkTableName    string `db:"pk_table_name"`
+	PkColumnName   string `db:"pk_column_name"`
+	DeleteRule     string `db:"delete_rule"`
+	UpdateRule     string `db:"update_rule"`
+}
+
+func showPrimaryKeysSQL(id SchemaObjectIdentifier) string {
```
sfc-gh-jcieslak: Those commands are for all tables, which doesn't make sense to name those files and other things are hybrid table specific. Those commands should be extracted and used in hybrid tables generator and be properly abstracted so that they could be reused by other table types.

pkg/sdk/hybrid_table_constraints_test.go
```diff
@@ -0,0 +1,108 @@
+package sdk
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHybridTableConstraints_ShowSQL(t *testing.T) {
```
sfc-gh-jcieslak: This file should be removed, as SDK unit tests look differently. This moves us back to the legacy pattern of producing SQL with plain strings, which should be adjusted as mentioned in other comments.

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/hybrid_table_desc_output_ext.go
```diff
@@ -0,0 +1,104 @@
...
+// physical column order returned by DESCRIBE TABLE).
+type HybridTableDescribeOutputAssert struct {
+	*assert.ResourceAssert
+	rowIndex int
+}
+
+func HybridTableDescribeOutput(t *testing.T, name string, rowIndex int) *HybridTableDescribeOutputAssert {
+	t.Helper()
+	return &HybridTableDescribeOutputAssert{
+		ResourceAssert: assert.NewResourceAssert(name, "describe_output"),
```
sfc-gh-jcieslak: instead of having field method called per assertion, I think you could just adjust the prefix here in resource assert constructor, and let's adjust the struct name to emphasize it's for assertions on single row rather than whole describe output. It's also a broader question, if we would like to have such specialized assertions or rather we should have describe assert struct that contans methods allowing for nested assertions (without struct scoped rowIndex that afffects assertions inside; in other cases we usually pass as parameters what given assertion needs, which means to stay consistent maybe we should pass rowIndex everywhere it's required instead).

pkg/acceptance/bettertestspoc/config/model/hybrid_table_model_ext.go
```diff
@@ -134,6 +134,43 @@ func (h *HybridTableModel) WithUniqueConstraint(columns []string) *HybridTableMo
...
+
+// WithIndex sets the index block from one or more index definitions.
+//
+// Uses ObjectVariable rather than MapVariable because each index block mixes a
+// string (name) with list values (columns, include_columns), which MapVariable's
+// MarshalJSON rejects ("maps must contain the same type").
+func (h *HybridTableModel) WithIndex(indexes ...HybridTableIndexConfig) *HybridTableModel {
+	objs := make([]tfconfig.Variable, len(indexes))
+	for i, idx := range indexes {
+		colVars := make([]tfconfig.Variable, len(idx.Columns))
```
sfc-gh-jcieslak: nit: `collections.Map` could be used in trasformations

pkg/resources/hybrid_table.go
```diff
@@ -124,8 +130,9 @@ var hybridTableSchema = map[string]*schema.Schema{
 				"name": {
 					Type:        schema.TypeString,
 					Optional:    true,
+					Computed:    true,
```
sfc-gh-jcieslak, line 133: We usually avoid specifying Optional + Computed field, is there a reason why this was transformed to Computed ?

pkg/resources/hybrid_table.go
```diff
@@ -142,14 +149,16 @@ var hybridTableSchema = map[string]*schema.Schema{
 		Type:        schema.TypeSet,
 		Optional:    true,
 		ForceNew:    true,
+		Set:         uniqueConstraintHash,
```
sfc-gh-jcieslak, line 152: Could you explain why do we need custom hashing algorithm for both unique_constraint, include_columns, foreign_key, and index? It was needed in any of the previous resources we've worked on I think.

pkg/resources/hybrid_table.go
```diff
@@ -710,13 +862,57 @@ func GetReadHybridTableFunc(withExternalChangesMarking bool) schema.ReadContextF
...
+		}
+
+		// Index read-back is best-effort: a SHOW INDEXES failure must not fail Read of
+		// an otherwise-healthy table. This mirrors the non-fatal SHOW IMPORTED KEYS sub-call
+		// inside GetConstraints, not the fatal GetConstraints call above.
+		var indexState []map[string]any
+		indexes, indexErr := client.HybridTables.ShowIndexes(ctx,
+			sdk.NewShowIndexesHybridTableRequest().WithIn(sdk.TableIn{Table: id}))
+		if indexErr != nil {
+			log.Printf("[WARN] SHOW INDEXES failed for %s; skipping index read-back: %v", id.FullyQualifiedName(), indexErr)
```
sfc-gh-jcieslak: Why do we log instead of returning an error?

pkg/resources/hybrid_table.go
```diff
@@ -710,13 +862,57 @@ func GetReadHybridTableFunc(withExternalChangesMarking bool) schema.ReadContextF
...
+		} else {
+			// FK constraints create a system-managed backing index in Snowflake that
```
sfc-gh-jcieslak: We don't need walls of comments to understand the code, comment only when something really requires some level of explaination, and when you comment try to make it crisp (very brief, but still enough for everyone to get what it's all about).

pkg/resources/hybrid_table.go
```diff
@@ -710,13 +862,57 @@ func GetReadHybridTableFunc(withExternalChangesMarking bool) schema.ReadContextF
...
+			fkNames := make(map[string]bool, len(constraints))
+			for _, c := range constraints {
+				if c.Kind == sdk.ColumnConstraintTypeForeignKey {
+					fkNames[strings.ToUpper(c.Name)] = true
+				}
+			}
+			var userIndexes []sdk.HybridTableIndex
+			for _, idx := range indexes {
+				upper := strings.ToUpper(idx.Name)
+				if fkNames[upper] || strings.HasPrefix(upper, "SYS_INDEX_") {
```
sfc-gh-jcieslak: Why do we need fkNames? we can just assert here the kind instead of building a map

pkg/resources/hybrid_table.go
```diff
@@ -710,13 +862,57 @@ func GetReadHybridTableFunc(withExternalChangesMarking bool) schema.ReadContextF
...
+					continue
+				}
+				userIndexes = append(userIndexes, idx)
+			}
+			indexState = buildIndexesStateFromShowIndexes(userIndexes)
```
sfc-gh-jcieslak: Can't buildIndexesStateFromShowIndexes be invoded below in `d.Set` like other constraints? It would be clearer

pkg/resources/hybrid_table.go
```diff
@@ -979,37 +1175,83 @@ func buildHybridColumnStateFromDescribe(details []sdk.HybridTableDetails, d *sch
...
-		{
-			"name": pkName,
-			"keys": pkKeys,
-		},
+// parseIndexColumnsString parses the bracketed, uppercase column list that
+// SHOW INDEXES returns for the `columns` and `included_columns` fields, e.g.
+// "[A, B, C]" -> {"A","B","C"} and the literal "[]" -> empty slice. The SDK does
+// not pre-split these (see hybridTableIndexRow.convert), so the resource layer
+// parses them here.
+func parseIndexColumnsString(raw string) []string {
```
sfc-gh-jcieslak: nit: could be inlined

pkg/sdk/testint/hybrid_table_constraints_integration_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/collections"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/require"
+)
+
+func TestInt_HybridTableConstraints(t *testing.T) {
```
sfc-gh-jcieslak: Same as with unit tests, this could be a part of already existing file where other hybrid table tests reside. We don't need to separate hybrid table tests into multiple files

pkg/testacc/resource_hybrid_table_acceptance_test.go
```diff
@@ -1359,6 +1355,254 @@ func TestAcc_HybridTable_PKNullableNoSpurious(t *testing.T) {
...
+				ImportStateVerifyIgnore: []string{"column"},
+			},
+		},
+	})
+}
+
+// hybridTableIDColumnDescribeOutputAssert returns a describe_output assertion for the
+// ID column (row 0): PK, NOT NULL, no default or comment. Shared across BasicUseCase,
+// CompleteUseCase, and any future test that has the same ID-column configuration.
+func hybridTableIDColumnDescribeOutputAssert(t *testing.T, ref string) assert.TestCheckFuncProvider {
```
sfc-gh-jcieslak: We usually don't extract such assertions, let's inline it

pkg/testacc/resource_hybrid_table_acceptance_test.go
```diff
@@ -144,16 +144,47 @@ func TestAcc_HybridTable_BasicUseCase(t *testing.T) {
...
 		// DESCRIBE normalizes types (e.g. INTEGER -> NUMBER(38,0)); DiffSuppressDataTypes
 		// handles this at plan time, but the raw state values differ after import.
 		"column",
-		// Constraint name may differ between config (empty) and what DESCRIBE returns.
-		"primary_key",
 	}
 
+	// Column-mutation models for the add/drop lifecycle steps absorbed from the
+	// standalone TestAcc_HybridTable_ColumnAdd / ColumnDrop tests.
+	colsWith2 := []sdk.TableColumnSignature{
```
sfc-gh-jcieslak: Could we move those new steps to another test that would be specifically checking column behavior?

pkg/testacc/resource_hybrid_table_acceptance_test.go
```diff
@@ -277,8 +388,19 @@ func TestAcc_HybridTable_CompleteUseCase(t *testing.T) {
...
 		// handles this at plan time, but the raw state values differ after import.
 		"column",
-		// Constraint name may differ between config (empty) and what DESCRIBE returns.
-		"primary_key",
+	}
+
+	// The EMAIL (row 2) describe_output row is unchanged across the create and update
+	// steps — only the NAME row's comment differs — so the helper is shared below.
+	// ID (row 0) uses the package-level hybridTableIDColumnDescribeOutputAssert.
+	assertEmailColumnDescribeOutput := func(ref string) assert.TestCheckFuncProvider {
```
sfc-gh-jcieslak: Let's not mix extracted and non-extracted assertions, let's inline this

pkg/testacc/resource_hybrid_table_acceptance_test.go
```diff
@@ -1359,6 +1355,254 @@ func TestAcc_HybridTable_PKNullableNoSpurious(t *testing.T) {
...
+	}
+	pk := []sdk.TableColumnSignature{{Name: "ID"}}
+
+	// Resource label "child" (rather than the file's usual "test") distinguishes the
+	// Terraform-managed child from the externally-created parent referenced by the FK.
+	//
+	// Two unique constraints: one with an explicit name, one without.
+	// Use WithUniqueConstraintValue to express both in a single SetVariable.
+	childModel := model.HybridTableFromId("child", id, cols, pk).
+		WithUniqueConstraintValue(
```
sfc-gh-jcieslak: This could be improved by extending model builder in _ext file to avoid using `tfconfig` package within test files

pkg/testacc/resource_hybrid_table_acceptance_test.go
```diff
@@ -1359,6 +1355,254 @@ func TestAcc_HybridTable_PKNullableNoSpurious(t *testing.T) {
...
+		CheckDestroy: CheckDestroy(t, resources.HybridTable),
+		Steps: []resource.TestStep{
+			// Step 1: Apply and assert constraint names are correctly read back.
+			// PK name is auto-generated (non-empty); explicit names round-trip exactly;
+			// the unnamed UNIQUE gets a server-generated SYS_CONSTRAINT_-prefixed name.
+			{
+				Config: accconfig.FromModels(t, childModel),
+				Check: assertThat(t,
+					// primary_key is TypeList — index 0 is reliable. Auto-generated name is non-empty.
+					assert.Check(resource.TestCheckResourceAttrSet(childModel.ResourceReference(), "primary_key.0.name")),
```
sfc-gh-jcieslak: Use generated assertions, and remove unnecessary comments

pkg/testacc/resource_hybrid_table_acceptance_test.go
```diff
@@ -1359,6 +1355,254 @@ func TestAcc_HybridTable_PKNullableNoSpurious(t *testing.T) {
...
+// Three steps:
+//  1. Apply: assert the PK name is non-empty (auto-generated), the named UNIQUE
+//     round-trips to "my_uq" while the unnamed UNIQUE gets a non-empty
+//     SYS_CONSTRAINT_-prefixed name, and the named FK round-trips to "my_fk"
+//     against the expected referenced table.
+//  2. Same config, expect empty plan — the key regression guard for the Computed +
+//     Set-hash work: a second apply must not see drift in constraint names.
+//  3. Import with ImportStateVerifyIgnore: []string{"column"} — constraint names
+//     must survive import without being suppressed.
+func TestAcc_HybridTable_ConstraintNameReadback(t *testing.T) {
```
sfc-gh-jcieslak: This test could be merged with CompleteUseCase one.  The step 2 is useless as changes are implicitly checked in the 1st step by testing framework, which makes is 2 step test that is the same as CompelteUseCase that **must** contain all optional fields.

pkg/testacc/resource_hybrid_table_acceptance_test.go
```diff
@@ -1359,6 +1355,254 @@ func TestAcc_HybridTable_PKNullableNoSpurious(t *testing.T) {
...
+		HasUniqueKey(false).
+		HasDefault("").
+		HasComment("")
+}
+
+// checkUniqueConstraintNamesReadBack asserts that every unique_constraint element in
+// state has a name that is either the explicit "my_uq" or a server-generated name with
+// the "SYS_CONSTRAINT_" prefix. An empty name (which would mean GetConstraints failed to
+// read back the auto-generated name of the unnamed UNIQUE) fails the check.
+func checkUniqueConstraintNamesReadBack(resourceRef string) resource.TestCheckFunc {
```
sfc-gh-jcieslak, line 1680: I don't get why this is needed, can't similar thing be achieved by generated (or extended) assertions? If not, do we really need this assert?

pkg/testacc/resource_hybrid_table_acceptance_test.go
```diff
@@ -1359,6 +1355,254 @@ func TestAcc_HybridTable_PKNullableNoSpurious(t *testing.T) {
...
+// TestAcc_HybridTable_Index verifies the create-only secondary-index block:
+//  1. Apply two indexes (one with INCLUDE); assert both round-trip into state
+//     via the ShowIndexes read-back (the index block stays out of
+//     ImportStateVerifyIgnore, so read-back must converge with config).
+//  2. Same config -> empty plan (regression guard for the indexHash + case
+//     suppression: a lowercase config must not churn against the uppercase
+//     SHOW INDEXES read-back).
+//  3. Import with ImportStateVerifyIgnore = {"column"} only; index names and
+//     columns must survive without being suppressed.
+func TestAcc_HybridTable_Index(t *testing.T) {
```
sfc-gh-jcieslak: This also seems like another test that should be merged with CompleteUseCase one that should contain all optional fields / behavior in one complete model

sfc-gh-jcieslak (changes requested):

pkg/resources/hybrid_table.go
sfc-gh-moczko: The `name` field on `primary_key` is `Optional + Computed + ForceNew` because Snowflake auto-generates a constraint name when the user omits it, and the Read function writes that server-generated name back into state via `buildPrimaryKeyStateFromConstraints`. Without `Computed: true`, Terraform would error at refresh time when the provider sets a value for a field the user left blank. The field description documents this: "If omitted, Snowflake auto-generates one (visible in state after the first apply)."

pkg/resources/hybrid_table.go
sfc-gh-moczko: Yeah, I was also thinking about it. Each hash function addresses a specific state-drift problem:

- `uniqueConstraintHash` and `foreignKeyHash` ignore name when building the hash because name is optional and can be filled in by the server. If we included it, something with no name in the config would end up with a different hash after Read writes back the generated name. That would make Terraform think it changed and trigger unnecessary churn or ForceNew on every later plan.
- `foreignKeyHash` additionally normalizes `table_id` to a canonical fully-qualified name to absorb quoting differences.
- `indexHash` and `indexIncludeColumnsHash` uppercase column names because `SHOW INDEXES` returns uppercase column names while a config may use any case — without this, a lowercase column in config and its uppercase server representation hash to different buckets, triggering a spurious ForceNew.

pkg/resources/hybrid_table.go
sfc-gh-moczko: Intentional best-effort behavior — a SHOW INDEXES failure must not fail the Read of an otherwise-healthy table. This mirrors the non-fatal SHOW IMPORTED KEYS sub-call inside `GetConstraints` (which also logs WARN and returns the partial result on failure). The comment block at lines 870–872 documents this contract explicitly.

Would you rather prefer to return an error?

pkg/testacc/resource_hybrid_table_acceptance_test.go
sfc-gh-moczko: It struck me, too, initially. 

The function checks a predicate that standard assertions cannot express: "all `unique_constraint.*.name` attributes are either the literal `my_uq` or match the `SYS_CONSTRAINT_*` prefix."

`TestCheckTypeSetElemNestedAttrs` can verify that *some* element in the set has a given value, but cannot assert that *every* element satisfies a pattern. Generated `resourceassert` methods offer typed empty/not-empty/exact-value checks per field — none support per-element prefix matching across a full TypeSet.

The specific scenario tested — one named constraint + one unnamed (system-generated) constraint in a 2-element set — has a non-deterministic hash suffix on the generated name, so we can only validate the prefix, not the full value.

pkg/resources/hybrid_table.go
sfc-gh-moczko: Done — `parseIndexColumnsString` deleted; both call sites now call `sdk.ParseCommaSeparatedStringArray` directly.

pkg/resources/hybrid_table.go
sfc-gh-moczko: Done — removed the `fkNames` map. The FK-backing filter now lives inside `readIndexState` as a nested loop that checks `c.Kind == sdk.ColumnConstraintTypeForeignKey` directly.

pkg/resources/hybrid_table.go
sfc-gh-moczko: Done — extracted `readIndexState(indexes, indexErr, constraints, id)` which handles the WARN log, FK-backing filter, and state build. `d.Set("index", readIndexState(...))` now sits alongside the other constraint setters.

pkg/sdk/testint/hybrid_table_constraints_integration_test.go
sfc-gh-moczko: Done — replaced the `fmt.Sprintf` + `client.ExecForTests()` block with `testClientHelper().HybridTable.CreateWithRequest` using `OutOfLineConstraint` entries for `pk_child`, `uq_named`, the anonymous UNIQUE (no `Name` field → Snowflake assigns `SYS_CONSTRAINT_*`), and `fk_named`.

pkg/sdk/testint/hybrid_table_constraints_integration_test.go
sfc-gh-moczko: Done — both `TestInt_HybridTableConstraints` and `TestInt_HybridTableShowIndexes` have been moved into the existing `hybrid_tables_integration_test.go`. The two standalone files are deleted.

pkg/sdk/testint/hybrid_table_indexes_integration_test.go
sfc-gh-moczko: Done — moved into `hybrid_tables_integration_test.go` alongside the other SDK integration tests; `hybrid_table_indexes_integration_test.go` deleted.

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/hybrid_table_desc_output_ext.go
sfc-gh-moczko: Done: renamed HybridTableDescribeOutputAssert to HybridTableDescribeOutputRowAssert and constructor to HybridTableDescribeOutputRow; prefix now includes the row index for clearer error messages.

pkg/acceptance/bettertestspoc/config/model/hybrid_table_model_ext.go
sfc-gh-moczko: Done: replaced all manual make+for loops in the model ext with collections.Map.

pkg/resources/hybrid_table.go
sfc-gh-moczko: Done: trimmed the buildIndexesStateFromShowIndexes docstring to a single line.

pkg/testacc/resource_hybrid_table_acceptance_test.go
sfc-gh-moczko: Done: inlined at all 5 call sites and deleted the extracted helper.

pkg/testacc/resource_hybrid_table_acceptance_test.go
sfc-gh-moczko: Done: extracted into TestAcc_HybridTable_ColumnBehavior.

pkg/testacc/resource_hybrid_table_acceptance_test.go
sfc-gh-moczko: Done: inlined at both call sites and removed the closure.

pkg/testacc/resource_hybrid_table_acceptance_test.go
sfc-gh-moczko: Done: added HybridTableUniqueConstraintConfig + WithUniqueConstraints and WithNamedForeignKey to the model builder; replaced the raw tfconfig blocks.

pkg/testacc/resource_hybrid_table_acceptance_test.go
sfc-gh-moczko: The constraint and FK assertions now use TestCheckResourceAttrSet, TestCheckResourceAttr, TestCheckTypeSetElemNestedAttrs, and checkUniqueConstraintNamesReadBack as part of the merged CompleteUseCase Create step.

pkg/testacc/resource_hybrid_table_acceptance_test.go
sfc-gh-moczko: Done: deleted TestAcc_HybridTable_ConstraintNameReadback; its constraint-name assertions, FK assertions, and empty-plan regression step are now part of TestAcc_HybridTable_CompleteUseCase.

pkg/testacc/resource_hybrid_table_acceptance_test.go
sfc-gh-moczko: Done: deleted TestAcc_HybridTable_Index; the index assertion (lowercase IDX_NAME/name for case-suppression) and empty-plan step are folded into TestAcc_HybridTable_CompleteUseCase.

pkg/sdk/hybrid_table_constraints.go
sfc-gh-moczko: Done — the hand-rolled SQL helpers have been moved to the generator definition file (`hybrid_tables_def.go`) as three `CustomShowOperationWithPairedStructs` entries (`ShowPrimaryKeys`, `ShowUniqueKeys`, `ShowImportedKeys`). The generated types and options structs now live in `hybrid_tables_gen.go`.

pkg/sdk/hybrid_table_constraints.go
sfc-gh-moczko: Agreed — the file has been deleted and the three commands (`SHOW PRIMARY KEYS`, `SHOW UNIQUE KEYS`, `SHOW IMPORTED KEYS`) are now generator entries in `hybrid_tables_def.go`. The resulting types (`TablePrimaryKey`, `TableUniqueKey`, `TableImportedKey`) can be reused by other table types as needed.

pkg/sdk/hybrid_table_constraints_test.go
sfc-gh-moczko: File removed. The SQL shape tests have been replaced with `assertOptsValidAndSQLEquals` calls in `hybrid_tables_gen_test.go` (the standard generator test pattern), and the merge logic unit tests have been moved to `hybrid_tables_ext_test.go`.

sfc-gh-kwasilewski (commented):
Please fix failing integration tests and solve merge conflicts. Once those are green, we can get this PR merged.

sfc-gh-kwasilewski (approved):

sfc-gh-jcieslak (approved):

PR #4850, opened 2026-06-12
chore: Move objects to SDK generator - database roles

sfc-gh-asawicki:
Move the `DatabaseRoles` interface to SDK generator
- validations are organized slightly differently (alinged with the SDK generator available logic)
- unit tests are not touched
- dto builders have slightly different API; the calling code was adjusted
- the change was split into: rename - def - regeneration phases, so it might be easier to check changes between specific comments instead of verifying all changes
- `SHOW` odes not return the database name, it was previously being filled in the function method body - there is no hook for that in the SDK gen; it will be added and adjusted in one of the follow-up PRs
- resource logic was adjusted


pkg/sdk/generator/defs/database_role_def.go
```diff
@@ -0,0 +1,148 @@
+//go:build sdk_generation
```
sfc-gh-jmichalak: Nit: this is not needed.

pkg/sdk/testint/database_role_integration_test.go
```diff
@@ -280,11 +292,11 @@ func TestInt_DatabaseRoles(t *testing.T) {
 		err := client.Grants.GrantPrivilegeToShare(ctx, []sdk.ObjectPrivilege{sdk.ObjectPrivilegeUsage}, &sdk.ShareGrantOn{Database: testClientHelper().Ids.DatabaseId()}, share.ID())
 		require.NoError(t, err)
 
-		grantRequest := sdk.NewGrantDatabaseRoleToShareRequest(roleId, share.ID())
+		grantRequest := sdk.NewGrantToShareDatabaseRoleRequest(roleId).WithShare(share.ID())
```
sfc-gh-jmichalak: According to https://docs.snowflake.com/en/sql-reference/sql/grant-database-role-share, the share is required.

pkg/sdk/database_roles_ext.go
```diff
@@ -0,0 +1,75 @@
...
+func (s *RevokeDatabaseRoleRequest) WithAccountRole(accountRole AccountObjectIdentifier) *RevokeDatabaseRoleRequest {
+	return s.WithFrom(*NewDatabaseRoleKindOfRoleRequest().WithAccountRoleName(accountRole))
+}
+
+// WithDatabaseRole is a convenience method for revoking a database role from another database role.
+func (s *RevokeDatabaseRoleRequest) WithDatabaseRole(databaseRole DatabaseObjectIdentifier) *RevokeDatabaseRoleRequest {
+	return s.WithFrom(*NewDatabaseRoleKindOfRoleRequest().WithDatabaseRoleName(databaseRole))
+}
+
+func (r databaseRoleDBRow) additionalConvert(_ *DatabaseRole) error {
```
sfc-gh-jmichalak: This function is not needed.

sfc-gh-jmichalak (changes requested):

pkg/sdk/database_roles_ext.go
sfc-gh-asawicki: True, removed

pkg/sdk/generator/defs/database_role_def.go
sfc-gh-asawicki: True, removed

pkg/sdk/testint/database_role_integration_test.go
sfc-gh-asawicki: Marked as required (in the revoke too) and adjusted the callers.

sfc-gh-jmichalak (approved):

PR #4852, opened 2026-06-12
feat: Add snowflake_storage_lifecycle_policy resource

sfc-gh-kwasilewski:
Add a new preview resource for managing Snowflake storage lifecycle policies, gated behind the snowflake_storage_lifecycle_policy_resource preview feature flag.


pkg/resources/storage_lifecycle_policy.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+var storageLifecyclePolicySchema = map[string]*schema.Schema{
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      blocklistedCharactersFieldDescription("Specifies the identifier for the storage lifecycle policy; must be unique for the database and schema in which the storage lifecycle policy is created."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"schema": {
```
sfc-gh-jcieslak, line 29: Shouldn't schema and database be marked as ForceNew?

pkg/resources/storage_lifecycle_policy.go
```diff
@@ -0,0 +1,279 @@
...
+		MinItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"name": {
+					Type:        schema.TypeString,
+					Required:    true,
+					Description: "The argument name.",
+					ForceNew:    true,
+				},
+				// TODO(SNOW-1596962): Fully support VECTOR data type sdk.ParseFunctionArgumentsFromString could be a base for another function that takes argument names into consideration.
```
sfc-gh-jcieslak: Do we want the limitations to be at resource-level docs too/instead? The resource-level warning blocks have better visibility 🤔

pkg/provider/previewfeatures/preview_features.go
```diff
@@ -107,6 +107,7 @@ const (
 	StorageIntegrationAzureResource             feature = "snowflake_storage_integration_azure_resource"
 	StorageIntegrationGcsResource               feature = "snowflake_storage_integration_gcs_resource"
 	StorageIntegrationsDatasource               feature = "snowflake_storage_integrations_datasource"
+	StorageLifecyclePolicyResource              feature = "snowflake_storage_lifecycle_policy_resource"
```
sfc-gh-asawicki, line 110: nit: leaving it as an acceptance-test-visible-only is the preferred way of adding new resources (IMO we can leave it with this PR as-is)

pkg/resources/storage_lifecycle_policy.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func UpdateStorageLifecyclePolicy(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+	id, err := sdk.ParseSchemaObjectIdentifier(d.Id())
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	if d.HasChange("name") || d.HasChange("database") || d.HasChange("schema") {
```
sfc-gh-asawicki: nit: This logic will fail (e.g. for the case when database is renamed). It's fine to ether leave it, as it will be solved with thehierarchy renames experiment, but it could be also okay to marke db and schema as forcenew

pkg/resources/storage_lifecycle_policy.go
sfc-gh-asawicki: I don't think that comment is still valid - vector data type should be already fully supported for the new data types

pkg/testacc/resource_storage_lifecycle_policy_acceptance_test.go
```diff
@@ -0,0 +1,443 @@
...
+			},
+			// Import without optionals
+			{
+				Config:            config.FromModels(t, basic),
+				ResourceName:      ref,
+				ImportState:       true,
+				ImportStateVerify: true,
+				// Snowflake does not persist the VARCHAR length in the policy signature, so the configured
+				// VARCHAR(200) cannot be reproduced on import (DESCRIBE returns an unbounded VARCHAR).
+				ImportStateVerifyIgnore: []string{"argument.0.type"},
```
sfc-gh-asawicki: 1. Let's at least check it with ImportCheck
2. It's forceNew, do we have a similar problem as recently with catalog integrations? Is there a step confirming it's not an issue?

sfc-gh-asawicki (changes requested):

pkg/sdk/generator/defs/storage_lifecycle_policies_def.go
```diff
@@ -110,7 +110,7 @@ var storageLifecyclePoliciesDef = g.NewInterface(
 			DataType("return_type").
 			Text("body").
 			OptionalNumber("archive_for_days").
-			Text("archive_tier"),
+			Text("archive_tier", g.WithValueAdjuster("normalizeStorageLifecyclePolicyArchiveTier")),
```
sfc-gh-jmichalak, line 113: Nit: this could be an enum of COOL | COLD, right?

sfc-gh-jmichalak (changes requested):

pkg/sdk/generator/defs/storage_lifecycle_policies_def.go
sfc-gh-kwasilewski: It's a non-nullable string column with three possible values: `NULL`, `COOL`, or `COLD`.

pkg/testacc/resource_storage_lifecycle_policy_acceptance_test.go
sfc-gh-kwasilewski: 2. No, the problem affected catalog integrations because not all fields were set in the Read function. Here we set all fields.

pkg/sdk/generator/defs/storage_lifecycle_policies_def.go
sfc-gh-kwasilewski: Defining it as `Enum` might not be easy without declaring `NULL` as an enum value, which we probably shouldn't do.

pkg/resources/storage_lifecycle_policy.go
sfc-gh-kwasilewski: Session policy resource doesn't mark them as ForceNew - we decided back then it was fine to do that. Should we revisit that decision?

pkg/provider/previewfeatures/preview_features.go
sfc-gh-kwasilewski: I don't think we'll release the next version soon, and even if we do, we should be fine with having this resource available in the production provider.

pkg/resources/storage_lifecycle_policy.go
sfc-gh-kwasilewski: I switched from `ParseTableColumnSignature` to a new `ParseTableColumnSignatureWithVectorSupport` parser. The comment is no longer necessary.

pkg/resources/storage_lifecycle_policy.go
sfc-gh-kwasilewski: Discussed offline - I marked them as ForceNew, as this is more reliable than not having ForceNew and relying on the current error-prone renaming workflow.

pkg/resources/storage_lifecycle_policy.go
sfc-gh-kwasilewski: Adjusted, see https://github.com/snowflakedb/terraform-provider-snowflake/pull/4852#discussion_r3420195467

pkg/testacc/resource_storage_lifecycle_policy_acceptance_test.go
sfc-gh-kwasilewski: Replaced ImportStateVerify with ImportCheck, where we check every field, including arguments.

pkg/testacc/resource_storage_lifecycle_policy_acceptance_test.go
```diff
@@ -66,9 +64,10 @@ func TestAcc_StorageLifecyclePolicy_BasicUseCase(t *testing.T) {
 		},
 		{
 			Name: "ID",
-			Type: testdatatypes.DataTypeNumber,
+			Type: testdatatypes.DataTypeVectorFloat_768,
 		},
 	}
+	importedArguments := expectedSignature
```
sfc-gh-jcieslak, line 70: nit: is this needed?

sfc-gh-jcieslak (approved):

pkg/testacc/resource_storage_lifecycle_policy_acceptance_test.go
sfc-gh-kwasilewski: No, I just wanted to have a handy alias rather than using `expectedSignature` in a place where arguments are expected.

pkg/provider/previewfeatures/preview_features.go
sfc-gh-asawicki: Yup, just remember to initially create resouyrce test-only in the future changes

sfc-gh-asawicki (approved):

sfc-gh-jmichalak (approved):

PR #4855, opened 2026-06-12
feat: add per-plan SHOW GRANTS caching for snowflake_grant_account_role

raffael-dzikowski:
## Summary

- Adds a new `GRANT_ACCOUNT_ROLE_SHOW_CACHING` experiment for `snowflake_grant_account_role`
- Without caching, every resource instance issues an independent `SHOW GRANTS OF ROLE <name>` call during Read. In large RBAC configurations (thousands of `snowflake_grant_account_role` resources sharing the same set of roles), this produces N identical round-trips returning the same full result set — only 1 is needed per unique role per plan
- When the experiment is enabled, the first Read for a given role fetches and caches the result; subsequent Reads in the same plan reuse it. Cache is invalidated on Create and Delete so mutations within a single apply remain correctly visible

## Changes

- `pkg/internal/provider/show_grants_cache.go` — new `ShowGrantsOfRoleCache` using `sync.RWMutex` with double-checked locking and error-not-cached semantics
- `pkg/internal/provider/show_grants_cache_test.go` — 8 unit tests (miss, hit, invalidation, error handling, key independence, concurrent read/write safety)
- `pkg/internal/provider/provider_context.go` — `GrantShowOfRoleCache` field added to `Context`
- `pkg/provider/experimentalfeatures/experimental_features.go` — `GRANT_ACCOUNT_ROLE_SHOW_CACHING` experiment registered
- `pkg/provider/provider.go` — cache initialised on provider startup
- `pkg/resources/grant_account_role.go` — Read gated behind experiment; Create/Delete invalidate the cache entry
- `MIGRATION_GUIDE.md` — entry added under v2.17.0 → v2.18.0
- Docs regenerated via `make pre-push`

## Context

Observed in a production RBAC configuration with ~15,000 `snowflake_grant_account_role` resources. Plan time was dominated by redundant `SHOW GRANTS OF ROLE` calls — many resources share the same handful of roles, so the provider was issuing thousands of identical round-trips. This experiment reduces those to one call per unique role per plan.

No acceptance tests are included (require a live Snowflake connection); happy to add them if the team can point to the right test setup or preferred pattern for experiment-gated resource tests.


sfc-gh-asawicki (changes requested):
Generally LGTM, a few small adjustments before we proceed:
- (must) acceptance tests - we should have a test with the experiment enabled, you can base it on [https://github.com/snowflakedb/terraform-provider-snowflake/blob/611f95bcd195f60ca[…]estacc/resource_schema_experimental_features_acceptance_test.go](https://github.com/snowflakedb/terraform-provider-snowflake/blob/611f95bcd195f60ca4763110a9ebc55762b14770/pkg/testacc/resource_schema_experimental_features_acceptance_test.go)
- (nice-to-have) conditional read
- (nice-to-have) cache parametrized - for easier future use (but it can be done later on if we proceed with adding this behavior to more resources).

pkg/provider/provider.go
```diff
@@ -795,7 +795,9 @@ func ConfigureProvider(_ context.Context, s *schema.ResourceData) (any, diag.Dia
 		}
 	}
 
-	providerCtx := &provider.Context{}
+	providerCtx := &provider.Context{
+		GrantShowOfRoleCache: provider.NewCache[[]sdk.Grant](),
```
sfc-gh-asawicki, line 803: nit: it shouldnt be a big obstacle but I would rather initialize the cache only if the experiment is enabled, we can adjust it on our side

pkg/testacc/resource_grant_account_role_show_caching_experimental_features_acceptance_test.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+// countShowGrantsOfRoleQueries returns how many `SHOW GRANTS OF ROLE <id>` statements were issued
+// (across all sessions of the test user) for the given, test-unique role. Because the role name is
+// randomly generated per test, this count is isolated to the queries produced by the test under
+// inspection. Used to prove the GRANT_ACCOUNT_ROLE_SHOW_CACHING experiment collapses N identical
+// SHOW calls into one.
+func countShowGrantsOfRoleQueries(t *testing.T, roleId sdk.AccountObjectIdentifier) int {
```
sfc-gh-asawicki, line 28: nit: Let's extract this as a invokeable assert - we will do it on our side, it's not blocking

sfc-gh-asawicki (dismissed):

sfc-gh-jmichalak (dismissed):

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

Issue #4844, opened 2026-06-11
Request to support AGENTS in ownership grant construct


divyeshramani:
### Company Name

Walt Disney - Disney Experiences.

### Use Cases or Problem Statement

I am trying to add future ownership grant for cortex agents using Snowflake terraform provider, but it seems like it doesn’t support ownership grants for agents yet.  Is it possible to prioritize for next provider release so we can allow to share ownership of agents with all members of team.

Error: expected on.0.future.0.object_type_plural to be one of ["AGGREGATION POLICIES" "ALERTS" "AUTHENTICATION POLICIES" "COMPUTE POOLS" "DATA METRIC FUNCTIONS" "DATABASES" "DBT PROJECTS" "DYNAMIC TABLES" "EVENT TABLES" "EXTERNAL TABLES" "EXTERNAL VOLUMES" "FAILOVER GROUPS" "FILE FORMATS" "FUNCTIONS" "GIT REPOSITORIES" "HYBRID TABLES" "ICEBERG TABLES" "IMAGE REPOSITORIES" "INTEGRATIONS" "MATERIALIZED VIEWS" "NETWORK POLICIES" "NETWORK RULES" "PACKAGES POLICIES" "PIPES" "PROCEDURES" "MASKING POLICIES" "PASSWORD POLICIES" "PROJECTION POLICIES" "REPLICATION GROUPS" "RESOURCE MONITORS" "ROLES" "ROW ACCESS POLICIES" "SCHEMAS" "SESSION POLICIES" "SECRETS" "SEMANTIC VIEWS" "SEQUENCES" "STAGES" "STREAMS" "TABLES" "TAGS" "TASKS" "USERS" "VIEWS" "WAREHOUSES"], got AGENTS

It seems like grant_ownership construct doesn't support AGENTS. https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/grant_ownership

### Category

category:resource

### Object type(s)

resource:grant_ownership

### Proposal

Please add support for AGENTS 

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-kwasilewski:
Hi @divyeshramani, thank you for the request.

Support for AGENT objects in snowflake_grant_ownership has just been merged and will be available in the next provider release (v2.18.0).

I'll leave this issue open until the release goes out. Thanks for your patience!

sfc-gh-kwasilewski:
[...]

PR #4859, opened 2026-06-15
feat: Allow granting ownership on agents

sfc-gh-kwasilewski:
This updates the snowflake_grant_ownership resource to allow granting ownership on AGENT objects, including single object grants, bulk grants (ALL AGENTS IN ...), and future grants (FUTURE AGENTS IN ...).

References: #4844


pkg/sdk/grants_validations.go
```diff
@@ -19,6 +19,7 @@ var (
 
 // based on https://docs.snowflake.com/en/sql-reference/sql/grant-ownership#required-parameters
 var validGrantOwnershipObjectTypes = []ObjectType{
+	ObjectTypeAgent,
```
sfc-gh-jcieslak, line 22: I think it would be nice to have tests for three variants: single object, bulk `on_all`, bulk `on_future`. As far as I remember MCP Server and Agent types are returned as `CORTEX AGENT` and `CORTEX SERVER` which means some adjustments other than those may be needed.

sfc-gh-jmichalak: \+1, maybe we should (later) start extracting them similarly to what we have in tag associations.

sfc-gh-jmichalak (changes requested):

pkg/sdk/grants_validations.go
sfc-gh-asawicki: +1

sfc-gh-asawicki (changes requested):

pkg/sdk/grants_validations.go
sfc-gh-kwasilewski: What kind of tests do you mean? I see that not all object types are covered by either unit tests (e.g., `TestGrants_GrantOwnership`) or acceptance tests (`pkg/testacc/resource_grant_ownership_acceptance_test.go`).

Regarding how they are returned, yes, it's `CORTEX_AGENT` and `CORTEX_AGENT_SERVER`, respectively. But this is already handled [here](https://github.com/snowflakedb/terraform-provider-snowflake/blob/219bfe892ccb34a2a803004f4c7c10ef7cc4245b/pkg/sdk/grants.go#L246).

sfc-gh-jcieslak: I meant the acceptance tests to test the resource behavior. It's handled, but the test would verify it. I think we could still add those tests (idk if all three variants or just one to confirm) as it's just non trivial object type (has some additional logic around it).

sfc-gh-kwasilewski: Added two tests: one for singular AGENT and another for plural AGENTS.

sfc-gh-jcieslak (dismissed):

pkg/testacc/testdata/TestAcc_GrantOwnership/OnAgent_ToAccountRole/test.tf
```diff
@@ -0,0 +1,11 @@
+resource "snowflake_account_role" "test" {
```
sfc-gh-jmichalak: We should use model builders instead.

pkg/sdk/grants_validations.go
sfc-gh-jmichalak: LGTM, but IMO integration tests should be enough, since this is an SDK-only change. See TestInt_TagsAssociations.

sfc-gh-jmichalak (dismissed):

pkg/testacc/testdata/TestAcc_GrantOwnership/OnAgent_ToAccountRole/test.tf
sfc-gh-asawicki: +1

sfc-gh-asawicki (dismissed):

pkg/sdk/grants_validations.go
sfc-gh-kwasilewski: Yeah, integration tests are enough - they correctly fail if `CORTEX_AGENT` -> `AGENT` mapping is removed. I added them and removed acceptance tests.

pkg/testacc/testdata/TestAcc_GrantOwnership/OnAgent_ToAccountRole/test.tf
sfc-gh-kwasilewski: No longer relevant

sfc-gh-jmichalak (approved):

PR #4860, opened 2026-06-15
feat: create api integration azure api management resource

sfc-gh-jcieslak:
## Changes
- Create azure api management resource

## Next prs
- Register resource as production resource (in provider.go) and as preview resource
- Add examples, documentation, and migration guide entry
- Add a deprecation notice to the old API integration resource / data source


pkg/resources/api_integration_azure_api_management.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, err
+	}
+	d.SetId(helpers.EncodeResourceIdentifier(id))
+
+	details, err := client.ApiIntegrations.DescribeAzureDetails(ctx, id)
+	if err != nil {
+		return nil, fmt.Errorf("could not describe API integration %s during import: %w", id.FullyQualifiedName(), err)
+	}
+
+	normalizedProvider, err := sdk.ToApiIntegrationAzureApiProviderType(details.ApiProvider)
```
sfc-gh-asawicki, line 93: nit: I suppose this logic, parametrized with type, will be used in all the api integration resources, right? Could we extract it?

pkg/testacc/resource_api_integration_azure_api_management_acceptance_test.go
```diff
@@ -0,0 +1,421 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAcc_ApiIntegrationAzureApiManagement_BasicUseCase(t *testing.T) {
+	id := testClient().Ids.RandomAccountObjectIdentifier()
+
+	const azureTenantId = "00000000-0000-0000-0000-000000000000"
```
sfc-gh-asawicki: nit: these helper values are copy-pasted in so many places right now. Could we maybe extract them, at least for the given package?

sfc-gh-asawicki (dismissed):

pkg/resources/api_integration_azure_api_management.go
sfc-gh-jcieslak: Yup, will apply once all variants are merged

pkg/testacc/resource_api_integration_azure_api_management_acceptance_test.go
sfc-gh-jcieslak: Extracted

pkg/resources/api_integration_azure_api_management.go
sfc-gh-asawicki: leaving it open then

sfc-gh-asawicki (approved):

PR #4865, opened 2026-06-16
chore: Move objects to SDK generator - warehouses

sfc-gh-asawicki:
Move the `Warehouses` interface to SDK generator

- validations are organized slightly differently (alinged with the SDK generator available logic)
- unit tests were not regenerated
- previously, interface did not use the request dtos but opts directly; it was adjusted
- previously, all the logic was in a single file; it was split with the generation
- all callers were adjusted, including the plugin framework poc tests


pkg/sdk/generator/defs/warehouse_def.go
```diff
@@ -0,0 +1,249 @@
...
+		OptionalNumberAssignment("MAX_CONCURRENCY_LEVEL", g.ParameterOptions()).
+		OptionalNumberAssignment("STATEMENT_QUEUED_TIMEOUT_IN_SECONDS", g.ParameterOptions()).
+		OptionalNumberAssignment("STATEMENT_TIMEOUT_IN_SECONDS", g.ParameterOptions()).
+		OptionalTags().
+		WithValidation(g.ValidIdentifier, "name").
+		WithValidation(g.ConflictingFields, "OrReplace", "IfNotExists").
+		WithAdditionalValidations(),
+).CustomOperation(
+	"CreateAdaptive",
+	"https://docs.snowflake.com/en/LIMITEDACCESS/adaptive-warehouses",
```
sfc-gh-kwasilewski: Adaptive warehouses are no longer in LIMITEDACCESS. The link redirects to https://docs.snowflake.com/en/user-guide/warehouses-adaptive.

pkg/sdk/generator/defs/warehouse_def.go
```diff
@@ -0,0 +1,249 @@
...
+		OptionalEnumAssignment("SCALING_POLICY", scalingPolicyEnum, g.ParameterOptions().SingleQuotes()).
+		OptionalNumberAssignment("AUTO_SUSPEND", g.ParameterOptions()).
+		OptionalBooleanAssignment("AUTO_RESUME", nil).
+		OptionalBooleanAssignment("INITIALLY_SUSPENDED", nil).
+		OptionalIdentifier("ResourceMonitor", g.KindOfTPointer[sdkcommons.AccountObjectIdentifier](), g.IdentifierOptions().SQL("RESOURCE_MONITOR").Equals()).
+		OptionalComment().
+		OptionalBooleanAssignment("ENABLE_QUERY_ACCELERATION", nil).
+		OptionalNumberAssignment("QUERY_ACCELERATION_MAX_SCALE_FACTOR", g.ParameterOptions()).
+		OptionalEnumAssignment("RESOURCE_CONSTRAINT", warehouseResourceConstraintEnum, g.ParameterOptions().SingleQuotes()).
+		WithField(g.OptionalEnumLegacy[sdkcommons.WarehouseGeneration]("Generation", g.ParameterOptions().SingleQuotes().SQL("GENERATION"))).
```
sfc-gh-kwasilewski, line 156: These two fields are listed after WAREHOUSE_SIZE and before MAX_CLUSTER_COUNT, see https://docs.snowflake.com/en/sql-reference/sql/create-warehouse#syntax

pkg/sdk/generator/defs/warehouse_def.go
```diff
@@ -0,0 +1,249 @@
...
+	"https://docs.snowflake.com/en/LIMITEDACCESS/adaptive-warehouses",
+	g.NewQueryStruct("CreateAdaptiveWarehouse").
+		Create().
+		OrReplace().
+		SQL("WAREHOUSE").
+		IfNotExists().
+		Name().
+		SQLWithCustomFieldName("warehouseType", "WAREHOUSE_TYPE = 'ADAPTIVE'").
+		OptionalComment().
+		OptionalEnumAssignment("MAX_QUERY_PERFORMANCE_LEVEL", maxQueryPerformanceLevelEnum, g.ParameterOptions().SingleQuotes()).
```
sfc-gh-kwasilewski, line 175: According to https://docs.snowflake.com/en/user-guide/warehouses-adaptive#sql-reference, this field doesn't require any quotes. Should we adjust it?

pkg/sdk/generator/defs/warehouse_def.go
```diff
@@ -0,0 +1,249 @@
...
+	g.NewQueryStruct("AlterWarehouse").
+		Alter().
+		SQL("WAREHOUSE").
+		IfExists().
+		Name().
+		OptionalSQL("SUSPEND").
+		OptionalSQL("RESUME").
+		OptionalSQL("IF SUSPENDED").
+		OptionalSQL("ABORT ALL QUERIES").
+		Identifier("NewName", g.KindOfTPointer[sdkcommons.AccountObjectIdentifier](), g.IdentifierOptions().SQL("RENAME TO")).
```
sfc-gh-kwasilewski: `OptionalIdentifier`?

sfc-gh-kwasilewski: Also, we use the "RenameTo" name almost everywhere else in the SDK.

pkg/testacc/20_plugin_framework_poc_warehouse_test.go
```diff
@@ -763,23 +761,19 @@ func (r *WarehouseResource) Update(ctx context.Context, request resource.UpdateR
...
 			response.Diagnostics.AddError("Could not update (alter set) warehouse PoC", err.Error())
 			return
 		}
 	}
-	if (unset != sdk.WarehouseUnset{}) {
-		err := r.client.Warehouses.Alter(ctx, id, &sdk.AlterWarehouseOptions{
-			Unset: &unset,
-		})
+	if *unset != *sdk.NewWarehouseUnsetRequest() {
+		err := r.client.Warehouses.Alter(ctx, sdk.NewAlterWarehouseRequest(id).WithUnset(*unset))
```
sfc-gh-jcieslak, line 776: Question: Was it intended to use here Alter and in the set case AlterWithSuspend?

pkg/sdk/generator/defs/warehouse_def.go
sfc-gh-jcieslak: nit: Could be changed to https://docs.snowflake.com/en/user-guide/warehouses-adaptive

sfc-gh-jcieslak (dismissed):

pkg/sdk/warehouses_ext.go
```diff
@@ -0,0 +1,354 @@
...
+	if other := strings.TrimSpace(r.Other); other != "" {
+		if val, err := strconv.ParseFloat(other, 64); err != nil {
+			return fmt.Errorf(`row 'other' has incorrect value '%s', %w`, other, err)
+		} else {
+			wh.Other = val
+		}
+	}
+
+	// ResourceMonitor - "null" means not set
+	if r.ResourceMonitor != "null" {
```
sfc-gh-kwasilewski: This can be handled by `g.WithCustomParser` and a parser that returns an empty string on "null".

pkg/sdk/warehouses_ext.go
```diff
@@ -0,0 +1,354 @@
...
+			errs = append(errs, fmt.Errorf("MinClusterCount must be less than or equal to MaxClusterCount"))
+		}
+	}
+	if v.AutoSuspend != nil {
+		if ok := validateIntGreaterThanOrEqual(*v.AutoSuspend, 0); !ok {
+			errs = append(errs, fmt.Errorf("AutoSuspend must be greater than or equal to 0"))
+		}
+	}
+	if v.QueryAccelerationMaxScaleFactor != nil {
+		if ok := validateIntInRangeInclusive(*v.QueryAccelerationMaxScaleFactor, 0, 100); !ok {
```
sfc-gh-kwasilewski, line 331: I thought we didn't usually perform such strict client-side validations.

pkg/testacc/20_plugin_framework_poc_warehouse_test.go
sfc-gh-asawicki: It should be `AlterWithSuspend`, I'll verify other invocations, good catch.

sfc-gh-asawicki: After giving it a second thought, the current suspend logic triggers only with Set being Set, so all Unset/Rename invocations can continue using Alter.

sfc-gh-asawicki: Adjusted a few invocations in integration tests.

pkg/sdk/generator/defs/warehouse_def.go
sfc-gh-asawicki: changed

sfc-gh-asawicki: changed

pkg/sdk/generator/defs/warehouse_def.go
sfc-gh-asawicki: This change was supposed to migrate the object in its current state, without adding missing fields/making additional adjustments. The ordering before was: https://github.com/snowflakedb/terraform-provider-snowflake/blob/219bfe892ccb34a2a803004f4c7c10ef7cc4245b/pkg/sdk/warehouses.go#L293.

Let me verify it in a separate change.

pkg/sdk/generator/defs/warehouse_def.go
sfc-gh-asawicki: As above, https://github.com/snowflakedb/terraform-provider-snowflake/blob/219bfe892ccb34a2a803004f4c7c10ef7cc4245b/pkg/sdk/warehouses.go#L350.

We can verify in a follow up.

pkg/sdk/generator/defs/warehouse_def.go
sfc-gh-asawicki: As it's naming only, and not behavior, I'll adjust it in this change.

pkg/sdk/warehouses_ext.go
sfc-gh-asawicki: There are a few places additionally handling "null" before conversion (e.g. functions, procedures), but in all cases the additional check is added in the _ext file. I can add the parse method adding the conditional "null" check. Still, I think that "null" verification should be handled on later on.

pkg/sdk/warehouses_ext.go
sfc-gh-asawicki: It depends - for some cases, we validate the accepted int values. This is what was there previously: https://github.com/snowflakedb/terraform-provider-snowflake/blob/219bfe892ccb34a2a803004f4c7c10ef7cc4245b/pkg/sdk/warehouses.go#L481

pkg/sdk/warehouses_ext.go
sfc-gh-asawicki: Done

pkg/sdk/generator/defs/warehouse_def.go
sfc-gh-asawicki: Done

pkg/sdk/warehouses_ext.go
sfc-gh-kwasilewski: Not everywhere. I've just had to deal with "NULL" strings in storage lifecycle policies, where I used `g.WithCustomParser`. IMO, manual converts should be used in more complex cases, where none of the other solutions work.

pkg/sdk/warehouses_ext.go
sfc-gh-kwasilewski: Fair enough, let's not change the behavior. Still, if I were to implement this resource for the first time, I'd only check if a number is positive, without hardcoding any values in the provider code.

sfc-gh-kwasilewski (approved):

PR #4871, opened 2026-06-16
chore: Add SYSTEM$CLUSTERING_INFORMATION SDK support and iceberg cluster-by validation

sfc-gh-jmichalak:
## Changes

- **`SystemFunctions.GetClusteringInformation`** — new SDK method wrapping `SYSTEM$CLUSTERING_INFORMATION`. The clustering key is not returned by `SHOW`/`DESCRIBE` for Iceberg tables, so this is the way to read/verify it. Accepts optional explicit columns (variadic) to compute clustering information against an arbitrary key. Parses the JSON payload into a typed `ClusteringInformation` struct (with unit test coverage for parsing).

- **`PARTITION BY` / `CLUSTER BY` mutual-exclusion validation** — Snowflake rejects setting both on `CREATE ICEBERG TABLE` with error `099207 (0A000): "Partition by and Cluster by cannot be specified together for Iceberg tables."`, even though the [docs](https://docs.snowflake.com/en/sql-reference/sql/create-iceberg-table-snowflake) list both options without that caveat. Added a `ConflictingFields` validation between `PartitionBy` and `ClusterBy` (regenerated via the SDK generator) and a TODO in the def file to report the docs discrepancy to Snowflake.

- **Integration tests** — split the clustering coverage out of the partitioned "all options" case into a dedicated cluster-by test (since the two are mutually exclusive), and assert the clustering key via `GetClusteringInformation`. Added `TestInt_GetClusteringInformation` covering the defined-key, explicit-columns, and unclustered (error) paths.

- **Unit tests** — added a conflicting-fields validation test and a valid `CLUSTER BY` SQL test for `CREATE ICEBERG TABLE`.

## Notes

- The SDK was regenerated with the object filter (`--filter-object-names=IcebergTables`); only the iceberg validations were affected.
- Live integration tests require a Snowflake account and were not run in this environment; unit tests pass and the `testint` package compiles.

### DIFF
--- a/pkg/sdk/generator/defs/iceberg_tables_def.go
+++ b/pkg/sdk/generator/defs/iceberg_tables_def.go
@@ -189,7 +189,7 @@ var icebergTablesDef = g.NewInterface(
 		QueryStructField("ColumnsAndConstraints", icebergTableColumnsAndConstraints, g.ListOptions().Parentheses().Required()).
 		ListQueryStructField("PartitionBy", icebergTablePartitionExpression, g.KeywordOptions().Parentheses().SQL("PARTITION BY")).
 		OptionalAssignment("PATH_LAYOUT", IcebergTablePathLayoutEnumDef.KindPtr(), g.ParameterOptions().NoQuotes()).
-		// TODO: report docs discrepancy to Snowflake - the CREATE ICEBERG TABLE docs list both PARTITION BY and
+		// TODO [next PRs]: report docs discrepancy to Snowflake - the CREATE ICEBERG TABLE docs list both PARTITION BY and
 		// CLUSTER BY as available options without noting they are mutually exclusive, but Snowflake rejects setting both with
 		// error 099207 (0A000): "Partition by and Cluster by cannot be specified together for Iceberg tables." This is enforced
 		// by the ConflictingFields validation below.
--- a/pkg/sdk/generator/defs/iceberg_tables_def.go
+++ b/pkg/sdk/generator/defs/iceberg_tables_def.go
@@ -189,6 +189,10 @@ var icebergTablesDef = g.NewInterface(
 		QueryStructField("ColumnsAndConstraints", icebergTableColumnsAndConstraints, g.ListOptions().Parentheses().Required()).
 		ListQueryStructField("PartitionBy", icebergTablePartitionExpression, g.KeywordOptions().Parentheses().SQL("PARTITION BY")).
 		OptionalAssignment("PATH_LAYOUT", IcebergTablePathLayoutEnumDef.KindPtr(), g.ParameterOptions().NoQuotes()).
+		// TODO: report docs discrepancy to Snowflake - the CREATE ICEBERG TABLE docs list both PARTITION BY and
+		// CLUSTER BY as available options without noting they are mutually exclusive, but Snowflake rejects setting both with
+		// error 099207 (0A000): "Partition by and Cluster by cannot be specified together for Iceberg tables." This is enforced
+		// by the ConflictingFields validation below.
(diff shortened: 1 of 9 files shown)


pkg/sdk/generator/defs/iceberg_tables_def.go
```diff
@@ -189,6 +189,10 @@ var icebergTablesDef = g.NewInterface(
 		QueryStructField("ColumnsAndConstraints", icebergTableColumnsAndConstraints, g.ListOptions().Parentheses().Required()).
 		ListQueryStructField("PartitionBy", icebergTablePartitionExpression, g.KeywordOptions().Parentheses().SQL("PARTITION BY")).
 		OptionalAssignment("PATH_LAYOUT", IcebergTablePathLayoutEnumDef.KindPtr(), g.ParameterOptions().NoQuotes()).
+		// TODO: report docs discrepancy to Snowflake - the CREATE ICEBERG TABLE docs list both PARTITION BY and
```
sfc-gh-asawicki: nit: TODO without jira ticket/next PRs marking

pkg/sdk/system_functions.go
```diff
@@ -237,3 +241,77 @@ func (c *systemFunctions) GetIcebergTableInformation(ctx context.Context, id Sch
...
+	}{}
+	var columnsArg string
+	if len(columns) > 0 {
+		columns = collections.Map(columns, func(col string) string {
+			return DoubleQuotes.Modify(col)
+		})
+		columnsArg = `, ` + SingleQuotes.Modify(fmt.Sprintf("(%s)", strings.Join(columns, ", ")))
+	}
+	idEscaped := SingleQuotes.Modify(id.FullyQualifiedNameEscaped())
+	sql := fmt.Sprintf(`SELECT SYSTEM$CLUSTERING_INFORMATION(%s%s) AS "CLUSTERING_INFORMATION"`, idEscaped, columnsArg)
```
sfc-gh-asawicki: we should try to model it using our normal SDK struct tags, so that the comma joining and escaping happens in the sql builder.

sfc-gh-asawicki (changes requested):

pkg/sdk/generator/defs/iceberg_tables_def.go
sfc-gh-jmichalak: Done.

pkg/sdk/system_functions.go
sfc-gh-jmichalak: Done.

pkg/sdk/system_functions.go
```diff
@@ -278,22 +278,39 @@ func (c *systemFunctions) GetClusteringInformation(ctx context.Context, id Schem
...
+	if err := c.client.queryOne(ctx, row, sql); err != nil {
+		return nil, err
+	}
 	return parseClusteringInformation(row.Info)
 }
 
+type getClusteringInformationOptions struct {
+	selectSystemClusteringInformation bool                       `ddl:"static" sql:"SELECT SYSTEM$CLUSTERING_INFORMATION"`
+	arguments                         *clusteringInformationArgs `ddl:"list,parentheses,must_parentheses"`
+	as                                bool                       `ddl:"static" sql:"AS \"CLUSTERING_INFORMATION\""`
```
sfc-gh-asawicki, line 302: q: do we need `\"` there? I mean, if done without double quoutes in means the same as its uppercased on SF side, right?

sfc-gh-asawicki (approved):

pkg/sdk/system_functions.go
sfc-gh-jmichalak: Maybe not, as this should get uppercased. Let's leave this for the system functions rework/rewrite.

PR #4874, opened 2026-06-16
feat: create api integration git repository oauth2 resource

sfc-gh-jcieslak:
## Changes
- Create git repository oauth2 api integration resource

## Next prs
- Document import limitation `oauth_client_secret` that after import the resource may still plan re-creation (unless we want to ignore changes from empty to filled that would be only possible if object was imported as during creation the value is required and should be non empty)
- Register resource as production resource (in provider.go) and as preview resource
- Add examples, documentation, and migration guide entry
- Add a deprecation notice to the old API integration resource / data source


pkg/resources/api_integration_git_repository_oauth2.go
```diff
@@ -0,0 +1,261 @@
...
+			Description:  "The OAuth 2.0 token endpoint for the Git repository.",
+		},
+		"oauth_client_id": {
+			Type:         schema.TypeString,
+			Required:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.StringIsNotEmpty,
+			Description:  "The client ID for the OAuth 2.0 application.",
+		},
+		"oauth_client_secret": {
```
sfc-gh-jcieslak, line 45: For this field, we should either:
- Document import limitation: oauth_client_secret, that after import, the resource may still plan re-creation 
- Ignore changes from empty to filled value (the value is required by create, so only the case of it missing would be during import)

pkg/resources/api_integration_git_repository_oauth2.go
```diff
@@ -0,0 +1,262 @@
...
+			Description:  "Specifies the validity period (in seconds) for the OAuth 2.0 access token.",
+		},
+		"oauth_refresh_token_validity": {
+			Type:         schema.TypeInt,
+			Optional:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.IntAtLeast(1),
+			Description:  "Specifies the validity period (in seconds) for the OAuth 2.0 refresh token.",
+		},
+		"oauth_allowed_scopes": {
```
sfc-gh-jmichalak, line 66: Q: does the order matter here? I suppose it shouldn't.

pkg/resources/api_integration_git_repository_oauth2.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	if err = errors.Join(
+		intAttributeCreate(d, "oauth_access_token_validity", &authReq.OauthAccessTokenValidity),
+		intAttributeCreate(d, "oauth_refresh_token_validity", &authReq.OauthRefreshTokenValidity),
+		stringAttributeCreate(d, "oauth_username", &authReq.OauthUsername),
+	); err != nil {
+		return diag.FromErr(err)
+	}
+
+	oauth2Params := sdk.NewGitHttpsApiOAuth2ParamsRequest().WithApiUserAuthentication(*authReq)
```
sfc-gh-jmichalak, line 181: Nit: I think the SDK is incorrect, and authReq should be required, basing on this example.

sfc-gh-jmichalak (dismissed):

pkg/resources/api_integration_git_repository_oauth2.go
```diff
@@ -0,0 +1,262 @@
...
+			Description:  "The OAuth 2.0 authorization endpoint for the Git repository.",
+		},
+		"oauth_token_endpoint": {
+			Type:         schema.TypeString,
+			Required:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.StringIsNotEmpty,
+			Description:  "The OAuth 2.0 token endpoint for the Git repository.",
+		},
+		"oauth_client_id": {
```
sfc-gh-kwasilewski, line 37: In other resources (catalog integrations, api authentication integration), this field is marked as sensitive.

pkg/resources/api_integration_git_repository_oauth2.go
```diff
@@ -0,0 +1,262 @@
...
+			Optional:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.IntAtLeast(1),
+			Description:  "Specifies the validity period (in seconds) for the OAuth 2.0 refresh token.",
+		},
+		"oauth_allowed_scopes": {
+			Type:        schema.TypeList,
+			Optional:    true,
+			ForceNew:    true,
+			Description: fmt.Sprintf("Specifies a list of scopes to use when making a request from the OAuth by a role with USAGE on the integration. Valid values are (case-insensitive): %s.", possibleValuesListed(sdk.AsStringList(sdk.AllApiIntegrationOauthAllowedScopes))),
```
sfc-gh-kwasilewski: nit: `enumValuesDescription` is a slightly better choice, because it also encapsulates "Valid values are (case-insensitive): " prefix.

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
```diff
@@ -0,0 +1,439 @@
...
+					},
+				},
+				Config: config.FromModels(t, withAllOptionals),
+				Check:  assertThat(t, assertWithAllOptionals...),
+			},
+		},
+	})
+}
+
+func TestAcc_ApiIntegrationGitRepositoryOauth2_CompleteUseCase(t *testing.T) {
```
sfc-gh-kwasilewski, line 233: How does this test differ from the last create test step above? It's all the same to me - except for the import with all optionals, but this can be moved.

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
```diff
@@ -0,0 +1,439 @@
...
+			{
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionUpdate),
+					},
+				},
+				Config: config.FromModels(t, basic),
+				Check:  assertThat(t, assertBasic...),
+			},
+			// Update - external changes to comment
```
sfc-gh-kwasilewski, line 172: We should also detect external changes to other settable attributes: enabled, api_allowed_prefixes, and api_blocked_prefixes.

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
```diff
@@ -0,0 +1,439 @@
...
+					require.ErrorIs(t, err, sdk.ErrObjectNotFound)
+				},
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectResourceAction(ref, plancheck.ResourceActionCreate),
+					},
+				},
+				Config: config.FromModels(t, withAllOptionals),
+				Check:  assertThat(t, assertWithAllOptionals...),
+			},
```
sfc-gh-kwasilewski: We need test steps for:
- updating ForceNew fields
- external changes to ForceNew fields

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
```diff
@@ -0,0 +1,439 @@
...
+		resourceshowoutputassert.ApiIntegrationGitHttpsApiDescribeOutput(t, ref).
+			HasUserAuthType(string(sdk.ApiIntegrationUserAuthTypeOauth2)).
+			HasOauthClientId(clientId).
+			HasOauthTokenEndpoint(tokenEndpoint).
+			HasOauthAuthorizationEndpoint(authorizationEndpoint).
+			HasOauthAccessTokenValidity(accessTokenValidity).
+			HasOauthRefreshTokenValidity(refreshTokenValidity).
+			HasOauthUsername(oauthUsername).
+			HasComment(comment),
+		objectassert.ApiIntegrationGitHttpsApiDetails(t, id).
```
sfc-gh-kwasilewski: Any reason for using objectassert and not resourceshowoutputassert?

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
```diff
@@ -0,0 +1,439 @@
...
+				ResourceName:  oauth2Model.ResourceReference(),
+				ImportState:   true,
+				ImportStateId: githubAppIntegration.ID().Name(),
+				ExpectError:   regexp.MustCompile("not compatible with snowflake_api_integration_git_repository_oauth2"),
+			},
+		},
+	})
+}
+
+func TestAcc_ApiIntegrationGitRepositoryOauth2_Import(t *testing.T) {
```
sfc-gh-kwasilewski, line 441: I don't think we need extra tests for import. ImportStateVerify from the previous tests already covers everything IMO (yes, there's one field in ImportStateVerifyIgnore, but that's totally fine since it isn't in DESC output).

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
```diff
@@ -0,0 +1,439 @@
...
+						planchecks.ExpectChange(testModel.ResourceReference(), "oauth_client_secret", tfjson.ActionDelete, nil, sdk.String(clientSecret)),
+					},
+					PostApplyPostRefresh: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+			},
+		},
+	})
+}
```
sfc-gh-kwasilewski, line 493: How about validation tests? We do have some fields with validations (non-empty strings, positive integers).

pkg/resources/api_integration_git_repository_oauth2.go
sfc-gh-jcieslak: Don't think so

pkg/resources/api_integration_git_repository_oauth2.go
sfc-gh-jcieslak: Good point, will adjust in the follow-up

pkg/resources/api_integration_git_repository_oauth2.go
sfc-gh-jcieslak: Will mark as sensitive then

pkg/resources/api_integration_git_repository_oauth2.go
sfc-gh-jcieslak: Good point, will adjust

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-jcieslak: That's what we added in other cases; this mostly checks the import with optionals. I usually approach this by handling non-forced new fields in the basic use case and handling imports in the complete use case to avoid re-creations.

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-jcieslak: Will adjust

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-jcieslak: Idk if `updating ForceNew fields` is necessary as output is known, `external changes to ForceNew fields` will be added as it checks if external changes for those fields are detected properly (would require plancheks on those field with Delete action).

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-jcieslak: I mean, I added them because of the comment from the previous resources. I think we should now add those additional import tests to make sure the import works as we expect to (like in the case of catalog integration).

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-jcieslak: Will add those

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-jcieslak: I used both, will make sure both have same amount of checks

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-kwasilewski: We could also add test steps for `StringIsNotEmpty` validations. This can be done in a follow-up.

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-kwasilewski: You're right that importing needs special care, but now we have two tests for the same case: `Import with all attributes` above and this one. I'm just wondering if we could remove one of them to keep the overall test execution speed under control. 

Since this test checks more than the other one (it expects destroy and create action, and expects a change in oauth_client_secret), we could keep it and remove the other one. I'll leave it to you; it doesn't block the PR.

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-kwasilewski: What I mean is that basically, the `withAllOptionals` model is the same as the `allAttributes` model. The basic use case ends with creating a model with all optionals, but that's what you're already doing in the complete use case.

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-jcieslak: I would leave both of them for an additional correctness check

pkg/testacc/resource_api_integration_git_repository_oauth2_acceptance_test.go
sfc-gh-jcieslak: Removed last steps and left complete use case test

sfc-gh-kwasilewski (dismissed):

sfc-gh-kwasilewski (approved):

PR #4881, opened 2026-06-17
chore: Rename in hierarchies experiment - 3-level hierarchy in snowflake_table

sfc-gh-asawicki:
Continuation to https://github.com/snowflakedb/terraform-provider-snowflake/pull/4825

- Logic extended to 3-level hierarchy
  - case A: database name changes, schema name unchanged - the logic is the same as for database-schema
  - case B: schema name changes, database name unchanged - the logic is the same as for database-schema
  - case C: both database and schema names change - there are 4 cases:
    - move + move
    - move + rename
    - rename + move
    - rename + rename
- The logic is currently directly inside the table resource - it will be extracted and simplified to be able to reuse it later in other resources - it will be done in the next PR

### DIFF
--- a/pkg/resources/table.go
+++ b/pkg/resources/table.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/helpers"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/internal/provider"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/experimentalfeatures"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/previewfeatures"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/schemas"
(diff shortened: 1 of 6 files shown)


pkg/resources/table.go
```diff
@@ -722,10 +725,201 @@ func ReadTable(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagn
...
+			oldDatabaseExists := errOldDb == nil
+
+			// Probe schemas
+			schemaInNewDbOldName := sdk.NewDatabaseObjectIdentifier(newDatabaseName, oldSchemaName)
+			schemaInNewDbNewName := sdk.NewDatabaseObjectIdentifier(newDatabaseName, newSchemaName)
+			schemaInOldDbOldName := sdk.NewDatabaseObjectIdentifier(oldDatabaseName, oldSchemaName)
+			schemaInOldDbNewName := sdk.NewDatabaseObjectIdentifier(oldDatabaseName, newSchemaName)
+
+			_, errSchemaNewDbOld := client.Schemas.ShowByID(ctx, schemaInNewDbOldName)
+			schemaNewDbOldExists := errSchemaNewDbOld == nil
```
sfc-gh-jmichalak, line 852: Please adjust the variable names. E.g. this one reads as "a new schema in an old db" which is not the case. Let's be more verbose here.

sfc-gh-jmichalak (approved):

pkg/resources/table.go
sfc-gh-asawicki: As discussed, I will add in the follow-up PR

sfc-gh-asawicki: Addressed in https://github.com/snowflakedb/terraform-provider-snowflake/pull/4891

PR #4884, opened 2026-06-18
feat: add snowflake_iceberg_table_from_delta_files resource (preview)

sfc-gh-jmichalak:
## Summary

Adds a new preview resource `snowflake_iceberg_table_from_delta_files` for managing Snowflake Iceberg tables created from Delta table files in an external volume ([CREATE ICEBERG TABLE ... DELTA](https://docs.snowflake.com/en/sql-reference/sql/create-iceberg-table-delta)).

Stacked on top of `iceberg-files-resource`.

## Changes

- **New resource** `snowflake_iceberg_table_from_delta_files` with full CRUD + import, registered as a preview feature.
  - `base_location` (ForceNew) with trailing-slash diff suppression.
  - `auto_refresh` write-only-style field (Snowflake does not read it back; only `auto_refresh_status` is exposed by `SHOW`).
- **Refactor:** extracted the schema, parameters, read/import/update/delete logic shared with `iceberg_table_from_files` into `pkg/resources/iceberg_table_commons.go`. The existing files resource is migrated onto it with no behavioral change.
- **SDK:** `auto_refresh_status` is now parsed from its JSON blob into a structured `IcebergTableAutoRefreshStatus` and exposed as a nested object in `show_output` (previously a raw string). Asserts and integration tests updated accordingly.
- **Generator:** the `JsonField` mapping in `convert.tmpl` now guards the `json.Unmarshal` for pointer-kind fields so an empty column no longer errors. This makes the previously hand-edited guard regeneration-safe (no manual tweak to a generated file).
- Docs, examples, migration guide, and preview-feature list updated.


templates/resources/iceberg_table_from_delta_files.md.tmpl
sfc-gh-kwasilewski, line 1: I'm pretty sure we don't need a specialized template, it looks almost identical to the default preview resource template.

pkg/resources/iceberg_table_from_delta_files.go
```diff
@@ -0,0 +1,147 @@
...
+	icebergTableCommonSchema(),
+	map[string]*schema.Schema{
+		"base_location": {
+			Type:         schema.TypeString,
+			Required:     true,
+			ForceNew:     true,
+			ValidateFunc: validation.StringIsNotEmpty,
+			Description:  "Specifies the relative path of the Delta table's directory in the external volume. Cannot be changed after creation.",
+			DiffSuppressFunc: SuppressIfAny(
+				IgnoreChangeToCurrentSnowflakeValueInShow("base_location"),
```
sfc-gh-kwasilewski: This field is normally set in the Read function, so this is not necessary, right?

pkg/resources/iceberg_table_from_delta_files.go
```diff
@@ -0,0 +1,147 @@
...
+
+		Description: "Resource used to manage an Iceberg table whose metadata is created from Delta table files in an external volume. For more information, check [the official documentation](https://docs.snowflake.com/en/sql-reference/sql/create-iceberg-table-delta).",
+
+		Schema: icebergTableFromDeltaFilesSchema,
+		Importer: &schema.ResourceImporter{
+			StateContext: TrackingImportWrapper(resources.IcebergTableFromDeltaFiles, importIcebergTable),
+		},
+		Timeouts: defaultTimeouts,
+		CustomizeDiff: customdiff.All(
+			ComputedIfAnyAttributeChanged(icebergTableFromDeltaFilesSchema, ShowOutputAttributeName, "comment", "base_location", "auto_refresh"),
```
sfc-gh-kwasilewski: nit: changes to `base_location` trigger recreation anyway

```suggestion
			ComputedIfAnyAttributeChanged(icebergTableFromDeltaFilesSchema, ShowOutputAttributeName, "comment", "auto_refresh"),
```

sfc-gh-jmichalak: Done.

templates/resources/iceberg_table_from_delta_files.md.tmpl
sfc-gh-jmichalak: Removed.

pkg/resources/iceberg_table_from_delta_files.go
sfc-gh-jmichalak: Removed.

sfc-gh-kwasilewski (dismissed):

sfc-gh-asawicki (dismissed):

sfc-gh-kwasilewski (approved):

PR #4889, opened 2026-06-18
feat: Allow assigning a storage lifecycle policy to a table in the SDK

sfc-gh-kwasilewski:
Add SDK support for ALTER TABLE ... ADD/DROP STORAGE LIFECYCLE POLICY plus integration tests.


pkg/sdk/tables_test.go
```diff
@@ -1449,6 +1470,40 @@ func TestTableAlter(t *testing.T) {
...
 	})
+
+	t.Run("add storage lifecycle policy", func(t *testing.T) {
+		storageLifecyclePolicyId := randomSchemaObjectIdentifier()
+
+		opts := &alterTableOptions{
+			name: id,
+			AddStorageLifecyclePolicy: &TableAddStorageLifecyclePolicy{
+				StorageLifecyclePolicy: storageLifecyclePolicyId,
+				On:                     []string{"FIRST_COLUMN"},
```
sfc-gh-jmichalak: I think this should be wrapped in double quotes, usually table columns are case sensitive.

sfc-gh-jmichalak (changes requested):

pkg/sdk/tables_test.go
sfc-gh-kwasilewski: Fixed

sfc-gh-jmichalak (approved):

PR #4891, opened 2026-06-18
chore: Rename in hierarchies - extract and refactor

sfc-gh-asawicki:
Continuation to https://github.com/snowflakedb/terraform-provider-snowflake/pull/4885

- The logic is cleaned up and extracted, so that we can much more easily add it to other resources later on.
- There are 2 methods: 1 for shallow hierarchy (2 levels), 1 for deeper hierarchy (3 levels, assuming database and schema to be the top ones).
- Table and schema resources adjusted.
- Naming cleaned up after review comment from #4885

### DIFF
--- a/pkg/resources/hierarchy_renames.go
+++ b/pkg/resources/hierarchy_renames.go
@@ -1,8 +1,10 @@
 package resources
 
 import (
+	"context"
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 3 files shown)


pkg/resources/hierarchy_renames.go
```diff
@@ -0,0 +1,267 @@
...
+// for a database-level object (database → object).
+// OR is the return type of objectShowByIDFn.
+func handleTwoLevelHierarchyRename[OR any](
+	ctx context.Context,
+	d *schema.ResourceData,
+	client *sdk.Client,
+	id *sdk.DatabaseObjectIdentifier,
+	objectRenameFn func(sdk.DatabaseObjectIdentifier, sdk.DatabaseObjectIdentifier) func() error,
+	objectShowByIDFn func(context.Context, sdk.DatabaseObjectIdentifier) (OR, error),
+	encodeIdFn func(sdk.DatabaseObjectIdentifier) string,
```
sfc-gh-jcieslak, line 24: Do we need to make it a parameter when it should be pretty much always EncodeResourceIdentifier?

pkg/resources/hierarchy_renames.go
```diff
@@ -0,0 +1,267 @@
...
+	_, errNewParent := parentShowByIDFn(ctx, newParentId)
+	newParentExists := errNewParent == nil
+	_, errOldParent := parentShowByIDFn(ctx, oldParentId)
+	oldParentExists := errOldParent == nil
+	_, errTargetObject := objectShowByIDFn(ctx, targetObjectId)
+	objectAtTargetExists := errTargetObject == nil
+	_, errCurrentObject := objectShowByIDFn(ctx, currentObjectId)
+	objectAtSourceExists := errCurrentObject == nil
+
+	capitalize := func(s string) string {
```
sfc-gh-jcieslak, line 148: nit: there's `strings.ToTitle`

sfc-gh-jcieslak (approved):

pkg/resources/hierarchy_renames.go
sfc-gh-asawicki: For now, we need to use the old table resource, as the new one is not available. Later on, we will be able to remove it.
I will add a TODO in the follow up PR, oke?

pkg/resources/hierarchy_renames.go
sfc-gh-asawicki: Follow-up, oke?

PR #4901, opened 2026-06-24
feat: Add snowflake_table_storage_lifecycle_policy_attachment resource

sfc-gh-kwasilewski:
Add a new preview resource that attaches a storage lifecycle policy to a table or a dynamic table.


examples/resources/snowflake_table_storage_lifecycle_policy_attachment/resource.tf
```diff
@@ -0,0 +1,36 @@
...
+resource "snowflake_table_storage_lifecycle_policy_attachment" "table_attachment" {
+  table_name                    = snowflake_table.table.fully_qualified_name
+  table_type                    = "TABLE"
+  storage_lifecycle_policy_name = snowflake_storage_lifecycle_policy.slp.fully_qualified_name
+  on                            = ["VAL"]
+}
+
+# Attaching a storage lifecycle policy to a dynamic table
+resource "snowflake_table_storage_lifecycle_policy_attachment" "dynamic_table_attachment" {
+  table_name                    = "\"prod\".\"security\".\"my_dynamic_table\""
```
sfc-gh-jmichalak: A dynamic table also has a fully qualified name field.

pkg/testacc/1_setup_provider_test.go
```diff
@@ -46,9 +46,7 @@ var (
 	tagsProviderFactory                              = providerFactoryUsingCache("Tags")
 	tagsWithExperimentFlagProviderFactory            = providerFactoryUsingCache("TagsWithExperimentFlag")
 	servicesProviderFactory                          = providerFactoryUsingCache("Services")
-	userPasswordPoliciesProviderFactory              = providerFactoryUsingCache("UserPasswordPolicies")
-	userAuthenticationPoliciesProviderFactory        = providerFactoryUsingCache("UserAuthenticationPolicies")
-	sessionPoliciesProviderFactory                   = providerFactoryUsingCache("SessionPolicies")
+	policiesProviderFactory                          = providerFactoryUsingCache("Policies")
```
sfc-gh-jmichalak: Nit: can we just have "warehouse required factory" or something? or have a TODO to do it later.

pkg/resources/table_storage_lifecycle_policy_attachment.go
```diff
@@ -0,0 +1,273 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+var tableStorageLifecyclePolicyAttachmentSchema = map[string]*schema.Schema{
+	"table_name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		Description:      blocklistedCharactersFieldDescription("Fully qualified name of the table (or dynamic table) the storage lifecycle policy is attached to."),
```
sfc-gh-jmichalak: Please remove blocklistedCharactersFieldDescription. In the examples, we even use these characters - we must allow them to reference objects outside TF. same below.

pkg/resources/table_storage_lifecycle_policy_attachment.go
```diff
@@ -0,0 +1,273 @@
...
+		DiffSuppressFunc: suppressIdentifierQuoting,
+		ValidateDiagFunc: IsValidIdentifier[sdk.SchemaObjectIdentifier](),
+	},
+	"table_type": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		Description:      "Specifies the type of the table referenced in `table_name`. " + enumValuesDescription(sdk.StorageLifecyclePolicySupportedTableTypes),
+		DiffSuppressFunc: NormalizeAndCompare(sdk.ToPolicyEntityDomain),
+		ValidateFunc: validation.StringInSlice(collections.Map(sdk.StorageLifecyclePolicySupportedTableTypes, func(v sdk.PolicyEntityDomain) string {
```
sfc-gh-jmichalak, line 34: Nit: why not a dedicated `To...` helper like with other enums?

pkg/resources/table_storage_lifecycle_policy_attachment.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	columns := expandStorageLifecyclePolicyColumns(d.Get("on").([]any))
+
+	if err := addStorageLifecyclePolicyToTable(ctx, client, tableType, tableName, storageLifecyclePolicyName, columns); err != nil {
+		return diag.FromErr(fmt.Errorf("error while attaching storage lifecycle policy to %s %s, err = %w", tableType, tableName.FullyQualifiedName(), err))
+	}
+
+	d.SetId(helpers.EncodeResourceIdentifier(tableName.FullyQualifiedName(), storageLifecyclePolicyName.FullyQualifiedName()))
```
sfc-gh-jmichalak, line 104: Q: should we also have the type in ID? I mean, is it possible, to have a table and a dynamic table with the same name? What about columns? Can we have two policies assigned for different columns?

pkg/resources/table_storage_lifecycle_policy_attachment.go
```diff
@@ -0,0 +1,273 @@
...
+
+	storageLifecyclePolicyReferences := make([]sdk.PolicyReference, 0)
+	for _, policyReference := range policyReferences {
+		if policyReference.PolicyKind == sdk.PolicyKindStorageLifecyclePolicy {
+			storageLifecyclePolicyReferences = append(storageLifecyclePolicyReferences, policyReference)
+		}
+	}
+
+	if len(storageLifecyclePolicyReferences) > 1 {
+		return diag.FromErr(fmt.Errorf("internal error: multiple storage lifecycle policy references attached to a table. This should never happen"))
```
sfc-gh-jmichalak, line 142: I guess this answer my question above, but we should have integration tests confirming this.

pkg/resources/table_storage_lifecycle_policy_attachment.go
```diff
@@ -0,0 +1,273 @@
...
+	if storageLifecyclePolicyReference.RefArgColumnNames != nil {
+		on = sdk.ParseCommaSeparatedStringArray(*storageLifecyclePolicyReference.RefArgColumnNames, true)
+	}
+
+	errs := errors.Join(
+		d.Set("table_type", storageLifecyclePolicyReference.RefEntityDomain),
+		d.Set(
+			"storage_lifecycle_policy_name",
+			sdk.NewSchemaObjectIdentifier(
+				*storageLifecyclePolicyReference.PolicyDb,
```
sfc-gh-jmichalak: We may handle it like this in other resources, but please dereference safely, as this could panic.

pkg/resources/table_storage_lifecycle_policy_attachment.go
```diff
@@ -0,0 +1,273 @@
...
+			).FullyQualifiedName()),
+		d.Set("on", on),
+	)
+	return diag.FromErr(errs)
+}
+
+func UpdateTableStorageLifecyclePolicyAttachment(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*provider.Context).Client
+
+	if d.HasChange("storage_lifecycle_policy_name") || d.HasChange("on") {
```
sfc-gh-jmichalak, line 187: Q: why not ForceNew on these fields then? It's still ALTER SET + ALTER UNSET.

pkg/testacc/resource_table_storage_lifecycle_policy_attachment_acceptance_test.go
```diff
@@ -0,0 +1,242 @@
...
+			{
+				PreConfig: func() {
+					testClient().Table.DropFunc(t, table2.ID())()
+				},
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+				Config: " ",
```
sfc-gh-jmichalak: Nit: I think we usually keep the config and add `Destroy: true`.

sfc-gh-jmichalak (changes requested):

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: > is it possible, to have a table and a dynamic table with the same name?

No, it's not.

> Can we have two policies assigned for different columns?

Only one policy can be assigned at the same time. Note that SLPs (storage lifecycle policies) are assigned to a table, not to columns.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: Will do, though it's unlikely to happen, since SLPs are session-level objects.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: Because we don't want to allow all enum values, we want to allow only two, so I can't reuse `ToPolicyEntityDomain`. I could implement a new `To...` specifically for these two allowed values, but the current approach seems as good to me.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: I'm not sure about `"`, but `|` is a resource ID delimiter, and `.` is an identifier delimiter, so these two characters cannot be used in this field for sure, which is why the warning seems to be rather correct.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: I initially marked all fields as ForceNew while working on an attachment resource for session policies, but then a review comment popped up that suggested something totally different: https://github.com/snowflakedb/terraform-provider-snowflake/pull/4666#discussion_r3109719686, and I applied the suggestion. Here I just replicated it.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: It's explicitly stated in documentation: https://docs.snowflake.com/en/user-guide/storage-management/storage-lifecycle-policies-create-manage#attach-a-policy-to-a-table, which is why there's no such test. Unless you think this can change in the future?

examples/resources/snowflake_table_storage_lifecycle_policy_attachment/resource.tf
sfc-gh-kwasilewski: Adjusted

pkg/testacc/1_setup_provider_test.go
sfc-gh-kwasilewski: Renamed to `warehouseRequiredProviderFactory` and added a comment documenting when it should be used

pkg/testacc/resource_table_storage_lifecycle_policy_attachment_acceptance_test.go
sfc-gh-kwasilewski: Changed

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-jmichalak: This is fine, then.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-jmichalak: Right, so maybe we should document that we disallow only `|`. We're still allowing the `.` and `"` because we're expecting the fully qualified name.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-jmichalak: I think we have the conversion approach in services or compute pools.  The reason is in future, it would be easier to e.g. generate the schemas automatically.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-jmichalak: I see, so why are table_name and table_type ForceNew? They could also be handled in Update.

sfc-gh-jmichalak (dismissed):

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: No particular reason. It's doable, but there's a risk of overcomplicating the update flow. My opinion is that while it makes sense for `storage_lifecycle_policy_name` to be non-ForceNew because the hypothetical `storage_lifecycle_policy` field on a table resource would also be non-ForceNew, the `table_name` is something different.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: Let me keep it that way, we can always change it later on.

pkg/resources/table_storage_lifecycle_policy_attachment.go
sfc-gh-kwasilewski: Done

sfc-gh-jmichalak (approved):

Issue #4730, opened 2026-05-12
Issue when adding boolean column to a table with default in one step


ayu-ss:
### Terraform CLI Version

1.13.0

### Terraform Provider Version

2.16.0

### Terraform Configuration

```terraform
resource "snowflake_table" "example" {
  database = "EXAMPLE_DB"
  schema   = "EXAMPLE_SCHEMA"
  name     = "EXAMPLE_TABLE"

  column {
    name     = "NEW_BOOLEAN_COLUMN"
    type     = "boolean"
    nullable = true

    default {
      constant = "FALSE"
    }
  }
}
```

### Category

category:snowflake

### Object type(s)

resource:table

### Expected Behavior

New boolean column added with a default value of false.

### Actual Behavior

When attempting to add a new boolean column with a default value to an existing Snowflake table in a single terraform apply, the Snowflake Terraform provider crashes with a nil-pointer dereference.

However, splitting the change into two steps — first adding the column without a default, then adding the default in a second run — works correctly.

This suggests the crash is due to unsafe handling of the default block for boolean types.

### Steps to Reproduce

1. Define a snowflake_table resource with an existing table.
2. Add a new column block with the following configuration:
```
Copy
column {
  name     = "NEW_BOOLEAN_COLUMN"
  type     = "boolean"
  nullable = true

  default {
    constant = "false"
  }
}
```
3. Run terraform apply.
4. Observe the crash with the following panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1bfe15b]

goroutine 115 [running]:
github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk.(*TableColumnActionRequest).toOpts(0xc000e2f1f0)
	github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk/tables_impl.go:380 +0x55b
```

### How much impact is this issue causing?

Medium

### Logs

https://gist.github.com/ayu-ss/db821cbbed2ea0627dfcc0f06b0e3f9a

### Additional Information

Splitting the change into two steps — first adding the column without a default, then adding the default in a second run — works correctly.

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jcieslak:
Hey @ayu-ss 👋
Thanks for reaching out to us.
[...]

PR #4914, opened 2026-06-27
fix: prevent nil-pointer panic when adding a column with a constant default

simond:
Adding a new column with a constant (or expression) default to an **existing**
table panics with a nil-pointer dereference. `TableColumnActionRequest.toOpts`
unconditionally dereferenced `DefaultValue.identity`, which is nil for any
non-identity default:

```
panic: runtime error: invalid memory address or nil pointer dereference
    pkg/sdk/tables_impl.go (*TableColumnActionRequest).toOpts
```

The add-column path now nil-guards the identity before building `ColumnIdentity`,
mirroring the existing `convertColumns` (create-table) logic. This also restores
the `Order`/`Noorder` fields, which the add path previously dropped.

Fresh table creation was unaffected because that path already guarded the nil,
which is why the two-step workaround (add the column, then set the default)
worked around the issue.

## Test Plan
* [x] unit tests - two new cases in `TestTableAlter`: adding a column with a
  constant default (the repro; panics without the fix) and with an identity
  default (regression guard)
* [x] acceptance tests
* [x] manually verified end-to-end against a live Snowflake account: the
  `ALTER TABLE ... ADD COLUMN <bool> DEFAULT FALSE` on an existing table now
  succeeds instead of crashing

## References
- #4730


sfc-gh-jmichalak:
Hi @simond 👋 

Thanks for your contribution! We'll discuss it internally first. Stay tuned!

sfc-gh-jmichalak (changes requested):
The implementation LGTM. Please also add

- Integration test in `pkg/sdk/testint/tables_integration_test.go` Add a test case that runs this query and verifies no error happens.
- Acceptance test in `pkg/testacc/resource_table_acceptance_test.go` In 1st step, create a simple table, and in the 2nd step add a new column with e.g. DEFAULT FALSE
- A short migration guide entry in `MIGRATION_GUIDE.MD`.

simond:
> The implementation LGTM. Please also add
> 
> * Integration test in `pkg/sdk/testint/tables_integration_test.go` Add a test case that runs this query and verifies no error happens.
> * Acceptance test in `pkg/testacc/resource_table_acceptance_test.go` In 1st step, create a simple table, and in the 2nd step add a new column with e.g. DEFAULT FALSE
> * A short migration guide entry in `MIGRATION_GUIDE.MD`.

Thanks for the review! I've pushed the three additions:

- Integration test in pkg/sdk/testint/tables_integration_test.go covering ALTER TABLE ... ADD COLUMN with a constant default.
- Acceptance test TestAcc_Table_AddColumnWithConstantDefault_issue4730 in pkg/testacc/resource_table_acceptance_test.go: step 1 creates a simple table, step 2 adds a BOOLEAN column with DEFAULT FALSE.
- A short *(bugfix)* entry in MIGRATION_GUIDE.MD.

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

Issue #4773, opened 2026-05-23
snowflake_oauth_integration_for_custom_clients should include allowed_role_list


npatel44:
### Use Cases or Problem Statement

snowflake_oauth_integration_for_custom_clients resource as well as snowflake official document doesnt mention OAUTH_ALLOWED_ROLES_LIST . however i was able to set that parameter via sql withotu any issue and was able to test that it get enforced too. If that is the case provider should support that variable as well.

  ALTER SECURITY INTEGRATION RUNLAYER_SNOWFLAKE_MCP_OAUTH
    SET ALLOWED_ROLES_LIST = ('DATA_TEAM'); <<this works
RUNLAYER_SNOWFLAKE_MCP_OAUTH is snowflake OATH and not external OAUTH

### Category

category:resource

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-kwasilewski:
Hi @npatel44, thanks for reaching out to us.

ALLOWED_ROLES_LIST is not in the official documentation for Snowflake OAuth integrations, so it is not included in the Snowflake Terraform provider schema for this integration type. Terraform support would require the feature to be formally documented first. This is because while the parameter may appear to work right now, it may change without notice in future Snowflake releases.

In the meantime, you can use the blocked_roles_list field as the documented and officially supported approach for restricting roles.

sfc-gh-kwasilewski:
Hi @npatel44, the new ALLOWED_ROLES_LIST property has been added to the official Snowflake documentation, which means we can add it to the Snowflake Terraform provider schema. I'll drop an update here once that's complete.

sfc-gh-kwasilewski:
Hi @npatel44,

allowed_roles_list has been added to `snowflake_oauth_integration_for_custom_clients` and `snowflake_oauth_integration_for_partner_applications` resources in [v2.19.0](https://github.com/snowflakedb/terraform-provider-snowflake/releases/tag/v2.19.0).

Closing this as completed. Feel free to reopen if you run into any issues.

PR #4937, opened 2026-07-02
feat: Add allowed_roles_list to oauth_integration_for_custom_clients

sfc-gh-kwasilewski:
Surfaces the ALLOWED_ROLES_LIST Snowflake property in the resource schema. Includes CustomizeDiff validation that the field requires oauth_use_secondary_roles = NONE, Read/Create/Update wiring, schema/describe changes, example and doc updates, and acceptance test coverage.

Adding allowed_roles_list to
snowflake_oauth_integration_for_partner_applications will be done in a follow-up.

References #4773


pkg/resources/oauth_integration_for_custom_clients.go
```diff
@@ -82,6 +82,15 @@ var oauthIntegrationForCustomClientsSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Description: relatedResourceDescription("A set of Snowflake roles that a user does not need to explicitly consent to using after authenticating.", resources.AccountRole),
 	},
+	"allowed_roles_list": {
```
sfc-gh-asawicki, line 85: question: in external_oauth_integration allowed_roles_list conflicts with blocked_roles_list - as I understand, it's not a problem here? at least based on the tests added

sfc-gh-asawicki, line 85: second question: what happens after the migration to this version when:
- the integration had some roles set up in Snowflake
- the integration had empty roles in Snowflake?

Asking because in the first case, the plan will not be empty, so customer should add it to their config then, right?

sfc-gh-asawicki (changes requested):

pkg/resources/oauth_integration_for_custom_clients.go
sfc-gh-kwasilewski: No, both allowed and blocked can be set at the same time. The only caveat is that if a specific role is in ALLOWED_ROLES_LIST, it cannot be added to BLOCKED_ROLES_LIST.

sfc-gh-kwasilewski: 1. We detect a drift, and the customer should adjust their config. Should we mention that in the migration guide? Right now, it says "No action is required", but that may not be true.
2. The plan should be empty. I think it's already covered by the BasicUseCase test, where we create an integration with empty roles, and then we import it.

sfc-gh-asawicki: 2 is OK
ad 1 We should mention it, and also have a test confirming it will really happen (we had a situation where our recommendation was not aligned with what really happens). There is also an edge case, when the customer imports it (using import block which behaves a bit differently than terraform import command - sharing a short doc internally). For that reason, I'm not sure if we maybe shouldnt hide this behavior behind experiment.

Let's add a test and update migration guide entry and discuss on Monday if we need to add an experiment, ok?

sfc-gh-kwasilewski: Added test and adjusted migration guide

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

PR #4945, opened 2026-07-03
chore: Remove comment interface from SDK

sfc-gh-asawicki:
- It's almost unused
- We don't have plans to have a dedicated comment resource
- Adjusted the only caller (dynamic tables)


pkg/sdk/dynamic_table_dto.go
```diff
@@ -30,6 +30,7 @@ type AlterDynamicTableRequest struct {
 	resume                     *bool
 	refresh                    *bool
 	set                        *DynamicTableSetRequest
+	comment                    *string
```
sfc-gh-kwasilewski, line 33: Any reason why it's not a part of DynamicTableSetRequest?

sfc-gh-asawicki: It's listed separately in https://docs.snowflake.com/en/sql-reference/sql/alter-dynamic-table#syntax (and it was previously implemented like this on the resource side). I will be moving dynamic tables to the SDK soon, and I can test it then.

sfc-gh-kwasilewski (approved):

PR #4947, opened 2026-07-03
feat: Add allowed_roles_list to oauth_integration_for_partner_apps

sfc-gh-kwasilewski:
Surface the ALLOWED_ROLES_LIST property on the
snowflake_oauth_integration_for_partner_applications resource, mirroring the custom clients implementation. The attribute can only be set when oauth_use_secondary_roles is NONE, enforced via CustomizeDiff.


pkg/resources/oauth_integration_for_custom_clients.go
```diff
@@ -183,7 +183,7 @@ func OauthIntegrationForCustomClients() *schema.Resource {
 		Description:   "Resource used to manage oauth security integration for custom clients objects. For more information, check [security integrations documentation](https://docs.snowflake.com/en/sql-reference/sql/create-security-integration-oauth-snowflake).",
 
 		CustomizeDiff: TrackingCustomDiffWrapper(resources.OauthIntegrationForCustomClients, customdiff.All(
-			oauthCustomClientsAllowedRolesListRequiresSecondaryRolesNone,
+			oauthAllowedRolesListRequiresSecondaryRolesNone,
```
sfc-gh-jmichalak: This should be moved and validated in ValidateRawResourceConfigFuncs.

sfc-gh-jmichalak (changes requested):

pkg/resources/oauth_integration_for_custom_clients.go
sfc-gh-asawicki: could be in follow-up

sfc-gh-asawicki (dismissed):

pkg/resources/oauth_integration_for_custom_clients.go
sfc-gh-kwasilewski: Done

sfc-gh-jmichalak (approved):

PR #4925, opened 2026-06-30
chore: Simplify resource assertions for other resource types part4 - resourceshowoutputassert package

sfc-gh-asawicki:
This PR completes Phase 3 of the resource assertion simplification: regenerating all objects in the `resourceshowoutputassert` package using the updated code generator templates.

### What changed

**All show output objects regenerated** — constructors updated from the old prefix-string pattern to `NewResourceShowOutputAssert` / `NewImportedResourceShowOutputAssert` / `NewDatasourceShowOutputAssert`.

**All describe output objects regenerated** — constructors updated to `NewResourceDescribeOutputAssert` / `NewImportedResourceDescribeOutputAssert` / `NewDatasourceDescribeOutputAssert`.

**All `_ext.go` files migrated** from old-style `AddAssertion(assert.ResourceShowOutputValueSet(...))` / `AddAssertion(assert.ResourceDescribeOutputValueSet(...))` to direct `StringValueSet(...)` / `BoolValueSet(...)` / `IntValueSet(...)` / `ValuePresent(...)` / `ValueNotSet(...)` / `SetContainsElem(...)` calls.

### Generator fixes

| File | Change |
|------|--------|
| `gen/model.go` | Added `IsSdkStruct bool` field — skips `Has*` generation for SDK struct/interface typed fields |
| `gen/templates/assertions.tmpl` | Added `(not .IsSdkStruct)` gate for Has* method generation |
| `gen/templates/definition.tmpl` | Datasource constructor uses `{{ .DataSourceName }}` for function naming |
| `gen/sdk_object_def.go` | Added `sdk.Role{}`, `sdk.SessionPolicy{}`, `sdk.ApiIntegrationAllDetails{}` to mappings; added `objectsNotBeingResources` entries |
| `objectassert/gen/sdk_object_def.go` | Added `sdk.ApiIntegrationAllDetails{}`, `sdk.SessionPolicy{}` to object list |

### Sub-object architecture

Sub-struct types do NOT have standalone assert types. Their assertions are methods on parent types:

| Sub-struct | Parent types | Approach |
|-----------|-------------|----------|
| `IcebergRestRestConfigDetails` | `CatalogIntegrationIcebergRestDescribeOutputAssert` | `objectsNotBeingResources`; methods in parent ext |
| `OpenCatalogRestConfigDetails` | `CatalogIntegrationOpenCatalogDescribeOutputAssert` | `objectsNotBeingResources`; methods in parent ext |
[...]

PR #4822, opened 2026-06-09
feat: create api integration amazon api gateway resource

sfc-gh-jcieslak:
## Changes
- Create amazon api gateway resource and common schema/operations for next api providers

## Next prs
- Register resource as production resource (in provider.go) and as preview resource
- Add examples, documentation, and migration guide entry
- Add a deprecation notice to the old API integration resource / data source

PR #4964, opened 2026-07-08
feat: add issuer to AWS default_workload_identity block

### LINKED ISSUE #4963: Support ISSUER in default_workload_identity AWS block for JWT-based WIF
### Company Name

Uzabase

The `snowflake_service_user` resource supports AWS Workload Identity Federation
(WIF) via the `default_workload_identity` → `aws` nested block, but that block
currently exposes only `arn`. There is no way to set `ISSUER`.

Docs for the current resource:
https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/service_user#nestedblock--default_workload_identity--aws

Snowflake's AWS WIF supports two attestation methods:
1. `GetCallerIdentity` (the current default) — ARN-only, works today.
2. `GetWebIdentityToken` / JWT-based — requires **both** `ARN` and `ISSUER` on
   the user. Snowflake recommends this method as more future-proof and
   standards-aligned, and its docs note it "may become the default in a future
   release."

Per Snowflake's docs ("Upgrade to JWT-based authentication (recommended)"), the
JWT method requires both `ARN` and `ISSUER`:
https://docs.snowflake.com/en/user-guide/workload-identity-federation

    ALTER USER <username> SET WORKLOAD_IDENTITY = (
      TYPE = AWS
      ARN = '<role_arn>'
      ISSUER = '<aws_issuer_url>'   -- not settable via Terraform today
    );

Because the provider can't set `ISSUER` on the AWS block, anyone adopting
JWT-based AWS WIF has to configure it manually (e.g. via `ALTER USER`) outside
of Terraform, which causes drift. For comparison, the `azure` block already
exposes `issuer` (and `subject`), so the AWS block missing `issuer` looks like
an oversight.

### Category

category:resource

### Object type(s)

resource:service_user

### Proposal

Add an `issuer` attribute to the `default_workload_identity.aws` nested block on
`snowflake_service_user` (and any other resource exposing an AWS WIF block),
mapping to the `ISSUER` parameter of `WORKLOAD_IDENTITY`. This mirrors the
existing `issuer` attribute already present on the `azure` block.

### Additional Information

The official drivers are converging on JWT-based AWS WIF as an opt-in expected
to become the default:
- Node.js (snowflake-sdk): opt-in via the `workloadIdentityAwsUseOutboundToken`
  connection option (default `false`).
- Python: opt-in via the `SNOWFLAKE_ENABLE_AWS_WIF_OUTBOUND_TOKEN=true`
  connector environment variable.

Snowflake's docs warn that `ISSUER` must be configured on the user *before*
enabling the driver-side JWT option, otherwise authentication fails immediately.

We first hit this with snowflake-sdk (Node.js) 3.0.0, where the JWT
(`GetWebIdentityToken`) path was briefly made the default and the SigV4
`GetCallerIdentity` path removed
(https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1415). That was
later reverted as an unintentional breaking change and the JWT method is now
opt-in (https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1437).
Even so, `ISSUER` needs to be settable via Terraform to adopt the recommended
JWT method without manual `ALTER USER` drift.

Related issues: #3942, #3998

References:
- Snowflake WIF docs / "Upgrade to JWT-based authentication (recommended)":
  https://docs.snowflake.com/en/user-guide/workload-identity-federation
- Current resource docs (only `arn` in the AWS block):
  https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/service_user#nestedblock--default_workload_identity--aws

- [x] Yeah, I'll take it 😎

edvakf:
Adds an optional `issuer` attribute to the `default_workload_identity.aws`
nested block on `snowflake_service_user` and `snowflake_legacy_service_user`,
mirroring the existing `azure.issuer` attribute.

It maps to the `ISSUER` parameter of `WORKLOAD_IDENTITY`:

```sql
ALTER USER <username> SET WORKLOAD_IDENTITY = (
  TYPE = AWS
  ARN = '<role_arn>'
  ISSUER = '<aws_issuer_url>'
);
```

This enables JWT-based (`GetWebIdentityToken`) AWS workload identity federation,
which Snowflake recommends and requires both `ARN` and `ISSUER`, in addition to
the existing `GetCallerIdentity` (arn-only) method. Previously `ISSUER` could
only be set via a manual `ALTER USER`, causing drift.

### Design notes (to pre-empt review questions)

- **Optional / additive / non-breaking.** `arn` stays `Required`; existing
  arn-only (`GetCallerIdentity`) configurations are unchanged.
- **`issuer` *is* returned by Snowflake** in `SHOW USER WORKLOAD IDENTITY
  AUTHENTICATION METHODS` (unlike `arn`, which is only returned split into
  `iamRole`/`awsAccount`/`awsPartition`, tracked by `TODO(SNOW-3003261)`). This
  is now captured in `UserWorkloadIdentityAuthenticationMethodsAwsAdditionalInfo.Issuer`
  and verified in both the SDK integration test and the acceptance test's
  `objectassert` checks. The Terraform resource's `Read` still does not flatten
  the AWS block into state (the existing `Type != WIFTypeAws` guard in
  `user.go`), so from Terraform's perspective `arn`/`issuer` remain write-only
  for now and there is **no perpetual diff**. Wiring up drift detection for
  `issuer` specifically (since Snowflake does return it faithfully) is possible
  as a follow-up but out of scope here.
- **No new validation test** because the field is `Optional` (no required /
  `ExactlyOneOf` rule applies). The arn-only → arn+issuer transition is covered
  by `TestAcc_*_WIF_AWS`; the required/`ExactlyOneOf` matrix stays covered by
  `TestAcc_*_WIF_Validations`.
- Applied consistently to both `service_user` and `legacy_service_user` (shared
  `defaultWorkloadIdentitySchema`).
- SDK change is made in the generator def
  (`userObjectWorkloadIdentityAwsStruct`) and the `*_gen.go` files were
  regenerated (`make generate-sdk-no-tests`); docs were regenerated via
  `make docs`.

## Test Plan
* [ ] acceptance tests (`TestAcc_ServiceUser_WIF_AWS`,
  `TestAcc_LegacyServiceUser_WIF_AWS` — extended with an
  "UPDATE - SWITCH TO JWT-BASED" step, now asserting `issuer` via
  `HasAwsAdditionalInfo`; require a live Snowflake account)
* [x] SDK unit tests (`go test ./pkg/sdk/ -run TestUser`) — new
  create/alter cases assert
  `WORKLOAD_IDENTITY = (TYPE = AWS ARN = '...' ISSUER = '...')`
* [x] SDK integration test (`TestInt_Users`) — `awsWifAssertion` now asserts
  `issuer` for both the arn-only and arn+issuer AWS cases
* [x] `make pre-push-check` clean (generator + docs produce no diff)
* [x] `go build ./...`, `make test-unit`

## References

* Closes #4963
* Snowflake WIF docs / "Upgrade to JWT-based authentication (recommended)":
  https://docs.snowflake.com/en/user-guide/workload-identity-federation
* Original WIF support: #4367

---

### Update: addressed review feedback

Thanks @sfc-gh-jmichalak and @sfc-gh-jcieslak for the review!

- Added `Issuer` to `UserWorkloadIdentityAuthenticationMethodsAwsAdditionalInfo`
  and asserted its value in `awsWifAssertion` (integration test) and in the
  acceptance tests' `objectassert` checks, per the request to verify `issuer`
  is actually returned by Snowflake.
- This also corrects the design note above: `issuer` is in fact returned by
  Snowflake for AWS (only `arn` has the SHOW-output round-trip gap), so I've
  updated that claim accordingly.
- Note: I ran `make generate-snowflake-object-assertions` as suggested, but
  didn't include its output — it isn't safe to run blindly here, for two
  pre-existing, unrelated reasons:
  1. The generator's registry
     (`objectassert/gen/sdk_object_def.go`) still declares
     `UserWorkloadIdentityAuthenticationMethod` with a plain
     `sdk.AccountObjectIdentifier`, but the checked-in assert file was
     hand-patched (`// adjusted manually`, 3 spots) to use the composite
     `UserWorkloadIdentityAuthenticationMethodsObjectIdentifier`
     (user + method name) instead, because a bare `AccountObjectIdentifier`
     can't uniquely identify a WIF method scoped to a user (tracked by
     `TODO(SNOW-1501905)`). The generator doesn't know about that override,
     so a blind run reverts it (renaming `...Methods...` → `...Method...`
     and dropping the composite identifier).
  2. It also emitted ~50 files for object types unrelated to this PR
     (e.g. `SessionPolicy`, `ApiIntegrationAwsDetails`) that are already in
     the registry (added by #4925, #4822) but were apparently never
     generated/committed — a pre-existing generation backlog, not something
     introduced here.

  Neither of these is related to the `issuer` change: `HasAwsAdditionalInfo`
  already compares the whole `AwsAdditionalInfo` struct by value, so adding
  the `Issuer` field is picked up automatically without regenerating anything.
- Left the vararg `issuer ...string` nit in `SetAwsWorkloadIdentity` as-is per
  @sfc-gh-jcieslak's comment that it's non-blocking and can be adjusted on
  your side.
- Have not yet implemented drift detection (Read/Update wiring) for `issuer`
  — happy to do this in this PR or a follow-up, whichever you'd prefer.


pkg/sdk/testint/users_integration_test.go
```diff
@@ -621,6 +626,24 @@ func TestInt_Users(t *testing.T) {
...
+					accountNumber: accountNumber,
+					arn:           fmt.Sprintf("arn:aws:iam::%s:role/test-role", accountNumber),
+					issuer:        "https://sts.amazonaws.com",
+				}
+			},
+			wifConfig: func(data any) *sdk.UserObjectWorkloadIdentityPropertiesRequest {
+				return awsWifConfig(data.(awsWifData).arn, data.(awsWifData).issuer)
+			},
+			wifAssertion: func(t *testing.T, assertion *objectassert.UserWorkloadIdentityAuthenticationMethodsAssert, data any) {
+				awsWifAssertion(t, assertion, data.(awsWifData).accountNumber)
```
sfc-gh-jmichalak: The issuer is returned from Snowflake:

- Add the issuer field to UserWorkloadIdentityAuthenticationMethodsAwsAdditionalInfo.
- Assert its value in awsWifAssertion.

pkg/resources/user_commons.go
```diff
@@ -34,6 +34,11 @@ var defaultWorkloadIdentitySchema = map[string]*schema.Schema{
 								Required:    true,
 								Description: "The ARN of the AWS IAM role to use for workload identity federation.",
 							},
+							"issuer": {
```
sfc-gh-jmichalak, line 37: General remark: external changes for the `arn` field are not detected due to Snowflake output limitations (as noted in the Registry docs), but the `issuer` is returned from `SHOW USER ...` so we could detect external changes. This is optional; we can merge without it.

pkg/testacc/resource_service_user_wif_acceptance_test.go
```diff
@@ -211,6 +217,24 @@ func TestAcc_ServiceUser_WIF_AWS(t *testing.T) {
...
+						HasNameString(id.Name()).
+						HasDefaultWorkloadIdentityAws(arn, issuer),
+					resourceshowoutputassert.UserShowOutput(t, userModelWithAWSAndIssuer.ResourceReference()).
+						HasHasWorkloadIdentity(true),
+					objectassert.UserDefaultWorkloadIdentityAuthenticationMethods(t, id).
+						HasName("DEFAULT").
+						HasType(sdk.WIFTypeAws).
+						HasNoComment().
+						HasLastUsedNotEmpty().
+						HasCreatedOnNotEmpty(),
```
sfc-gh-jmichalak: After this field is added to the helper struct as I wrote above, the assertions should be regenerated with `make generate-snowflake-object-assertions` (but only for this struct). Then, we can add the missing value assertion here.

sfc-gh-jmichalak (changes requested):
Hi, thanks for your contribution! The change looks good overall, but to get this merged, please just add a simple assertion in the integration test. Thanks!

pkg/resources/user_commons.go
sfc-gh-jcieslak: To make it detect and handle external changes, you have to modify Read and Update functions, where in Read value from Snowflake would be set to issuer field and in Update you have to modify ALTER request in SET (as I saw in docs to unset ISSUER the SET without it has to be issued). As mentioned above, it's optional, but changes in Snowflake won't be detected without it.

pkg/acceptance/helpers/user_client.go
```diff
@@ -276,16 +276,21 @@ func (c *UserClient) SetAzureWorkloadIdentity(t *testing.T, userId sdk.AccountOb
 }
 
 // SetAwsWorkloadIdentity sets the AWS workload identity configuration for a user.
-func (c *UserClient) SetAwsWorkloadIdentity(t *testing.T, userId sdk.AccountObjectIdentifier, arn string) {
+// issuer is optional (used for JWT-based / GetWebIdentityToken workload identity federation).
+func (c *UserClient) SetAwsWorkloadIdentity(t *testing.T, userId sdk.AccountObjectIdentifier, arn string, issuer ...string) {
```
sfc-gh-jcieslak, line 280: nit: there can be only one issuer. I know vararg is used here to make the field optional, but we could have just another function that supports the issuer. Non-blocker, we can adjust this later on our side.

sfc-gh-jcieslak (commented):
I don't have any additional blocking comments other than the ones left by @sfc-gh-jmichalak. Thanks for the high quality contribution and detailed issue description 🙏

pkg/acceptance/helpers/user_client.go
sfc-gh-jmichalak: This method is not used because the external changes are not detected. I left this for now.

pkg/resources/user_commons.go
sfc-gh-jmichalak: Left for now, handling the changes on the entire block will be done in SNOW-3003261.

sfc-gh-jmichalak (approved):

sfc-gh-jcieslak (approved):

edvakf:
Thanks @sfc-gh-jmichalak and @sfc-gh-jcieslak for the review and for merging this! Appreciate the guidance on scoping the issuer/arn drift detection to SNOW-3003261 and leaving the vararg helper as-is for now.

PR #4970, opened 2026-07-08
chore: Fix task ownership cleanup in "on all tasks - with operate" test

sfc-gh-kwasilewski:
The cleanup transferred task ownership back to the current role while the session was switched to the owning role, which could fail with:

  Task ownership can only be transferred to a role that is activated in
  the current session or by a role with account-level MANAGE GRANTS.

Run the hand-back as ACCOUNTADMIN instead of switching to the owning role. The destination is then the active role, so the cleanup no longer hits this error.

### DIFF
--- a/pkg/sdk/testint/grants_integration_test.go
+++ b/pkg/sdk/testint/grants_integration_test.go
@@ -2180,10 +2180,8 @@ func TestInt_GrantOwnership(t *testing.T) {
 
 		t.Cleanup(func() {
 			currentRole := testClientHelper().Context.CurrentRole(t)
-			usePreviousRole := testClientHelper().Role.UseRole(t, role.ID())
 			grantOwnershipToRole(t, currentRole, ownershipGrantOnTask(task), sdk.Pointer(sdk.Revoke))
 			grantOwnershipToRole(t, currentRole, ownershipGrantOnTask(secondTask), sdk.Pointer(sdk.Revoke))
-			usePreviousRole()
 		})
 
 		usePreviousRole = testClientHelper().Role.UseRole(t, taskRole.ID())
--- a/pkg/sdk/testint/grants_integration_test.go
+++ b/pkg/sdk/testint/grants_integration_test.go
@@ -2108,6 +2108,8 @@ func TestInt_GrantOwnership(t *testing.T) {
 	})
 
 	t.Run("on all tasks - with operate", func(t *testing.T) {
+		currentRole := testClientHelper().Context.CurrentRole(t)
+
 		taskRole, taskRoleCleanup := testClientHelper().Role.CreateRoleGrantedToCurrentUser(t)
 		t.Cleanup(taskRoleCleanup)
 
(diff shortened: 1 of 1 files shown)


pkg/sdk/testint/grants_integration_test.go
sfc-gh-asawicki, line 2127: The problem with this test is that there is no role cleanup, so in case of any error, the role might not be reverted to the one used in our test connection. There are more uses of `UseRole` in this or other tests in this file (maybe even beyond). Let's make sure that each of them has t.Cleanup.

pkg/sdk/testint/grants_integration_test.go
```diff
@@ -2180,10 +2180,8 @@ func TestInt_GrantOwnership(t *testing.T) {
 
 		t.Cleanup(func() {
 			currentRole := testClientHelper().Context.CurrentRole(t)
-			usePreviousRole := testClientHelper().Role.UseRole(t, role.ID())
```
sfc-gh-asawicki, line 2183: The role is changed a few times during this test. So, we will not always have an ACCOUNTADMIN role here, given the current setup.

Let's iterate on it, and when we have the agreement for the solution, let's verify if there are no more tests in this file needing similar adjustment.

sfc-gh-asawicki (changes requested):

pkg/sdk/testint/grants_integration_test.go
sfc-gh-kwasilewski: I registered the cleanups via t.Cleanup on the caller side - this is what other integration tests that utilize UseRole do, and I wanted to keep consistency.

pkg/sdk/testint/grants_integration_test.go
sfc-gh-kwasilewski: Fixed, every UseRole call in this test is made from an ACCOUNTADMIN baseline, so each cleanup reverts the role to ACCOUNTADMIN.

pkg/sdk/testint/grants_integration_test.go
```diff
@@ -2119,14 +2121,11 @@ func TestInt_GrantOwnership(t *testing.T) {
 
 		// grantTaskRole grants the necessary privileges to a role to be able to create task
 		grantTaskRole(t, taskRole.ID())
-
-		currentRole := testClientHelper().Context.CurrentRole(t)
-
 		grantTaskRole(t, role.ID())
-		grantTaskRole(t, currentRole)
```
sfc-gh-kwasilewski, line 2126: Not needed - ACCOUNTADMIN already has EXECUTE TASK and USAGE on warehouse privileges.

pkg/sdk/testint/grants_integration_test.go
```diff
@@ -2220,14 +2217,14 @@ func TestInt_GrantOwnership(t *testing.T) {
...
 		usePreviousRole = testClientHelper().Role.UseRole(t, role.ID())
+		t.Cleanup(usePreviousRole)
 		currentTask, err = client.Tasks.ShowByID(ctx, task.ID())
 		require.NoError(t, err)
 		require.Equal(t, sdk.TaskStateSuspended, currentTask.State)
 
 		currentSecondTask, err = client.Tasks.ShowByID(ctx, secondTask.ID())
 		require.NoError(t, err)
 		require.Equal(t, sdk.TaskStateSuspended, currentSecondTask.State)
-		usePreviousRole()
```
sfc-gh-kwasilewski, line 2230: Not needed now that we registered cleanups

sfc-gh-asawicki (approved):

PR #4981, opened 2026-07-09
chore: Adjust docker compose envs

sfc-gh-jcieslak:
For the preprod tests to correctly pick up the cloud resources envs, they have to be explicitly passed through in Docker Compose


sfc-gh-jcieslak:
PR moved

PR #4993, opened 2026-07-10
chore: Ignore blocked_roles_list in external OAuth import verify

sfc-gh-kwasilewski:
TestAcc_ExternalOauthIntegration_CompleteUseCase sets only external_oauth_allowed_roles_list. On older Snowflake versions the blocked list then stayed empty, but on newer versions both lists are always populated, so ACCOUNTADMIN and SECURITYADMIN now show up in the blocked list. ImportStateVerify sees the extra roles and fails.

Ignoring external_oauth_blocked_roles_list on import mirrors the other import steps in this file. It is safe to do so because the read path suppresses the privileged roles via
IgnoreValuesFromSetIfParamSet.

### DIFF
--- a/pkg/testacc/resource_external_oauth_integration_acceptance_test.go
+++ b/pkg/testacc/resource_external_oauth_integration_acceptance_test.go
@@ -315,6 +315,7 @@ func TestAcc_ExternalOauthIntegration_CompleteUseCase(t *testing.T) {
 					"external_oauth_any_role_mode",
 					"external_oauth_scope_delimiter",
 					"external_oauth_allowed_roles_list",
+					"external_oauth_blocked_roles_list",
 				},
 			},
 		},


pkg/testacc/resource_external_oauth_integration_acceptance_test.go
```diff
@@ -315,6 +315,7 @@ func TestAcc_ExternalOauthIntegration_CompleteUseCase(t *testing.T) {
 					"external_oauth_any_role_mode",
 					"external_oauth_scope_delimiter",
 					"external_oauth_allowed_roles_list",
+					"external_oauth_blocked_roles_list",
```
sfc-gh-jcieslak, line 318: An alternative solution (I think) would be to filter out those roles during the import. Not that it matters, the only difference what's in the state (this approach: additional roles, filtering in import: only user-set roles as previously). I'm fine with both, as the behavior is the same.

sfc-gh-jcieslak (approved):

pkg/testacc/resource_external_oauth_integration_acceptance_test.go
sfc-gh-kwasilewski: We're going to look into imports and how to test them soon, so let's leave it as it is for now.

PR #4998, opened 2026-07-13
chore: Rename FileFormats to FileFormatsLegacy

sfc-gh-jmichalak:
## Summary
- Renames the SDK's `FileFormats` client and related types (`FileFormat`, `CreateFileFormatOptions`, `AlterFileFormatOptions`, `DropFileFormatOptions`, `ShowFileFormatOptions`, `LegacyFileFormatTypeOptions`, etc.) to their `*Legacy` counterparts, freeing up the unqualified names for an upcoming non-legacy SDK implementation.
- Updates all callers across resources, data sources, schemas, acceptance helpers, and integration tests to use the new `*Legacy` names.


pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/table_check_constraint_desc_output_gen.go
```diff
@@ -0,0 +1,89 @@
+// Code generated by resource show output assertions generator (v0.1.0); DO NOT EDIT.
+
+package resourceshowoutputassert
```
sfc-gh-asawicki: q: I don't see this object being added to any object list, so why was it generated in this PR?

sfc-gh-asawicki (changes requested):

pkg/acceptance/bettertestspoc/assert/resourceshowoutputassert/table_check_constraint_desc_output_gen.go
sfc-gh-jmichalak: It was not added before.

sfc-gh-asawicki (approved):

PR #4987, opened 2026-07-09
feat: add interactive warehouse to the SDK

sfc-gh-uvarma:
## Description

Adds SDK support for [interactive warehouses](https://docs.snowflake.com/en/user-guide/warehouses-interactive). This is the first PR in a series adding first-class Terraform support for the `INTERACTIVE` warehouse type; it covers only the SDK layer (definitions + unit tests). Integration tests and the resource follow in subsequent PRs.

## Changes

- **New warehouse type + create operation.** Add `INTERACTIVE` to the warehouse type enum and a `CreateInteractive` custom operation.
- **ALTER extensions.** `ALTER WAREHOUSE` gains `ADD TABLES (...)` / `DROP TABLES (...)` (each its own top-level clause) and `SET`/`UNSET FALLBACK_WAREHOUSE`.
- **SHOW parsing.** The `SHOW WAREHOUSES` row parses the `tables` column (a comma-separated list of fully-qualified table names) into `[]SchemaObjectIdentifier`.
   - Parsing is **quote-aware**: identifiers can contain commas when quoted (e.g. `DB.SCHEMA."a,b"`), so a naive `strings.Split(",")` would break them apart. Existing methods don't handle this correctly.

## Test Plan

- Added unit tests for the generated DDL asserting the exact SQL.
- Added `TestSplitCommaOutsideQuotes` covering the quote-aware splitter, including commas inside quoted identifiers and doubled-quote (`""`) escaping.

## References

Part of the interactive-warehouse feature. Follow-up PRs: SDK integration tests, then the `snowflake_warehouse_interactive` resource.

PR #5000, opened 2026-07-13
feat: add SDK integration tests for interactive warehouses

sfc-gh-uvarma:
## Description

PR 2 (of 3) in the interactive-warehouse series. Adds SDK integration tests that validate the interactive-warehouse SDK behavior (added in #4987) against a live Snowflake connection.

## Changes

- **`TestInt_Warehouses_Interactive`** covering:
  - `CreateInteractive` with and without table associations,
  - `ALTER ... ADD TABLES` / `DROP TABLES` and asserting the resulting `Tables` on `SHOW`,
  - `SET`/`UNSET FALLBACK_WAREHOUSE`, asserting the value via `SHOW PARAMETERS ... IN WAREHOUSE` (fallback is exposed as a warehouse **parameter**, not a `SHOW WAREHOUSES` column),
- **`Warehouse.IsInteractiveWarehouse()`** helper introduced. It reports `type == INTERACTIVE`; also used by the resource (next PR)
- **Test client helpers:**
  - `Warehouse.CreateInteractive` / `CreateInteractiveWithRequest`
  - `Table.CreateInteractiveTable` — creates an interactive table fixture via raw SQL.

## Test Plan

Run against an account with interactive warehouses enabled:

```
go test -tags account_level_tests ./pkg/sdk/testint/ -run '^TestInt_Warehouses_Interactive$'
```

Validated against a QA account — all subtests pass.

## References

Part of the interactive-warehouse feature. Builds on the SDK PR; the `snowflake_warehouse_interactive` resource follows.

### DIFF
--- a/pkg/sdk/testint/warehouses_integration_test.go
+++ b/pkg/sdk/testint/warehouses_integration_test.go
@@ -967,18 +967,6 @@ func TestInt_Warehouses_Interactive(t *testing.T) {
 	client := testClient(t)
 	ctx := testContext(t)
 
-	// createInteractiveTable creates an interactive table fixture and returns its id with a cleanup.
-	createInteractiveTable := func(t *testing.T) sdk.SchemaObjectIdentifier {
-		t.Helper()
-		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
-		_, err := client.ExecForTests(ctx, fmt.Sprintf(`CREATE INTERACTIVE TABLE %s (id INT) CLUSTER BY (id)`, id.FullyQualifiedName()))
-		require.NoError(t, err)
-		t.Cleanup(func() {
-			_, _ = client.ExecForTests(ctx, fmt.Sprintf(`DROP TABLE IF EXISTS %s`, id.FullyQualifiedName()))
-		})
-		return id
-	}
-
 	t.Run("create interactive: minimal", func(t *testing.T) {
 		id := testClientHelper().Ids.RandomAccountObjectIdentifier()
 		err := client.Warehouses.CreateInteractive(ctx, sdk.NewCreateInteractiveWarehouseRequest(id))
@@ -993,8 +981,10 @@ func TestInt_Warehouses_Interactive(t *testing.T) {
 	})
 
 	t.Run("create interactive: with tables", func(t *testing.T) {
-		table1 := createInteractiveTable(t)
-		table2 := createInteractiveTable(t)
+		table1, table1Cleanup := testClientHelper().Table.CreateInteractiveTable(t)
+		t.Cleanup(table1Cleanup)
+		table2, table2Cleanup := testClientHelper().Table.CreateInteractiveTable(t)
+		t.Cleanup(table2Cleanup)
 
 		id := testClientHelper().Ids.RandomAccountObjectIdentifier()
 		err := client.Warehouses.CreateInteractive(ctx, sdk.NewCreateInteractiveWarehouseRequest(id).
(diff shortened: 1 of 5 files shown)


pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -962,3 +962,109 @@ func TestInt_Warehouses_Experimental(t *testing.T) {
 		assert.Equal(t, warehouseId2.Name(), warehouses[0].Name)
 	})
 }
+
+func TestInt_Warehouses_Interactive(t *testing.T) {
+	client := testClient(t)
+	ctx := testContext(t)
+
+	// createInteractiveTable creates an interactive table fixture and returns its id with a cleanup.
+	createInteractiveTable := func(t *testing.T) sdk.SchemaObjectIdentifier {
```
sfc-gh-kwasilewski: Le'ts use the `CreateInteractiveTable` function that you added to the table client.

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -962,3 +962,109 @@ func TestInt_Warehouses_Experimental(t *testing.T) {
...
+
+	t.Run("create interactive: minimal", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomAccountObjectIdentifier()
+		err := client.Warehouses.CreateInteractive(ctx, sdk.NewCreateInteractiveWarehouseRequest(id))
+		require.NoError(t, err)
+		t.Cleanup(testClientHelper().Warehouse.DropWarehouseFunc(t, id))
+
+		w, err := client.Warehouses.ShowByID(ctx, id)
+		require.NoError(t, err)
+		assert.Equal(t, id.Name(), w.Name)
```
sfc-gh-kwasilewski: 1. Ideally, we should assert all warehouse properties and parameters, not only those closely related to interactive warehouses. 
2. Let's use generated object assertions (`objectassert.Warehouse`) - they're better than raw assertions due to their fluent API that aggregates all mismatches into one error report with rich and uniform error messages. 

Take a look at `create adaptive: minimal` test - it implements both of these points.

This also applies to `create interactive: with tables`

pkg/sdk/testint/warehouses_integration_test.go
```diff
@@ -962,3 +962,109 @@ func TestInt_Warehouses_Experimental(t *testing.T) {
...
+			WithTables([]sdk.SchemaObjectIdentifier{table1}))
+		require.NoError(t, err)
+		t.Cleanup(testClientHelper().Warehouse.DropWarehouseFunc(t, id))
+
+		err = client.Warehouses.Alter(ctx, sdk.NewAlterWarehouseRequest(id).WithAddTables([]sdk.SchemaObjectIdentifier{table2}))
+		require.NoError(t, err)
+
+		w, err := client.Warehouses.ShowByID(ctx, id)
+		require.NoError(t, err)
+		assert.ElementsMatch(t, []sdk.SchemaObjectIdentifier{table1, table2}, w.Tables)
```
sfc-gh-kwasilewski: Just wanted to confirm: are these table identifiers always returned in a specific order?

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-uvarma: Sorry I missed this! Fixed

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-uvarma: Thanks for the pointer! Done.

pkg/sdk/testint/warehouses_integration_test.go
sfc-gh-uvarma: Good callout. Order is not guaranteed by Snowflake (in some cases in *can* change without user action). The new `HasTables` helper is order-insensitive and I used it here.

pkg/sdk/warehouses_ext.go
```diff
@@ -379,3 +379,9 @@ func (s *CreateAdaptiveWarehouseRequest) ID() AccountObjectIdentifier {
 func (s *CreateInteractiveWarehouseRequest) ID() AccountObjectIdentifier {
 	return s.name
 }
+
+// IsInteractiveWarehouse reports whether the warehouse is interactive. Snowflake surfaces this
+// through the type column (type = INTERACTIVE); there is no separate is_interactive column.
+func (w *Warehouse) IsInteractiveWarehouse() bool {
```
sfc-gh-kwasilewski, line 385: Now it's really not used :) Let's remove it in a follow-up PR.

sfc-gh-kwasilewski (approved):

pkg/sdk/warehouses_ext.go
sfc-gh-kwasilewski: ...unless you use it in the resource - in that case, it's fine.

PR #5001, opened 2026-07-13
fix(cache): GRANT_ACCOUNT_ROLE_SHOW_CACHING no longer blocks parallel role lookups

raffael-dzikowski:
## Summary

`Cache.GetOrLoad` (added in #4855, `GRANT_ACCOUNT_ROLE_SHOW_CACHING`) held a single process-wide write lock across the entire `loadFn` call on a cache miss, including the network round-trip to Snowflake. The provider framework reads resources with worker-pool parallelism; this serialized cache misses on *different* keys behind one lock, defeating parallelism for exactly the calls that most need it — first-time lookups.

This surfaced as a real-world regression downstream: rolling out the experiment across a large RBAC configuration (thousands of `snowflake_grant_account_role` resources across several Terraform layers) showed CI plan times regressing sharply in layers with low role reuse — up to 4.1x slower in one layer where every grant targets a distinct role — while layers with high role reuse were only mildly affected. Comparing `SHOW GRANTS OF ROLE` call counts and Snowflake-side execution time via `QUERY_HISTORY` before/after confirmed the cache was doing exactly what it's supposed to (fewer calls, cheaper calls, less total Snowflake-side compute) — the extra wall-clock time was being spent waiting on the lock, not on Snowflake.

## Changes

- `pkg/internal/provider/cache.go` — replaces the global lock-across-`loadFn` with `golang.org/x/sync/singleflight.Group`. Concurrent misses on the *same* key still share one in-flight call and its result (dedup preserved, including on error — singleflight forgets a key once its call completes, so the next call always retries fresh). Misses on *different* keys now run fully in parallel. The mutex is retained only for the short critical sections that read/write the map, never across `loadFn`.
- `pkg/internal/provider/cache_test.go` — two new deterministic unit tests (no `time.Sleep`, no wall-clock thresholds):
  - `TestCache_ConcurrentMissesOnDifferentKeysRunInParallel`: a rendezvous/barrier where every `loadFn` signals it has started, then blocks until all of them have. Under the old implementation only one goroutine ever reaches `loadFn`; the rest block on the outer lock and never signal, so the barrier never fills and the test fails via a bounded timeout rather than hanging. I verified this manually — reverted `GetOrLoad` to the old body locally, confirmed this test fails with the expected timeout message, then restored the fix.
  - `TestCache_ConcurrentMissesOnSameKeyCollapseToOneLoad`: guards that same-key dedup still holds under real concurrency (not just the sequential case `TestCache_HitSkipsLoadFn` already covers).
- `MIGRATION_GUIDE.md` — brief entry under the existing `v2.18.x ➞ v2.19.0` heading; no configuration changes required.
- `go.mod` — `golang.org/x/sync` promoted from `// indirect` to direct (it was already fully resolved in `go.sum`; no version change).

Public API of `Cache[T]` is unchanged, so `grant_account_role.go`, `provider_context.go`, and `provider.go` needed no changes. The existing acceptance tests (`resource_grant_account_role_show_caching_experimental_features_acceptance_test.go`) already prove the round-trip-reduction property against live Snowflake via `QUERY_HISTORY` counts and are untouched.

## Testing

- `go test -race -run TestCache -v ./pkg/internal/provider/...` — all pass, including the two new tests, no data races.
- `go vet ./pkg/provider/... ./pkg/resources/... ./pkg/datasources/... ./pkg/sdk/... ./pkg/sdk/testint/... ./pkg/testacc/...` and the `non_account_level_tests`-tagged variant — clean.
- `gofumpt -l` on the changed files — clean, no diff.
- `golangci-lint run` scoped to the package — 0 issues.
- `go run ./pkg/internal/tools/doc-gen-helper/` — no diff in generated docs (the experiment's doc string describes only user-facing behavior, which is unchanged).


pkg/internal/provider/cache.go
```diff
@@ -28,30 +39,48 @@ func NewCache[T any]() *Cache[T] {
...
 		return zero, err
 	}
-	c.data[key] = v
-	return v, nil
+	// Do returns any (it predates generics). Every non-error return path above returns a
+	// concrete T, so this assertion is total.
+	return v.(T), nil
 }
 
 // Invalidate removes the cached result for key, forcing the next GetOrLoad call
```
sfc-gh-jmichalak, line 104: A few small things that could be addressed later:

1. Please call c.group.Forget here. Otherwise, a slow in-flight load that started before a Create/Delete mutation can finish after that mutation's Invalidate() call and silently re-store stale data.
2. This method should also cancel the loadFn's context.
3. Previously each blocked caller retried loadFn independently once it got the lock; now one transient SHOW GRANTS error fails every resource sharing that role in the same plan.

sfc-gh-jmichalak (dismissed):
LGTM with small improvements proposal. Let's wait for @sfc-gh-asawicki's review.

MIGRATION_GUIDE.md
```diff
@@ -48,6 +48,10 @@ These features will be marked as stable in future releases. To use them, add `sn
 
 Stay tuned for the next variants of Iceberg Tables support in the provider!
 
+### *(bugfix)* snowflake_grant_account_role SHOW GRANTS caching no longer serializes parallel reads
```
sfc-gh-asawicki: nit: I would argue if this is a bugfix or improvement

sfc-gh-asawicki (dismissed):

sfc-gh-jmichalak (approved):

sfc-gh-asawicki (approved):

PR #5002, opened 2026-07-14
chore: add snowflake defaults account

sfc-gh-jcieslak:
## Changes
- Add snowflake defaults test account profile to test Snowflake default behavior (e.g. in case of "BUILT-IN" authentication policy set on account when no user-defined policy is attached)


pkg/sdk/README.md
```diff
@@ -2,6 +2,8 @@
 
 [secondary_test_account] credentials are required in the Snowflake profile if running integration tests that provision resources in multiple accounts:
 
+[snowflake_defaults_test_account] credentials are required in non-prod environments for integration tests that check for Snowflake default behavior (e.g. having the built-in account authentication policy when no user-defined policy is attached).
```
sfc-gh-asawicki, line 7: nit: can we add azure profile here too?

pkg/testacc/0_setup_test.go
```diff
@@ -42,9 +42,10 @@ type snowflakeTestEnvironmentContext struct {
 }
 
 type acceptanceTestContext struct {
```
sfc-gh-asawicki, line 44: q: what about sweepers?

sfc-gh-asawicki (changes requested):

pkg/sdk/README.md
sfc-gh-jcieslak: Added

pkg/testacc/0_setup_test.go
sfc-gh-jcieslak: Forgot about them, added missing sweepers (for both azure and new account).

sfc-gh-asawicki (dismissed):

sfc-gh-asawicki (approved):

PR #5005, opened 2026-07-14
chore: Add mcp server sdk

sfc-gh-jcieslak:
## Changes
- Add MCP SDK
- Add integration tets with generated assertions
- Adjust test client

## To discuss
The controvertial part is that the spec is passed to the Snowflake object in YAML format and returned as JSON (which is subset of YAML format, so "still" YAML is returned, but in "different formatting"). To address later resource handling and current comparison in integration test assertions I introduced normalization function that uses already used yaml library to parse either input (YAML) or output (YAML in JSON formatting) and use json.Marshal to output final output as it sorts and serializes the maps which are output from yaml parser (which fits the form later used in resources for sdkValidation and NormalizeAndCompare).

## TODO
- Discuss with mcp team missing COMMENT property in docs that can be set


pkg/sdk/generator/defs/mcp_servers_def.go
```diff
@@ -0,0 +1,70 @@
...
+).
+	CreateOperation(
+		"https://docs.snowflake.com/en/sql-reference/sql/create-mcp-server",
+		g.NewQueryStruct("CreateMcpServer").
+			Create().
+			OrReplace().
+			SQL("MCP SERVER").
+			IfNotExists().
+			Name().
+			OptionalTextAssignment("COMMENT", g.ParameterOptions().SingleQuotes()).
```
sfc-gh-kwasilewski: nit

```suggestion
			OptionalComment().
```

pkg/sdk/testint/mcp_servers_integration_test.go
```diff
@@ -0,0 +1,139 @@
...
+	require.NoError(t, err)
+
+	defaultSpec := testClientHelper().McpServer.DefaultSpec()
+	normalizedDefaultSpec, err := sdk.NormalizeMcpServerSpecification(defaultSpec)
+	require.NoError(t, err)
+
+	t.Run("create: basic", func(t *testing.T) {
+		id := testClientHelper().Ids.RandomSchemaObjectIdentifier()
+
+		cleanup := testClientHelper().McpServer.CreateWithRequest(t, sdk.NewCreateMcpServerRequest(id, defaultSpec))
```
sfc-gh-kwasilewski: Let's not use the test client in create test, use `client.McpServers.Create` instead. Also applies to `"create: complete"`

pkg/sdk/testint/mcp_servers_integration_test.go
```diff
@@ -0,0 +1,139 @@
...
+				HasDatabaseName(id.DatabaseName()).
+				HasSchemaName(id.SchemaName()).
+				HasOwner(snowflakeroles.Accountadmin.Name()).
+				HasComment(comment).
+				HasServerSpec(normalizedComplexSpec).
+				HasCreatedOnNotEmpty(),
+		)
+	})
+
+	t.Run("show", func(t *testing.T) {
```
sfc-gh-kwasilewski, line 100: It'd be good to test all possible `in` variants: account, database, schema.

pkg/sdk/generator/defs/mcp_servers_def.go
sfc-gh-jcieslak: Used

pkg/sdk/testint/mcp_servers_integration_test.go
sfc-gh-jcieslak: True, refactored

pkg/sdk/testint/mcp_servers_integration_test.go
sfc-gh-jcieslak: Done

pkg/sdk/testint/mcp_servers_integration_test.go
```diff
@@ -100,15 +102,37 @@ func TestInt_McpServers(t *testing.T) {
...
+			})
+		}
+
+		t.Run("in_account", func(t *testing.T) {
+			showResults, err := client.McpServers.Show(ctx, sdk.NewShowMcpServerRequest().WithIn(sdk.In{Account: sdk.Bool(true)}))
+			require.NoError(t, err)
 
-		result, err := collections.FindFirst(showResults, func(result sdk.McpServer) bool {
-			return result.ID().FullyQualifiedName() == id.FullyQualifiedName()
+			require.Len(t, showResults, 1)
```
sfc-gh-kwasilewski: `GreaterOrEqual`? It's safer not to assume there are no other MCP servers in the account.

sfc-gh-jcieslak: Changed (also added in other cases the len check)

sfc-gh-kwasilewski (approved):

PR #5009, opened 2026-07-14
feat: add snowflake_warehouse_interactive resource

sfc-gh-uvarma:
## Description

Final PR in the interactive-warehouse series (the SDK PR #4987 and integration-test PR #5000 are already merged). Adds a new **preview** resource, `snowflake_warehouse_interactive`, for managing [interactive warehouses](https://docs.snowflake.com/en/user-guide/warehouses-interactive). Gated behind the `snowflake_warehouse_interactive_resource` preview feature.

## Changes

- **New resource `snowflake_warehouse_interactive`** with full CRUD + import
- **Data source:** the existing `snowflake_warehouses` data source now surfaces the interactive `tables` field via the shared show schema.
- **Registration & scaffolding:** preview feature (`snowflake_warehouse_interactive_resource`), provider/resource registration, bettertestspoc model / assert / show-output generators, issue labels.
- **Acceptance tests, docs, examples, migration guide.**

## Test Plan

- Acceptance tests pass against a QA account with the preview feature enabled.
- `make pre-push` (check-compilation, all generators, `mod`, `fmt`, docs, `golangci-lint`, architecture tests) — all clean, no drift.

## Notes for reviewers

- `fallback_warehouse` drift detection relies on `SHOW PARAMETERS`; identifiers containing double quotes are unsupported provider-wide (`SNOW-1571674`), which also bounds the `tables` parsing.


templates/resources/warehouse_interactive.md.tmpl
```diff
@@ -0,0 +1,37 @@
+---
+page_title: "{{.Name}} {{.Type}} - {{.ProviderName}}"
```
sfc-gh-kwasilewski, line 2: This template is effectively the same as the default template for preview resources (`templates/resources.md.tmpl`). Let's use the default one (you just need to remove the specialized template, and the default template will be used automatically).

MIGRATION_GUIDE.md
```diff
@@ -50,6 +50,22 @@ Stay tuned for the next variants of Iceberg Tables support in the provider!
 
 ## v2.17.x ➞ v2.18.0
 
+### *(new preview resource)* New interactive warehouse resource
```
sfc-gh-kwasilewski: Wrong place - v2.18.0 has already been released. Put it on the  top of the file.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+		Optional:         true,
+		Default:          string(sdk.WarehouseSizeXSmall),
+		ValidateDiagFunc: sdkValidation(sdk.ToWarehouseSize),
+		DiffSuppressFunc: SuppressIfAny(NormalizeAndCompare(sdk.ToWarehouseSize), IgnoreChangeToCurrentSnowflakeValueInShow("size")),
+		Description:      fmt.Sprintf("Specifies the size of the interactive warehouse. Valid values are (case-insensitive): %s.", possibleValuesListed(sdk.ValidWarehouseSizesString)),
+	},
+	"max_cluster_count": {
+		Type:             schema.TypeInt,
+		Optional:         true,
+		Computed:         true,
```
sfc-gh-kwasilewski: Why computed? This property is returned in the SHOW output, so we always know its value. We only use Computed for Snowflake parameters and special attributes (`ShowOutputAttributeName`, `ParametersAttributeName`, and `FullyQualifiedNameAttributeName`).

Also applies to `min_cluster_count` and `auto_suspend`.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+	"name": {
+		Type:             schema.TypeString,
+		Required:         true,
+		Description:      blocklistedCharactersFieldDescription("Identifier for the interactive warehouse; must be unique for your account."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"warehouse_size": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		Default:          string(sdk.WarehouseSizeXSmall),
```
sfc-gh-kwasilewski: That's not how we handle default values in the provider. In general, we avoid hardcoding Snowflake defaults in the resource's schema, because they may change, and that will be problematic for the provider. 

Take a look at how the standard warehouse resource handles this attribute.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+// interactiveWarehouseMinAutoSuspend is the minimum AUTO_SUSPEND (in seconds) accepted by Snowflake
+// for interactive warehouses.
+const interactiveWarehouseMinAutoSuspend = 86400
```
sfc-gh-kwasilewski: I know it's specified in the docs, but are you sure this limit will never change and it will always be the same for all customers? If not, it's better not to validate this at the client-side, and instead check if the value is positive (`validation.IntAtLeast(1)`).

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+
+	name := d.Get("name").(string)
+	id, err := sdk.ParseAccountObjectIdentifier(name)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	req := sdk.NewCreateInteractiveWarehouseRequest(id)
+
+	if v, ok := d.GetOk("tables"); ok {
```
sfc-gh-kwasilewski, line 281: Please use create builders defined in `resource_helpers_create.go`.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+			if err := client.Warehouses.Alter(ctx, sdk.NewAlterWarehouseRequest(id).WithDropTables(toDrop)); err != nil {
+				return diag.FromErr(fmt.Errorf("error dropping tables from interactive warehouse %s: %w", id.FullyQualifiedName(), err))
+			}
+		}
+	}
+
+	set := sdk.NewWarehouseSetRequest()
+	unset := sdk.NewWarehouseUnsetRequest()
+
+	if d.HasChange("warehouse_size") {
```
sfc-gh-kwasilewski, line 460: Please use update helpers defined in `resource_helpers_update.go`.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+		fallbackWarehouse := fallbackWarehouseFromParameters(warehouseParameters)
+
+		if withExternalChangesMarking {
+			sizeVal, sizeStr := optionalStringOutputMapping(w.Size)
+			autoResumeVal, autoResumeStr := w.AutoResume, booleanStringFromBool(w.AutoResume)
+			if err = handleExternalChangesToObjectInShow(
+				d,
+				outputMapping{"size", "warehouse_size", sizeStr, sizeVal, nil},
+				outputMapping{"auto_resume", "auto_resume", autoResumeVal, autoResumeStr, nil},
+				outputMapping{"resource_monitor", "resource_monitor", w.ResourceMonitor.Name(), w.ResourceMonitor.Name(), nil},
```
sfc-gh-kwasilewski, line 386: All optional properties with Snowflakes defaults needs be to handled here. You're  now missing `max_cluster_count`, `min_cluster_count`, and `auto_suspend`.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+			minClusterCountValue = *w.MinClusterCount
+		}
+
+		errs := errors.Join(
+			d.Set("name", w.Name),
+			d.Set("comment", w.Comment),
+			d.Set("warehouse_type", effectiveType),
+			d.Set("auto_suspend", autoSuspendValue),
+			d.Set("max_cluster_count", maxClusterCountValue),
+			d.Set("min_cluster_count", minClusterCountValue),
```
sfc-gh-kwasilewski: Once you add these three attributes to `handleExternalChangesToObjectInShow`, please remove them from here.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+				d,
+				outputMapping{"size", "warehouse_size", sizeStr, sizeVal, nil},
+				outputMapping{"auto_resume", "auto_resume", autoResumeVal, autoResumeStr, nil},
+				outputMapping{"resource_monitor", "resource_monitor", w.ResourceMonitor.Name(), w.ResourceMonitor.Name(), nil},
+			); err != nil {
+				return diag.FromErr(err)
+			}
+		}
+
+		if err = setStateToValuesFromConfig(d, warehouseInteractiveSchema, []string{
```
sfc-gh-kwasilewski: This a legacy function which we no longer use in new resources, you should be able to safely remove it.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	w, err := client.Warehouses.ShowByID(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	if !w.IsInteractiveWarehouse() {
+		return nil, fmt.Errorf("warehouse %s is not an interactive warehouse; use snowflake_warehouse instead", id.FullyQualifiedName())
```
sfc-gh-kwasilewski:
```suggestion
		return nil, fmt.Errorf("warehouse %s is not an interactive warehouse; use snowflake_warehouse or snowflake_warehouse_adaptive instead", id.FullyQualifiedName())
```

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+	}
+
+	tables := make([]string, len(w.Tables))
+	for i, table := range w.Tables {
+		tables[i] = table.FullyQualifiedName()
+	}
+
+	errs := errors.Join(
```
sfc-gh-kwasilewski, line 251: Terraform executes the `Read` function right after the `Import` function, so we don't need to set attributes here.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+		DeleteContext: PreviewFeatureDeleteContextWrapper(string(previewfeatures.WarehouseInteractiveResource), TrackingDeleteWrapper(resources.WarehouseInteractive, deleteFunc)),
+		Description:   "Resource used to manage interactive warehouse objects. Interactive warehouses are optimized for low-latency, high-concurrency queries against a defined set of tables. For more information, check [interactive warehouse documentation](https://docs.snowflake.com/en/user-guide/warehouses-interactive).",
+
+		Schema: warehouseInteractiveSchema,
+		Importer: &schema.ResourceImporter{
+			StateContext: TrackingImportWrapper(resources.WarehouseInteractive, ImportWarehouseInteractive),
+		},
+
+		CustomizeDiff: TrackingCustomDiffWrapper(resources.WarehouseInteractive, customdiff.All(
+			ComputedIfAnyAttributeChanged(warehouseInteractiveSchema, ShowOutputAttributeName, "name", "warehouse_size", "max_cluster_count", "min_cluster_count", "auto_suspend", "auto_resume", "resource_monitor", "comment", "tables", "fallback_warehouse"),
```
sfc-gh-kwasilewski: `fallback_warehouse` is not a part of `show_output` because it's a parameter. Instead, it should be handled below, together with other parameters.

pkg/schemas/warehouse_gen.go
```diff
@@ -156,10 +156,28 @@ var ShowWarehouseSchemaAdaptive = collections.MergeMaps(showWarehouseSchemaCommo
 
 var _ = ShowWarehouseSchemaAdaptive
 
-// ShowAllWarehousesSchema merges common, regular, and adaptive fields.
+// showWarehouseSchemaInteractive contains fields only present for interactive warehouses.
```
sfc-gh-kwasilewski, line 159: There's a single `SHOW WAREHOUSE` command for all warehouse types, so we should have a single show schema in the provider as well.

MIGRATION_GUIDE.md
```diff
@@ -50,6 +50,22 @@ Stay tuned for the next variants of Iceberg Tables support in the provider!
...
+
+We have added a new preview resource for managing interactive warehouses: [snowflake_warehouse_interactive](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/warehouse_interactive) ([Snowflake docs](https://docs.snowflake.com/en/user-guide/warehouses-interactive)).
+
+This feature will be marked as stable in future releases. To use it, add `snowflake_warehouse_interactive_resource` to the `preview_features_enabled` field in the provider configuration.
+
+A few interactive-specific constraints apply:
```
sfc-gh-jcieslak: The limitations are usually a part of resource specific documentation rather than in migration guide. Let's document them in templates/resources/warehouse_interactive.md.tmpl and let's say/link that more information on the resource can be read in the resource documentation.

templates/resources/warehouse_interactive.md.tmpl
sfc-gh-jcieslak: The limitations should be put here, so let's leave it and modify it (then, it won't be the same as default)

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,574 @@
...
+	"min_cluster_count": {
+		Type:             schema.TypeInt,
+		Optional:         true,
+		Computed:         true,
+		ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(1)),
+		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("min_cluster_count"),
+		Description:      "Specifies the minimum number of server clusters for the interactive warehouse (only applies to multi-cluster warehouses). Snowflake always assigns a value, so this field is computed when not set.",
+	},
+	"auto_suspend": {
+		Type:             schema.TypeInt,
```
sfc-gh-jcieslak, line 56: This also may need IntDefault (look at warehouse.go implementation of this field)

pkg/testacc/resource_warehouse_interactive_acceptance_test.go
```diff
@@ -0,0 +1,219 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/bettertestspoc/config/model"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/acceptance/helpers/random"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/resources"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+func TestAcc_WarehouseInteractive_Basic(t *testing.T) {
```
sfc-gh-jcieslak: Look at other recent examples of acceptance tests and see how ...BasicUseCase and ...CompleteUseCase tests are composed (which models and steps are used), e.g. test steps in pkg/testacc/resource_api_integration_amazon_api_gateway_acceptance_test.go. We're missing some of the steps in this file (e.g. setting all optional fields from model with no optional fields and vice-versa). Other, more concrete tests like:
TestAcc_WarehouseInteractive_TablesDelta
TestAcc_WarehouseInteractive_Import_WrongWarehouseType
TestAcc_WarehouseInteractive_Validations
Could be left as they are, the TestAcc_WarehouseInteractive_FallbackWarehouse could be removed once this test would be more full as in other acceptance tests with both ...BasicUseCase and ...CompleteUseCase tests.

MIGRATION_GUIDE.md
sfc-gh-uvarma: Done. Thank you!

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Thanks. This makes sense. Removed `Computed: true`.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Done.

MIGRATION_GUIDE.md
sfc-gh-uvarma: Fixed. Sorry, I didn't resolve the merge conflict correctly.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Makes sense. Done.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Thanks. This makes sense to me. Removed defaults.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Done.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Thanks. I assume `warehouse.go` has this because of a legacy issue, but we don't need it anymore?

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Done. Thanks!

pkg/schemas/warehouse_gen.go
sfc-gh-uvarma: That makes sense to me, but we've got `ShowWarehouseSchemaAdaptive` above. It takes the common schema and adds `max_query_performance_level` and `query_throughput_multiplier`. I figured doing the same for interactive (adding `tables`) made sense.

How would you like to reconcile this? Adaptive has already split this, so I was following the pattern. As part of addressing code review comments, I realized that `FALLBACK_WAREHOUSE` was not correctly being treated as a parameter, so I'm also having to split `ShowWarehouseParametersSchema` as well.

I'm happy to do whatever you think is best: I don't know what the best practices are when making multiple resource types in Terraform for variants of the same entity type in Snowflake.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Done.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Done.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Done.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Mostly done. `tables`, `initially_suspended`, and `resource_monitor` don't have existing builders that I can use. If you'd like, I can add builders for those in this PR and use them.

* `tables`: this is unique, might not be useful elsewhere
* `resource_monitor`: it looks like other callsites could be migrated to a new builder if we build one (obviously, that migration should happen in a separate PR)
* `initially_suspended`: there is one for booleans represented as strings but not for native booleans.

pkg/testacc/resource_warehouse_interactive_acceptance_test.go
sfc-gh-uvarma: Done! This actually uncovered a subtle bug in this PR: `CREATE INTERACTIVE WAREHOUSE` implicitly sets the current warehouse to an interactive warehouse. When we subsequently run `ALTER` statements, they sometimes timeout because now we have the 5s timeout that interactive warehouses have. I fixed this by remembering the previous warehouse just before creating an interactive warehouse and then using the (previous) warehouse after creating the new warehouse.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Done. This surfaced an issue where the SDK didn't actually treat `fallback_warehouse` as a parameter. I've fixed the SDK in this PR as well. This breaks the contributing guidelines (resource and SDK changes in separate PRs). Let me know if you'd like me to split that out into a separate PR.

templates/resources/warehouse_interactive.md.tmpl
sfc-gh-uvarma: Thanks! Done.

pkg/resources/warehouse_interactive.go
sfc-gh-kwasilewski: * `tables`: I agree it's too specific, let's skip it.
* `resource_monitor`: The question is whether such a builder could be used for other attributes than `resource_monitor`. If not necessarily, let's keep it as it is.
* `initially_suspended`: We could add a builder for native booleans, but I'll let you decide - I'm already fine with these changes so far.

pkg/resources/warehouse_interactive.go
```diff
@@ -32,34 +28,31 @@ var warehouseInteractiveSchema = map[string]*schema.Schema{
...
 	"min_cluster_count": {
 		Type:             schema.TypeInt,
 		Optional:         true,
-		Computed:         true,
 		ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(1)),
 		DiffSuppressFunc: IgnoreChangeToCurrentSnowflakeValueInShow("min_cluster_count"),
-		Description:      "Specifies the minimum number of server clusters for the interactive warehouse (only applies to multi-cluster warehouses). Snowflake always assigns a value, so this field is computed when not set.",
+		Description:      "Specifies the minimum number of server clusters for the interactive warehouse (only applies to multi-cluster warehouses).",
 	},
 	"auto_suspend": {
```
sfc-gh-kwasilewski, line 55: I just noticed that auto-suspend can be disabled by setting the value to NULL, as per https://docs.snowflake.com/en/sql-reference/sql/create-interactive-warehouse. I'm afraid this attribute doesn't allow for that in it's current form. 

Does setting AUTO_SUSPEND to 0 is the same as setting it to NULL? If so, we could just slightly modify validation and allow zeros.

pkg/resources/warehouse_interactive.go
sfc-gh-kwasilewski: Yes. It's just unnecessary duplication.

pkg/resources/warehouse_interactive.go
```diff
@@ -387,51 +380,31 @@ func ReadWarehouseInteractiveFunc(withExternalChangesMarking bool) schema.ReadCo
...
-		}
-		minClusterCountValue := 0
-		if w.MinClusterCount != nil {
-			minClusterCountValue = *w.MinClusterCount
-		}
-
 		errs := errors.Join(
 			d.Set("name", w.Name),
 			d.Set("comment", w.Comment),
 			d.Set("warehouse_type", effectiveType),
```
sfc-gh-kwasilewski, line 395: I must have missed that - since warehouse_type has a Snowflake default value, and it's handled by `handleExternalChangesToObjectInShow`, we don't need to set it in the state.

pkg/resources/warehouse_interactive.go
sfc-gh-kwasilewski: It's ok, this addition is so tiny that there's no need to split.

pkg/resources/warehouse_interactive.go
```diff
@@ -325,11 +301,28 @@ func CreateWarehouseInteractive(ctx context.Context, d *schema.ResourceData, met
 		req.WithStatementTimeoutInSeconds(*v)
 	}
 
+	// CREATE WAREHOUSE makes the newly created warehouse the current warehouse for the session
+	// connection. For an interactive warehouse this forces a 5-second statement timeout, which would
+	// then apply to the follow-up ALTER, the post-create Read, and any later statement that reuses the
+	// same pooled connection (e.g. a subsequent apply or import). Capture the current warehouse first
+	// and restore it right after creating, so DDL/metadata operations keep running on the original
+	// warehouse rather than the interactive one.
+	previousWarehouse, err := client.ContextFunctions.CurrentWarehouse(ctx)
```
sfc-gh-kwasilewski: This is tricky. We don't usually require a warehouse in the session, because the provider mostly executes DDL statements (there are some exception, e.g. when we need to read something from INFORMATION_SCHEMA). So this can return nothing.

pkg/resources/warehouse_interactive.go
```diff
@@ -170,14 +163,17 @@ func WarehouseInteractive() *schema.Resource {
 		},
 
 		CustomizeDiff: TrackingCustomDiffWrapper(resources.WarehouseInteractive, customdiff.All(
-			ComputedIfAnyAttributeChanged(warehouseInteractiveSchema, ShowOutputAttributeName, "name", "warehouse_size", "max_cluster_count", "min_cluster_count", "auto_suspend", "auto_resume", "resource_monitor", "comment", "tables", "fallback_warehouse"),
+			ComputedIfAnyAttributeChanged(warehouseInteractiveSchema, ShowOutputAttributeName, "name", "warehouse_size", "max_cluster_count", "min_cluster_count", "auto_suspend", "auto_resume", "resource_monitor", "comment", "tables"),
 			ComputedIfAnyAttributeChanged(
 				warehouseInteractiveSchema, ParametersAttributeName,
 				strings.ToLower(string(sdk.WarehouseParameterMaxConcurrencyLevel)),
 				strings.ToLower(string(sdk.WarehouseParameterStatementQueuedTimeoutInSeconds)),
 				strings.ToLower(string(sdk.WarehouseParameterStatementTimeoutInSeconds)),
```
sfc-gh-kwasilewski, line 179: `fallback_warehouse` needs to be added here, just like other parameters.

pkg/schemas/warehouse_gen.go
sfc-gh-kwasilewski: Sorry, I missed that. I thought more about the date source, where it makes sense to have a single, unified schema, but we already have it: `ShowAllWarehousesSchema`. Your changes are fine, please disregard my earlier comment.

pkg/testacc/resource_warehouse_interactive_acceptance_test.go
```diff
@@ -0,0 +1,357 @@
...
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config:      accconfig.FromModels(t, modelInvalidAutoSuspend),
+				PlanOnly:    true,
+				ExpectError: regexp.MustCompile(`expected auto_suspend to be at least \(1\), got 0`),
```
sfc-gh-kwasilewski, line 399: There are more attributes with validators, let's test all of them here. PlanOnly tests are cheap, as there are no interactions with Snowflake.

pkg/testacc/resource_warehouse_interactive_acceptance_test.go
sfc-gh-kwasilewski: We could also increase timeouts. 5s timeout is not a big issue. See my [comment](https://github.com/snowflakedb/terraform-provider-snowflake/pull/5009#discussion_r3596178736) where I explained why remembering the previous warehouse might not be possible.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: For `resource_monitor` and `initially_suspended`, yes. I can do this in a follow up PR.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Makes sense.

pkg/schemas/warehouse_gen.go
sfc-gh-uvarma: Sounds good. Thanks. OK to close this comment thread?

pkg/testacc/resource_warehouse_interactive_acceptance_test.go
sfc-gh-uvarma: I've removed the hack to try to restore the previous warehouse (like you point out, if there isn't a current warehouse, we haven't fixed the problem). In my testing, the acceptance tests are now flaky because of the 5 second timeout.

I had the same idea as you to increase the statement timeout at the session level, but the semantics for this parameter are `min(warehouse_level, session_level)`.  https://docs.snowflake.com/en/sql-reference/parameters#statement_timeout_in_seconds

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Makes sense. Thanks. OK to resolve this thread?

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Thanks. The whole reason I did the SDK change was to fix this but then... I didn't fix this. 🤦 

Fixed. Required a couple more SDK changes.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Yeah, as discussed in the other thread, I'm not sure what we can do here. I agree that this is fragile (breaks when there is no current warehouse when we read this line). What would you think about creating a new session/client just for the `CREATE INTERACTIVE WAREHOUSE` statement?

To the best of my knowledge, our options are:

 1. accept flakiness (note that if there are N interactive warehouses we need to create in a plan, the (N+1)-th attempt to `apply` will not hit this problem); this isn't a good user experience.
 2. use a separate session for the `CREATE INTERACTIVE WAREHOUSE` statements. From the digging that I've done, there's just no way to unset the current warehouse in a session.
 3. use the bandaid I had before (restore the previous warehouse, if one was set). This masks the problem when the user has a warehouse selected (but the problem is still visible when the user does not have a current warehouse).

pkg/testacc/resource_warehouse_interactive_acceptance_test.go
sfc-gh-uvarma: Done!

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: I'm working on getting AUTO_SUSPEND=0 to work just like regular warehouses in Snowflake. Can we assume that will land soon and let this ship as-is? Today, if you leave it unspecified in a terraform config, the provider will `ALTER WAREHOUSE ... UNSET AUTO_SUSPEND`, which sets it to `NULL`, so we implicitly get the `null` in terraform -> `NULL` in Snowflake behavior.

Once Snowflake accepts 0, there will be a way to explicitly disable auto-suspend from terraform (rather than rely on the current default value).

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: IIUC, that won't work. In regular warehouses, we have `Optional: true` but not `Computed: true` (this is because you can switch between standard and snowpark optimized). For interactive warehouses, we have `Computed: true`. `handleExternalChangesToObjectInShow` won't set the value if there is no drift.

pkg/testacc/resource_warehouse_interactive_acceptance_test.go
sfc-gh-asawicki: > CREATE INTERACTIVE WAREHOUSE implicitly sets the current warehouse to an interactive warehouse. 

We have to mind other tests too, as this will impact them in the current setup. PLease create a dedicated ProviderFactory and replace: `ProtoV6ProviderFactories: TestAccProtoV6ProviderFactories,`.

> I had the same idea as you to increase the statement timeout at the session level, but the semantics for this parameter are min(warehouse_level, session_level).

I don't think the session level is the problem here, we have statements running longer, so let's just set it on the interactive warehouse level for each test involving running any alters. Let's have a separate test with just pure creation, without the setting. Let's document it in tests and, more importantly, in the user facing docs.

> When we subsequently run ALTER statements, they sometimes timeout because now we have the 5s timeout that interactive warehouses have.

Regardless of the above and how we handle this on the provider side, that's a pretty poor experience even through SQL if I create a warehouse and then can't even alter it by default because I hit the timeout on ALTER.

pkg/resources/warehouse_interactive.go
sfc-gh-kwasilewski: Please take a look at [Artur's comment](https://github.com/snowflakedb/terraform-provider-snowflake/pull/5009#discussion_r3601691578) - applying his suggestions should get rid of test flakiness.

pkg/resources/warehouse_interactive.go
sfc-gh-kwasilewski: Ok, makes sense.

pkg/resources/warehouse_interactive.go
```diff
@@ -0,0 +1,510 @@
...
+		Required:         true,
+		Description:      blocklistedCharactersFieldDescription("Identifier for the interactive warehouse; must be unique for your account."),
+		DiffSuppressFunc: suppressIdentifierQuoting,
+	},
+	"warehouse_size": {
+		Type:             schema.TypeString,
+		Optional:         true,
+		ValidateDiagFunc: sdkValidation(sdk.ToWarehouseSize),
+		DiffSuppressFunc: SuppressIfAny(NormalizeAndCompare(sdk.ToWarehouseSize), IgnoreChangeToCurrentSnowflakeValueInShow("size")),
+		Description:      fmt.Sprintf("Specifies the size of the interactive warehouse. Valid values are (case-insensitive): %s. Note: removing the size from config will result in the resource recreation.", possibleValuesListed(sdk.ValidWarehouseSizesString)),
```
sfc-gh-kwasilewski: Let's adjust the description - changing the size also recreates a warehouse.

pkg/resources/warehouse_interactive.go
sfc-gh-kwasilewski: > Today, if you leave it unspecified in a terraform config, the provider will ALTER WAREHOUSE ... UNSET AUTO_SUSPEND, which sets it to NULL

Yes, but only if you update a warehouse that previously had auto_suspend specified in the config. If you create a new warehouse without auto_suspend in the config, you'll end up with the default value (86400).

Usually, the provider expects that unsetting a property (ALTER ... UNSET ...) reverts its value to the default value, and that's how it usually works in Snowflake. For warehouses (including interactive), it seems it works differently: ALTER ... UNSET ... sets auto_suspend to NULL, which is different than default.

I'm afraid this might be a source of frustration for customers using this resource, because they can get different results depending on the operation type (create or update).

(BTW, I wonder why tests didn't detect that)

If we want to get this right, in Update we need to do exactly the same thing as in standard warehouse: https://github.com/snowflakedb/terraform-provider-snowflake/blob/abf4a892302e082acf019bbf5a1a106cbec1c3c2/pkg/resources/warehouse.go#L590-L598

pkg/testacc/resource_warehouse_interactive_acceptance_test.go
sfc-gh-uvarma: > so let's just set it on the interactive warehouse level for each test involving running any alters

That's the problem: [you can't](https://docs.snowflake.com/en/sql-reference/sql/create-interactive-warehouse#:~:text=Interactive%20warehouses%20have%20a%20maximum%20timeout%20interval%20of%20five%20seconds.): 

> Interactive warehouses have a maximum timeout interval of five seconds. Any larger values are ignored.

So, it's not like we can modify our tests so that we're only testing with larger timeouts.

> that's a pretty poor experience even through SQL if I create a warehouse and then can't even alter it by default because I hit the timeout on ALTER

Yes, I definitely agree.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Thanks for calling this out. I've updated this to match the regular warehouse's behavior.

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Artur's suggestions help but don't eliminate the problem:

 * by using a separate provider, we avoid causing problems in other tests, but the interactive warehouse tests are still flaky.
 * we cannot set the statement timeout to be greater than 5 seconds on interactive warehouses

pkg/resources/warehouse_interactive.go
sfc-gh-uvarma: Fixed. Also, I'm no longer recreating: I'm suspending during ALTER and then resuming (if it was running at the start).

pkg/resources/warehouse_interactive.go
sfc-gh-kwasilewski: As I stated elsewhere, ALTER statements don't appear to suffer from timeout errors on real prod accounts. I think we've covered everything needed here.

Thanks for switching to a separate provider.

MIGRATION_GUIDE.md
```diff
@@ -26,6 +26,24 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
...
+
+This feature will be marked as stable in future releases. To use it, add `snowflake_warehouse_interactive_resource` to the `preview_features_enabled` field in the provider configuration.
+
+Interactive warehouses behave differently from standard warehouses in a few ways that are described in the [resource documentation](https://registry.terraform.io/providers/snowflakedb/snowflake/latest/docs/resources/warehouse_interactive).
+
+The `show_output` field in the `snowflake_warehouses` data source now includes an additional computed `tables` attribute surfacing interactive warehouse table associations.
+
+No changes are required for existing configurations unless you want to adopt this preview feature with Terraform.
+
+### *(new feature)* `issuer` added to `default_workload_identity.aws` on `snowflake_service_user` and `snowflake_legacy_service_user`
```
sfc-gh-kwasilewski: I think this might have been duplicated by mistake

sfc-gh-uvarma: Sorry that I missed this. Thanks for catching! Fixed.

sfc-gh-kwasilewski (approved):

PR #5015, opened 2026-07-15
chore: Regenerate object assertions cleanly (part 4)

sfc-gh-asawicki:
Continuing the effort to make `make generate-snowflake-object-assertions` idempotent — this part adds the remaining generator capabilities needed to eliminate `// adjusted manually` patches from all touched objects.                           
                                                                                                                                                                                                                                                  
  **New generator flags:**                                                                                                                                                                                                                          
  - `NoIdentifiableObject` — for list-returned describe results with no meaningful identifier (`IcebergTableDetails`, `TableSearchOptimizationDetails`); generates `FromObject` with placeholder but no constructor and no bare `New*Assert()`.   
  Cleanly separates from `IsSubStruct` (true embedded fields) and `NoShowById` (standalone objects without simple Show).                                                                                                                            
  - `ShowByParentIdDef` — generates a parent-ID constructor for objects that require a parent to be listed (`ProgrammaticAccessToken`, `UserWorkloadIdentityAuthenticationMethod`); groups `ParentIdType`, `ClientName`, `ShowMethodName` together
  as they are only meaningful in combination.                                                                                                                                                                                                       
  - `DescribeOverrideDef` — overrides the default `testClient.<Name>.Describe` naming in `IsDataSourceOutput` constructors when the actual helper doesn't follow the convention (catalog integration variants).                                   
  - `NestedAssertFields` / `SkipFields` — `NestedAssertFields` generates nested assertion adapters instead of simple `Has*`; `SkipFields` fully suppresses generation for fields with richer hand-written ext implementations (e.g. `TaskRelations` 
  uses order-insensitive `ElementsMatch`).                                                                                                                                                                                                          
                                                                                                                                                                                                                                                    
  **Auto-detected comparison method (`ComparisonMethod`):**                                                                                                                                                                                         
  Generator previously always emitted `!=`, causing compile errors for struct types and wrong results for `datatypes.DataType`. Now detects `reflect.DeepEqual` for structs, `datatypes.AreTheSame` for `datatypes.DataType`, and `!=` for        
  everything else. Adds `ToSql` mapper so `DataType` error messages print human-readable SQL type names.                                                                                                                                            
                                                                                                                                                                                                                                                  
  **Nested assertion adapter pattern:**                                                                                                                                                                                                             
  - Non-pointer fields: `Has<Field>With(subAssert)` — iterates sub-asserter's assertions as independent entries in the parent list (no early exit).                                                                                               
  - Pointer fields: `Has<Field>()` nil check + `Has<Field>With(subAssert)` that delegates to the nil check method to avoid duplication.                                                                                                             
  - Slice fields: variadic `Has<Field>With(subAsserts ...)` with strict count validation; each sub-asserter's assertions are independent entries with index bounds guard.                                                                           
  - All variants wrap errors with the field path (and slice index) so failing assertions are immediately traceable without counting assertion indices.                                                                                              
  - `AssertionOnPointerField[F]` helper centralises nil-check-then-assert for pointer fields.                                                                                                                                                       
                                                                                                                                                                                                                                                    
  **`SubStructAssert[T]`:**                                                                                                                                                                                                                         
  Sub-struct asserters previously embedded `SnowflakeObjectAssert[T, I]`, accidentally satisfying `VerifyAll`/`ToTerraformTestCheckFunc` and allowing standalone misuse. The new type exposes only `AddAssertion` and `GetAssertions` — standalone  
  use is now a compile error. `GetAssertions()` and `NewSnowflakeObjectAssertEmpty` are removed from `SnowflakeObjectAssert` as they are no longer needed there.                                                                                    
                                                                                                                                                                                                                                                  
  **SDK changes:**                                                                                                                                                                                                                                  
  - `CatalogIntegrationOpenCatalogDetails.ID()` and `CatalogIntegrationIcebergRestDetails.ID()` — needed for generated `FromObject`.                                                                                                              
  - `ExternalVolumeDetails`: adds `Id AccountObjectIdentifier` field, `ID()` method, and `DescribeDetails` (parses + sets Id from caller's identifier in one call, consistent with `PasswordPolicies`/`StorageLifecyclePolicies` pattern).          
                                                                                                                                                                                                                                                    
  **Test migrations:**                                                                                                                                                                                                                              
  - Catalog integration tests: two-call `FromObject` pattern replaced with chained `HasRestConfigWith`/`HasRestAuthenticationWith`/`HasSigV4RestAuthenticationWith` etc.                                                                            
  - ExternalVolume tests: per-index `FromObject` calls replaced with `HasStorageLocationsWith(...)` including nested pointer sub-field assertions.                                                                                                  
  - WIF gen file renamed from wrongly-named plural (`_methods_`) to correct singular; `ShowUserWorkloadIdentityAuthenticationMethod` helper added.


pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
```diff
@@ -316,6 +355,12 @@ var allStructs = []SdkObjectDef{
 		IsDataSourceOutput: true,
 		IsSubStruct:        true,
 	},
+	{
+		IdType:             "sdk.AccountObjectIdentifier",
+		ObjectStruct:       sdk.BearerRestAuthenticationDetails{},
```
sfc-gh-kwasilewski: The only field this struct has is bearer_token, but since it's a secret, we can't assert it reliably. Not needed.

pkg/acceptance/bettertestspoc/assert/objectassert/cortex_agent_snowflake_gen.go
```diff
@@ -101,7 +102,7 @@ func (c *CortexAgentAssert) HasComment(expected string) *CortexAgentAssert {
 func (c *CortexAgentAssert) HasProfile(expected sdk.CortexAgentProfile) *CortexAgentAssert {
 	c.AddAssertion(func(t *testing.T, o *sdk.CortexAgent) error {
 		t.Helper()
-		if o.Profile != expected {
+		if !reflect.DeepEqual(o.Profile, expected) {
```
sfc-gh-kwasilewski, line 105: It'd nice to scan through _ext.go files and remove all unneeded extra assertions that were added due to this issue, like here (there's HasCortexAgentProfile in the _ext.go that now is the same as the generated version).

pkg/acceptance/bettertestspoc/assert/objectassert/gen/sdk_object_def.go
sfc-gh-asawicki: Removed the dedicated assert and added a check in ext.

pkg/acceptance/bettertestspoc/assert/objectassert/cortex_agent_snowflake_gen.go
sfc-gh-asawicki: As there are 2 more PRs incoming, ending with a full clean regeneration, I will do it as a continuation (it'll be easier to do when we have a clean state that can be fully regeneratable).

sfc-gh-kwasilewski (approved):

PR #5026, opened 2026-07-17
feat: Detect external drift for inherited account role grants

sfc-gh-kwasilewski:
Read inherited grants back from SHOW GRANTS TO ROLE and match them by object type and container using the inherited-specific columns (is_inherited, inherited_from, inherited_from_database, inherited_from_schema).

Add unit tests for the drift-detection logic and acceptance tests for non-strict re-granting and strict-mode detection of externally added privileges.


pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -1195,9 +1199,9 @@ func computePrivileges(id GrantPrivilegesToAccountRoleId, grants []sdk.Grant, gr
 		// the default application named SNOWFLAKE that could be granted with `object_type = "APPLICATION"`.
 		// To make the logic simpler, we do not allow it and `object_type = "DATABASE"` should be used for all applications.
 		// TODO When implementing SNOW-991421 see if logic added in SNOW-887897 could be moved to the SDK to simplify the resource implementation.
-		if grantedOn == sdk.ObjectTypeDatabase && (sdk.ObjectTypeApplication == grant.GrantedOn || sdk.ObjectTypeApplication == grant.GrantOn) {
+		if *grantedOn == sdk.ObjectTypeDatabase && (sdk.ObjectTypeApplication == grant.GrantedOn || sdk.ObjectTypeApplication == grant.GrantOn) {
```
sfc-gh-asawicki, line 1202: q: I haven't analyzed all the input paths but shouldn't we have a nil-check somewhere on the line?

sfc-gh-asawicki (approved):

pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -1213,15 +1217,86 @@ func computePrivileges(id GrantPrivilegesToAccountRoleId, grants []sdk.Grant, gr
...
+		if !grant.IsInherited {
+			continue
+		}
+		if (grant.GrantTo != sdk.ObjectTypeRole && grant.GrantedTo != sdk.ObjectTypeRole) || grant.GranteeName.Name() != id.RoleName.Name() {
+			continue
+		}
+		if grant.GrantedOn != grantedOn || !inheritedAccountRoleGrantMatchesContainer(grant, container, database, schema) {
+			continue
+		}
+		if !strictPrivilegeManagement && !slices.Contains(id.Privileges, grant.Privilege) {
```
sfc-gh-jmichalak, line 1236: Nit: all these conditions could be done in one if.

pkg/resources/grant_privileges_to_account_role.go
```diff
@@ -1213,15 +1217,86 @@ func computePrivileges(id GrantPrivilegesToAccountRoleId, grants []sdk.Grant, gr
...
+			trimIdentifierQuotes(*grant.InheritedFromDatabase) == schema.DatabaseName() &&
+			trimIdentifierQuotes(*grant.InheritedFromSchema) == schema.Name()
+	default:
+		return false
+	}
+}
+
+// trimIdentifierQuotes strips the surrounding double quotes Snowflake may add to the inherited_from_*
+// identifier columns, so the values can be compared against the parsed identifier names.
+func trimIdentifierQuotes(s string) string {
```
sfc-gh-jmichalak, line 1286: Nit: Shouldn't it be handled by the already existing identifier parsers?

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-kwasilewski: The only scenario where this is nil is for inherited grants. However, those are handled early by `computeInheritedAccountRolePrivileges`, which returns before reaching this line.

pkg/testacc/resource_grant_privileges_to_account_role_acceptance_test.go
```diff
@@ -3044,6 +3044,171 @@ func TestAcc_GrantPrivilegesToAccountRole_StrictRoleManagement_BasicOnCreate(t *
...
+						HasAlwaysApply(false).
+						HasStrictPrivilegeManagement(true),
+					assert.Check(queriedAccountRolePrivilegesEqualTo(t, roleId, string(configuredPrivilege))),
+				),
+			},
+		},
+	})
+}
+
+func TestAcc_GrantPrivilegesToAccountRole_OnAccountObject_Inherited_ExternalDrift(t *testing.T) {
```
sfc-gh-jmichalak, line 3139: Nit: This could be a part of TestAcc_GrantPrivilegesToAccountRole_OnAccountObject_Inherited

sfc-gh-jmichalak (approved):

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-kwasilewski: Yes, this will be rewritten in a follow-up PR.

pkg/resources/grant_privileges_to_account_role.go
sfc-gh-kwasilewski: It will look even worse if I collapse all these if statements into one big blob, plus I wanted this function to look similar to the existing `computePrivileges` that already has multiple if statements.

PR #5031, opened 2026-07-20
chore: Add missing file formats integration tests

sfc-gh-jmichalak:
## Summary
- Add `DISABLE_SNOWFLAKE_DATA` to the XML file format CREATE/ALTER options, describe struct, and describe parser — found missing from the SDK by comparing `DESCRIBE FILE FORMAT` output against the SDK for every file format type.
- Convert the manually maintained `FileFormatCsv/Json/Avro/Orc/Parquet/Xml/AllDetails` object assertions into fully generated ones by wiring `DescribeOverride` in the object assertion generator definitions, removing the hand-written `_ext.go` assertion files.
- Update `FileFormatClient` describe helpers to return `(*T, error)` to match the generator's expected provider signature.
- Expand file format integration tests so every SDK field is exercised in at least one CREATE and one ALTER test, assert all available `FileFormat` fields (adding a `HasCreatedOnNotEmpty` extension), and use the generated Describe-backed assert constructors directly instead of `FromObject` + manual Describe calls.


pkg/sdk/testint/file_formats_gen_integration_test.go
```diff
@@ -0,0 +1,1103 @@
...
+			HasCreatedOnNotEmpty().
+			HasName(id.Name()).
+			HasDatabaseName(id.DatabaseName()).
+			HasSchemaName(id.SchemaName()).
+			HasType(sdk.FileFormatTypeCsv).
+			HasOwner(snowflakeroles.Accountadmin.Name()).
+			HasOwnerRoleType("ROLE").
+			HasComment(""))
+
+		assertThatObject(t, objectassert.FileFormatCsv(t, id).
```
sfc-gh-jcieslak, line 37: Hmm, I think for "parameter-based" describe outputs where default is returned, we could somehow use it to generate HasAllDefaults assertions. Just an idea.

pkg/sdk/file_formats_ext_test.go
```diff
@@ -6,7 +6,244 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestParseNullIfProperty(t *testing.T) {
-	require.Equal(t, []NullString{{"a"}, {"b"}}, parseNullIfProperty("[a, b]"))
-	require.Equal(t, []NullString{}, parseNullIfProperty("[]"))
+func TestParseFileFormatCsv(t *testing.T) {
+	id := randomSchemaObjectIdentifier()
+	properties := []FileFormatProperty{
```
sfc-gh-jcieslak, line 11: not all mapping paths are tested; applies to other types (ik they're tested in integration tests, but maybe let's have them here too, they're fast and it's cheap to add missing cases)

pkg/sdk/file_formats_gen_test.go
```diff
@@ -582,12 +582,12 @@ func TestFileFormats_AlterOrc(t *testing.T) {
 
 	t.Run("all options", func(t *testing.T) {
 		opts := defaultOpts()
-		opts.IfExists = new(true)
+		opts.IfExists = Bool(true)
```
sfc-gh-jcieslak: nit: I think we would like to ultimately move to `new` instead of custom pointer wrappers, so I would limit such conversions

pkg/sdk/testint/file_formats_gen_integration_test.go
sfc-gh-jmichalak: Well, we could do this even for SHOW/DESC objects. I'll leave this as a future improvement.

pkg/sdk/file_formats_gen_test.go
sfc-gh-jmichalak: Oops, it was claude's idea. Reverted.

pkg/sdk/file_formats_ext_test.go
sfc-gh-jmichalak: Done.

sfc-gh-jcieslak (approved):

PR #5037, opened 2026-07-21
feat: Expose all warehouse parameters in snowflake_warehouses data source

sfc-gh-kwasilewski:
The snowflake_warehouses data source now uses a single unified parameters schema covering every warehouse type and populates the FALLBACK_WAREHOUSE parameter whenever SHOW PARAMETERS returns it.

- Add ShowAllWarehouseParametersSchema and AllWarehouseParametersToSchema and use them in the data source.
- Generate WarehousesDatasourceParameters from the WarehouseInteractive parameter definition so the datasource parameters assert covers all parameters, including fallback_warehouse.
- Expand the data source acceptance tests for standard, adaptive, and interactive warehouses.


pkg/acceptance/bettertestspoc/assert/resourceparametersassert/gen/model.go
```diff
@@ -25,10 +25,10 @@ type ResourceParameterAssertionModel struct {
...
 
 var dataSourceParametersMapping = map[string]string{
-	"Database":  "Databases",
-	"Task":      "Tasks",
-	"User":      "Users",
-	"Warehouse": "Warehouses",
+	"Database":             "Databases",
+	"Task":                 "Tasks",
+	"User":                 "Users",
+	"WarehouseInteractive": "Warehouses",
```
sfc-gh-jcieslak, line 31: Q: Is there a reason we remove regular Warehouse case?

sfc-gh-jcieslak (approved):

pkg/acceptance/bettertestspoc/assert/resourceparametersassert/gen/model.go
sfc-gh-kwasilewski: Yes, the regular Warehouse doesn't define `FALLBACK_WAREHOUSE` parameter. Only WarehouseInteractive has all warehouse parameters. So if we want to include all warehouse parameters in the data source, we have to use WarehouseInteractive in this mapping.

PR #5047, opened 2026-07-22
feat: Add inherited grants support to grant_privileges_to_database_role

sfc-gh-kwasilewski:
Support GRANT INHERITED on all current and future schemas and schema objects (in a database or a schema) for the
snowflake_grant_privileges_to_database_role resource, gated behind the INHERITED_GRANTS experiment. Shared inherited-grant logic used by both the account role and database role resources is extracted into grant_privileges_commons.go.


pkg/resources/grant_privileges_commons.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/provider/experimentalfeatures"
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// inheritedGrantsRequireExperiment returns a CustomizeDiff that fails the plan when an `inherited`
+// block is used in any of the given blocks while the INHERITED_GRANTS experiment is not enabled.
+func inheritedGrantsRequireExperiment(blockNames ...string) schema.CustomizeDiffFunc {
```
sfc-gh-kwasilewski, line 18: These are the functions moved from `grant_privileges_to_account_role.go`:

* `inheritedGrantsRequireExperiment` -> `inheritedGrantsRequireExperiment` (now a factory taking blockNames ...string)
* `rawConfigHasInheritedGrant` (now takes blockNames ...string)
* `computeInheritedAccountRolePrivileges` -> `computeInheritedPrivileges`
* `inheritedAccountRoleGrantScope` -> `inheritedGrantScopeFromData`
* `inheritedAccountRoleGrantMatchesContainer` -> `inheritedGrantMatchesContainer`
* `getInheritedGrantContainer` (unchanged)

pkg/resources/grant_privileges_commons_test.go
```diff
@@ -0,0 +1,237 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/Snowflake-Labs/terraform-provider-snowflake/pkg/sdk"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestComputeInheritedPrivileges(t *testing.T) {
```
sfc-gh-kwasilewski, line 11: This isn't a new test, it's a test from the renamed `grant_privileges_to_account_role_test.go` file with two new test cases.

docs/resources/grant_privileges_to_database_role.md
```diff
@@ -210,6 +236,7 @@ Optional:
 
 - `all_schemas_in_database` (String) The fully qualified name of the database.
 - `future_schemas_in_database` (String) The fully qualified name of the database.
+- `inherited` (String) Configures an inherited privilege to be granted on all current and future schemas in a database. See [Inherited grants](https://docs.snowflake.com/en/user-guide/inherited-grants-using) for more details.
```
sfc-gh-jcieslak, line 239: I think we could also update migration guide, always_apply field documentation, and maybe some docs in guides directory to inform customers that better alternative is there instead of having on_all + on_future combo. There are also a few issues that are about having one grant resource where both could be set up, I think after releasing inherited grants we could inform the users there and even perhaps close those threads as setting both would be then possible (just through different path that is more Snowflake native).

pkg/resources/grant_privileges_commons.go
```diff
@@ -0,0 +1,179 @@
...
+			resp.Diagnostics = append(resp.Diagnostics, diag.Diagnostic{
+				Severity: diag.Error,
+				Summary:  "Invalid with_grant_option",
+				Detail:   "`with_grant_option` cannot be used together with an `inherited` block, because inherited grants do not support the WITH GRANT OPTION clause.",
+			})
+		}
+		// always_apply re-grants the configured privileges on every apply. Inherited grants already
+		// cover all current and future objects in the container, so re-granting serves no purpose and
+		// is disallowed to keep the resource behavior clear.
+		if alwaysApply := req.RawConfig.GetAttr("always_apply"); !alwaysApply.IsNull() && alwaysApply.IsKnown() && alwaysApply.True() {
```
sfc-gh-jcieslak, line 53: I'm also wondering if this is needed after inherited grants (maybe should be deprecated and planned for removal in v3?)

pkg/resources/grant_privileges_commons.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+		actualPrivileges = append(actualPrivileges, grant.Privilege)
+	}
+
+	return actualPrivileges
+}
+
+// inheritedGrantScopeFromData returns the object type and container an inherited grant's identifier data
+// is scoped to, used to match the rows returned by SHOW GRANTS TO [DATABASE] ROLE.
+func inheritedGrantScopeFromData(data fmt.Stringer) (grantedOn sdk.ObjectType, container InheritedContainerKind, database *sdk.AccountObjectIdentifier, schema *sdk.DatabaseObjectIdentifier) {
```
sfc-gh-jcieslak, line 121: nit: Shouldn't data be just `any`?

sfc-gh-jmichalak (approved):

pkg/testacc/resource_grant_privileges_to_database_role_acceptance_test.go
```diff
@@ -676,6 +681,375 @@ func TestAcc_GrantPrivilegesToDatabaseRole_OnSchemaObject_OnFunctionWithoutArgum
...
+				},
+				Check: assertThat(
+					t,
+					resourceassert.GrantPrivilegesToDatabaseRoleResource(t, ref).
+						HasDatabaseRoleName(roleId.FullyQualifiedName()).
+						HasPrivileges(privilege).
+						HasAllPrivileges(false).
+						HasWithGrantOption(false).
+						HasAlwaysApply(false),
+					assert.Check(resource.TestCheckResourceAttr(ref, "on_schema.0.inherited", databaseId.FullyQualifiedName())),
```
sfc-gh-jcieslak, line 740: Why not through the re-generated assertions?

docs/resources/grant_privileges_to_database_role.md
sfc-gh-kwasilewski: Let me do that as a follow-up. Do you have links to those issues?

pkg/resources/grant_privileges_commons.go
sfc-gh-kwasilewski: It's a good idea, but we should do that once we remove or turn on by default the INHERITED_GRANTS experiment.

pkg/resources/grant_privileges_commons.go
sfc-gh-kwasilewski: They could, but `fmt.Stringer` is also valid because all these types have a String method.

pkg/testacc/resource_grant_privileges_to_database_role_acceptance_test.go
sfc-gh-kwasilewski: We don't have them because the generator doesn't support blocks, we'd need to manually add them to _ext files.

docs/resources/grant_privileges_to_database_role.md
sfc-gh-jcieslak: Generally they should be here: https://github.com/snowflakedb/terraform-provider-snowflake/issues?q=is%3Aissue%20state%3Aopen%20label%3Acategory%3Agrants

I remember this one: https://github.com/snowflakedb/terraform-provider-snowflake/issues/1993
maybe this also applies: https://github.com/snowflakedb/terraform-provider-snowflake/issues/2668

pkg/resources/grant_privileges_commons.go
sfc-gh-jcieslak: Yup, let's just note that somewhere in v3 doc as another point to discuss / do

pkg/testacc/resource_grant_privileges_to_database_role_acceptance_test.go
sfc-gh-jcieslak: We generally do in such cases, but since it's a single case idc that much

sfc-gh-jcieslak (approved):

Issue #4415, opened 2026-02-05
[Feature]: Missing feature, snowflake intelligence support


jakerohald:
### Company Name

Xero

### Use Cases or Problem Statement

The terraform resources required for snowflake intelligence support are not currently supported.

Requirements as per snowflake documentation: https://docs.snowflake.com/en/user-guide/snowflake-cortex/snowflake-intelligence#grant-sf-intelligence-privileges

Resource 1: Creating Snowflake Intelligence object 

(something like)

```
resource "snowflake_intelligence" "si_object" {
  name = "SNOWFLAKE_INTELLIGENCE"
}

```
Resource 2/3: Grant privileges

```
resource "snowflake_grant_privileges_to_account_role" "snowflake_intelligence_modify_privileges" {
  with_grant_option = false
  privileges        = ["MODIFY"]
  account_role_name = "ROLE_NAME"
  on_schema_object {
    object_type = "SNOWFLAKE INTELLIGENCE"
    object_name = "SNOWFLAKE_INTELLIGENCE"
  }
}
```

```
resource "snowflake_grant_privileges_to_account_role" "snowflake_intelligence_usage_grant" {
  with_grant_option = false
  privileges        = ["USAGE"]
  account_role_name = "ROLE_NAME"
  on_schema_object {
    object_type = "SNOWFLAKE INTELLIGENCE"
    object_name = "SNOWFLAKE_INTELLIGENCE"
  }
}

```
### Category

category:resource

### How much impact is this issue causing?

Medium

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-jmichalak:
Hi @jakerohald 👋 

Thanks for reporting this. This is currently not on our roadmap, but I added your request to our internal list. Please also approach your Account Manager to help prioritize this.

Note that the SQL command reference documentation seems to be missing for this object, so I also requested it.

PR #5049, opened 2026-07-22
feat: Support SNOWFLAKE INTELLIGENCE object type in grants

sfc-gh-jmichalak:
## Summary
- `SNOWFLAKE INTELLIGENCE` is an account-level object and cannot be granted to database roles (verified manually), so `snowflake_grant_privileges_to_database_role` is intentionally unchanged.
- Manually verified with the Snowflake CLI against a live test account that `GRANT`/`REVOKE MODIFY ON SNOWFLAKE INTELLIGENCE` and `GRANT OWNERSHIP ON SNOWFLAKE INTELLIGENCE` work as expected (the latter isn't yet reflected in the public `GRANT OWNERSHIP` docs page).
- Adds acceptance tests for both resources.
- Adds a `MIGRATION_GUIDE.md` entry and regenerates resource docs.
- This is now missing in docs, but this feature has been requested multiple times. I've asked to update Snowflake docs.

Partially addresses #4415 — that issue also requests a dedicated `snowflake_intelligence`-style resource for creating the Snowflake Intelligence object itself, which is not part of this PR.


sfc-gh-asawicki (dismissed):

pkg/sdk/grants_validations.go
```diff
@@ -239,14 +238,18 @@ var invalidGrantToFutureObjectTypes = []ObjectType{
 	ObjectTypeStorageLifecyclePolicy,
 	ObjectTypeTag,
 	ObjectTypeWarehouse,
-	ObjectTypeWorkspace,
+}
+
+// A list of object types that are not supported for grant to all or future.
+var invalidGrantToPluralObjectTypes = []ObjectType{
```
sfc-gh-asawicki, line 244: nit: we already have invalidGrantToAllObjectTypes and invalidGrantToFutureObjectTypes (below for schema objects). Maybe we could use these lists instead of adding a new one?

sfc-gh-asawicki (approved):

pkg/sdk/grants_validations.go
sfc-gh-jmichalak: We could but they separate ON ALL and ON FUTURE, and we don't need this separation for now. Additionally, they have ObjectTypeWarehouse for some reason, so I'd say this needs an overall cleanup.

PR #5053, opened 2026-07-23
feat: Add AUTHENTICATOR_EXPLICIT_ONLY experiment to disable implicit authenticator derivation

sfc-gh-asawicki:
## Changes

The provider currently auto-derives `authenticator = OAUTH` when `token` or `token_accessor` is configured without an explicit `authenticator` field. This implicit behavior is confusing and scheduled for removal in v3. The new experiment lets users opt into the v3 behavior early.

- Add `AUTHENTICATOR_EXPLICIT_ONLY` experimental feature constant and registration
- Gate the token→OAuth derivation in `ConfigureProvider` with experiment check
- Gate the token_accessor→OAuth derivation in `getDriverConfigFromTerraform` with experiment check
- Read `experimental_features_enabled` early in `ConfigureProvider` (before derivation runs) and reuse it for `providerCtx.EnabledExperiments`
- Add acceptance test proving the experiment prevents OAuth derivation (expects password error instead of OAuth token error)
- Add migration guide entry with upgrade instructions
- Regenerate docs

When the experiment is enabled, setting `token` or `token_accessor` without `authenticator` falls through to the `SNOWFLAKE` default instead of implicitly selecting `OAUTH`. Users must add `authenticator = "OAUTH"` explicitly.

## References

- Preparation for v3 authenticator behavior change


pkg/testacc/provider_acceptance_test.go
```diff
@@ -944,6 +945,38 @@ func TestAcc_Provider_testFixedDefaultAuthenticatorForToken(t *testing.T) {
...
+			t.Setenv(snowflakeenvs.ConfigPath, random.String())
+		},
+		ProtoV6ProviderFactories: providerFactoryWithoutCache(),
+		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
+			tfversion.RequireAbove(tfversion.Version1_5_0),
+		},
+		Steps: []resource.TestStep{
+			{
+				Config:      providerConfig,
+				ExpectError: regexp.MustCompile("260002: password is empty"),
```
sfc-gh-jmichalak, line 974: Nit: here we can assert the actual values used in the driver config - check other tests.

sfc-gh-jmichalak (approved):

pkg/testacc/provider_acceptance_test.go
sfc-gh-asawicki: Tried it:
- `Check` is not run when `ExpectError` is present
- if we drop `ExpectError` and leave `Check` only - the test fails due to connection issues

Leaving `ExpectError` only

PR #5060, opened 2026-07-24
chore: Ignore show_output.0.running in warehouse adaptive import tests

sfc-gh-kwasilewski:
The running field is non-deterministic (reflects active query count at
read time) and can differ between plan and import, causing flaky import
verification failures. Ignore it in ImportStateVerifyIgnore for all
import steps across TestAcc_WarehouseAdaptive_BasicUseCase and
TestAcc_WarehouseAdaptive_CompleteUseCase.

### DIFF
--- a/pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
+++ b/pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
@@ -132,7 +132,7 @@ func TestAcc_WarehouseAdaptive_BasicUseCase(t *testing.T) {
 				ResourceName:            warehouseModel.ResourceReference(),
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"max_query_performance_level", "query_throughput_multiplier"},
+				ImportStateVerifyIgnore: []string{"max_query_performance_level", "query_throughput_multiplier", "show_output.0.running"},
 			},
 			// set query_throughput_multiplier to 0 (explicit value, distinct from IntDefault sentinel -1)
 			{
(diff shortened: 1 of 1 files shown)


pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
```diff
@@ -261,9 +262,10 @@ func TestAcc_WarehouseAdaptive_CompleteUseCase(t *testing.T) {
 			},
 			// import and verify state matches
 			{
-				ResourceName:      warehouseModelComplete.ResourceReference(),
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            warehouseModelComplete.ResourceReference(),
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"show_output.0.running"},
```
sfc-gh-jcieslak, line 268: Question: is `running` only in adaptive warehouses or other warehouses also have this in their show_output? If so, maybe we should add ignore there too?

sfc-gh-jcieslak (approved):

pkg/testacc/resource_warehouse_adaptive_acceptance_test.go
sfc-gh-kwasilewski: All warehouses have this field in their show_output.

I don't know why only adaptive tests are failing. It’s possible there’s something specific to these tests causing `running` to be non-zero.

For now, let's keep the fix scoped to adaptive warehouses only. If we see failures in other tests, we can expand the fix.

PR #5062, opened 2026-07-24
feat: add SDK definition for ExternalAccessIntegration

sfc-gh-pcichon:
## Changes
- Add definition file and generate SDK code for External Access Integration.


pkg/sdk/generator/defs/external_access_integrations_def.go
```diff
@@ -0,0 +1,123 @@
...
+	"https://docs.snowflake.com/en/sql-reference/sql/alter-external-access-integration",
+	g.NewQueryStruct("AlterExternalAccessIntegration").
+		Alter().
+		SQL("EXTERNAL ACCESS INTEGRATION").
+		IfExists().
+		Name().
+		OptionalQueryStructField(
+			"Set",
+			g.NewQueryStruct("ExternalAccessIntegrationSet").
+				ListAssignment("ALLOWED_NETWORK_RULES", g.KindOfT[sdkcommons.SchemaObjectIdentifier](), g.ParameterOptions().Parentheses()).
```
sfc-gh-jcieslak, line 60: Question: Docs seem strange https://docs.snowflake.com/en/sql-reference/sql/alter-external-access-integration did you check if at least one allowed_network_rule is needed? The syntax `ALLOWED_NETWORK_RULES = (<rule_name> [ , <rule_name> ... ])` looks like at least one item should be there.

pkg/sdk/external_access_integrations_gen_test.go
```diff
@@ -0,0 +1,294 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, ErrNilOptions)
+	})
+
+	t.Run("validation: valid identifier for [opts.name]", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.name = emptyAccountObjectIdentifier
+		assertOptsInvalidJoinedErrors(t, opts, ErrInvalidObjectIdentifier)
+	})
+
+	t.Run("validation: exactly one field from [opts.AllowedApiAuthenticationIntegrations.None opts.AllowedApiAuthenticationIntegrations.Integrations] should be present", func(t *testing.T) {
```
sfc-gh-kwasilewski, line 39: The generator doesn't provide a template for that, but we usually add a variant where more than one fields are present.

pkg/sdk/generator/defs/external_access_integrations_def.go
```diff
@@ -0,0 +1,123 @@
...
+					g.KeywordOptions().SQL("ALLOWED_API_AUTHENTICATION_INTEGRATIONS ="),
+				).
+				OptionalQueryStructField(
+					"AllowedAuthenticationSecrets",
+					allowedAuthenticationSecrets(),
+					g.KeywordOptions().SQL("ALLOWED_AUTHENTICATION_SECRETS ="),
+				).
+				OptionalBooleanAssignment("ENABLED", g.ParameterOptions()).
+				OptionalComment().
+				PredefinedQueryStructField("SetTags", "[]TagAssociation", g.KeywordOptions().SQL("TAG")).
```
sfc-gh-kwasilewski: Use `OptionalSetTags()` on the `AlterExternalAccessIntegration` level, next to "Set" and "Unset". It's not only about consistency (every entity in the provider use this helper), but also about correctness - looks like ALTER fails if you try to set both a tag and something else, like comment, in the same statement.

pkg/sdk/generator/defs/external_access_integrations_def.go
```diff
@@ -0,0 +1,123 @@
...
+			g.KeywordOptions().SQL("SET"),
+		).
+		OptionalQueryStructField(
+			"Unset",
+			g.NewQueryStruct("ExternalAccessIntegrationUnset").
+				OptionalSQL("ALLOWED_NETWORK_RULES").
+				OptionalSQL("ALLOWED_API_AUTHENTICATION_INTEGRATIONS").
+				OptionalSQL("ALLOWED_AUTHENTICATION_SECRETS").
+				OptionalSQL("COMMENT").
+				PredefinedQueryStructField("UnsetTags", "[]ObjectIdentifier", g.KeywordOptions().SQL("TAG")).
```
sfc-gh-kwasilewski: Use `OptionalUnsetTags`

pkg/sdk/generator/defs/external_access_integrations_def.go
sfc-gh-jcieslak: Same in ALLOWED_API_AUTHENTICATION_INTEGRATIONS and ALLOWED_AUTHENTICATION_SECRETS

sfc-gh-jcieslak: This is usually not a concern in our resources as we fill those with at least one value during the Update, but we can discuss if we want to do something with this fact (if really at least one value is required)

pkg/sdk/generator/defs/external_access_integrations_def.go
sfc-gh-jcieslak: nit: I think we have predefined method for this (`OptionalSetTags`?)

sfc-gh-jcieslak: same for UnsetTags

pkg/sdk/generator/defs/external_access_integrations_def.go
```diff
@@ -0,0 +1,123 @@
...
+		Text("name").
+		Text("type").
+		Text("category").
+		Bool("enabled").
+		OptionalText("comment", g.WithRequiredInPlain()),
+	g.NewQueryStruct("ShowExternalAccessIntegrations").
+		Show().
+		SQL("EXTERNAL ACCESS INTEGRATIONS").
+		OptionalLike(),
+).DescribeOperationWithPairedStructs(
```
sfc-gh-jcieslak, line 112: This is very generic describe, look at e.g. api integration, along side of this one we define externally implemented Describe methods that are using concrete structs instead of generic properties. This can be done either now or in a follow-up PR.

pkg/sdk/external_access_integrations_gen_test.go
sfc-gh-jcieslak, line 39: nit: with this assertion we usually do none set or two set to make sure both are validated correct (one set is correct and would be tested below in valid cases)

pkg/sdk/external_access_integrations_gen_test.go
```diff
@@ -0,0 +1,294 @@
...
+		assertOptsValidAndSQLEquals(t, opts, `CREATE EXTERNAL ACCESS INTEGRATION %s ALLOWED_NETWORK_RULES = (%s) ENABLED = true`, id.FullyQualifiedName(), networkRuleId.FullyQualifiedName())
+	})
+
+	t.Run("all options", func(t *testing.T) {
+		apiAuthId := randomAccountObjectIdentifier()
+		secretId := randomSchemaObjectIdentifier()
+		opts := defaultOpts()
+		opts.OrReplace = Bool(true)
+		opts.AllowedApiAuthenticationIntegrations = &ExternalAccessIntegrationAllowedApiAuthenticationIntegrations{
+			Integrations: []AccountObjectIdentifier{apiAuthId},
```
sfc-gh-jcieslak: for slices let's have at least two items (in other cases below too)

pkg/sdk/external_access_integrations_gen_test.go
```diff
@@ -0,0 +1,294 @@
...
+		assertOptsInvalidJoinedErrors(t, opts, errExactlyOneOf("AlterExternalAccessIntegrationOptions", "Set", "Unset"))
+	})
+
+	t.Run("validation: at least one of the fields [opts.Set.AllowedNetworkRules opts.Set.AllowedApiAuthenticationIntegrations opts.Set.AllowedAuthenticationSecrets opts.Set.Enabled opts.Set.Comment opts.Set.SetTags] should be set", func(t *testing.T) {
+		opts := defaultOpts()
+		opts.Set = &ExternalAccessIntegrationSet{}
+		assertOptsInvalidJoinedErrors(t, opts, errAtLeastOneOf("AlterExternalAccessIntegrationOptions.Set", "AllowedNetworkRules", "AllowedApiAuthenticationIntegrations", "AllowedAuthenticationSecrets", "Enabled", "Comment", "SetTags"))
+	})
+
+	t.Run("validation: exactly one field from [opts.Set.AllowedApiAuthenticationIntegrations.None opts.Set.AllowedApiAuthenticationIntegrations.Integrations] should be present", func(t *testing.T) {
```
sfc-gh-jcieslak, line 166: same here, look at other files with exactly one field validation tests, there are usually two for a given "exactly one of" validation

pkg/sdk/generator/defs/external_access_integrations_def.go
sfc-gh-jcieslak: Also set and unset tags usually cannot be used with other set / unset parameters (e.g. look at api integration, validation for alter). You can check manually if it's different here, or validations for alter need to be adjusted here.

pkg/sdk/generator/defs/external_access_integrations_def.go
sfc-gh-pcichon: ### Manual testing investigation
- `ALLOWED_NETWORK_RULES` needs to be present (it is a required filed) in the CREATE command and the list supplied needs to be non-empty
- `ALLOWED_API_AUTHENTICATION_INTEGRATIONS` and `ALLOWED_AUTHENTICATION_SECRETS` are do not need to be present (they are optional fields) in neither CREATE nor ALTER commands
  - in case they are being supplied, it is possible to supply an empty list `()`

### Conclusions (status: done)
_(after an offline discussion)_
- We should enforce non-empty `ALLOWED_NETWORK_RULES` to be present in the CREATE command
  - Added `WithValidation(g.AtLeastOneValueSet, "AllowedNetworkRules")` for that
- We should enforce non-empty `ALLOWED_NETWORK_RULES` to be present in the ALTER SET command unless it is not present at all
- We should not enforce `ALLOWED_API_AUTHENTICATION_INTEGRATIONS` or `ALLOWED_AUTHENTICATION_SECRETS` to be non-empty

pkg/sdk/generator/defs/external_access_integrations_def.go
sfc-gh-pcichon: You are right, made the suggested change. I was referencing the [Snowflake docs](https://docs.snowflake.com/en/sql-reference/sql/alter-external-access-integration) but it seems like they might not be accurate. After doing some manual testing, I can confirm that (un)setting tags and some other fields cannot be done at the same time, with one command.

sfc-gh-pcichon: Same as above

> You are right, made the suggested change. I was referencing the [Snowflake docs](https://docs.snowflake.com/en/sql-reference/sql/alter-external-access-integration) but it seems like they might not be accurate. After doing some manual testing, I can confirm that (un)setting tags and some other fields cannot be done at the same time, with one command.

pkg/sdk/generator/defs/external_access_integrations_def.go
sfc-gh-pcichon: Done

pkg/sdk/generator/defs/external_access_integrations_def.go
sfc-gh-pcichon: Good suggestion, I will add it in the following PRs.

pkg/sdk/external_access_integrations_gen_test.go
sfc-gh-pcichon: Added validation unit tests for both cases: no field present, more than one field present. For all `ExactlyOneValueSet` validations.

sfc-gh-pcichon: Same as above

> Added validation unit tests for both cases: no field present, more than one field present. For all `ExactlyOneValueSet` validations.

pkg/sdk/external_access_integrations_gen_test.go
sfc-gh-pcichon: Agreed, done.

pkg/sdk/external_access_integrations_gen_test.go
sfc-gh-pcichon: Same as above

> Added validation unit tests for both cases: no field present, more than one field present. For all `ExactlyOneValueSet` validations.

pkg/sdk/generator/defs/external_access_integrations_def.go
```diff
@@ -80,12 +80,15 @@ var externalAccessIntegrationsDef = g.NewInterface(
...
-				WithValidation(g.AtLeastOneValueSet, "AllowedNetworkRules", "AllowedApiAuthenticationIntegrations", "AllowedAuthenticationSecrets", "Comment", "UnsetTags"),
+				WithValidation(g.AtLeastOneValueSet, "AllowedNetworkRules", "AllowedApiAuthenticationIntegrations", "AllowedAuthenticationSecrets", "Comment"),
 			g.ListOptions().NoParentheses().SQL("UNSET"),
 		).
+		OptionalSetTags().
+		OptionalUnsetTags().
 		WithValidation(g.ValidIdentifier, "name").
-		WithValidation(g.ExactlyOneValueSet, "Set", "Unset"),
+		WithValidation(g.ConflictingFields, "IfExists", "SetTags").
+		WithValidation(g.ConflictingFields, "IfExists", "UnsetTags").
```
sfc-gh-kwasilewski: Is there a conflict between IF EXISTS and SET/UNSET TAG? Documentation says nothing about it, and the SQL statement executes successfully.

pkg/sdk/external_access_integrations_validations_ext.go
```diff
@@ -0,0 +1,8 @@
+package sdk
+
+func (opts *ExternalAccessIntegrationSet) additionalValidations() error {
+	if opts.AllowedNetworkRules != nil && len(opts.AllowedNetworkRules) == 0 {
+		return NewError("AllowedNetworkRules must not be empty when provided")
```
sfc-gh-pcichon, line 5: _Why do we return a custom error instead of one from existing error helpers?_
The existing helpers (`errAtLeastOneOf`, `errNotSet`) imply absence, but here the field is explicitly provided (non-nil) yet empty. None of the standard helpers express exactly that. Adding an additional error helper for this specific type of scenario might be an overkill.

pkg/sdk/generator/defs/external_access_integrations_def.go
sfc-gh-pcichon: Good catch, removed these validations.

pkg/sdk/external_access_integrations_validations_ext.go
sfc-gh-kwasilewski, line 5: nit: Create operation could also benefit from this check

sfc-gh-kwasilewski (approved):

pkg/sdk/external_access_integrations_validations_ext.go
sfc-gh-pcichon: For CREATE we do the check through `WithValidation(g.AtLeastOneValueSet, "AllowedNetworkRules")`. This does exactly what we need but I agree that it might not be very explicit. This is done in other places in the codebase like in [API integrations](https://github.com/snowflakedb/terraform-provider-snowflake/blob/dev/pkg/sdk/generator/defs/api_integrations_def.go#L283). 

The reason why we can use it for CREATE but not for ALTER is that for ALTER the field is optional, meaning that the validation for non-emptiness should only be considered iff the field is supplied. Whereas for CREATE the field is required and we can use `WithValidation(g.AtLeastOneValueSet, "AllowedNetworkRules")` which **always** checks for non-emptiness.

pkg/sdk/generator/defs/external_access_integrations_def.go
```diff
@@ -0,0 +1,125 @@
...
+					g.KeywordOptions().SQL("ALLOWED_API_AUTHENTICATION_INTEGRATIONS ="),
+				).
+				OptionalQueryStructField(
+					"AllowedAuthenticationSecrets",
+					allowedAuthenticationSecrets(),
+					g.KeywordOptions().SQL("ALLOWED_AUTHENTICATION_SECRETS ="),
+				).
+				OptionalBooleanAssignment("ENABLED", g.ParameterOptions()).
+				OptionalComment().
+				WithValidation(g.AtLeastOneValueSet, "AllowedNetworkRules", "AllowedApiAuthenticationIntegrations", "AllowedAuthenticationSecrets", "Enabled", "Comment").
```
sfc-gh-jcieslak, line 73: This validation is wrong, only AllowedNetworkRules should be here or should be only validated through additional validations

sfc-gh-jcieslak (approved):

sfc-gh-pcichon:
Closing this PR, as the merging is blocked due to migration to a private development repo. An equivalent PR is opened on that repo and will be merged there.

PR #5066, opened 2026-07-27
fix: Recreate warehouse when type changes externally to INTERACTIVE

sfc-gh-kwasilewski:
Snowflake does not allow changing WAREHOUSE_TYPE via ALTER to INTERACTIVE. Previously the warehouse and warehouse_adaptive resources planned an in-place update when the underlying object was changed externally to INTERACTIVE, which fails on apply.

Both resources now force resource recreation in that case:
- warehouse: a ForceNewIfChange custom diff on warehouse_type.
- warehouse_adaptive: HandleWarehouseExternalTypeChange forces recreation when the external type is INTERACTIVE.

### DIFF
--- a/pkg/resources/custom_diffs.go
+++ b/pkg/resources/custom_diffs.go
@@ -333,8 +333,15 @@ func RecreateWhenStageCloudChangedExternally(stageCloud sdk.StageCloud) schema.C
 	return RecreateWhenResourceTypeChangedExternally("cloud", stageCloud, sdk.ToStageCloud)
 }
 
+// RecreateWhenWarehouseTypeChangedExternally recreates a warehouse when argument warehouseType is different than in the state.
+func RecreateWhenWarehouseTypeChangedExternally(warehouseType sdk.WarehouseType) schema.CustomizeDiffFunc {
+	return RecreateWhenResourceTypeChangedExternally("warehouse_type", warehouseType, sdk.ToWarehouseType)
+}
+
 // HandleWarehouseExternalTypeChange detects when the warehouse type has been changed externally
 // and plans an update to restore it to warehouseType (without forcing recreation).
+// One exception is when the type was changed externally to INTERACTIVE - because Snowflake does not
+// allow changing WAREHOUSE_TYPE via ALTER to INTERACTIVE, the resource has to be recreated instead.
 func HandleWarehouseExternalTypeChange(warehouseType sdk.WarehouseType) schema.CustomizeDiffFunc {
 	return func(_ context.Context, diff *schema.ResourceDiff, _ any) error {
 		if n := diff.Get("warehouse_type"); n != nil {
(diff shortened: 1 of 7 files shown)


pkg/sdk/warehouses_ext.go
```diff
@@ -279,7 +279,8 @@ func (r warehouseDBRow) additionalConvert(wh *Warehouse) error {
 				return fmt.Errorf("invalid resource constraint: %s", r.ResourceConstraint.String)
 			}
 		case WarehouseTypeAdaptive:
-			// Adaptive warehouses don't use resource constraints; ignore.
+		case WarehouseTypeInteractive:
```
sfc-gh-jmichalak: Nit: maybe we should rewrite this switch to a simple if and allow unknown warehouse types? Especially that we use a primitive cast instead of a proper parsing function.

pkg/resources/custom_diffs.go
```diff
@@ -333,8 +333,15 @@ func RecreateWhenStageCloudChangedExternally(stageCloud sdk.StageCloud) schema.C
...
+// RecreateWhenWarehouseTypeChangedExternally recreates a warehouse when argument warehouseType is different than in the state.
+func RecreateWhenWarehouseTypeChangedExternally(warehouseType sdk.WarehouseType) schema.CustomizeDiffFunc {
+	return RecreateWhenResourceTypeChangedExternally("warehouse_type", warehouseType, sdk.ToWarehouseType)
+}
+
 // HandleWarehouseExternalTypeChange detects when the warehouse type has been changed externally
 // and plans an update to restore it to warehouseType (without forcing recreation).
+// One exception is when the type was changed externally to INTERACTIVE - because Snowflake does not
+// allow changing WAREHOUSE_TYPE via ALTER to INTERACTIVE, the resource has to be recreated instead.
 func HandleWarehouseExternalTypeChange(warehouseType sdk.WarehouseType) schema.CustomizeDiffFunc {
```
sfc-gh-jmichalak, line 340: We can just use this function for all the warehouse types and place all the logic there. Because in the current approach we use different functions for each warehouse type:

- forceNewIfChangedToInteractive
- RecreateWhenWarehouseTypeChangedExternally
- HandleWarehouseExternalTypeChange

sfc-gh-jmichalak (changes requested):

pkg/sdk/warehouses_ext.go
sfc-gh-kwasilewski: Done

pkg/resources/custom_diffs.go
sfc-gh-kwasilewski: Removed `RecreateWhenWarehouseTypeChangedExternally`.

As for the `forceNewIfChangedToInteractive`, it needs to stay, because `HandleWarehouseExternalTypeChange` forces the type toward a single expected value, but the normal warehouse resource supports two types (STANDARD and SNOWPARK-OPTIMIZED)

sfc-gh-jmichalak (approved):

PR #5039, opened 2026-07-21
feat: add snowflake_file_format_json resource

sfc-gh-jmichalak:
## Summary
- Adds a new preview resource `snowflake_file_format_json` for managing JSON file formats (create/read/update/delete, import, external-change detection for both attribute drift and type changes).
- Adds SDK-backed schemas, bettertestspoc assert/model generators, docs, and examples for the new resource.
- Adds acceptance tests covering basic/complete use cases, external drift (attribute-level and type-level), and validation errors.

PR #5067, opened 2026-07-27
feat: add snowflake_file_format_csv resource

sfc-gh-jmichalak:
## Summary

Adds a new preview resource `snowflake_file_format_csv` for managing CSV file formats, following the `snowflake_file_format_json` resource added in #5039 as the template.

## Changes

**SDK**
- `csvFileFormatOptionFields` now uses `OptionalQueryStructField("NullIf", nullIfList, ...)` instead of a plain `ListAssignment`, so `nil` means "untouched" (omitted from the statement) while a non-nil empty list renders `NULL_IF = ()`. This matches what JSON already does and is what the resource needs to unset `null_if` on `ALTER`.
- Regenerated `FileFormats` (`--filter-object-names=FileFormats`). The diff is CSV-only; `NullIfList`/`NullIfListRequest` simply move earlier in the generated files because CSV now precedes JSON.
- Callers updated for the new type: `stage_file_format.go`, `stages_gen_test.go`, the testint integration test, `stage_common.go`, and the internal-stage acceptance test.
- New unit test: `set NullIf to an empty list` for `AlterCsv`.

**Resource**
- New `pkg/resources/file_format_csv.go` — create/read/update/delete, import, external-change detection for attribute drift, and `RecreateWhenResourceTypeChangedExternally` for type changes.
- New `pkg/schemas/file_format_csv.go` — `DescribeFileFormatCsvSchema` and `FileFormatCsvToSchema`, reusing the existing `StageFileFormatCsvToSchema` mapping.
- `csvFileFormatSchema` moved from `stage_file_format.go` into `file_format_common.go` as a prefix-parameterized function, so `parse_header`/`skip_header` `ConflictsWith` paths resolve both at the resource top level and nested under `file_format.0.csv.0.`. `stageCsvFileFormatToSchema` became the shared `csvFileFormatToSchema`.
- Added `fileFormatStringOrNoneMapper` / `isFileFormatNoneSentinel` alongside the existing `Auto` equivalents, and pointed `parseStageFileFormatStringOrNone` at the new sentinel helper.
- Registered in the provider, preview features (`snowflake_file_format_csv_resource`), resource enum, issue labels, and the 4 issue templates.

**Tests, docs, examples**
- Generated resourceassert and config model builders; `null_if` ext helpers for the model and describe-output assert.
- `pkg/testacc/resource_file_format_csv_acceptance_test.go` covering basic/complete/altered configs, import, external attribute drift, external type change, unset, rename, and validations; plus the check-destroy entry.
- Examples, doc template, generated docs, `MIGRATION_GUIDE.md` entry, and the preview-resources list.

## Notes for reviewers

Three behaviours worth a look:

- **`skip_header` on import** is set to `IntDefault` (`-1`) rather than Snowflake's actual `0`, mirroring how the boolean fields use `BooleanDefault`. Without this, importing a config that omits `skip_header` shows a permanent diff. The acceptance test's import steps ignore `skip_header`, `escape`, and `escape_unenclosed_field` (the latter two are returned escaped by Snowflake) — the same workaround `resource_internal_stage_acceptance_test.go` already uses.
- **Unsetting `null_if`** falls back to `NULL_IF = ('\\N')`, Snowflake's CSV default, rather than the empty list JSON uses — for JSON, empty *is* the Snowflake default. The escaped-form fallbacks (`\n` for `record_delimiter`, `\\` for `escape_unenclosed_field`, `\\N` for `null_if`) were derived from the existing integration-test assertions, so those round-trips are the part most worth confirming when the acceptance tests run against a real account.
- **`parse_header` ↔ `skip_header`** are mutually exclusive in Snowflake, and the `ALTER` path can emit both in one statement when switching between them (removing one from config sends its fallback while the other is being set), which SDK validation rejects. This is the same latent limitation the JSON resource has with `ignore_utf8_errors`/`replace_invalid_characters`; and the JSON resource will be fixed in a follow-up.

## Verification

`go build ./...`, `go vet ./...`, `golangci-lint` (0 issues), unit tests (7513 pass), `make test-architecture`, and the `generate-sdk-no-tests-check` / `generate-all-config-model-builders-check` / `generate-resource-assertions-check` / `generate-resource-show-output-assertions-check` / `docs-check` / `generate-issue-labels-check` / `generate-docs-additional-files-check` idempotency targets all pass. Acceptance and integration tests need a live Snowflake account and were not run.


pkg/resources/file_format_common.go
```diff
@@ -64,6 +65,19 @@ func fileFormatStringOrAutoMapper(v string) (sdk.StageFileFormatStringOrAutoRequ
 	return *sdk.NewStageFileFormatStringOrAutoRequest().WithValue(v), nil
 }
 
+// isFileFormatNoneSentinel reports whether v is the special "NONE" value Snowflake uses to
+// mean "no delimiter/escape/enclosing character", as opposed to a literal character.
+func isFileFormatNoneSentinel(v string) bool {
```
sfc-gh-asawicki, line 70: nit: if this is special Snowflake value, then it should reside in the SDK

sfc-gh-asawicki (dismissed):

pkg/resources/file_format_common.go
sfc-gh-jmichalak: I'll do this in a follow-up.

sfc-gh-asawicki (approved):

PR #5078, opened 2026-07-28
feat: Rework account and user authentication policy attachment resources

sfc-gh-kwasilewski:
- Add UpdateContext to both resources so changing the policy updates in-place instead of forcing destroy + recreate.
- Rework Read to query POLICY_REFERENCES, enabling drift detection when the policy is changed or removed outside Terraform.
- Fix account attachment ID format from legacy pipe-separated to fully qualified name; state migrates automatically on next plan/apply.
- Add DiffSuppressFunc for identifier quoting on all policy/user fields.
- Add CheckDestroy helpers for account-level policy attachment tests and fix build tags on account-level test files.
- Generate resourceassert and config model helpers for both resources.
- Document import syntax for both resources.
- Add required-warehouse note to account authentication and session policy attachment doc templates.
- Add migration guide entry for both resources.


pkg/resources/account_authentication_policy_attachment.go
```diff
@@ -41,36 +42,101 @@ func AccountAuthenticationPolicyAttachment() *schema.Resource {
...
+	client := meta.(*provider.Context).Client
+
+	if d.HasChange("authentication_policy") {
+		newAuthenticationPolicyName, err := sdk.ParseSchemaObjectIdentifier(d.Get("authentication_policy").(string))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		if err := client.Accounts.Alter(ctx, sdk.NewAlterAccountRequest().
+			WithUnset(*sdk.NewAccountUnsetRequest().WithAuthenticationPolicy(true))); err != nil {
```
sfc-gh-jmichalak, line 124: Q: can't this be done in 1 call with FORCE?

pkg/resources/user_authentication_policy_attachment.go
```diff
@@ -129,15 +141,52 @@ func ReadUserAuthenticationPolicyAttachment(ctx context.Context, d *schema.Resou
...
+		}
+
+		newAuthenticationPolicyName, err := sdk.ParseSchemaObjectIdentifier(d.Get("authentication_policy_name").(string))
+		if err != nil {
+			return diag.FromErr(err)
+		}
+
+		if err := client.Users.Alter(ctx, sdk.NewAlterUserRequest(userName).
+			WithIfExists(true).
+			WithUnset(*sdk.NewUserUnsetRequest().WithAuthenticationPolicy(true))); err != nil {
```
sfc-gh-jmichalak, line 163: ditto

pkg/resources/account_authentication_policy_attachment.go
```diff
@@ -41,36 +42,101 @@ func AccountAuthenticationPolicyAttachment() *schema.Resource {
...
-	err := client.Accounts.Alter(ctx, sdk.NewAlterAccountRequest().
+	err = client.Accounts.Alter(ctx, sdk.NewAlterAccountRequest().
 		WithSet(*sdk.NewAccountSetRequest().WithAuthenticationPolicy(authenticationPolicy)))
 	if err != nil {
-		return diag.FromErr(err)
+		return diag.FromErr(fmt.Errorf("error while creating authentication policy attachment, err = %w", err))
 	}
 
-	d.SetId(helpers.EncodeSnowflakeID(authenticationPolicy))
+	d.SetId(helpers.EncodeResourceIdentifier(authenticationPolicy))
```
sfc-gh-jmichalak, line 59: We need state upgraders for this, IIRC we have a generic upgrader for changing IDs. And acceptance tests for the migration.

sfc-gh-jmichalak (changes requested):

pkg/resources/account_authentication_policy_attachment.go
sfc-gh-kwasilewski: It can, but we don't have FORCE in SDK. I could add it, but for sure not in this PR (it's already a bit too big)

pkg/resources/user_authentication_policy_attachment.go
sfc-gh-kwasilewski: As above

pkg/resources/account_authentication_policy_attachment.go
sfc-gh-kwasilewski: Do we really need a state upgrader? None of the functions currently parse d.Id():
* Create / Update derive the policy from `d.Get("authentication_policy")`
* Delete just issues an ALTER ACCOUNT ... UNSET
* Read retrieves the account locator from `client.GetAccountLocator()` and queries POLICY_REFERENCES

sfc-gh-jmichalak: You're right - I was thinking of the old Read, which did helpers.DecodeSnowflakeIDLegacy(d.Id()).

Things I'd still like:

1. A migration acceptance test, just asserting an empty plan after the upgrade.
2. Comments in code pointing out that we shouldn't read Id in the future anymore because it can be in the old format.
3. Optionally, we don't even need setting the policy Id, the account ID/locator should be enough, given that an account can only have one policy attached.

Not blocking though.

sfc-gh-jmichalak (approved):

pkg/resources/account_authentication_policy_attachment.go
sfc-gh-kwasilewski: Let me do that in a follow-up.

pkg/resources/account_authentication_policy_attachment.go
sfc-gh-kwasilewski: Done in a follow-up PR.

pkg/resources/user_authentication_policy_attachment.go
sfc-gh-kwasilewski: Done in a follow-up PR.

pkg/resources/account_authentication_policy_attachment.go
sfc-gh-kwasilewski: Done in a follow-up PR.

Issue #3860, opened 2025-07-15
[Bug]: Semantic views not supported in permissions


ms32035:
### Terraform CLI Version

1.12.2

### Terraform Provider Version

2.3.0

### Terraform Configuration

```terraform
resource "snowflake_grant_privileges_to_account_role" "all_semantic_views_in_database_all_privileges" {

  privileges        = ["ALL PRIVILEGES"]
  account_role_name = snowflake_account_role.role.name

  with_grant_option = false
  count             = var.semantic_views ? 1 : 0

  on_schema_object {
    all {
      object_type_plural = "SEMANTIC VIEWS"
      in_database        = data.snowflake_database.database.name
    }
  }
}

resource "snowflake_grant_privileges_to_account_role" "future_semantic_views_in_database_all_privileges" {

  privileges        = ["ALL PRIVILEGES"]
  account_role_name = snowflake_account_role.role.name

  with_grant_option = false
  count             = var.semantic_views ? 1 : 0

  on_schema_object {
    future {
      object_type_plural = "SEMANTIC VIEWS"
      in_database        = data.snowflake_database.database.name
    }
  }
}
```

### Category

category:grants

### Expected Behavior

Grants created

### Actual Behavior

Error: expected [{{} on_schema_object} {{} {{{{} 78}} 0xc00149fad0}} {{} future} {{} {{{{} 78}} 0xc000f08c90}} {{} object_type_plural}] to be one of ["ALERTS" "AUTHENTICATION POLICIES" "CORTEX SEARCH SERVICES" "DATA METRIC FUNCTIONS" "DYNAMIC TABLES" "EVENT TABLES" "EXTERNAL TABLES" "FILE FORMATS" "FUNCTIONS" "GIT REPOSITORIES" "HYBRID TABLES" "ICEBERG TABLES" "MATERIALIZED VIEWS" "MODELS" "NETWORK RULES" "NOTEBOOKS" "PASSWORD POLICIES" "PIPES" "PROCEDURES" "SECRETS" "SERVICES" "SEQUENCES" "SNAPSHOTS" "STAGES" "STREAMS" "TABLES" "TASKS" "VIEWS" "STREAMLITS" "DATASETS"], got SEMANTIC VIEWS

### Steps to Reproduce

try to create a grant resource as above

### How much impact is this issue causing?

Medium

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

[...]

PR #5081, opened 2026-07-29
feat: Add backup policies SDK

manshahH:
* Added [BACKUP POLICY](https://docs.snowflake.com/en/sql-reference/sql/create-backup-policy) object to the SDK

Three things below disagree with the SQL reference. All were checked against a live Business Critical account, and the first two were confirmed by @sfc-gh-asawicki on the issue thread before I started.

* **`DROP BACKUP POLICY` supports `IF EXISTS`.** The docs give the grammar as just `DROP BACKUP POLICY <name>`. On a live account, dropping the same policy twice with `IF EXISTS` returns `Drop statement executed successfully (DROP_SYNTAX_CHECK already dropped)` on the second call — the standard no-op response. `IfExists()` is included, matching neighbouring objects.
* **`has_retention_lock` is a `VARCHAR` holding `true`/`false`, not `Y`/`N`.** `DESCRIBE RESULT` on `SHOW BACKUP POLICIES` reports `VARCHAR(16777216)`. It is mapped with `BoolFromText(..., g.WithBoolParsed())` so it goes through `strconv.ParseBool` — the default `== "Y"` comparison would silently report every policy as unlocked.
* **`SNAPSHOT POLICY` appears to be the deprecated former name of this object.** [CREATE SNAPSHOT POLICY](https://docs.snowflake.com/en/sql-reference/sql/create-snapshot-policy) is now marked *Deprecated*: "The CREATE SNAPSHOT POLICY command is deprecated. See CREATE BACKUP POLICY." The provider currently has `ObjectTypeSnapshotPolicy` and `ObjectTypeSnapshotSet` as grantable object types (added in #3860) but no BACKUP equivalents. I have added `ObjectTypeBackupPolicy` alongside rather than renaming theirs, since renaming a grantable object type would be breaking. Happy to open a separate issue if you would like that reconciled.

Two choices I would rather flag than assume — both are settled easily in the integration test PR:

* `comment` and `schedule` are typed nullable (`sql.NullString`). I only ran `SHOW BACKUP POLICIES` against policies that had both set, so I do not know whether an unset `schedule` returns `NULL` or an empty string. Nullable survives either.
* `UNSET` is modelled with `AtLeastOneValueSet`, matching STORAGE LIFECYCLE POLICY. The docs write it as `UNSET { COMMENT | SCHEDULE | EXPIRE_AFTER_DAYS }`, which could mean exactly one. Untested — I will confirm and tighten it if needed.

Per the PR split agreed on the issue, integration tests, the resource and the data source follow in separate PRs.

## Test Plan

* [x] unit tests
* [ ] integration tests (next PR)
* [ ] acceptance tests (resource PR)

## References

* https://github.com/snowflakedb/terraform-provider-snowflake/issues/4294
* https://github.com/snowflakedb/terraform-provider-snowflake/issues/5038#issuecomment-5116029570


sfc-gh-jmichalak:
Hi @manshahH 👋 

Thanks for your contribution! We'll discuss it internally on Monday.

manshahH:
> Hi @manshahH 👋
> 
> Thanks for your contribution! We'll discuss it internally on Monday.

Hello @sfc-gh-jmichalak 
take your time, enjoy the weekend. i look forward to contributing more!

sfc-gh-jmichalak (approved):
LGTM, thanks for your contribution!

I'm also observing the drift from the documentation. We'll push this internally, and this is not blocking. However, we're changing our contributing flow, so we'll be able to merge this PR with a delay.

sfc-gh-jcieslak:
Closing as this change was synced through this commit (with `Co-Authored-By` marking): https://github.com/snowflakedb/terraform-provider-snowflake/commit/4ddefa4029a97012273678f75d795ec166d63d49

PR #5095, opened 2026-08-05
feat: extend SHOW GRANTS caching to snowflake_grant_privileges_to_account_role and snowflake_grant_ownership

raffael-dzikowski:
## Changes

Extends the existing `GRANT_ACCOUNT_ROLE_SHOW_CACHING` pattern to two more resources that show the same shape of redundancy: `snowflake_grant_privileges_to_account_role` and `snowflake_grant_ownership` each independently issue a `SHOW GRANTS ON <object>` / `SHOW FUTURE GRANTS IN <container>` call per resource instance during Read, even when many instances resolve to the exact same underlying SHOW statement — e.g. several privilege grants on one schema, or several future-grant roles on one database. Adds a new `GRANTS_SHOW_CACHING` experiment that caches these per plan/apply cycle.

- New `GRANTS_SHOW_CACHING` experiment, independent of `GRANT_ACCOUNT_ROLE_SHOW_CACHING` (kept as-is, unchanged, for `snowflake_grant_account_role`) — both can be enabled together
- `snowflake_grant_privileges_to_database_role` and `snowflake_grant_privileges_to_share` don't cache their own Reads (out of scope here), but now invalidate the shared cache after their own mutations, since they write into the same `SHOW GRANTS ON <object>` key space the cached resources read from
- `pkg/internal/provider/cache.go` — `GetOrLoad` now takes the caller's ctx and derives the cancellable load ctx from it (previously always `context.Background()`), so Terraform Timeouts/plan cancellation propagate into a cache-miss load

## How

- **Cache key**: the rendered SQL of the SHOW statement (new `sdk.ShowGrantOptionsToSQL`, wrapping the same `structToSQL` that `Grants.Show` itself uses to build its query), not a hand-rolled identifier. Two `ShowGrantOptions` values that render to the same SQL are guaranteed to return the same result set, so this is a provably-correct key rather than something that could drift out of sync with what's actually queried.
- **Cross-resource sharing falls out of that for free**: because the key is the SQL itself, `snowflake_grant_ownership` and `snowflake_grant_privileges_to_account_role` targeting the same object naturally collapse to one cached entry, even though they're different resource types. This is exercised directly in the new acceptance tests.
- **`snowflake_grant_account_role`'s existing cache is migrated onto this same keying scheme** — it previously kept its own `roleIdentifier.FullyQualifiedName()` key in a separate cache field. Behavior and its flag name are unchanged; this is purely an internal simplification down to one cache field, done here because the resource's Read/Create/Delete needed touching anyway for the ctx fix below.
- **Why the ctx fix is bundled in**: extending the cache to more call sites made the existing `context.Background()`-only behavior more visible, so it's fixed once here rather than deferred.
- **Scope boundary**: `snowflake_grant_privileges_to_account_role` also makes a second, currently-uncached round trip per instance (an account-role existence check) that this PR does not address — see #5096, a separate, independent experiment covering that lookup.

## Tests

- Two new deterministic unit tests for the ctx propagation/cancellation change, plus a new `TestShowGrantOptionsToSQL` covering the rendering delegation and determinism — the two properties the cache key correctness relies on. Full per-grant-kind SQL rendering coverage already exists in `TestGrantShow`.
- New acceptance test file (`resource_grants_show_caching_experimental_features_acceptance_test.go`) mirroring the existing `GRANT_ACCOUNT_ROLE_SHOW_CACHING` test shape, including one test proving the cross-resource-type cache sharing claim above.

## References

- #5096 — companion PR caching the account-role existence/properties lookup (`Roles.ShowByID`/`ShowByIDSafely`), the second uncached round trip noted in the scope boundary above. Independent and reviewable separately, but touches an adjacent line in `ReadGrantPrivilegesToAccountRole` — whichever of the two merges second will need a small, mechanical rebase there.


pkg/resources/grant_account_role.go
```diff
@@ -129,13 +129,27 @@ func CreateGrantAccountRole(ctx context.Context, d *schema.ResourceData, meta an
 		// computed/server-default fields to populate, so the Read is redundant here. With caching
 		// enabled we skip it to avoid an extra SHOW GRANTS during apply. We still invalidate so any
 		// later Read for this role in the same plan observes the new grant.
-		providerCtx.GrantShowOfRoleCache.Invalidate(roleIdentifier.FullyQualifiedName())
+		if cacheKey, keyErr := showGrantsOfRoleCacheKey(roleIdentifier); keyErr == nil {
+			providerCtx.GrantShowCache.Invalidate(cacheKey)
+		}
```
sfc-gh-jmichalak: Nit: the abstraction level is a bit inconsistent - in other resources we have nice helpers for invalidating the cache, checking constraints (experiment level, cache key). I know these resources use different experiments but I believe we could still improve readability :) Not blocking

pkg/sdk/grants_test.go
```diff
@@ -1286,6 +1289,59 @@ func TestGrantShow(t *testing.T) {
...
+	t.Run("matches Grants.Show's own rendering", func(t *testing.T) {
+		expected, err := structToSQL(opts)
+		require.NoError(t, err)
+
+		actual, err := ShowGrantOptionsToSQL(opts)
+		require.NoError(t, err)
+		assert.Equal(t, expected, actual)
+	})
+
+	t.Run("deterministic across repeated calls and independently-built equivalent values", func(t *testing.T) {
```
sfc-gh-jmichalak: Nit: IMO this case is unnecessary.

pkg/testacc/resource_grants_show_caching_experimental_features_acceptance_test.go
```diff
@@ -0,0 +1,254 @@
...
+			},
+		},
+	})
+}
+
+// TestAcc_GrantsShowCaching_AccountRolePrivileges_SharedObjectIssuesShowPerInstanceWithoutExperiment
+// is the contrast to the test above: without the experiment, the same database targeted by two
+// separate grants is shown multiple times (once per Read of each instance), which is exactly the
+// redundancy the experiment removes.
+func TestAcc_GrantsShowCaching_AccountRolePrivileges_SharedObjectIssuesShowPerInstanceWithoutExperiment(t *testing.T) {
```
sfc-gh-jmichalak: Nit: this case could be bundled with the one above as separate steps. We'd get rid of the dependency setup.

pkg/testacc/resource_grants_show_caching_experimental_features_acceptance_test.go
```diff
@@ -0,0 +1,254 @@
...
+			{
+				Config: config.FromModels(t, experimentProviderModel, grantWithUsage),
+				Check:  resource.TestCheckResourceAttr(resourceName, "privileges.#", "1"),
+			},
+			{
+				Config: config.FromModels(t, experimentProviderModel, grantWithUsageAndCreateSchema),
+				Check:  resource.TestCheckResourceAttr(resourceName, "privileges.#", "2"),
+			},
+			// If Update failed to invalidate the cache, this refresh would still see the pre-Update,
+			// single-privilege SHOW GRANTS result and produce a non-empty plan (drift back to 1).
```
sfc-gh-jmichalak: Nit: we can also assert the number of SHOW queries here.

sfc-gh-jmichalak (approved):
LGTM, a couple of nits you can address later, or we can do it ourselves. A kind reminder that we're changing our contributing flow, so we can't merge it for now. We'll take care of it :)

pkg/internal/provider/provider_context.go
```diff
@@ -3,8 +3,12 @@ package provider
...
-	EnabledFeatures      []string
-	EnabledExperiments   []string
-	GrantShowOfRoleCache *Cache[[]sdk.Grant]
+	Client             *sdk.Client
+	EnabledFeatures    []string
+	EnabledExperiments []string
+	// GrantShowCache caches SHOW GRANTS results, keyed by the rendered SQL of the ShowGrantOptions
+	// that produced them (see sdk.ShowGrantOptionsToSQL). Shared by snowflake_grant_account_role
+	// (gated by GRANT_ACCOUNT_ROLE_SHOW_CACHING), and snowflake_grant_privileges_to_account_role /
+	// snowflake_grant_ownership (gated by GRANTS_SHOW_CACHING).
```
sfc-gh-asawicki: nit: Please reduce the commentary to the bear minimum (here and in other places). IMO this one could be fully removed.

pkg/resources/grant_account_role.go
```diff
@@ -129,13 +129,27 @@ func CreateGrantAccountRole(ctx context.Context, d *schema.ResourceData, meta an
...
 	}
 	return ReadGrantAccountRole(ctx, d, meta)
 }
 
+// showGrantsOfRoleCacheKey renders the SHOW GRANTS OF ROLE <role> statement for roleIdentifier to
+// the canonical cache key shared with the GrantShowCache. sdk.ShowGrantOptionsToSQL only fails on a
+// struct-rendering error from the reflection-based SQL builder; the literal ShowGrantOptions built
+// here is always well-formed, so an error is not expected in practice. Callers should still treat
+// it as fall back to an uncached SHOW rather than fail the whole Read/Create/Delete over a cache-key
+// rendering problem.
```
sfc-gh-asawicki: nit: Please reduce the commentary to the bear minimum (here and in other places). IMO this one could be fully removed.

pkg/resources/grant_helpers.go
```diff
@@ -15,6 +19,49 @@ func toPrivileges(privileges []string) ([]string, error) {
...
+		return providerCtx.Client.Grants.Show(ctx, opts)
+	}
+	return providerCtx.GrantShowCache.GetOrLoad(ctx, key, func(loadCtx context.Context) ([]sdk.Grant, error) {
+		return providerCtx.Client.Grants.Show(loadCtx, opts)
+	})
+}
+
+// invalidateGrantsShowCache invalidates the GrantShowCache entry for opts, if the GRANTS_SHOW_CACHING
+// experiment is enabled. A no-op (not an error) if opts fails to render or the cache was never
+// populated for it — Invalidate on an absent key is already a documented no-op.
```
sfc-gh-asawicki: nit: this comment should be remvoed

pkg/resources/grant_helpers.go
```diff
@@ -15,6 +19,49 @@ func toPrivileges(privileges []string) ([]string, error) {
 	return collections.MapErr(privileges, sdk.ToPrivilege)
 }
 
+// showGrantsCached issues opts via client.Grants.Show, transparently caching the result in
```
sfc-gh-asawicki: nit: Please reduce the commentary to the bear minimum (here and in other places). I think this one could be almost fully deleted too

pkg/resources/grant_ownership.go
```diff
@@ -655,6 +666,21 @@ func prepareShowGrantsRequestForGrantOwnership(id *GrantOwnershipId) (*sdk.ShowG
 	return opts, expectedGrantedOn
 }
 
+// invalidateShowGrantsForOwnershipId invalidates the cached SHOW GRANTS result (if any) for the
+// statement that Read would issue for id, using the same request-building logic as Read
+// (prepareShowGrantsRequestForGrantOwnership) so the invalidated key always matches the cached key.
+// Because that key is shared with other grant resources targeting the same object/container (see
+// the comment at the CreateGrantOwnership call site), this also invalidates their cached view of
+// this object, not just grant_ownership's own. A no-op for id.Kind values whose Read issues no
+// SHOW at all (opts == nil, e.g. OnAllGrantOwnershipKind).
```
sfc-gh-asawicki: nit: Please reduce the commentary to the bear minimum (here and in other places). IMO this one could be fully removed.

pkg/resources/grant_ownership.go
```diff
@@ -273,6 +274,14 @@ func CreateGrantOwnership(ctx context.Context, d *schema.ResourceData, meta any)
 
 	d.SetId(id.String())
 
+	// A GrantOwnership call can change more than the OWNERSHIP grant itself: depending on
+	// outbound_privileges (REVOKE|COPY) it can also change other privileges on the same object,
+	// which is exactly the SHOW GRANTS result that snowflake_grant_privileges_to_account_role
+	// (and the other grant resources sharing this object) may have cached. Because the cache key
+	// is the rendered SQL of the SHOW statement, invalidating with our own opts invalidates the
+	// same entry any of those resources would have populated for this object/container.
+	invalidateShowGrantsForOwnershipId(providerCtx, id)
```
sfc-gh-asawicki: let's shorten this comment and apply the same short one in other places (like 1 or 2 lines max)

pkg/resources/grant_privileges_to_share.go
```diff
@@ -555,6 +562,22 @@ func getShareGrantOn(d *schema.ResourceData) (*sdk.ShareGrantOn, error) {
...
 
+// invalidateShowGrantsForShareId invalidates the cached SHOW GRANTS result (if any) for the
+// statement that Read would issue for id, using the same request-building logic as Read
+// (prepareShowGrantsRequestForShare) so the invalidated key always matches the cached key. This
+// resource does not itself cache its Read (it is not in scope for the GRANTS_SHOW_CACHING
+// experiment), but it mutates the same SHOW GRANTS ON <object> key space that
+// snowflake_grant_privileges_to_account_role and snowflake_grant_ownership do cache under that
+// experiment — without this, a grant/revoke here could leave a stale cached entry unnoticed for
+// the rest of the plan/apply cycle. A no-op for id.Kind values whose Read issues no SHOW at all.
+func invalidateShowGrantsForShareId(providerCtx *provider.Context, id GrantPrivilegesToShareId) {
```
sfc-gh-asawicki: nit: a few handling functions are pretty similar; instead of having them spread across different resources, let's maybe extract a single file for the cache-experiment related methods and put them there? It may be also easier to find how we could make them even more generic when yhey are in the same place

pkg/sdk/grants.go
```diff
@@ -286,6 +286,15 @@ type ShowGrantsOf struct {
 	Share           AccountObjectIdentifier  `ddl:"identifier" sql:"SHARE"`
 }
 
+// ShowGrantOptionsToSQL renders opts to the exact SQL string that Grants.Show would issue for it
+// (Grants.Show itself goes through the same structToSQL call, see validateAndQuery in
+// helpers_proposal.go). Two ShowGrantOptions values are safe to treat as equivalent for caching
+// purposes if and only if they render to the same string, since identical SQL against an
+// unchanged Snowflake account is guaranteed to return an identical result set.
+func ShowGrantOptionsToSQL(opts *ShowGrantOptions) (string, error) {
```
sfc-gh-asawicki: nit: maybe instead of exposing such a function per type, we could extract one method taking any? It could be even just a duplicate StructToSql delegating to structToSQL.

@sfc-gh-jmichalak WDYT? do you see any risks with such an approach?

pkg/sdk/grants_test.go
```diff
@@ -1286,6 +1289,59 @@ func TestGrantShow(t *testing.T) {
...
 }
 
+// TestShowGrantOptionsToSQL covers the properties ShowGrantOptionsToSQL is relied on for as a
+// cache key (see pkg/internal/provider.Cache and its callers in pkg/resources): it must render
+// exactly what Grants.Show would issue, and it must be deterministic across repeated calls on an
+// equivalent value, since two ShowGrantOptions built independently (e.g. by different resources
+// reading the same object) need to collapse to the same cache key. Coverage of the rendering
+// itself, across every ShowGrantOptions shape, already lives in TestGrantShow above (it exercises
+// the same structToSQL that ShowGrantOptionsToSQL wraps) — this test intentionally does not
+// duplicate that.
```
sfc-gh-asawicki: This comment is not necessary

pkg/testacc/resource_grants_show_caching_experimental_features_acceptance_test.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+// countShowGrantsOnDatabaseQueries returns how many `SHOW GRANTS ON DATABASE <id>` statements were
+// issued (across all sessions of the test user) for the given, test-unique database. Because the
+// database name is randomly generated per test, this count is isolated to the queries produced by
+// the test under inspection. Used to prove the GRANTS_SHOW_CACHING experiment collapses N identical
+// SHOW calls into one, including across different resource types (see the mirror-key comment on
+// showGrantsCached in pkg/resources/grant_helpers.go).
+func countShowGrantsOnDatabaseQueries(t *testing.T, databaseId sdk.AccountObjectIdentifier) int {
```
sfc-gh-asawicki, line 20: the comment is not necessary

pkg/testacc/resource_grants_show_caching_experimental_features_acceptance_test.go
```diff
@@ -0,0 +1,254 @@
...
+		return strings.Contains(h.QueryText, needle)
+	}))
+}
+
+// TestAcc_GrantsShowCaching_AccountRolePrivileges_SharedObjectIssuesSingleShow proves the cache
+// effect for snowflake_grant_privileges_to_account_role: with the experiment enabled, two grants on
+// the same database to two different roles result in exactly one `SHOW GRANTS ON DATABASE` call for
+// that database over the provider's lifetime, instead of one per instance.
+func TestAcc_GrantsShowCaching_AccountRolePrivileges_SharedObjectIssuesSingleShow(t *testing.T) {
+	database, databaseCleanup := testClient().Database.CreateDatabase(t)
```
sfc-gh-asawicki: Use CreateDatabaseWithParametersSet instead of CreateDatabase in every place used

pkg/testacc/resource_grants_show_caching_experimental_features_acceptance_test.go
sfc-gh-asawicki, line 20: This method is almost the same as countShowGrantsOfRoleQueries, let's generify

pkg/testacc/resource_grants_show_caching_experimental_features_acceptance_test.go
```diff
@@ -0,0 +1,254 @@
...
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						plancheck.ExpectEmptyPlan(),
+					},
+				},
+				Check: func(_ *terraform.State) error {
+					// Without the experiment, each instance issues its own SHOW on every Read pass, so the
+					// shared database is shown well more than once. We assert "more than one" rather than
+					// an exact number to stay robust against Terraform's refresh cadence.
+					if got := countShowGrantsOnDatabaseQueries(t, database.ID()); got <= 1 {
```
sfc-gh-asawicki: nit: got < 2 would be a bit more self-explanatory

pkg/testacc/resource_grants_show_caching_experimental_features_acceptance_test.go
```diff
@@ -0,0 +1,254 @@
...
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: config.FromModels(t, experimentProviderModel, ownershipModel, privilegesModel),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("snowflake_grant_ownership.ownership", "account_role_name", roleOwner.ID().Name()),
+					resource.TestCheckResourceAttr("snowflake_grant_privileges_to_account_role.privileges", "account_role_name", roleGrantee.ID().Name()),
+				),
+			},
+			// Both resources' Read must converge on a second refresh, and — the point of this test —
```
sfc-gh-asawicki: I dont think that this comment in every test is necessary; I would remove

sfc-gh-asawicki (changes requested):

pkg/sdk/grants.go
sfc-gh-jmichalak: SG

raffael-dzikowski:
Thank you both again for the in-depth review. All mentioned points should now be addressed.

pkg/testacc/resource_grant_account_role_show_caching_experimental_features_acceptance_test.go
```diff
@@ -20,20 +20,24 @@ import (
...
+func countShowQueries(t *testing.T, needle string) int {
 	t.Helper()
 	queryHistory := testClient().InformationSchema.GetQueryHistory(t, 1000)
-	needle := fmt.Sprintf("SHOW GRANTS OF ROLE %s", roleId.FullyQualifiedName())
 	return len(collections.Filter(queryHistory, func(h helpers.QueryHistory) bool {
 		return strings.Contains(h.QueryText, needle)
 	}))
 }
 
+// countShowGrantsOfRoleQueries returns how many `SHOW GRANTS OF ROLE <id>` statements were issued
```
sfc-gh-asawicki: This is comment is just restating the implementation, it should be deleted.

sfc-gh-asawicki (approved):
LGTM

sfc-gh-jmichalak (approved):

pkg/testacc/resource_grant_account_role_show_caching_experimental_features_acceptance_test.go
raffael-dzikowski: Thanks. I just removed the comment.

sfc-gh-jcieslak:
Closing as this change was synced through this commit (with `Co-Authored-By` marking): https://github.com/snowflakedb/terraform-provider-snowflake/commit/230809d1859d2d38444bf69882bad621c8599337

PR #5096, opened 2026-08-06
feat: cache the account-role existence/properties lookup (Roles.ShowByID/ShowByIDSafely)

raffael-dzikowski:
## Changes

`snowflake_account_role`, `snowflake_grant_application_role`, and `snowflake_grant_privileges_to_account_role` each independently call `Roles.ShowByID`/`ShowByIDSafely` (`SHOW ROLES LIKE '<name>'`) to look up an account role — once per resource instance, even when many instances reference the same role. This is a common shape in RBAC-style configurations: dozens of grant resources targeting the same handful of roles, each triggering its own redundant round trip during Read. Adds a new `ACCOUNT_ROLE_SHOW_CACHING` experiment that caches this lookup for the duration of a single plan/apply cycle, so N resource instances referencing the same role collapse to one round trip instead of N.

- New `RoleShowCache *Cache[*sdk.Role]` field on `provider.Context`, reusing the existing generic `Cache[T]`
- Wraps the `Roles.ShowByID`/`ShowByIDSafely` call in all three resources above behind the cache, gated on the new experiment
- Invalidates on `snowflake_account_role` Update (rename and/or comment change) and Delete — the only resource whose CRUD can make a cached entry stale
- `snowflake_account_role`'s `DeleteContext` moves from the shared generic `ResourceDeleteContextFunc` helper to a bespoke `DeleteAccountRole`, since the shared helper is used by other resources and isn't the right place for role-cache-specific invalidation
- This is a separate, independent flag from `GRANT_ACCOUNT_ROLE_SHOW_CACHING` (which caches `SHOW GRANTS OF ROLE`, not the role object itself) — both can be enabled together

## How

- **Cache key**: the role's fully qualified name directly. Unlike `SHOW GRANTS`, `ShowByID`/`ShowByIDSafely` always resolve to the same statement for a given identifier, so there's no need for a canonical SQL-renderer here — the identifier itself is already the correct, unique key.
- **Why no invalidation on Create**: `ShowByID`/`ShowByIDSafely` return `(nil, ErrObjectNotFound)` on a miss, and `Cache.GetOrLoad` never caches an error result — so a pre-creation "not found" is never cached in the first place, and there's nothing stale to clear once the role exists.
- **Why Update invalidates unconditionally, not just on rename**: `ShowByIDSafely` returns the full `Role` struct including `Comment`, so a comment-only change can make a cached entry just as stale as a rename. The old identifier is invalidated right after a successful rename; the current identifier is invalidated unconditionally before the trailing Read, covering both cases.
- **Scope check**: audited all callers of the equivalent SDK methods. The parallel database-role family (`DatabaseRoles.ShowByID`, used by `snowflake_database_role` and `snowflake_grant_privileges_to_database_role`) isn't touched here — no config we tested against exercises it, so it's left out rather than built speculatively. Happy to extend this PR (or follow up) if there's real demand.

## Tests

- New acceptance test file (`resource_account_role_show_caching_experimental_features_acceptance_test.go`) mirroring the existing `GRANT_ACCOUNT_ROLE_SHOW_CACHING` test shape: a shared-role-issues-single-show test, a without-experiment contrast, and a rename-invalidates-cache test (no precedent in the existing caching tests, since `snowflake_grant_account_role` has no Update)
- `go build`/`go vet`/`golangci-lint`/`gofumpt` clean, with and without `-tags non_account_level_tests`; acceptance tests compile but were not run against a live account from this environment

## References

- Companion PR, same motivation and profiling exercise: https://github.com/snowflakedb/terraform-provider-snowflake/pull/5095 (extends `SHOW GRANTS` caching to `snowflake_grant_privileges_to_account_role`/`snowflake_grant_ownership`). Independent and reviewable separately, but touches an adjacent line in `ReadGrantPrivilegesToAccountRole` — whichever of the two merges second will need a small, mechanical rebase there.
- Profiled against a large, real-world RBAC configuration where a handful of roles are referenced by a very large number of grant resources; the redundant role lookups were a measurable share of plan time in that setup, in the same vein as the existing `GRANT_ACCOUNT_ROLE_SHOW_CACHING` experiment.


pkg/internal/provider/provider_context.go
```diff
@@ -7,4 +7,9 @@ type Context struct {
 	EnabledFeatures      []string
 	EnabledExperiments   []string
 	GrantShowOfRoleCache *Cache[[]sdk.Grant]
+	// RoleShowCache caches the result of looking up an account role by identifier (SHOW ROLES LIKE
+	// '<name>', via Roles.ShowByID/ShowByIDSafely), keyed by the role's fully qualified name. Shared
+	// by snowflake_account_role, snowflake_grant_application_role, and
+	// snowflake_grant_privileges_to_account_role (gated by ACCOUNT_ROLE_SHOW_CACHING).
```
sfc-gh-asawicki: let's drop this comment

pkg/testacc/resource_account_role_show_caching_experimental_features_acceptance_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfversion"
+)
+
+// countShowRolesLikeQueries returns how many `SHOW ROLES LIKE '<name>'` statements were issued
+// (across all sessions of the test user) for the given, test-unique role. Because the role name is
+// randomly generated per test, this count is isolated to the queries produced by the test under
+// inspection. Used to prove the ACCOUNT_ROLE_SHOW_CACHING experiment collapses N identical SHOW
+// calls into one.
+func countShowRolesLikeQueries(t *testing.T, roleId sdk.AccountObjectIdentifier) int {
```
sfc-gh-asawicki, line 24: nit: the same comment as in #5095

sfc-gh-asawicki (changes requested):

sfc-gh-jmichalak (approved):

raffael-dzikowski:
Thank you both. This is now addressed analogously to #5095.

sfc-gh-asawicki (approved):
LGTM

sfc-gh-jmichalak (approved):

sfc-gh-jcieslak:
Closing as this change was synced through this commit (with `Co-Authored-By` marking): https://github.com/snowflakedb/terraform-provider-snowflake/commit/2e501f10b0b78349553aa55cc969e07bd5aa5af1

PR #5099, opened 2026-08-13
fix: (SNOW-3953574) Handle unprefixed database role grantee names in SHOW GRANTS (BCR-2371)

sfc-gh-dszmolka:
### Issue
SNOW-3953574 (customer complaint)

The 2026_06 bundle (BCR-2371) changes `SHOW GRANTS OF DATABASE ROLE` so a database role granted to another database role is returned without the database prefix (PARENT_DR) instead of fully qualified (MYDB.PARENT_DR).

`ParseDatabaseObjectIdentifier` requires two parts and errored on the unprefixed name, aborting the whole `Show()` call. 
`ReadGrantDatabaseRole` then cleared the resource ID, causing a perpetual re-create, "Provider produced inconsistent result after apply" errors, and failed imports.

The grantee name is now normalized to a fully qualified identifier in both bundle states: when the prefix is absent it is reconstructed from the queried role's database (a database role can only be granted to another database role in the same database), keeping the grantee stable and importable whether or not the bundle is enabled.

Adds an SDK integration test and a MIGRATION_GUIDE.md entry.

### References
* BCR-2371 - SHOW GRANTS output consistency improvement - https://docs.snowflake.com/en/release-notes/bcr-bundles/2026_06/bcr-2371

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -220,6 +220,26 @@ Now, a column type that cannot be parsed no longer fails the whole describe call
 
 References: [#5090](https://github.com/snowflakedb/terraform-provider-snowflake/issues/5090)
 
+### *(bug fix)* Fixed handling of unprefixed database role grantee names in `SHOW GRANTS` (2026_06 bundle / BCR-2371)
+
+Previously, when a database role was granted to another database role, the provider expected the `SHOW GRANTS OF DATABASE ROLE` output to return the grantee database role as a fully qualified name (`<database>.<database_role>`). The 2026_06 bundle (BCR-2371) changes this output so that the grantee database role is returned without the database prefix (just `<database_role>`). With the bundle enabled, the provider could not parse the grantee name, so the Read operation of the `snowflake_grant_database_role` resource failed to find the grant and marked the resource as deleted. `terraform plan` then showed a permanent diff recreating the grant, and `terraform apply` returned an error like
+
+```
+│ Error: Provider produced inconsistent result after apply
+│
+│ When applying changes to snowflake_grant_database_role.example, provider "provider[\"registry.terraform.io/snowflakedb/snowflake\"]" produced an unexpected new value: Root object was present, but now absent.
+│
+│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
+```
+
+Importing such a resource failed for the same reason (`Cannot import non-existent remote object`).
+
+In this release, the grantee database role name is normalized to a fully qualified identifier regardless of whether the bundle is enabled: when the database prefix is missing, it is reconstructed from the database of the queried database role (a database role can only be granted to another database role in the same database). The `snowflake_grant_database_role` resource is no longer recreated on every plan and can be imported again.
+
+No changes in configuration are required.
+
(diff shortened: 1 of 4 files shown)


pkg/sdk/testint/grants_integration_test.go
```diff
@@ -2514,6 +2514,33 @@ func TestInt_ShowGrants(t *testing.T) {
 		assert.Equal(t, testClientHelper().Ids.SnowflakeApplicationId().Name(), grants[0].GranteeName.Name())
 	})
 
+	t.Run("grantee name of a database role granted to another database role is fully qualified", func(t *testing.T) {
```
sfc-gh-asawicki, line 2517: The bundle is not yet enabled by default. I don't think we currently have any "live" BCR bundle test, but we usually:
1. Use secondary test account for such tests.
2. Turn the bundle on.
3. Test the new behavior.
Later on, the test will be removed/adjusted when the Bundle is dead.

pkg/sdk/testint/grants_integration_test.go
```diff
@@ -2514,6 +2514,33 @@ func TestInt_ShowGrants(t *testing.T) {
...
+		// Regression test for the 2026_06 bundle (BCR-2371): SHOW GRANTS OF DATABASE ROLE returns the
```
sfc-gh-asawicki, line 2518: nit: we put these comments at the top of the t.Run

pkg/sdk/grants_impl.go
```diff
@@ -405,11 +405,20 @@ func (v *grants) Show(ctx context.Context, opts *ShowGrantOptions) ([]Grant, err
 			}
 			resultList[i].GranteeName = NewAccountObjectIdentifier(granteeName)
 		} else if !slices.Contains([]ObjectType{ObjectTypeRole, ObjectTypeShare, ObjectTypeUser, ObjectTypeApplication}, grant.GrantedTo) {
-			id, err := ParseDatabaseObjectIdentifier(granteeNameRaw)
+			// Before BCR-2371 the grantee (another database role) is fully qualified (<database>.<database_role>); after, only <database_role>.
+			// A database role can only be granted within the same database, so reconstruct the missing prefix from the queried role.
+			granteeNameParts, err := ParseIdentifierString(granteeNameRaw)
```
sfc-gh-asawicki: `ParseIdentifierString` has a comment that it will be hidden, let's maybe change the logic to:
1. try ParseDatabaseObjectIdentifier
2. try ParseAccountObjectIdentifier if err
3. err otherwise
WDYT?

MIGRATION_GUIDE.md
```diff
@@ -220,6 +220,26 @@ Now, a column type that cannot be parsed no longer fails the whole describe call
 
 References: [#5090](https://github.com/snowflakedb/terraform-provider-snowflake/issues/5090)
 
+### *(bug fix)* Fixed handling of unprefixed database role grantee names in `SHOW GRANTS` (2026_06 bundle / BCR-2371)
```
sfc-gh-asawicki, line 223: With such changes, we should have also an acceptance test (also on the secondary account, with the dedicated provider factory), checking that old version (strict 2.19) errors out and then snext step showing that with the new version it works.

sfc-gh-asawicki (changes requested):

PR #5027, opened 2026-07-20
feat: Add catalog-linked database support to the SDK

Benjamin0313:
## Summary

- `CreateCatalogLinked` operation covering the full [CREATE DATABASE (catalog-linked)](https://docs.snowflake.com/en/sql-reference/sql/create-database-catalog-linked) syntax: `LINKED_CATALOG` parameters, `EXTERNAL_VOLUME`, `COMMENT`, `TAG`, and `CATALOG_CASE_SENSITIVITY`
- `AlterCatalogLinked` operation per [ALTER DATABASE (catalog-linked)](https://docs.snowflake.com/en/sql-reference/sql/alter-database-catalog-linked): ADD/REMOVE/UNSET for allowed and blocked namespaces, and SET for `SYNC_INTERVAL_SECONDS` / `ALLOWED_WRITE_OPERATIONS`; `RENAME TO` and `COMMENT` are already covered by the existing generic `Alter`
- New enums: `CatalogLinkedDatabaseNamespaceMode`, `CatalogLinkedDatabaseAllowedWriteOperations`, `DatabaseCatalogCaseSensitivity`
- `SYSTEM$GET_CATALOG_LINKED_DATABASE_CONFIG` and `SYSTEM$CATALOG_LINK_STATUS` wrappers in `system_functions.go` (following the `SYSTEM$CLUSTERING_INFORMATION` pattern) — needed for drift detection since `LINKED_CATALOG` settings are not exposed via SHOW/DESCRIBE

Notes on decisions:
- The generated `databases_gen_test.go` is not committed, matching the convention of recent generator migrations (#4975, #4985); unit tests live in the handwritten `databases_test.go`, including enum conversion tests.
- Quoting of identifiers inside `LINKED_CATALOG` (`CATALOG = '"NAME"'`) follows the existing Iceberg tables pattern; live verification lands with the integration tests PR.
- Per the agreed scope, `WITH CONTACT`, `SUSPEND | RESUME DISCOVERY`, and the ALTER-only parameters (`BASE_LOCATION_PREFIX`, `ICEBERG_VERSION_DEFAULT`, `ICEBERG_MERGE_ON_READ_BEHAVIOR`) are intentionally excluded from this PR.

Subsequent PRs per the approved plan: (2) SDK integration tests, (3) the `snowflake_catalog_linked_database` resource, (4) follow-up wiring of the three parameters above.

## Test Plan
[...]

Issue #4201, opened 2025-11-16
[Feature]: Support Catalog-Linked Databases


MrChadMWood:
### Use Cases or Problem Statement

Snowflake has a new “catalog-linked databases” feature. 
> https://docs.snowflake.com/en/sql-reference/sql/create-database-catalog-linked

A catalog-linked database means a Snowflake database object that is linked to an external Iceberg REST catalog. Snowflake will automatically sync namespaces and tables from that external catalog into the database.

Currently, the `snowflake_database` resource supports a `catalog` parameter. To avoid confusion, this seems to mean it’s setting a default catalog integration that Iceberg tables in that database will use—not a support mechanism for the “caralog-linked database.”

This feature request is to ask for first-hand support of the catalog-linked database via SnowflakeDB Terraform provider.

### Category

category:resource

### How much impact is this issue causing?

Low

### Would you like to implement a fix?

- [ ] Yeah, I'll take it 😎

sfc-gh-mkowalski:
Hi @MrChadMWood, thanks for reaching out!

Current Terraform Provider [roadmap](https://github.com/snowflakedb/terraform-provider-snowflake/blob/main/ROADMAP.md) does not include this feature for now. Your request has been added to our internal list and will be discussed in the future.

jl178:
@sfc-gh-mkowalski - Any updates/news on this feature?

sfc-gh-kwasilewski:
Hi @jl178! 

It's not on our roadmap at the moment, but I bumped this feature's priority on our internal list to ensure it gets a closer look during the next planning round.

jochem725:
@sfc-gh-kwasilewski I'm also interested in this - it'd be very useful to have this to provision Iceberg resources end-to-end with AWS Glue / IAM.

marco-scatassi-nimbus:
@sfc-gh-kwasilewski I follow to say it would be a great feature to have.

DiogoReisPinto:
@sfc-gh-kwasilewski Also looking for this - Any update on wether this made the roadmap or not?

sfc-gh-kwasilewski:
@jochem725 @marco-scatassi-nimbus @DiogoReisPinto 

Support for catalog-linked databases is still not yet on our roadmap. That said, we are currently working on adding support for Iceberg tables, which is a closely related area.

We'll also be publishing an updated roadmap soon, so please keep an eye out for it for the latest on what's planned and prioritized.

[...]

PR #5168, opened 2026-08-24
chore: Add SDK integration tests for catalog-linked databases

Benjamin0313:
## Summary

SDK integration tests for catalog-linked databases — step (2) of the [approved plan](https://github.com/snowflakedb/terraform-provider-snowflake/issues/4201#issuecomment-4951657803), following up on #5027.

- `TestInt_DatabasesCatalogLinked_WithAdditionalDependencies`: `CreateCatalogLinked` (basic + complete with all options), every `AlterCatalogLinked` action, rename / comment through the generic `Alter`, and both system functions (`SYSTEM$GET_CATALOG_LINKED_DATABASE_CONFIG` for read-back verification, `SYSTEM$CATALOG_LINK_STATUS` polled with `EventuallyWithT`)
- The tests need a preconfigured external Iceberg REST catalog, so they are gated on two new env variables (`TEST_SF_TF_CATALOG_LINKED_DATABASE_CATALOG_INTEGRATION` / `..._EXTERNAL_VOLUME`) via `testenvs.GetOrSkipTest` and skip without them
- New `CreateCatalogLinkedDatabaseWithRequest` helper in `pkg/acceptance/helpers`

Intentionally not asserted: `catalog_name` (fixture-dependent), `is_suspended` (SUSPEND/RESUME DISCOVERY is out of scope), and the failure-path fields of `SYSTEM$CATALOG_LINK_STATUS` (require a broken link; parsing is unit-tested).

## Test Plan
* [x] build, vet (all build-tag combinations), and `pkg/sdk` unit tests pass
* [ ] live run — I have no external Iceberg REST catalog fixture; happy to iterate if a run on your side surfaces issues

## Questions
* Do you have a fixture on your side these could run against (and would you prefer different env variable names)?
* Is env-var gating over the iceberg tests' unconditional `t.Skip` the right direction here? If so, migrating `TestInt_IcebergTables_WithAdditionalDependencies` to the same mechanism could be a small follow-up.

## References
* Part of #4201 · follow-up to #5027
* [CREATE DATABASE (catalog-linked)](https://docs.snowflake.com/en/sql-reference/sql/create-database-catalog-linked) / [ALTER DATABASE (catalog-linked)](https://docs.snowflake.com/en/sql-reference/sql/alter-database-catalog-linked)


pkg/sdk/testint/databases_integration_test.go
```diff
@@ -937,3 +938,193 @@ func TestInt_DatabasesDescribe(t *testing.T) {
...
+
+	// Shared database for the subtests that only read it or fully revert their changes.
+	sharedDatabase := createCatalogLinkedDatabase(t)
+
+	t.Run("create: basic", func(t *testing.T) {
+		assertThatObject(t, objectassert.DatabaseFromObject(t, sharedDatabase).
+			HasKind(sdk.DatabaseKindCatalogLinkedDatabase))
+
+		config := getConfig(t, sharedDatabase.ID())
+		assert.Equal(t, restCatalogId, config.CatalogIntegration)
```
sfc-gh-kwasilewski: Please use generated `objectassert` assertions here and in every other test

pkg/sdk/testint/databases_integration_test.go
```diff
@@ -937,3 +938,193 @@ func TestInt_DatabasesDescribe(t *testing.T) {
...
+		require.NotNil(t, config.ExternalVolume)
+		assert.Equal(t, externalVolumeId, *config.ExternalVolume)
+		assert.Empty(t, config.AllowedNamespaces)
+		assert.Empty(t, config.BlockedNamespaces)
```
sfc-gh-kwasilewski: We should assert all properties and parameters, not only these we explicitly set.

pkg/sdk/testint/databases_integration_test.go
```diff
@@ -937,3 +938,193 @@ func TestInt_DatabasesDescribe(t *testing.T) {
...
+
+		err = client.Databases.Alter(ctx, sdk.NewAlterDatabaseRequest(newName).WithUnset(*sdk.NewDatabaseUnsetRequest().WithComment(true)))
+		require.NoError(t, err)
+
+		unset, err := client.Databases.ShowByID(ctx, newName)
+		require.NoError(t, err)
+		assert.Empty(t, unset.Comment)
+	})
+
+	t.Run("get catalog link status", func(t *testing.T) {
```
sfc-gh-kwasilewski, line 1162: We're missing tests for the `GetCatalogLinkedDatabaseConfig` system function

pkg/sdk/testint/databases_integration_test.go
Benjamin0313: added generated asserts for both system-function outputs (`CatalogLinkedDatabaseConfig`, `CatalogLinkStatus`) and switched every object-state assertion in the test to `objectassert` / `objectparametersassert`. The only raw testify left is operation-success checks and the `EventuallyWithT` polling closure, which cannot take the assert builders.

pkg/sdk/testint/databases_integration_test.go
Benjamin0313: the complete-create test and the new dedicated config subtest now assert all 11 config fields, and the SHOW output is asserted field by field. Two notes:
- Intentionally left out: `is_current`, `origin`, `options`, `retention_time`, `resource_group` (session/account-dependent) and `lastLinkAttemptStartTime` (timing-dependent). For database parameters only `EXTERNAL_VOLUME` is asserted — `HasAllDefaults()` doesn't fit since the external volume is set, and without a live catalog I can't tell whether the linked catalog shows up in the `CATALOG` parameter. Happy to extend after a live run.
- The unset defaults in the config subtest follow the docs (`ALLOWED_WRITE_OPERATIONS = ALL`, no flatten delimiter in `IGNORE_NESTED_NAMESPACE` mode, `SYNC_INTERVAL_SECONDS = 30`); they may need adjustment once the tests run against a real external catalog.

pkg/sdk/testint/databases_integration_test.go
Benjamin0313: added a dedicated `get catalog linked database config` subtest asserting the full config (including documented defaults for everything not set at creation), next to the existing link-status one.

pkg/sdk/testint/databases_integration_test.go
```diff
@@ -937,3 +939,227 @@ func TestInt_DatabasesDescribe(t *testing.T) {
...
+	getConfig := func(t *testing.T, id sdk.AccountObjectIdentifier) *sdk.CatalogLinkedDatabaseConfig {
+		t.Helper()
+		config, err := client.SystemFunctions.GetCatalogLinkedDatabaseConfig(ctx, id)
+		require.NoError(t, err)
+		require.NotNil(t, config)
+		return config
+	}
+
+	// Shared database for the subtests that only read it or fully revert their changes.
+	sharedDatabase := createCatalogLinkedDatabase(t)
```
sfc-gh-jmichalak: Please don't use a shared object like this; instead create it in each test.

pkg/sdk/testint/databases_integration_test.go
sfc-gh-jmichalak: `get catalog linked database config` should be merged with "basic" test.

pkg/sdk/testint/databases_integration_test.go
```diff
@@ -937,3 +939,227 @@ func TestInt_DatabasesDescribe(t *testing.T) {
...
+
+	restCatalogId := sdk.NewAccountObjectIdentifier(testenvs.GetOrSkipTest(t, testenvs.CatalogLinkedDatabaseCatalogIntegration))
+	externalVolumeId := sdk.NewAccountObjectIdentifier(testenvs.GetOrSkipTest(t, testenvs.CatalogLinkedDatabaseExternalVolume))
+
+	createCatalogLinkedDatabase := func(t *testing.T) *sdk.Database {
+		t.Helper()
+		id := testClientHelper().Ids.RandomAccountObjectIdentifier()
+		database, databaseCleanup := testClientHelper().Database.CreateCatalogLinkedDatabaseWithRequest(t, sdk.NewCreateCatalogLinkedDatabaseRequest(id).
+			WithLinkedCatalog(*sdk.NewLinkedCatalogRequest().WithCatalog(restCatalogId)).
+			WithExternalVolume(externalVolumeId))
```
sfc-gh-jmichalak: Nit: we should also test the variant without ExternalVolume (can be done as a follow-up).

pkg/sdk/testint/databases_integration_test.go
```diff
@@ -937,3 +939,227 @@ func TestInt_DatabasesDescribe(t *testing.T) {
...
+// REST catalog integration and external volume, which we can't provision dynamically for now.
+// TODO(SNOW-3725859): Provide them dynamically and move these tests to the main test suite.
+func TestInt_DatabasesCatalogLinked_WithAdditionalDependencies(t *testing.T) {
+	client := testClient(t)
+	ctx := testContext(t)
+
+	restCatalogId := sdk.NewAccountObjectIdentifier(testenvs.GetOrSkipTest(t, testenvs.CatalogLinkedDatabaseCatalogIntegration))
+	externalVolumeId := sdk.NewAccountObjectIdentifier(testenvs.GetOrSkipTest(t, testenvs.CatalogLinkedDatabaseExternalVolume))
+
+	createCatalogLinkedDatabase := func(t *testing.T) *sdk.Database {
```
sfc-gh-jmichalak: Nit: this func should be moved to `helpers` package, as a Database's method.

pkg/sdk/testint/databases_integration_test.go
```diff
@@ -937,3 +939,227 @@ func TestInt_DatabasesDescribe(t *testing.T) {
...
+				WithSyncIntervalSeconds(120).
+				WithAllowedWriteOperations(sdk.CatalogLinkedDatabaseAllowedWriteOperationsNone)))
+		require.NoError(t, err)
+
+		assertThatObject(t, objectassert.CatalogLinkedDatabaseConfigFromObject(t, getConfig(t, database.ID())).
+			HasSyncIntervalSeconds(120).
+			HasAllowedWriteOperations(sdk.CatalogLinkedDatabaseAllowedWriteOperationsNone))
+	})
+
+	t.Run("alter: rename and comment through the generic alter", func(t *testing.T) {
```
sfc-gh-jmichalak, line 1133: Some fields are still missing from tests:
```sql
ALTER DATABASE [ IF EXISTS ] <name> SET [ BASE_LOCATION_PREFIX = '<string>' ]
                                        [ COMMENT = '<string_literal>' ]
                                        [ CONTACT <purpose> = <contact_name> [ , <purpose> = <contact_name> ... ] ]
                                        [ ICEBERG_VERSION_DEFAULT = <integer> ]
                                        [ ICEBERG_MERGE_ON_READ_BEHAVIOR = { 'AUTO' | 'ENABLED' | 'DISABLED' } ]
                                        [ ENABLE_ICEBERG_MERGE_ON_READ = { TRUE | FALSE } ]

ALTER DATABASE [ IF EXISTS ] <name> UNSET { BASE_LOCATION_PREFIX         |
                                            COMMENT                      |
                                            CONTACT                      |
                                            ICEBERG_VERSION_DEFAULT      |
                                            ICEBERG_MERGE_ON_READ_BEHAVIOR |
                                            ENABLE_ICEBERG_MERGE_ON_READ
                                          }
```
They are also missing from the SDK so they can be added in the next PR.

pkg/sdk/testint/databases_integration_test.go
sfc-gh-jmichalak: Nit: in general we also assert all the parameters (besides session-dependent), even in "basic" scenario". Let's add the remaining ones - and we can skip the ones you listed.

sfc-gh-jmichalak (commented):
LGTM with a couple of requests that can be done in a follow-up.

sfc-gh-jmichalak (approved):

pkg/acceptance/helpers/database_client.go
```diff
@@ -87,6 +87,32 @@ func (c *DatabaseClient) CreateDatabaseWithRequest(t *testing.T, request *sdk.Cr
...
+	ctx := context.Background()
+
+	id := request.ID()
+	err := c.client().CreateCatalogLinked(ctx, request)
+	require.NoError(t, err)
+	cleanup := c.DropDatabaseFunc(t, id)
+
+	database, err := c.client().ShowByID(ctx, id)
+	if err != nil {
+		// Register the cleanup here so a failing ShowByID does not leak the database.
```
sfc-gh-jmichalak, line 108: Nit: there is no need because we return the cleanup function anyway. It should be called outside.

pkg/sdk/testint/databases_integration_test.go
sfc-gh-jmichalak: Also, please bundle the changes so we for example test SET on all fields at the same time, to reduce test running time.

pkg/acceptance/helpers/database_client.go
Benjamin0313: Done — removed the guard; the helper now matches its siblings and returns the cleanup for the caller to register.

pkg/sdk/testint/databases_integration_test.go
Benjamin0313: Done — the three ALTER subtests are now one `alter: update linked catalog`, applying ADD for both namespace lists plus SET for all fields before a single assertion, then REMOVE for both, then UNSET for both. That takes the databases created against the external catalog from 7 down to 5 per run (each ALTER still needs its own statement since only one action is accepted at a time).

sfc-gh-kwasilewski (approved):

sfc-gh-kwasilewski:
Hi @Benjamin0313. Thanks for the integration tests, this is looking good and we’re close to merging it.

With this PR wrapping up, I wanted to check in on the next step from the plan: the snowflake_catalog_linked_database resource. No rush at all, but it would help us with planning if you could share when you expect to start on it and roughly what your timeline looks like.

Benjamin0313:
@sfc-gh-kwasilewski san
The resource PR is up: #5216. It follows the scope we agreed in #4201 — no data source, and contact/tag/suspend-resume plus the ALTER-only parameters left for follow-ups.

One heads-up: it includes a small SDK fix, since SYSTEM$GET_CATALOG_LINKED_DATABASE_CONFIG returns the namespaces single-quoted and the read-back never matched the configuration without it.

Issue #5208, opened 2026-09-15
[Feature]: snowflake_cortex_search_service resource is missing PRIMARY KEY and AUTO_SUSPEND configuration


allllllllez:
### Use Cases or Problem Statement

`CREATE CORTEX SEARCH SERVICE` and `ALTER CORTEX SEARCH SERVICE` support `PRIMARY KEY` and `AUTO_SUSPEND`, but the `snowflake_cortex_search_service` Terraform resource does not expose either attribute. The SDK layer (`pkg/sdk/cortex_search_services_gen.go`) already builds this SQL, but `pkg/resources/cortex_search_service.go` does not surface it in the resource schema.

### Category

category:resource, category:preview

### Object type(s)

resource:cortex_search_service

### Proposal

Add two optional, `ForceNew` attributes to `snowflake_cortex_search_service`:

- `primary_key` (list/set of string) — maps to `PRIMARY KEY (...)` in `CREATE CORTEX SEARCH SERVICE`.
- `auto_suspend` (number, with `IntDefault`-style handling) — maps to `AUTO_SUSPEND` in both `CREATE` and `ALTER ... SET`.

Note: per an existing TODO in `pkg/sdk/generator/defs/cortex_search_services_def.go`, `AUTO_SUSPEND` (and `REQUEST_LOGGING`) is not exposed in `SHOW`/`DESCRIBE` output, so drift detection for `auto_suspend` will be limited to what Terraform itself tracks in state.

### Would you like to implement a fix?

- [x] Yeah, I'll take it 😎

sfc-gh-kwasilewski:
Hi @allllllllez,

Sorry for the delayed response.

Thanks for picking this up in #5209. I have reviewed the change and left comments on the PR. Please continue the discussion there.

PR #5209, opened 2026-09-15
feat: Add primary_key and auto_suspend support to snowflake_cortex_search_service

allllllllez:
## Summary
- Add `primary_key` and `auto_suspend` to `snowflake_cortex_search_service`. The SDK already generates this SQL; the resource schema never exposed it.
- Expose `primary_key_columns` in `describe_output` (parsed by the SDK, dropped by the schema mapper).
- Resetting `auto_suspend` uses a separate `ALTER ... SET AUTO_SUSPEND = NULL` call, since `SET` and `SET DEFAULTS` are mutually exclusive in Snowflake's grammar.
- Regenerated assertions, config model builders, and docs. Added a Migration Guide entry and extended the acceptance test.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run` (0 issues on touched packages)
- [x] `make test-unit` (green on touched packages; pre-existing unrelated failure in `pkg/internal/oswrapper`)
- [x] Manually verified against a live Snowflake account: create, reset-to-default, re-set, and delete of both `primary_key` and `auto_suspend` all succeeded

## References

- #5208


pkg/resources/cortex_search_service.go
```diff
@@ -42,6 +43,13 @@ var cortexSearchServiceSchema = map[string]*schema.Schema{
 		Description: "Specifies the column to use as the search column for the Cortex search service; must be a text value.",
 		ForceNew:    true,
 	},
+	"primary_key": {
+		Type:        schema.TypeSet,
+		Optional:    true,
+		Elem:        &schema.Schema{Type: schema.TypeString},
+		Description: "Specifies the column(s) in the base table that uniquely identify each row, used to enable optimized (incremental) refreshes when the underlying data changes. All primary key columns must use the TEXT data type.",
+		ForceNew:    true,
```
sfc-gh-kwasilewski: Snowflake allows modifying the set of primary key columns via ALTER SET and ALTER UNSET, both of which are supported by our SDK. Please implement an update operation for this attribute.

pkg/resources/cortex_search_service.go
```diff
@@ -72,6 +80,13 @@ var cortexSearchServiceSchema = map[string]*schema.Schema{
 		Optional:    true,
 		Description: "Specifies a comment for the Cortex search service.",
 	},
+	"auto_suspend": {
+		Type:             schema.TypeInt,
+		Optional:         true,
+		Default:          IntDefault,
```
sfc-gh-kwasilewski: We only need `IntDefault` if 0 is a valid value for a property on Snowflake side.  Please confirm and remove it if needed.

pkg/resources/cortex_search_service.go
```diff
@@ -72,6 +80,13 @@ var cortexSearchServiceSchema = map[string]*schema.Schema{
...
+		ValidateDiagFunc: validation.ToDiagFunc(validation.IntAtLeast(1800)),
```
sfc-gh-kwasilewski: We avoid strict client-side validations here. If Snowflake updates its allowed value ranges in the future, any released version of the provider with hardcoded limits would immediately break. Please change it to:

```go
validation.ToDiagFunc(validation.IntAtLeast(0))
```

or

```go
validation.ToDiagFunc(validation.IntAtLeast(1))
```

if you confirm that 0 is an invalid value.

pkg/resources/cortex_search_service.go
```diff
@@ -72,6 +80,13 @@ var cortexSearchServiceSchema = map[string]*schema.Schema{
...
+		Description:      "Specifies the number of seconds of inactivity after which the Cortex search service automatically suspends its serving compute, releasing the resources. Minimum value is 1800 (30 minutes). If unset, Snowflake defaults to no automatic suspension.",
```
sfc-gh-kwasilewski: Please use `externalChangesNotDetectedFieldDescription` doc helper.

MIGRATION_GUIDE.md
```diff
@@ -24,6 +24,17 @@ for changes required after enabling given [Snowflake BCR Bundle](https://docs.sn
 > [!TIP]
 > If you're still using the `Snowflake-Labs/snowflake` source, see [Upgrading from Snowflake-Labs Provider](./SNOWFLAKEDB_MIGRATION.md) to upgrade to the snowflakedb namespace.
 
+## v2.21.x ➞ v2.22.0
```
sfc-gh-kwasilewski, line 27: Please pull in newest changes from dev; there's already a v2.22.0 section in the guide.

pkg/resources/cortex_search_service.go
```diff
@@ -252,12 +267,18 @@ func CreateCortexSearchService(ctx context.Context, d *schema.ResourceData, meta
 	if v, ok := d.GetOk("embedding_model"); ok {
 		request.WithEmbeddingModel(v.(string))
 	}
+	if v, ok := d.GetOk("primary_key"); ok && len(v.(*schema.Set).List()) > 0 {
```
sfc-gh-kwasilewski, line 286: Please implement drift detection for this attribute (you need to update the Read function).

pkg/resources/cortex_search_service.go
allllllllez: Fixed in 1403b9390.

Removed `ForceNew` from `primary_key` and implemented the update path. Snowflake exposes `SET PRIMARY KEY (...)` and `UNSET PRIMARY KEY` as dedicated clauses that are mutually exclusive with the generic `SET`, so the update issues them as a separate `ALTER`. `describe_output` is now recomputed when `primary_key` changes, and the acceptance test covers update, unset, and re-set in place.

pkg/resources/cortex_search_service.go
allllllllez: Fixed in 1403b9390.

Confirmed that 0 is not valid on the Snowflake side: `AUTO_SUSPEND` requires a minimum of 1800 seconds, or `NULL` to disable auto-suspend. The zero value is therefore enough to mean unset, so `Default: IntDefault` is removed and create/update now use `d.GetOk` and `intAttributeUpdate` respectively.

pkg/resources/cortex_search_service.go
allllllllez: Fixed in 1403b9390.

Changed to `validation.ToDiagFunc(validation.IntAtLeast(1))`, since 0 is invalid on the Snowflake side. The 1800 minimum is no longer enforced client-side; it is only mentioned in the field description and left to Snowflake to validate.

pkg/resources/cortex_search_service.go
allllllllez: Fixed in 1403b9390.

The description now goes through `externalChangesNotDetectedFieldDescription`. That is accurate here: Snowflake returns `auto_suspend` on neither `SHOW CORTEX SEARCH SERVICES` nor `DESCRIBE CORTEX SEARCH SERVICE`, so drift on this field cannot be detected.

MIGRATION_GUIDE.md
allllllllez: Fixed in 7ce7c40c0, which merges the latest dev, so the entry now sits under the existing v2.22.0 section instead of introducing a second one.

1403b9390 also reworks the wording, because `primary_key` is no longer recreated on change and `auto_suspend` drift is not detected.

pkg/resources/cortex_search_service.go
allllllllez: Fixed in 1403b9390.

`Read` now marks external changes to `primary_key` by comparing against `describe_output.primary_key_columns`, using the deep-equal variant of the handler since the value is a list. `ImportCortexSearchService` sets the field as well, so `primary_key` could be dropped from `ImportStateVerifyIgnore` in the acceptance test.

sfc-gh-kwasilewski (approved):

sfc-gh-kwasilewski:
@allllllllez Could you solve merge conflicts? I attempted to import the PR, and even though it said it succeeded, the result wasn't quite right.

m-kajiya-datum:
@sfc-gh-kwasilewski 
Sorry, I ran into a conflict after updating the branch... I'll fix it and let you know!

allllllllez:
@sfc-gh-kwasilewski
I've fixed it! Please take care of the follow-up tasks!

